PR #23, opened 2021-05-07
entire review the parameters for the query

Nexucis:
The new format can be : 

```yaml
kind: "Query"
parameter:
  label_names:
    matchers:
    - "up"
  capturing_regexp: ".*"
```
or

```yaml
kind: "Query"
parameter:
 label_values:
   label_name: "instance"
   matchers:
   - "up"
 capturing_regexp: ".*"
```

or 

```yaml
kind: "Query"
parameter:
  expr: "up{instance='localhost:8080'}"
  capturing_regexp: ".*"
```

Each possibility is related to way to get the data from Prometheus (aka or by getting the label names, or by getting the label values for a specific label_name, or by performing a PromQL Query)


Nexucis:
Note: the usage of `matchers` is optional in `label_names` or in `label_values`

PR #43, opened 2021-05-27
[FRONT] variables management for the dashboard

Nexucis:
This PR is providing the way to manage the variables (readonly) in the dashboard.
It should be merged after #41 because it requires to be integrated with panels management.

here a short video of how it looks like on a medium  screen.

https://user-images.githubusercontent.com/4548045/119805199-55b86900-bee1-11eb-96fd-e83725ebae30.mp4

@AntoineThebaud maybe it can help you for the creation of the component `dashboard-sections`
@celian-garcia FYI.


internal/api/front/perses/src/app/project/prometheusrule/prometheusrule.module.ts
```diff
@@ -15,20 +15,16 @@ import { NgModule } from '@angular/core';
...
 import { MatExpansionModule } from '@angular/material/expansion';
-import { PageModule } from '../../shared/component/page/page.module';
 import { PrometheusruleLayoutComponent } from './prometheusrule-layout/prometheusrule-layout.component';
 import { PrometheusruleDetailsComponent } from './prometheusrule-details/prometheusrule-details.component';
 
 @NgModule({
   declarations: [PrometheusRuleListComponent, PrometheusruleLayoutComponent, PrometheusruleDetailsComponent],
   imports: [
-    MatDividerModule,
     MatExpansionModule,
```
Nexucis, line 25: ah I think I can move the `MatExpansionModule` to the `SharedModule` since it is used by two different module now.

Nexucis:
lol it make me think also it will be cool to have something that would deploy the PR so we can play with UI.

Will look into that if it's easy to do. Prometheus has something like that, so maybe we can do the same.

Nexucis:
mmm so Netlify is not able to build the backend. So I will have to find a way to deploy the backend somewhere and then we will be able to use Netlify to run the front-end.

Nexucis:
@celian-garcia  @AntoineThebaud should be good now for a review :).
It still remain some optimization, but I believe it can be done in another PR.

Tell me if you want to review the PR or not. I can understand you don't have so much time to look at it.

Nexucis:
This PR is also integrated the usage of the variable in the dashboard-section. And I also took sometime to fix some issue / code it was easy to solve in the dashboard-section component.

For example, the component were subscribing to the `dashboardFeedService` without destroying these subsribes when the component is destroyed.

AntoineThebaud (approved):

PR #51, opened 2021-06-07
upgrade angular to v12

Nexucis:


Nexucis:
through this PR, the styles had to move outside of the assets folder. More info here: https://github.com/angular/angular-cli/issues/20812#issuecomment-843861249

PR #62, opened 2021-08-05
Add initial React app stub

juliusv:
Initially created by:

```bash
npx create-react-app web --template typescript
```

...and then adjusted.

It doesn't do anything yet besides rendering an empty "Perses" app bar with Material UI.

To start (on http://localhost:3000/):

```bash
npm i
npm run start
```


web/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "private": true,
+  "dependencies": {
+    "@material-ui/core": "^4.12.3",
+    "@types/node": "^12.0.0",
+    "@types/react": "^17.0.15",
+    "@types/react-dom": "^17.0.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-scripts": "4.0.3",
+    "typescript": "^4.1.2",
```
Nexucis, line 13: let's put directly typescript a devDependencies instead :)

web/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "perses",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@material-ui/core": "^4.12.3",
+    "@types/node": "^12.0.0",
+    "@types/react": "^17.0.15",
+    "@types/react-dom": "^17.0.0",
```
Nexucis, line 9: Hopefully the @types should also be part of the devDependencies.

Nexucis (commented):
Some comments regarding the different dependencies :).

Would be cool to have the CI to build the react-app. But if you want I can do it in another PR if you don't want to be bothered by that :).

Wondering how you think to serve the webapp through `echo` (the http framework), since it's quite outside of the `internal` package.

Also any reason to not put it in the `internal` package ?

Thanks for initializing that !!

web/README.md
```diff
@@ -0,0 +1,46 @@
+# Getting Started with Create React App
```
Nexucis, line 1: do we want to keep the generated readme ?

web/package.json
juliusv: Ok yeah, `create-react-app` just doesn't care much about `devDependencies` and puts everything under `dependencies` by default, but will change.

web/README.md
juliusv: I can nuke it.

juliusv:
> Would be cool to have the CI to build the react-app. But if you want I can do it in another PR if you don't want to be bothered by that :).

Later :)

> Wondering how you think to serve the webapp through echo (the http framework), since it's quite outside of the internal package. (It's ok to think about it later)

I haven't looked into that yet, I was planning on using https://github.com/rakyll/statik to bake the React assets into the Go code.

> Also any reason to not put it in the internal package ?

IMO `internal` should only be for Go packages that are intentionally hidden for outside consumers, I prefer keeping the web frontend code outside of that and ideally easily accessible from the repo root.

juliusv:
@Nexucis An unrelated Go test is failing:

```
=== RUN   TestUpdateDatasource
    datasource_test.go:122: 
        	Error Trace:	datasource_test.go:122
        	Error:      	Received unexpected error:
        	            	invalid character '}' after top-level value
        	Test:       	TestUpdateDatasource
--- FAIL: TestUpdateDatasource (0.00s)
```

Nexucis:
> I haven't looked into that yet, I was planning on using https://github.com/rakyll/statik to bake the React assets into the Go code.

ah yeah interesting. For Angular I use a code generation, but certainly better to not use an artisanal way.

> IMO internal should only be for Go packages that are intentionally hidden for outside consumers, I prefer keeping the web frontend code outside of that and ideally easily accessible from the repo root.

Like I said in a previous conversation, I have been told it was painful to have the UI so deep in the arbo. And on my side I wasn't so confortable to have it in the `internal` package etheir. So yeah totally fine with that. I was just curious :).

> @Nexucis An unrelated Go test is failing

yeah that's odd. I just rerun the pipeline and it works. Will have to find what happens one day.

Nexucis (approved):
thanks for the PR !

juliusv:
Thanks for merging!

> Like I said in a previous conversation, I have been told it was painful to have the UI so deep in the arbo. And on my side I wasn't so confortable to have it in the internal package etheir. So yeah totally fine with that. I was just curious :).

Yes, and I would also not see this primarily as a Go repo, but primarily as a web / React repo, with some Go pieces to support the web application :)

Nexucis:
> Yes, and I would also not see this primarily as a Go repo, but primarily as a web / React repo, with some Go pieces to support the web application :)

Haha I will have a different opinion here. But I think for the moment it doesn't really matter :)

PR #63, opened 2021-08-06
remove prometheus rule management

Nexucis:
This PR is removing the PrometheusRules management as it has nothing related to the dashboard features.

/cc @juliusv


juliusv:
:+1:

PR #65, opened 2021-08-09
replace Amadeus copyright by Perses Authors

Nexucis:


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2021 Amadeus s.a.s
+   Copyright [yyyy] The Perses Authors
```
juliusv, line 189: I'm not sure that the LICENSE itself should mention the copyright owners at all, as these are just general instructions on how to apply the Apache 2 license. See also: https://github.com/prometheus/prometheus/blob/ee7e0071d10e795cb78a8e50764e45cbcaf8c29a/LICENSE#L189

juliusv:
:+1: otherwise

PR #66, opened 2021-08-09
Remove angular app and enable react build

Nexucis:


Nexucis:
...
I should have deleted these files before having changed the copyright

juliusv:
:+1:

Issue #55, opened 2021-06-09
Layout managing / review the datamodel of the dashboard


Nexucis:
The idea is to think about how we can integrate a layout management and probably a better way to manage the different panels/ components.

Instead of considering you have different sections and different components, maybe it could be interesting to considere everything as a component. Like that you can manage of course different chart, but also different layout. Like a grid, an expand component ...etc.

It's also the opportunity to introduce a way to reference different panels that has been created in others projects.

Based on this different output, here a first draft of a new datamodel for a dashboard:

```json
{
  "kind": "Dashboard",
  "metadata": {
    "name": "NodeExporter",
    "project": "perses"
  },
  "spec": {
    "datasource": "PrometheusDemo",
    "duration": "6h",
    "variables": {
      "job": {
        "kind": "LabelValuesQuery",
        "hide": false,
        "parameter": {
          "label_name": "job",
          "matchers": [
            "node_uname_info"
          ],
          "capturing_regexp": "(.*)"
        }
      },
      "node": {
        "kind": "LabelValuesQuery",
        "hide": false,
        "parameter": {
          "label_name": "instance",
          "matchers": [
            "node_uname_info{job=\"$job\"}"
          ],
          "capturing_regexp": "(.*)"
        }
      }
    },
    "panels": [
      {
        "kind": "Panel",
        "metadata": {
          "name": "myAwesomePanel"
        },
        "spec": {
          "kind": "LineChart",
          "chart": {
            "show_legend": true,
            "lines": [
              {
                "expr": "up",
                "legend": "{{job}}[{{instance}}]"
              }
            ]
          }
        }
      },
      {
        "kind": "Panel",
        "metadata": {
          "name": "CPU Busy"
        },
        "spec": {
          "kind": "GaugeChart",
          "chart": {
            "expr": "(((count(count(node_cpu_seconds_total{instance=\"$node\",job=\"$job\"}) by (cpu))) - avg(sum by (mode)(rate(node_cpu_seconds_total{mode='idle',instance=\"$node\",job=\"$job\"}[5m])))) * 100) / count(count(node_cpu_seconds_total{instance=\"$node\",job=\"$job\"}) by (cpu))"
          }
        }
      },
      {
        "kind": "Panel",
        "metadata": {
          "name": "Sys Load (5m avg)"
        },
        "spec": {
          "kind": "GaugeChart",
          "chart": {
[...]

PR #69, opened 2021-08-11
Implement the layout on the backend side

Nexucis:
this PR adds the datamodel for the Dashboard Layout on the backend side. (related to #55)

It checks that the json references are pointing to an existing object.

Note: I'm not sure it will be useful in the future to set the pointer in the JSONRef. I did it today, probably this set won't be useful at all

### DIFF
--- a/pkg/model/api/v1/jsonref.go
+++ b/pkg/model/api/v1/jsonref.go
@@ -19,7 +19,12 @@ import (
 	"regexp"
 )
 
-var jsonRefRegexp = regexp.MustCompile(`^#/([a-zA-Z0-9_-]+)(?:/([a-zA-Z0-9_-]+))*$`)
+var (
+	// jsonRefMatching is only used to validate the whole reference.
+	jsonRefMatching = regexp.MustCompile(`^#?/([a-zA-Z0-9_-]+)(?:/([a-zA-Z0-9_-]+))*$`)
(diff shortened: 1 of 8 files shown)


pkg/model/api/v1/jsonref.go
```diff
@@ -0,0 +1,69 @@
...
+
+package v1
+
+import (
+	"encoding/json"
+	"fmt"
+	"regexp"
+)
+
+var jsonRefRegexp = regexp.MustCompile(`^#/([a-zA-Z0-9_-]+)(?:/([a-zA-Z0-9_-]+))*$`)
```
Nexucis, line 22: well looks like I have an issue with the question `How to capture multiple repeated groups`.
Quite interesting actually, but not really how to solve it.

The problem currently is that:

for the reference `#/spec/panels/name` it will capture `spec` and `name` not the `panels`. Which looks like a common mistake called `Repeating a Capturing Group vs. Capturing a Repeated Group`

Nexucis: mmm it's not solvable like that, but I could actually simplify the thing and it should solve the issue.

Nexucis: voilà should be better now

Nexucis:
merging this PR. Changes can be performed if there is anything odd in what I dit

PR #73, opened 2021-08-19
upgrade golang and golangci-lint to last version

Nexucis:


juliusv:
Go 1.17 now has some Go code formatting diffs (see failing tests). Specifically, the formatting for build tags has changed:

> Go versions 1.16 and earlier used a different syntax for build constraints, with a "// +build" prefix. The gofmt command will add an equivalent //go:build constraint when encountering the older syntax.

(from https://pkg.go.dev/cmd/go#hdr-Build_constraints)

Nexucis:
Indeed I saw it too. I just didn't have the time to fix it. I will do it when coming back. Or if someone is eager to do it, we can close it in favor of the new fresh working one.

PR #79, opened 2021-09-08
implement a proxy middleware that redirect requests to the datasource

Nexucis:
For the moment it redirects all request to the datasource.

HTTP method whitelist will come later when redoing the datamodel of the  datasource.

### How to use it

* standard usage, target a known datasource and used correctly the api of the datasource

```bash
curl -XGET http://localhost:8080/proxy/datasources/PrometheusDemo/api/v1/labels

{
"status": "success",
"data": [
"__name__",
"action",
"address",
"alertmanager"
]
}
```

* target a not known datasource

```bash
curl -XGET http://localhost:8080/proxy/datasources/test/api/v1/labels

{
"message": "unable to forward the request to the datasource 'test', datasource doesn't exist"
}
```

it returns the 502 error (bad gateway)

* target a known datasource, but with a wrong usage of the API. It returns the error from the datasource

```bash
curl -XGET http://localhost:8080/proxy/datasources/PrometheusDemo/foo

404 page not found
```

### DIFF
--- a/internal/api/core/middleware/proxy.go
+++ b/internal/api/core/middleware/proxy.go
@@ -70,6 +70,8 @@ func buildProxy(c echo.Context, dao datasource.DAO) (*url.URL, error) {
 		logrus.WithError(err).Errorf("unable to find the datasource '%s', something wrong with the database", datasourceName)
 		return nil, echo.NewHTTPError(http.StatusInternalServerError, "internal server error")
 	}
+	// Based on the HTTP 1.1 RFC, a `/` should be the minimum path.
+	// https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2
 	path := "/"
 	if len(matchingGroups[0]) > 1 {
 		path = matchingGroups[0][2]
--- a/internal/api/core/middleware/proxy.go
+++ b/internal/api/core/middleware/proxy.go
@@ -1,0 +1,118 @@
+package middleware
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"regexp"
+	"time"
+
+	"github.com/labstack/echo/v4"
+	"github.com/perses/common/etcd"
+	"github.com/perses/perses/internal/api/interface/v1/datasource"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	proxyMatcher                       = regexp.MustCompile(`/proxy/datasources/([a-zA-Z-0-9_-]+)(/.*)`)
+	defaultTransport http.RoundTripper = &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			DualStack: true,
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		TLSHandshakeTimeout: 10 * time.Second,
+		MaxIdleConns:        100,
+		IdleConnTimeout:     90 * time.Second,
+		ForceAttemptHTTP2:   true,
+	}
+)
+
+func Proxy(dao datasource.DAO) echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			targetURL, err := buildProxy(c, dao)
+			if err != nil {
+				return err
+			}
+			if targetURL == nil {
+				return next(c)
+			}
+			return serveProxy(c, targetURL)
+		}
+	}
+}
+
+func buildProxy(c echo.Context, dao datasource.DAO) (*url.URL, error) {
+	req := c.Request()
+	requestPath := req.URL.Path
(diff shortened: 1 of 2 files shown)


internal/api/core/middleware/proxy.go
```diff
@@ -0,0 +1,118 @@
...
+	dts, err := dao.Get(datasourceName)
+	if err != nil {
+		if etcd.IsKeyNotFound(err) {
+			logrus.Debugf("unable to find the Datasource '%s'", datasourceName)
+			return nil, echo.NewHTTPError(http.StatusBadGateway, fmt.Sprintf("unable to forward the request to the datasource '%s', datasource doesn't exist", datasourceName))
+		}
+		logrus.WithError(err).Errorf("unable to find the datasource '%s', something wrong with the database", datasourceName)
+		return nil, echo.NewHTTPError(http.StatusInternalServerError, "internal server error")
+	}
+	var path string
```
robskillington, line 73: nit: Maybe path should be `/` if there is nothing passed on the URL params?

internal/api/core/middleware/proxy.go
```diff
@@ -0,0 +1,118 @@
...
+	proxy.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, err error) {
+		desc := targetURL.String()
+		msg := fmt.Sprintf("remote %s is unreachable, could not forward: %v", desc, err)
+		logrus.Errorf(msg)
+		c.Set("_error", echo.NewHTTPError(http.StatusBadGateway, msg))
+	}
+	// use the dedicated HTTP transport to avoid any TSL encrypt issue
+	proxy.Transport = defaultTransport
+	return proxy
+}
```
robskillington, line 118: Instead of using the `.Set` and `.Get` for the context using a non-typed string as the key lookup there I'd just inline the proxy request method and use a lambda variable to capture the error. I would also make error wrapping and unwrapping possible you'll want to use `fmt.Errorf` and `%w` instead of `%v` to capture the error:
```suggestion
func serveProxy(c echo.Context, targetURL *url.URL) error {
	req := c.Request()
	res := c.Response()

	// We have to modify the HOST of the request in order to match the host of the targetURL
	// So far I'm not sure to understand exactly why, but if you are going to remove it, be sure of what you are doing.
	// It has been done to fix an error returned by Openshift itself saying the target doesn't exist.
	// Since we are using HTTP/1, setting the HOST is setting also an header so if the host and the header is blocked
	// then maybe it is blocked by the Openshift router.
	req.Host = targetURL.Host
	// Fix header
	if len(req.Header.Get(echo.HeaderXRealIP)) == 0 {
		req.Header.Set(echo.HeaderXRealIP, c.RealIP())
	}
	if len(req.Header.Get(echo.HeaderXForwardedProto)) == 0 {
		req.Header.Set(echo.HeaderXForwardedProto, c.Scheme())
	}
	
	var proxyErr error
	proxy := httputil.NewSingleHostReverseProxy(targetURL)
	proxy.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, err error) {
		desc := targetURL.String()
		proxyErr = fmt.Errorf("error proxying, remote unreachable: target=%s, err=%w", desc, err)
		logrus.Errorf(proxyErr.Error())
		proxyErr = err
	}
	// Use default HTTP transport to proxy with HTTP scheme instead of HTTPS.
	proxy.Transport = defaultTransport
	
	// Reverse proxy request.
	proxy.ServeHTTP(res, req)

	// Return any error handled during proxying request.
	return proxyErr
}
```instead of using instead of using

File: internal/api/core/middleware/proxy.go (hunk 1 above)
Comment (robskillington) [reply internal/api/core/middleware/proxy.go:73]: As per HTTP 1.1 RFC:
```
   Note that the absolute path
   cannot be empty; if none is present in the original URI, it MUST be
   given as "/" (the server root).
```

Relevant section:
https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2

**Review by robskillington — APPROVED** [review]
LGTM other than the minor comments I left

File: internal/api/core/middleware/proxy.go (hunk 2 above)
Comment (Nexucis) [reply internal/api/core/middleware/proxy.go:118]: ah yeah that's simpler. Thanks !

Comment (Nexucis) [reply internal/api/core/middleware/proxy.go:118]:
```go
	// Use default HTTP transport to proxy with HTTP scheme instead of HTTPS.
	proxy.Transport = defaultTransport
```

acutally it can proxy to https no issue on that. `defaultTransport` is actually a custom transport that would avoid the TLS encrypt issue that can happen when the certificate is not trusted.

For the moment it's hardcoded, but I think it's something that will be defined in the datasource configuration.

internal/api/core/middleware/proxy.go
Nexucis: done

internal/api/core/middleware/proxy.go
Nexucis: perhaps it's a bit overkilled to wrappe the proxy error just to log it. But I'm fine for that, it's not a big changes anyway.

PR #81, opened 2021-09-10
Editorial changes to README.md

juliusv:
I changed the language around a lot in some places, but hopefully the
overall result is smoother to read and easier to understand, while
conveying the same meaning.


Nexucis (approved):
it's indeed so much better. Thanks a lot @juliusv

PR #83, opened 2021-09-15
move dashboard spec component in a dedicated package

Nexucis:


Nexucis:
I think like that, it's cleaner, and the name of the different struct are not prefixed by `Dashboard`. Like that the structs are shorter.

juliusv:
:+1:

PR #84, opened 2021-09-16
Reworking the datasource format

Nexucis:
This PR is changing deeply the datamodel of the datasource. The idea is to introduce the concept of multi different type of datasource.

1. Every different datasources will share these attributes : 

```golang
type BasicDatasource struct {
	Kind    Kind `json:"kind" yaml:"kind"`
	Default bool `json:"default" yaml:"default"`
}
```

* `Kind` will give you the type of the datasource.
* `Default` will say if it's the default datasource for the same type. Meaning that you will be able to define one default datasource per type of datasource. One default datasource for Prometheus, one for MySQL ...etc.

2. This PR is also changing the configuration for the Prometheus datasource itself by introducing more complexe HTTP configuration: 

```golang
type HTTPConfiguration struct {
	// URL is the url required to contact the datasource
	URL *url.URL `json:"url" yaml:"url"`
	// The way the UI will contact the datasource. Or through the Backend or directly.
	// By default, Access is set with the value 'server'
	Access HTTPAccess `json:"access,omitempty" yaml:"access,omitempty"`
	// WhiteList is a list of tuple of http method and http endpoint that will be accessible.
	// These parameters are only used when access is set to 'server'
	WhiteList []HTTPWhiteListConfig `json:"white_list" yaml:"white_list"`
	// Auth is holding any security configuration for the http configuration.
	// When defined, it's impossible to set the value of Access with 'browser'
	Auth *HTTPAuth `json:"auth,omitempty" yaml:"auth,omitempty"`
	// Headers can be used to provide additional header that needs to be forwarded when requesting the datasource
	// When defined, it's impossible to set the value of Access with 'browser'
	Headers map[string]string `yaml:"headers,omitempty"`
}
```
As you can see, it introduced:
* an HTTP datasource could be used directly in the UI
* some auth configuration (basic auth, bearer token, ca cert)

The Prometheus Datasource define a default `whiteList` configuration. That's the only thing really specialize for Prometheus itself.

It also introduced the notion of the local datasource. You can define a datasource in your project. The datasource definition is exactly the same as the one you can define globaly. 
Only dashboard in the same project will be able to use it. Dashboards from other project won't be able to use it. Only the global datasources are shared across the different dashboards / projects.

A consequency of changing the spec of the datasource and introducing the local datasource is the recode of the Proxy.
For the moment, we only support Prometheus which exposed an HTTP API, but at some point we likely have not HTTP datasource.
The proxy is taking care of that, by introducing a proxy factory. Depending of the datasource type, an http proxy will be created or something else.

I will add more unit test, but in the meantine I will be interesting by your opinion about these changes @juliusv and @robskillington :).


internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,225 @@
...
-	req := c.Request()
-	requestPath := req.URL.Path
-	if !proxyMatcher.MatchString(requestPath) {
-		// request is not matching the proxy path. Probably it's a request to the API itself
-		// That's why no error is returned in order to let the request passed.
-		return nil, nil
-	}
-	logrus.Tracef("'%s' is a request to a datasource", requestPath)
-	matchingGroups := proxyMatcher.FindAllStringSubmatch(requestPath, -1)
+func catchDatasourceAndPath(c echo.Context, dts datasource.DAO, globalDTS globaldatasource.DAO) (v1.DatasourceSpec, string, error) {
```
juliusv, line 58: `catch` sounds a bit unusual for this kind of operation, how about `extract`?

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,225 @@
...
+	// Set up the proxy
+	var proxyErr error
+	reverseProxy := httputil.NewSingleHostReverseProxy(h.config.URL)
+	reverseProxy.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, err error) {
+		desc := h.config.URL.String()
+		proxyErr = fmt.Errorf("error proxying, remote unreachable: target=%s, err=%w", desc, err)
+		logrus.Errorf(proxyErr.Error())
+		proxyErr = err
+	}
+	// use a dedicated HTTP transport to avoid any TSL encrypt issue
```
juliusv, line 162:
```suggestion
	// use a dedicated HTTP transport to avoid any TLS encryption issues
```

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,225 @@
...
+	// Return any error handled during proxying request.
+	return proxyErr
+}
+
+func (h *httpProxy) prepareRequest(c echo.Context) error {
+	req := c.Request()
 	// We have to modify the HOST of the request in order to match the host of the targetURL
 	// So far I'm not sure to understand exactly why, but if you are going to remove it, be sure of what you are doing.
 	// It has been done to fix an error returned by Openshift itself saying the target doesn't exist.
 	// Since we are using HTTP/1, setting the HOST is setting also an header so if the host and the header are different
```
juliusv, line 175: Aside: In HTTP/1.1 the host is *only* sent in the `Host:` header, so that's why it definitely needs to be set so that the backend router / proxy knows who you want to talk to.

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,225 @@
...
+	req.Host = h.config.URL.Host
 	// Fix header
 	if len(req.Header.Get(echo.HeaderXRealIP)) == 0 {
 		req.Header.Set(echo.HeaderXRealIP, c.RealIP())
 	}
 	if len(req.Header.Get(echo.HeaderXForwardedProto)) == 0 {
 		req.Header.Set(echo.HeaderXForwardedProto, c.Scheme())
 	}
+	// set header according to the configuration
+	if len(h.config.Headers) > 0 {
```
juliusv, line 186: Do we want to allow overriding even special headers like the user agent or the host header here? Maybe fine for now, but I'd add a TODO to think about this later.

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,225 @@
...
+	}
+	authConfig := h.config.Auth
+	if authConfig != nil {
+		if len(authConfig.BearerToken) > 0 {
+			req.Header.Set(echo.HeaderAuthorization, fmt.Sprintf("Bearer %s", authConfig.BearerToken))
+		}
+		if authConfig.BasicAuth != nil {
+			password, err := authConfig.BasicAuth.GetPassword()
+			if err != nil {
+				logrus.WithError(err).Error("unable to get access to the password for the basic auth")
```
juliusv, line 199: "unable to get access" sounds like a permissions / auth problem... looking at the code of `GetPassword()` the most likely reason would be if there's a file configured and probably the file doesn't exist or has a wrong name. "unable to read" / "unable to retrieve"?

internal/api/interface/v1/datasource/interface.go
```diff
@@ -24,17 +24,20 @@ type Query struct {
 	// Name is a prefix of the Datasource.metadata.name that is used to filter the list of the Datasource.
 	// Name can be empty in case you want to return the full list of Datasource available.
 	Name string `query:"name"`
+	// Project is the exact name of the project.
+	// The value can come or from the path of the URL or from the query parameter
```
juliusv, line 28:
```suggestion
	// The value can come from the path of the URL or from the query parameter
```

internal/api/interface/v1/globaldatasource/interface.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/perses/common/etcd"
+	"github.com/perses/perses/internal/api/shared"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+type Query struct {
+	etcd.Query
+	// Name is a prefix of the GlobalDatasource.metadata.name that is used to filter the list of the GlobalDatasource.
```
juliusv, line 24: `NamePrefix` could be a less confusing name for this field then?

pkg/model/api/v1/datasource.go
```diff
@@ -16,117 +16,142 @@ package v1
...
-	urlAsString := ""
-	if d.URL != nil {
-		urlAsString = d.URL.String()
-	}
-	tmp := &tmpDatasourceSpec{
-		URL: urlAsString,
+func unmarshalDatasourceSpec(spec map[string]interface{}, staticMarshal func(interface{}) ([]byte, error), staticUnmarshal func([]byte, interface{}) error) (DatasourceSpec, error) {
+	if specKind, ok := spec["kind"]; !ok {
+		return nil, fmt.Errorf("attribute 'kind' not found in 'datasource.spec'")
+	} else {
```
juliusv, line 39: The `else` isn't needed here, as you return early in the first branch anyway, so you can keep things less nested / indented.

juliusv: Ah use it to access `specKind` later, ok.

internal/api/core/middleware/proxy.go
Nexucis: indeed we shouldn't have a list of header that cannot be overridden

Nexucis: will put a todo

internal/api/core/middleware/proxy.go
Nexucis: extract is definitely better :). Thanks !

internal/api/core/middleware/proxy.go
Nexucis: ah that's where it comes from. Perhaps I should change the comment then ?

Nexucis:
alright I think I tackled all comments @juliusv. I also renamed the `WhiteList` attribute with `AllowedEndpoints`. I think it describes a bit better the purpose of this attribute.

pkg/model/api/v1/datasource_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/perses/perses/pkg/model/api/v1/datasource"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v2"
+)
+
+func TestUnmarshallJSONDatasource(t *testing.T) {
```
juliusv, line 27:
```suggestion
func TestUnmarshalJSONDatasource(t *testing.T) {
```

juliusv: (and everywhere else)

Nexucis: fixed

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,226 @@
...
+	req.URL.Path = h.path
+
+	// Set up the proxy
+	var proxyErr error
+	reverseProxy := httputil.NewSingleHostReverseProxy(h.config.URL)
+	reverseProxy.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, err error) {
+		desc := h.config.URL.String()
+		proxyErr = fmt.Errorf("error proxying, remote unreachable: target=%s, err=%w", desc, err)
+		logrus.Errorf(proxyErr.Error())
+		proxyErr = err
```
robskillington, line 160: Hm this seems to clobber the definition proxyErr assignment on line 158?

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,226 @@
...
+		return nil, "", echo.NewHTTPError(http.StatusBadGateway, "unable to forward the request to the datasource, request not properly formatted")
+	}
+	projectName := matchingGroups[0][1]
+	datasourceName := matchingGroups[0][2]
+	// getting the datasource object
+	dts, err := dao.Get(projectName, datasourceName)
+	if err != nil {
+		if etcd.IsKeyNotFound(err) {
+			logrus.Debugf("unable to find the Datasource '%s' in project '%s'", datasourceName, projectName)
+			return nil, "", echo.NewHTTPError(http.StatusBadGateway, fmt.Sprintf("unable to forward the request to the datasource '%s', datasource doesn't exist", datasourceName))
```
robskillington, line 110: nit: Should this be http.StatusNotFound?

internal/api/core/middleware/proxy.go
```diff
@@ -1,118 +1,226 @@
...
+		return nil, "", echo.NewHTTPError(http.StatusBadGateway, "unable to forward the request to the datasource, request not properly formatted")
 	}
 	datasourceName := matchingGroups[0][1]
 	// getting the datasource object
 	dts, err := dao.Get(datasourceName)
 	if err != nil {
 		if etcd.IsKeyNotFound(err) {
 			logrus.Debugf("unable to find the Datasource '%s'", datasourceName)
-			return nil, echo.NewHTTPError(http.StatusBadGateway, fmt.Sprintf("unable to forward the request to the datasource '%s', datasource doesn't exist", datasourceName))
+			return nil, "", echo.NewHTTPError(http.StatusBadGateway, fmt.Sprintf("unable to forward the request to the datasource '%s', datasource doesn't exist", datasourceName))
```
robskillington, line 84: nit: Should this be http.StatusNotFound?

internal/api/impl/v1/datasource/persistence.go
```diff
@@ -41,13 +41,13 @@ func (d *dao) Update(entity *v1.Datasource) error {
 	return d.client.Upsert(key, entity)
 }
 
-func (d *dao) Delete(name string) error {
-	key := v1.GenerateDatasourceID(name)
+func (d *dao) Delete(project string, name string) error {
```
robskillington, line 44: nit: Can be `project, name string` here (since types are same for the two args)

internal/api/impl/v1/datasource/persistence.go
```diff
@@ -41,13 +41,13 @@ func (d *dao) Update(entity *v1.Datasource) error {
...
+	key := v1.GenerateDatasourceID(project, name)
 	return d.client.Delete(key)
 }
 
-func (d *dao) Get(name string) (*v1.Datasource, error) {
-	key := v1.GenerateDatasourceID(name)
+func (d *dao) Get(project string, name string) (*v1.Datasource, error) {
```
robskillington, line 49: nit: Can be `project, name string` here (since types are same for the two args)

internal/api/interface/v1/datasource/interface.go
```diff
@@ -21,20 +21,23 @@ import (
...
-	return v1.GenerateDatasourceID(q.Name), nil
+	return v1.GenerateDatasourceID(q.Project, q.NamePrefix), nil
 }
 
 type DAO interface {
 	Create(entity *v1.Datasource) error
 	Update(entity *v1.Datasource) error
-	Delete(name string) error
-	Get(name string) (*v1.Datasource, error)
+	Delete(project string, name string) error
```
robskillington, line 39: nit: Can be `project, name string` here (since types are same for the two args)

internal/api/interface/v1/datasource/interface.go
```diff
@@ -21,20 +21,23 @@ import (
...
+	Get(project string, name string) (*v1.Datasource, error)
```
robskillington, line 40: nit: Can be `project, name string` here (since types are same for the two args)

pkg/model/api/v1/datasource/http.go
```diff
@@ -0,0 +1,325 @@
...
+	if err := unmarshal(&tmp); err != nil {
+		return err
+	}
+	if err := h.validate(tmp); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (h *HTTPConfig) validate(conf tmpHTTPConfig) error {
```
robskillington, line 300: nit: Hm this method also seems to be setting variables on the `HTTPConfig` itself. Perhaps this would be better named `validateAndParse(...)` or something?

pkg/model/api/v1/datasource/prometheus.go
```diff
@@ -0,0 +1,84 @@
...
+	},
+	{
+		Endpoint: "/api/v1/query",
+		Method:   http.MethodPost,
+	},
+	{
+		Endpoint: "/api/v1/query_range",
+		Method:   http.MethodPost,
+	},
+}
```
robskillington, line 42: Probably want to add the `/api/v1/label/FOO/values` endpoint too?

pkg/model/api/v1/datasource/prometheus.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+var defaultPrometheusAllowedEndpoints = []HTTPAllowedEndpoint{
+	{
+		Endpoint: "/api/v1/labels",
+		Method:   http.MethodPost,
```
robskillington, line 24: Hm, is this not GET?

robskillington (approved):
LGTM other than minor comments

pkg/model/api/v1/datasource/prometheus.go
Nexucis: yeah that's a TODO (I forgot to add it), we should be able to managed endpoint with variable inside.

pkg/model/api/v1/datasource/prometheus.go
juliusv: We support both methods:

https://github.com/prometheus/prometheus/blob/d77c985f8c122cf5947b0eb621999fcc8dcca9bf/web/api/v1/api.go#L294-L295

...I think someone in the past once ran into problems with `GET` and the URL length limit for match parameters, so for newer Prometheus versions, we support `POST` as well for some of these endpoints.

Nexucis: actually both are supported. It's just POST is avoiding the size limit of the URL, because the parameters are passed in the body

pkg/model/api/v1/datasource/http.go
Nexucis: mmm I don't know, this 'convention' is used on every object. Well it's fine to change it everywhere, perhaps I'm a bit too familiar that in a `validate` method you have code that is changing and setting data. It's a sort of way to validate the data, because at the end, well your data is valid and contained the proper value.
That's the way I see it. WDYT ?

internal/api/core/middleware/proxy.go
Nexucis: that's a good question. Looking at the definition of the error `BadGateway` probably it's indeed better to return the status not found

pkg/model/api/v1/datasource/prometheus.go
Nexucis: and this is not yet used by the Proxy

PR #86, opened 2021-09-28
Fix import / TypeScript error issues

juliusv:
The import line change in
https://github.com/perses/perses/commit/bcb29c7b98aa880ce662ee842eed53a9e013866f
breaks the app for some reason ("Uncaught TypeError: Cannot read property
'useChartQuery' of undefined"), although there is an index.ts that exports
everything. Let's just revert that change for now to make things run again.


Nexucis:
thanks @juliusv !

PR #87, opened 2021-09-28
Add TODO comment to plugin registry import line

juliusv:


Nexucis:
ah yeah I was a bit too fast. thanks !

PR #88, opened 2021-09-28
add sample-data

Nexucis:

### DIFF
--- a/ui/app/sample-data/dashboard.ts
+++ b/ui/app/sample-data/dashboard.ts
@@ -19,7 +19,7 @@ const cadvisor: DashboardResource = {
     name: 'Kubelet - cAdvisor',
   },
   spec: {
-    datasource: { name: 'PromLabs Prometheus' },
+    datasource: { name: 'Public Prometheus Demo Server' },
     // TODO: Should duration actually be a time range?
     duration: '3h',
     variables: {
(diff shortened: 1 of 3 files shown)


ui/app/sample-data/dashboard.ts
```diff
@@ -0,0 +1,273 @@
+import { DashboardResource, AnyVariableDefinition } from '@perses-ui/core';
+
+const cadvisor: DashboardResource = {
+  kind: 'Dashboard',
+  metadata: {
+    name: 'Kubelet - cAdvisor',
+  },
+  spec: {
+    datasource: { name: 'PromLabs Prometheus' },
```
juliusv, line 9:
```suggestion
    datasource: { name: 'Public Prometheus Demo Server' },
```

ui/app/sample-data/dashboard.ts
```diff
@@ -0,0 +1,273 @@
+import { DashboardResource, AnyVariableDefinition } from '@perses-ui/core';
+
+const cadvisor: DashboardResource = {
+  kind: 'Dashboard',
+  metadata: {
+    name: 'Kubelet - cAdvisor',
```
juliusv, line 6: The public demo server doesn't have Kubernetes data (nor does my PromLabs one), so it would be better to check in a Node Exporter dashboard or something that works against the data in that server.

Nexucis: well the goal is more to use the json you have in the directory `./dev`. These data will just stay until the frontend is connected to the backend

Nexucis: here it's just adding data to be able to build and start the app

juliusv:
:+1: as a temporary workaround to be able to start the app.

PR #92, opened 2021-09-30
fix checklicense to work on different shell

Nexucis:
hopefully fixing the script checklicense to work on multiple different shell. (at least works again on mine ...)

Also it ignores all sub folder `dist` `node_modules` instead of just the one at the root package.

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-exclude_directories=("./ui/node_modules" ".git" ".idea" ".github")
+exclude_directories=("*/node_modules/*" ".git/*" ".idea/*" ".github/*" "*/dist/*")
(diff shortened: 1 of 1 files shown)


check_license.sh
```diff
@@ -21,17 +21,18 @@ license_header="// Copyright ${year} ${license_copyright}
 
 function buildExcludeDirectories() {
   local result
```
juliusv, line 23: Is this line still needed if you have the other `local result=...` line?

juliusv:
Not the shell expert, but :+1: if it works for you :)

check_license.sh
Nexucis: oups, indeed not needed anymore.

PR #94, opened 2021-10-04
use golang regexp to allow variable usage in allowedEndpoint

Nexucis:
This PR is changing the attribute `Endpoint` in the struct `AllowedEndpoint` from `string` to `regexp.Regexp`.

It's a way to be able to write endpoint with variable like `/api/v1/label/<label_name>/values`. 

Of course moving from a `string` to `regexp.Regexp` allows more than just the usage of variable. But I think it's more interesting like that. It opens the possibility to write something like that `/api/v1/label(s)?/*`


internal/api/core/middleware/proxy.go
```diff
@@ -145,12 +143,23 @@ func (h *httpProxy) serve(c echo.Context) error {
 	req := c.Request()
 	res := c.Response()
 
+	isAllowed := false
+	for i := 0; i < len(h.config.AllowedEndpoints) && !isAllowed; i++ {
+		allowedEndpoint := h.config.AllowedEndpoints[i]
+		isAllowed = isAllowed || (allowedEndpoint.Method == req.Method && len(allowedEndpoint.Endpoint.FindAllString(h.path, -1)) > 0)
```
juliusv, line 149: I think you can remove the `isAllowed ||` here because it will always be false anyway (otherwise we don't enter the loop).

Overall, would a normal `range` loop be simpler and clearer?

```go
isAllowed := false
for _, allowedEndpoint := range h.config.AllowedEndpoints {
    if allowedEndpoint.Method == req.Method && len(allowedEndpoint.Endpoint.FindAllString(h.path, -1)) > 0 {
        isAllowed = true
        break;
    }
}
```

internal/api/core/middleware/proxy.go
```diff
@@ -145,12 +143,23 @@ func (h *httpProxy) serve(c echo.Context) error {
...
+		return echo.NewHTTPError(http.StatusForbidden, fmt.Sprintf("you are not allowed to use this endpoint '%s' with the HTTP method %s", h.path, req.Method))
+	}
+
 	if err := h.prepareRequest(c); err != nil {
 		return err
 	}
 
 	// redirect the request to the datasource
 	req.URL.Path = h.path
+	logrus.Debugf("request will be redirected to '%s'", h.config.URL.String())
```
juliusv, line 162: Btw. here and in other places: I'd normally use `%q` instead of `'%s'` to automatically correctly quote+escape complex unknown string values, could make sense here too?

pkg/model/api/v1/datasource/http.go
```diff
@@ -70,43 +71,74 @@ func (h *HTTPAccess) validate() error {
 }
 
 type HTTPAllowedEndpoint struct {
+	Endpoint *regexp.Regexp `json:"endpoint" yaml:"endpoint"`
```
juliusv, line 74: I'd feel renaming this to `EndpointPattern` or something like that would be clearer now with this change?

internal/api/core/middleware/proxy.go
juliusv: Hm, `range` loops are way more Go-y and simpler to read to me, even with breaks... but don't feel too strongly either.

pkg/model/api/v1/datasource/http.go
Nexucis: ah yes, good idea!

Nexucis:
@juliusv should be good now

pkg/model/api/v1/datasource/http.go
```diff
@@ -70,49 +71,80 @@ func (h *HTTPAccess) validate() error {
 }
 
 type HTTPAllowedEndpoint struct {
-	Endpoint string `json:"endpoint" yaml:"endpoint"`
-	Method   string `json:"method" yaml:"method"`
+	EndpointPattern *regexp.Regexp `json:"endpoint_pattern" yaml:"endpoint_pattern"`
+	Method          string         `json:"method" yaml:"method"`
+}
+
+type tmpHTTPAllowedEndpoint struct {
```
juliusv, line 78: I kind of like this approach to making regex fields YAML-marshalable: https://github.com/prometheus/prometheus/blob/60d099088692b6b5bf4b8f4e13f76bd3921da99e/pkg/relabel/relabel.go#L135-L180

And then it's just used like this: https://github.com/prometheus/prometheus/blob/60d099088692b6b5bf4b8f4e13f76bd3921da99e/pkg/relabel/relabel.go#L80-L81

Then it can also be reused easily in other places if defined like that. What do you think?

Nexucis: ah yeah that's definitely better and smarter. Thanks !

PR #95, opened 2021-10-06
set up tsconfig and webpack to build and hot rebuild all packages

Nexucis:


ui/app/webpack.common.ts
```diff
@@ -98,7 +62,6 @@ export const commonConfig: Configuration = {
         test: /\.(png|jpg|gif)$/,
         type: 'asset',
       },
-      // SVG as React components
```
juliusv, line 101: Doesn't this comment still apply?

ui/app/webpack.dev.ts
```diff
@@ -55,18 +36,11 @@ const devConfig: Configuration = {
   },
 
   devServer: {
-    port: parseInt(process.env.PORT ?? '3001'),
+    port: parseInt(process.env.PORT ?? '3000'),
     open: true,
-    https: getHttpsConfig(),
```
juliusv, line 60: Does this HTTPS change relate to the rest of the PR?

ui/app/webpack.common.ts
Nexucis: ah yeah it is indeed

ui/app/webpack.common.ts
```diff
@@ -12,76 +12,40 @@
...
   resolve: {
-    // Add ts/tsx to list of files that can be imported without including the extension
-    extensions: ['.ts', '.tsx', '.js', '.jsx', '.json'],
+    extensions: ['.ts', '.tsx', '.js', 'jsx', '.json'],
   },
-
   plugins: [
-    new CleanWebpackPlugin(),
+    new CleanWebpackPlugin({
+      dangerouslyAllowCleanPatternsOutsideProject: true,
```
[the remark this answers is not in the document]
Nexucis: well I tried without it, indeed you have a warning, but it works well. Sooo guess it's fine :D

ui/app/webpack.common.ts
```diff
@@ -12,76 +12,40 @@
...
-
-      {
-        test: /\.tsx?$/,
-        //include: [FilePaths.SRC],
         use: [
           {
             loader: 'ts-loader',
             options: {
               // Type-checking happens in separate plugin process
-              transpileOnly: true,
```
LukeTillman, line 84: Doing type-checking in a separate process/plugin is supposed to make the build faster. Is that not necessary any more?

ui/app/webpack.dev.ts
```diff
@@ -25,22 +22,6 @@ declare module 'webpack' {
...
-  ) {
-    return true;
-  }
-
-  return {
-    key: fs.readFileSync(process.env.SSL_KEY_FILE),
-    cert: fs.readFileSync(process.env.SSL_CRT_FILE),
-  };
-}
-
```
LukeTillman, line 43: Don't think we should remove this either. As someone who does a lot of `localhost` dev across different codebases/projects, adding a common SSL cert to my local machine's keychain to stop Chrome from complaining is something I definitely use. It really comes in handy if you have some project that you do local dev on that adds an HSTS to Chrome for localhost (which means Chrome always redirects to HTTPS for localhost).

LukeTillman (commented):
Left a few comments inline. Thanks for tackling this!

ui/app/webpack.dev.ts
```diff
@@ -11,9 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-require('dotenv-defaults').config();
```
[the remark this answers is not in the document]
Nexucis: my question would be more why do we need it ? :)

ui/app/webpack.dev.ts
Nexucis: I remove it because it's bothering me and I don't have your setup and I'm pretty sure planty of user won't have it. We can keep it if you have a way to handle it with no particular setup. For example, with a condition that would enable it if you have the correct environment.

ui/app/tsconfig.json
```diff
@@ -1,10 +1,29 @@
...
+  ],
   "compilerOptions": {
     "sourceMap": true,
-    "noEmit": true
+    "outDir": "./dist",
+    "rootDir": "../",
+    "tsBuildInfoFile": "../.build-cache/app.tsbuildinfo",
+    "incremental": true
   },
+  "references": [
```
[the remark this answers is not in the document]
Nexucis: yeah that's the key to make it work.
This and 

```
    "paths": {
      "@perses-ui/*": ["./*/src"]
    }
```
in the root tsconfig

Nexucis:
After a tchat with @LukeTillman, I will put back the eslint, but in a different process in order to not impact the build time.

PR #100, opened 2021-10-18
implement a simple header and a footer

Nexucis:
This PR is adding a Top bar that is displaying the name of the project, and later the logo (once we have one haha). It also included the button to switch the theme. Currently it doesn't do anything. Theming will come a bit after.

This PR is also adding a footer that is containing the copyright, the link to the project (probably later to a website if we have one), and the version. The version is for the moment hardcoded, but soon it will call the API to get the version. (already available under the endpoint `/api/v1/health`

Temporaly it deactivated the display of the variable as I didn't find a simple way to not having the right bar over the footer and the header. Probably something to tweak or perhaps develop a new component.

I believe this right bar should be contained between the header and the footer when you are on a laptop. And then once you are on phone, it should be displayed like it is today. But then you should be able to collapse it.


ui/app/src/App.tsx
```diff
@@ -28,14 +29,16 @@ function App() {
...
   return (
-    <PluginRegistry
-      loadingFallback="Loading..."
-      ErrorFallbackComponent={AlertErrorFallback}
-    >
-      <DataSourceRegistry>
-        <DashboardView resource={dashboard} />
-      </DataSourceRegistry>
-    </PluginRegistry>
+    <CoreView>
```
juliusv, line 32: I think `<Layout>` or so would be a more common naming for the overall page layout template?

Nexucis: `Layout` could be in conflict with the dasboard code / datamodel as you have this notion as well :(.
Maybe another name ?

ui/app/src/views/core/Core.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { HeaderView } from './Header';
+import { FooterView } from './Footer';
+
+export interface CoreProps {
+  children: ReactNode;
+}
+
+export default function CoreView(props: CoreProps) {
+  return (
+    <div>
```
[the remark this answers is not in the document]
Nexucis: ah yeah it's just to wrap the multiple components. I forgot about this way. Thanks for the reminder !

ui/app/src/views/core/Footer.tsx
```diff
@@ -0,0 +1,42 @@
...
+  '& li + li:before': {
+    content: '"|"',
+    padding: '3px',
+  },
+  '& a:visited': {
+    color: 'inherit',
+  },
+};
+
+export function FooterView(): JSX.Element {
```
juliusv, line 24: Wouldn't you want to make this the default export, since there's only one thing? Same for the header.

Apropos this signature (with explicit return value), I'm more used to React components being typed and written like:

```
import React, { FC } from 'react';

export default Footer: FC = () => <Box>...</Box>
```

...then the return value is inferred from that, and you can also include typing for props like `FC<FooProps>` when needed.

But seems like besides the typing, for components we so far use the `function` keyword rather than arrow functions in Perses (in Prometheus we do it the other way around)?

ui/app/src/views/core/Core.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <HeaderView />
```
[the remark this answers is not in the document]
juliusv: Ah, I assumed the `View` part in there was similar to why you had the `View` in your components, because `Dashboard` is already a widely used term in the rest of the project as well.

ui/app/src/App.tsx
LukeTillman: Yeah, I typically start the same way (`Header` and `Footer` directly in the root `App` component), so if you wanted to do it that way, totally cool.

ui/app/src/views/core/Core.tsx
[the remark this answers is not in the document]
LukeTillman: I think of `View` basically being synonymous with "Page" (which is why I used that naming convention for `DashboardView` since it's a main "page" in the app). Not sure how many views we're going to have initially to be honest (Dashboard may be the only one for awhile). But since these are just components, totally cool with just calling them `Header` and `Footer` if you want?

ui/app/src/views/core/Footer.tsx
LukeTillman: Just my $0.02 but I think either syntax is fine for components (`function Blah() {}` or `const Blah = () => {}`). Most React codebases I've been in have tended to mix and match. Personally I gravitate towards `function`, but it's totally just a style thing I probably picked up years ago, maybe before we had arrow functions. 🤷‍♂️ 

RE: Default exports though, that's actually a convention that I'm pretty sure React devs picked up because if you ever go to do code-splitting, [`React.lazy`](https://reactjs.org/docs/code-splitting.html#reactlazy) for lazy-loading components only works with components that are default exports out of the box, so I think we all got in the habit of just always making components be default exports.

ui/app/src/views/core/Core.tsx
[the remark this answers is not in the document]
Nexucis: I think we should have these views : 
* home page (no idea what we would have inside, but could be cool)
* the dashboard display (of course!)
* a page to display the datasource
* Perhaps the metric explorer ? (could be cool to have it)

Probably I will call them `Header` and `Footer` since they are just components and not the entire page.

PR #102, opened 2021-10-19
Fetch the version from the server

Nexucis:
This PR introduced the dev proxy required to query the backend from the frontend and fetch the `health` endpoint that contain the version of Perses.

I'm wondering if we should display the commit ID. What do you think ?

### DIFF
--- a/ui/app/src/Footer.tsx
+++ b/ui/app/src/Footer.tsx
@@ -1,6 +1,7 @@
 import { Box, Theme } from '@mui/material';
 import { SxProps } from '@mui/system/styleFunctionSx/styleFunctionSx';
 import { Github } from 'mdi-material-ui';
+import useHealth from './health-client';
 
 const style: SxProps<Theme> = {
   display: 'flex',
@@ -22,6 +23,7 @@ const style: SxProps<Theme> = {
 };
 
 export default function Footer(): JSX.Element {
+  const { response } = useHealth();
   return (
     <Box sx={style}>
       <ul>
@@ -35,7 +37,11 @@ export default function Footer(): JSX.Element {
             <Github sx={{ verticalAlign: 'bottom' }} />
           </a>
         </li>
-        <li>v0.1.0</li>
+        <li>
+          {response && response.version.length > 0
+            ? response.version
+            : 'development version'}
+        </li>
       </ul>
     </Box>
   );
--- a/ui/app/src/Footer.tsx
+++ b/ui/app/src/Footer.tsx
@@ -1,7 +1,8 @@
-import { Box, Theme } from '@mui/material';
+import { Box, CircularProgress, Theme } from '@mui/material';
 import { SxProps } from '@mui/system/styleFunctionSx/styleFunctionSx';
 import { Github } from 'mdi-material-ui';
 import useHealth from './health-client';
+import Toast from './components/Toast';
 
 const style: SxProps<Theme> = {
   display: 'flex',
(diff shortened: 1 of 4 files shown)


Nexucis:
on the paper, I would say why not. I'm just curious about the use case you have in mind where you need to compare the timestamp of the server with your local one.

roidelapluie:
Prometheus does this: we run a time() query in the UI. If the time difference is huge, we inform the user that what they see in the graphs might not be accurate, and that they might not query the data they think they query.

ui/app/src/health-client.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright 2021 The Perses Authors
```
juliusv, line 1: Would it make sense to put this file into the `core/src/hooks` dir along with the other hooks in there, or is this not "core"-specific enough? Not 100% sure in general what should go into `core` vs. the `app`, I guess we'll be figuring that out :)

juliusv:
> Prometheus does this: we run a time() query in the UI. If the time difference is huge, we inform the user that what they see in the graphs might not be accurate, and that they might not query the data they think they query.

To expand on the reasoning for this: sometimes people have very unsynchronized clocks between the browser and the server (also sometimes due to timezone screwups), and the browser by default sends the browser's current time as the `end` time for a graph, and if that's e.g. in the future from the perspective of the server, people get a completely empty graph and wonder why.

ui/app/src/health-client.ts
Nexucis: and I'm fine with this approach. At least it's simpler \o/

Nexucis:
ah that's interesting. Thanks for the thoughts @roidelapluie @juliusv. Now I'm wondering if it really makes sense to store the time in the `health` endpoint.

Well just because, the perses server could have a different time than Prometheus as well don't you think? Or we assume that all servers are well synchronized ?

ui/app/src/Footer.tsx
```diff
@@ -22,6 +23,7 @@ const style: SxProps<Theme> = {
 };
 
 export default function Footer(): JSX.Element {
+  const { response } = useHealth();
```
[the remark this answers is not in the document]
Nexucis: like that : https://mui.com/components/snackbars/#customized-snackbars

But I think it's not possible to stack the message. I think it's in the material specification that the toast shouldn't be stacked.

ui/app/src/health-client.ts
juliusv: Ah ok, fine by me :)

ui/app/src/Footer.tsx
```diff
@@ -22,21 +24,33 @@ const style: SxProps<Theme> = {
...
+            <a
+              href="https://github.com/perses/perses"
+              target="_blank"
+              rel="noreferrer"
+            >
+              <Github sx={{ verticalAlign: 'bottom' }} />
+            </a>
+          </li>
+          <li>
+            {loading ? (
```
LukeTillman, line 43: Love it. This "if loading" thing is something I see in React all the time, so think you nailed it.

ui/app/src/components/Toast.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Alert, Snackbar } from '@mui/material';
+import { SyntheticEvent, useEffect, useState } from 'react';
+
+export interface ToastProperties {
+  loading: boolean;
+  message?: string;
+  severity: 'error' | 'warning' | 'info' | 'success';
+}
+
+export default function Toast(props: ToastProperties) {
```
LukeTillman, line 23: 👏

LukeTillman: I have some more "common" code here that I might be able to contribute to the project once I talk to Rob around showing errors via a Toast globally. Will follow up.

ui/app/src/health-client.ts
LukeTillman: Yeah, I can talk you through the philosophy I usually follow with data-fetching stuff like this maybe next time we get together. Might even be worth putting it into a GitHub discussion. For `app` vs `core`, I'm thinking "do plugins need this too" and if so, it goes in `core`, otherwise starts in `app`.

Once you decide which project/package, then it's a question of "where in the package do I put it", so maybe the more interesting question here is "where in app does it go"? And I think that's an open question (I have thoughts), but if we're going to potentially reorganize the folder structure, may have a different answer once we figure that out. 😂

ui/app/src/components/Toast.tsx
Nexucis: yeah that would be great to have a common way to handle the error. I think here it's a bit too local and doesn't really scale

PR #106, opened 2021-10-21
Add project list in the top bar

Nexucis:
This PR is adding the project list in the top bar.

It will be used later to navigate from a project to another.

to be merged after #105 

image


Nexucis:
this PR is ready to review @juliusv @LukeTillman

ui/app/src/components/Header.tsx
```diff
@@ -11,17 +11,99 @@
...
+          />
+          {loading ? (
+            <CircularProgress size="1rem" />
+          ) : (
+            response && <ProjectMenu projectList={response} />
+          )}
+        </Box>
         <Switch />
       </Toolbar>
+      {/* TODO manage in a single place the pull of fetching error */}
```
juliusv, line 105: What do you mean with "pull" in "pull of fetching error"? Pool as in multiple errors?

ui/app/src/components/Header.tsx
```diff
@@ -11,17 +11,99 @@
...
+      <Toast loading={loading} severity={'error'} message={error?.message} />
```
juliusv, line 106:
```suggestion
      <Toast loading={loading} severity="error" message={error?.message} />
```

I see there's no `{error && ...}` here, does the toast disappear automatically if the `message` is undefined?

ui/app/src/components/Header.tsx
```diff
@@ -11,17 +11,99 @@
...
+import { ProjectModel, useProjectQuery } from '../model/project-client';
+import Toast from './Toast';
+
+function ProjectMenu(props: { projectList: ProjectModel[] }): JSX.Element {
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+
+  const handleMenu = (event: MouseEvent<HTMLElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+  const handleCloseMenu = () => {
```
juliusv, line 39: I'd just inline these two single-line functions in the single place where they're used?

ui/app/src/components/Header.tsx
```diff
@@ -11,17 +11,99 @@
...
+    setAnchorEl(null);
+  };
+  return (
+    <>
+      <Button
+        aria-label="list of the available project"
```
juliusv, line 45:
```suggestion
        aria-label="List of the available projects"
```

ui/app/src/components/Header.tsx
```diff
@@ -11,17 +11,99 @@
...
+      </Button>
+      <Menu
+        id="menu-project-list-appbar"
+        anchorEl={anchorEl}
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'left',
+        }}
+        keepMounted
+        open={Boolean(anchorEl)}
```
juliusv, line 62: This may be clearer / more explicit:

```suggestion
        open={anchorEl !== null}
```

ui/app/src/components/Header.tsx
juliusv: Ok `handleCloseMenu()` is actually used in two places... but meh.

juliusv:
Generally cool, just a bunch of nits again :)

ui/app/src/components/Header.tsx
Nexucis: ah yeah `pool of error` as it can comes from multiple different sources. It won't scale if each component has to integrate the way to display the error

juliusv: Yep, makes sense!

ui/app/src/components/Header.tsx
Nexucis: It's actually in the `Toast` component that the check occurred. So if the `message` is undefined, then it won't show up.
Here: https://github.com/perses/perses/blob/main/ui/app/src/components/Toast.tsx#L26 you have the check

ui/app/src/components/Header.tsx
Nexucis: I think it won't be used anymore once we have the code to switch from a workspace to another one (with routing and so one)

Nexucis:
Note: probably we should merge this PR after #108, so we can adapt this code with the new way to fetch

LukeTillman (approved):
One minor suggestion (and one question that might warrant a discussion) inline, but overall looks great. And good idea on switching to `react-query` before merging. 👍

ui/app/src/components/Header.tsx
Nexucis: I kind of a big fan of toast. I find it fit nicely to show fetch error. Probably we could add a message that is wrapping the error message. So you can identify where does it come from

juliusv: For the top menu I don't mind too much where the error is displayed, but when it comes to showing errors of individual PromQL errors, it would be nice to show them close to where the query is defined, because otherwise it's going to be harder to find out which one's failing. (I also prefer the inline query error in Prometheus over the one in Grafana, where you have to hover over a tiny area at the top left of a graph pane to see what the actual error is.)

Nexucis: yeah in the case of the promql query, it would make sense + we will have the place to display the error properly.
In the Top bar it would be weird / ugly to show an error don't you think ? 

I think we can say, we use the Toast to display the fetch error when you are querying the Perses backend server.
When you are targeting the datasource, probably you want to display it in a different way for better understanding.

By the way, the toast can be used to display some success message. Like `datasource has been successfully reached`

juliusv: Yep, that sounds good to me at least.

ui/app/src/components/Header.tsx
```diff
@@ -11,15 +11,108 @@
...
+              {value.metadata.name}
+            </MenuItem>
+          );
+        })}
+      </Menu>
+    </>
+  );
+}
+
+function ProjectMenu(): JSX.Element {
```
LukeTillman, line 80: I think you can just inline the `ProjectMenuComponent` directly into this one. No need to have them be separate. You can also just use an early return for the loading indicator (pretty common pattern in React). For example:

```tsx
if (isLoading) {
  return <CircularProgress />;
}

// Regular rendering after loading has finished...
return (
  <>
    <Button />
    { /* ... rest of regular JSX markup ... */
  </>
);
```

Would also be cool to break it out into its own file, but don't feel like you have to if you like it next to `Header`. Get the feeling we might end up doing it eventually if this header keeps growing.

LukeTillman (commented):
Left another couple minor suggestions inline, but still overall looks good to merge when you're ready. I'm about to put a PR for global snackbar/toasts, so I'll adapt this code too once it's merged in that PR.

ui/app/src/components/Header.tsx
Nexucis: ah yeah that's cleaner. will do that. Thanks for the tips !

Nexucis: > Would also be cool to break it out into its own file
yeah probably it will make sense once we would add another component in the `Header`

ui/app/src/model/project-client.ts
```diff
@@ -0,0 +1,42 @@
...
+import buildURL from './url-builder';
+
+const resource = 'projects';
+
+export interface ProjectModel {
+  kind: 'Project';
+  metadata: Metadata;
+}
+
+type ProjectListOptions = Omit<
```
Nexucis, line 25: I'm feeling we will duplicate each time this kind of type. Maybe worth to have one generic.

Nexucis:
ready for another round of review @LukeTillman :D

ui/app/src/model/project-client.ts
LukeTillman: Yeah, good idea. Might be worth making a shorthand generic type for that.

LukeTillman: I _really_ wish we had some codegen for a bunch of these types based on some API spec. Definitely one thing I miss about not having a GraphQL API + schema. 😂

LukeTillman (approved):
:shipit:

ui/app/src/model/project-client.ts
Nexucis: will put that on my todo list :D

PR #107, opened 2021-10-22
UI File Reorganization, Guidelines Explaining Layout

LukeTillman:
Like we discussed during the working group, this reorganizes the files in the `app` to more closely reflect the intended use of the current folder structure in the UI. I also added a `ui-guidelines.md` explaining some of the structure and providing some guidance for people new to the codebase.

I know the guidelines in particular may warrant some discussion, so definitely feel free to make suggestions. I largely took what we use at Chronosphere as the base for that and modified it slightly to make it more appropriate for Perses. It's really meant to be a living document that should definitely evolve over time as we build things out.

Let me know if you have questions or I'll be on Matrix tomorrow if you want to chat about it!

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -21,6 +21,8 @@ are commonly used across multiple packages in the workspace.
 
 ## Development Environment
 
+First, check out the [UI Guidelines](./ui-guidelines.md) for some information
+on how the codebase is organized and our approach to development in the UI.
 You'll need to have the following installed locally before you can start
 developing:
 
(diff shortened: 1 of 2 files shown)


ui/app/src/App.tsx
```diff
@@ -12,16 +12,14 @@
 // limitations under the License.
 
 import { DashboardResource } from '@perses-ui/core';
-// TODO: Figure out why the import path './context/plugin-registry' doesn't work
-// (the app throws a TypeScript error in the browser) although there is an index.ts
-// file in that directory that should expose everything.
-import { PluginRegistry } from './context/plugin-registry/PluginRegistry';
```
LukeTillman, line 18: There was a circular dependency of imports causing this issue, which I fixed.

PR #109, opened 2021-11-02
Add global snackbar provider, convenience hook

LukeTillman:
This adds a global `SnackbarProvider` which is basically just the provider from [`notistack`](https://iamhosseindhv.com/notistack) which works with Material UI Snackbars/Toasts. This also contains some convenience methods on the `useSnackbar` hook that we've found useful when working with it in the past. Turns out that these methods actually work really well when attached to `react-query`'s `onError` option. 🎉 

Here's what it looks like in action (when failing to fetch version info for the health check because the server is turned off):

https://user-images.githubusercontent.com/428023/139941378-52789bbe-5450-44dd-a1c2-2bea0017d828.mp4

I've left the default `notistack` options for now, but we can always revisit/tweak those later if we want to.

### DIFF
--- a/ui/app/src/App.tsx
+++ b/ui/app/src/App.tsx
@@ -12,7 +12,6 @@
 // limitations under the License.
 
 import { DashboardResource } from '@perses-ui/core';
-import { QueryClient, QueryClientProvider } from 'react-query';
 import { PluginRegistry } from './context/plugin-registry';
 import ViewDashboard from './views/dashboard/ViewDashboard';
 import AlertErrorFallback from './components/AlertErrorFallback';
@@ -22,8 +21,6 @@ import Header from './components/Header';
 import Footer from './components/Footer';
 import { pluginRuntime } from './model/plugin-runtime';
 
-const queryClient = new QueryClient();
-
 function App() {
   const dashboard = useSampleData<DashboardResource>('dashboard');
   if (dashboard === undefined) {
@@ -31,7 +28,7 @@ function App() {
   }
 
   return (
-    <QueryClientProvider client={queryClient}>
+    <>
       <Header />
       <PluginRegistry
         loadingFallback="Loading..."
@@ -43,7 +40,7 @@ function App() {
         </DataSourceRegistry>
       </PluginRegistry>
       <Footer />
-    </QueryClientProvider>
+    </>
   );
 }
 
--- a/ui/app/src/components/Footer.tsx
+++ b/ui/app/src/components/Footer.tsx
@@ -1,8 +1,8 @@
 import { Box, CircularProgress, Theme } from '@mui/material';
 import { SxProps } from '@mui/system/styleFunctionSx/styleFunctionSx';
 import { Github } from 'mdi-material-ui';
+import { useSnackbar } from '../context/SnackbarProvider';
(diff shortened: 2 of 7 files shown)


ui/app/src/App.tsx
```diff
@@ -22,16 +21,14 @@ import Header from './components/Header';
...
-const queryClient = new QueryClient();
-
 function App() {
   const dashboard = useSampleData<DashboardResource>('dashboard');
   if (dashboard === undefined) {
     return null;
   }
 
   return (
-    <QueryClientProvider client={queryClient}>
```
LukeTillman, line 34: I moved this up to `index.tsx` since that's typically where those sorts of providers live (probably should have put it there to begin with).

ui/app/src/components/Footer.tsx
```diff
@@ -24,7 +24,8 @@ const style: SxProps<Theme> = {
 };
 
 export default function Footer(): JSX.Element {
-  const { data, isLoading, error } = useHealth();
+  const { exceptionSnackbar } = useSnackbar();
+  const { data, isLoading } = useHealth({ onError: exceptionSnackbar });
```
LukeTillman, line 28: Works nicely with `react-query` as just the `onError` option.

ui/app/src/context/SnackbarProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * Gets the SnackbarContext with methods for displaying snackbars/toasts.
+ */
+export function useSnackbar(): SnackbarContext {
+  const { enqueueSnackbar, closeSnackbar } = useNotistack();
+
+  // Create variant-specific callbacks
+  const errorSnackbar = useEnqueueFunction(enqueueSnackbar, 'error');
+  const infoSnackbar = useEnqueueFunction(enqueueSnackbar, 'info');
+  const warningSnackbar = useEnqueueFunction(enqueueSnackbar, 'warning');
+  const successSnackbar = useEnqueueFunction(enqueueSnackbar, 'success');
```
LukeTillman, line 59: We've found these super useful for a couple reasons:

1. We often want the ability to potentially show different types of snackbars in a single page/hook, so being able to grab what you need (including multiple) is nice, e.g.:
```ts
const { successSnackbar, exceptionSnackbar } = useSnackbar();
```
2. We almost never need to create and pass an options object which tends to look a lot less verbose. So instead of:
```ts
enqueueSnackbar('Something really bad happened', { 
  variant: 'error'
});
```
We can just do:
```ts
errorSnackbar('Something really bad happened');
```

Nexucis:
looks cool :). Just wondering if the notification shouldn't be on the right instead. I'm feeling it's more natural to look at the right.

juliusv:
Yeah, I think bottom right seems more intuitive to me as well.

Looks good otherwise!

Nexucis:
probably we can remove the `Toast`. It was "designed" and used only for displaying the error and later a success message. But the snackbar you add has the same purpose. So no need to keep the `Toast`

LukeTillman:
Cool, I'll remove `Toast` and update the options to put it in the bottom right.

Nexucis (approved):
nice thanks !

PR #110, opened 2021-11-03
Expose time range as absolute

LukeTillman:
This is to start to address the discussion in #91. Here I'm doing three things:

1. Always expose the dashboard's current time range as an absolute time to plugins so we don't have the issue with different ranges being used across different panels for relative times.
2. Moves the logic to align the time range with the step out of the Prom API client that does fetching (see my `TODO` note there)
3. Upgrades all our webpack dev dependencies to their latest versions.

I did the third one there because I've noticed locally that I sometimes get weird build errors when making changes to the plugins while the dev server is started and it look like a recent update to webpack may have fixed it. (Side note: might be nice to find some bot or something to prompt us to do dependency upgrades regularly...)

I think once I start to dig into implementing the `LineChart` panel, I'm probably going to want to change up the query hook's API (e.g. maybe not `DataFrame`s any more, maybe providing the current time range as an argument, etc.), but we'll see as that evolves.


ui/prometheus-plugin/src/plugins/range-chart-query.ts
```diff
@@ -54,13 +55,32 @@ export function usePrometheusRangeChartQuery(
   const timeRange = useDashboardPrometheusTimeRange();
   const step = usePanelRangeStep(timeRange, minStep);
 
+  // Align the time range so that it's a multiple of the step (TODO: we may
+  // ultimately want to return this from the hook so that charts will know what
+  // time range was actually used?)
```
juliusv, line 60: Yep! We'll eventually want that knowledge available to whatever paints the chart, so it can show exactly that range, no matter how much data is being returned :)

juliusv:
Looks good to me besides that TODO that I assume is for later.

LukeTillman:
Yeah, I figured that `TODO` can happen once I try to implement the `LineChart` and probably need to change up the API to do it.

juliusv:
:+1:

ui/prometheus-plugin/src/model/time.ts
```diff
@@ -33,11 +31,15 @@ export interface PrometheusTimeRange {
...
+  const { start, end } = useDashboardTimeRange();
 
-  // Only recalculate the time range if the value on the dashboard changes,
-  // otherwise relative time ranges will change every time this hook is called
-  return useMemoized(() => toPrometheusTimeRange(timeRange), [timeRange]);
+  // Only recalculate the time range if the value on the dashboard changes
+  return useMemoized(() => {
+    return {
+      start: Math.ceil(getUnixTime(start)),
+      end: Math.ceil(getUnixTime(end)),
```
Nexucis, line 40: Don't we think this thing should be at the  `core` level / global level ? I don't see why it is related to Prometheus (particularly).

ui/prometheus-plugin/src/plugins/range-chart-query.ts
```diff
@@ -54,13 +55,32 @@ export function usePrometheusRangeChartQuery(
...
+  const { start, end } = useMemoized(() => {
```
Nexucis, line 61: same question than before, I have the feeling this should be a global hook. I don't really see why it's relative to Prometheus. Probably I'm missing something here. (it was simpler in my head haha)

Nexucis:
looks cleaner overall ! just some questions on my side (as always I think haha)

> (Side note: might be nice to find some bot or something to prompt us to do dependency upgrades regularly...)

Perhaps we can enable dependent bot. I'm wondering how it can work with a monorepo.

ui/prometheus-plugin/src/model/time.ts
LukeTillman: I feel like unless we end up with multiple data source that deal with things in unix timestamp seconds (and not ms or some other date format), this is definitely Prometheus-specific. In Grafana, for example, this logic is Prom-specific and in the Prom DataSource, not in the core libraries.

ui/prometheus-plugin/src/plugins/range-chart-query.ts
LukeTillman: Same thing with this--I see this as Prom-specific behavior at least for now. (And same as above, Grafana treats this as Prom-specific behavior as well and this is in the Prom DataSource.) I think we'll know more when we get to other data sources down the road whether they need/want similar behavior. May not be necessary for them and they may also be dealing in different date/step formats that aren't unix time in seconds.

Nexucis (approved):
thanks for the answer @LukeTillman. Looks good to me.

PR #112, opened 2021-11-09
Initial LineChartPanel prototype

LukeTillman:
This is an initial prototype for the `LineChartPanel` plugin. I started as simple as I could and deferred figuring out tooltips, legends, etc. for now. I figured we can start with something like this, then iterate on it to add features. I implemented the chart rendering with both Apache ECharts and uPlot. It looks something like this:

https://user-images.githubusercontent.com/428023/140957808-ac14b1e3-f2be-4f33-9d96-406729abdc90.mp4

As you can see, it's using ECharts by default, but if you add `uplot=true` to the query string, it will switch to using uPlot instead. I figure that we can maintain both for a short period of time until we get a better feeling for the performance and are comfortable picking a "winner", at which point we can retire the other implementation.

As we discussed in Matrix chat, I renamed the query plugin to `GraphQuery` instead of `ChartQuery` to try and better differentiate between range/instant query types, although I'm honestly not sure if that's any better (naming is hard 😂 ). I also got rid of the data frames columnar based model and switched to a row-based one that's friendlier to both the Prom return format and the format expected by ECharts (less transformation involved this way). I used `Iterable` in those interfaces to allow us to potentially defer some of the processing in a smarter way at some point in the future if we want to.


ui/panels-plugin/src/plugins/line-chart/LineChart.tsx
```diff
@@ -11,23 +11,124 @@
...
+    const dataset: EChartsOption['dataset'] = [];
+    const series: EChartsOption['series'] = [];
+
+    for (const query of queries) {
+      // Skip queries that are still loading and don't have data
+      if (query.loading || query.data === undefined) continue;
+
+      // For every series that comes back from a query, add a Dataset and a Series
+      // to the chart
+      let id = 0;
```
juliusv, line 53: Should the `id` be incremented per query, not per series? I'm getting a `Duplicated option on id "0"` error when trying to run multiple queries currently.

juliusv: Although I guess that error would indicate that it really needs to be different for every series, even across queries?

PR #127, opened 2021-11-12
simplify the update of the npm dependencies

Nexucis:


Nexucis:
dependabot is handling properly the way to update dependencies when you are using `npm workspaces`. See :
* https://github.com/perses/perses/pull/119
* https://github.com/perses/perses/pull/117

/cc @juliusv @roidelapluie, based on that it will be nice to enable it in prometheus/prometheus, will do it once I finished to do it with perses

PR #139, opened 2021-11-15
disable host check and disable https config in gitpod env

Nexucis:


Nexucis:
it's based on https://github.com/webpack/webpack-dev-server/issues/883

`disableHostCheck` has been changed in favor of `allowedHosts` in webpack-dev-server v4

and I just added a special environment in case we specifically want to deactivate the HTTPS config in the webpack-dev config.
With these two options added, the preview in gitpod is working again.

Nexucis:
Yes good idea. I will change it

PR #141, opened 2021-11-15
Improve wordings in CONTRIBUTING.md

juliusv:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,76 +1,74 @@
 Contributing
 ============
 
-As Perses is still in progress, process to contribute is not entirely in place.
+As Perses is still a work in progress, the contribution process is still evolving.
 
-We are using GitHub for the development.
+We are using GitHub as our main development and discussion forum.
 
-* All PRs should go there
-* We are opening issue when we are sure about a feature, and we want to trace it. We also open issues when the feature proposed should be small and doesn't require huge talk.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -82,26 +80,8 @@ The secure way to configure your monitoring.               <\
...
-```
-
-* Finally, start the web-app:
-
-```bash
-npm start
-```
-
-The web-app is now available using the url http://localhost:3000.
+See the [ui/README.md](./ui/README.md) file for details around the build process and the structure of the web UI. The web UI currently still shows a canned test dashboard that is not loaded from the Perses API backend yet, but it already runs health checks that require the backend to be running in order to not show errors in the UI.
```
Nexucis, line 87: perhaps we should keep the minimum version required to run the frontend don't you think ?
Here or in the `ui/readme` ?

Nexucis: I'm wondering if we should put nodejs 16 instead. Not sure if it really works with nodejs 14.
Since we are running nodejs 16 in the CI, probably better to aligned the version with it.

Nexucis (approved):
nice, thanks a lot for the rewritting !! 
I just left some comments, but overall it looks already super good to me :)

CONTRIBUTING.md
juliusv: Added that to `ui/README.md` and specified Node 16 there.

PR #166, opened 2021-11-18
Initial radial gauge chart 

sjcobb:
Still a work in progress (I left TODO items in comments), but want to go ahead and get some feedback.

Current state:

gauge-ex-scrnsht

no-data_message_gauge


Nexucis:
that's so cool, thank you @sjcobb !

Instead of adding an example of Gauge, do you think we could instead add the gauche chart related to the node exporter dashboard.

Because the one you modified is the one representing the node exporter dashboard. And you should find some gauge to display :).

Last question : the gauge chart here is done with Echart, are you planning to do it as well with uplot ? @LukeTillman I supposed that's what we need to do for the moment right ?

sjcobb:
Thanks! From what I can see on their [demo page](https://leeoniya.github.io/uPlot/demos/), it doesn't look like uPlot supports radial gauge charts. Which is by design, see their creator's [response here](https://github.com/leeoniya/uPlot/issues/546#issuecomment-877502513), "uPlot is unlikely to do radial/polar charts like pie, radar, gauge"

Nexucis:
ah right, thanks for the info @sjcobb. If you are ok to complete the node exporter dashboard, that would be cool.

I didn't look into the details, but it looks good to me overall. And yup definitely now it would make sense to have a common code to wrape the echart so it can decrease the amount of code (and potentially duplicated code) between the Gauge and Line chart.

ui/panels-plugin/src/plugins/gauge-chart/GaugeChart.tsx
```diff
@@ -11,22 +11,262 @@
...
-  return <div>{JSON.stringify(props)}</div>;
+// TODO (sjcobb): take threshold default colors from theme (green, red, yellow, gray)
+const thresholdColors = {
+  green: 'rgb(115, 191, 105)',
+  yellow: 'rgba(237, 129, 40, 0.89)',
+  red: 'rgba(245, 54, 54, 0.9)',
+  gray: '#e1e5e9',
+};
+
+const noDataOption = {
```
LukeTillman, line 53: Nice, like this idea.

ui/panels-plugin/src/plugins/gauge-chart/GaugeChart.tsx
```diff
@@ -11,22 +11,262 @@
...
+  },
+  series: [],
+};
+
+function GaugeChart(props: GaugeChartProps) {
+  const { query, width, height, calculation, unit } = props;
+  const { data } = useGraphQuery(query);
+
+  const option: EChartsOption = useMemo(() => {
+    // TODO (sjcobb): add loading spinner, share noDataOption with other charts
```
LukeTillman, line 78: Yeah, we're gonna have to figure out an approach to loading pretty soon. I was thinking maybe making use of MUI's `Skeleton` component.

LukeTillman (approved):
Looks like a great initial prototype, nice work! I think we can follow-up in another PR and work on de-duping some of this with `LineChart` (e.g. extract some of the common e-charts wrapper code). Once this is merged, I'll update my PR that's converting Grafana JSON to convert Gauge Charts to this format which should give us some "real" charts to look at.

PR #167, opened 2021-11-19
Fixed Grid Layout prototype

LukeTillman:
Like I mentioned in Perses chat, this is a prototype for a fixed grid layout using CSS Grid. The grid has 24 columns and a variable number of rows. This allows us to "absolutely position" a panel on the grid using `x/y/width/height` (in terms of rows/columns), allowing for more complicated layouts and empty spacing between panels in a pretty flexible way.

I also added a small layout demo sample dashboard to show some of the changes off a bit:

https://user-images.githubusercontent.com/428023/142703218-4ed028a6-a8c2-4204-97f5-ebad646f8474.mp4

One other thing I changed was adding the `title`/`collapse` properties to the grid itself, since I figured that was the primary use case we were probably going for. All of this is up for discussion, so let me know what you're thinking.


sjcobb (approved):
Really cool! It's crazy all the stuff CSS Grid can do.

Later on, we can do chart-level customizations too so they adjust when shown at smaller sizes (ECharts [media query](https://echarts.apache.org/en/tutorial.html#Responsive%20Mobile-End), [responsive chart example](https://echarts.apache.org/examples/en/editor.html?c=doc-example/pie-media)

Nexucis (approved):
really cool indeed.
I'm just wondering if `x` `y` `height` and `width` are really good name to explain what does it do.

But perhaps with an accurate documentation it would do the job

Nexucis:
@LukeTillman could you report the changes you made regarding the layout in the issue https://github.com/perses/perses/issues/55.

It's just for the followup to keep the issue up to date. Also at some point backend will need to updated with the changes did

LukeTillman:
Yeah, RE: naming, I started with something that was a lot like CSS Grid's syntax, but that didn't seem quite right. As an alternative if folks find it confusing, we would maybe do something like:

```ts
type GridPosition = {
  // Currently 'x'
  column_start: number;
  // Currently 'width'
  column_span: number;
  // Currently 'y'
  row_start: number;
  // Currently 'height'
  row_span: number;
}
```

Not sure if that's clearer 🤷‍♂️

Nexucis:
mmm I'm not sure either it's clearer haha.
Perhaps @juliusv you will have an opinion on that.

I would prefere to keep the system "0 based". But Perhaps I'm too "developper" in my head.

Ever way, I think it's not a blocking point for this PR. It can be changed later.

PR #168, opened 2021-11-20
Basic JSON conversion

LukeTillman:
Building on the CSS grid-based layout in #167, I was able to get some really basic Grafana JSON conversion working. This could eventually help whenever we get around to doing some proper conversion tooling (although I suspect we may want to rewrite it for a CLI), but in the short-term should help us with trying out different options on chart types as we iterate on those.

I went ahead and added the "full" [Node Exporter](https://grafana.com/grafana/dashboards/6014) dashboard's JSON to `sample-data` and here's what it looks like being loaded now with conversion from the JSON:

https://user-images.githubusercontent.com/428023/142740259-8c1458b6-8ab8-462e-9465-2632e74de673.mp4

I made a few changes to the UI layout so that we could use the variable selectors again (otherwise you don't get any graphs). That was just some quick/rough changes, so I definitely plan on going back and cleaning that up some more later.


sjcobb (approved):

Nexucis (approved):
O M G, this is incredible !!

Yeah for sure we will migrate this to a CLI. But thanks to yout work we won't have to think again about the convertion logic.
In my head, CLI would come a bit later, but probably we can spead up and already starting it with the command `migrate` for example.

Wondering if you could integrate already the GaugeChart ? it would be interesting to see them in action as well.

LukeTillman:
Yeah, I'll merge this into that other PR, then add Gauge Chart to the conversion code before I merge it that other PR.

PR #179, opened 2021-11-22
fetch the globalDatasources

Nexucis:
This PR is proposing a way to get the GlobalDatasources from the Perses backend.

Globally it just changes a bit how to select a datasource and refactor the datamodel of the Datasource (and the PrometheusDatasource)

It is still in progress as I would like to implement the way the frontend should use the Perses backend as a proxy to contact the Datasource (of course depending of the datasource configuration)

I'm still opening the PR to have your thought on the actual work.

Note that after this PR, using the Perses backend will be mandatory. Otherwise, the frontend won't be able to have access to the datasource definition.


sjcobb:
Looks good! I'm still trying to wrap my head around a lot of it though. 

What does this mean for local development / does anything need to be updated in CONTRIBUTING.md?

Nexucis:
no issue @sjcobb don't worry :).

Before this PR, if you didn't run the backend, you could still see the chart filled.
After this PR, if you don't run the backend, well you will have an error for each charts displayed saying it cannot get the data.

So it becomes a bit mandatory to run the backend for the development.

sjcobb (approved):

ui/core/src/model/http.ts
```diff
@@ -11,22 +11,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { DataSourceResource } from '@perses-ui/core';
+export interface BasicAuth {
```
LukeTillman, line 14: Do you feel like we can hold off on auth options until we need them? Feels a little premature to me. (I suspect we'll probably never want `BasicAuth` unless someone requests it, but 🤷‍♂️ ).

ui/core/src/model/datasource.ts
```diff
@@ -11,22 +11,23 @@
...
+  metadata: ProjectMetadata;
+  spec: AnyDatasourceSpecDefinition;
 }
 
-export type AnyDataSourceDefinition = DataSourceDefinition<string, JsonObject>;
+export interface GlobalDatasourceModel {
+  kind: 'GlobalDatasource';
+  metadata: Metadata;
+  spec: AnyDatasourceSpecDefinition;
+}
```
LukeTillman, line 33: Not sure I understand the difference between "global" and not global?

ui/app/src/context/DataSourceRegistry.tsx
```diff
@@ -30,22 +31,25 @@ export interface DataSourceRegistryProps {
  */
 export function DataSourceRegistry(props: DataSourceRegistryProps) {
   const { children } = props;
+  const { exceptionSnackbar } = useSnackbar();
+  const { data, isLoading } = useGlobalDatasourceQuery({
```
LukeTillman, line 35: I mentioned this in the meeting this morning, but do you think we should change this up to wait until `isLoading` is `false` before we start rendering children? Feels like it might simplify things a bit since we probably don't want panels calling `getDataSources` before loading has completed.

ui/prometheus-plugin/src/model/prometheus-client.ts
```diff
@@ -83,35 +83,36 @@ export function useLabelNames(
...
-  apiUrl: string,
+  datasourceSelector: DatasourceSelector,
+  apiURI: string,
   params: T,
   rename?: KeyNameMap<T>,
   queryOptions?: QueryOptions
 ) {
-  const { base_url: baseUrl } = usePrometheusConfig(dataSource);
-  const key = [baseUrl, apiUrl, params] as const;
+  const httpConfig = (<PrometheusSpecDatasource>usePrometheusConfig(datasourceSelector)).http;
```
LukeTillman, line 108: Feels like something isn't quite right if you're having to cast to `PrometheusSpecDataSource` here...

LukeTillman (approved):
Overall, looks good to me. Left a few questions/suggestions inline. 👍

ui/core/src/model/http.ts
Nexucis: The auth options are just here for the future form of the datasource in the UI. I can remove it for sure.
It's only used in the backend, but by looking about how people are protecting Prometheus, `BasicAuth` is maybe the simplest way. So pretty sure it will be required directly.
And it's already considered / handled in the Perses backend here: https://github.com/perses/perses/blob/main/internal/api/core/middleware/proxy.go#L205-L218

ui/core/src/model/datasource.ts
Nexucis: A `GlobalDatasource` is a datasource that can be used by any dashboards the different `project`.
A not global `Datasource` is a datasource scoped to a `project`. So only the dashboards that are part of the same project than the not global `Datasource` can use it.

ui/app/src/context/DataSourceRegistry.tsx
Nexucis: I'm a bit partial here.
Yes it totally makes sense that the panels panels shouldn't run the query if the datasource doesn't load. But it would still be nice to load the dashboard in order to edit it (when it will be possible).
In Grafana you have a load step before accessing the dashboard, and if this step doesn't work, then it's impossible to edit the dashboard. So I would like to avoid it.

ui/core/src/model/datasource.ts
Nexucis: For the moment the not global `Datasource`s are not supported. Since they are scoped to a project selected, that means the list will change you are changing the project.
I think it will be nice if we can change the way to get the datasources. Like it's simply fetch the datasources when you go on a dashboard page. 
If the code is fetching the `GlobalDatasources` then it put in a cache. Like that it won't be fetched everytime you are changing the dashboard, which was the objective of the datasource registering. 

Would you be ok with this approach ?

ui/app/src/context/DataSourceRegistry.tsx
Nexucis: I have the feeling it will require a refactoring of how we load / register the datasources (see my other comment about how we could manage it)

ui/prometheus-plugin/src/model/prometheus-client.ts
Nexucis: Yup totally. But I don't know how to do it better. Probably I messed up the creation of the types/interfaces ?

Nexucis:
I'm squashing this PR as I'm feeling the questions raised here shouldn't be solved in this PR but probably in a deep refactoring.
Happy to continue the discussion even if it's merged of course !

LukeTillman:
Yeah, totally cool to merge. ❤️  Sorry I didn't respond sooner. We can always refactor if necessary!

PR #203, opened 2021-11-29
Gauge chart: threshold steps reorg, spacing fix for small panels

sjcobb:
For radial gauge charts, the thresholds.steps property has been redefined so it's more intuitive. Previously, steps were defined using the same structure as EChart's [axisLine.lineStyle.color](https://echarts.apache.org/en/option.html#series-gauge.axisLine.lineStyle.color). This meant each color was mapped to the max value for a given color range.

Now, colors do not need to be set explicitly (see predefined ThresholdColors) and are mapped to the min value of the step range, example:

```js
thresholds: {
    // default_color: '#000', // optional (sets progress color, which defaults to ThresholdColors.GREEN)
    steps: [
      {
        value: 85,
        // color: '#FFA500', // optional, overrides defaultWarningColor
      },
      {
        value: 95,
        // color: '#FF0000', // optional, overrides defaultAlertColor
      }
    ]
  }
}
```
Other commits:
- convert-panels.ts convertGaugePanel function has been updated to filter out nulls and unnecessary colors
  - null was previously used in Grafana definition to set green default color 
- gauge radius and series widths / fonts adjusted to look better when used in small panels

### TODOs (for future PR)
- support min, max, and units other than percents
- allow selection of pre-defined colors from chart theme (ThresholdColors is temporary)


ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
```diff
@@ -11,20 +11,73 @@
...
-export const ThresholdColorsMap = {
-  green: 'rgb(115, 191, 105)',
-  orange: 'rgba(237, 129, 40, 0.89)',
-  red: 'rgba(245, 54, 54, 0.9)',
+import { merge } from 'lodash-es';
+
+export const ThresholdColors = {
+  GREEN: 'rgba(115, 191, 105, 1)',
+  ORANGE: 'rgba(237, 129, 40, 0.9)',
+  RED: 'rgba(245, 54, 54, 0.9)',
```
Nexucis, line 19: does these colors will change depending of the theme ?

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
```diff
@@ -11,20 +11,73 @@
...
+      },
+      {
+        value: 95,
+        color: defaultAlertColor,
+      },
+    ],
+  };
+
+  // converts to 2d array structure needed for series-gauge.axisLine.lineStyle.color
+  const stepsArr = thresholds.steps.map((step: StepOptions, index) => {
```
Nexucis, line 59: don't you want to put this function outside of this function ? I'm thinking about unit test, and this function would be a good candidate. And  it would be cool to have some

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
```diff
@@ -11,20 +11,73 @@
...
+export type StepOptions = {
   value: number;
+  color: ThresholdColorsType;
 };
+
+export function convertThresholds(
+  thresholds: ThresholdOptions = {
+    steps: [{ value: 0, color: ThresholdColors.GREEN }],
+  }
+) {
```
Nexucis, line 38: could you put what is the type returned ? I think it will help to understand what is expecting

Nexucis (commented):
looks cleaner definitely ! Thanks for doing that :).
I just have some comments / notes

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
```diff
@@ -11,20 +11,73 @@
...
+    if (index >= 1) {
+      const prevItem = arr[index - 1] || defaultThresholdArr;
+      const prevItemColor = prevItem[1];
+      const offsetItem: [number, string] = [item[0], prevItemColor];
+      return offsetItem;
+    } else {
+      const firstItem: [number, string] = [item[0], defaultThresholdColor];
+      return firstItem;
+    }
+  });
```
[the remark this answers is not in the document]
sjcobb: I updated convertThresholds to use the zip approach, you're right it's a lot better / easier to understand

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
```diff
@@ -11,20 +11,73 @@
...
+  thresholds: ThresholdOptions = {
+    steps: [{ value: 0, color: ThresholdColors.GREEN }],
+  }
+) {
+  const defaultThresholdColor = thresholds.default_color || ThresholdColors.GREEN;
+  const defaultThresholdArr: [number, string] = [0, defaultThresholdColor];
+  const defaultWarningColor = ThresholdColors.ORANGE;
+  const defaultAlertColor = ThresholdColors.RED;
+
+  const defaultThresholds: ThresholdOptions = {
```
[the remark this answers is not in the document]
sjcobb: Yeah I agree, I went ahead and made thresholds optional--it ended up being easier to do this with the zip approach instead of merging defaultThresholds like it was before.

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
sjcobb: thanks Augustin, I updated this

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
sjcobb: Not currently, this ThresholdColors implementation is temporary--I added a TODO item for a future PR when I will look into extending MUI themes to work with ECharts themes ([1](https://echarts.apache.org/en/api.html#echarts.registerTheme), [2](https://echarts.apache.org/en/theme-builder.html))

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
sjcobb: Unit test is added. Luke's zip approach suggestion reduced the number of lines in convertThresholds in half, so I left it as one function.

ui/panels-plugin/src/plugins/gauge-chart/thresholds.ts
```diff
@@ -11,20 +11,73 @@
...
   value: number;
+  color: ThresholdColorsType;
 };
+
+export function convertThresholds(
+  thresholds: ThresholdOptions = {
+    steps: [{ value: 0, color: ThresholdColors.GREEN }],
+  }
+) {
+  const defaultThresholdColor = thresholds.default_color || ThresholdColors.GREEN;
```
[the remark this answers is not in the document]
sjcobb: Thanks Luke, I fixed these and will look into adding a eslint rule

Nexucis (approved):
nice !!

Issue #183, opened 2021-11-23
Manage a sub-folder system


Nexucis:
I think it will be nice to be able to have sub-folder in the project.

For the moment, the idea I have in mind is to create an object `folder` that contains the folder tree. Something like that:

```json
{
  "kind": "Folder",
  "metadata": {
     "name": "my root folder name",
     "project": "perses"
   },
   "spec": [
     {
        "kind": "Dashboard",
        "name": "myDashboard" #a ref to the dashboard already stored
     },
     {
        "kind": "Folder",
        "name": "another sub-sub-folder",
        "spec": [
             {
                 "kind": "Dashboard",
                 "name": "anotherDashboard"
             },
            {
               "kind": "Folder",
               "name": "anotherFolder"
              "spec": [
              ]
            }
        ]
     }
   ]
}
```

like that you can have the arborescence you want and that's basically a different way to display the dashboard.
You still will be able to get all dashboards just like that `/api/v1/projects/perses/dashboards`.

shaikiam:
Upvoting.
Multiple folder structure would be very helpful to maintain a meaningful Hierarchy within the organization. Coming from Grafana this is a feature that will be a starting point while we migrate 100s of dashboards into Perses.

Nexucis:
on that topic the backend is ready since a long time now. The remaining thing to do is to be able to display it.
If someone is eager to contribute and make significant changes, that would be the one.

adrianSepiol:
> on that topic the backend is ready since a long time now. The remaining thing to do is to be able to display it. If someone is eager to contribute and make significant changes, that would be the one.

I can give it a go.

jgbernalp:
@Nexucis  Would we check the folders when a dashboard name changes?

Nexucis:
well in the perses app it is not really possible to change the perses name. But if a dashboard is deleted, I am not sure the folder pointed to this dashboard is updated.

Nexucis:
Just to add more details that can help to implement this:

From my point of view, the UI will have to get the list of the dashboards (only the metadata)  and the list of the folders, then based on that it will build the view of the dashboard list. It includes these rules:

- A dashboard can be part of multiple different folders
- If a dashboard is not part of any folders, then it should be displayed / listed at the root of the view.

adrianSepiol:
> Just to add more details that can help to implement this:
> 
[...]

PR #205, opened 2021-11-29
API: add folder management

Nexucis:
This PR is adding a way to manage a sub-folder system in the backend. In the backend it's just a document that describes a folder tree with some references to the dashboard.

It will be used by the frontend to sort the dashboard in the different folders defined.

Note: related to the issue #183 

FYI: @LukeTillman @sjcobb 

### DIFF
--- a/internal/api/core/server.go
+++ b/internal/api/core/server.go
@@ -18,6 +18,7 @@ import (
 	echoUtils "github.com/perses/common/echo"
 	"github.com/perses/perses/internal/api/impl/v1/dashboard"
 	"github.com/perses/perses/internal/api/impl/v1/datasource"
+	"github.com/perses/perses/internal/api/impl/v1/folder"
 	"github.com/perses/perses/internal/api/impl/v1/globaldatasource"
 	"github.com/perses/perses/internal/api/impl/v1/health"
 	"github.com/perses/perses/internal/api/impl/v1/project"
@@ -38,6 +39,7 @@ func NewPersesAPI(serviceManager dependency.ServiceManager) echoUtils.Register {
 	endpoints := []endpoint{
 		dashboard.NewEndpoint(serviceManager.GetDashboard()),
 		datasource.NewEndpoint(serviceManager.GetDatasource()),
+		folder.NewEndpoint(serviceManager.GetFolder()),
 		globaldatasource.NewEndpoint(serviceManager.GetGlobalDatasource()),
 		health.NewEndpoint(serviceManager.GetHealth()),
 		project.NewEndpoint(serviceManager.GetProject()),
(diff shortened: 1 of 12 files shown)


juliusv:
The PR generally seems fine to me, but introducing a feature as central as Folders is something that the other more active devs really need to give high-level input on as well.

One thing to be aware of: this introduces more references between things (previously we only had dashboards and datasources), so manually managing referential integrity without a relational database is already becoming one step more complex.

Nexucis:
> The PR generally seems fine to me, but introducing a feature as central as Folders is something that the other more active devs really need to give high-level input on as well.

Of course ! I didn't try to it as a ghost haha.

> One thing to be aware of: this introduces more references between things (previously we only had dashboards and datasources), so manually managing referential integrity without a relational database is already becoming one step more complex.

Yeah this one is introducing more manual references, but I think it is still fine and it's quite easy to manage.
But that's also why it's easy to implement.  If we were using a referential databases, it's hard to manage this kind of document. I mean having a document that have a recursive definition with multiple references.

And again, we are using a K/V database (for the moment), because then moving / supporting CRD in k8s will be an easy move. Which is not when you are using a relational database.

Yes at some points we should probably support others databases :). That can make sense

juliusv:
> But that's also why it's easy to implement. If we were using a referential databases, it's hard to manage this kind of document. I mean having a document that have a recursive definition with multiple references.

True. The current model you have here allows a dashboard to be referenced from multiple folders at the same time, kind of like a hardlink or symlink on Unix. Do you think in the end, when there's UI-based-editing features, it will also make sense to expose that kind of flexibility to the user (i.e. create a dashboard, associate it with multiple folders, almost like GMail labels)?

Nexucis:
I think we should already prevent the possibility to link a dashboard to multiple different sub-folders. It's a sort of technical limitation because it would be hard for the UI to determinate from which folder the dashboard is coming from.

You will likely have this link `https://perses-dev/project/<your_project>/folders/<folder_name>/<dashboard_name>`. Because the number of folders you can describe in this document is not limited but the URL is limited, you won't be able to put the folder tree in the URL.

So if the dashboard is referenced in multiple sub-folder, the UI won't be able to know from which folder the dashboard is coming from. 

But if the dashboard is referenced only one time, then the UI will be able to re-build the accurate `BreadCrum` based on the URL + based on the `Folder` document.

Nexucis:
I think I will implement something to be sure you will have only one references of a dashboard per document.

I'm wondering if we should already limit the depth you can have in term of folder hierarchy. And if yes what would be the limits ? 5, 10 ? Any opinion around this topic @juliusv @sjcobb  @LukeTillman ?

juliusv:
For now I wouldn't worry about limiting the depth, it's something you can get back to if / when needed IMO.

PR #206, opened 2021-12-03
Small Plugin Cleanup

LukeTillman:
This is a small clean-up to the plugin types to remove the `Kind` type parameter that's being used everywhere. It's not actually adding anything useful since `Kind` will always just be a `string` at runtime when dealing with it in the Perses app code and so it was making those types a lot more verbose and confusing than they actually needed to be.

I also cleaned up a few other small things and I'll try and leave comments inline to describe what's up. I think I'll have a slightly larger refactoring of those plugin types coming soon that should really clean things up and make them less confusing and verbose.


dev/data/globaldatasource.json
```diff
@@ -12,6 +12,20 @@
       }
     }
   },
+  {
```
LukeTillman, line 15: Just adding a browser data source to the sample data here...

ui/app/src/context/plugin-registry/registry-state.ts
```diff
@@ -21,6 +21,7 @@ import {
   PluginType,
   AnyPluginImplementation,
   JsonObject,
+  ALL_PLUGIN_TYPES,
```
LukeTillman, line 24: This change to have a runtime list of all plugin types simplifies some of the registry code below so that now there are fewer places to go make changes whenever we add a new plugin type.

ui/app/src/index.tsx
```diff
@@ -20,7 +20,7 @@ import App from './App';
 import { createTheme } from './styles/theme';
 import { SnackbarProvider } from './context/SnackbarProvider';
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({ defaultOptions: { queries: { refetchOnWindowFocus: false } } });
```
LukeTillman, line 23: This just turns off the default `react-query` behavior that was causing PromQL queries, etc. to be refetched anytime you focus the browser window.

ui/app/webpack.common.ts
```diff
@@ -35,6 +35,7 @@ export const commonConfig: Configuration = {
     new ForkTsCheckerWebpackPlugin({
       typescript: {
         configFile: path.resolve(__dirname, './tsconfig.json'),
+        build: true, // Since we use project references...
```
LukeTillman, line 38: This _should_ fix the error that we see sometimes locally where TypeScript complains that "XXX has not been built from source file YYY".

ui/app/webpack.dev.ts
```diff
@@ -60,6 +60,7 @@ const devConfig: Configuration = {
     port: parseInt(process.env.PORT ?? '3000'),
     open: true,
     https: getHttpsConfig(),
+    http2: process.env.HTTP2 === 'true',
```
LukeTillman, line 63: Allows enabling HTTP2 for the webpack dev server (like I mentioned in chat).

sjcobb (approved):

ui/tsconfig.base.json
```diff
@@ -11,8 +11,7 @@
     "isolatedModules": true,
     "noUncheckedIndexedAccess": true,
     "composite": true,
-    "paths": {
-      "@perses-ui/*": ["./*/src"]
```
Nexucis, line 15: the unit tests are failing. I'm wondering if it's related to this change.

I'm surprised we don't need it because when I worked around this topic and the possibility to run `npm start`, this part was the key to make it works.

Nexucis (commented):
Nice :). Thanks for doing that. Looking forward for the big refactoring. I think it will be a really nice thing to do.

I'm a a bit worried about the tsconfig you changed and maybe it can explain why the unit tests are failing.

ui/app/src/index.tsx
Nexucis: cool ! That was simple at then. Nice catch !

Do you know what this behavior is the one by default ? It's a bit odd to have it by default, but probably I miss some context

LukeTillman: Yeah, I felt like it was a weird choice to have by default too, so don't think you're missing anything. The docs have a bit of an explanation of some of the defaults:

https://react-query.tanstack.com/guides/important-defaults

ui/tsconfig.base.json
LukeTillman: Oh man, this sent me down a rabbit hole. 😂 

OK, so you're right, this change _is_ why the tests are failing to compile. But with Typescript project references, this setting isn't actually supposed to be necessary. With project references, the `references` config in `tsconfig.json` points to the other projects that a given package depends on. Then when you run `tsc --build` (you have to use `--build` to get project references to compile correctly), the Typescript compiler takes into account the `references` config and builds those projects first.

The problem is that `jest` and `ts-jest` in particular doesn't currently deal with project references. Since it's a jest transformer that runs "on-demand" for individual files, it doesn't actually have a way to use the `--build` flag that would take care of building project references first. There's an [open issue in ts-jest](https://github.com/kulshekhar/ts-jest/issues/1648) currently where they're discussing how to deal with project references and has some workarounds.

I'm going to think about this a bit more, but I think for now, I may just use one of the workarounds (e.g. `moduleNameMapper`) suggested in that issue...

LukeTillman:
Sorry about the CI noise if anyone else is getting notified, trying out a few things.

ui/tsconfig.base.json
LukeTillman: I ended up bringing this config back in a `tsconfig.test.json` for running unit tests. We'll see what `ts-jest` ends up ultimately doing to better support project references.

PR #227, opened 2021-12-08
LineChart: echarts trigger item tooltip, uplot custom tooltip

sjcobb:
When comparing our LineChart implementation with ECharts benchmarks in Chronosphere's UI, I noticed some performance discrepancies. The main issue is [a bug](https://github.com/apache/echarts/issues/14305) in how `xAxis.type: 'time'` works with axisLabel intervals, which results in symbols always rendering since [series-line.showAllSymbol](https://echarts.apache.org/en/option.html#series-line.showAllSymbol) relys on that interval strategy to determine when to hide symbols. This meant line charts with 1500+ series were taking ~30 seconds to load--_now ECharts is competitive with uPlot in performance again_.

Previously, the workaround here was to hide symbols using `symbol: 'none' or showSymbol: false`, but this is no longer an option since [trigger: 'item'](https://echarts.apache.org/en/option.html#tooltip.trigger) tooltips need symbols to be visible in order to show tooltip content on hover (and trigger 'axis' tooltips can't be used with so many series). LineChart now uses the more stable 'category' xAxis.type implementation, as well as passing data in series.data instead of dataset. Other proposed performance improvements are to enable the LTTB [downsampling property](https://echarts.apache.org/en/option.html#series-line.sampling).

See: https://localhost:3000/?dashboard=benchmarkDashboard

echarts_tooltip_screenshot

**TODO**: 
- tooltip content: fix metric name showing beside marker 
- show closest series in tooltip even when not focused on data point ([this feature](https://github.com/apache/echarts/issues/15080) is on EChart's roadmap already)
- separate PR with uPlot tooltip for comparison

### DIFF
--- a/ui/app/sample-data/benchmarkDashboard.ts
+++ b/ui/app/sample-data/benchmarkDashboard.ts
@@ -23,7 +23,6 @@ const benchmarkDashboard: DashboardResource = {
   },
   spec: {
     datasource: { kind: 'Prometheus', name: 'PrometheusDemo', global: true },
-    // TODO: Should duration actually be a time range?
     duration: '24h',
     variables: {
       job: {
@@ -103,9 +102,9 @@ const benchmarkDashboard: DashboardResource = {
           unit: { kind: 'Bytes' },
         },
       },
-      filesystemFullness: {
+      basicEx: {
         kind: 'LineChart',
-        display: { name: 'Filesystem Fullness' },
+        display: { name: 'Single Query' },
         options: {
           queries: [
             {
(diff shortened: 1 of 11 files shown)


Nexucis:
Oh we don't need this PR anymore ?

I know I'm a bit late in the review for Perses, my apologies for that.

sjcobb:
No problem! I should have commented before--found a better approach that doesn't require refactoring the xAxis type and gives us more control over customization. I just need to clean up my branch a little bit so I'll submit that PR after the holidays. Hope you enjoy your break!

PR #266, opened 2022-01-05
echarts and uplot custom tooltips / focused series logic

sjcobb:
For better performance and more control over focused series logic / tooltip functionality, this PR implements custom tooltips using EChart's zrender coordinates. I also refactored the uPlot tooltips from [my previous PR](https://github.com/perses/perses/pull/227) to share most of the same React components. 

Bonus features:
- Show non-truncated series name when hovering inside tooltip
- Resizable tooltip (in addition to being able to enter and scroll inside tooltip)

_screen recording from: https://localhost:3000/?dashboard=benchmarkDashboard_
[tooltip_demo.mp4](https://user-images.githubusercontent.com/9369625/148558753-79fb6822-ddc6-48f8-b7c0-983c4556ec70.mp4
)

screenshot when hovering over series name:

tooltip_hover_screenshot

PR #285, opened 2022-01-28
Update tooltip visual style and layout

jhardy:
This PR updates the line tooltip with the following modifications:
 -  use a `display: table` layout strategy 
 - Update overall visual style
 - Update display when there is one series 
 - Ability to wrap labels via a prop
 

image

image


ui/panels-plugin/src/plugins/line-chart/tooltip/TooltipContent.tsx
```diff
@@ -11,39 +11,64 @@
...
+      {seriesTime && <Typography variant="caption">{formatTimeSeriesHeader(seriesTime)}</Typography>}
+      <Divider sx={{ borderColor: 'grey.800' }} />
+      <Box
+        sx={{
+          display: 'table',
+        }}
+      >
+        {focusedSeries.map(({ datumIdx, seriesIdx, seriesName, y, markerColor }) => {
+          if (datumIdx === null || seriesIdx === null) return null;
+          const key = seriesIdx.toString() + datumIdx.toString();
```
jhardy, line 50: @sjcobb I still need to resolve this commented out section, my goal here was to keep all the `SeriesInfo` elements grouped together to have more consistent display of them, but I am not sure if my approach is the best. Would love any feedback here.

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
```diff
@@ -13,13 +13,14 @@
...
-export const TOOLTIP_MIN_WIDTH = 360;
+export const TOOLTIP_MAX_WIDTH = 700;
+export const TOOLTIP_MAX_HEIGHT = 600;
 
 export const TOOLTIP_MAX_ITEMS = 50;
 
 export const TOOLTIP_DATE_FORMAT = new Intl.DateTimeFormat(undefined, {
   year: 'numeric',
-  month: 'numeric',
+  month: 'short',
```
jhardy, line 23: any objections to using the `short` format here? I think it reads nice over all, but would be good to hear the opinions of others.

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
```diff
@@ -13,13 +13,14 @@
 
 import { FocusedSeriesArray } from '../utils/focused-series';
 
-export const TOOLTIP_MIN_WIDTH = 360;
+export const TOOLTIP_MAX_WIDTH = 700;
+export const TOOLTIP_MAX_HEIGHT = 600;
```
jhardy, line 18: I updated the max width and max height considerably, but I feel like this is a better default strategy than allowing user resizing.

ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx
```diff
@@ -28,14 +28,17 @@ function assembleTransform(coords: Coordinate, chartWidth: number) {
 
 interface TooltipProps {
   tooltipData: TooltipData;
+  /**
+   * Specify if the labels should wrap instead of truncate.
+   */
+  wrapLabels?: boolean;
```
jhardy, line 34: I was debating whether or not to include label wrapping in this PR, so if out of scope or if we want to take another approach I am open to removing this for now.

ui/panels-plugin/src/plugins/line-chart/tooltip/SeriesMarker.tsx
```diff
@@ -22,19 +22,16 @@ function SeriesMarker(props: SeriesMarkerProps) {
...
-    >
-      <Box
-        sx={{
-          width: '10px',
-          height: '10px',
-          backgroundColor: markerColor,
-        }}
-      ></Box>
-    </Box>
+      style={{ backgroundColor: markerColor }}
```
jhardy, line 33: I moved the `markerColor` to an inline style to avoid excessive class generation by styled components.

juliusv:
Beautiful! I still think that at least for the single-series view it would be nicer to display the labels like in Prometheus (bold label name, colon instead of `=`, losing quotes around the value - just makes stuff more readable IMO): https://demo.promlabs.com/graph?g0.expr=histogram_quantile(0.9%2C%20rate(demo_api_request_duration_seconds_bucket%5B5m%5D))&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
juliusv: I like it!

ui/panels-plugin/src/plugins/line-chart/tooltip/SeriesMarker.tsx
juliusv: Ahh, was that what was causing a new class on the element every time you moved the mouse by like 1px?

sjcobb:
Looks great! An edge case I noticed in seriesTestAlt from [benchmarkDashboard](https://localhost:3000/?dashboard=benchmarkDashboard) is when you hover close to 0, you can't always see the values without scrolling right slightly and when the charts first load, before activating the tooltip, I can sometimes see 24px empty box (see [screen capture](https://user-images.githubusercontent.com/9369625/151873183-739b615e-4dc6-49a6-8425-ad56193c7711.mp4))

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
sjcobb: Yeah for sure, the user resizing thing was a dead-end that I should have just removed at the time :smile:

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
sjcobb: Looks good to me!

ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx
sjcobb: Sounds good to be a separate PR / something we iterate on later. 

I am pleasantly surprised with how the wrapped labels look in Figma. My opinion would be to reduce TOOLTIP_MAX_WIDTH some and allow labels to wrap to two lines (then in the rare case that it needs to wrap to three lines, they get truncated). But definitely go with whatever you think is best, this looks way better than what I came up with!

ui/panels-plugin/src/plugins/line-chart/tooltip/TooltipContent.tsx
sjcobb: For context, one thing I had been playing around with was increasing xBuffer to show more nearby points. So potentially multiple dates could show in the tooltip (which required sorting differently). I think that is overkill though so all the lastDate stuff can probably be removed.

ui/panels-plugin/src/plugins/line-chart/tooltip/SeriesInfo.tsx
```diff
@@ -11,59 +11,101 @@
...
-            margin: '0 0 2px',
             display: 'flex',
             flexDirection: 'row',
             alignItems: 'center',
             justifyContent: 'left',
           }}
         >
           <SeriesMarker markerColor={markerColor} />
-          <Box component="span">value: {y}</Box>
+          <Box component="span" color="grey.300">
```
Nexucis, line 41: Wouldn't be a good idea to have these colors not harcoded ? I'm thinking when we will have a theme in place, these hardcoded color could be an issue no ?

ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx
Nexucis: yeah I think it will be interesting to do it in another PR. Like that it will be easy to see the before/after the changes

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
Nexucis: It looks even better like that ! So cool thanks for doing that :)

ui/panels-plugin/src/plugins/line-chart/tooltip/TooltipContent.tsx
Nexucis: Is it something that should be solved before merging this PR, or we could keep it for later ?

Nexucis (commented):
Pretty cool @jhardy ! Thank you for doing this :).

I have a weird behaviour. When I have the cursor on a graph, the tooltip is shown as expected, but when I'm leaving it, I have these small black screen that appears.

I didn't test on the master branch, so perhaps it's not coming with these changes.

image

Nexucis:
also I don't know if you saw it, but the unit tests are failing + you have to sign your commits @jhardy  (so the DCO can be green)

Nexucis:
Last thing, is it normal that the tooltip doesn't occupy the whole space allocated to it ?

image

In the above screen, we can see that we have some white border on the right and at the bottom of the tooltip.

ui/panels-plugin/src/plugins/line-chart/tooltip/SeriesInfo.tsx
sjcobb: I'll take care of this, I have a local branch where I started playing around with dark/light theme

ui/panels-plugin/src/plugins/line-chart/tooltip/TooltipContent.tsx
sjcobb: Yeah it looks mostly done already, mainly just removing the commented out block

juliusv:
Hmm, on Chrome / Linux it looks like this for me, with currently non-active hover details lurking around in the top left of a graph and the active one having both X and Y scroll bars:

hover-detail-glitches

juliusv:
If I remove the `overflow: scroll` from https://github.com/perses/perses/blob/6577dcd8e5a6f7bf6ca35118b56990a11180463c/ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx#L62, this gets rid of the scrollbars at least.

Nexucis:
ah yeah I forgot to mention I tested on MacOS / Chrome

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
```diff
@@ -45,7 +46,7 @@ export const defaultCursorData = {
 
 export const emptyTooltipData = {
   cursor: defaultCursorData,
-  focusedSeries: [{ seriesIdx: null, datumIdx: null, date: '', seriesName: '', x: 0, y: 0, markerColor: '' }],
+  focusedSeries: null, //[{ seriesIdx: null, datumIdx: null, date: '', seriesName: '', x: 0, y: 0, markerColor: '' }],
```
jhardy, line 49: @sjcobb to prevent the tooltip from rendering I updated `focusedSeries` to be null by default. Before I delete the old default I wanted to get your take 

It solves this little black square showing up cc @Nexucis  image

ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx
```diff
@@ -54,20 +53,20 @@ function Tooltip(props: TooltipProps) {
...
-          width: TOOLTIP_MIN_WIDTH,
-          height: resizeDir === 'vertical' ? '180px' : 'auto',
+          borderRadius: '6px',
+          maxWidth: TOOLTIP_MAX_WIDTH,
+          maxHeight: TOOLTIP_MAX_HEIGHT,
           position: 'absolute',
           top: 0,
           left: 0,
-          overflow: 'scroll',
+          overflow: 'hidden',
```
jhardy, line 62: Fixes the scroll bar issues that @Nexucis and @juliusv were seeing

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
sjcobb: Looks good! Not sure why I did it that way in the first place

ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx
```diff
@@ -13,7 +13,7 @@
 
 import { Box } from '@mui/material';
 import { useEffect, useState } from 'react';
-import { Coordinate, TooltipData, TOOLTIP_MIN_WIDTH } from './tooltip-model';
+import { Coordinate, TooltipData, TOOLTIP_MAX_HEIGHT, TOOLTIP_MAX_WIDTH } from './tooltip-model';
 import TooltipContent from './TooltipContent';
 
 function assembleTransform(coords: Coordinate, chartWidth: number) {
```
sjcobb, line 19: Since TOOLTIP_MAX_WIDTH has increased, how `flipTooltipPosThreshold` is calculated may need to be adjusted so the tooltip doesn't get cut off by surrounding components like OptionsDrawer. Definitely not a blocker for this PR, I can take care of it when I look into using Portals so it attaches to the body (see [Luke comment](https://github.com/perses/perses/pull/266#discussion_r781511724) for reference)

ui/panels-plugin/src/plugins/line-chart/tooltip/SeriesInfo.tsx
```diff
@@ -21,49 +21,112 @@ interface SeriesInfoProps {
...
+          maxWidth: '520px',
+        }}
+      >
+        <SeriesMarker markerColor={markerColor} />
+        <Box
+          component="span"
+          sx={{
+            color: 'grey.300',
+            display: 'inline-block',
+            maxWidth: '550px',
```
sjcobb, line 109: Would it make sense to calculate this maxWidth relative to TOOLTIP_MAX_WIDTH (in case that constant changes in the future)?

sjcobb (approved):

ui/panels-plugin/src/plugins/line-chart/tooltip/Tooltip.tsx
jhardy: I agree, there needs to be some adjustments here, I wonder if it's some combination of tooltip container size, mouse position and browser window width  & height ?

sjcobb: Yeah that's true, so many factors!

sjcobb (approved):

ui/panels-plugin/src/plugins/line-chart/tooltip/tooltip-model.ts
Nexucis: perfect !

Nexucis:
I found a corner case, when you are displaying multiple timeseries in the tooltip, it's a long timeries and it goes behind the variables bar.
Perhaps it's not a big issue if we think the place of the variables are going to change.

image

Nexucis (approved):
based on what we discussed yesterday, I think we can merge it now.

PR #297, opened 2022-02-02
Add Login cmd to CLI

Nexucis:
This PR is adding a `login` command to the CLI.
Currently it's just used to save the URL that would be used later to contact the Perses' API.

The URL is saved in a file that would be stored in ${USERHOME}/.perses/config


Nexucis:
Closing this PR as I will have another one which might me more coherent (and a bit bigger unfortunately) that this one

PR #307, opened 2022-02-03
stat chart add sparkline, reorg charts, enhance formatting

sjcobb:
Overview:
- reorg gauge-chart and stat-chart to new components folder structure 
- pass query data from StatChartPanel, GaugeChartPanel
- rework stat chart to use echarts, support sparklines
- initial echarts wrapper
- Skeleton loading state in GaugeChartPanel and StatChartPanel
- install @formatjs/intl-numberformat for future unit conversion
  - support [units from ECMAScript standard](https://tc39.es/proposal-unified-intl-numberformat/section6/locales-currencies-tz_proposed_out.html#sec-issanctionedsimpleunitidentifier)
- add echarts media query to adjust textStyle.fontSize
- decimalFormatter using Intl.NumberFormat
- default unitOptions.unitDisplay to narrow
- stat chart examples in benchmarkDashboard, nodeExporterDashboard

Responsive example:
stat-chart-sparkline-ex


sjcobb:
Current state of node-exporter-full_rev23.json conversion:
node-exporter-conversion

ui/app/src/views/dashboard/Panel.tsx
```diff
@@ -29,7 +29,8 @@ export interface PanelProps extends CardProps {
 function Panel(props: PanelProps) {
   const { definition, ...others } = props;
   const [contentElement, setContentElement] = useState<HTMLDivElement | null>(null);
-
+  const showPanelHeader = definition.display.show_panel_header ?? true;
```
sjcobb, line 32: Went with this so StatChart's backgroundColor could extend to the edge of the panel without any whitespace (and so display.name isn't duplicated), thoughts on a better name or approach?

ui/package.json
```diff
@@ -17,6 +17,7 @@
     "panels-plugin"
   ],
   "devDependencies": {
+    "@formatjs/intl-numberformat": "^7.4.1",
```
sjcobb, line 20: Could potentially remove this dependency since I only ended up using one method that validates against [unit identifier standard](https://tc39.es/proposal-unified-intl-numberformat/section6/locales-currencies-tz_proposed_out.html#sec-issanctionedsimpleunitidentifier), but there may be other [future use-cases for it](https://github.com/formatjs/formatjs)

ui/panels-plugin/src/components/echarts-wrapper/EChartsWrapper.tsx
```diff
@@ -0,0 +1,102 @@
...
+export type onEventFunction = () => void;
+
+export type onEventsType = Record<string, onEventFunction>;
+
+export interface EChartsWrapper {
+  sx: SxProps<Theme>;
+  option: EChartsCoreOption;
+  theme?: string;
+  onChartReady?: (instance: ECharts) => void;
+  onEvents?: onEventsType;
```
sjcobb, line 29: Took part of this from [echarts-for-react](https://github.com/hustcc/echarts-for-react/), planning to improve these optional props when I refactor LineChart to use EChartsWrapper and implement chart themes

ui/app/src/views/dashboard/Panel.tsx
sjcobb: As discussed on call, planning to remove this and instead use theme to customize.

LukeTillman: 👍

ui/package.json
LukeTillman: I'm cool with keeping it, but is this a root dev dependency? Feels like this actually belongs as a dependency of a specific package?

ui/panels-plugin/src/components/echarts-wrapper/EChartsWrapper.tsx
```diff
@@ -0,0 +1,101 @@
...
+    for (const eventName in events) {
+      if (Object.prototype.hasOwnProperty.call(events, eventName)) {
+        const customEvent = events[eventName] ?? null;
+        if (customEvent) {
+          bindEvent(eventName, customEvent);
+        }
+      }
+    }
+    // TODO (sjcobb): unbind events
+  }
```
LukeTillman, line 70: Does this need to be inline in the component? I'm not 100% sure, but it doesn't look like it closes over anything from the component and could just be extracted outside the component body?

ui/panels-plugin/src/components/gauge-chart/GaugeChart.tsx
```diff
@@ -54,23 +45,25 @@ const noDataOption = {
...
-function GaugeChart(props: GaugeChartProps) {
-  const { query, width, height, calculation, unit } = props;
-  const thresholds = props.thresholds ?? defaultThresholdInput;
-  const { data } = useGraphQuery(query);
+export type GaugeChartData = number | null | undefined;
 
-  const option: EChartsOption = useMemo(() => {
-    // TODO (sjcobb): add loading spinner, share noDataOption with other charts
-    if (data === undefined) return {};
+interface GaugeChartProps extends JsonObject {
```
LukeTillman, line 50: Love the more-specific chart props and component. 👍

ui/panels-plugin/src/model/units.ts
```diff
@@ -11,9 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { IsSanctionedSimpleUnitIdentifier } from '@formatjs/ecma402-abstract';
```
LukeTillman, line 14: Holy function name, Batman! 😂

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
```diff
@@ -24,29 +26,57 @@ export type GaugeChartPanelProps = PanelProps<GaugeChartOptions>;
...
+    return value;
+  }, [data, calculation]);
+
+  if (error) throw error;
+
+  if (contentDimensions === undefined) return <></>;
+
+  if (loading === true) {
+    return (
+      <Skeleton
```
LukeTillman, line 63: Nice use of Skeleton. 👍

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
```diff
@@ -11,45 +11,101 @@
...
+  name: string;
+  show_panel_header: string;
   query: AnyGraphQueryDefinition;
   calculation: CalculationType;
-  unit?: UnitOptions;
+  unit: UnitOptions;
+  thresholds?: ThresholdOptions;
+  sparkline?: {
+    show: boolean;
+  };
```
LukeTillman, line 42: Curious, what other options do you envision for `sparkline`? Would be nice if we didn't also need this boolean `show` property (i.e. if `sparkline` has an options object, then we infer that it should be shown rather than needing a separate boolean flag).

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
```diff
@@ -11,45 +11,101 @@
...
+  const chartData = useChartData(data, calculation, name, showName);
 
-  const displayValue = useMemo(() => {
-    if (data === undefined) return 'No data';
+  if (error) throw error;
 
-    // TODO: something smarter with iterable?
-    const series = Array.from(data.series)[0];
-    if (series === undefined) return 'No data';
+  if (contentDimensions === undefined) return <></>;
```
LukeTillman, line 61: This should probably just be `return null`?

LukeTillman (approved):
Overall, I think this looks great. Should definitely bring back the panel header/remove that option and I left a few questions inline. Nice work!

ui/panels-plugin/src/components/echarts-wrapper/EChartsWrapper.tsx
sjcobb: Yeah good call, I moved this

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
sjcobb: Yeah true, I'll change this so an empty sparkline object is enough to show the line series and then [default colors](https://github.com/perses/perses/blob/61be6d5a3e3747818a98ba1d4f09b6b4114af979/ui/panels-plugin/src/components/stat-chart/StatChart.tsx#L79) can be overriden using something like: 
```
sparkline: {
  line_color?: string;
  line_width?: number;
  area_color?: string; // defaults to same as line_color
  // add more as needed
};
```
Which in StatChart would translate to [lineStyle.color](https://echarts.apache.org/en/option.html#series-line.lineStyle.color), lineStyle.width, areaStyle.color, etc. Also, any thoughts @jhardy?

ui/package.json
sjcobb: Whoops, just moved to panels-plugin's package.json!

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
sjcobb: I [went ahead with the options above](https://github.com/perses/perses/pull/307/commits/a275bbd95597713be0b89c5ca1d770382e846f10) and also added line_opacity and area_opacity. I have some ideas about panel-level vs. chart-level props that I want to discuss with the team before going any further.

PR #308, opened 2022-02-07
Handle config in CLI with Login and Project command

Nexucis:
This PR aims to introduce two new commands: 
* login: `percli login http://localhost:8080`. For the moment it's just used to indicate which URL the CLI should use for the later commands. Later probably it will integrate a way to ask to the API for a valid token
* project: `percli project`. This command will be used to store the default project to be used for the later commands.

These two commands required the introduction of a config management. The CLI config is stored in `${USERHOME}/.perses/config`

I also took the opportunity to print the version of the server when you use the command `percli version`. 

Finally, the little script that is populating the database is able to use podman instead of docker (in case like me you had to switch to podman).


internal/cli/cmd/login/login.go
```diff
@@ -0,0 +1,71 @@
...
+			InsecureTLS: o.insecureTLS,
+		},
+	})
+}
+
+func NewCMD() *cobra.Command {
+	o := &option{}
+	cmd := &cobra.Command{
+		Use:   "login URL",
+		Short: "Lo in to the Perses' API",
```
juliusv, line 58:
```suggestion
		Short: "Log in to the Perses API",
```

internal/cli/cmd/login/login.go
```diff
@@ -0,0 +1,71 @@
...
+			URL:         o.url,
+			InsecureTLS: o.insecureTLS,
+		},
+	})
+}
+
+func NewCMD() *cobra.Command {
+	o := &option{}
+	cmd := &cobra.Command{
+		Use:   "login URL",
```
juliusv, line 57: I'd do `login [URL]` like in the example in https://github.com/spf13/cobra/blob/master/user_guide.md#example, so that it's clear that it's a placeholder.

internal/cli/cmd/login/login.go
```diff
@@ -0,0 +1,71 @@
...
+# Log in to the given server
+percli login https://perses.dev
+`,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdUtils.HandleError(o.complete(args))
+			cmdUtils.HandleError(o.validate())
+			cmdUtils.HandleError(o.execute())
+		},
+	}
+	cmd.Flags().BoolVar(&o.insecureTLS, "insecure-skip-tls-verify", o.insecureTLS, "if true the server's certificate will not be checked for validity. This will make your HTTPS connections insecure")
```
juliusv, line 69: If we're putting multiple sentences in flag descriptions, it's better if we start each sentence capitalized and end both with a period (and then it probably makes sense to do that for all flags).

internal/cli/cmd/project/project.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type option struct {
+	projectName string
+	apiClient   v1.ProjectInterface
+}
+
+func (o *option) complete(args []string) error {
+	if len(args) > 1 {
+		return fmt.Errorf("only the project's can be specified as an argument")
```
juliusv, line 33:
```suggestion
		return fmt.Errorf("only the project can be specified as an argument")
```

internal/cli/utils/config.go
```diff
@@ -0,0 +1,142 @@
...
+		Project: project,
+	})
+}
+
+// WriteConfig writes the configuration file in the path {USER_HOME}/.perses/config
+// if the directory doesn't exist, the function will create it
+func WriteConfig(config *CLIConfig) error {
+	path := filepath.Join(getRootFolder(), pathConfig)
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
juliusv, line 112: Don't you need to check errors other than `is.IsNotExist()` here too? Like if the stat just completely failed...

juliusv (commented):
Just a quick review, good to merge IMO for now after those comments...

internal/cli/utils/config.go
Nexucis: ah yup, indeed I forgot to check the rest of the errors. Thanks !

internal/cli/cmd/login/login.go
Nexucis: Yup totally makes sense. And actually it is already the case for the others. So it's missed on my side.

internal/cli/cmd/login/login.go
Nexucis: yup. I'm just wondering for later how we could say that a placeholder is optional.

Nexucis:
thanks for the review @juliusv!

PR #314, opened 2022-02-28
Manage properly the CLI error and implement unit test for the CLI

Nexucis:
This PR is changing and introducing different things:
1. First it changes the way to manage the error in the CLI. Basically all errors are reported to the root CMD which is able to handle and print the error in the correct file descriptor.
2. It introduces a way to inject an `io.writer` in the cmd so we can intercept what is printed on stdout / stderr and verify it is the correct message (for testing purpose)
3. It provides also a fake http client that should be used when you want to mock the connection to the API. It's easier to use than to use an actual mock system.
4. It also provides an interface that each command must implement.

### DIFF
--- a/internal/cli/utils/test/test.go
+++ b/internal/cli/utils/test/test.go
@@ -50,10 +50,11 @@ type Suite struct {
 	IsErrorExpected bool
 }
 
-func ExecuteSuiteTest(t *testing.T, cmd *cobra.Command, suites []Suite) {
+func ExecuteSuiteTest(t *testing.T, newCMD func() *cobra.Command, suites []Suite) {
 	for _, test := range suites {
 		t.Run(test.Title, func(t *testing.T) {
 			buffer := bytes.NewBufferString("")
+			cmd := newCMD()
 			cmd.SetOut(buffer)
 			cmd.SetErr(buffer)
 			cmd.SetArgs(test.Args)
--- a/internal/cli/utils/test/test.go
+++ b/internal/cli/utils/test/test.go
@@ -1,0 +1,73 @@
+// Copyright 2022 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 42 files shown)


internal/cli/utils/test/test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/perses/perses/pkg/client/api"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v2"
+)
+
+func JSONMarshalStrict(obj interface{}) []byte {
+	if data, err := json.Marshal(obj); err != nil {
+		panic(err)
+	} else {
```
juliusv, line 31: Omit the else when possible after error handling, see https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

internal/cli/utils/test/test.go
```diff
@@ -0,0 +1,74 @@
...
+		return data
+	}
+}
+
+func YamlMarshalStrict(obj interface{}) []byte {
```
juliusv, line 36: `Yaml` -> `YAML` (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/client/fake_api/fake_v1/folder.go
```diff
@@ -0,0 +1,85 @@
...
+	modelV1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+func FolderList(project string, prefix string) []*modelV1.Folder {
+	initialList := []*modelV1.Folder{
+		{
+			Kind: modelV1.KindFolder,
+			Metadata: modelV1.ProjectMetadata{
+				Metadata: modelV1.Metadata{
+					Name: "FF15",
```
juliusv, line 29: Final Fantasy 15?

juliusv (commented):
:+1: besides nits

pkg/client/fake_api/fake_v1/folder.go
Nexucis: haha yes :).  I was searching some inspirations ^^

Nexucis:
thanks @juliusv !

PR #315, opened 2022-03-01
Reorganize UI into new @perses-dev packages

LukeTillman:
First, sorry about this being a giant PR. Unfortunately, reorganizing things into new packages also involved rewriting the APIs of a few things along the way and if you add on changing the package names from `@perses-ui` to `@perses-dev`, you end up with something like this. 🙃 

At a high level, here's what this does:

1. Switches our npm organization to `@perses-dev` and sets all package versions to `0.1.0` (instead of the `1.0.0` used for the old packages, which was definitely a mistake).
2. Reorganizes the existing code to better support embedding scenarios for outside consumers while at the same time, still allowing us to have an app.
3. Rewrites some of the public APIs offered by those components to reflect that they might be reused outside of the Perses app.

At this point, here's what the package structure looks like now and what I think each package represents:

- `app`: still the overall Perses application where we stitch everything else together. I anticipate this containing routing, fetching, and other logic specific to the overall application while composing components from other packages for features.
- `components`: common components available to the app, plugins, or people who want to embed common Perses UI elements into their own applications. This is pretty sparse right now, but I think the next step is to move some of our common charting components in here so they can work outside of plugins.
- `core`: still common model definitions, utilities, etc. that are used by both the app and plugins
- `dashboards`: the dashboards feature with a bunch of components exported that make it possible for folks to reuse in their own applications. You could (in theory) reuse the entire Dashboard view, just the Dashboard's content (without variable inputs), just a Panel, or even just the grid layout "engine".
- `panels-plugin`: still the plugins for our common visualizations. As mentioned above, I think we should consider moving some common chart types into the `components` package so people can use our charts (e.g. Prometheus!) without needing the whole plugin system.
- `plugin-system`: All the type definitions and components that power the plugins we support. Plugin authors would consume this package for the type definitions while consumers could use this to host Perses plugins in their own applications if so desired. This also includes the definitions for the runtime available to plugins (e.g. the current time range state, the current template variable state), but doesn't include the implementations of those.
- `prometheus-plugin`: still the plugins for loading data from Prometheus.

Overall, this is already a pretty huge change and there are still a ton more things that need follow-ups, but I want to do those in smaller PRs that build on top of this big change. Here are a few things that I plan to follow-up on:

1. Move some of the "dumb" charting components out of `panels-plugin` and into `components` so that someone can use our charts without being coupled to our plugin system/data fetching logic.
2. Fix the PluginRegistry API so it's possible to render a dashboard while the list of installed plugins is still loading rather than waiting for it to load.
3. Consider changing up some of the plugin APIs so that we control loading state rather than the plugin (i.e. we can use `react-query` in Perses code, not in the plugins themselves, and they just provide the async function, similar to how the PluginRegistry works in this PR).
4. Flesh out/discuss DataSources in more detail since I don't think we've quite nailed down how those should work or how they fit into the overall plugin architecture.
5. Figure out how much to automate the package building/publishing workflow as part of CI/CD and implement that automation.
6. Finally start putting some tests around these newly organized components. I think it would be also cool to have a few examples (maybe including cypress tests against those examples/demos) of how to embed a few common components into an application other than its use in the Perses app.

If it's helpful, happy to host a Zoom walkthrough with other folks as part of the review process here since I know this a lot to look at. Hopefully this is the last of any giant reorgs like this. 🙏


ui/.eslintrc.base.js
```diff
@@ -13,7 +13,13 @@
 
 // Base eslint configuration for typescript projects
 module.exports = {
-  extends: ['plugin:@typescript-eslint/recommended', 'plugin:prettier/recommended'],
+  extends: [
```
LukeTillman, line 16: We used to have two "base" configs for eslint while we really only needed one, so I just condensed them down early on into a single config. (Sorry about the additional noise in this PR).

Nexucis:
> If it's helpful, happy to host a Zoom walkthrough with other folks as part of the review process here since I know this a lot to look at. Hopefully this is the last of any giant reorgs like this. 🙏

On my side, yeah I would be happy to have a short "demo" / explanation about what you did and/or about what does it solve, what does it improve, how it will make our life easier after that ...etc. For example, as a tiny react developer, if tomorrow I want to connect the frontend to the backend regarding the dashboard, I would like to know if it's easier now :).
Because it wasn't today ^^

Nexucis:
And unfortunately you have a conflict in this PR @LukeTillman :(, and you should sign your commits so the DCO won't block the PR.

LukeTillman:
> And unfortunately you have a conflict in this PR @LukeTillman :(, and you should sign your commits so the DCO won't block the PR.

Yeah, I will fix the DCO and merge conflict today. I'm not surprised I forgot to sign a few over the course of the last few (six!) weeks. 😂

LukeTillman:
Ahh, crap, screwed up my rebase. Never mind the above, working on fixing...

LukeTillman:
Sorry about all the noise, merge conflict and DCO are fixed. 🎉

Nexucis:
nice :) thanks for this huge rework @LukeTillman, I will try to review as much as I can soon, expect from me a bunch of silly questions :P.

ui/app/src/model/bundled-plugins.ts
```diff
@@ -0,0 +1,53 @@
...
+  importPluginModule: () => import('@perses-dev/prometheus-plugin'),
+});
+BUNDLED_PLUGINS.set(panelsPackage.perses as PluginResource, {
+  importPluginModule: () => import('@perses-dev/panels-plugin'),
+});
+
+/**
+ * Returns props for the PluginRegistry allowing it to load plugins that are bundled with the app.
+ */
+export function useBundledPlugins(): Pick<PluginRegistryProps, 'getInstalledPlugins' | 'importPluginModule'> {
```
sjcobb, line 38: This is really cool, looks like it will be easy to add more types of plugins in the future

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * The View for viewing a Dashboard.
+ */
+function ViewDashboard() {
+  const dashboard = useSampleData<DashboardResource>(
+    new URLSearchParams(window.location.search).get('dashboard') || 'node-exporter-full'
+  );
+
+  // TODO: Loading indicator
```
sjcobb, line 27: I'm interested to discuss the plan for loading states, since I've been unsure what's the best approach in some previous PRs

ui/components/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@perses-dev/components",
```
sjcobb, line 2: I imagine I'm going to be spending **_a lot_** of time in this package :smile:
I can take care of moving some of the chart components over if you'd like?

ui/components/src/utils/combine-sx.ts
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+
+import { SxProps, Theme } from '@mui/material';
+
+/**
+ * Combines sx props from multiple sources. Useful when creating a component and
+ * you want consumers to be able to provide SxProps that should be combined with
+ * some built-in styles.
+ */
+export function combineSx(...sxProps: Array<SxProps<Theme> | undefined>): SxProps<Theme> {
```
sjcobb, line 21: I've seen this in cloud, but I need to start using it

ui/components/package.json
LukeTillman: Yeah, that would be great. Once we get this PR merged, you can definitely do that. 👍

ui/app/src/views/ViewDashboard.tsx
LukeTillman: Yeah, feels like we may need some kind of common `LoadingPane` component or something similar for full-page loading like this. I really like the `Skeleton` usage so far for other stuff. But might be something for Jared to take a crack at.

ui/app/src/model/bundled-plugins.ts
LukeTillman: Yeah, my thinking here was that someone doing an embed of Perses could do their own version of this hook and bundle whatever plugins they want in their own build/environment.

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
```diff
@@ -35,11 +37,12 @@ export function GaugeChartPanel(props: GaugeChartPanelProps) {
     definition: {
       options: { query, calculation },
     },
+    contentDimensions,
   } = props;
   const unit = props.definition.options.unit ?? { kind: 'Percent', decimal_places: 1 };
   const thresholds = props.definition.options.thresholds ?? defaultThresholdInput;
-  const { contentDimensions } = usePanelState();
-  const { data, loading, error } = useGraphQuery(query);
+  const suggestedStepMs = useSuggestedStepMs(contentDimensions?.width);
```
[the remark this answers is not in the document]
LukeTillman: Wasn't sure if this was the best way to go, but seemed to work for now.

ui/core/src/model/time.ts
```diff
@@ -83,3 +83,13 @@ export function isDurationString(maybeDuration: string): maybeDuration is Durati
   if (maybeDuration === '') return false;
   return DURATION_REGEX.test(maybeDuration);
 }
+
+/**
+ * Gets a suggested step/interval size for a time range based on the width of a visual component.
+ */
+export function getSuggestedStepMs(timeRange: AbsoluteTimeRange, width: number) {
+  // TODO: Should we try to suggest more "rounded" step values based around
```
sjcobb, line 91: Yeah I like this idea, I can look into this if we decide to

sjcobb (approved):
I did it!  all-files-viewed-scrnsht

Going ahead and approving since this looks great and I'm not sure if I'll be able to join the meeting Mon or not.

ui/dashboards/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "build": "tsc",
+    "test": "echo 'no test to run' && exit 0",
+    "lint": "eslint src --ext .ts,.tsx",
+    "lint:fix": "eslint --fix src --ext .ts,.tsx"
+  },
+  "dependencies": {
+    "@mui/material": "^5.4.0",
+    "@perses-dev/components": "^0.1.0",
+    "@perses-dev/core": "^0.1.0",
+    "@perses-dev/plugin-system": "^0.1.0",
```
Nexucis, line 18: ah interesting to see that this package doesn't depend on `prometheus-plugin`

ui/plugin-system/src/components/PluginLoadingBoundary/index.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2021 The Perses Authors
```
Nexucis, line 1: mmm that's weird the syntax highlight of github doesn't work anymore here :/

ui/prometheus-plugin/package.json
```diff
@@ -12,13 +12,16 @@
...
+    "@perses-dev/core": "^0.1.0",
+    "@perses-dev/plugin-system": "^0.1.0",
     "date-fns": "^2.28.0",
     "lezer": "^0.13.5",
     "lezer-promql": "^0.20.0",
-    "react": "^17.0.2",
     "react-query": "^3.34.12"
   },
+  "peerDependencies": {
+    "react": "^17.0.2"
```
Nexucis, line 23: I guess you should add it as a `devDependency` as well

ui/panels-plugin/package.json
```diff
@@ -14,7 +14,8 @@
   "dependencies": {
     "@formatjs/intl-numberformat": "^7.4.1",
     "@mui/material": "^5.4.0",
-    "@perses-ui/core": "^1.0.0",
+    "@perses-dev/core": "^0.1.0",
+    "@perses-dev/plugin-system": "^0.1.0",
     "date-fns": "^2.28.0",
     "echarts": "^5.3.0",
     "lodash-es": "^4.17.21",
```
Nexucis, line 21: don't you want to consider the `react` dependency as a `peerDependency` too ? Like it is the case in others packages ?

Nexucis (commented):
It looks interesting, I just have few questions or notes :)

ui/app/src/App.tsx
```diff
@@ -11,48 +11,38 @@
...
       >
-        <PluginRegistry
-          loadingFallback="Loading..."
-          ErrorFallbackComponent={AlertErrorFallback}
-          runtime={pluginRuntime}
-        >
-          <DataSourceRegistry>
-            <ViewDashboard resource={dashboard} />
-          </DataSourceRegistry>
+        <PluginRegistry getInstalledPlugins={getInstalledPlugins} importPluginModule={importPluginModule}>
```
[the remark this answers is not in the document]
LukeTillman: Yeah, I'm not sure about Datasources yet to be honest, but the `PluginRegistry` here is because I was thinking of "what plugins are available and loaded" as a piece of global state that's shared across the application. Basically we set it up here once and then all the package underneath it can use plugins. Datasources are here for the same reason (for now) but maybe we'll find out that they're feature-specfic. I feel like I need to think a bit more on what a "Datasource" actually is/should be and start a discussion.

ui/app/src/context/DataSourceRegistry.tsx
```diff
@@ -51,19 +46,13 @@ export function DataSourceRegistry(props: DataSourceRegistryProps) {
...
-}
+  const context = useMemo(() => {
+    const defaultDatasource = datasourceList.find((ds) => ds.spec.default);
+    if (defaultDatasource === undefined) return undefined;
+    return { defaultDatasource, getDatasources };
+  }, [getDatasources, datasourceList]);
 
-export function useDataSources(selector: DatasourceSelector) {
-  return useDataSourceRegistry().getDataSources(selector);
+  // TODO: Show a loading indicator? Change up the API to be async?
```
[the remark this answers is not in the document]
LukeTillman: Yup, agreed on needing a conversation on loading state. I'm working on a PR right now to make this API better so we don't have to block loading of the page and can render (for example) a Dashboard while these are still loading.

ui/core/src/model/dashboard.ts
```diff
@@ -32,7 +32,7 @@ export interface DashboardResource {
 export interface DashboardSpec {
   datasource: DatasourceSelector;
   duration: DurationString;
-  variables: Record<string, AnyVariableDefinition>;
-  panels: Record<string, AnyPanelDefinition>;
+  variables: Record<string, VariableDefinition>;
   layouts: LayoutDefinition[];
+  panels: Record<string, PanelDefinition>;
```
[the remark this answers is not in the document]
LukeTillman: Yeah, I remember that in the original spec. I remember discussing it when we did the change to support fancier CSS grid layouts and I think that's when I changed this so that `layouts` is now basically `entrypoint`.

ui/dashboards/package.json
LukeTillman: Yeah, it can just depend on the `plugin-system` package and not know anything about what plugin is being used to do the querying for data. So in the (far) future, we could render data from things like Graphite, etc.

ui/panels-plugin/package.json
LukeTillman: Yup, good call out. Will fix this.

ui/plugin-system/src/components/PluginLoadingBoundary/index.ts
LukeTillman: Haha, same for me. PR is too big. 😂

ui/prometheus-plugin/package.json
LukeTillman: Yup, good catch.

ui/components/package.json
Nexucis: yup I opened a discussion here : https://github.com/perses/perses/discussions/331. I tried to find a clean solution and came up with different option.
It will be cool to have a consensus around this topic before starting any work around the CI/CD (which shouldn't be a big effort, since it just misses the release job(s))

ui/dashboards/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@perses-dev/dashboards",
```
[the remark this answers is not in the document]
LukeTillman: Yeah, documentation is on my list as well. We discussed maybe doing Storybook, but not sure if it's appropriate. I will add some basic `README`s though as a start.

Nexucis:
@LukeTillman once you fix the dependencies "issues" I found, I think we are good to merge this PR.

Readme on each package can certainly come in different PR.

PR #335, opened 2022-03-15
add the release documentation

Nexucis:
As part of #331, I'm creating this documentation that is indicating how to release Perses. I took some inspirations from the Prometheus release process.

Probably we can still improve it and automatise some steps.

~~I would be interested if there is some magic command to upgrade the npm dependencies when you are using a monorepo. Would be great to know for Prometheus too.~~

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -26,12 +26,19 @@ In case you would like to update manually the dependencies :
 ```bash
 make update-go-deps
 git add go.mod go.sum
-git commit -m "Update dependencies"
+git commit -m "Update Go dependencies"
 ```
 
 ### Updating npm dependencies
 
-TBD
+```bash
+make update-npm-deps
+git add ./ui/package-lock.json ./**/package.json
+git commit -m "Update npm dependencies"
+```
+
+Note: in case you feel it, you can also upgrade the npm dependencies to their latest release (major or minor)
+with `make upgrade-npm-deps`
 
 ### 1. Prepare your release
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -44,7 +44,7 @@ with `make upgrade-npm-deps`
 
 You should start to create a branch that follows the pattern `release/v<X.Y.Z>`
 
-Update the file `Changelog.md` and the different `package.json` with the corresponding version:
+Update the file `CHANGELOG.md` and the different `package.json` with the corresponding version:
 
 ```bash
 version=v0.2.0
@@ -57,7 +57,7 @@ git add ./ui/package-lock.json ./**/package.json
 Do this in a proper PR pointing to the release branch as this gives others the opportunity to chime in on the release in
 general and on the addition to the changelog in particular.
 
-Entries in the `Changelog.md` are meant to be in this order:
+Entries in the `CHANGELOG.md` are meant to be in this order:
 
 * `[FEATURE]`
 * `[ENHANCEMENT]`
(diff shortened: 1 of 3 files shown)


RELEASE.md
```diff
@@ -0,0 +1,73 @@
...
+git push origin "${tag}"
+```
+
+Signing a tag with a GPG key is appreciated, but in case you can't add a GPG key to your Github account using the
+following [procedure](https://docs.github.com/en/authentication/managing-commit-signature-verification), you can replace
+the `-s` flag by `-a` flag of the git tag command to only annotate the tag without signing.
+
+Once a tag is created, the release process through the Github Actions will be triggered for this tag.
+
+You should then create the Github release associated to the tag and put the content of the `Changelog.md` there.
```
Nexucis, line 73: this last line will certainly be changed. Goreleaser is able to create the github release and push a Changelog. Only thing remaining is to find a way to put the correct Changelog instead of the generated one.

RELEASE.md
```diff
@@ -0,0 +1,88 @@
...
+general and on the addition to the changelog in particular.
+
+Entries in the `Changelog.md` are meant to be in this order:
+
+* `[FEATURE]`
+* `[ENHANCEMENT]`
+* `[BUGFIX]`
+* `[BREAKINGCHANGE]`
+
+As we have many libraries we aime to expose, it's better if you also put a clear indication about what lib is concerning
```
LukeTillman, line 67:
```suggestion
As we have many libraries we publish, it's better if you also put a clear indication about what library is affected
```

npm-deps.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+function ncu() {
+    target=$1
+    npx npm-check-updates -u --target "${target}"
```
LukeTillman, line 5: We could prob add this to our `devDependencies` in the root `package.json` at some point since I imagine it might be a common thing we use (and thus avoid needing to use `npx`).

LukeTillman (approved):
Docs looks good, going to take a look at the other release PR next...

npm-deps.sh
Nexucis: ah yeah indeed that could be another way. Mmm no idea what's the best. I supposed npx is also quite common in a web environment right ?

AntoineThebaud (approved):

RELEASE.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+Note: in case you feel it, you can also upgrade the npm dependencies to their latest release (major or minor)
+with `make upgrade-npm-deps`
+
+### 1. Prepare your release
+
+You should start to create a branch that follows the pattern `release/v<X.Y.Z>`
+
+Update the file `Changelog.md` and the different `package.json` with the corresponding version:
```
sjcobb, line 47: Is it capitalized CHANGELOG.md like in Prometheus?

RELEASE.md
```diff
@@ -0,0 +1,88 @@
...
+
+```bash
+version=v0.2.0
+./ui_release.sh --release "${version}"
+cd ui/
+npm install
+git add ./ui/package-lock.json ./**/package.json
```
sjcobb, line 54: Could there be an npm script or command in Makefile for this eventually?

sjcobb (approved):
Couple small non-blocking comments--thanks for putting this together!

RELEASE.md
Nexucis: ah yeah you are totally right !
Thanks for spotting it !

RELEASE.md
Nexucis: I managed to add it to the Makefile in the PR https://github.com/perses/perses/pull/336.
I will change this procedure there with the Makefile command.

Spoiler alert it will be like that: 

```bash
make release version="v0.2.0"
```

Nexucis:
thanks for the review @sjcobb @LukeTillman @AntoineThebaud !

PR #355, opened 2022-03-21
move StatChart and GaugeChart to perses-dev/components

sjcobb:
Initial follow up to _Reorganize UI into new @perses-dev packages_ [PR 315](https://github.com/perses/perses/pull/315) to begin moving certain components to perses-dev/components. This PR starts with GaugeChart, StatChart, and EChartsWrapper.

While refactoring and updating imports, I moved some conversion logic from chart-level to panel-level. For example, thresholds are now converted in GaugeChartPanel and StatChartPanel. This provides more flexibility in the props for GaugeChart and StatChart.

Commits
- copy EChartsWrapper, GaugeChart, StatChart to perses-dev/components, update imports
- refactor GaugeChart so thresholds converted in GaugeChartPanel
- change StatChart props so backgroundColor is passed from StatChartPanel
- fix components exports after moving model/units
- delete ui/panels-plugin/src/model/units.ts since moved to components


Nexucis:
mmm the build failed for the frontend.
Not sure if it's a caching issue (like it happened before) or if it's a real issue related to this PR.
I already re-started the job twice and it still failed, so perhaps there is an issue with this PR. Usually it doesn't fail so often.

I'm wondering if we should keep this cache for the UI, it doesn't seem it works well with npm workspace specially when the workspaces are changing heavily.

ui/components/src/StatChart.tsx
```diff
@@ -56,6 +53,15 @@ const noDataOption = {
...
 };
 
+export type UnixTimeMs = number;
+
+export type GraphSeriesValueTuple = [timestamp: UnixTimeMs, value: number];
+
+export interface GraphSeries {
+  name: string;
+  values: Iterable<GraphSeriesValueTuple>;
+}
```
sjcobb, line 63: This is temporary to fix circular dependency errors btw, planning to move these out of StatChart

Nexucis (approved):
nice improvement !

LukeTillman (approved):
Seems pretty straightforward. Thanks for the much smaller PR, was easier to see what's going on. 👍

PR #361, opened 2022-03-24
fix dependency job in release flow

Nexucis:
This should fix the release flow issue spotted here: https://github.com/perses/perses/actions/runs/2036668112


LukeTillman (approved):
🎉

PR #363, opened 2022-03-25
Try setting env var that setup-node writes to .npmrc

LukeTillman:
The `setup-node` action writes an `.npmrc` file with `NODE_AUTH_TOKEN` as a placeholder for use when doing things like publishing packages. See https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm for an example. Let's see if this fixes the auth issue in the libs release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -49,16 +49,19 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: "ui/.nvmrc"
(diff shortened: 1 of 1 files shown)


sjcobb (approved):

Nexucis (approved):

Nexucis:
note that it won't solve the issue on golang side

PR #371, opened 2022-03-28
Fix plugin registry to render children while waiting on plugin list load

LukeTillman:
This is a small improvement to plugin loading that allows the `PluginRegistry` component to render its children while it's waiting on the list of available plugins to load. This way we don't have to block the rendering of a screen (e.g. a Dashboard) while that list is still being fetched.


Nexucis (approved):
oh yeah good idea!

PR #373, opened 2022-03-29
improve the way to create tag and bump the version

Nexucis:
This PR is a proposition to improve a bit more our release flow by adding more command to reduce the potential mistake when creating a tag / bump a new version.

Let me know what do you think.


scripts/ui_release.sh
```diff
@@ -74,8 +74,8 @@ function clean() {
   done
 }
 
-function release() {
-  version=${1}
+function bumpVersion() {
```
LukeTillman, line 77: Nit: Should we think about renaming this script to just `release.sh` or something now that it's doing more than just UI-related release tasks?

LukeTillman (approved):
Looks good to me. Like having fewer commands to screw up. 😂

scripts/ui_release.sh
Nexucis: actually instead of renaming I splitted in two different scripts. Just to isolate what is for the ui libs itself and for the release in general.
It's here: https://github.com/perses/perses/pull/374

PR #375, opened 2022-03-29
fallback to go 1.17

Nexucis:
<del>A tentative to reduce the time taken to build the go binaries</del>


Nexucis:
After analysing what changed comparing when the buildtime was acceptable, and I came to the conclusion it was the Go version.

So my last commit is fallbacking the Go version, and instantly the buildtime moved from 10min to 2min and didn't fail.

I opened an issue on `goreleaser` side (https://github.com/goreleaser/goreleaser/issues/3011). Let's see if there is something to tweak somewhere.

PR #387, opened 2022-04-02
Datasources API Update

LukeTillman:
Like we discussed in Matrix chat, this updates the Datasources API to use JSON refs rather than our bespoke "metadata query" API which could have returned more than one Datasource for a query. I pulled in a new library called `jsonref` that takes care of resolving the reference to the Datasource (including fetching from the Perses API if that's what's referenced). One other nice side-effect of this is that the `DatasourcesRegistry` no longer needs to block rendering while loading the entire list of Datasources. Instead, they're loaded on demand as the panels request them.

Some of the sample data was a little out of date, so I updated that to fix a few compile errors and use the new ref API. Let me know if you have questions!


sjcobb (approved):
LGTM!

ui/app/sample-data/benchmarkDashboard.ts
```diff
@@ -22,7 +22,7 @@ const benchmarkDashboard: DashboardResource = {
     updated_at: '2021-11-09',
   },
   spec: {
-    datasource: { kind: 'Prometheus', name: 'PrometheusDemo', global: true },
+    datasource: { $ref: '/api/v1/globaldatasources/PrometheusDemoBrowser#' },
```
Nexucis, line 25: is it the regular syntax to have a `#` at the end of the URI ?

ui/app/src/context/DatasourceRegistry.tsx
```diff
@@ -0,0 +1,77 @@
...
+const DEFAULT_DATASOURCE_CACHE_KEY = '__DEFAULT_DATASOURCE__';
+
+export interface DatasourceRegistryProps {
+  children: React.ReactNode;
+}
+
+/**
+ * Makes DataSource resources available to children.
+ */
+export function DatasourceRegistry(props: DatasourceRegistryProps) {
```
Nexucis, line 29: Since `jsonRef` can load on demand the datasources, I'm wondering if we still need a datasource registry.

ui/app/sample-data/benchmarkDashboard.ts
Nexucis: Something just came to my mind:  how do we use the default datasource here ?

Currently you can have one default datasource per type of datasource. So I'm wondering how you can say I want to use the Prometheus default datasource.

ui/prometheus-plugin/src/model/prometheus-client.ts
```diff
@@ -65,21 +65,23 @@ function useQueryWithGet<T extends RequestParams<T>, TResponse>(
   params: T,
   queryOptions?: QueryOptions
 ) {
-  const config = usePrometheusConfig(queryOptions?.datasource);
-  const datasourceURL = buildDatasourceURL(config.metadata.name, config.spec.http);
-  const key = [datasourceURL, apiURI, params] as const;
+  const getPrometheusDatasource = useGetPrometheusConfig();
```
Nexucis, line 68: a bit overkill to create an alias of the function don't you think ?

Nexucis (commented):
overall looks fine I just have a couple of questions / notes

ui/app/sample-data/benchmarkDashboard.ts
LukeTillman: > is it the regular syntax to have a `#` at the end of the URI ?

Yeah, in json ref terms it just means "the root node of the document" because in theory this could actually be `/my/url#/path/in/document/to/datasource`. But maybe it's not necessary for the root in this case and it will default to it without the `#`? I can try it without and see if it works. 🤷‍♂️ 

> Something just came to my mind: how do we use the default datasource here ?
> 
> Currently you can have one default datasource per type of datasource. So I'm wondering how you can say I want to use the Prometheus default datasource.

Interesting. I was under the impression that there were two "defaults" in play:

1. The default for the dashboard, specified in the dashboard spec as `datasource`. Right now that field is required although I was thinking maybe it should be optional in which case it would fall back to...
2. The system-wide default datasource which the `DataSourceRegistry` is calculating by looking at the list of datasources and finding the one with the default flag (regardless of kind).

You think there should be a default per kind?

ui/app/src/context/DatasourceRegistry.tsx
LukeTillman: Right now, we still need something to do caching of queries and default data source calculations (although see my question above about default datasources). Without the caching in the registry, all the visible panels would be rendering in parallel and ask for their datasource and then you would get 10 or 12 or however many panels all making HTTP requests to resolve that ref (which would be bad 😂 ).

ui/prometheus-plugin/src/model/prometheus-client.ts
LukeTillman: Good question, not sure if you're asking:

1. Shouldn't this line just be `const getPrometheusConfig = useGetPrometheusConfig()` or
2. Why not just call `useGetPrometheusConfig()` to get the config instead of a callback to get it?

The answer to number 2 is a little more complicated and has to do with the rules of hooks in React, so happy to explain if it's the latter. If it's number 1, yeah, probably should just be the same name. 😂

Nexucis: ah I was more on the number 2.
But if it's actually a necessary line of code and not just there to rename the function, then maybe just give the same name no ?

ui/app/src/context/DatasourceRegistry.tsx
Nexucis: ah yeah I forgot about the caching.

I thought it would be simply possible to do the caching with the libs react-fetch and so we can remove this.

ui/app/sample-data/benchmarkDashboard.ts
Nexucis: > Yeah, in json ref terms it just means "the root node of the document" because in theory this could actually be /my/url#/path/in/document/to/datasource

Well if it's coming from the `jsonref` specification, we should follow it even it doesn't really make sense for us on this particular point.

> You think there should be a default per kind?

That was something I found cool. But it can be changed if it's too hard to managed on frontend side or totally overkilled.

In my mind saying, I want to use the default Prometheus datasource, and then default Log Server datasource for my dashboard sounds cool to me.

ui/app/src/context/DatasourceRegistry.tsx
LukeTillman: Yeah, let me think on it some more. Maybe there's a way to do this with just a hook that doesn't need a registry... 🤔

ui/app/sample-data/benchmarkDashboard.ts
LukeTillman: So, if we wanted to do a "default per kind", I think we're back to the question we discussed in Matrix chat which is "how do you point at a Datasource"? I don't think using a json reference works in that case.

So how would I say, "I want to use a Prometheus Datasource" or "I want to use a Log Server Datasource". The previous thing we had with a "Datasource selector" was OK, but the syntax felt a little weird and the API when you query returned multiple Datasources, so it felt like there wasn't really a way to point at a specific one.

LukeTillman: BTW, I'm not at all tied to using json refs, that was just my suggestion for solving the problem of needing to point at a specific datasource. If some sort of "selector" API is better or more desirable, we should definitely do that and figure out how to make it possible to make sure you only get a single Datasource back when you use a selector.

Nexucis: >  The previous thing we had with a "Datasource selector" was OK, but the syntax felt a little weird and the API when you query using the selector could potentially return multiple Datasources that matched the selector, so it felt like there wasn't really a way to point at a specific one

mmm the previous selector is this one:

```go
type Datasource struct {
	// Name is the name of the datasource
	Name string `json:"name" yaml:"name"`
	// Kind is the datasource kind
	Kind datasource.Kind `json:"kind" yaml:"kind"`
	// If global is true, we are referencing a global datasource.
	// When set to false, we are referencing a datasource in the same project as the current dashboard.
	Global bool `json:"global" yaml:"global"`
}
```

`Kind` is only necessary in case you would like to use the default datasource that is matching this type.

In case you want to specify the datasource you would like to use, you can give the name as it is the unique key.

Then if the value `Global` is set to false, then the datasource should be retrieved in the same project where the dashboard belongs.

But before going back to a selector or moving forward with the json ref, we should assess if:
* We want a default datasource
* We want a default datasource per kind.

Both are ok from my perspective. Grafana is doing the first one only if I remember correctly. 

I'm really sorry because I totally forgot about the `default` feature. But now it is the good time to assess if we want to keep it and to remove it :).

Last thing: if the selector wasn't clear at all how to use it, definitely we should change it to make it clearer.

LukeTillman:
Closing this while we discuss in #404

PR #391, opened 2022-04-04
Remove uplot

LukeTillman:
Since we've gone all in on E-Charts at this point, this just removes the `uplot` dependency and related code. I feel a little bit like I'm stealing Steven's thunder here by being the one to do this. 😆


Nexucis (approved):
🍿

sjcobb (approved):
You beat me to it! :joy:

PR #392, opened 2022-04-04
Minor improvements for NPM package consumers

LukeTillman:
This makes a couple very minor improvements for consumers of the NPM packages:

1. Removes our reliance on Map/Set in the `immer` library. This means someone using `@perses-dev/plugin-system` doesn't have to know to call `enableMapSet()` somewhere (nor do they need to add `immer` as a dependency to their app to do that).
2. Moves `@mui/material` and `react-query` from `dependencies` to `peerDependencies` so that folks can bring their own compatible version if they already have one. Since both of these rely on React context (the theme context in MUI's case and the query client context in react-query's case), it makes sense to have these be peers so that the context works properly in apps that already have these libraries.

A "TIL" side note on number 2: In `npm` versions > 7, it now automatically installs `peerDependencies` if they're not already installed/specified somewhere higher up in the dependency tree, so no need to duplicate things in `devDependencies`. 🎉

### DIFF
--- a/ui/app/src/render-app.tsx
+++ b/ui/app/src/render-app.tsx
@@ -14,7 +14,6 @@
 import { CssBaseline, ThemeProvider } from '@mui/material';
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { enableMapSet } from 'immer';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import App from './App';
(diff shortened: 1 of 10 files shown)


Nexucis (approved):
SGTM

sjcobb (approved):

PR #409, opened 2022-04-12
fix eslint issue after adding new LineChart prop

sjcobb:
Fix overlooked dataZoomEnabled eslint issue from my previous sjcobb/components/reorg-line [PR 403](https://github.com/perses/perses/pull/403)


LukeTillman (approved):
💥

PR #411, opened 2022-04-14
remove @formatjs/intl-numberformat dependency

sjcobb:
In release v0.3.0, units.ts moved from perses-dev/panels-plugin to perses-dev/components, but I forgot to update each package.json with the @formatjs/intl-numberformat dependency that moved (formatjs was added in prev StatChart [PR 307](https://github.com/perses/perses/pull/307)).

To simplify things, I've decided to remove the @formatjs/intl-numberformat dependency since it was only being used for one function.

Commits:
- remove @formatjs/intl-numberformat dependency
- copy IsSanctionedSingleUnitIdentifier util that checks against SANCTIONED_UNITS from: https://tc39.es/ecma402/#table-sanctioned-simple-unit-identifiers

### DIFF
--- a/ui/components/src/model/units.ts
+++ b/ui/components/src/model/units.ts
@@ -11,7 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { IsSanctionedSimpleUnitIdentifier } from '@formatjs/ecma402-abstract';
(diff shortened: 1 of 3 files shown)


Nexucis (approved):
👍

LukeTillman (approved):

PR #415, opened 2022-04-18
add dark mode support, theme toggle in header

sjcobb:
This PR adds support for switching between light and dark mode using MUI theme defaults. There are boilerplate override examples for future customization. Let me know if there are default colors you'd like to change (in meantime before we have mockups to reference).

Theme preference is currently stored in localStorage.

_A future PR will generate ECharts themes from MUI theme colors / fonts._

Current state:

dark_mode_ex

https://user-images.githubusercontent.com/9369625/163816287-40229504-f109-42ee-ac48-36b52ce0535e.mov


Nexucis:
the dark theme itself is so sexy 🥳 !!!
I just find it strange that we reload the entire UI when we switch the theme. Is it a technical limitation ? I think it's smooth if the switch of the theme just change the theme and doesn't reload the entire UI which induce asking again the data to fill the dashboards

sjcobb:
> the dark theme itself is so sexy 🥳 !!! I just find it strange that we reload the entire UI when we switch the theme. Is it a technical limitation ? I think it's smooth if the switch of the theme just change the theme and doesn't reload the entire UI which induce asking again the data to fill the dashboards

Yeah you're right @Nexucis, I referenced part of this implementation from Chronosphere, but took out the part where we use GraphQL to save the user's theme preference to the db. The reload was needed in that case, but not here so I'll take a look at removing it!

ui/app/src/utils/browser-storage.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright 2021 The Perses Authors
```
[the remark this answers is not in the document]
sjcobb: :pray: thanks I'll definitely start using that!

LukeTillman:
Yeah, that reload in our UI is because of some weird `iframe` considerations, so hopefully not necessary here.

ui/app/src/context/DarkMode.tsx
```diff
@@ -0,0 +1,65 @@
...
+  isDarkModeEnabled: boolean;
+  setDarkMode: (pref: boolean) => Promise<void>;
+}
+
+export const DarkModeContext = createContext<DarkModeContext>({
+  isDarkModeEnabled: false,
+  setDarkMode: () => {
+    throw new Error('Error: Dark mode not implemented');
+  },
+});
```
LukeTillman, line 31: Minor suggestion: I hate that React makes you define a "default" context implementation like this where, for example, the `setDarkMode` doesn't really do anything. In my opinion, for contexts like this where a default doesn't really work, we should be throwing if the Provider is missing not giving them a default value.

What I've been doing instead is defining the content as `| undefined` and then exporting a hook for using it that does the check for `undefined` and throws a helpful error message. Would be cool to follow that pattern here too. For example:

https://github.com/perses/perses/blob/1f420c168d062bd9bffdae84da15ee1145247811/ui/plugin-system/src/runtime/time-range.ts#L21-L32

ui/components/src/StatChart.tsx
```diff
@@ -189,7 +191,7 @@ export function StatChart(props: StatChartProps) {
       },
       series: statSeries,
       textStyle: {
-        color: backgroundColor === 'transparent' ? '#000000' : '#FFFFFF',
+        color: backgroundColor === 'transparent' ? theme.palette.text.primary : '#FFFFFF',
```
LukeTillman, line 194: Looks like the linter caught a missing `theme` dependency here for the `useMemo`.

LukeTillman (approved):
Overall looks great! One minor suggestion inline.

sjcobb:
Getting rid of location.reload isn't as easy as I thought it'd be--I'm going to tackle that issue in a separate PR as part of the ECharts theme generation task I've been working on in a different branch.

PR #421, opened 2022-04-22
Switch bundle output directory back

LukeTillman:
I'm an idiot and left the wrong bundle output directory in webpack from when I added all the `@perses-dev` packages in #315. 😂   Sorry for breaking this!


Nexucis (approved):
no worries, thanks for the fix :)

PR #425, opened 2022-04-25
bugfix echarts wrapper disposes canvas too frequently 

sjcobb:
Currently, our echarts wrapper component disposes and re-initializes the canvas when events are fired and in certain cases when data / props do not change. This PR, refactors the dispose logic and adds refs for tracking when to call init and setOption. useDeepMemo is also added to LineChart.tsx when the echarts option is set.

Tooltip styles were also adjusted since I noticed they looked off in dark mode and dependabot PRs [417](https://github.com/perses/perses/pull/417) and [418](https://github.com/perses/perses/pull/418) were reverted since those versions of @testing-library/react are only compatible with React v18.


sjcobb:
Will fix the failing build tomorrow, I probably created a circular dependency by accident

ui/components/src/EChart.tsx
```diff
@@ -11,49 +11,30 @@
...
 // limitations under the License.
 
-import React, { useState, useLayoutEffect } from 'react';
+import React, { useCallback, useEffect, useLayoutEffect, useRef } from 'react';
 import debounce from 'lodash/debounce';
 import { ECharts, EChartsCoreOption, init } from 'echarts/core';
 import { Box, SxProps, Theme } from '@mui/material';
+import { isEqual } from 'lodash-es';
 
+// https://echarts.apache.org/en/api.html#events
```
sjcobb, line 20: Removed these comments since they are duplicated verbatim from [this link](https://echarts.apache.org/en/api.html#events)

ui/components/src/EChart.tsx
```diff
@@ -121,62 +102,80 @@ export type OnEventsType<T> = {
...
-  const [echart, setChart] = useState<ECharts | undefined>(undefined);
-  // Create a chart instance in the container
-  useLayoutEffect(() => {
-    if (containerRef === null) return;
-
-    const chart = init(containerRef);
-    setChart(chart);
-    if (_instance !== undefined) {
-      _instance.current = chart;
+export const EChart = React.memo(function EChart<T>({
```
sjcobb, line 110: Renamed to avoid confusion / needing an alias to fix collision with _ECharts from 'echarts/core'_ 
Also since this component is for rendering a single chart

ui/components/src/EChart.tsx
```diff
@@ -121,62 +102,80 @@ export type OnEventsType<T> = {
...
-    onChartInitialized?.(chart);
-
-    return () => {
-      chart.dispose();
-    };
-  }, [containerRef, _instance, onChartInitialized, onEvents]);
+  useEffect(() => {
+    const chart = chartElement.current;
+    if (chart === null || onEvents === undefined) return;
+    bindEvents(chart, onEvents);
```
LukeTillman, line 134: Shouldn't this be cleaning up event handlers also in a dispose since it's the effect that's binding them? Not sure if that's a thing since I'm not super familiar with the echarts API.

ui/components/src/EChart.tsx
```diff
@@ -121,62 +102,80 @@ export type OnEventsType<T> = {
...
+  const initChartDom = useCallback(() => {
+    if (containerRef.current !== null && chartElement.current === null) {
+      chartElement.current = init(containerRef.current);
+      onChartInitialized?.(chartElement.current);
+      if (_instance !== undefined) {
+        _instance.current = chartElement.current;
+      }
     }
-    bindEvents(chart, onEvents);
+  }, [_instance, onChartInitialized]);
```
LukeTillman, line 129: Since this function is only used in one place, wonder if it should just be inlined into the spot where it's used?

ui/components/src/EChart.tsx
```diff
@@ -121,62 +102,80 @@ export type OnEventsType<T> = {
...
     updateSize();
-    return () => window.removeEventListener('resize', updateSize);
-  }, [echart]);
+    return () => {
+      window.removeEventListener('resize', updateSize);
+      if (chartElement.current !== null) {
+        chartElement.current.dispose();
+        chartElement.current = null;
+      }
+    };
```
LukeTillman, line 162: This feels weird to me because it's dispose function is cleaning up something from (I think?) another effect? Feels like each effect should be cleaning up after itself otherwise this is going to become really hard to reason about.

ui/components/src/utils/memo.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright 2022 The Perses Authors
```
LukeTillman, line 1: Was there a circular dependency issue with `components` just using this from `core`? I think a dependency from components -> core is totally fine if we need to. Feels like `core` is always gonna be our "lowest level" package that others might depend on.

LukeTillman (approved):
Looks like this was a "fun" issue to try and solve. 😂  Left some suggestions inline. In particular, I'm a little worried about the chain of effects in `EChart` becoming hard to reason about (feels like those always are), so anything we can do to make that better would be a great thing. Thanks for digging into this!

ui/components/src/EChart.tsx
sjcobb: We're definitely in undocumented territory, but I have [an approach now](https://github.com/perses/perses/blob/fcc5354c64e22a7cb90fcc743dc54a0c396c6888/ui/components/src/EChart.tsx#L160) that will work for unbinding events individually!

ui/components/src/EChart.tsx
sjcobb: Yeah good call, and now that event unbinding works, I was able to clean this up :tada:

ui/components/src/utils/memo.ts
sjcobb: It doesn't seem to be a circular dependency, but I'm seeing [these errors](https://github.com/perses/perses/runs/6181994115?check_suite_focus=true) in CI. I wasn't seeing them locally earlier, but I'll try again to reproduce it.

sjcobb: Fixed in [this commit](https://github.com/perses/perses/pull/425/commits/ab2d991821f70e8c2d5be04b1b3ef1d325549520) so now components imports useDeepMemo from core

PR #427, opened 2022-05-01
fix go build: set parallelism build to 1

Nexucis:
This PR should fix back the build of the binaries for the release


sjcobb (approved):
looks good!

PR #430, opened 2022-05-03
fix package.json so core is dependency of components

sjcobb:
This PR fixes issue noted in #perses-dev chat that core is currently is dependency of itself. Both package.json files are updated so now the components package instead pulls from core.

Issue was introduced by me in [PR 425](https://github.com/perses/perses/pull/425)

### DIFF
--- a/ui/components/package.json
+++ b/ui/components/package.json
@@ -23,6 +23,7 @@
     "lint:fix": "eslint --fix src --ext .ts,.tsx"
   },
   "dependencies": {
+    "@perses-dev/core": "^0.4.0",
     "echarts": "^5.3.2",
     "lodash-es": "^4.17.21",
     "react-error-boundary": "^3.1.4"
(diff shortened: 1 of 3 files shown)


ui/components/package.json
```diff
@@ -23,6 +23,7 @@
     "lint:fix": "eslint --fix src --ext .ts,.tsx"
   },
   "dependencies": {
+    "@perses-dev/core": "^0.4.0",
```
Nexucis, line 26: are you sure you want to remove the dependency in the `component` package too ?

sjcobb: This is adding it to components package.json right? Intention was to make core a dependency of components since LineChart [imports useDeepMemo here](https://github.com/perses/perses/blob/v0.4.0/ui/components/src/LineChart.tsx#L15).

Nexucis: ah sorry I read it wrongly. I thought it was removed :(

Nexucis (approved):
lgtm. Thanks for the fix @sjcobb !

ui/components/package.json
sjcobb: no worries!

PR #432, opened 2022-05-06
CLI: lint cmd can load CUE schema to validate dashboards locally

Nexucis:


AntoineThebaud (approved):
great!

PR #435, opened 2022-05-06
Update CUE schemas with frontend definition

Nexucis:
This PR aims to update the cue schema with the current definition defined in the frontend.


schemas/plugins/gauge/panel.cue
```diff
@@ -1,11 +1,24 @@
...
+	options: {
+		query:       #query
+		calculation: common.#calculation
+		unit?:       common.#unit
+		thresholds?: common.#threasholds
+	}
+}
+
+#query: {
+
```
Nexucis, line 21: this is where we have to be able to inject new definition. Here for example we should be able to inject the prometheus query definition.

I'm wondering how we will do that

schemas/plugins/common/unit.cue
```diff
@@ -0,0 +1,19 @@
...
+#percentUnit: {
+	kind:           "Percent" | "PercentDecimal"
+	decimal_places: number
+}
+
+#decimalUnit: {
+	#kind:          "Decimal"
+	decimal_places: number
+	suffix?:        "string"
+	unitDisplay?:   "short" | "long" | "narrow"
```
Nexucis, line 18: perhaps we should name it `unit_display` to align the syntax on other fields.

Nexucis (commented):
I finished this PR.
Hopefully I didn't miss anything and after this PR the datamodel should be aligned.

@sjcobb @LukeTillman can you tell me if you are fine with this?

schemas/base.cue
```diff
@@ -1,5 +1,7 @@
 {
-  displayed_name: string
-  kind: string
-  chart: _ // any type
-}
\ No newline at end of file
+	kind: string
+	display: {
+		name: string
```
AntoineThebaud, line 4: Just for you to know, like this the `display` struct is not closed, if you wanted it to be so you'd need to do something like
```
{
	kind: string
	display: #display
	options: _ // any type
}

#display: {
        name: string
}
```

but actually the wraping base def is already opened and should be in the current case:
```
#panel: {
        displayed_name: string
	kind: string
	chart: _ // any type
}

#panel
```


...but doing so was causing the validation of plugins to fail so I removed it "temporarily". Now I have again time to allocate to this, so you can keep it as is for now and I'll find a solution hopefully :D

schemas/plugins/common/calculation.cue
```diff
@@ -0,0 +1,3 @@
+package common
```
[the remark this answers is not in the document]
AntoineThebaud: lgtm!

schemas/base.cue
Nexucis: I'm not sure I'm following what's the issue here ^^. 

In the meantime, since we talking about `display`, for the moment there is only one attribute in this object, @LukeTillman @sjcobb do you foresee we will have more fields later ? Or do you think we could make it flat ? 

```cue
{
    kind: string
    display_name: string
    char: _
}
```

AntoineThebaud: It's not an issue in your PR, this was just to share information ;)

Nexucis:
I just added some "unit" test on the chart schemas in order to verify that the schemas validate a valid JSON file.
And here the errors I got: 

```bash
testing charts gauge
options.thresholds.steps.1.color: incomplete value string:
    ../common/threshold.cue:5:9
options.unit.decimal_places: incomplete value number:
    ../common/unit.cue:11:18
```

I took the JSON example from the file `nodeExporterDashboard.ts` and it seems there are others things missing.
Here for instance, I have this JSON file: 

```JSON
{
  "kind": "GaugeChart",
  "display": {
    "name": "CPU Busy"
  },
  "options": {
    "query": {
      "kind": "PrometheusGraphQuery",
      "options": {
        "query": "(((count(count(node_cpu_seconds_total{job='node',instance='$instance'}) by (cpu))) - avg(sum by (mode)(rate(node_cpu_seconds_total{mode='idle',job='node',instance='$instance'}[$interval])))) * 100) / count(count(node_cpu_seconds_total{job='node',instance='$instance'}) by (cpu))"
      }
    },
    "calculation": "LastNumber",
    "unit": {
      "kind": "Percent"
    },
    "thresholds": {
      "steps": [
        {
          "value": 85
        },
        {
          "value": 95
        }
      ]
    }
  }
}
```

And CUE is yelling that `thresholds` are missing the field `color`.  So my question is, who is wrong ? The JSON file is incorrect and CUE is correct (so we have to provide the field `color`) or is it the opposite ? (so I have to change the CUE schema and potentially the typescript too)

Same question for the `unit` that expects `decimal_places` to be filled.

As a final note, I noticed in the `DecimalUnitOptions` we have this field `suffix?: 'string';`.
So from my understanding, `suffix` is an optional field that can have a unique value `string`. That doesn't look like it's a super useful field don't you think ? Or if it's just, we need to know if the user needs it or not, perhaps a boolean could simply do the job.

@sjcobb or @LukeTillman if you have some times to look at it, that would be cool. Thank you :)

sjcobb:
@Nexucis for thresholds color, it looks like the [type in thresholds.ts](https://github.com/perses/perses/blob/v0.4.0/ui/panels-plugin/src/model/thresholds.ts#L33) is wrong since it should be optional as [shown here](https://github.com/perses/perses/blob/v0.4.0/ui/app/sample-data/nodeExporterDashboard.ts#L92). Sorry about that! I can make a separate PR with that fix too if you want to go ahead and merge this one?

schemas/plugins/common/unit.cue
sjcobb: Or I could have just named it `display` since we're already inside of unit so it can be inferred (for context I took the values from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#unitdisplay))

sjcobb (approved):

schemas/plugins/common/unit.cue
Nexucis: ah that's a good idea. I will update it with this new name

schemas/base.cue
Nexucis: alright let's keep it like that. Thanks!

Nexucis:
cool, thanks @sjcobb for your feedback.

> The thought with suffix was it would support any of the units shown in this [ECMA standard](https://tc39.es/proposal-unified-intl-numberformat/section6/locales-currencies-tz_proposed_out.html#sec-issanctionedsimpleunitidentifier), I played around with it some in [this dashboard for example](https://github.com/perses/perses/blob/v0.4.0/ui/app/sample-data/benchmarkDashboard.ts#L210).
> 
> On second thought, It actually conflicts with how kinds are defined though, so I'd be good with getting rid of it.

Perhaps on this particular point, is it ok if you do it ? I'm not sure to know what are the consequences to remove the field `suffix` from the datamodel. I supposed there will be some changes in the typescript.

Nexucis:
alright, thanks a lot for your review and feedback @sjcobb @AntoineThebaud. Let's merge this one to unlock the situation.

PR #439, opened 2022-05-11
fix line chart xAxis date formatting

sjcobb:
LineChart xAxis labels now show days / seconds when applicable instead of only hours / minutes.

**Screenshot**:

graph-step-days-ex

### DIFF
--- a/ui/components/src/LineChart.tsx
+++ b/ui/components/src/LineChart.tsx
@@ -309,9 +309,9 @@ function getFormattedDate(value: number, rangeMs: number) {
     minute: 'numeric',
     hour12: false,
   };
-  const fiveMinMs = 300000;
+  const thirtyMinMs = 1800000;
   const dayMs = 86400000;
-  if (rangeMs <= fiveMinMs) {
+  if (rangeMs <= thirtyMinMs) {
     dateFormatOptions.second = 'numeric';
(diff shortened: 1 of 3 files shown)


ui/components/src/LineChart.tsx
```diff
@@ -290,11 +293,27 @@ export function LineChart({
   );
 }
 
-function getFormattedDate(value: number) {
-  const XAXIS_DATE_FORMAT = new Intl.DateTimeFormat(undefined, {
+function getStepInterval(data: number[]) {
```
sjcobb, line 296: This only gets called when stepMs isn't passed from panel container (makes embedding the chart independently a little easier)

LukeTillman (approved):
Looks good to me, but wondering if having this be based on the `stepMs` really makes sense long-term? Feels like the fact that the `stepMs` scales currently for larger time ranges is more of a Prometheus plugin implementation detail. Maybe this should really be based on the overall range of values being shown on the x axis and not the step between data points?

Nexucis (approved):
pretty cool !

sjcobb:
> Looks good to me, but wondering if having this be based on the `stepMs` really makes sense long-term? 

Yeah that makes sense @LukeTillman, I just updated it to determine the time granularity based on the xAxis range instead!

PR #444, opened 2022-05-17
components package readme with LineChart import example

sjcobb:
Added a README.md to the ui/components folder that links to new LineChart import example: 
https://codesandbox.io/s/perses-line-chart-component-using-echarts-1tg92v?file=/src/App.js

### DIFF
--- a/ui/components/README.md
+++ b/ui/components/README.md
@@ -10,6 +10,6 @@ To use the LineChart component, begin by importing as shown:
 import { LineChart } from "@perses-dev/components";
 ```
 
-For a full example of one way to map the query data to an [ECharts-compatible format](https://echarts.apache.org/en/option.html#series-line.type), **see [this CodeSandbox link](https://codesandbox.io/s/perses-line-chart-component-using-echarts-1tg92v?file=/src/App.js)**.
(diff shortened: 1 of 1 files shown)


sjcobb:
Yeah this example definitely isn't enough on its own @Nexucis, I'm planning to add additional docs, a page for local dev (eventually Storybook), tests, etc. This was just what I had time for before meetings the other day.

For now I'll update it to pull utils from Perses instead of one-off functions. But to clarify, the use-case you're describing potentially sounds like something you'd want a full panel for?

Nexucis:
> This was just what I had time for before meetings the other day.

Oh I don't blame you, I'm sorry if you feel it this way ! I'm really grateful you had a bit of time to start doing that !
And by the way it would be really a shame from my side to blame you considering the number of times I'm taking to write a simple doc around k8s 😭 

> For now I'll update it to pull utils from Perses instead of one-off functions.

Alright !

> But to clarify, the use-case you're describing potentially sounds like something you'd want a full panel for?

mmm probably. What's the different from your point of view between the simple usage of a `LineChart` and a full panel ? 

As an end user, I'm wondering how much it is useful to just have the `LineChart`. Tell me if I'm wrong, I have the feeling to be able to use the `LineChart` you actually need more stuff than just the actual `LineChart`.

sjcobb:
Yeah the LineChart on it's own is a pretty specific need, but it has been helpful in the Chronosphere app lately, we are using it as a stepping stone until Perses can take over more of the page. But just getting our tooltip in use has been an improvement from what was there before.

Nexucis (approved):
thanks @sjcobb :).
Looking forward for the next docs ;)

PR #448, opened 2022-05-24
add documentation regarding how Perses might work on k8s

Nexucis:

### DIFF
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -1,0 +1,84 @@
+Perses on Kubernetes
+====================
+
+This document aims to describe our vision about how Perses will work on kubernetes.
+
+## 1. Database schema of Perses (outside kubernetes)
+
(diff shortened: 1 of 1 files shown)


Nexucis:
Hopefully I didn't miss a part on how it works, and I hope it will help you :).

Nexucis:
yeah you are right, but I think in that case we should rely on the kubernetes permission system. So you are back in the first point where Perses is using the user token.

The second point is when you give a full access to everybody to the dashboards and this is something that cannot be done using the user token, because like explained it's unlikely a k8s SRE will give this kind of right to all users

AntoineThebaud (approved):

sjcobb (approved):
LGTM thanks for writing this up!

LukeTillman (approved):
Thanks for taking the time to write this up!

PR #453, opened 2022-05-25
add a version number at metadata level an increment it at each update of the resource

Nexucis:


AntoineThebaud (approved):

chucknthem:
Should the update request also fail if the incoming version doesn't match the version in the db? (it means someone else had updated the version since the dashboard was fetched).

That's what grafana does by default unless you tell it to force override.

Nexucis:
Sorry for the long delay to answer you @chucknthem, I forgot a bit :(.

Sooo today it is quite simple, the latest update is the latest version. Meaning that the latest update will override regardless the changes in the previous version.

The backend will just get the previous version of the document directly from the DB to know the version number, it increments this number and then apply the changes and push back the document to the DB

Which means the user don't even have to provide the version number in their documents, which makes sense for me if you think it term of dashboard as code.

Since your document will be in git, you don't really care about the version number, you just want to push the document you have in git in Perses regardless the current version of the document in Perses.

PR #454, opened 2022-05-27
align layout definition between frontend and backend

Nexucis:

### DIFF
--- a/pkg/model/api/v1/dashboard.go
+++ b/pkg/model/api/v1/dashboard.go
@@ -35,11 +35,10 @@ type DashboardSpec struct {
 	Datasource dashboard.Datasource `json:"datasource" yaml:"datasource"`
 	// Duration is the default time you would like to use to looking in the past when getting data to fill the
 	// dashboard
-	Duration   model.Duration                 `json:"duration" yaml:"duration"`
-	Variables  map[string]*dashboard.Variable `json:"variables,omitempty" yaml:"variables,omitempty"`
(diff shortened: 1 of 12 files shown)


AntoineThebaud (approved):

pkg/model/api/v1/dashboard.go
```diff
@@ -142,30 +138,21 @@ func (d *Dashboard) validate() error {
...
-					}
-				}
-			}
-		case *dashboard.ExpandLayoutParameter:
-			for _, ref := range parameter.Children {
-				if err := d.checkAndSetRef(ref); err != nil {
+		switch spec := layout.Spec.(type) {
+		case *dashboard.GridLayoutSpec:
+			for _, item := range spec.Items {
+				if err := d.checkAndSetRef(item.Content); err != nil {
```
sjcobb, line 144: I'm curious long-term what validation lives on the FE vs. BE--for instance, does any of the current FE validation need to be updated as well like this [resolvePanelRef function](https://github.com/perses/perses/blob/v0.4.2/ui/core/src/utils/panel-refs.ts#L24) or does that remain the same?

Nexucis: The minimum required is that the BE contains all necessary validation so it can ensure that the data are valid and can be read and used by the FE.

Then for reactivity and nice user experience, we could have the same amount of validation on the FE.
Looking at the function you pointed @sjcobb I think we are good. I think this question should be raised again when the UI will be able to create / edit a dashboard. Because then this kind of validation will be necessary to validate the form before sending it to the BE

PR #456, opened 2022-05-30
update Dockerfile and archive to have a ready to use container / binaries

Nexucis:
This PR is updating the Dockerfile and the archives built in order to provide a container and an archive ready to use.

The idea is if you do `docker run perses/perses:latest` it's working without any additional parameter, configuration ...etc.


Nexucis:
I'm adding an another image that will be available with the tag `:<version>-debug`. It will contain a bash so you can navigate in the image and probably investigate if you have an issue

AntoineThebaud (approved):

PR #465, opened 2022-06-13
Add ability to show info icon next to panel header 

eunicorn:
If display.description is defined in the dashboard definition, the panel header should show an info icon. When hovering over the info icon, it should display a tooltip with the description.

- moved name and description from each #panel definition to common/display.cue 
- created a Tooltip component 
- fixed minor stylings in Panel 

image

### DIFF
--- a/ui/components/src/InfoTooltip/InfoTooltip.md
+++ b/ui/components/src/InfoTooltip/InfoTooltip.md
@@ -12,9 +12,10 @@ Info tooltip displays informative text when users hover over, focus on, or tap a
 
 ## Props
 
-| Name        |                                                        |
-| ----------- | ------------------------------------------------------ |
-| description | informative text in the tooltip                        |
-| children    | the element with the tooltip                           |
-| title       | [optional] title of the tooltip content                |
-| placement   | [optional] placement of the tooltip. default is bottom |
+| Name        |                                                                                               |
+| ----------- | --------------------------------------------------------------------------------------------- | --- |
+| description | informative text in the tooltip                                                               |
+| children    | the element with the tooltip                                                                  |
+| id          | [optional] used to help implement the accessibility logic. default is a randomly generated id |     |
+| title       | [optional] title of the tooltip content                                                       |
+| placement   | [optional] placement of the tooltip. default is bottom                                        |
--- a/ui/components/src/InfoTooltip/InfoTooltip.test.tsx
+++ b/ui/components/src/InfoTooltip/InfoTooltip.test.tsx
@@ -11,9 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React from 'react';
-import { act, render } from '@testing-library/react';
-import { fireEvent, screen } from '@testing-library/dom';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { screen } from '@testing-library/dom';
 import { InfoTooltip } from './InfoTooltip';
 
 describe('InfoTooltip', () => {
@@ -26,15 +26,8 @@ describe('InfoTooltip', () => {
       </InfoTooltip>
     );
     const tooltipAnchor = screen.getByText('tooltipAnchor');
-    act(() => {
-      fireEvent(
-        tooltipAnchor,
-        new MouseEvent('mouseover', {
-          bubbles: true,
-        })
-      );
-    });
+    userEvent.hover(tooltipAnchor);
     await screen.findByText(title);
-    await screen.findByText(description);
+    screen.findByText(description);
   });
 });
--- a/ui/components/src/InfoTooltip/InfoTooltip.tsx
+++ b/ui/components/src/InfoTooltip/InfoTooltip.tsx
@@ -30,13 +30,19 @@ export enum TooltipPlacement {
 interface InfoTooltipProps {
   description: string;
   children: React.ReactNode;
+  id?: string;
   title?: string;
   placement?: TooltipPlacement;
 }
 
-export const InfoTooltip = ({ title, description, placement, children }: InfoTooltipProps) => {
+export const InfoTooltip = ({ id, title, description, placement, children }: InfoTooltipProps) => {
   return (
-    <StyledTooltip arrow placement={placement} title={<TooltipContent title={title} description={description} />}>
+    <StyledTooltip
+      arrow
+      id={id}
+      placement={placement}
+      title={<TooltipContent title={title} description={description} />}
+    >
       <div>{children}</div>
     </StyledTooltip>
   );
@@ -51,7 +57,6 @@ const TooltipContent = ({ title, description }: Pick<InfoTooltipProps, 'title' |
           sx={(theme) => ({
             color: theme.palette.text.primary,
             fontWeight: theme.typography.fontWeightMedium,
-            lineHeight: '1.25rem',
           })}
         >
           {title}
@@ -61,7 +66,6 @@ const TooltipContent = ({ title, description }: Pick<InfoTooltipProps, 'title' |
         variant="caption"
         sx={(theme) => ({
           color: theme.palette.text.primary,
-          lineHeight: '1.38rem',
         })}
       >
         {description}
@@ -78,7 +82,7 @@ const StyledTooltip = styled(({ className, ...props }: MuiTooltipProps) => (
     color: theme.palette.grey[900],
     maxWidth: '300px',
     padding: theme.spacing(1),
-    boxShadow: '0px 1px 1px 0px #00000033',
+    boxShadow: theme.shadows[1],
   },
   [`& .${tooltipClasses.arrow}`]: {
     color: theme.palette.background.paper,
--- a/ui/dashboards/src/components/Panel/Panel.test.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.test.tsx
@@ -57,6 +57,6 @@ describe('Panel', () => {
   it('should render name and info icon', async () => {
     renderPanel();
     await screen.findByText('Fake Panel');
-    await screen.findByTestId('InformationOutlineIcon');
+    screen.queryByLabelText('info-tooltip');
   });
 });
--- a/ui/dashboards/src/components/VariableOptionsDrawer/VariableOptionsDrawer.test.tsx
+++ b/ui/dashboards/src/components/VariableOptionsDrawer/VariableOptionsDrawer.test.tsx
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { fireEvent, screen } from '@testing-library/react';
-import { act } from 'react-dom/test-utils';
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { JsonObject, VariableDefinition } from '@perses-dev/core';
 import { PluginRegistrationConfig, PluginRegistry } from '@perses-dev/plugin-system';
 import { mockPluginRegistryProps, renderWithContext } from '../../test';
@@ -83,16 +83,9 @@ describe('VariableOptionsDrawer', () => {
     it('should display correct options', async () => {
       renderVariableOptionsDrawer();
       const openButton = await screen.findByRole('button', { name: 'Open' });
-      act(() => {
-        fireEvent(
-          openButton,
-          new MouseEvent('click', {
-            bubbles: true,
-          })
-        );
-      });
+      userEvent.click(openButton);
       await screen.findByText('all');
-      await screen.findByText('node');
+      screen.getByText('node');
     });
   });
 });
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -70,8 +70,14 @@ export function Panel(props: PanelProps) {
               {definition.display.name}
             </Typography>
             {definition.display.description && (
-              <InfoTooltip description={definition.display.description} placement={TooltipPlacement.Right}>
+              <InfoTooltip
+                id="info-tooltip"
+                description={definition.display.description}
+                placement={TooltipPlacement.Right}
+              >
                 <InformationOutlineIcon
+                  aria-describedby="info-tooltip"
+                  aria-hidden={false}
                   sx={{ fontSize: '1rem', position: 'relative', left: '4px', cursor: 'pointer' }}
                 />
               </InfoTooltip>


Nexucis:
hey  @eunicorn thank you for this nice PR!

I unlocked the CI and some are failing.

Can you run the following commands at the root of the project:

```
make fmt
make fixlicense
```

and then commit the changes. That should fix the errors :).

ui/dashboards/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+            color: theme.palette.mode === 'dark' ? theme.palette.common.white : '#2A2E42',
+            fontWeight: '700',
+            fontSize: '14px',
+            lineHeight: '20px',
+          })}
+        >
+          {title}
+        </Typography>
+      )}
+      <Typography
```
Nexucis, line 61: Why not having the check if the description exists here ?

ui/dashboards/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const Tooltip = ({ title, description, placement, children }: TooltipProps) => {
+  return (
+    <StyledTooltip arrow placement={placement} title={<TooltipContent title={title} description={description} />}>
+      <div>{children}</div>
+    </StyledTooltip>
+  );
+};
+
+const TooltipContent = ({ title, description }: Pick<TooltipProps, 'title' | 'description'>) => {
```
Nexucis, line 45: do you foresee a situation where you would have a title and a description at the same time ? Just curious :)

ui/dashboards/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <>
+      {title && (
+        <Typography
+          variant="h6"
+          sx={(theme) => ({
+            color: theme.palette.mode === 'dark' ? theme.palette.common.white : '#2A2E42',
+            fontWeight: '700',
+            fontSize: '14px',
```
Nexucis, line 54: don't you want to have a `fontSize` depending on the size of the screen or coming from material ui (if it makes sense/ it's possible)

ui/dashboards/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+            lineHeight: '20px',
```
Nexucis, line 55: same question for the `lineHeight`. Don't you want something depending on the size of the screen.

ui/dashboards/src/components/Panel.tsx
```diff
@@ -52,20 +54,37 @@ export function Panel(props: PanelProps) {
...
+                whiteSpace="nowrap"
+                overflow="hidden"
+                textOverflow="ellipsis"
+              >
+                {definition.display.name}
+              </Typography>
+              {definition.display.description && (
+                <Tooltip description={definition.display.description} placement={TooltipPlacement.Right}>
+                  <InformationOutlineIcon
+                    sx={{ width: '16px', height: '16px', position: 'relative', left: '4px', cursor: 'pointer' }}
```
Nexucis, line 77: same question here. Don't you want to make the size depending on the size of the screen ? Like should it be. the same size for a small screen or on a phone ?

I'm even wondering if on a phone screen, it makes sense to display the tooltip

Nexucis (commented):
thanks for fixing the CI errors @eunicorn :).
I just added a couple of comments and notes. Let me know what do you think about them.

ui/dashboards/src/components/Tooltip.tsx
eunicorn: Description is a mandatory prop since tooltip should always have a description, so there is no need to check if description exists

ui/dashboards/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+      </Typography>
+    </>
+  );
+};
+
+const StyledTooltip = styled(({ className, ...props }: MuiTooltipProps) => (
+  <MuiTooltip {...props} classes={{ popper: className }} />
+))(({ theme }) => ({
+  [`& .${tooltipClasses.tooltip}`]: {
+    backgroundColor: theme.palette.mode === 'dark' ? '#3E4662' : '#F0F1F6',
```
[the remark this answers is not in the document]
eunicorn: Agreed, I added these colors to the theme :)

ui/dashboards/src/components/Tooltip.tsx
eunicorn: I think currently in our chronosphere UI we have tooltips with title and description, so we might want to use this tooltip instead in the future

sjcobb:
I was thinking more about it and I'm going to rename my [Tooltip](https://github.com/perses/perses/tree/bff747fae16a2fe397ce3d77ec9a04b56883f47e/ui/components/src/Tooltip) component / folder to either GraphTooltip or TimeseriesTooltip, could you rename your's to InfoTooltip and move it to the components package? Or if you can think of a better name than InfoTooltip (DescriptionTooltip seems kind of long) that works too, we just need some way to distinguish it from the one used by LineChart

ui/dashboards/src/components/Panel.tsx
```diff
@@ -52,20 +54,37 @@ export function Panel(props: PanelProps) {
...
-            component="h2"
-            variant="body2"
-            fontWeight={(theme) => theme.typography.fontWeightMedium}
-            whiteSpace="nowrap"
-            overflow="hidden"
-            textOverflow="ellipsis"
-          >
-            {definition.display.name}
-          </Typography>
+          <>
```
sjcobb, line 57: Is the outer <> fragment needed here?

ui/app/sample-data/benchmarkDashboard.ts
```diff
@@ -155,7 +155,7 @@ const benchmarkDashboard: DashboardResource = {
       },
       doubleQueries: {
         kind: 'LineChart',
-        display: { name: 'Double Queries' },
+        display: { name: 'Double Queries', description: 'This is a line chart!!!' },
```
sjcobb, line 158: I love the enthusiasm about being a line chart :laughing::rofl:

ui/dashboards/src/components/Tooltip.tsx
sjcobb: Yeah I need to be better about this, I've used px in some places but agree using theme.typography / MUI Typography variants or rem if needed would be better

ui/dashboards/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+            fontSize: '14px',
+            lineHeight: '20px',
+          })}
+        >
+          {title}
+        </Typography>
+      )}
+      <Typography
+        sx={(theme) => ({
+          color: theme.palette.mode === 'dark' ? theme.palette.common.white : '#2A2E42',
```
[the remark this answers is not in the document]
sjcobb: The theme is pretty bare bones right now so we definitely need to work on building it out / deciding on the Perses color palette. I did add a few things in my [panel enhancements PR](https://github.com/perses/perses/pull/462/files#diff-bb0956411b8189d60e881bfe6005b17b03f397e3d621f88dd1af8c36325af6f3) too like typography.ts (I'll work on getting that branch updated / address feedback later today)

ui/dashboards/src/components/Tooltip.tsx
eunicorn: There is a check if description exists in the Panel component. If description doesn't exist, then we will not show the info icon & tooltip

ui/dashboards/src/components/Tooltip.tsx
eunicorn: yep! I updated the code to use rem/variants instead

ui/dashboards/src/components/Panel.tsx
eunicorn: removed

ui/dashboards/src/components/Tooltip.tsx
Nexucis: Wouldn't be more interesting to use the `them.typography` ? Like that we are aligned with material UI.
Just asking as a UI noob with 2 cents of knowledge :)

ui/dashboards/src/components/Tooltip.tsx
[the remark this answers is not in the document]
Nexucis: cool, thank you :)

ui/dashboards/src/components/Tooltip.tsx
eunicorn: Yes that's what i meant by using variants instead. Each variant should map to the correct font type/size/weight in `theme.typography`

ui/components/src/test/render.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { render, RenderOptions } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from 'react-query';
+
+const queryClient = new QueryClient({ defaultOptions: { queries: { refetchOnWindowFocus: false } } });
+
+/**
+ * Test helper to render a React component with some common app-level providers wrapped around it.
+ */
+export function renderWithContext(ui: React.ReactElement, options?: Omit<RenderOptions, 'queries'>) {
```
Nexucis, line 22: it fiels weird to have this function in this package. Wouldn't it make more sense to have it in the `core` package ?

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -11,19 +11,11 @@
...
-import { Panel } from './Panel';
-
-const TEST_DEFINITION: PanelDefinition = {
-  kind: 'FakePanel',
-  display: {
-    name: 'My test panel',
-  },
-  options: {},
-};
+import { renderWithContext, mockPluginRegistryProps } from '../../test';
```
Nexucis, line 17: for the moment these are coming from the package `@perses-dev/component` so it's better to reflect that in the package import instead of using the relative package important.

ui/dashboards/src/components/VariableOptionsDrawer/VariableOptionsDrawer.test.tsx
```diff
@@ -0,0 +1,85 @@
+import { fireEvent, screen } from '@testing-library/react';
```
Nexucis, line 1: missing license header. `make fixlicense` should fix it.

ui/components/package.json
```diff
@@ -18,7 +18,8 @@
     "clean": "rimraf dist/",
     "build": "tsc --build",
     "build:cjs": "tsc --project ./tsconfig.cjs.json",
-    "test": "echo 'no test to run' && exit 0",
+    "test": "TZ=UTC jest",
+    "test:watch": "TZ=UTC jest --watch",
```
Nexucis, line 22: it would be cool if this script is added in other package where the tests are actually running like in `panels-plugin`

ui/components/src/InfoTooltip/InfoTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const StyledTooltip = styled(({ className, ...props }: MuiTooltipProps) => (
+  <MuiTooltip {...props} classes={{ popper: className }} />
+))(({ theme }) => ({
+  [`& .${tooltipClasses.tooltip}`]: {
+    backgroundColor: theme.palette.secondary.main,
+    color: theme.palette.grey[900],
+    maxWidth: '300px',
+    padding: theme.spacing(1),
+    boxShadow: '0px 1px 1px 0px #00000033',
```
Nexucis, line 81: the box shadow should depend on the theme right ?

Nexucis (commented):
thanks for your continus effort on this PR @eunicorn and many thanks for the new unit tests !

I just added some new comments and questions.

ui/dashboards/src/components/Tooltip.tsx
Nexucis: cool :) thanks for your answer!

ui/components/src/test/render.tsx
eunicorn: Right, that makes sense to me since we also have the same function in `ui/dashboards/src/test/render.tsx`.

@sjcobb @LukeTillman  what do you think?

ui/components/src/InfoTooltip/InfoTooltip.md
```diff
@@ -0,0 +1,20 @@
+# Info Tooltip
```
LukeTillman, line 1: Love that you're adding docs already. 👍

ui/components/src/InfoTooltip/InfoTooltip.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    );
+    const tooltipAnchor = screen.getByText('tooltipAnchor');
+    act(() => {
+      fireEvent(
+        tooltipAnchor,
+        new MouseEvent('mouseover', {
+          bubbles: true,
+        })
+      );
+    });
```
LukeTillman, line 36: Totally a side note, but we should probably look at adding [`@testing-library/user-event`](https://www.npmjs.com/package/@testing-library/user-event) if we haven't yet for these kinds of interactions.

ui/components/src/InfoTooltip/InfoTooltip.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    await screen.findByText(title);
+    await screen.findByText(description);
```
LukeTillman, line 38: Nit: does this need to be async also since you already awaited for the title to show? (Just a minor thing that might keep tests faster if it doesn't need to be the async version).

ui/components/src/InfoTooltip/InfoTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+const TooltipContent = ({ title, description }: Pick<InfoTooltipProps, 'title' | 'description'>) => {
+  return (
+    <>
+      {title && (
+        <Typography
+          variant="body2"
+          sx={(theme) => ({
+            color: theme.palette.text.primary,
+            fontWeight: theme.typography.fontWeightMedium,
+            lineHeight: '1.25rem',
```
LukeTillman, line 54: These line-heights feel a little bit like one-off "magic numbers" to me. 😂  Should we be trying to adjust the theme typography or something instead so this stays consistent or at least relies on the theme instead of being hardcoded? There's also some MUI `Typography` props for adding a gutter to the bottom of spacing is the issue here.

ui/components/src/test/render.tsx
LukeTillman: Yeah, just to chime-in (sorry I've been a little incommunicado lately), I've been trying to keep `core` to just common data models/utils/etc. rather than anything React-specific, that way we can avoid having yet another package with a `react` library dependency. A test utils package might be another way to tackle sharing some of this type of stuff. Grafana does a similar thing with [one of their packages](https://www.npmjs.com/package/@grafana/e2e-selectors).

I didn't realize we'd need these common test things in multiple packages so quickly or I probably would've just set it up that way last week. I guess that's a good thing though since it means we're adding tests faster than I thought. 😂

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -36,23 +28,35 @@ const FAKE_PANEL_PLUGIN: PluginRegistrationConfig<JsonObject> = {
...
 
-  it('should render Panel', async () => {
+  it('should render name and info icon', async () => {
     renderPanel();
-
-    const chart = await screen.findByRole('figure');
-    expect(chart).toBeInTheDocument();
-    expect(chart).toHaveTextContent('FakePanel chart');
+    await screen.findByText('Fake Panel');
+    await screen.findByTestId('InformationOutlineIcon');
```
LukeTillman, line 60: Two quick nits here:

1. Same comment as above about whether this can be sync instead of async since you already did an async call right before this.
2. Any way to do this without test IDs (ideally maybe with a Role query)? Test IDs are really a last ditch thing that I'd love if we largely try to avoid like we have at Chronosphere. Totally cool if not (and I know you had no way of knowing this beforehand).

For more details on number 2, see [the testing library docs](https://testing-library.com/docs/queries/about#priority) on query priorities which we've largely tried to follow in the past.

ui/dashboards/src/components/VariableOptionsDrawer/VariableOptionsDrawer.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+      act(() => {
+        fireEvent(
+          openButton,
+          new MouseEvent('click', {
+            bubbles: true,
+          })
+        );
+      });
+      await screen.findByText('all');
+      await screen.findByText('node');
```
LukeTillman, line 95: Same nit as above around whether this needs to be async.

LukeTillman (approved):
This looks awesome. Thank you for adding so many new tests. Everything I called out inline is a nit I think, so feel free to ignore. And congrats on your first PR in Perses! 😄

ui/components/src/InfoTooltip/InfoTooltip.test.tsx
eunicorn: done :)

ui/components/src/InfoTooltip/InfoTooltip.tsx
eunicorn: I removed the line height for now since you're right that this should be addressed by theme typography

ui/dashboards/src/components/Panel/Panel.test.tsx
eunicorn: For #2, I was able to use queryByLabelText after adding aria-labelledBy to the information icon, which is actually the correct way to do accessibility for tooltip :)

ui/components/src/InfoTooltip/InfoTooltip.test.tsx
LukeTillman: Nice! I thought we might do that in a follow-up, but I like it! 😂

ui/dashboards/src/components/Panel/Panel.test.tsx
LukeTillman: Awesome. And TIL how to correctly do accessibility for tooltip.

ui/dashboards/src/components/Panel/Panel.test.tsx
Nexucis: Fair enough :).

ui/components/src/test/render.tsx
Nexucis: sounds good to me :).

Looking forward to see the improvement around this topic !

Nexucis (approved):
thanks for this work @eunicorn. I have the feeling that all issue and questions raised in this PR have been tackled.
So I'm happy to merge your first PR whenever you feel it's finished!

eunicorn:
> thanks for this work @eunicorn. I have the feeling that all issue and questions raised in this PR have been tackled. So I'm happy to merge your first PR whenever you feel it's finished!

@Nexucis yes that'd be great, thank you :)

Nexucis:
congrats @eunicorn :)

PR #467, opened 2022-06-15
Jest test setup for Dashboards package

LukeTillman:
We've been slacking on writing tests in the UI while things were still changing very quickly, but I think the time has come to start doing that regularly. (For the record, I'm very guilty of this 😂 ). I know Eunice hit a few issues when trying to add some Jest tests as part of #465, so this attempts to get all the plumbing setup so we can start adding tests to our React packages. I added one simple test for `Panel` to the `@perses-dev/dashboards` package as a proof of concept just to show that this is working and to make sure I covered most of the bases on common setup.

Hopefully this makes it easier for us to start expanding the UI integration test suite. 🎉


sjcobb (approved):
:partying_face:

LukeTillman:
Gonna go ahead and merge this to unblock Eunice, but @Nexucis if you see anything that needs addressing, happy to follow-up in a separate PR on it.

Nexucis:
thanks for doing that @LukeTillman.
It looks ok to me. I just think once we have a lot of tests, probably we will have to run them in a single band. When they are running in parallel, apparently it can cause some weird behaviours and your tests are failing without a good indication.

PR #478, opened 2022-07-05
Backend: support charts without datasource + add TextPanel plugin

AntoineThebaud:
- support panels without a datasource
- add TextPanel plugin
- renaming to use the more generic term "panel" rather than "chart" (keep chart where applicable)


AntoineThebaud:
@Nexucis @sjcobb @chucknthem @LukeTillman just a non-critical topic raised by this PR, I think we have a problem of wording :x it was decided at the time to go with the term "Chart", thus we got "LineChart", "GaugeChart", "StatChart".. but a Text panel isn't a chart, so I'm wondering how to properly name this new plugin? (+ on the long run it wont be the only panel in that case I guess)
Should we go with the term "panel" instead and rename the kinds to simply "Line", "Gauge", "Text"... i.e basically what Grafana is doing?

chucknthem:
> @Nexucis @sjcobb @chucknthem @LukeTillman just a non-critical topic raised by this PR, I think we have a problem of wording :x it was decided at the time to go with the term "Chart", thus we got "LineChart", "GaugeChart", "StatChart".. but a Text panel isn't a chart, so I'm wondering how to properly name this new plugin? (+ on the long run it wont be the only panel in that case I guess) Should we go with the term "panel" instead and rename the kinds to simply "Line", "Gauge", "Text"... i.e basically what Grafana is doing?

I'd be in favor of dropping the `Chart` suffix which seems a bit redundant

sjcobb:
I personally prefer 'LineChart' so it matches the underlying component and wouldn’t want to rename the component to Line.tsx since it isn’t descriptive enough. In this case, lately Grafana calls the panel type 'timeseries' and fieldConfig.defaults.custom.drawStyle is 'line', but I like that our approach is simpler. 

For this panel, we were planning to call it 'Markdown', Luke is working on a brief about security considerations that he’s going to share. And there will definitely be other panel types w/o the Chart suffix, for instance, 'Table' is one we want to add soon as well.

Nexucis (approved):
PR looks good to me, thanks @AntoineThebaud !

Regarding the above discussion about the wording, it can be tackle in a different PR once we are all agree on what to do.

PR #481, opened 2022-07-14
gauge panel support additional units

sjcobb:
Currently, our GaugeChart panel only really works correctly with `unit: { kind: 'Percent' }` 

This PR adds support for our other supported units, which required adding the ability to set thresholds.max (needed when unit.kind is _not_ 'Percent' or 'PercentDecimal')

Screenshots:
gauge_bytes_ex

gauge_decimal_ex

### DIFF
--- a/schemas/panels/gauge/gauge.cue
+++ b/schemas/panels/gauge/gauge.cue
@@ -25,8 +25,7 @@ import (
 		calculation: common.#calculation
 		unit?:       common.#unit
 		thresholds?: common.#thresholds
-		// determines end value of last threshold color segment when unit is not a percent
-		max?:        number
+		max?:        number // determines end value of last threshold color segment when unit is not a percent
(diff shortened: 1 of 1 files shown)


LukeTillman (approved):
Looks great! 👍

schemas/common/threshold.cue
```diff
@@ -21,5 +21,6 @@ package common
 
 #thresholds: {
 	default_color?: string
+	max?:           number
```
Nexucis, line 24: it would be cool to have a description as comment to explain the purpose of this field.

sjcobb: yeah good point, max isn't relevant for thresholds in line charts so I moved it to gauge.cue and added a comment with more details

Nexucis (approved):
perfect, thanks !

PR #486, opened 2022-07-21
fix install errors in prometheus-plugin

sjcobb:
Switch prometheus-plugin package.json and imports to use @prometheus-io/lezer-promql

### DIFF
--- a/ui/prometheus-plugin/package.json
+++ b/ui/prometheus-plugin/package.json
@@ -26,7 +26,8 @@
     "@perses-dev/core": "^0.4.2",
     "@perses-dev/plugin-system": "^0.4.2",
     "@prometheus-io/lezer-promql": "^0.37.0",
-    "date-fns": "^2.28.0"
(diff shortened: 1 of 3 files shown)


ui/prometheus-plugin/package.json
```diff
@@ -23,11 +23,13 @@
...
   "dependencies": {
-    "@lezer/lr": "^0.16.3",
+    "@lezer/highlight": "^1.0.0",
+    "@lezer/lr": "^1.2.0",
     "@perses-dev/core": "^0.4.2",
     "@perses-dev/plugin-system": "^0.4.2",
+    "@prometheus-io/lezer-promql": "^0.37.0",
     "date-fns": "^2.28.0",
-    "lezer-promql": "^0.22.0"
+    "lezer": "^0.13.5"
```
Nexucis, line 32: this can be removed. It should have been replaced by `@lezer/lr`

Nexucis (approved):
nice, thanks !

PR #489, opened 2022-07-29
review the way to cross build Perses

Nexucis:
It's just reuse more what is defined in the Makefile, so the CI/CD is a bit more reproductible.


sjcobb (approved):
👍

PR #495, opened 2022-08-04
build errors fix - clean dist folders before start

sjcobb:
There are currently intermittent errors when running Perses on our main branch, this PR ensures all previous dist folders are cleaned before `npm run start -w app` is run.

Related to [this discussion](https://matrix.to/#/!XNmgYIAndZOkEvQAbb:matrix.org/$ZAAq5_auqCXrTz79ufphLMKtyqVj39gFvl4iQOA4qnE?via=matrix.org) in #perses-dev.

Install errors Karan ran into: [main_branch_npm_errors_ex.txt](https://github.com/perses/perses/files/9260024/main_branch_npm_errors_ex.txt)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,7 +9,7 @@
     "test": "npm test --workspaces",
     "lint": "npm run lint --workspaces",
     "lint:fix": "npm run lint:fix --workspaces",
-    "start": "npm run start -w app",
+    "start": "npm run clean && npm run start -w app",
     "clean": "rimraf .build-cache && npm run clean --workspaces"
   },
   "workspaces": [


Nexucis (approved):
ah it's all coming from a missing cleanup :o.
Nice catch thanks @sjcobb !

ui/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "npm test --workspaces",
     "lint": "npm run lint --workspaces",
     "lint:fix": "npm run lint:fix --workspaces",
-    "start": "npm run start -w app",
+    "start": "npm run clean && npm run start -w app",
```
eunicorn, line 12: will there be cases when we don't want to run clean before start? is it better to have a separate command that runs both?

eunicorn (approved):

LukeTillman:
Not sure if we _really_ want to always be doing this before start since it basically eliminates all the benefits we get from Typescript's project cache.

LukeTillman:
Maybe we just update the `README` or other dev instructions to let folks know they can do this (along with removing `node_modules` and running `npm install`) to reset their environments if they have issues?

sjcobb:
Was thinking more and since this happens occasionally in CI too, it’d be good to have some non-manual way to address this—what if we moved ‘npm run clean’ to run after all modules are installed (maybe in postinstall? @LukeTillman)

That wouldn’t affect local dev as much since we don’t reinstall that often and I think this comes about because of weird peerDependency issues

PR #498, opened 2022-08-08
Move variable inputs to the top of dashboard view

LukeTillman:
While we're waiting on some final designs for what variable inputs are going to look like in v1, this just moves them out of the weird drawer we've had for awhile and onto the top of the page when viewing a dashboard. We're talking about doing some more complicated things in a v1 version (e.g. pinning them to the top as you scroll, making them able to be hidden, etc.), but it looks like they're going to be located up top. Here's what this looks like, with bonus collapsing to vertical alignment behavior on mobile:

https://user-images.githubusercontent.com/428023/183473885-dcbcc545-e99f-4cba-9994-ead36ade8d63.mp4


ui/dashboards/src/components/VariableAutocomplete.tsx
```diff
@@ -25,14 +25,17 @@ export interface VariableAutocompleteProps {
...
+
+  className?: AutocompleteProps['className'];
+  sx?: AutocompleteProps['sx'];
 }
 
 /**
  * A MUI Autocomplete that displays variable options, loaded from the
  * appropriate plugin.
  */
 export function VariableAutocomplete(props: VariableAutocompleteProps) {
```
sjcobb, line 37: Curious if you know what's going on with this loading state in nodeExporterDashboard and benchmarkDashboard, is that Interval variable defined correctly?

loading_state_ex

sjcobb (approved):
LGTM! I really like how they look up top, just noted one really small issue when testing with longer values

LukeTillman:
> Tested with longer variable values and looks like in one case the little arrow overlaps with the text:
> 
> long_label_overlap

So I set a min width on these of 250px which is what I think they're currently rendering at. I think for the "real" version of this, we're gonna need to figure out how to properly size these AND what to do with long values (like you just tested). Since we don't currently know what these are even going to look like while we're waiting on Mohit's designs, I'm gonna hold off on trying to address this. I'm just afraid we may end up with some completely different control other than Autocomplete or some other look for showing the "selected" state.

PR #499, opened 2022-08-12
line chart tooltip performance improvement

sjcobb:
We noticed performance issues when using Perses on dashboard with many charts, where multiple queries returned thousands of series. The way the tooltip is implemented is the main cause, this PR doesn't fix the issue completely but has a few improvements. I'm planning to look into additional tooltip fixes and adding lazy loading [using IntersectionObserver](https://stackoverflow.com/questions/53757229/reactjs-how-to-render-a-component-only-when-scroll-down-and-reach-it-on-the-page/65628717#65628717)

### Commits Summary
- add React.memo in Tooltip.tsx
- remove an unnecessary prop
- add check in Tooltip.tsx for if the cursor is over a canvas element (using mousePos.target)
  - there is already a check in getFocusedSeriesData for whether the current target is a chart, but adding the condition in Tooltip.tsx allows getFocusedSeriesData to not be called as often

### DIFF
--- a/ui/components/src/Tooltip/Tooltip.tsx
+++ b/ui/components/src/Tooltip/Tooltip.tsx
@@ -29,7 +29,10 @@ interface TooltipProps {
 const Tooltip = React.memo(function Tooltip({ chartRef, chartData, wrapLabels }: TooltipProps) {
   const [pinnedPos, setPinnedPos] = useState<CursorCoordinates | null>(null);
   const mousePos = useMousePosition();
+
   if (mousePos === null || mousePos.target === null) return null;
+
+  // ensure user is hovering over a chart before checking for nearby series
   if ((mousePos.target as HTMLElement).tagName !== 'CANVAS') return null;
 
   const chart = chartRef.current;
--- a/ui/components/src/Tooltip/Tooltip.tsx
+++ b/ui/components/src/Tooltip/Tooltip.tsx
@@ -16,28 +16,21 @@ import { Box, Portal } from '@mui/material';
 import { ECharts as EChartsInstance } from 'echarts/core';
 import { EChartsDataFormat } from '../model/graph';
 import { getFocusedSeriesData } from './focused-series';
-import {
-  CursorCoordinates,
-  CursorData,
-  TooltipData,
-  TOOLTIP_MAX_HEIGHT,
-  TOOLTIP_MAX_WIDTH,
-  useMousePosition,
-} from './tooltip-model';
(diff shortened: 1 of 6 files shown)


ui/components/src/Tooltip/Tooltip.tsx
```diff
@@ -16,28 +16,21 @@ import { Box, Portal } from '@mui/material';
...
 }
 
-export function Tooltip(props: TooltipProps) {
-  const { chartRef, chartData } = props;
+const Tooltip = React.memo(function Tooltip({ chartRef, chartData, wrapLabels }: TooltipProps) {
   const [pinnedPos, setPinnedPos] = useState<CursorCoordinates | null>(null);
   const mousePos = useMousePosition();
-  if (mousePos === null) return null;
+  if (mousePos === null || mousePos.target === null) return null;
+  if ((mousePos.target as HTMLElement).tagName !== 'CANVAS') return null;
```
sjcobb, line 33: main improvement is here to first check if the cursor is above a canvas element before calling getFocusedSeriesData

eunicorn:
looks good :)

ui/components/src/Tooltip/utils.ts
```diff
@@ -0,0 +1,57 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CursorCoordinates, CursorData, TOOLTIP_MAX_WIDTH } from './tooltip-model';
+
+export function assembleTransform(
```
LukeTillman, line 16: Nit: would be great to just get a quick summary comment on what this is doing. (Bonus points for adding tests 😂 )

LukeTillman (approved):
Looks like a good low cost fix. Have you thought at all about whether chart tooltips could actually be some kind of "global" thing (i.e. we only ever render one on a dashboard that the charts can then hook into)? Not sure if you'd ever want more than one on the screen at once...

sjcobb:
> Have you thought at all about whether chart tooltips could actually be some kind of "global" thing (i.e. we only ever render one on a dashboard that the charts can then hook into)? Not sure if you'd ever want more than one on the screen at once...

Yeah I actually started looking at that but got nervous since it was such a huge change--but you're right, if this doesn't improve performance enough, I'll need to change the approach (my browser doesn't crash anymore, but it could be better).

I also have been thinking it'd be cool to have an option to show multiple tooltips in sync--similar to how Grafana has this 'Shared Tooltip' setting at the dashboard level:
grafana_shared_tooltip

PR #500, opened 2022-08-12
lazy load panels on scroll

sjcobb:
To help with performance this PR lazy loads panel content on scroll using react-intersection-observer

### DIFF
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -39,8 +39,8 @@ export function Panel(props: PanelProps) {
   }, [width, height]);
 
   const { ref, inView } = useInView({
-    threshold: 0,
-    initialInView: true,
+    threshold: 0.3,
+    initialInView: false,
     triggerOnce: true,
   });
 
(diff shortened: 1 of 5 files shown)


ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -37,11 +38,18 @@ export function Panel(props: PanelProps) {
     return { width, height };
   }, [width, height]);
 
+  const { ref, inView } = useInView({
+    threshold: 0,
+    initialInView: true,
+    triggerOnce: true,
```
LukeTillman, line 44: Ahh nice. Does this option mean that once it's in view and rendered, it doesn't unmount (and then get re-rendered later)?

LukeTillman (approved):
💥 This is gonna be great for dashboards with lots of charts, nice work.

eunicorn:
nice, the api is so simple!

ui/dashboards/src/components/Panel/Panel.tsx
sjcobb: Yep that's right! Don't want to rerun a bunch of queries if we don't have to

PR #501, opened 2022-08-12
Release v0.5.2

sjcobb:
- increase versions to 0.5.2
- update CHANGELOG.md with performance improvement notes


LukeTillman:
Probably also need to include this in the release notes: https://github.com/perses/perses/pull/498

Technically, that's probably a breaking change since it removed the old drawer-related components from the dashboards package, so may want to do this as `0.6.0`. (Not sure if we've already merged the last release branch back into `main`?)

LukeTillman (approved):
Nevermind, as you helpfully pointed out, those other two PRs were against the `0.5` branch. Sorry for the confusion!

PR #505, opened 2022-08-15
fix tooltip glitch, add feat to pin and enter tooltip on click

sjcobb:
### Overview

Last week's tooltip performance [PR 499](https://github.com/perses/perses/pull/499) caused a regression where the tooltip is no longer enterable and sometimes gets stuck when moving your cursor quickly. An additional check is added to the 'user is hovering over a canvas element' condition to fix this.

While fixing this issue, I also added a new feature--the ability to pin the tooltip on click. When a users clicks close to a series, the tooltip will remain in place. The user can then enter the tooltip and copy labels (or other actions in the future like drill-down, link to query builder, etc). To unpin the tooltip, double click or move your cursor away from the active chart.

### Screen Recording

https://user-images.githubusercontent.com/9369625/184645715-4801c222-72a8-42e3-905b-d01c2cb10d63.mp4


ui/components/src/LineChart/LineChart.tsx
```diff
@@ -100,7 +101,18 @@ export function LineChart({ height, data, unit, grid, legend, visualMap, onDataZ
     enableDataZoom(chartRef.current);
   }
 
+  const handleOnClick = () => {
+    if (pinTooltip === false) {
+      setPinTooltip(true);
+    } else {
+      setPinTooltip(false);
+    }
```
LukeTillman, line 109: Nit: I think you want to use the callback form of `setState` here just to be 100% sure you're always seeing the latest state, e.g.:

```
setPinTooltip(current => !current);
```

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -100,7 +101,18 @@ export function LineChart({ height, data, unit, grid, legend, visualMap, onDataZ
...
+  };
+
   const handleOnDoubleClick = () => {
+    if (pinTooltip === true) {
+      setPinTooltip(false);
```
LukeTillman, line 114: Same nit here, should use the callback form since you're relying on the current state value.

LukeTillman (approved):
Looks good, one minor nit in a couple places. 👍

PR #509, opened 2022-08-24
time range selector v1: relative time dropdown, abs start / end calendar

sjcobb:
Version 1 of time range selection tasks, which includes:
- dropdown to select from a predefined list of time ranges, ex) 'Last 1 hour', 'Last 6 hours'
- re-render data from selected time range
- input custom time ranges by selecting specific dates/times for start and end
- custom start and end calendar UX using MUI StaticDateTimePicker

Upcoming v2 TODOs:
- shareable URLs
- zoom event on a chart should update the active time range and re-render panels
- form validation improvements
- create separate component for customized StaticDateTimePicker 
- support passing relative time ranges in setTimeRange
- add tests
- UX refinements

Screen recording:

https://user-images.githubusercontent.com/9369625/186436527-2aabbbf6-f919-400a-8e75-9cb1b9dc7e50.mp4


ui/components/src/TimeRangeSelector/AbsoluteTimePicker.tsx
```diff
@@ -0,0 +1,149 @@
...
+          >
+            <h3>Select End Time</h3>
+            <StaticDateTimePicker
+              displayStaticWrapperAs="desktop"
+              openTo="day"
+              disableHighlightToday={true}
+              value={initialTimeRange.end}
+              onChange={(newValue) => {
+                if (newValue === null) return;
+                setTimeRange({ start: timeRange.start, end: newValue });
```
LukeTillman, line 100: Same nit as above.

ui/components/src/TimeRangeSelector/AbsoluteTimePicker.tsx
```diff
@@ -0,0 +1,149 @@
...
+            if (isValidDateRange === true) {
+              const updatedRange = { start: startDate, end: timeRange.end };
+              setTimeRange(updatedRange);
+              onChange(updatedRange);
+            }
+          }}
+          value={format(timeRange.start, DATE_TIME_FORMAT)}
+          label="Start Time"
+          placeholder="mm/dd/yyyy hh:mm"
+          type="tel"
```
LukeTillman, line 122: Should this actually be `tel` since it's a date input?

ui/components/src/TimeRangeSelector/TimeRangeSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+        open={showCustomDateSelector}
+        onClose={() => setShowCustomDateSelector(false)}
+        sx={(theme) => ({
+          padding: theme.spacing(2),
+        })}
+      >
+        <AbsoluteTimePicker initialTimeRange={absoluteTimeRange} onChange={handleTimePickerChange} />
+      </Popover>
+      <FormControl fullWidth>
+        <InputLabel>{FORM_CONTROL_LABEL}</InputLabel>
```
LukeTillman, line 83: Is it possible to leave the `FormControl` and `InputLabel` stuff to the consumer and just make this the input component?

ui/core/src/model/time.ts
```diff
@@ -26,6 +26,14 @@ export interface RelativeTimeRange {
...
 }
 
+export interface TimeOption {
+  from: string;
+  to: string;
+  display: string;
+}
+
+export type DateTimeFormat = number | string;
+
```
LukeTillman, line 36: Feels like this is specific to the UI component and maybe should be part of its public API?

ui/dashboards/src/components/TimeRangeControls.tsx
```diff
@@ -0,0 +1,38 @@
...
+  { from: 'now-30m', to: 'now', display: 'Last 30 minutes' },
+  { from: 'now-1h', to: 'now', display: 'Last 1 hour' },
+  { from: 'now-6h', to: 'now', display: 'Last 6 hours' },
+  { from: 'now-12h', to: 'now', display: 'Last 12 hours' },
+  { from: 'now-1d', to: 'now', display: 'Last 1 day' },
+  { from: 'now-7d', to: 'now', display: 'Last 7 days' },
+  { from: 'now-14d', to: 'now', display: 'Last 14 days' },
+];
+
+export function TimeRangeControls() {
```
LukeTillman, line 32: Love how you split this into Dashboards from the core component. Nice work.

ui/core/src/model/time.ts
```diff
@@ -26,6 +26,14 @@ export interface RelativeTimeRange {
   pastDuration: DurationString;
 }
 
+export interface TimeOption {
+  from: string;
+  to: string;
+  display: string;
+}
```
LukeTillman, line 33: Suggestion: rather than having a new string format for this, why not use the existing `RelativeTimeRange` type from core? For example:

```ts
export interface TimeOption {
  value: RelativeTimeRange;
  display: string;
}
```

The new from/to string formats feel unstructured to me and it feels like there's no reason we need to do this like Grafana?

ui/dashboards/src/context/TimeRangeStateProvider.tsx
```diff
@@ -26,15 +26,41 @@ export interface TimeRangeProviderProps {
...
 
-  return <TimeRangeContext.Provider value={ctx}>{children}</TimeRangeContext.Provider>;
+  const ctx = useMemo(() => ({ timeRange, defaultDuration: initialValue.pastDuration }), [timeRange, initialValue]);
+
+  const setters = useMemo(() => ({ setTimeRange }), [setTimeRange]);
+
+  return (
+    <TimeRangeSetterContext.Provider value={setters}>
+      <TimeRangeContext.Provider value={ctx}>{children}</TimeRangeContext.Provider>
+    </TimeRangeSetterContext.Provider>
```
LukeTillman, line 44: Going to think about it some more, but I wonder if there's some way to do this doesn't involve two different contexts/providers. The plugin runtime stuff definitely makes things more complicated...

(Side note: this is totally cool for now, just thinking out loud.)

ui/components/src/TimeRangeSelector/TimeRangeSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { AbsoluteTimeRange, TimeOption, convertTimeShortcut, parseDurationString } from '@perses-dev/core';
+import { AbsoluteTimePicker } from './AbsoluteTimePicker';
+
+const DATE_TIME_FORMAT = 'yyyy-MM-dd HH:mm:ss';
+const FORM_CONTROL_LABEL = 'Time Range';
+
+interface TimeRangeSelectorProps {
+  timeOptions: TimeOption[];
+  onChange: (value: AbsoluteTimeRange) => void;
+  defaultTimeOption?: TimeOption;
```
LukeTillman, line 26: This part of the API feels a little weird to me for two reasons:

1. I was expecting that the value could be an absolute or relative time range (since the user can select either).
2. I was expecting this to have both `onChange` and a `value` prop so that the state was controlled completely outside this input component.

I don't think we want to have an `onChange` without the component's value also being controlled from outside. Otherwise we're duplicating state, having to sync stuff, etc.

LukeTillman: Just as an example, basically this is what I was kinda expecting as the API:

```ts
interface TimeRangeSelectorProps {
  timeOptions: TimeOption[];
  value: AbsoluteTimeRange | RelativeTimeRange;
  onChange: (value: AbsoluteTimeRange | RelativeTimeRange) => void;
}
```

Maybe you also end up with something for that union type of relative/absolute too. (e.g. `type TimeRangeValue = AbsoluteTimeRange | RelativeTimeRange`

LukeTillman (approved):
Overall, this looks great. I left a few nits inline, but I think my two big concerns are:

1. The props API and where the state lives. Would be great if it was entirely external to the component.
2. The options API and whether we really need a new string format that kinda feels unstructured/magic.

Whenever you feel like this is stable enough, would be great to add some tests too. But overall, this looks awesome, nice work!

ui/dashboards/src/context/TimeRangeStateProvider.tsx
sjcobb: Yeahh I agree it's confusing, wonder if this could be refactored when we bring in Zustand?

ui/core/src/model/time.ts
sjcobb: The thought was keeping TIME_OPTIONS similar to how Grafana does it would make implementing shareable URLs easier since I got feedback from Charles that it'd be good to keep our URL structure the same as Grafana. But yeah, the approach currently isn't great since that needs to be customizable too so I'll change this

ui/components/src/TimeRangeSelector/AbsoluteTimePicker.tsx
sjcobb: Yeah this is weird, I'll add a comment. Basically I inspected the regular MUI DateTimePicker props to build this StaticDateTimePicker and it actually uses 'tel' so that's what I landed on when 'date' was working as expected

ui/components/src/TimeRangeSelector/AbsoluteTimePicker.tsx
```diff
@@ -0,0 +1,149 @@
...
+              onChange={(newValue) => {
+                if (newValue === null) return;
+                setTimeRange({ start: newValue, end: timeRange.end });
+              }}
+              onAccept={() => handleOnStartAccept()}
+              renderInput={(params) => <TextField {...params} />}
+            />
+          </Box>
+        )}
+        {showEndCalendar && (
```
eunicorn, line 78: Do we need showEndCalendar? Can we just do `!showStartCalendar`?

ui/core/src/model/time.ts
LukeTillman: Yeah, for URLs, we can always serialize/deserialize values to the query string however we want. Maybe those do end up looking more like what Grafana does. 🤷‍♂️

ui/components/src/TimeRangeSelector/AbsoluteTimePicker.tsx
sjcobb: Good call! I changed this so only showStartCalendar is used now

ui/components/src/TimeRangeSelector/TimeRangeSelector.tsx
sjcobb: Yup good call, I moved all of this to TimeRangeControls.tsx

ui/components/src/TimeRangeSelector/TimeRangeSelector.tsx
sjcobb: I made progress on this, but left a few TODOs for a follow up PR.

PR #510, opened 2022-08-26
Implemented edit mode and added react-grid-layout

eunicorn:
When edit button is clicked, dashboard is in edit mode. You can then reposition and resize panels. 

Added [zustand](https://github.com/pmndrs/zustand) for state management and [react-grid-layout](https://github.com/react-grid-layout/react-grid-layout) for drag and drop. 

edit-mode

### DIFF
--- a/ui/app/src/views/ViewDashboard.tsx
+++ b/ui/app/src/views/ViewDashboard.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { DashboardResource } from '@perses-dev/core';
-import { DashboardProvider, ViewDashboard as DashboardView } from '@perses-dev/dashboards';
+import { ViewDashboard as DashboardView } from '@perses-dev/dashboards';
 import Footer from '../components/Footer';
 import { useSampleData } from '../utils/temp-sample-data';
 
@@ -30,11 +30,9 @@ function ViewDashboard() {
   }
 
   return (
-    <DashboardProvider initialState={{ dashboardSpec: dashboard.spec }}>
-      <DashboardView dashboardResource={dashboard}>
-        <Footer />
-      </DashboardView>
-    </DashboardProvider>
+    <DashboardView dashboardResource={dashboard}>
+      <Footer />
+    </DashboardView>
   );
 }
 
--- a/ui/app/src/views/ViewDashboard.tsx
+++ b/ui/app/src/views/ViewDashboard.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { DashboardResource } from '@perses-dev/core';
-import { ViewDashboard as DashboardView } from '@perses-dev/dashboards';
+import { DashboardProvider, ViewDashboard as DashboardView } from '@perses-dev/dashboards';
 import Footer from '../components/Footer';
 import { useSampleData } from '../utils/temp-sample-data';
 
@@ -30,9 +30,11 @@ function ViewDashboard() {
   }
 
   return (
-    <DashboardView dashboardResource={dashboard}>
-      <Footer />
-    </DashboardView>
+    <DashboardProvider initialState={{ dashboardSpec: dashboard.spec }}>
+      <DashboardView dashboardResource={dashboard}>
+        <Footer />
+      </DashboardView>
+    </DashboardProvider>
   );
 }
 
--- a/ui/components/src/EChart.tsx
+++ b/ui/components/src/EChart.tsx
@@ -174,6 +174,14 @@ export const EChart = React.memo(function EChart<T>({
     };
   }, [onEvents]);
 
+  useEffect(() => {
+    const updateSize = debounce(() => {
+      if (chartElement.current === null) return;
+      chartElement.current.resize();
+    }, 200);
+    updateSize();
+  }, [sx]);
+
   return <Box ref={containerRef} sx={sx}></Box>;
 });
 
--- a/ui/dashboards/src/components/GridLayout/GridLayout.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridLayout.tsx
@@ -10,13 +10,17 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+import 'react-grid-layout/css/styles.css';
+import 'react-resizable/css/styles.css';
 
 import { useState } from 'react';
+import { Responsive, WidthProvider } from 'react-grid-layout';
 import { Box, BoxProps, Collapse } from '@mui/material';
 import { GridDefinition, GridItemDefinition } from '@perses-dev/core';
+import { useEditMode } from '../../context';
 import { GridTitle } from './GridTitle';
 
-const COLUMNS = 24;
+const ResponsiveGridLayout = WidthProvider(Responsive);
 
 export interface GridLayoutProps extends BoxProps {
   definition: GridDefinition;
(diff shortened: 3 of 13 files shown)


ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -41,23 +45,36 @@ describe('Panel', () => {
...
-    renderPanel();
+    renderPanel(initialState);
     await screen.findByText('Fake Panel');
     screen.queryByLabelText('info-tooltip');
   });
+
+  it('should render edit panel icon when in edit mode', async () => {
+    initialState.isEditMode = true;
+    renderPanel(initialState);
+    await screen.getByTestId('PencilIcon');
```
eunicorn, line 78: will update this test to not use testId when we implement the edit panel feature

ui/components/src/EChart.tsx
```diff
@@ -174,6 +174,14 @@ export const EChart = React.memo(function EChart<T>({
     };
   }, [onEvents]);
 
+  useEffect(() => {
```
LukeTillman, line 177: Oooh, good idea!

ui/dashboards/src/components/GridLayout/GridLayout.tsx
```diff
@@ -10,13 +10,17 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+import 'react-grid-layout/css/styles.css';
+import 'react-resizable/css/styles.css';
```
LukeTillman, line 14: I suspect, but don't know, that this might be an issue for library/embed use cases. I'm interested to see what the library build output looks like cause the ability to load CSS like this typically comes from an app's webpack build. We may need to think about how to do this in a way that still works with Dashboards the library.

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -66,6 +81,7 @@ export function Panel(props: PanelProps) {
             sx={{
               display: 'flex',
               alignItems: 'center',
+              height: '20px',
```
LukeTillman, line 84: Nit: This feels a little like a "magic number" to me and like we don't want to hardcode it, but maybe I'm wrong? If it's hardcoded, could be difficult for folks with a different theme to make it look correct (e.g. different typography, etc.)

ui/dashboards/src/views/ViewDashboard.tsx
```diff
@@ -27,9 +29,45 @@ export interface ViewDashboardProps extends BoxProps {
...
+              </Button>
+              <Button
+                variant="contained"
+                sx={{
+                  margin: '0 4px',
+                }}
+              >
+                Save
+              </Button>
+            </Box>
```
LukeTillman, line 68: Nit/suggestion: might consider MUI `Stack` here for the spacing between buttons as an alternative to the custom styles.

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -30,9 +30,11 @@ function ViewDashboard() {
   }
 
   return (
-    <DashboardView dashboardResource={dashboard}>
-      <Footer />
-    </DashboardView>
+    <DashboardProvider initialState={{ dashboardSpec: dashboard.spec }}>
```
LukeTillman, line 33: Do you think this should actually go inside `DashboardView` from the `dashboards` package? Not clear to me really, but thought I should ask since the other providers are all in there.

ui/dashboards/src/views/ViewDashboard.tsx
```diff
@@ -49,8 +87,31 @@ export function ViewDashboard(props: ViewDashboardProps) {
...
+              flexDirection: 'column',
             }}
           >
+            {isEditMode ? (
+              <Button
+                sx={{
+                  alignSelf: 'flex-end',
+                }}
+              >
+                <AddIcon sx={{ fontSize: '1rem', marginRight: '8px' }} />
```
LukeTillman, line 100: A little worried with all the `fontSize: '1rem'` custom styles that I've seen for icons that maybe something's off with our theme or something? Just noticing a pattern.

LukeTillman (approved):
This is SO awesome and I am super excited to see visual editing getting started! I left a few suggestions inline, but nothing huge honestly. Awesome work getting this started. Let me know if I can clarify anything.

ui/app/src/views/ViewDashboard.tsx
eunicorn: I had to put the `DashboardProvider` here because `DashboardView` needs it to access the store.

I can move other providers to `ViewDashboard` though 😜

ui/dashboards/src/views/ViewDashboard.tsx
eunicorn: I removed all the custom styles for Icons and created overrides for it

ui/dashboards/src/components/Panel/Panel.tsx
eunicorn: I updated it to using `minHeight`. This is because there is a bug right now where the panel header height is different depending if there is the info icon.

ui/dashboards/src/components/GridLayout/GridLayout.tsx
eunicorn: follow up ticket - https://app.shortcut.com/chronosphere/story/29863/convert-react-grid-layout-css-files-to-using-mui-global-styles

ui/dashboards/src/components/GridLayout/GridLayout.tsx
```diff
@@ -77,20 +66,18 @@ export function GridLayout(props: GridLayoutProps) {
...
-              sm: 36,
-            },
-            columnGap: (theme) => theme.spacing(1),
-            rowGap: (theme) => theme.spacing(1),
-          }}
+        <ResponsiveGridLayout
+          className="layout"
+          breakpoints={{ lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0 }}
+          cols={{ lg: 24, md: 24, sm: 24, xs: 24, xxs: 2 }}
+          rowHeight={30}
```
cndonovan, line 73: I feel like I'm just going to make random comments because I'm still learning the whole codebase, but do you think it's worth storing the breakpoints as constants somewhere? 

Might be fine as is

ui/dashboards/src/context/DashboardProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface DashboardActions {
+  setEditMode: (isEditMode: boolean) => void;
+  setLayouts: (layouts: LayoutDefinition[]) => void;
+  setPanels: (panels: Record<string, PanelDefinition>) => void;
+  addPanel: (name: string, panel: PanelDefinition) => void;
+}
+
+export type DashboardStoreState = DashboardState & DashboardActions;
+
+export interface DashboardStoreProp {
```
cndonovan, line 35: DashboardStoreProps?

ui/dashboards/src/components/GridLayout/GridLayout.tsx
eunicorn: that makes sense, but for now i'll leave it as is since i made up these breakpoints haha we can revisit when we fine tune this

ui/app/src/views/ViewDashboard.tsx
saminzadeh: So when integrating Perses as a framework into external apps, we import the [`Dashboard`](https://github.com/perses/perses/blob/164c34505d55867283f10acc760e154604b7ff5a/ui/dashboards/src/components/Dashboard.tsx#L26) component from the `dashboards` package then wire up the Providers by hand. Is our thought that `ViewDashboard` can be seen as a "helper preset component" to do this for you. If so, I can see this Provider moving into that component.

ui/dashboards/src/context/DashboardProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import create from 'zustand';
+import type { StoreApi } from 'zustand';
+import createZustandContext from 'zustand/context';
+import produce from 'immer';
+import { DashboardSpec, LayoutDefinition, PanelDefinition } from '@perses-dev/core';
+
+interface DashboardState {
```
saminzadeh, line 20: For my own reference, are you thinking that this is where we will store most global dashboard editing state? For example, when we add template variables and we need to show a modal/drawer to edit those values.

saminzadeh (approved):
Looks great @eunicorn - I'm also excited to give Zustand a shot in a real world app.

ui/app/src/views/ViewDashboard.tsx
eunicorn: I don't think we really talked about that yet, but you make a good point. 

I can move the logic that needs access to the store into a separate component so that I can move `DashboardProvider` to `ViewDashboard`

ui/dashboards/src/context/DashboardProvider.tsx
eunicorn: correct :)

PR #513, opened 2022-09-01
Fix bug where bytes would be formatted with 'undefined' as unit

cndonovan:
We noticed a bug where bytes were sometimes formatted incorrectly:
Screen Shot 2022-09-01 at 10 17 55 AM

Turns out, it happens when the number of byes is very small -- between -1 and 1. 
Screen Shot 2022-09-01 at 10 18 12 AM

With the fix, we see the correct value and correct unit for bytes:
Screen Shot 2022-09-01 at 9 00 22 AM

## Why?
First, we have an array of units like `['Bytes', 'KB', 'MB', ...]`.

Then, we have a function for getting the index of the unit to use based on the number of bytes. 

When the number of bytes is between -1 and 1, our function returns -1. 

-1 is not an index, and so we get `undefined` as the unit. 

I did a little Wolfram Alpha graphing of this function to see this edge case too:
Screen Shot 2022-09-01 at 11 33 05 AM

### DIFF
--- a/ui/components/src/model/units.ts
+++ b/ui/components/src/model/units.ts
@@ -227,7 +227,9 @@ function formatBytes(bytes: number, decimals = 2) {
   const dm = decimals < 0 ? 0 : decimals;
   const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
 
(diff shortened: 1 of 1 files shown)


LukeTillman (approved):
The Wolfram Alpha graph was a nice touch 😂  Thanks for fixing!

PR #517, opened 2022-09-06
Add Panel

eunicorn:
In edit mode, user can add new group or add new panel. Clicking add panel button will open up the add panel drawer where you can input the panel name and description. 

add panel

### DIFF
--- a/ui/dashboards/src/components/AddPanel/AddPanel.tsx
+++ b/ui/dashboards/src/components/AddPanel/AddPanel.tsx
@@ -25,7 +25,7 @@ import {
   Typography,
 } from '@mui/material';
 import { Drawer } from '@perses-dev/components';
-import { ChangeEvent, useState } from 'react';
+import { ChangeEvent, FormEvent, useState } from 'react';
 import { useLayouts, usePanels } from '../../context';
 import { removeWhiteSpacesAndSpecialCharacters } from '../../utils/functions';
 
@@ -54,8 +54,10 @@ const AddPanel = ({ isOpen, onClose }: AddPanelProps) => {
     setPanelDescription(e.target.value);
   };
 
-  const onAddPanelClick = () => {
-    addPanel(removeWhiteSpacesAndSpecialCharacters(panelName), {
+  const onAddPanelClick = (e: FormEvent) => {
+    e.preventDefault();
+    const panelKey = removeWhiteSpacesAndSpecialCharacters(panelName);
+    addPanel(panelKey, {
       kind: 'EmptyChart',
       display: { name: panelName, description: panelDescription },
       options: {},
@@ -73,7 +75,7 @@ const AddPanel = ({ isOpen, onClose }: AddPanelProps) => {
       y: maxY + 1,
       width: 12,
       height: 6,
-      content: { $ref: `#/spec/panels/${panelName}` },
+      content: { $ref: `#/spec/panels/${panelKey}` },
     });
     onClose();
   };
--- a/ui/dashboards/src/components/AddPanel/AddPanel.tsx
+++ b/ui/dashboards/src/components/AddPanel/AddPanel.tsx
@@ -1,0 +1,138 @@
+// Copyright 2022 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {
+  MenuItem,
+  Stack,
+  Select,
+  TextField,
+  InputLabel,
+  FormControl,
+  Grid,
+  SelectChangeEvent,
+  Box,
+  Button,
+  Typography,
+} from '@mui/material';
+import { Drawer } from '@perses-dev/components';
+import { ChangeEvent, FormEvent, useState } from 'react';
+import { useLayouts, usePanels } from '../../context';
+import { removeWhiteSpacesAndSpecialCharacters } from '../../utils/functions';
+
+interface AddPanelProps {
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+const AddPanel = ({ isOpen, onClose }: AddPanelProps) => {
+  const { layouts, addItemToLayout } = useLayouts();
+  const { addPanel } = usePanels();
+
+  const [group, setGroup] = useState(0);
+  const [panelName, setPanelName] = useState('');
+  const [panelDescription, setPanelDescription] = useState('');
+
+  const onSelectGroupChange = (e: SelectChangeEvent<number>) => {
+    setGroup(e.target.value as number);
+  };
+
+  const onPanelNameChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setPanelName(e.target.value);
+  };
+
+  const onPanelDescriptionChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setPanelDescription(e.target.value);
+  };
+
+  const onAddPanelClick = () => {
+    addPanel(removeWhiteSpacesAndSpecialCharacters(panelName), {
+      kind: 'EmptyChart',
+      display: { name: panelName, description: panelDescription },
+      options: {},
+    });
+
+    // find maximum y so new panel is added to the end of the grid
+    let maxY = 0;
+    layouts[group]?.spec.items.forEach((layout) => {
+      if (layout.y > maxY) {
+        maxY = layout.y;
+      }
+    });
+    addItemToLayout(group, {
+      x: 0,
+      y: maxY + 1,
+      width: 12,
+      height: 6,
+      content: { $ref: `#/spec/panels/${panelName}` },
+    });
+    onClose();
+  };
+
+  return (
+    <Drawer isOpen={isOpen} onClose={onClose}>
+      <form onSubmit={onAddPanelClick}>
+        <AddPanelHeader onClose={onClose} />
+        <Grid container spacing={2}>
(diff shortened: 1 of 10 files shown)


ui/dashboards/src/context/DashboardProvider.tsx
```diff
@@ -49,33 +52,56 @@ export function usePanels() {
...
 
 export function useEditMode() {
   const { isEditMode, setEditMode } = useStore(({ isEditMode, setEditMode }) => ({ isEditMode, setEditMode }));
   return { isEditMode, setEditMode };
 }
 
+export function useDashboard() {
+  const getDashboard = useStore((state) => state.getDashboard);
+  return { getDashboard };
+}
```
cndonovan, line 74: I might be missing something, but why not return the dashboard itself, vs. a function to call to get the dashboard?

I ask because I noticed that in other cases, like useLayouts and usePanels, we just get the object back, vs. we get a getter and then call the getter.

I'd be down with either pattern of "getting", just wondering if it can be the same pattern everywhere.

ui/dashboards/src/context/DashboardProvider.tsx
```diff
@@ -49,33 +52,56 @@ export function usePanels() {
...
   return (
     <Provider
       createStore={() =>
-        create((set) => ({
+        create((set, get) => ({
           layouts,
           panels,
           isEditMode: !!isEditMode,
           setEditMode: (isEditMode: boolean) => set({ isEditMode }),
           setLayouts: (layouts: LayoutDefinition[]) => set({ layouts }),
```
cndonovan, line 92: I wonder if we'll even need setLayouts, if we'll just use addLayout / removeLayout type of functions instead 🤔 

We can wait and see how it turns out, and you might have a plan for it. Mostly just noting to myself that maybe we'll end up not needing this.

cndonovan (approved):
So exciting!

ui/dashboards/src/components/AddPanel/AddPanel.tsx
```diff
@@ -0,0 +1,138 @@
...
+    setPanelName(e.target.value);
+  };
+
+  const onPanelDescriptionChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setPanelDescription(e.target.value);
+  };
+
+  const onAddPanelClick = (e: FormEvent) => {
+    e.preventDefault();
+    addPanel(panelName, {
```
LukeTillman, line 58: Rather than using the raw panel name input, do you think it might be worthwhile to try and generate a "key" here that avoids using spaces and other special characters?

ui/dashboards/src/components/AddPanel/AddPanel.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const onPanelNameChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setPanelName(e.target.value);
+  };
+
+  const onPanelDescriptionChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setPanelDescription(e.target.value);
+  };
+
+  const onAddPanelClick = (e: FormEvent) => {
+    e.preventDefault();
```
LukeTillman, line 57: Just curious, why prevent default?

ui/dashboards/src/components/Dashboard.tsx
```diff
@@ -26,14 +27,19 @@ export interface DashboardProps extends BoxProps {
...
     <Box {...others}>
       <ErrorBoundary FallbackComponent={ErrorAlert}>
-        {spec.layouts.map((layout, idx) => (
+        {layouts.map((layout, idx) => (
           <GridLayout
             key={idx}
             definition={layout}
-            renderGridItemContent={(definition) => <GridItemContent content={definition.content} spec={spec} />}
+            renderGridItemContent={(definition) => (
+              <GridItemContent content={definition.content} spec={getDashboard() || spec} />
```
LukeTillman, line 41: Not sure I understand this? Is `getDashboard` just for edit mode while `spec` is for view mode? Feels like maybe this component should just render whatever is passed in via props? If I were using the component (passing in a spec) but then it stopped rendering my spec, I think I'd be surprised by that behavior?

ui/dashboards/src/components/DashboardToolbar.tsx
```diff
@@ -40,14 +62,20 @@ export const DashboardToolbar = () => {
...
-            Add Panel
-          </Button>
+            <Button onClick={onAddGroup}>
+              <AddIcon sx={{ marginRight: '8px' }} />
+              Add Group
+            </Button>
+            <Button onClick={onAddPanel}>
+              <AddIcon sx={{ marginRight: '8px' }} />
+              Add Panel
+            </Button>
```
LukeTillman, line 77: I think you want to use the `startIcon` property on `Button` here rather than putting it into children. Should take care of the margin for you without needing custom styles.

ui/dashboards/src/context/DashboardProvider.tsx
```diff
@@ -26,8 +26,11 @@ interface DashboardState {
 interface DashboardActions {
   setEditMode: (isEditMode: boolean) => void;
   setLayouts: (layouts: LayoutDefinition[]) => void;
+  addLayout: (layout: LayoutDefinition) => void;
+  addItemToLayout: (index: number, item: GridItemDefinition) => void;
   setPanels: (panels: Record<string, PanelDefinition>) => void;
   addPanel: (name: string, panel: PanelDefinition) => void;
+  getDashboard: () => DashboardSpec;
```
LukeTillman, line 33: I'd second Christine's question about why this is a function rather than just a property on the state? I know you pulled layouts/panels out in your initial PR, but maybe the overall spec should just be in the state instead?

ui/dashboards/src/context/DashboardProvider.tsx
```diff
@@ -84,6 +110,14 @@ export function DashboardProvider(props: DashboardProviderProps) {
               }, {})
             );
           },
+          getDashboard: () => {
+            const panels = get().panels;
+            return produce(dashboardSpec, (draftState) => {
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              draftState.panels = panels as any;
```
LukeTillman, line 117: Feels like something weird is going on if we need to cast to `any`?

LukeTillman (approved):
Overall, this looks great. I left a few comments/questions inline. Let me know if you want to talk any of it through and once you think this is ready to merge. 🎉

ui/dashboards/src/context/DashboardProvider.tsx
eunicorn: updated it to return the dashboard itself

ui/dashboards/src/context/DashboardProvider.tsx
eunicorn: my original thinking is that we'll have to update the whole layout when user repositions the panels in a grid, but that's only one layout, so we probably have to refactor this :)

ui/dashboards/src/components/AddPanel/AddPanel.tsx
eunicorn: this prevents url from changing when submitting a form

ui/dashboards/src/components/AddPanel/AddPanel.tsx
eunicorn: i think in the future we also need to add validations, like check for duplicate panel name

ui/dashboards/src/components/Dashboard.tsx
eunicorn: you're right, i reverted the change and had to move the components out of <ViewDashboard /> component into <DashboardApp /> so that I could call `useDashboard` inside the `DashboardStoreProvider` and pass the dashboard spec into this Dashboard component

ui/dashboards/src/views/DashboardApp.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Box } from '@mui/material';
+import { combineSx } from '@perses-dev/components';
+import { useState } from 'react';
+import { VariableList, Dashboard } from '../components';
+import AddPanel from '../components/AddPanel/AddPanel';
+import { DashboardToolbar } from '../components/DashboardToolbar';
+import { useDashboard } from '../context';
+import { ViewDashboardProps } from './ViewDashboard';
+
+export const DashboardApp = (props: ViewDashboardProps) => {
```
sjcobb, line 23: nit but this seems a little weird to name this DashboardApp since there is the separate perses-dev/app package, would something like ‘DashboardContent.tsx’ or just ‘Dashboard.tsx’ work (can the old Dashboard component be deleted now btw?)

sjcobb: nvm DashboardContent isn't great either since it would be outside of Dashboard.tsx, can't think of anything better at the moment

ui/dashboards/src/components/AddPanel/AddPanel.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const panelKey = removeWhiteSpacesAndSpecialCharacters(panelName);
+    addPanel(panelKey, {
+      kind: 'EmptyChart',
+      display: { name: panelName, description: panelDescription },
+      options: {},
```
sjcobb, line 63: This is really cool, I'm interested in how we'll store default panel options for other panel types--which I guess comes with the plugin-system enhancements @LukeTillman is proposing.

sjcobb (approved):
🚀 this is great, really exciting to see! Sorry for causing a bunch of merge conflicts 😆

sjcobb:
Ah dang, it looks like the `fixed form submit error` commit needs to be signed off on

eunicorn:
> Ah dang, it looks like the `fixed form submit error` commit needs to be signed off on

oops i always forget! done

ui/dashboards/src/views/DashboardApp.tsx
eunicorn: The idea is that `ViewDashboard` will have all the providers, `DashboardApp` is where we have local states, and `Dashboard` is just a component. I think it's also confusing how we have two ViewDashboard files.

i'm open to new name proposals :)

PR #518, opened 2022-09-07
[app] Move footer component into App vs. ViewDashboard

cndonovan:

# Changes
* Move footer component from `ViewDashboard.tsx` to `App.tsx`
* Add semantic tags for `main` and `footer`
* Add constant for the default dashboard ID

# Before
Screen Shot 2022-09-06 at 6 01 00 PM
Screen Shot 2022-09-06 at 6 00 47 PM

# After
Screen Shot 2022-09-06 at 6 00 19 PM
Screen Shot 2022-09-06 at 6 00 30 PM


Nexucis (approved):
ah yeah good point indeed !
I'm wondering if we will keep the footer in long term.

sjcobb (approved):
LGTM!

PR #519, opened 2022-09-07
[contributing.md] Add explanation for jq package

cndonovan:

I added bullet point about `jq`. I want people to know they can delete this if we ever stop using it in the `populate.sh` script.


Nexucis:
lgtm.

But I don't see why we would stop using it.

sjcobb (approved):
👍 👍

PR #523, opened 2022-09-09
governance doc table update

sjcobb:
Update table in GOVERNANCE.md to match MAINTAINERS.md. 

@Nexucis are both of these tables with info about team members needed? Seems like the one in MAINTAINERS.md will be easier to remember to update

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -42,6 +42,7 @@ The current team members are:
 | Steven Cobb | [sjcobb](https://github.com/sjcobb) | Chronosphere |
 | Julius Volz | [juliusv](https://github.com/juliusv) | PromLabs |
 | Antoine Thebaud | [AntoineThebaud](https://github.com/AntoineThebaud) | Amadeus IT Group |
+| Eunice Wong | [eunicorn](https://github.com/eunicorn) | Chronosphere |
 
 ## Maintainers
 


eunicorn (approved):

GOVERNANCE.md
```diff
@@ -42,6 +42,7 @@ The current team members are:
 | Steven Cobb | [sjcobb](https://github.com/sjcobb) | Chronosphere |
 | Julius Volz | [juliusv](https://github.com/juliusv) | PromLabs |
 | Antoine Thebaud | [AntoineThebaud](https://github.com/AntoineThebaud) | Amadeus IT Group |
+| Eunice Wong | [eunicorn](https://github.com/eunicorn) | Chronosphere |
```
eunicorn, line 45: 🥳

Nexucis:
yeah both are mandatory.

That's because the Maintainers list can be different from the Core Team list.
In our case since we have only one important repo, it's not super useful. But it can change once we will have others repo.
Besides, the Maintainers list could explained the areas of the different maintainers. Like that: https://github.com/prometheus/prometheus/blob/main/MAINTAINERS.md

It doesn't list everybody from the Prometheus Team, just some key members on different areas. We could do the same if we think it is relevant and associate different team members on different areas.

PR #524, opened 2022-09-09
Removed react-grid-layout and react-resizable css imports

eunicorn:
```
import 'react-grid-layout/css/styles.css';
import 'react-resizable/css/styles.css';
```

We needed to import the two css files in order for react-grid-layout to work. This meant that users consuming the dashboard library would also need to import these files in order for dashboard to work or have css loader in webpack

This PR fixed this issue by passing the css to MUI Global Styles.


sjcobb (approved):
LGTM!

PR #526, opened 2022-09-12
Remove unused validate prop from plugin definitions

LukeTillman:
There's currently an unused prop on plugins called `validate` which originally (in theory) was supposed to hook into some sort of runtime validation that plugin authors could provide to validate plugin options objects. It was sort of a half-baked idea I had at the time I did the original plugin system implementation. We haven't really spent a lot of time thinking about options validation yet (although it's possible we get to something cuelang based?), and this prop is currently complicating expanding plugins to support visual editing.

Let's just remove it for now so we can move forward on expanding plugins and revisit validation as a separate topic.


LukeTillman:
Still have no idea why front-end builds seem to randomly fail on the first run for a PR, then work fine when you re-run them. 🙃

sjcobb (approved):
👍

eunicorn (approved):

Nexucis:
> Still have no idea why front-end builds seem to randomly fail on the first run for a PR, then work fine when you re-run them.

I'm wondering if it comes with the cache used in the github action 
https://github.com/perses/perses/blob/main/.github/workflows/build_and_release.yaml#L20-L25

Even if I don't see the connections, but in npm worlds I usually don't see it haha

Nexucis:
nice cleanup otherall :)

PR #527, opened 2022-09-12
Update usePlugin to only need kind, not definition

LukeTillman:
When using a plugin (i.e. the `usePlugin` hook), you really only need the plugin type and kind, not a full-blown definition object. This just updates that API to reflect that. This will be particularly important when you _don't_ have a definition object yet, for example, when you're adding a panel and you need to get the plugin in order to create the initial options for the panel (which ultimately results in a definition object).


Nexucis:
ah yeah that makes sense ! Cool :)

PR #528, opened 2022-09-12
Extend panel plugins to add placeholders for visual editing

LukeTillman:
This extends panel plugins to add placeholders for what I _think_ are going to be the APIs needed for visual editing. Basically, it extends the definition of a `PanelPlugin` to add:

- `createInitialOptions`: a callback to create the initial/empty options object for a panel
- `PanelOptionsEditor`: a React component for visually editing the Options object

I went ahead and put some "stub" implementations in for our existing panel plugins. Since this was already a decent-sized PR, I didn't try integrating this with the visual editing functionality that's there yet. I'll try and do that in a follow-up PR shortly.

### DIFF
--- a/ui/panels-plugin/src/index.ts
+++ b/ui/panels-plugin/src/index.ts
@@ -1,4 +1,4 @@
-// Copyright 2021 The Perses Authors
+// Copyright 2022 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -12,41 +12,49 @@
 // limitations under the License.
 
 import { PluginSetupFunction } from '@perses-dev/plugin-system';
-import { EmptyChart, EmptyChartKind } from './plugins/empty-chart/EmptyChart';
-import { GaugeChartPanel, GaugeChartKind } from './plugins/gauge-chart/GaugeChartPanel';
-import { LineChartPanel, LineChartKind } from './plugins/line-chart/LineChartPanel';
-import { StatChartKind, StatChartPanel } from './plugins/stat-chart/StatChartPanel';
+import { createInitialEmptyChartOptions, EmptyChart, EmptyChartOptionsEditor } from './plugins/empty-chart';
+import { createInitialGaugeChartOptions, GaugeChartPanel, GaugeChartOptionsEditor } from './plugins/gauge-chart';
+import { LineChartPanel, LineChartOptionsEditor, createInitialLineChartOptions } from './plugins/line-chart';
+import { createInitialStatChartOptions, StatChartOptionsEditor, StatChartPanel } from './plugins/stat-chart';
 
 export const setup: PluginSetupFunction = (registerPlugin) => {
   registerPlugin({
     pluginType: 'Panel',
-    kind: LineChartKind,
+    kind: 'LineChart',
     plugin: {
       PanelComponent: LineChartPanel,
+      OptionsEditorComponent: LineChartOptionsEditor,
+      createInitialOptions: createInitialLineChartOptions,
     },
   });
 
   registerPlugin({
     pluginType: 'Panel',
-    kind: GaugeChartKind,
+    kind: 'GaugeChart',
     plugin: {
       PanelComponent: GaugeChartPanel,
+      OptionsEditorComponent: GaugeChartOptionsEditor,
+      createInitialOptions: createInitialGaugeChartOptions,
     },
   });
 
   registerPlugin({
     pluginType: 'Panel',
-    kind: StatChartKind,
+    kind: 'StatChart',
     plugin: {
       PanelComponent: StatChartPanel,
+      OptionsEditorComponent: StatChartOptionsEditor,
+      createInitialOptions: createInitialStatChartOptions,
     },
   });
 
   registerPlugin({
     pluginType: 'Panel',
-    kind: EmptyChartKind,
+    kind: 'EmptyChart',
     plugin: {
       PanelComponent: EmptyChart,
+      OptionsEditorComponent: EmptyChartOptionsEditor,
+      createInitialOptions: createInitialEmptyChartOptions,
     },
   });
 };
--- a/ui/panels-plugin/src/plugins/empty-chart/EmptyChart.tsx
+++ b/ui/panels-plugin/src/plugins/empty-chart/EmptyChart.tsx
@@ -1,4 +1,4 @@
-// Copyright 2021 The Perses Authors
+// Copyright 2022 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -12,14 +12,10 @@
 // limitations under the License.
 
 import { Box } from '@mui/material';
-import { JsonObject } from '@perses-dev/core';
 import { PanelProps } from '@perses-dev/plugin-system';
-
-export const EmptyChartKind = 'EmptyChart' as const;
+import { EmptyChartOptions } from './empty-chart-model';
 
 export type EmptyChartProps = PanelProps<EmptyChartOptions>;
-
-type EmptyChartOptions = JsonObject;
 
 export function EmptyChart(props: EmptyChartProps) {
   return <Box sx={{ overflow: 'hidden' }}>{props.definition.kind}</Box>;
--- a/ui/panels-plugin/src/plugins/empty-chart/empty-chart-model.ts
+++ b/ui/panels-plugin/src/plugins/empty-chart/empty-chart-model.ts
@@ -11,8 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export * from './graph-queries';
-export * from './panels';
-export * from './plugins';
-export * from './variables';
-export * from './visual-editing';
+import { JsonObject } from '@perses-dev/core';
+
+export type EmptyChartOptions = JsonObject;
+
+export function createInitialEmptyChartOptions(): EmptyChartOptions {
+  return {};
+}
--- a/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
@@ -1,4 +1,4 @@
-// Copyright 2021 The Perses Authors
+// Copyright 2022 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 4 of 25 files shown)


LukeTillman:
I integrated this with the existing visual editing functionality in #529, so check this one out first, then that one.

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
```diff
@@ -0,0 +1,43 @@
...
+  thresholds?: ThresholdOptions;
+  max?: number;
+}
+
+/**
+ * Creates the initial/empty options for a GaugeChart panel.
+ */
+export function createInitialGaugeChartOptions(): GaugeChartOptions {
+  return {
+    // TODO: How do you represent an initially empty/unset graph query?
```
sjcobb, line 36: I compared with Grafana and it looks like they completely leave out the `targets` array before a query is set, which seems counterintuitive

ui/panels-plugin/src/index.ts
```diff
@@ -12,41 +12,49 @@
...
 
 export const setup: PluginSetupFunction = (registerPlugin) => {
   registerPlugin({
     pluginType: 'Panel',
-    kind: LineChartKind,
+    kind: 'LineChart',
     plugin: {
       PanelComponent: LineChartPanel,
+      OptionsEditorComponent: LineChartOptionsEditor,
+      createInitialOptions: createInitialLineChartOptions,
```
sjcobb, line 27: This setup will work basically the same for our "third party" plugins right or are there any differences between them and our 4 core panel plugins?

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
LukeTillman: Yeah, I think we need to think about "empty states" in the JSON a bit since I suspect this is going to come up a bunch during visual editing.

ui/panels-plugin/src/index.ts
LukeTillman: Yup, should be the same for all plugins.

sjcobb (approved):
🚀 played around with your other branch locally, this is awesome!

ui/plugin-system/src/model/visual-editing.ts
```diff
@@ -0,0 +1,35 @@
...
+  // TODO: These are temporary and may not actually make sense, so replace
+  // with whatever makes sense as visual editing evolves
+  value: Options;
+  onChange: (next: Options) => void;
+}
+
+/**
+ * Callback for creating initial/empty options for a plugin.
+ */
+export type InitialOptionsCallback<Options extends JsonObject> = () => Options;
```
eunicorn, line 35: Is there a reason why this needs to be a callback? Can the type just be `Options`?

eunicorn (approved):

ui/panels-plugin/src/index.ts
```diff
@@ -12,41 +12,49 @@
...
 
 import { PluginSetupFunction } from '@perses-dev/plugin-system';
-import { EmptyChart, EmptyChartKind } from './plugins/empty-chart/EmptyChart';
-import { GaugeChartPanel, GaugeChartKind } from './plugins/gauge-chart/GaugeChartPanel';
-import { LineChartPanel, LineChartKind } from './plugins/line-chart/LineChartPanel';
-import { StatChartKind, StatChartPanel } from './plugins/stat-chart/StatChartPanel';
+import { createInitialEmptyChartOptions, EmptyChart, EmptyChartOptionsEditor } from './plugins/empty-chart';
+import { createInitialGaugeChartOptions, GaugeChartPanel, GaugeChartOptionsEditor } from './plugins/gauge-chart';
+import { LineChartPanel, LineChartOptionsEditor, createInitialLineChartOptions } from './plugins/line-chart';
+import { createInitialStatChartOptions, StatChartOptionsEditor, StatChartPanel } from './plugins/stat-chart';
```
cndonovan, line 18: nice! (love index files imports)

ui/panels-plugin/src/index.ts
```diff
@@ -12,41 +12,49 @@
...
 
 export const setup: PluginSetupFunction = (registerPlugin) => {
   registerPlugin({
     pluginType: 'Panel',
-    kind: LineChartKind,
+    kind: 'LineChart',
```
cndonovan, line 23: Nice I'll make sure MarkdownPanel is just 'Markdown' for kind too

ui/panels-plugin/src/plugins/empty-chart/EmptyChart.tsx
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 The Perses Authors
+// Copyright 2022 The Perses Authors
```
cndonovan, line 1: Haha this must be out-of-date in a few places huh

ui/panels-plugin/src/plugins/empty-chart/empty-chart-model.ts
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { JsonObject } from '@perses-dev/core';
+
+export type EmptyChartOptions = JsonObject;
```
cndonovan, line 16: I like the having explicit name for this type (even tho it's just JsonObject) and the explict model file!

ui/panels-plugin/src/plugins/empty-chart/EmptyChart.tsx
```diff
@@ -12,15 +12,11 @@
...
 import { Box } from '@mui/material';
-import { JsonObject } from '@perses-dev/core';
 import { PanelProps } from '@perses-dev/plugin-system';
-
-export const EmptyChartKind = 'EmptyChart' as const;
+import { EmptyChartOptions } from './empty-chart-model';
 
 export type EmptyChartProps = PanelProps<EmptyChartOptions>;
 
-type EmptyChartOptions = JsonObject;
```
cndonovan, line 22: Yay this file got simpler in nice way

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
```diff
@@ -12,27 +12,17 @@
...
 
 export type GaugeChartPanelProps = PanelProps<GaugeChartOptions>;
 
-interface GaugeChartOptions extends JsonObject {
-  query: GraphQueryDefinition;
-  calculation: CalculationType;
-  unit?: UnitOptions;
-  thresholds?: ThresholdOptions;
-  max?: number;
-}
```
cndonovan, line 34: Again, dope how this file got simper as all the type concerns move into the model file!

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
cndonovan, line 36: The stuff below itself doesn't look bad to me :)

ui/panels-plugin/src/plugins/line-chart/LineChartOptionsEditor.tsx
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+import { OptionsEditorProps } from '@perses-dev/plugin-system';
+import { Box } from '@mui/material';
+import { LineChartOptions } from './line-chart-model';
+
+export type LineChartOptionsEditorProps = OptionsEditorProps<LineChartOptions>;
+
+export function LineChartOptionsEditor(props: LineChartOptionsEditorProps) {
+  const { value } = props;
```
cndonovan, line 21: Btw, what is value in all these? Like what is it representing? I can probably also just look this info up by looking at some models though.

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
cndonovan: I think I remember my last company had a utility function for generating a default empty query too, and we used that in different places (like maybe we'd want to use between these different charts)

ui/plugin-system/src/model/visual-editing.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A component for visual editing of a plugin's options.
+ */
+export type OptionsEditor<Options extends JsonObject = JsonObject> = React.ComponentType<OptionsEditorProps<Options>>;
+
+/**
+ * Common props passed to options editor components.
+ */
+export interface OptionsEditorProps<Options extends JsonObject = JsonObject> {
```
cndonovan, line 25: I'm pretty Typescript-new, but is it something like this:

`Options` is a generic type that will depend on the panel itself, and then this interface somehow guarantees that we use the same `Options` between panel and panel options editor?

cndonovan (approved):
much approve

ui/plugin-system/src/model/visual-editing.ts
LukeTillman: Great question. The TLDR is: I'm worried about having a static object floating around that someone could (accidentally or otherwise) modify and cause all sorts of weird issues. Seemed better to have a factory function here and avoid it.

ui/panels-plugin/src/plugins/empty-chart/EmptyChart.tsx
LukeTillman: So. Many. Places. 😂

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
LukeTillman: That's a good idea. I've definitely seen that kinda pattern too before.

ui/panels-plugin/src/plugins/line-chart/LineChartOptionsEditor.tsx
LukeTillman: Value is just the generic `Options` type that plugin authors control. At runtime (in the app/framework code), we don't know what they used, so it's just a `JsonObject` then.

ui/plugin-system/src/model/visual-editing.ts
LukeTillman: Yup, you got it!

LukeTillman: If you're curious, it's the `Options` parameter here on `PanelPlugin` (that's then used on all its props) that ensures for a given plugin, they use the same options type for all the pieces:

https://github.com/perses/perses/pull/528/files#diff-0598e07a5c7c0745e67a4438e870edef8e13758c88f3ae2a0b111453c5d06041R21-R25

cndonovan: Ah cool

PR #530, opened 2022-09-13
time range selector v2: support data zoom, shareable links

sjcobb:
Follow up to PR #509, includes support for: 
- [FEATURE] time range selection shareable URLs
- [FEATURE] zoom event on a chart should update the active time range and re-render panels

Additional details:
- routing using react-router v6 is added (but only in ui/app/package.json so apps other than our's can use their own solution)
- now only one TimeRangeContext instead of separate for setter
- ViewDashboard inside app has been refactored
  - Open question: should this be renamed since there is also a ViewDashboard inside the dashboards package?
- onTimeRangeChange callback added to TimeRangeStateProvider
- setQueryParams is used to update the URL on data zoom

### Future TODOs
- [x] add QueryStringProvider so apps can choose their own query param implementations, if undef fallback to setTimeRange 
- [ ] ~~add undo when LineChart is double clicked~~ see PR #537 
  - leaving double click behavior undefined for now, to discuss other unzoom options
- [ ] setup dashboard routing correctly and completely remove old dashboard url param approach
- [ ] add more tests
- [ ] UX and form validation refinements
- [ ] add zoom transition using an [ECharts animation](https://echarts.apache.org/handbook/en/how-to/animation/transition/)

Screen recording:

https://user-images.githubusercontent.com/9369625/189939044-b78378a0-161c-4d55-b215-a44c15129002.mov


ui/app/src/views/ViewDashboard.tsx
```diff
@@ -21,16 +25,54 @@ const DEFAULT_DASHBOARD_ID = 'node-exporter-full';
...
   // TODO: Loading indicator
   if (dashboard === undefined) {
     return null;
   }
 
-  return <DashboardView dashboardResource={dashboard} />;
+  const handleOnDateRangeChange = (event: TimeRangeValue) => {
+    // TODO: refactor, preserve all existing params
+    if (isRelativeValue(event)) {
+      setSearchParams({
```
sjcobb, line 54: This PR isn't ready to review yet, I'm planning to move this to the other ViewDashboard.tsx inside the dashboards package and use this [query-string package](https://www.npmjs.com/package/query-string) instead of having separate implementations for React Router v5 and v6. This will allow for a more generalizable solution that we can also make use of for template variables. 



useQueryStringState

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const DEFAULT_DASHBOARD_ID = 'node-exporter-full';
+
+function HomeDashboard() {
+  const { getInstalledPlugins, importPluginModule } = useBundledPlugins();
+
+  const [searchParams] = useSearchParams();
+
+  const dashboardParam = searchParams.get('dashboard');
+  const dashboard = useSampleData<DashboardResource>(dashboardParam || DEFAULT_DASHBOARD_ID);
```
sjcobb, line 45: Planning to remove this temporary ?dashboard=<dashboard name> URL params approach now that react-router v6 is added

ui/app/src/views/ViewDashboard.tsx
sjcobb: See PR description--discussed this with Luke and decided to stick with the onDateRangeChange approach in this PR for now and add a better long term approach in the future using either a new QueryStringProvider or adding a queryParamOptions prop (this would also need to work with apps that use react router v5 and not break the back button / useLocation history)

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { useBundledPlugins } from '../model/bundled-plugins';
+import { useSampleData } from '../utils/temp-sample-data';
+
+// app specific echarts option overrides, empty since perses uses default
+// https://apache.github.io/echarts-handbook/en/concepts/style/#theme
+const ECHARTS_THEME_OVERRIDES: EChartsTheme = {};
+
+const DEFAULT_DASHBOARD_ID = 'node-exporter-full';
+
+function HomeDashboard() {
```
sjcobb, line 39: Any thoughts on how to name these components? I like DashboardApp.tsx better but that is used in the dashboards package. Also, it's confusing to me that we have a ViewDashboard.tsx in ui/app and another in ui/dashboards, but not sure of better names / which should be changed?

ui/app/src/components/Header.tsx
```diff
@@ -108,12 +108,18 @@ export default function Header(): JSX.Element {
     <AppBar position="relative">
       <Toolbar>
         <Box sx={style} flexGrow={1}>
-          <Typography variant="h6" sx={{ marginRight: '1rem' }}>
```
sjcobb, line 111: Fixes 'Perses' logo text since it was slightly too high and was bothering me 😆

ui/app/src/components/Header.tsx
```diff
@@ -108,12 +108,18 @@ export default function Header(): JSX.Element {
...
+          <Typography variant="h6" sx={{ margin: '1rem' }}>
             Perses
           </Typography>
           <Divider orientation="vertical" flexItem sx={{ borderRightColor: 'rgba(255,255,255,0.2)' }} />
           <ProjectMenu />
         </Box>
+        <Button href="/?dashboard=demoDashboard" color="inherit">
```
sjcobb, line 117: I'm planning to implement dashboard routing properly in a follow-up PR

ui/plugin-system/src/runtime/time-range.ts
```diff
@@ -12,11 +12,13 @@
...
 
 import { createContext, useContext } from 'react';
-import { AbsoluteTimeRange, DurationString } from '@perses-dev/core';
+import { AbsoluteTimeRange, TimeRangeValue } from '@perses-dev/core';
 
 export interface TimeRange {
-  timeRange: AbsoluteTimeRange;
-  defaultDuration: DurationString;
+  timeRange: TimeRangeValue;
+  absoluteTimeRange: AbsoluteTimeRange;
```
saminzadeh, line 19: If a plugin is only concerned with the resolved timestamps (the date object/unix epoch ms), do they use `absoluteTimeRange`?

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
```diff
@@ -0,0 +1,59 @@
...
+// limitations under the License.
+
+import React, { useState, useMemo, useCallback } from 'react';
+import { TimeRangeValue, AbsoluteTimeRange, toAbsoluteTimeRange, isRelativeTimeRange } from '@perses-dev/core';
+import { TimeRange, TimeRangeContext } from '../runtime/time-range';
+
+export interface TimeRangeProviderProps {
+  initialValue: TimeRangeValue;
+  children?: React.ReactNode;
+  onDateRangeChange?: (e: TimeRangeValue) => void;
```
saminzadeh, line 21: nit: Since the type is called `TimeRange`, is it better for this to be called `onTimeRangeChange` ?

saminzadeh (approved):
LGTM - left a small suggestion

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
sjcobb: yeah good call, this is updated now

ui/app/src/views/HomeDashboard.tsx
eunicorn: Yeah DashboardApp here makes sense too. Naming is tough lol

Can we combine the two ViewDashboard into one file so all the providers are in one place?

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
```diff
@@ -0,0 +1,59 @@
...
+export interface TimeRangeProviderProps {
+  initialValue: TimeRangeValue;
+  children?: React.ReactNode;
+  onTimeRangeChange?: (e: TimeRangeValue) => void;
+}
+
+/**
+ * Provider implementation that supplies the TimeRangeState at runtime.
+ */
+export function TimeRangeStateProvider(props: TimeRangeProviderProps) {
```
eunicorn, line 27: oh cool now everything is in one provider! i like this approach!

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -76,32 +56,25 @@ export function TimeRangeControls() {
...
-              setAbsoluteTime({ start: timeRange.start, end: timeRange.end });
-            }
-            setSelectedTimeRange(timeRange);
             setShowCustomDateSelector(false);
           }}
         />
       </Popover>
       <FormControl fullWidth>
-        <InputLabel>{FORM_CONTROL_LABEL}</InputLabel>
+        <InputLabel id="select-time-range" variant="filled">
```
eunicorn, line 64: should the InputLabel be part of TimeRangeSelector?

eunicorn (approved):

ui/app/src/views/HomeDashboard.tsx
sjcobb: Agreed I hate naming things 😆 

So the reason TimeRangeStateProvider is separate and inside the Perses app is since I wanted to use [useSearchParams](https://reactrouter.com/en/6.4.0/hooks/use-search-params) from React Router v6 in handleonTimeRangeChange and I didn't want to make that a required dep in @perses-dev/dashboards (especially since we're on v5 in our UI)

eunicorn: oh i see, how about PersesApp.tsx?

ui/plugin-system/src/runtime/time-range.ts
sjcobb: yeah that's right, maybe I should rename or potentially consolidate these 🤔  (I also just added previousTimeRange in my [unzoom PR](https://github.com/perses/perses/pull/537))

ui/app/src/views/HomeDashboard.tsx
Nexucis: I wouldn't name it `PersesApp`, just because it's not the home page of Perses which is already called `App.tsx`.

Perhaps we should rename the file in the package `dashboard` and have here `DashboardViews` or `DashboardPage`.

There is a documentation about the naming I think written by @LukeTillman [here](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md). Perhaps it will help :)

ui/app/src/views/Docs.tsx
```diff
@@ -0,0 +1,18 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function Docs() {
+  return <h1>Docs</h1>;
+}
+
+export default Docs;
```
Nexucis, line 18: this page exists just for the purpose of testing the routing right ?

ui/app/src/App.tsx
```diff
@@ -11,53 +11,16 @@
...
-                <ViewDashboard />
-              </DataSourceRegistry>
-            </PluginBoundary>
-          </PluginRegistry>
-        </ChartsThemeProvider>
-      </Box>
-      <Footer />
-    </Box>
+    <Routes>
+      <Route path="/" element={<HomeDashboard />} />
```
Nexucis, line 21: Ideally the routing should be something like `/projects/<project_name>/dashboards/<dashboard_name>`.

Like that when sharing URL, the FE will be able to know what is the dashboard to load

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,89 @@
...
+  // TODO: completely remove old dashboard url param approach
+  const dashboardParam = searchParams.get('dashboard');
+  const dashboard = useSampleData<DashboardResource>(dashboardParam || DEFAULT_DASHBOARD_ID);
+
+  const muiTheme = useTheme();
+  const chartsTheme: PersesChartsTheme = useMemo(() => {
+    return generateChartsTheme('perses', muiTheme, ECHARTS_THEME_OVERRIDES);
+  }, [muiTheme]);
+
+  // TODO: Loading indicator
```
Nexucis, line 53: ah yeah that would be so cool to have a `loader`.

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,89 @@
...
+  if (dashboard === undefined) {
+    return null;
+  }
+
+  return (
+    <Box
+      sx={{
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100vh',
```
Nexucis, line 63: here the height shouldn't depend on the screen size ?

ui/core/src/model/time.ts
```diff
@@ -31,12 +32,26 @@ export interface RelativeTimeRange {
...
  */
-export function isRelativeValue(timeRange: TimeRangeValue): timeRange is RelativeTimeRange {
+export function isRelativeTimeRange(timeRange: TimeRangeValue): timeRange is RelativeTimeRange {
   return (timeRange as RelativeTimeRange).pastDuration !== undefined;
 }
 
+/**
+ * Determine whether a given time range is in Grafana format
+ */
+export function isGrafanaRelativeTimeRange(from: string, to: string): boolean {
```
Nexucis, line 44: `isGrafana` ? :D. Maybe another name would be more appropriate don't you think ?

ui/core/src/model/time.ts
```diff
@@ -104,3 +119,16 @@ export function getSuggestedStepMs(timeRange: AbsoluteTimeRange, width: number)
...
 }
+
+/**
+ * Gets the default time range taking into account URL params
+ */
+export function getDefaultTimeRange(from: string, to: string, dashboard?: DashboardResource): TimeRangeValue {
+  const dashboardDuration = dashboard?.spec.duration ?? '1h';
+  const parsedParam = from !== null ? from.split('-')[1] : dashboardDuration;
+  const pastDuration = parsedParam && isDurationString(parsedParam) ? parsedParam : dashboardDuration;
+  if (from === '' || to === '') return { pastDuration };
```
Nexucis, line 130: That's strange the linter doesn't yell at this syntax.
It's usually a nice practice to put `return ...` in the next line. It will be easier to read

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+import userEvent from '@testing-library/user-event';
+import { TimeRangeValue } from '@perses-dev/core';
+import { TimeRangeStateProvider } from '@perses-dev/plugin-system';
+import { screen } from '@testing-library/react';
+import { renderWithContext } from '../../test';
+import { TimeRangeControls } from './TimeRangeControls';
+
+describe('TimeRangeControls', () => {
```
Nexucis, line 21: cool to see new unit test here !

Nexucis:
otherall that's a cool PR @sjcobb.
On top on my others comments and questions, I didn't see anything related to the time zone. Do we assume to display everything in UTC ? I know many of people would love to see that (i.e. not able to change the time zone and that everything is in UTC)

ui/app/src/views/Docs.tsx
sjcobb: Yeah that’s right, I guess for actual docs we should start using GitHub Pages

ui/app/src/App.tsx
sjcobb: Maybe we can discuss this again at the next Perses meeting, this would need to happen when we connect the BE and FE right? I really just wanted react-router in this PR since [useSearchParams](https://reactrouter.com/en/6.4.0/hooks/use-search-params) made things easy

ui/app/src/views/HomeDashboard.tsx
sjcobb: So `vh` is specific to viewport height, this would be equal to 100% of the viewport (really I just left these lines how they were before btw)

ui/core/src/model/time.ts
sjcobb: Are we trying to avoid using 'Grafana' in our codebase? For me, the `from: 'now-<duration>', to: 'now'` structure is kind of specific to Grafana and I needed some way to distinguish it from the isRelativeTimeRange util

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { initialValue, children, onTimeRangeChange } = props;
+
+  const [timeRange, setActiveTimeRange] = useState<TimeRangeValue>(initialValue);
```
LukeTillman, line 30: Back to my earlier comment, the whole state flow here still feels very confusing to me. As I'm looking (and I could be wrong, so apologies in advance), it looks like we have:

1. A `useState` out in the view with a time range (maybe only used as an initial time range, but not sure?)
2. Two `useState` calls in this provider
3. The values in the query string from react router out in the view

To me, it feels like "what is the currently selected time range" is all just one piece of state? There might be other versions we can derive from that (e.g. always converting the currently selected time range to absolute if it's relative), but still feels like there's only one piece of state. I just wonder if there's a way to simplify that flow so we really only have one "source of truth" here.

By "source of truth" I basically mean, only one `useState` or one `useQueryParams` or whatever it is that is responsible for getting/setting the value. That being said, I could also be totally misunderstanding what's going on or what the various pieces of state are for. 😂

ui/plugin-system/src/runtime/time-range.ts
```diff
@@ -12,11 +12,13 @@
...
+  initialValue: TimeRangeValue;
```
LukeTillman, line 20: Is `initialValue` here just for resetting/undo of some kind? If so, would it be better to just expose a `reset` API rather than expose this?

ui/plugin-system/src/runtime/time-range.ts
LukeTillman, line 19: Do you think there's a use case for needing access to both of these? Or do consumers only care about the "current value" in absolute terms?

LukeTillman (approved):
Shoo, what a gargantuan effort. Kudos for taking on introducing routing for the first time to Perses! 🎉  I left a bunch of comments/questions inline on things that were either confusing to me, or I think might be able to be improved. The state flow in particular (and having it potentially duplicated in multiple places) is really the biggest thing I think we should try and iron out. Let me know if you want to chat more about any of this.

Nice work!

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
saminzadeh: +1 to this. I think managing one piece is state is cleaner then deriving other state via useMemo

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -11,26 +11,50 @@
...
+  const fromParam = searchParams.get('from') ?? '';
+  const toParam = searchParams.get('to') ?? '';
+  const defaultTimeRange = getDefaultTimeRange(fromParam, toParam, dashboardResource);
+  const [activeTimeRange, setActiveTimeRange] = useState<TimeRangeValue>(defaultTimeRange);
 
-  return <DashboardView dashboardResource={dashboard} />;
+  const handleonTimeRangeChange = (event: TimeRangeValue) => {
+    if (isRelativeTimeRange(event)) {
+      searchParams.set('from', `now-${event.pastDuration}`);
+      searchParams.set('to', 'now');
```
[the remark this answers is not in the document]
saminzadeh: Do you mean `from=-2h&to=now` instead of `?from=now-2h&to=now`?

ui/app/src/App.tsx
Nexucis: yeah that's why I said "ideally", but definitely that would have to happen when we will connect BE and FE. 
We can talk about it Monday no issue.

It can be done later for sure and move on following the changes you wanted to perform

ui/app/src/views/ViewDashboard.tsx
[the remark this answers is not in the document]
sjcobb: So @chucknthem had the strongest opinion about this awhile back, I'll try to find that feedback again and summarize 

I could go either way, changing to `start=-2h&end=now` is another option

ui/plugin-system/src/runtime/time-range.ts
sjcobb: yeah I was thinking we might need it later for undo / restoring to original state, but a reset API makes more sense so just removed it

ui/app/src/views/ViewDashboard.tsx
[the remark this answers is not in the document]
LukeTillman: I was thinking you could just totally omit `end` if it's always inferred to be "now". 🤷‍♂️  Then `from` is just a duration string for relative time ranges.

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,89 @@
...
+        sx={{
+          flexGrow: 1,
+          overflow: 'hidden',
+        }}
+      >
+        <ChartsThemeProvider themeName="perses" chartsTheme={chartsTheme}>
+          <PluginRegistry getInstalledPlugins={getInstalledPlugins} importPluginModule={importPluginModule}>
+            <PluginBoundary loadingFallback="Loading..." ErrorFallbackComponent={ErrorAlert}>
+              <DataSourceRegistry>
+                <ViewDashboard dashboardResource={dashboard} />
```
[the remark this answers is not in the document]
sjcobb: So I separated them since I wanted to ensure dashboardResource was defined since spec.duration is needed when determining the defaultTimeRange for TimeRangeStateProvider. That part will change though when we remove the useSampleData approach and connect BE & FE.

I'm [still confused ](https://github.com/perses/perses/pull/530#discussion_r971240326)with how things are named in ui/dashboards though, didn't we have a view rendering another view before my PR? Might need a call to talk this through 😄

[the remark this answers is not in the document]
sjcobb: Maybe something like [this commit ](https://github.com/perses/perses/commit/94ba4885a02423906de500d10577e3fe1aa337d7#diff-c8d994442c02c8a3708b3011ec892bc3b5514bf9ba2150e965b179d0490c7a37) could work, ended up renaming DashboardApp in ui/dashboards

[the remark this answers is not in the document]
eunicorn: i think it's confusing too, we have too many "views" 😆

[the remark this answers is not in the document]
LukeTillman: Oh, I think I'm starting to understand. So I would create a folder for the overall DashboardView (e.g. `DashboardView` or whatever) and then the main view component is the export from that folder (probably in an `index.ts` file). Then if you need sub-components here at the view level, that's totally cool, you can put them in the folder next to the main view. Same pattern we use for components that have sub-components in separate files basically. We do this for view-specific stuff in our internal codebase too.

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,89 @@
...
+const DEFAULT_DASHBOARD_ID = 'node-exporter-full';
+
+function HomeDashboard() {
+  const { getInstalledPlugins, importPluginModule } = useBundledPlugins();
+
+  const [searchParams] = useSearchParams();
+
+  // TODO: completely remove old dashboard url param approach
+  const dashboardParam = searchParams.get('dashboard');
+  const dashboard = useSampleData<DashboardResource>(dashboardParam || DEFAULT_DASHBOARD_ID);
```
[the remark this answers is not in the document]
sjcobb: Refactored what was in App.tsx so hopefully this makes more sense now

ui/app/src/views/HomeDashboard.tsx
```diff
@@ -0,0 +1,89 @@
...
+        component="main"
+        sx={{
+          flexGrow: 1,
+          overflow: 'hidden',
+        }}
+      >
+        <ChartsThemeProvider themeName="perses" chartsTheme={chartsTheme}>
+          <PluginRegistry getInstalledPlugins={getInstalledPlugins} importPluginModule={importPluginModule}>
+            <PluginBoundary loadingFallback="Loading..." ErrorFallbackComponent={ErrorAlert}>
+              <DataSourceRegistry>
```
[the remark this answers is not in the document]
sjcobb: Yeah this was bad idea, kind of rushed this part when I realized setting up routing properly was going to take too long--moved it back to App.tsx!

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
sjcobb: Changed this back so now there isn't timeRange and absoluteTimeRange, I think a lot of the confusion is due to what I [described here](https://github.com/perses/perses/pull/530/#discussion_r972118025) about not wanting to make React Router v6 a required dependency in @perses-dev/dashboards, @perses-dev/plugin-system, etc. Ideally we would just be able to set the query params directly from TimeRangeControls / other components. 

Dang, maybe I should just go ahead and change to the QueryStringProvider approach we talked about before 🤔

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -11,26 +11,50 @@
...
+      const endUnixMs = getUnixTime(event.end) * 1000;
+      searchParams.set('from', startUnixMs.toString());
+      searchParams.set('to', endUnixMs.toString());
+      setSearchParams(searchParams);
+      setActiveTimeRange({ start: event.start, end: event.end });
+    }
+  };
+
+  return (
+    <TimeRangeStateProvider initialValue={activeTimeRange} onTimeRangeChange={handleonTimeRangeChange}>
```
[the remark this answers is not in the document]
sjcobb: Cleaned this up and just passing defaultTimeRange in directly now!

ui/core/src/model/time.ts
Nexucis: well from my side, it was weird to the `Grafana` named here. It's really a typed :).

if `to` is always equal to `now` in that method, why not rename it with something like `TimeFromPastToNow` (not super english though)

ui/app/src/views/ViewDashboard.tsx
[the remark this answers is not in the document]
Nexucis: haha that's what I just suggested in another comment for the method `GrafanaRelativeTimeRange` 🤣

ui/core/src/model/time.ts
sjcobb: That makes sense, this util actually isn't needed anymore based on the other thread so ended up removing it completely!

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
sjcobb: PR is now updated with the QueryStringProvider approach--useQueryParams is the single source of truth unless an app does not provide setQueryParams, then setTimeRange is used as a fallback.

ui/app/src/views/ViewDashboard.tsx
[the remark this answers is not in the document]
sjcobb: Updated this to what we discussed, relative time ranges use `start=1h` and no more redundant 'end'

ui/app/src/views/HomeDashboard.tsx
sjcobb: Ended up changing this back to how it was before since the app component reorg isn't necessary since I switched to the QueryStringProvider approach (TimeRangeStateProvider is back inside dashboards package)

sjcobb:
> I didn't see anything related to the time zone. Do we assume to display everything in UTC

@Nexucis yup that's right, UTC is used in the Perses app currently, planning to think more about time zones later. At that point, we'll add another prop to TimeRangeProvider so apps can implement their own solutions for how they handle user settings

ui/app/src/views/HomeDashboard.tsx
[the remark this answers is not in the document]
sjcobb: See comment [here](https://github.com/perses/perses/pull/530#discussion_r975450689), I ended up not needing to do the reorg as part of this PR, but I think we should look into simplifying those view components later.

PR #531, opened 2022-09-13
Use external file for plugin module resources, add metadata for individual plugins

LukeTillman:
Up until now, we've kept some very basic metadata about plugin modules (i.e. packages) in the `package.json` for each package under the `perses` key. This PR does three things to that:

1. Removes that metadata from `package.json` and into its own `plugin.json` file in each package (which will be distributed along with the package when it's published).
2. Changes the format of plugin module metadata so that individual plugins inside the module can have their own metadata. This is so that we can (eventually) show that metadata in the UI to the user (e.g. when selecting a panel type, we should be able to display a list of friendly names for each panel plugin).
3. Updates the `PluginRegistry` to work with the new format.

At some point in the near future, I'll be adding some APIs to the plugin system to allow our code to "introspect" or query that metadata to, for example, get the metadata for all plugins of a certain plugin type.

### DIFF
--- a/ui/app/src/model/bundled-plugins.ts
+++ b/ui/app/src/model/bundled-plugins.ts
@@ -12,9 +12,9 @@
 // limitations under the License.
 
 // Eagerly load the metadata for the bundled plugins, but lazy-load the plugins
-import prometheusPackage from '@perses-dev/prometheus-plugin/package.json';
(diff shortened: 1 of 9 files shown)


sjcobb (approved):
Nice! I like this, definitely more intuitive than having it in each individual package.json

PR #532, opened 2022-09-14
Create the skeleton of the Markdown panel

cndonovan:
Adds a component, but it's doesn't do anything yet:

Screen Shot 2022-09-14 at 7 53 43 AM

### DIFF
--- a/ui/app/sample-data/benchmarkDashboard.ts
+++ b/ui/app/sample-data/benchmarkDashboard.ts
@@ -71,6 +71,16 @@ const benchmarkDashboard: DashboardResource = {
       } as AnyVariableDefinition,
     },
     panels: {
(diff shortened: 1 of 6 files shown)


LukeTillman (approved):
Looks like a great start 👍

sjcobb (approved):
LGTM! As you iterate on this, you'll also need to add markdown to our cuelang schema: https://github.com/perses/perses/tree/main/schemas/panels

We should go over this on a call and I can describe that workflow in more detail and our future plans for how it'll be used to generate types / keep the BE and FE in sync

eunicorn (approved):

Nexucis:
ah it has been already merged :(.
It would have been nice to have the cuelang schema added in the same PR

PR #533, opened 2022-09-14
Use React 18 for development and allow React 17 backward compatability

crob611:
This updates the main package.json and the app package.json to use React 18.  

All the other packages are updated to have a peer dependency on React 17 OR React 18.  

This should allow development against 18, while also allowing an external app using React 17 to use Perses and satisfy all peer dependencies 🤞 

Had to upgrade the package `use-resize-observer` to version 9 to support React 18. It's still backwards compatible with 17, but no changes should be an issue in this case [Change Log](https://github.com/ZeeCoder/use-resize-observer/blob/master/CHANGELOG.md#900-2022-05-15)

Also including a patch to get the app running with React 17 again in case anyone wants to test it against the older version and verify it still works.  

[react-17.patch.zip](https://github.com/perses/perses/files/9567707/react-17.patch.zip)


Nexucis:
That's pretty cool indeed. But I'm wondering how much it will be hard to support the two versions in long term.
Does the plan is to support the last two versions of React ? Like once react v19 is out, we will support react 18 and react 19 and drop react 17 ?

crob611:
I don't know what the long term plan for supporting both should be. React's Component API has been fairly stable for almost 4 years now (since the release of Hooks) and there are plenty of libraries whose React dependency is just > 16.8.  

This at least in the short term will allow apps that consume Perses packages to upgrade themselves to React 18.  Having the peer dependency on just React 17 could prevent that otherwise.

eunicorn (approved):
thanks for doing this!

Nexucis (approved):
> there are plenty of libraries whose React dependency is just > 16.8.

Interesting, thanks for sharing. I'm wondering if we could do the same. Perhaps when React 19 is out we could redo the same PR and see what's happening.

Anyway just thinking about the implications, and I don't have any bad feeling against that. So all good for me :)

LukeTillman (approved):
Awesome, thanks a ton @crob611 for tackling this.

PR #534, opened 2022-09-14
Simplify plugin module format, use metadata to drive plugin loading

LukeTillman:
My ultimate goal is to simplify the plugin loading code a bunch, but the first step in doing that is to simplify the module format used in plugins. Before this PR, the format for plugin modules was to export a `setup` function, that then used a `register` callback in that function to register plugins. There were a few downsides to that:

1. It was more complicated than it needed to be. There was nothing stopping someone from registering plugins that weren't in the metadata.
2. We were duplicating information in those register calls (like the plugin type and kind that a plugin handles) that was already in the plugin metadata.
3. We weren't doing any sort of validation that the plugin metadata was correct (i.e. all the plugins the metadata said were in the module, were actually in the module we imported).

This updates the plugin module format so that plugins are now just named exports that correspond to the `kind` they handle. The import process for a module is now driven by the plugin metadata, only looking for named exports from that module's metadata. I think this is a lot simpler and should allow for some further simplification of the plugin loading code.

One other side-effect of this change is it forced me to treat plugin modules being imported as `unknown`, which is actually probably more representative of what's happening at runtime (i.e. we have no idea what we're going to get inside the dynamically imported JS module). I think in the future we could add some runtime validation for the various plugin types to try and make sure that the imported code is at least on the surface, what we expect for that kind of plugin.

### DIFF
--- a/ui/app/src/model/bundled-plugins.ts
+++ b/ui/app/src/model/bundled-plugins.ts
@@ -14,11 +14,11 @@
 // Eagerly load the metadata for the bundled plugins, but lazy-load the plugins
 import prometheusResource from '@perses-dev/prometheus-plugin/plugin.json';
 import panelsResource from '@perses-dev/panels-plugin/plugin.json';
-import { PluginRegistryProps, PluginModuleResource, PluginModule } from '@perses-dev/plugin-system';
+import { PluginRegistryProps, PluginModuleResource } from '@perses-dev/plugin-system';
 import { useCallback } from 'react';
 
 interface BundledPlugin {
-  importPluginModule: () => Promise<PluginModule>;
+  importPluginModule: () => Promise<unknown>;
 }
 
 /**
--- a/ui/panels-plugin/plugin.json
+++ b/ui/panels-plugin/plugin.json
@@ -36,6 +36,14 @@
           "name": "Empty Chart",
           "description": ""
         }
+      },
+      {
+        "pluginType": "Panel",
+        "kind": "Markdown",
+        "display": {
+          "name": "Markdown",
+          "description": ""
+        }
       }
     ]
   }
--- a/ui/plugin-system/src/components/PluginRegistry/registry-state.ts
+++ b/ui/plugin-system/src/components/PluginRegistry/registry-state.ts
@@ -15,7 +15,7 @@ import { useCallback, useMemo, useRef } from 'react';
 import { useImmer } from 'use-immer';
 import { JsonObject } from '@perses-dev/core';
 import {
-  PluginRegistrationConfig,
+  Plugin,
   PluginModuleResource,
   PluginType,
   ALL_PLUGIN_TYPES,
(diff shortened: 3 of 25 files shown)


LukeTillman:
I'll make sure to update the Markdown panel skeleton's format as well once #532 is merged before I merge this.

cc @cndonovan

ui/app/src/model/bundled-plugins.ts
```diff
@@ -14,11 +14,11 @@
 // Eagerly load the metadata for the bundled plugins, but lazy-load the plugins
 import prometheusResource from '@perses-dev/prometheus-plugin/plugin.json';
 import panelsResource from '@perses-dev/panels-plugin/plugin.json';
-import { PluginRegistryProps, PluginModuleResource, PluginModule } from '@perses-dev/plugin-system';
+import { PluginRegistryProps, PluginModuleResource } from '@perses-dev/plugin-system';
 import { useCallback } from 'react';
 
 interface BundledPlugin {
-  importPluginModule: () => Promise<PluginModule>;
+  importPluginModule: () => Promise<unknown>;
```
sjcobb, line 21: Sounds a little scary, “we’re entering the unknown, import if you dare!!”

ui/panels-plugin/plugin.json
```diff
@@ -36,6 +36,14 @@
           "name": "Empty Chart",
           "description": ""
         }
+      },
+      {
+        "pluginType": "Panel",
+        "kind": "Markdown",
+        "display": {
```
sjcobb, line 43: Are we also going to need ‘icon’ eventually (to show above the display name during add/edit flow)?

ui/plugin-system/src/components/PluginRegistry/registry-state.ts
```diff
@@ -77,49 +75,53 @@ export function useRegistryState(installedPlugins?: PluginModuleResource[]) {
...
 
-      // Call the setup function and remember it's been registered
-      pluginModule.setup(registerPlugin);
+      // Treat plugin module as JS module with named exports that are each a Plugin
+      const plugins = pluginModule as Record<string, Plugin<JsonObject>>;
+
+      setPlugins((draft) => {
+        // Look for all the plugins specified in the metadata
+        for (const pluginMetadata of resource.spec.plugins) {
+          // Assume that plugins will be exported under the same named export as the kind they handle
```
sjcobb, line 93: This reminds me, I was thinking earlier this week about if it would make sense to make `kind: 'LineChart'` have an alias of `kind: 'TimeSeries'` somehow. That way we could gradually make that switch instead of introducing the breaking change right away. 

It seems like maybe that wouldn't be very straightforward based on these lines?

ui/panels-plugin/plugin.json
LukeTillman: Yeah, I think that's probably likely. Not exactly sure how that's going to work quite yet though since images are another thing (like we had with external CSS style sheets for react grid layout) that are handled by webpack loaders.

ui/plugin-system/src/components/PluginRegistry/registry-state.ts
LukeTillman: Good question. I think it would still be pretty easy to do. You'd basically just add to the module exports from `panels-plugin`:

```
export { LineChart as TimeSeries }
```

And then make sure to include an entry in the `plugin.json` for it.

ui/panels-plugin/plugin.json
sjcobb: Wonder if we could get by with inline SVG if the icons are simple enough, but yeah something to figure out later--I see they're gone in the most recent mockups anyways and it's just a dropdown

sjcobb (approved):
You weren't kidding about the breaking changes 😂 but this is great, definitely simplifies things!

PR #535, opened 2022-09-14
Add Template Variable Support to Perses

saminzadeh:
The following implements the core changes related to this [design doc](https://docs.google.com/document/d/1JaU_to9xikifQktm5-16scFy2TfW0cFD7KGQtIbJ_Qw/edit).

- Defines the template variables definitions for List and Text variables, along with the schema changes
- Adds support for Text variables
- Sets up plugin structure for variables
- Removes all existing Template variable prototype code

Examples

**TextVariable**
```json
{
    "kind" : "TextVariable",
    "name": "foo",
    "options": {
        "value": "bar"
    }
}
```

**ListVariable**
```json
{
    "kind" : "ListVariable",
    "name": "foo",
    "options": {
        "kind": "MyCustomListVariable",
        "options": {
            "value": [1,2]
        }
    }
}
```

TODO/Coming:
- Resolve List variable options via plugin
- Build Prom variable plugins


Nexucis:
I think from the datasource designed, we were more or less agree to replace `options` by `spec` to be consistent with the rest of the objects defined

Nexucis:
based on the panel definition, like the [gauge](https://github.com/perses/perses/blob/main/schemas/panels/gauge/gauge.cue), we called `datasource` the field used to select a datasource to be used for the panel.

I think it should be the same for the variable.

I would see it at the root of the variable to be consistent with the Panel. Like that:

```json
{
    "kind" : "ListVariable",
    "name": "foo",
    "datasource": {}
    "options": {
        "kind": "MyCustomListVariable",
        "options": {
            "value": [1,2]
        }
    }
}
```

That should be then the same selector definition used for the panel.
For the variables, probably it will be optional and it will condition the value for the field `kind` you have at the next level.

saminzadeh:
> based on the panel definition, like the [gauge](https://github.com/perses/perses/blob/main/schemas/panels/gauge/gauge.cue), we called `datasource` the field used to select a datasource to be used for the panel.
> 
> I think it should be the same for the variable.
> 
> I would see it at the root of the variable to be consistent with the Panel. Like that:
> 
> ```json
> {
>     "kind" : "ListVariable",
>     "name": "foo",
>     "datasource": {}
>     "options": {
>         "kind": "MyCustomListVariable",
>         "options": {
>             "value": [1,2]
>         }
>     }
> }
> ```
> 
> That should be then the same selector definition used for the panel. For the variables, probably it will be optional and it will condition the value for the field `kind` you have at the next level.


I see, that makes sense - at least from a consistency perspective with Panels that are interested in datasources. In my current implementation, the plugin only receives the `variables.options` or `ListVariableOptions` payload so I'll have to modify that to receive the full variable definition.  Otherwise, it would make sense for the `datasource` property to be in the options payload of a ListVariable

```json
{
    "kind" : "ListVariable",
    "name": "foo",
    "options": {
        "datasource": {}
        "kind": "MyCustomListVariable",
        "options": {
            "value": [1,2]
        }
    }
}
```

saminzadeh:
> I think from the datasource designed, we were more or less agree to replace `options` by `spec` to be consistent with the rest of the objects defined

Cool - that sounds good. @LukeTillman do you recommend I do that in this PR or should I continue to extend the `Definition` type and we can update them all together?
https://github.com/perses/perses/blob/78ab4c0f0490ecaff9b3ca6a205563e3529c9a36/ui/core/src/model/definitions.ts#L29

ui/dashboards/src/templating/components.tsx
```diff
@@ -0,0 +1,164 @@
+// Copyright 2022 The Perses Authors
```
LukeTillman, line 1: Nit: would be great to break these out into the appropriate folders once you think you have the logic down (e.g. `components`, `model`, etc.) Same with the `core` file and other stuff in this `templating` folder.

ui/dashboards/src/templating/core.tsx
```diff
@@ -0,0 +1,269 @@
...
+        if (!varState) {
+          return;
+        }
+        vars[name] = varState;
+      });
+      return vars;
+    },
+    (left, right) => {
+      return JSON.stringify(left) === JSON.stringify(right);
+    }
```
LukeTillman, line 66: That's interesting. I assume this is some kind of compare function? What's it used for? (Just for my own curiosity.)

ui/dashboards/src/templating/core.tsx
```diff
@@ -0,0 +1,269 @@
...
+  );
+  return state;
+}
+
+export function useTemplateVariable(name: string) {
+  const store = useTemplateVariableSrvContext();
+  return useStore(store, (s) => {
```
LukeTillman, line 73: Curious to get your thoughts on Zustand at some point now that you've built all this on top of it.

ui/dashboards/src/templating/core.tsx
```diff
@@ -0,0 +1,269 @@
...
+function createTemplateVariableSrvStore({ initialVariableDefinitions = [] }: TemplateVariableSrvArgs) {
+  const store = createStore<TemplateVariableSrv>()(
+    devtools(
+      immer((set, get) => ({
+        variableState: hydrateTemplateVariableStates(initialVariableDefinitions),
+        variableDefinitions: initialVariableDefinitions,
+
+        // Actions
+        loadTemplateVariable: async (name: VariableName) => {
+          const def = get().variableDefinitions.find((v) => v.name === name) as ListVariableDefinition;
```
LukeTillman, line 121: Why the cast?

ui/plugin-system/src/runtime/template-variables.ts
```diff
@@ -12,28 +12,36 @@
...
+    return state;
+  }
+
+  const values: VariablesState = {};
+  names.forEach((name) => {
+    const s = state[name];
+    if (s) {
+      values[name] = s;
+    }
+  });
```
LukeTillman, line 44: Do you think there's any value in memoizing this? (I always wonder when I see a loop like this to produce a value.)

ui/prometheus-plugin/src/model/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Returns a list of template variables
+ */
+export const parseTemplateVariables = (text: string) => {
+  const regex = TEMPLATE_VARIABLE_REGEX;
+  let m;
+  const matches = [];
+  // find all matches in regex and return them
+  do {
```
LukeTillman, line 52: I love seeing a `do...while` loop in front-end code. Always makes me a little nervous though that we're gonna forget something and end up with an infinite loop. 😂

LukeTillman (commented):
Overall I think this looks fantastic. I want to think about the type definitions/format a little more since those still feel a little funky to me, but awesome job. Left a few comments/questions inline.

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,54 @@
...
+  error?: Error | null;
+};
+
+export type VariableOption = { label: string; value: string };
+
+export type VariablesState = Record<VariableName, VariableState>;
+
+export type VariableValue = string | string[] | null;
+
+export interface IVariable<TKind extends string, TOptions extends JsonObject = JsonObject>
```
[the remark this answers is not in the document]
saminzadeh: Yeah, let me know if you brainstorm something cleaner. It made sense to me since the Variable type expects a `kind`

ui/dashboards/src/templating/components.tsx
saminzadeh: Ack - planning on doing that in the non-draft PR. I was trying to limit the merge conflicts while prototyping. That being said, at some point do you think it will make sense to take a [feature specific file approach](https://reactjs.org/docs/faq-structure.html#grouping-by-features-or-routes) as we grow so files are closer to the modules that need them? I've always struggled with this lol 🤷

ui/dashboards/src/templating/core.tsx
saminzadeh: I don't have too strong of an opinion. I didn't really feel like it made things too much easier over simple hooks and providers + immer. It lacks strong opinions on selectors, immutable state, etc. I actually think I prefer a Provider context model over global state for "lib" use cases over "app" use cases. Probably need to give it a few more goes to get a better sense. Curious to hear everyones thoughts on it with they give it a go

ui/dashboards/src/templating/core.tsx
saminzadeh: loadingTemplateVariables is only required for ListVariable since TextVariables don't have options. I actually may need to change this logic when I implement dynamic plugin resolvers.

ui/prometheus-plugin/src/model/utils.ts
saminzadeh: haha i have to admit this was copy and paste from the regex exporter lol. Happy to clean it up to more "modern" JS

ui/plugin-system/src/runtime/template-variables.ts
saminzadeh: Yeah definitely, I'm thinking of adding in all the memoization optimizations in a follow to get the flow of logic correct first.

ui/dashboards/src/templating/core.tsx
saminzadeh: This was kind of a left over artifact of testing zustand selectors and how they worked. Basically, return the same state if the template variables that are selected didn't change. I'll probably clean this up

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,54 @@
...
+export type TextVariableDefintion = IVariable<'TextVariable', TextVariableOptions>;
 
-export type SingleSelectOptions = {
-  default_value: string;
+export type ListVariableOptions<TKind extends string, TOptions extends JsonObject> = {
+  allowMultiple?: boolean;
+  allowAllValue?: boolean;
+  customAllValue?: string;
+  kind: TKind;
+  options: TOptions;
```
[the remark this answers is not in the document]
saminzadeh: Yeah I'm open to suggestion on what definition is cleaner (and more consistent) here. 

So instead of this:
```json
{
  "kind": "ListVariable",
  "name": "MyVariable",
  "options": {
    "allowMultiple": true,
    "kind": "PromQLLabelValues",
    "options": {
      "label": "job"
    }
  }
}
```

Do this?
```json

{
  "kind": "ListVariable",
  "name": "MyVariable",
  "options": {
    "allowMultiple": true,
    "optionsLoader": {
      "kind": "PromQLLabelValues",
      "options": {
        "label": "job"
      }
    }
  }
}
```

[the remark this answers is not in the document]
LukeTillman: Yeah, basically that. It just felt weird to me to have `options.kind` and `options.options` 😂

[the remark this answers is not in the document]
Nexucis: well actually I would rather prefer to go with the first solution since it is consistent with the Panels, the datasource and basically the whole rest of object definition.

`kind` always defined the type of `options` => aka `spec` ;)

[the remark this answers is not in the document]
Nexucis: beside the first solution is flatten than the second one. So less painful to write if you write the JSON by yourself ;)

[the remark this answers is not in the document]
Nexucis: Anything common to the ListVariable should go on the first `options`. The second level is conditioned by the `kind`.
Since `kind` is common to the `ListVariable` definition, it should go in the first `options` level

ui/prometheus-plugin/src/model/utils.ts
Nexucis: That would be great to modernise it. My stressed is reaching an high level when seing that :D

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,55 @@
...
+export type VariableState = {
+  value: VariableValue;
+  options?: VariableOption[];
+  loading: boolean;
+  error?: Error;
+};
+
+export type VariableOption = { label: string; value: string };
+
+export type VariablesState = Record<VariableName, VariableState>;
```
cndonovan, line 29: the plural `s` in `VariablesState` is so subtle i feel like i could miss it, but trying to think of a good alternative too...

throwing it out there, `VariableStates`?

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,55 @@
...
 
-export interface VariableDisplayOptions extends JsonObject {
-  hide?: boolean;
-  label: string;
+export interface TextVariableOptions extends JsonObject {
+  value: string;
 }
 
-export type VariableSelectionOptions = SingleSelectOptions | MultiSelectOptions;
+export type TextVariableDefintion = Variable<'TextVariable', TextVariableOptions>;
```
cndonovan, line 48: typo TextVariableDefintion (missing i)

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const allowMultiple = definition?.options.allowMultiple === true;
+  useEffect(() => {
+    loadTemplateVariable(name);
+  }, [name, loadTemplateVariable]);
+
+  let value = ctx.state?.value;
+
+  if (allowMultiple && !Array.isArray(value)) {
+    value = [];
+  }
```
cndonovan, line 50: it'd be cool if we could set this default somehow in types, but not sure that's possible

ui/dashboards/src/context/TemplateVariableProvider.tsx
```diff
@@ -0,0 +1,268 @@
...
+  VariablesState,
+  VariableState,
+  VariableName,
+  VariableValue,
+  ListVariableDefinition,
+  VariableDefinition,
+  DEFAULT_ALL_VALUE as ALL_VALUE,
+} from '@perses-dev/core';
+
+type TemplateVariableSrv = {
```
cndonovan, line 30: what does `srv` stand for?

ui/app/sample-data/convert/convert-variables.ts
```diff
@@ -11,97 +11,107 @@
...
+    // if (grafanaVariable.name === '') continue;
 
-    // TODO: This is just a static list of options, so we should support
-    if (grafanaVariable.type === 'custom') continue;
+    // // Just punt on these for now
+    // if (grafanaVariable.type === 'datasource') continue;
 
-    const { name, query, label, hide } = grafanaVariable;
+    // // TODO: This is just a static list of options, so we should support
+    // if (grafanaVariable.type === 'custom') continue;
```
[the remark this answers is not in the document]
saminzadeh: I don't have a preference, my thought was to comment it out so when I need to implement it, the reference is already there. But if it feels dirty, I can remove. There will be follow up PRs after this core one that will handle the convert logic too.

ui/core/src/model/dashboard.ts
```diff
@@ -32,7 +32,7 @@ export interface DashboardResource {
 export interface DashboardSpec {
   datasource: DatasourceSelector;
   duration: DurationString;
-  variables: Record<string, VariableDefinition>;
+  variables: VariableDefinitions;
```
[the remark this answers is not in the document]
saminzadeh: That's a good point, I'll change that.

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,54 @@
...
-/**
- * Variable definition options that are common to all variables.
- */
-export interface VariableDefinition<Options extends JsonObject = JsonObject> extends Definition<Options> {
-  display: VariableDisplayOptions;
-  selection: VariableSelectionOptions;
-  capturing_regexp?: string;
+export const DEFAULT_ALL_VALUE = '$__all' as const;
+
+export type VariableName = string;
```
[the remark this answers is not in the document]
saminzadeh: Yeah exactly, I'm not sure the exact way of doing it. But I have seen some projects do with types like Duration. But it might be too much and a string is simple enough

ui/core/src/model/variables.ts
saminzadeh: Ah yeah I felt the same way. That could work, or maybe even `VariableStateMap` ?

ui/core/src/model/variables.ts
saminzadeh: haha nice catch!

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,55 @@
...
+
+export interface Variable<Kind extends string, TOptions extends JsonObject = JsonObject> extends Definition<TOptions> {
+  kind: Kind;
+  name: VariableName;
+  display?: {
+    label?: string;
+    hidden?: boolean;
+  };
+  defaultValue?: VariableValue;
+  options: TOptions;
```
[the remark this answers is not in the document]
saminzadeh: I was trying to not confuse the VariableOption with this generic type. I meant to drop the `T` prefix but i'll update that.

ui/dashboards/src/components/Variables/Variable.tsx
saminzadeh: Yeah good call, I think we can add a default. I'll probably do that in the next PR the implements List variable loading.

ui/dashboards/src/context/TemplateVariableProvider.tsx
saminzadeh: Lol yeah that's not super clear, probably need a better name. It means `service` which is used in angular (aka Grafana). Is `TemplateVariableState` or like `TemplateVariableStateManager` better?

ui/app/sample-data/convert/convert-variables.ts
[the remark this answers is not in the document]
cndonovan: gotcha gotcha, that seems fine

ui/core/src/model/variables.ts
cndonovan: i like that too, `SomethingMap` was coming up when i was running through names

ui/dashboards/src/context/TemplateVariableProvider.tsx
cndonovan: Both sound good to me!

ui/core/src/model/variables.ts
```diff
@@ -13,29 +13,54 @@
...
+export type TextVariableDefinition = Variable<'TextVariable', TextVariableOptions>;
 
-export type SingleSelectOptions = {
-  default_value: string;
+export type ListVariableOptions<Kind extends string, Options extends JsonObject> = {
+  allowMultiple?: boolean;
+  allowAllValue?: boolean;
+  customAllValue?: string;
+  optionsLoader: {
+    kind: Kind;
```
LukeTillman, line 55: Think `kind` here can just be `string` (you don't need the extra generic param) since this is always a plugin kind that the framework code doesn't know anything about.

ui/dashboards/src/context/TemplateVariableProvider.tsx
```diff
@@ -0,0 +1,267 @@
...
+    };
+  });
+}
+
+export function useTemplateVariableDefinitions() {
+  const store = useTemplateVariableStore();
+  return useStore(store, (s) => s.variableDefinitions);
+}
+
+export function useTemplateVariableSrv() {
```
LukeTillman, line 97: Nit: Saw you renamed to remove the `Srv` part of the name (I'm having Angular flashbacks), might want to clean up a few more usages of `Srv` in here to reflect the new name?

LukeTillman (approved):
Just a couple minor nits, but awesome work. Definitely something we can iterate on. Let me know when you're ready to merge and I'll knock it out.

ui/dashboards/src/context/TemplateVariableProvider.tsx
saminzadeh: Good catch!

PR #537, opened 2022-09-15
add onDoubleClick callback to LineChart component for future use

sjcobb:
Follow up to PR #530 that add support for: 
- add ability to override default double click behavior to LineChart.tsx component
  - leaving undefined for now, while we discuss ideal unzoom feature
- ~~add undo when LineChart is double clicked~~
- ~~keep track of previousTimeRange in case plugins need it again~~

### DIFF
--- a/ui/plugin-system/src/context/TimeRangeStateProvider.tsx
+++ b/ui/plugin-system/src/context/TimeRangeStateProvider.tsx
@@ -29,8 +29,6 @@ export function TimeRangeStateProvider(props: TimeRangeProviderProps) {
 
   const [timeRange, setActiveTimeRange] = useState<TimeRangeValue>(initialValue);
 
-  const [previousTimeRange, setPreviousTimeRange] = useState<TimeRangeValue | null>(null);
-
   const defaultTimeRange: AbsoluteTimeRange = isRelativeTimeRange(initialValue)
     ? toAbsoluteTimeRange(initialValue)
     : initialValue;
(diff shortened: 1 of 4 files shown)


ui/plugin-system/src/context/TimeRangeStateProvider.tsx
```diff
@@ -29,13 +29,16 @@ export function TimeRangeStateProvider(props: TimeRangeProviderProps) {
 
   const [timeRange, setActiveTimeRange] = useState<TimeRangeValue>(initialValue);
 
+  const [previousTimeRange, setPreviousTimeRange] = useState<TimeRangeValue | null>(null);
```
saminzadeh, line 32: IMO, this `previousTimeRange` doesn't belong in the plugin runtime sdk. "Undo" is a feature of the LineChart, therefore it should manage the previous state based on subscribing to timeRange state changes. You can argue that we should manage previous state values for everything to support "undo" use-cases otherwise.

Nexucis:
is it really intuitive to double click and it undo something ? Just asking as a UX noob design.

I would expect it is zooming when I'm double clicking.

ui/plugin-system/src/context/TimeRangeStateProvider.tsx
sjcobb: Yeah that's true, I had it closer to that at one point but then starting overthinking it and it's a little unclear to me what goes in the plugin runtime sdk and what doesn't. For now, I'm going to just leave the optional onDoubleClick callback undefined and come back to unzoom later since maybe it shouldn't be on double click

sjcobb:
> is it really intuitive to double click and it undo something

Going to leave this out for now so we can compare options and think about undo more holistically. 

Merging with just the optional onDoubleClick callback in case someone wants to override LineChart's default behavior

PR #538, opened 2022-09-15
Updated React Grid Layout placeholder to use the theme color and fixed no kebab case error

eunicorn:
**Bug**
`'-webkit-user-select': 'none',`
Using kebab-case for css properties in objects is not supported. 

**Enhancement**
before the placeholder color used to be red, now it'll use the theme primary color.

overlay


sjcobb (approved):

LukeTillman (approved):
Thanks a ton for fixing this! The console noise was bugging me way more than it probably should have. 😂

LukeTillman:
Side note: using the theme color was a nice touch. Looks great.

PR #539, opened 2022-09-15
Move plugin registry into legacy folder to prep for new implementation

LukeTillman:
This just moves the existing `PluginRegistry` into a `legacy` folder in preparation for another PR where I'm going to do a new implementation to make plugin loading much simpler. Just a few simple file renames, so isn't a breaking change of any kind since we're still publicly exporting the legacy version. I think doing this will make the diffs easier to read/review for the future PRs so I'm doing this first.


sjcobb (approved):
👍 👍

eunicorn (approved):

Nexucis:
Hehe I can get what is coming: A lot of breaking changes in coming 🤣 

👍 waiting to see what you have in mind to simplify the load.

PR #540, opened 2022-09-15
Refactored DashboardProvider

eunicorn:
**DashboardProvider Improvements**
- useZustandContext is deprecated, so I refactored DashboardProvider to [use the recommended approach](https://github.com/pmndrs/zustand#react-context).
- moved layouts states and app states into their own slices so it's more clean and modular. 
- used zustand/immer to eliminate using `produce` in setters


ui/dashboards/src/components/AddPanel/AddPanel.tsx
```diff
@@ -28,18 +28,16 @@ import { Drawer, ErrorAlert } from '@perses-dev/components';
 import { JsonObject } from '@perses-dev/core';
 import { PluginBoundary } from '@perses-dev/plugin-system';
 import { ChangeEvent, FormEvent, useState } from 'react';
-import { useLayouts, usePanels } from '../../context';
+import { useDashboardApp, useLayouts, usePanels } from '../../context';
 import { removeWhiteSpacesAndSpecialCharacters } from '../../utils/functions';
 import { PanelOptionsEditor, PanelOptionsEditorProps } from './PanelOptionsEditor';
 
-interface AddPanelProps {
```
eunicorn, line 35: Removed these props and storing these states in zustand instead because there are a few places in the app where you can open the add panel drawer

ui/dashboards/src/context/DashboardAppSlice.tsx
```diff
@@ -0,0 +1,45 @@
...
+    setIsOpen: (isOpen: boolean, index?: number) => void;
+    index?: number;
+  };
+}
+
+export const createDashboardAppSlice = immer<DashboardAppSlice>((set) => ({
+  addPanelComponent: {
+    isOpen: false,
+    groupIndex: 0,
+    setIsOpen: (isOpen: boolean, groupIndex = 0, panelRef?: string) =>
```
eunicorn, line 22: Since the add panel component is used for adding new panel and editing existing one, we need know when to open Add Panel vs Edit Panel.

if `panelRef` is undefined, it means we're adding a new panel

ui/dashboards/src/context/DashboardAppSlice.tsx
```diff
@@ -0,0 +1,45 @@
...
+      set((state) => {
+        state.addPanelComponent.isOpen = isOpen;
+        state.addPanelComponent.groupIndex = groupIndex;
+        state.addPanelComponent.panelRef = panelRef;
+      }),
+  },
+  addGroupComponent: {
+    isOpen: false,
+    isEdit: false,
+    setIsOpen: (isOpen: boolean, index?: number) =>
```
eunicorn, line 32: similar to add panel component, we also need to know when the add group component is used for adding a new panel group vs editing an existing panel group

ui/dashboards/src/context/LayoutsSlice.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export interface LayoutsSlice {
+  updateLayout: (layout: LayoutDefinition, index?: number) => void;
+  addItemToLayout: (index: number, item: GridItemDefinition) => void;
+}
+
+export const createLayoutsSlice: StateCreator<LayoutsSlice> = (set) => ({
+  updateLayout: (layout: LayoutDefinition, index?: number) =>
+    set(
+      produce((state) => {
```
saminzadeh, line 27: You can also use the immer-middleware so you don't have to do this in each set:
https://github.com/pmndrs/zustand#immer-middleware

eunicorn: i need access to `layouts` in this slice, but i'm not sure how i can pass `layouts` or 'props` using the immer middleware since i need to initialize `layouts` from props

any idea?

LukeTillman (approved):
Thanks for talking through this with me on Zoom, I appreciate it. I left one minor suggestion inline, but LGTM.

ui/dashboards/src/context/LayoutsSlice.tsx
saminzadeh: I think you can use `get()`? 
https://github.com/pmndrs/zustand#read-from-state-in-actions

eunicorn: ok i finally figured it out - i just had to cast the `state` type to `DashboardStoreState`

eunicorn: get() is read only

PR #542, opened 2022-09-16
New Plugin Registry Implementation

LukeTillman:
In the continuing effort to simplify plugin loading, this is the `PluginRegistry` implementation. In this PR we have the new implementation for the registry, as well as the new hooks for consuming it:

- `usePlugin`: this is mostly the same as the current API for using a plugin at runtime, it just has a new implementation that consumes the new PluginRegistry API. It's now using `react-query` to execute the async callback which should allow consumers to do things like disable loading, add callbacks on load, etc.
- `useListPluginMetadata`: this is the new API for querying/introspecting plugin metadata. You pass it a plugin type and it will give you back the metadata for all the plugins of that type (e.g. "give me a list of all the Panel plugins"). We'll be able to use this in visual editing to allow users to select things like the type of panel they want to add to a dashboard.

One thing to note is that the props accepted by the new `PluginRegistry` implementation are the same as the old implementation, so when I make the switch in a follow-up PR, this shouldn't break consumers. I purposefully did not switch to the new implementation in this PR because:

1. It's already pretty big
2. When I make the switch, it's going to get rid of the things like the `PluginLoadingBoundary` and allow consumers to choose to display their own loading indicators while a plugin is loading.

I'll make the switch to this implementation in another PR shortly.


ui/plugin-system/src/runtime/plugins.ts
```diff
@@ -0,0 +1,49 @@
...
+  UseQueryOptions<PluginImplementation<T, JsonObject>, unknown, PluginImplementation<T, JsonObject>, string>,
+  'queryKey' | 'queryFn'
+>;
+
+/**
+ * Loads a plugin and returns the plugin implementation, along with loading/error state.
+ */
+export function usePlugin<T extends PluginType>(pluginType: T, kind: string, options?: UsePluginOptions<T>) {
+  const { getPlugin } = usePluginRegistry();
+  const queryKey = `${pluginType}:${kind}`;
```
eunicorn, line 30: can you use`getTypeAndKindKey` here?

ui/plugin-system/src/components/PluginRegistry/plugin-indexes.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Returns an async callback for getting indexes of the installed plugin data.
+ */
+export function usePluginIndexes(getInstalledPlugins: GetInstalledPlugins) {
+  // Creates indexes from the installed plugins data (does useEvent because this accesses the getInstalledPlugins prop
+  // and we want a stable reference for the callback below)
+  const createPluginIndexes = useEvent(async (): Promise<PluginIndexes> => {
+    const installedPlugins = await getInstalledPlugins();
```
eunicorn, line 34: do we need to catch any errors here?

ui/plugin-system/src/components/PluginRegistry/plugin-indexes.ts
```diff
@@ -0,0 +1,89 @@
...
+  // De-dupe creating plugin indexes (i.e. requests to get installed plugins)
+  const pluginIndexesCache = useRef<Promise<PluginIndexes> | undefined>(undefined);
+  const getPluginIndexes = useCallback(() => {
+    let request = pluginIndexesCache.current;
+    if (request === undefined) {
+      request = createPluginIndexes();
+      pluginIndexesCache.current = request;
+
+      // Remove failed requests from the cache so they can potentially be retried
+      request.catch(() => pluginIndexesCache.current === undefined);
```
eunicorn, line 76: oh we catch it here?

eunicorn (approved):

ui/plugin-system/src/runtime/plugins.ts
LukeTillman: Yeah, I totally could. Good catch!

ui/plugin-system/src/components/PluginRegistry/plugin-indexes.ts
LukeTillman: Good question: I decided to let errors just bubble up to the call site for now (i.e. wherever `usePlugin` or `useListMetadata` are called). But totally possible we want to report this once via some "global" mechanism and then let it bubble up. We may need to think about having some plugin runtime error reporting context that folks could implement when embedding (i.e. "use a snackbar" or soemthing)...

ui/plugin-system/src/components/PluginRegistry/plugin-indexes.ts
LukeTillman: Yeah, this will only delete the failed Promise from the cache. The error still bubbles up to `useQuery` (which will then report it on its `error` prop).

ui/plugin-system/src/components/PluginRegistry/test-plugins/bert/index.ts
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { PanelPlugin } from '../../../../model';
+
+// Dummy plugins to test loading
```
sjcobb, line 16: Love the new tests!

bert-ernie

ui/plugin-system/src/runtime/plugins.ts
```diff
@@ -0,0 +1,49 @@
...
+// Allow consumers to pass useQuery options from react-query when listing metadata
+type UseListPluginMetadataOptions = Omit<
+  UseQueryOptions<PluginMetadata[], unknown, PluginMetadata[], string>,
+  'queryKey' | 'queryFn'
+>;
+
+/**
+ * Gets a list of plugin metadata for the specified plugin type and returns it, along with loading/error state.
+ */
+export function useListPluginMetadata(pluginType: PluginType, options?: UseListPluginMetadataOptions) {
```
sjcobb, line 44: Nice, this looks really straightforward to use

sjcobb (approved):
🚀 🚀  
Can I get a summary of your last PRs for the changelog when I go to publish 0.8? Doesn't look like this PR has breaking changes, but interested in which to categorize as [BREAKINGCHANGE] so I don't forget any

LukeTillman:
@sjcobb Definitely, will ping you directly with a list/summary.

PR #543, opened 2022-09-17
upgrade to go 1.19

Nexucis:


Nexucis:
I will just go ahead on this, so I can update the PR #497.
If someone is not agree with this change, we can talk and revert if necessary

PR #544, opened 2022-09-17
test build parallel

Nexucis:


Nexucis:
just checking if the parallel build is working with go 1.19. I might consider to remove goreleaser if it doesn't. 10min to build just 3 different binaries is so painful

PR #547, opened 2022-09-20
Add static variables plugin for ListVariables

saminzadeh:
Video demo: https://www.loom.com/share/a94abddac40249f18e7f5db8c02bb6ae

- Changes the Variable plugin to use a promise-based resolver instead of React hooks
- Support for loading List variable options via plugins
- Adds a StaticListVariable which supports the following:

```ts
// Example just providing values
const v = {
  name: "instance",
  kind: "ListVariable",
  options: {
    optionsLoader: {
      kind: "StaticListVariable",
      options: {
        values: ["demo.do.prometheus.io:3000", "demo.do.prometheus.io:9100"],
      },
    },
  },
};

// Example by providing a label and value
const v = {
  name: "instance",
  kind: "ListVariable",
  options: {
    optionsLoader: {
      kind: "StaticListVariable",
      options: {
        values: [{ label: "My Label", value: "value"}, "another-value"],
      },
    },
  },
};
 ```

TODO:

- [ ] Add tests once interfaces and definitions are stable
- [ ] Move StaticListVariable to a better package, currently its in the Prometheus one 

### CODE REVIEW

File: ui/plugin-system/src/model/variables.ts [hunk 1]
```diff
@@ -18,29 +18,30 @@ import { usePlugin } from '../components/PluginLoadingBoundary';
...
+  const plugin = usePlugin('Variable', definition.options.optionsLoader.kind);
   if (plugin === undefined) {
     // Provide default values while the plugin is being loaded
-    return { data: [], loading: true };
+    return;
   }
-  return plugin.useVariableOptions(definition);
+
+  return {
+    getVariableOptions: () => {
```
saminzadeh, line 43: This is probably best to be memoized but its a bit funky with the conditional hooks. I'll address t his optimization in a follow up PR

ui/plugin-system/src/model/variables.ts
```diff
@@ -18,29 +18,30 @@ import { usePlugin } from '../components/PluginLoadingBoundary';
...
+  getVariableOptions: GetVariableOptions<Options>;
 }
 
 /**
  * Plugin hook responsible for getting the options of a custom variable
  * definition.
  */
-export type UseVariableOptionsHook<Options extends JsonObject = JsonObject> = (
+export type GetVariableOptions<Options extends JsonObject = JsonObject> = (
   definition: ListVariableDefinition<Options>
```
saminzadeh, line 29: I'm wondering if this should receive the full ListVariable definition or just the Options for the variable kind?

ui/plugin-system/src/model/variables.ts
```diff
@@ -18,29 +18,30 @@ import { usePlugin } from '../components/PluginLoadingBoundary';
...
-};
+) => Promise<{ data: VariableOption[] }>;
 
 /**
  * Use the variable options from a variable plugin at runtime.
  */
-export const useVariableOptions: VariablePlugin['useVariableOptions'] = (definition) => {
-  const plugin = usePlugin('Variable', definition.kind);
+export const useVariablePlugin = (definition: ListVariableDefinition) => {
+  const plugin = usePlugin('Variable', definition.options.optionsLoader.kind);
```
saminzadeh, line 36: Should this plugin be renamed to `ListVariable` or is `Variable` still ok?

ui/app/sample-data/demoDashboard.ts
```diff
@@ -35,16 +35,26 @@ const demoDashboard: DashboardResource = {
       },
       {
         name: 'instance',
-        kind: 'TextVariable',
+        kind: 'ListVariable',
         options: {
-          value: 'demo.do.prometheus.io:9100',
+          optionsLoader: {
+            kind: 'StaticListVariable',
```
saminzadeh, line 41: How do we feel about the `StaticListVariable` name?

ui/plugin-system/src/model/variables.ts
```diff
@@ -18,29 +18,30 @@ import { usePlugin } from '../components/PluginLoadingBoundary';
...
-  useVariableOptions: UseVariableOptionsHook<Options>;
+  getVariableOptions: GetVariableOptions<Options>;
 }
 
 /**
  * Plugin hook responsible for getting the options of a custom variable
  * definition.
  */
-export type UseVariableOptionsHook<Options extends JsonObject = JsonObject> = (
+export type GetVariableOptions<Options extends JsonObject = JsonObject> = (
```
saminzadeh, line 28: Also, I didn't pass any context yet since I didn't need it for the plugins yet. But I imagine, i'll inject the current template variable states (needed for chaining variables) and time range (so template variables that use time can be recomputed)

ui/plugin-system/src/model/variables.ts
```diff
@@ -18,29 +18,30 @@ import { usePlugin } from '../components/PluginLoadingBoundary';
  * Plugin for handling custom VariableDefinitions.
  */
 export interface VariablePlugin<Options extends JsonObject = JsonObject> {
-  useVariableOptions: UseVariableOptionsHook<Options>;
+  getVariableOptions: GetVariableOptions<Options>;
```
sjcobb, line 21: nit, maybe a comment here could be useful on why this is promise-based instead of a hook or it can wait as long as we document that decision later when we add actual plugin docs.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -37,42 +37,88 @@ export function TemplateVariable({ name }: TemplateVariableProps) {
...
+    }
+    setVariableLoading(name, true);
+    try {
+      const { data } = await variablePlugin.getVariableOptions();
+      setVariableOptions(name, data);
+    } catch (e) {
+      console.error('Failed to load template variable options', e);
+    }
+    setVariableLoading(name, false);
+  }, [definition]);
```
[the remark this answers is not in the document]
sjcobb: Oh whoops nvm, I see it's a warning, not an error is why build didn't fail

sjcobb (approved):
🚀

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -37,42 +37,88 @@ export function TemplateVariable({ name }: TemplateVariableProps) {
...
+      console.error('Failed to load template variable options', e);
+    }
+    setVariableLoading(name, false);
+  }, [variablePlugin, definition, name, setVariableLoading, setVariableOptions]);
+
   useEffect(() => {
-    loadTemplateVariable(name);
-  }, [name, loadTemplateVariable]);
+    loadOptions();
+  }, [loadOptions]);
```
LukeTillman, line 62: I wonder if there's some way to (eventually) move this logic up into the store? Or maybe make use of `useQueries` from `react-query` to run multiple options loading callbacks and manage the loading/data state? Anyway, just thinking out loud, nothing to do here.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -37,42 +37,88 @@ export function TemplateVariable({ name }: TemplateVariableProps) {
...
+            // Must be selected
+            if (e.target.value === null || e.target.value.length === 0) {
+              return;
+            }
+            setVariableValue(name, e.target.value as VariableValue);
           }}
           multiple={allowMultiple}
         >
-          {ctx.state?.options?.map((option) => (
+          {loading && (
```
LukeTillman, line 109: Does this `Select` need to become an `Autocomplete` eventually to support multiple selections? Was thinking about it while looking at loading state because I think that control also has a built-in loading state/indicator.

LukeTillman (approved):
Looks great. Thanks for fixing up with the conflicts from the big API change.

ui/dashboards/src/components/Variables/Variable.tsx
saminzadeh: Yeah agreed, I tried to do this at first but it felt kind of awkward since I need to resolve the plugin through a hook . I was hoping that I could load it purely through async promises. That way you could load the options without mounting any components.

```ts
async loadVariable(name: string) {
   const def = getVariableDefinitionByName(name)
   const plugin = await loadPlugin(def.kind) 
}
```

The other idea was to load the plugin via a hook then set getVariableOptions function in the store. 
```tsx
store.setVariableLoader('myVar', plugin.getVariableOptions)

// Then later you can call it 
store.loadVariable('myVar')

```

ui/dashboards/src/components/Variables/Variable.tsx
saminzadeh: Yeah I think I might need to change this to Autocomplete. I don't think the Select allows you to "create options" on the fly, which is probably needed too. I'll put out a follow up PR if thats the case.

ui/dashboards/src/components/Variables/Variable.tsx
LukeTillman: Yeah, the new `PluginRegistry` implementation I added should let you do that (part of the reason I redid the API was this use case). It's not in there quite yet, but once it is, you could probably use the context object directly:

https://github.com/perses/perses/blob/ec485059dbb767993ad071077c10c8fc3a32ce6f/ui/plugin-system/src/components/PluginRegistry/plugin-registry-model.ts#L17-L20

PR #549, opened 2022-09-21
Be consistent with kind and spec in our API

LukeTillman:
As discussed on Matrix, this updates all our API definitions in the front-end to be consistent with their usage of `kind` and `spec`. The high level changes were basically:

1. Switch plugins to use a `spec` field instead of an `options` field.
2. Add `kind: 'Panel'` to all panel definitions. Move common options under `spec` and plugin specifics under `spec.plugin`.
3. Add `kind: 'GraphQuery'` to all graph query definitions. Move common options under `spec` (there aren't any right now) and plugin specifics under `spec.plugin`.
4. Just use `unknown` rather than `JsonObject` to represent plugin spec types at runtime when we don't know what they are in the framework code.

While the number of lines changed here looks like a lot, the large majority of the changes were really to demo/test dashboard data.


Nexucis:
Looks pretty good for me.

We will have to update the Cuelang schemas to be aligned (cc @AntoineThebaud), but otherwise that's nice.

Nexucis (approved):
I'm wondering if `JSONObject` has still a purpose, but otherwise looks good to me

Nexucis:
nice it's so beautiful with `plugin` 🤩

Nexucis (approved):
let's merge it if you don't have anything else to change :)

ui/dashboards/src/test/testDashboard.ts
```diff
@@ -28,108 +28,163 @@ const testDashboard: DashboardResource = {
     duration: '24h',
     variables: [
```
Nexucis, line 29: Just seeing that, sorry for the late.

Wouldn't be easier to put the variables as a map like the panels. So it's easier to reference it.
That would be consistent also.
@saminzadeh @LukeTillman @sjcobb

Nexucis:
I don't see the datasource selector also :o. Does it disappear ?

ui/dashboards/src/test/testDashboard.ts
cndonovan: If we did / do ever go with second approach, maybe we'd want something like:
```
variables: { ... },
variableLayout: [...],
panels: { ... },
panelLayouts: [ ... ],
```

Basically, prefixing `layouts` so that one is for panels and one is for variables

cndonovan (approved):

LukeTillman:
> I don't see the datasource selector also :o. Does it disappear ?

👍  Good point. Like we talked about in chat, hoping to fix this soon when I implement the new Datasources spec we agreed on a few weeks ago.

ui/dashboards/src/test/testDashboard.ts
LukeTillman: Let's definitely follow-up on this if you all think it merits more discussion. We can definitely do it in a follow-up PR if we feel like we need to.

Nexucis: maybe managing a layout for the variable might be painful for the users. I'm quite lazy, so the less I have to do to configure my dashboard, the better I would be 🤣

PR #552, opened 2022-09-21
Swap sed and npm version calls when bumping version

LukeTillman:
This just swaps the order of the `sed` and `npm version` calls when bumping the UI package versions so that when `npm version` runs, the dependencies are already updated and don't try to resolve package dependencies against the old verison. (This was causing peer dependency conflicts in the latest attempt at a release because we upgraded the React peer dependency.)


Nexucis (approved):

LukeTillman:
Thanks @Nexucis !

PR #553, opened 2022-09-21
Markdown panel cleanup: Styling, memoization, and cuelang schemas

cndonovan:
This PR:
* Add styles for panel elements (specifically, `<code>` and `<table>`, which looked pretty bare-bone without styles)
* Memoize `marked.parse()` and `DOMPurify.sanitize()` (i.e. don't rerun these on re-render unless `text` actually changed)
* Add cuelang schemas
Screen Shot 2022-09-21 at 9 19 26 AM


Nexucis:
is it possible to see a screenshot showing how it looks like with your PR ?

Nexucis (approved):
LGTM

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -11,17 +11,58 @@
...
+import { Box, Theme } from '@mui/material';
 import { PanelProps } from '@perses-dev/plugin-system';
 import { MarkdownPanelOptions } from './markdown-panel-model';
 
 export type MarkdownPanelProps = PanelProps<MarkdownPanelOptions>;
 
-// Overall, this panel does:
-// 1. text --> html (with support for original markdown and GFM)
-// 2. sanitize the html to prevent XSS attacks
+function createMarkdownPanelStyles(theme: Theme) {
```
sjcobb, line 23: Oh cool, I've had to do styles this way a few times to override lib styles, I should have remembered to suggest it too!

schemas/panels/markdown/markdown.cue
```diff
@@ -14,9 +14,8 @@
 package text
 
 #panel: {
-	kind: "TextPanel"
+	kind: "Markdown"
```
sjcobb, line 17: 🚀 thanks for updating these

sjcobb (approved):
LGTM!

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
cndonovan: Yeah, I think my mental blocker was getting the insight that I can apply styling in all the ways people normally do with `mui`, it doesn't need to be different because I'm writing code that's part of library that will be used by other projects. 

(And realizing that that assumption works because we're requiring that people install `mui` to use our code.)

schemas/panels/markdown/markdown.cue
cndonovan: welcome! #cuelang!

LukeTillman:
Nice work, thanks @cndonovan!

PR #554, opened 2022-09-21
Remove old datasources, shim enough to keep UI functioning

LukeTillman:
This just removes the old Datasources code, type definitions, etc. so that we can implement the new spec documented in #404. I left a very basic implementation in place for the legacy version just to let the UI continue functioning while we work on implementing it for real.


Nexucis:
Nice :).
Maybe for this one, let's try to connect FE and BE directly. WDYT ?
Hopefully next week we should be able to deliver the new implem of thé datasource BE side

LukeTillman:
> Nice :). Maybe for this one, let's try to connect FE and BE directly. WDYT ? Hopefully next week we should be able to deliver the new implem of thé datasource BE side

Yeah, definitely possible. My plan is to basically do the implementation with stubbed out API calls so we can just swap it out for a real `fetch` call to connect them.

Nexucis:
> My plan is to basically do the implementation with stubbed out API calls so we can just swap it out for a real fetch call to connect them.

Interesting ! Do you think it would be possible to do the same for the loading of the dashboard ?

Nexucis (approved):

LukeTillman:
> > My plan is to basically do the implementation with stubbed out API calls so we can just swap it out for a real fetch call to connect them.
> 
> Interesting ! Do you think it would be possible to do the same for the loading of the dashboard ?

Yeah, probably could do the same there. One thing I'd love to do when we want to make that change is to add a proper URL/route to the UI for a dashboard so we could grab the "dashboard ID" from the URL path.

Nexucis:
> One thing I'd love to do when we want to make that change is to add a proper URL/route to the UI for a dashboard so we could grab the "dashboard ID" from the URL path.

Totally inline with that.
From my side a proposition for the URL would be `/projects/<project_name>/dashboards/<dashboard_name>`.
With `project_name` and `dashboard_name` you are able to find/get the dashboard.

PR #555, opened 2022-09-21
Make it possible to edit markdown

cndonovan:
This is a start at making it possible to edit the markdown panel: 

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
@@ -54,6 +54,7 @@ const PanelDrawer = () => {
   // TODO: Get rid of that useEffect :)
   const defaultOptions = useMemo(() => panelDefinition?.spec.plugin.spec ?? {}, [panelDefinition]);
 
+  console.log({ defaultGroup, defaultPanelName, defaultPanelDescription, defaultKind, defaultOptions });
   const [group, setGroup] = useState(defaultGroup);
   const [panelName, setPanelName] = useState(defaultPanelName);
   const [panelDescription, setPanelDescription] = useState(defaultPanelDescription);
--- a/ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
@@ -40,18 +40,25 @@ const PanelDrawer = () => {
   const { panels, updatePanel } = usePanels();
   const { panelDrawer, closePanelDrawer } = useDashboardApp();
 
-  let defaultPanelName = '';
-  let defaultDescription = '';
-  if (panelDrawer?.panelKey) {
-    // editing an existing panel
-    defaultPanelName = panels[panelDrawer.panelKey]?.spec.display.name ?? '';
-    defaultDescription = panels[panelDrawer.panelKey]?.spec.display.description ?? '';
-  }
-  const [group, setGroup] = useState(panelDrawer?.groupIndex);
+  const panelKey = panelDrawer?.panelKey;
+  const panelGroupIndex = panelDrawer?.groupIndex;
+  const panelDefinition = panelKey ? panels[panelKey] : undefined;
+
+  const defaultGroup = panelGroupIndex ?? 0;
+  const defaultPanelName = panelDefinition?.spec.display.name ?? '';
+  const defaultPanelDescription = panelDefinition?.spec.display.description ?? '';
+  const defaultKind = panelDefinition?.spec.plugin.kind ?? '';
+  const defaultOptions = panelDefinition?.spec.plugin.spec;
+
+  const [group, setGroup] = useState(defaultGroup);
   const [panelName, setPanelName] = useState(defaultPanelName);
-  const [panelDescription, setPanelDescription] = useState(defaultDescription);
-  const [kind, setKind] = useState('');
-  const [options, setOptions] = useState<unknown>({});
+  const [panelDescription, setPanelDescription] = useState(defaultPanelDescription);
+  const [kind, setKind] = useState(defaultKind);
+  // Plugin options depends on the plugin kind ('LineChart', 'Markdown', et cetera).
+  // As the user switches between kinds, we keep track of the options
+  // so that we can go back to wherever they left off.
+  const [kindToOptions, setKindToOptions] = useState<Record<string, unknown>>({ [defaultKind]: defaultOptions });
+  const options = kindToOptions[kind];
 
   // TODO: What should we show if either of these error out?
   const { data: pluginMetadata } = useListPluginMetadata('Panel');
@@ -59,7 +66,9 @@ const PanelDrawer = () => {
     enabled: kind !== '',
     onSuccess: (plugin) => {
       // Set initial options whenever plugin kind changes and a plugin is loaded
-      setOptions(plugin.createInitialOptions());
+      if (!kindToOptions[kind]) {
+        setKindToOptions({ ...kindToOptions, [kind]: plugin.createInitialOptions() });
+      }
     },
   });
   const OptionsEditorComponent = plugin?.OptionsEditorComponent;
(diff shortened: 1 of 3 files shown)


ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -41,33 +41,44 @@ const PanelDrawer = () => {
...
+    defaultOptions = panelDefinition.spec.plugin.spec;
   }
   const [group, setGroup] = useState(panelDrawer?.groupIndex);
   const [panelName, setPanelName] = useState(defaultPanelName);
-  const [panelDescription, setPanelDescription] = useState(defaultDescription);
-  const [kind, setKind] = useState('');
-  const [options, setOptions] = useState<unknown>({});
+  const [panelDescription, setPanelDescription] = useState(defaultPanelDescription);
+  const [kind, setKind] = useState(defaultKind);
+  const [options, setOptions] = useState<unknown>(defaultOptions);
```
cndonovan, line 62: We'll probably want to rename `option` in light of the change to `spec`, but I'll save that for another day

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -41,33 +41,44 @@ const PanelDrawer = () => {
...
 
   // TO DO: we might want to make the form a sub component we don't need this useEffect
   // currently, we need to reset the states whenever panelDrawer is reopened
   // since this component does not get remounted when it open/closes (otherwise we lose the animation of sliding in/out)
   useEffect(() => {
```
cndonovan, line 67: Might look into following this comment so we don't need this useEffect, it feels strange to set the values twice.

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -164,21 +184,12 @@ const PanelDrawer = () => {
...
-                variant="outlined"
-                onChange={handlePanelDescriptionChange}
-              />
-            </Stack>
+            <TextField
+              label="Panel Description"
+              value={panelDescription}
+              variant="outlined"
+              onChange={handlePanelDescriptionChange}
+            />
```
cndonovan, line 192: Moved these `<input>`s around to match the Chronosphere design more

ui/dashboards/src/components/PanelDrawer/PanelOptionsEditor.tsx
```diff
@@ -23,10 +23,14 @@ export interface PanelOptionsEditorProps {
...
   const { OptionsEditorComponent, createInitialOptions } = usePanelPlugin(kind);
+  const [currentKind, setCurrentKind] = useState(kind);
 
   // When the kind changes, re-init options
   useEffect(() => {
-    onChange(createInitialOptions());
+    if (kind !== currentKind) {
+      onChange(createInitialOptions());
+      setCurrentKind(kind);
+    }
```
cndonovan, line 33: Without this, this would overwrite the existing values (reset them to empty values), because the `useEffect` was called multiple times with "Markdown" as the same value for `kind`.

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -188,10 +199,11 @@ const PanelDrawer = () => {
                 <MenuItem value="LineChart">Line Chart</MenuItem>
                 <MenuItem value="GaugeChart">Gauge Chart</MenuItem>
                 <MenuItem value="StatChart">Stat Chart</MenuItem>
+                <MenuItem value="Markdown">Markdown</MenuItem>
```
cndonovan, line 202: Add Markdown option!

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
Nexucis: mmm that would be cool to do it asap. Like that the model doesn't diverge :).

ui/dashboards/src/components/PanelDrawer/PanelOptionsEditor.tsx
```diff
@@ -23,10 +23,14 @@ export interface PanelOptionsEditorProps {
...
 
     // TODO: See if we can switch up plugin loading so this happens as part of selecting the plugin kind so we don't
     // need this effect at all
```
cndonovan, line 36: I also thought about moving the `createInitialOptions` into the `onPanelKindChange` in the parent (PanelDrawer), as a way to not have this useEffect. Because at a high-level, we want: `kind` changes ==> change the options. 

But figured I'd leave it here since this comment hints that things might change and plus, maybe we want this part of the logic to be part of this generic editor... like, no matter where PanelOptionsEditor is used, it will ensure that the options are reset correctly (instead of relying on the caller)

LukeTillman: I think the changes I've made in #559 should take care of this issue (and your other comment about it above). Do you mind if we merge that first, then you can see what impact that has to these changes?

LukeTillman: Also, feel free to leave comments on that PR if you have a chance to review!

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
LukeTillman: Yeah, agreed. This was something that Eunice and I talked about during code review, so she left a TODO. If you want to try and tackle it, go for it.

ui/dashboards/src/components/PanelDrawer/PanelOptionsEditor.tsx
cndonovan: That sounds great (re: merge yours in first, see where things land / what impact it has here, and maybe I don't need to add this code)

ui/panels-plugin/src/plugins/markdown/MarkdownPanelOptionsEditor.tsx
```diff
@@ -11,13 +11,24 @@
...
 export function MarkdownPanelOptionsEditor(props: MarkdownPanelOptionsEditorProps) {
-  const { value } = props;
-  return <Box>{JSON.stringify(value)}</Box>;
+  const {
+    onChange,
+    value: { text },
+  } = props;
+
+  console.log('in MarkdownPanelOptionsEditor');
+  console.log({ text });
```
cndonovan, line 28: oops need to remove

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -40,26 +40,35 @@ const PanelDrawer = () => {
...
   const [panelName, setPanelName] = useState(defaultPanelName);
-  const [panelDescription, setPanelDescription] = useState(defaultDescription);
-  const [kind, setKind] = useState('');
-  const [options, setOptions] = useState<unknown>({});
+  const [panelDescription, setPanelDescription] = useState(defaultPanelDescription);
+  const [kind, setKind] = useState(defaultKind);
+  // Plugin options depends on the plugin kind ('LineChart', 'Markdown', et cetera).
+  // As the user switches between kinds, we keep track of the options
+  // so that we can go back to wherever they left off.
+  const [kindToOptions, setKindToOptions] = useState<Record<string, unknown>>({ [defaultKind]: defaultOptions });
```
LukeTillman, line 60: Oooh, nice. That's an interesting idea and definitely a better experience than the current one. 👍

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -68,16 +77,12 @@ const PanelDrawer = () => {
...
-    if (panelDrawer?.panelKey) {
-      // display panel name and description in text fields when editing an existing panel
-      setPanelName(panels[panelDrawer.panelKey]?.spec.display.name ?? '');
-      setPanelDescription(panels[panelDrawer.panelKey]?.spec.display.description ?? '');
-    } else {
-      setPanelName('');
-      setPanelDescription('');
-    }
-  }, [panelDrawer, panels]);
+    setGroup(defaultGroup);
```
LukeTillman, line 80: Love that this got simpler. We can definitely still follow-up on trying to get rid of the effect completely.

LukeTillman (approved):
Overall looks great! Love the idea for preserving options. Once you get rid of the `console` statements, I'll go ahead and merge.

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
cndonovan: Yay, glad you like. It was kind of a necessity too, because the createInitialOptions() would overwrite the initial data otherwise 😄 We had a similar thing at my last job and we did the "keep track of current kind and compare when it changes" solution, but I think I like this way more.

sjcobb: +1 really like this idea!

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
sjcobb: Nice! In the future, we'll probably want to move some of these into @perses-dev/components so they're easier to share (this PR looks good, I'm thinking about more complicated controls in the future that diverge from MUI)

sjcobb (approved):
🚀

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
cndonovan: Ya, like start having our own form field components in the components package?

saminzadeh (approved):
This is great- Nice work Christine!

sjcobb:
Just re-ran the failed go job to see if it will pass, also looks like the last two commits need to be rebased with signoff added, l can merge after!

cndonovan:
> Just re-ran the failed go job to see if it will pass, also looks like the last two commits need to be rebased with signoff added, l can merge after!

Ah thank you! I just came back to this to fix the test failure, and now I'll fix that too

PR #559, opened 2022-09-22
Integrate new PluginRegistry, remove legacy code

LukeTillman:
This switches all `PluginRegistry` usages over to the new implementation created in #542 and removes all the legacy code. Now that all plugins with hooks have been switched to Promise-based APIs (see #556 and #547), the tricks we were doing with the `PluginLoadingBoundary` to avoid violating the rules of hooks are no longer necessary. I replaced those boundaries with regular `ErrorBoundary` components since I think we still want to catch errors in those places.


ui/dashboards/src/components/Panel/PanelContent.tsx
```diff
@@ -11,16 +11,34 @@
...
- * definition's kind. Used so that a PluginLoadingBoundary can be wrapped around this for fallback UI while
- * the plugin is loading.
+ * definition's kind. Used so that an ErrorBoundary can be wrapped around this.
  */
 export function PanelContent(props: PanelContentProps) {
-  const { PanelComponent } = usePanelPlugin(props.definition.spec.plugin.kind);
+  const {
+    definition: {
+      spec: {
+        plugin: { kind: pluginKind },
```
cndonovan, line 27: Wouldn't be opposed to calling this `panelKind` or even `panelPluginKind` to distinguish from the overall plugin kind ('Panel')

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -184,17 +194,20 @@ const PanelDrawer = () => {
...
             </FormControl>
           </Grid>
           <Grid item xs={8}>
-            <PluginBoundary loadingFallback="Loading..." ErrorFallbackComponent={ErrorAlert}>
-              {kind !== '' && <PanelOptionsEditor kind={kind} value={options} onChange={handleOptionsChange} />}
-            </PluginBoundary>
+            <ErrorBoundary FallbackComponent={ErrorAlert}>
+              {OptionsEditorComponent !== undefined && (
+                <OptionsEditorComponent value={options} onChange={handleOptionsChange} />
+              )}
```
cndonovan, line 209: Part of me wants to keep PanelOptionsEditor because I'm imagining a dev who is developing their own UI for editing panels... (i.e. they wouldn't just reach for PanelDrawer, they would build it out and PanelOptionsEditor would be a piece of it)

But, maybe we do just want them to reach for PanelDrawer, and we can assume that that's their point of entry into the editing experience. 

Which is making me realize I want to rename it something like `PanelEditor` or `EditPanelDrawer` (something to make it clear that it's the key to editing panels)

cndonovan (approved):

ui/app/src/App.tsx
```diff
@@ -57,18 +63,20 @@ function App() {
...
+          <ChartsThemeProvider themeName="perses" chartsTheme={chartsTheme}>
+            <PluginRegistry getInstalledPlugins={getInstalledPlugins} importPluginModule={importPluginModule}>
               <LegacyDataSourceRegistry>
                 <QueryStringProvider queryString={searchParams} setQueryString={setSearchParams}>
-                  {/* temp fix to ensure dashboard refreshes when URL changes since setQueryString not reloading as expected  */}
-                  <ViewDashboard key={location.key} />
+                  <ErrorBoundary FallbackComponent={ErrorAlert}>
+                    {/* temp fix to ensure dashboard refreshes when URL changes since setQueryString not reloading as expected  */}
+                    <ViewDashboard key={location.key} />
+                  </ErrorBoundary>
```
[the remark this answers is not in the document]
LukeTillman: Yeah, visibly it would be the same. The difference would be whether the providers get unmounted (and thus lose any state in them, etc.).

ui/dashboards/src/components/Panel/PanelContent.tsx
LukeTillman: Sounds good to me!

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
LukeTillman: Yeah, I can see a use case for maybe bringing it back in some way just to have self-contained loading state and UI for when the plugin component is loading. There's nothing stopping someone who's building their own UI from doing the same thing we're doing either to get the component (i.e. `const plugin = usePlugin('Panel', panelPluginKind)`).

sjcobb (approved):
💥

PR #560, opened 2022-09-22
add Array checks after template variable changes

sjcobb:
- Add Array checks before calling forEach and map so existing dashboards don't break
- Prevents `definitions.forEach is not a function` and similar .map error
- see related PR #561

### DIFF
--- a/ui/dashboards/src/components/Variables/VariableList.tsx
+++ b/ui/dashboards/src/components/Variables/VariableList.tsx
@@ -45,14 +45,15 @@ export function TemplateVariableList() {
                 </TableRow>
               </TableHead>
               <TableBody>
-                {variableDefinitions.map((v) => (
-                  <TableRow key={v.name} sx={{ '&:last-child td, &:last-child th': { border: 0 } }}>
-                    <TableCell component="th" scope="row" sx={{ fontWeight: 'bold' }}>
-                      {v.name}
-                    </TableCell>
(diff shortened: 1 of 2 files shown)


ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -45,14 +45,15 @@ export function TemplateVariableList() {
...
               <TableBody>
-                {variableDefinitions.map((v) => (
-                  <TableRow key={v.name} sx={{ '&:last-child td, &:last-child th': { border: 0 } }}>
-                    <TableCell component="th" scope="row" sx={{ fontWeight: 'bold' }}>
-                      {v.name}
-                    </TableCell>
-                    <TableCell align="right">{v.kind}</TableCell>
-                  </TableRow>
-                ))}
+                {Array.isArray(variableDefinitions) &&
```
cndonovan, line 48: U think it's worth leaving a comment to remove these checks after we've migrated dashboards? Will we migrate dashboards??

Trying to think of having less noise in the code, but also the comment itself could become noise

cndonovan (approved):

LukeTillman:
As alternative, could we just wrap part of the component tree in an `ErrorBoundary`? Feels like that's probably what we want in the future anyway so that an error in that part of the page doesn't break everything?

LukeTillman:
Going to close this and use #562 instead

PR #561, opened 2022-09-22
make row display.title and display.collapse optional 

sjcobb:
- Allow panels to still render if spec.display (inside layouts array) is not populated. 
- This PR defaults rows to open with an empty title
- see related PR #560

### DIFF
--- a/ui/dashboards/src/components/GridLayout/GridLayout.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridLayout.tsx
@@ -37,7 +37,15 @@ export function GridLayout(props: GridLayoutProps) {
     ...others
   } = props;
 
-  const [isOpen, setIsOpen] = useState(!!spec.display?.collapse?.open);
+  // allow display to be optional, default to open state
+  const rowDisplay = spec.display ?? {
+    title: '',
+    collapse: {
+      open: true,
+    },
+  };
+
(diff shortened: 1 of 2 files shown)


ui/dashboards/src/components/GridLayout/GridLayout.tsx
```diff
@@ -37,7 +37,15 @@ export function GridLayout(props: GridLayoutProps) {
...
-  const [isOpen, setIsOpen] = useState(!!spec.display?.collapse?.open);
+  // allow display to be optional, default to open state
+  const rowDisplay = spec.display ?? {
+    title: '',
+    collapse: {
+      open: true,
+    },
+  };
+
+  const [isOpen, setIsOpen] = useState(rowDisplay.collapse?.open ?? true);
```
LukeTillman, line 48: Was it just this state initializer (i.e. the `!!`) causing the issue? Feels weird to default the whole `display` property and show a title if there isn't one.

LukeTillman (approved):
One question inline. Thanks for all the work on integration.

cndonovan (approved):

ui/dashboards/src/components/GridLayout/GridLayout.tsx
LukeTillman: (We're gonna want to support groups without titles eventually anyway)

LukeTillman:
Going to close this and fix it in #562

PR #563, opened 2022-09-22
Update styles for markdown panel links

cndonovan:
Showed some screenshots to Mohit, he reminded me to check dark mode, which made me see that the links looked weird.

## Changes
* For links, get the color from `theme`
* For code blocks, the border-radius is fixed instead of coming from `theme.borderRadius` 
  * Even if people are theme-ing their buttons/containers to have a certain border-radius, I don't think this should apply to the code formatting
* For code blocks + tables, the padding depends on font-size instead of `theme.spacing()`
  * theme.spacing() can defined be in terms of `px`, I think the padding for this formatting should be set to always scale with the font-size

## Screenshots
Screen Shot 2022-09-22 at 3 10 28 PM
Screen Shot 2022-09-22 at 3 10 40 PM


sjcobb (approved):
💥 I still forget about dark mode sometimes 😂

LukeTillman (approved):

PR #564, opened 2022-09-23
add a second build command that will build the missing package and then rebuild the app

Nexucis:


Nexucis:
actually it doesn't fix anything strangely. I'm currently changing `goreleaser` with `promu` and it is still failing.

Building `@perses-dev/prometheus-plugin` with `tsc` doesn't solve the issue because running after `npm run build` still fail

And regarding the last proposition in #568 this doesn't make sense anymore

PR #565, opened 2022-09-23
Add prometheus template variable support

saminzadeh:
Video demo: https://www.loom.com/share/7abb86efb73f4753bfb01aad589bf044

This PR adds support for loading ListVariable options through PromQL queries. This first iteration supports 2 types of queries, `LabelNames` (grafana: `label_names()`)  and `LabelValues` (grafana: `label_values(label)` or `label_values(metric, label)` (see [Grafana docs ](https://grafana.com/docs/grafana/latest/datasources/prometheus/#query-variable)for these).

Here is an example of the queries:

```ts
      {
        kind: 'ListVariable',
        spec: {
          name: 'instance',
          plugin: {
            kind: 'PrometheusLabelValuesVariable',
            spec: {
              label_name: 'instance',
              matchers: ['up{job="node"}'],
            },
          },
        },
      },
      {
        kind: 'ListVariable',
        spec: {
          name: 'instance',
          plugin: {
            kind: 'PrometheusLabelNamesVariable',
            spec: {},
          },
        },
      }
```

TODO/Follow Ups:

- [ ]  Add variable chaining / dependencies between definitions (See [grafana](https://grafana.com/docs/grafana/v9.0/variables/variable-types/chained-variables/))
- [ ] Ability to switch from the default datasource


ui/app/sample-data/demoDashboard.ts
```diff
@@ -27,20 +27,32 @@ const demoDashboard: DashboardResource = {
     duration: '30m',
     variables: [
       {
-        kind: 'TextVariable',
+        kind: 'ListVariable',
         spec: {
           name: 'job',
-          value: 'node',
+          plugin: {
```
saminzadeh, line 33: I think the big question is here, should each variable type be its own variable plugin? So instead of:
```ts
{
   kind: "PrometheusVariable",
   spec: {
      datasource: "prom-demo",
      kind: "LabelValues",
      spec: { 
         label: "job"
      }
   }
}
```

It would look something like this:

```ts
{
   kind: "PrometheusLabelValuesVariable",
   spec: {
      datasource: "prom-demo",
      label: "job"
   }
}
```

I have a slight preference for the first approach (basically overloading the Prom Variable plugin) and treating the types of queries as properties in the spec. That way, they can be visually grouped together and the datasource selection is preserved.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -40,6 +40,7 @@ function ListVariable({ name }: TemplateVariableProps) {
   const { data: variablePlugin } = usePlugin('Variable', definition.spec.plugin.kind);
 
   const { setVariableValue, setVariableLoading, setVariableOptions } = useTemplateVariableActions();
+  const datasources = useLegacyDatasources();
```
saminzadeh, line 43: Will this eventually allow us to choose different datasources? It seems like it only supports `defaultDatasource` right now?

Nexucis:
Well instead of taking the same datamodel than Grafana, we could instead review a bit the struct of LabelValues / LabelNames and actually be closer to what the Prometheus API is proposing.

This would be more coherent as a Prometheus user : 

```json
{
  "kind": "PrometheusLabelNamesVariable",
  "spec" : {
       "matchers": [
           "up"
       ],
      "capturing_regexp": ".*"
  }
}
```

```json
{
  "kind": "PrometheusLabelValuesVariable",
  "spec" : {
       "label_name": "instance",
       "matchers": [
           "up"
       ],
      "capturing_regexp": ".*"
  }
}
```

where `matchers` is optional in both case

```json
{
  "kind": "PrometheusQueryVariable",
  "spec" : {
    "expr": "up{instance='localhost:8080'}",
    "label_name": "instance",
    "capturing_regexp": ".*"
  }
}
```

saminzadeh:
> Well instead of taking the same datamodel than Grafana, we could instead review a bit the struct of LabelValues / LabelNames and actually be closer to what the Prometheus API is proposing.
> 
> This would be more coherent as a Prometheus user :
> 
> ```json
> {
>   "kind": "PrometheusLabelNamesVariable",
>   "spec" : {
>        "matchers": [
>            "up"
>        ],
>       "capturing_regexp": ".*"
>   }
> }
> ```
> 
> ```json
> {
>   "kind": "PrometheusLabelValuesVariable",
>   "spec" : {
>        "label_name": "instance",
>        "matchers": [
>            "up"
>        ],
>       "capturing_regexp": ".*"
>   }
> }
> ```
> 
> where `matchers` is optional in both case
> 
> ```json
> {
>   "kind": "PrometheusQueryVariable",
>   "spec" : {
>     "expr": "up{instance='localhost:8080'}",
>     "label_name": "instance",
>     "capturing_regexp": ".*"
>   }
> }
> ```

That makes sense too, I feel like `capturing_regexp` could be a ListVariable feature rather than specific to the Prom variables. That way, you can filter values from any variable / datasource even if their backend or plugin doesn't directly support it.

In this example, are suggesting that these are children of a "PrometheusVariable" kind or they are top level variable kinds?

ui/app/sample-data/demoDashboard.ts
cndonovan: The first one feels best to me too!

sjcobb: Yeah I like the first version as well, seems more consistent with Luke's recent changes too. Would there need to be a `plugin` added too though (similar to [ex here](https://github.com/perses/perses/blob/main/ui/app/sample-data/benchmarkDashboard.ts#L114))?

saminzadeh: I'm not sure if it makes sense to include `plugin` since it's the plugin kind, the `PrometheusVariable` is the actual plugin kind. If that's a bit unclear, we can also just change `kind` to `type` in the sub config to differentiate between an entity definition.

```ts
{
   kind: "PrometheusVariable",
   spec: {
      datasource: "prom-demo",
      type: "LabelValues", // or LabelNames
      query: { 
         label: "job"
      }
   }
}
```

Nexucis:
> That makes sense too, I feel like `capturing_regexp` could be a ListVariable feature rather than specific to the Prom variables.

I didn't think that much, but you are probably right. I'm just wondering if it should be a mandatory field. Or if it is optional it is actually replaced by `.*` so it captured everything.

> In this example, are suggesting that these are children of a "PrometheusVariable" kind or they are top level variable kinds?

Good question. I have the feeling these could be a top level variable kinds to simplify it.
I thought first it would make sense to make it part of a `PrometheusVariable`, but actually you will still have three different definitions (one for each variable) and so three different plugin variable.  So you can't actually have a single variable definition that would contain the three different one.
With that in mind, the `PrometheusVariable` doesn't make sense for me. 

And regarding your arg in https://github.com/perses/perses/pull/565#discussion_r978877823 for the fact it can be visually grouped together, well I assumed if you provide a list of variable definition, you will have to provide the UI as well. So probably you can already grouped them in the UI. Specially since you are coding a dedicated UI for each variables. So if you are in that mode, then you can say "I have three different variable definitions, but actually they are part of the same system, so I have only one UI for the three of them, regardless their def."

So I would see something like this: 

```json
{
    "kind": "ListVariable",
    "spec": {
      "name": "instance",
      "capturing_regexp": ".*"
      "display": {
        "name" : "Incredible Variable",
        "description": "veni vidi vici"
      },
      "plugin": {
        "kind": "PrometheusLabelNamesVariable",
        "spec" : {
          "datasource": {},
          "matchers": [
            "up"
          ]
        }
    }
  }
}
```

ui/app/sample-data/demoDashboard.ts
saminzadeh: We went with the one plugin per query type approach as discussed in the sync.

ui/dashboards/src/components/Variables/Variable.tsx
LukeTillman: Yup. I'm starting to work on "Datasources v2", so we should be able to plug that in here sometime in the next few weeks.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -49,13 +50,13 @@ function ListVariable({ name }: TemplateVariableProps) {
     }
     setVariableLoading(name, true);
     try {
-      const { data } = await variablePlugin.getVariableOptions(definition);
+      const { data } = await variablePlugin.getVariableOptions(definition, { datasources });
```
LukeTillman, line 53: Not important right now, but want to pick your brain about this style of dependency injection a bit more at some point. (This is just a reminder for us to talk about it at some point 😂 )

LukeTillman (approved):
This implementation turned out very straightforward, love it. Nice work.

PR #567, opened 2022-09-24
Try out some new UI build tooling

LukeTillman:
The randomly failing UI builds have been bothering me, so I decided to try out a few alternative front-end build tools to try and make things better. This makes it so we stop using the (very slow) Typescript compiler except for type-checking (i.e. we no longer use it for transpiling code) and remove the project references. Instead, this has been replaced with:

- [Turborepo](https://turborepo.org/): this is basically a task runner for UI monorepos that is aware of your package structure/dependencies and can then run tasks in the correct order. This means, for example, each package can have its own build script and Turborepo will figure out what order to run them in based on `dependencies` in the `package.json` files.
- ~~[tsup](https://tsup.egoist.dev/): this compiles our library packages (i.e. everything but `app`) using `esbuild` underneath to transform the code. It outputs both ESM and CJS builds in one go, meaning we no longer need a separate build step in CI for CJS builds.~~
- ~~[esbuild](https://esbuild.github.io/): this is a really fast Typescript compiler written in Golang. It doesn't do type-checking (or emit type declarations) so we still have to do that separately using the official `tsc`, but the performance gains are pretty great. I've plugged this into our webpack build and like I mentioned above, it also powers `tsup`.~~
- [swc](https://swc.rs/): a really fast Typescript compiler written in Rust. It doesn't do type-checking or emit type declarations, so we still do that separately using the official `tsc`. There's also a loader for webpack, so we're using that for the `app` code.
- [@swc/jest](https://swc.rs/docs/usage/jest): a drop-in replacements for `ts-jest` that uses `swc` to compile our Jest tests.

I did a few totally unscientific tests on my M1 Mac locally just to see if this is actually any faster. 

## Dev Server Times

How long does `npm run start` take before webpack serves the bundle, both for an initial run with no cache, then quitting the initial run and restarting the server with a cache from the first run.

| Description | No Cache | Restart with Cache |
| ----------- | -------- | ------------------ |
| Old         | 18s      | 11s                |
| New         | 16s      | 7s                 |

Overall a modest improvement.

## Build Times

Basically, how long does `npm run build` take. Nothing is cached (just like on the CI server).

| Description | Build for ESM | Build for CJS | Total |
| ----------- | ------------- | ------------- | ----- |
| Old         | 30s           | 12s           | 42s   |
| New         | 18s           | 0s            | 18s   |

Here we can really see the benefits, partly because we can do both ESM and CJS builds in one go now. Hopefully this should help a lot with our CI jobs.

Speed improvements weren't my main goal here, although I suspect the difference will probably become more pronounced the larger the codebase gets (and the more work `tsc` would have to do to compile things). Stability was my main goal and trying to prevent some of the flakiness we've been seeing with builds. I'm hoping this will have a positive impact there.

I'd like for us to try this setup out for awhile and see if it's actually a better experience. If there are issues that crop up and we can't fix them, we can always just revert and go back to `tsc` and project references.

### Future Improvement Ideas

- See if TurboRepo's [remote caching](https://turborepo.org/docs/core-concepts/remote-caching) feature is worth enabling. This would allow us to share a remote cache across both dev machines and CI servers.
- Investigate `vite` or `parcel` as an alternative to using `webpack`. I don't think we'd get a ton of gain from this yet since most of our code is in packages that are compiled separately, but it might be worth checking into eventually. `parcel` is intriguing (although not nearly as popular as `vite`) because it uses `swc` underneath for compilation and looks like it's also capable of building libraries.
~~- See about using something like `@swc/jest` to speed up Jest test compilation. Our test suites are pretty small right now so it's not super urgent, but this could become a bigger issue soon.~~ This is done now.

**Update**: after merging in the changes from #568, I've gone ahead and updated the description here with the final new toolchain.


LukeTillman:
The CI results look promising. Here's a [recent run before the change](https://github.com/perses/perses/actions/runs/3109414385/jobs/5039730011) where it looks like the 2 UI build tasks took about 2 minutes total to finish. It looks like it takes about 1 minute now in this PR.

PR #568, opened 2022-09-24
Try out some new UI build tooling (part 2)

LukeTillman:
This is a follow-on to #567 (I've setup this PR to be targeted at merging into that PR's branch). Here, I wanted to see what it would be like to use `swc` rather than `tsup` and `esbuild` to power all the Typescript transformations, including adding it to Jest. `swc` is a Typescript transpiler built in Rust and so it's also very fast. A few things I like about this setup:

1. Only one compiler toolchain is being used everywhere, including in Jest. With the `esbuild` toolchain, there isn't really a great way to add it to Jest, something we tried at Chronosphere and failed at (ending up with `@swc/jest` instead).
2. The build output from `swc` seems a little bit better / more intuitive than `tsup` because it outputs individual files for each file in our libraries, rather just a single bundle output file.
3. The performance numbers in the other PR seem to be the same, so we still get the same speed gains. (Note: this is particularly true for "cached" runs since Turborepo does task output caching in both of these setups.)

Really the only thing I don't love about this setup is that the `package.json` setup is a little bit more verbose for each package because we have separate tasks (that are all run concurrently) for building ESM, CJS, and Typescript type definitions. But on balance, I think this is the way to go because of the advantages I pointed out above.

## Why did the tests change?

When adding `@swc/jest`, some of our tests started failing. This was because `swc` (and `esbuild` for that matter) both follow the specs for ES Modules which say that the `export` object from an ES Module is not writable. This means that some of our tests which were trying to do `jest.spyOn` to mock the behavior of exports from a module were not able to "rewrite" that export with the mock behavior.

The official Typescript compiler (that we were using in Jest before this change) doesn't currently implement that part of the spec correctly, and so exports were still rewritable in that world. They plan to change that at some point, but the timeline isn't clear right now. Jest's official support for ES Modules will likely also enforce the spec correctly since it's based on the underlying Node.js implementation of ESM (so if Node enforces it, it's likely to also be enforced in Jest).

The "correct" way to fix this is in our test code by following the spec and not trying to mock exports from ES modules. (Note: it _is_ possible in an ESM world to still mock an entire module as evidenced by Jest's current unstable implementation.) I rewrote the tests to not rely on mocking those exports. I _think_ they ended up in a slightly better state as a result where they aren't so coupled to implementation details. I still don't love manipulating the store (since that also feels a bit like an implementation detail), but we can continue to iterate on strategies to make that better.

## Did the tests get faster?

They did! Here's the timing for running all of our test suites in all packages.

| Description | Time |
|-|-|
| Old (`ts-jest` with Typescript compiler) | ~13s |
| New (`@swc/jest` with swc compiler) | ~5s |

Our test suites are _really_ small right now so it's not super obvious, but I think this is going to have a huge impact over time. This is definitely something we've found internally at Chronosphere where we made this same switch on a codebase with 100s of tests and saved minutes.


Nexucis:
my only concern regarding this, is more concerning the long term.
We may face some issues that will be difficult to solve because of lack of support / usage of this non regular toolchain.

I already thought even for Jest and ESM it is a nightmare and both are standard :/. So would it be worse by using a different toolchain with Jest ?

Nexucis:
I'm currently trying as well to reduce the time taken to build the perses application (BE + FE) and it seems it's now impossible to build the FE without a cache mechanism that required to be warm up.

So if this new way to build the FE guarantees build stabilities and that works with a fresh clone with no caching, then I will be ok with that

LukeTillman:
> my only concern regarding this, is more concerning the long term. We may face some issues that will be difficult to solve because of lack of support / usage of this non regular toolchain.
> 
> I already thought even for Jest and ESM it is a nightmare and both are standard :/. So would it be worse by using a different toolchain with Jest ?

Yeah, we basically use this toolchain internally at Chronosphere (minus Turborepo), and it's worked great. It's definitely a risk, but `swc` is pretty popular and supported by some of the "next gen" build tooling companies and projects (e.g. Vercel, Deno, etc.) so I'm not super worried. It would also be pretty straightforward to switch back to the vanilla `tsc` at some point if we needed to. I keep hoping Microsoft will finally listen to the community and rewrite the official compiler in something faster.

Jest's official support for ESM is still experimental. That's largely been because Node's support for it has been experimental too. I look at `@swc/jest` as just another option for how to do Typescript and jest together. There really hasn't been an "official" way, you've always had different options (e.g. `ts-jest`, `babel`, etc.).

LukeTillman:
> I'm currently trying as well to reduce the time taken to build the perses application (BE + FE) and it seems it's now impossible to build the FE without a cache mechanism that required to be warm up.
> 
> So if this new way to build the FE guarantees build stabilities and that works with a fresh clone with no caching, then I will be ok with that

Yeah, the only caching we have turned on in the CI pipeline (I think?) is npm package caching to speed up the `npm install` call. This shouldn't impact that. And builds have gotten faster. 🎉

Nexucis (approved):
LGTM then :)

sjcobb (approved):

PR #570, opened 2022-09-26
implement the new datasource format

Nexucis:
This PR is implementing the new datasource definition we drafted in July.

Like for the panels and later for the variables, I'm using CUE to define the Prometheus datasource.
Since it includes a http configuration, the server needs to be aware about the config. That's why we can find as well the go struct that is matching this http config.
Which means modifying the Cue definition of the http proxy cannot be done without doing the same in the Go code.

As the HTTP Proxy configuration can be found at any depth in the plugin, there is a method using go reflect that will search in the entire datasource definition if the proxy configuration is defined.

Hopefully after this PR, the FE should be able to contact the API to get the datasource definition. The e2e tests has been updated to reflect to new way to define the datasource and it is working. So the FE shouldn't have an issue to contact the API.


internal/api/shared/schemas/schemas.go
```diff
@@ -32,20 +32,27 @@ var baseQueryDef []byte
 //go:embed query_disjunction_generator.cue
 var queryDisjunctionGenerator []byte
 
+var cueOption = []cue.Option{
```
AntoineThebaud, line 35: I'd call this one cueValidateOptions or cueValidationOptions

internal/api/shared/schemas/schemas.go
```diff
@@ -82,65 +94,102 @@ func New(conf config.Schemas) Schemas {
...
-		}
+	// Then merge with the queries disjunction schema
+	pluginSchema = enrichSchema(pluginSchema)
+	if pluginSchema.Err() != nil {
+		logrus.WithError(pluginSchema.Err()).Errorf("Error enriching %s schema", modelKind)
+		return pluginSchema.Err()
+	}
+
+	// do the validation using the main #panel def of the schema as entrypoint
+	unified := value.Unify(pluginSchema.LookupPath(cue.ParsePath("")))
```
AntoineThebaud, line 186: Ah I missed this one in the previous PR, you can replace `pluginSchema.LookupPath(cue.ParsePath(""))` by simply `pluginSchema` ;)

internal/api/shared/schemas/schemas.go
```diff
@@ -82,65 +94,102 @@ func New(conf config.Schemas) Schemas {
...
-			cue.Hidden(true),
-		}
-		err = unified.Validate(opts...)
-		if err != nil {
-			logrus.Debug(errors.Details(err, nil))
-			//TODO: return errors.Details(err, nil) to get a more meaningful error, but should be cleaned of line numbers & server file paths!
-			err = fmt.Errorf("invalid panel %s: %s", panelName, err) // enrich the error message returned by cue lib
-			return err
-		}
+	// Then merge with the queries disjunction schema
```
AntoineThebaud, line 178: since this function is now more generic/abstracted, this comment should be moved or removed (dosnt apply to datasources validation)

internal/api/shared/schemas/schemas.go
```diff
@@ -82,65 +94,102 @@ func New(conf config.Schemas) Schemas {
...
+	pluginSchema = enrichSchema(pluginSchema)
+	if pluginSchema.Err() != nil {
+		logrus.WithError(pluginSchema.Err()).Errorf("Error enriching %s schema", modelKind)
+		return pluginSchema.Err()
+	}
+
+	// do the validation using the main #panel def of the schema as entrypoint
```
AntoineThebaud, line 185: this comment is no longer relevant

AntoineThebaud (changes requested):

AntoineThebaud (approved):

PR #571, opened 2022-09-26
Add eslint:recommended, enable pretty output from Typescript

LukeTillman:
This just makes two minor tweaks to our dev tool configs:

1. Turns on `eslint:recommended` which was somehow missing (definitely my fault 😂 )
2. Set `pretty: true` in `tsconfig` so that the Typescript compiler outputs colors for compile errors, making that more obvious in our log output

### DIFF
--- a/ui/tsconfig.base.json
+++ b/ui/tsconfig.base.json
@@ -15,7 +15,8 @@
     "isolatedModules": true,
     "noUncheckedIndexedAccess": true,
     "declaration": true,
-    "declarationMap": true
+    "declarationMap": true,
+    "pretty": true
   },
   // For builds that use ts-node to compile the configs (e.g. webpack, jest)
   "ts-node": {
--- a/ui/.eslintrc.base.js
+++ b/ui/.eslintrc.base.js
@@ -14,6 +14,7 @@
 // Base eslint configuration for typescript projects
 module.exports = {
   extends: [
+    'eslint:recommended',
     'plugin:react/recommended',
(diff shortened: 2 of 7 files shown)


Nexucis (approved):

ui/tsconfig.base.json
```diff
@@ -15,7 +15,8 @@
     "isolatedModules": true,
     "noUncheckedIndexedAccess": true,
     "declaration": true,
-    "declarationMap": true
+    "declarationMap": true,
+    "pretty": true
   },
   // For builds that use ts-node to compile the configs (e.g. webpack, jest)
```
saminzadeh, line 21: Unrelated to this PR, but are comments supported in .json? My editor shows linting errors

saminzadeh (approved):

ui/tsconfig.base.json
LukeTillman: Yeah, `tsconfig` is JSONC and supports them. Not all JSON though obviously. If you're using VS Code and it doesn't do it for you automatically, you can switch the file type down in the bottom right corner by clicking on it:

Screen Shot 2022-09-26 at 2 12 52 PM

saminzadeh: Ah I see, the file I was noticing had the issue was actually turbo.json which VSCode doesn't automatically switch. Seems like[ they fixed it ](https://github.com/Microsoft/vscode/issues/50974) for tsconfig though

PR #573, opened 2022-09-28
Rename GraphQuery to TimeSeriesQuery

LukeTillman:
As discussed in chat, this renames the very generic `GraphQuery` to the more descriptive `TimeSeriesQuery` which better reflects the output format. I tried to also rename all the related types, but I may have missed a few which we can clean-up as we find them. There are [a few lines of refactoring](https://github.com/perses/perses/pull/573/commits/8b08c2461ce9be760188df7239a07796928fbc16) in here to plugin-related types that just simplified a few of the generics that aren't helpful at runtime, but most of this is just renames.

### DIFF
--- a/ui/app/sample-data/benchmarkDashboard.ts
+++ b/ui/app/sample-data/benchmarkDashboard.ts
@@ -83,10 +83,10 @@ const benchmarkDashboard: DashboardResource = {
             spec: {
               queries: [
                 {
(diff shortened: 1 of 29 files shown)


sjcobb (approved):
💥 thanks for updating this, I like that the name is more explicit now! About to go ahead and merge, and then I'll rename LineChart to TimeSeriesChart

PR #575, opened 2022-09-28
rename LineChart panel to TimeSeriesChart

sjcobb:
As [discussed in the Matrix chat](https://matrix.to/#/!XNmgYIAndZOkEvQAbb:matrix.org/$IVaC8SRjGxTE7tcwTe9uahcY1RZu7aoTu3pJH52F9UU?via=matrix.org), this PR renames our `kind: 'LineChart'` panel to `kind: 'TimeSeriesChart'`. This is a follow up to #573 which makes our names more explicit about the type of data they support. CUE schema is also updated and I did a find & replace in BE code (hopefully I didn't miss any)


docs/dashboard.md
```diff
@@ -213,7 +213,7 @@ Example:
 
 ##### LineChart
```
Nexucis, line 214: `TimeseriesChart` ?

dev/data/dashboard.json
```diff
@@ -39,7 +39,7 @@
       },
       "panels": {
         "upByLabel": {
-          "kind": "LineChart",
+          "kind": "TimeSeriesChart",
```
Nexucis, line 42: I would have written `Timeseries` I thought it was a single word, but I'm not the best in english here 😅

Nexucis (approved):
Thank you so much @sjcob you changed it everywhere including in the cue schema and in the go code !

LukeTillman (approved):
Oh man, you're making me look bad. I should've updated cue too. 😂

dev/data/dashboard.json
sjcobb: Yeahh `TimeseriesChart` looks really weird to me and I kind of piggybacked off @LukeTillman's `TimeSeriesQuery` change too so those would be consistent

eunicorn (approved):

dev/data/dashboard.json
sjcobb: Maybe worth another discussion, but I'm going to go ahead and merge this--if we decide to make the 'S' lowercase, I think that should be a follow-up PR where TimeSeriesQuery is also changed

PR #578, opened 2022-09-28
Upgrade to react-query v4, clean-up test logs

LukeTillman:
This updates our `react-query` dependency to the latest major version (v4) which helps clean-up our Jest test logs in the dashboards package where we were getting some console error messages related to that package. I also cleaned up one other warning that was getting logged by MUI due to the panel types being loaded dynamically.

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
@@ -28,7 +28,7 @@ import PanelDrawer from './PanelDrawer';
 describe('Panel Drawer', () => {
   const renderPanelDrawer = () => {
     const { addMockPlugin, pluginRegistryProps } = mockPluginRegistryProps();
-    addMockPlugin('Panel', 'FakePanel', FAKE_PANEL_PLUGIN);
+    addMockPlugin('Panel', 'LineChart', FAKE_PANEL_PLUGIN);
 
     const { store, DashboardProviderSpy } = createDashboardProviderSpy();
 
--- a/ui/dashboards/src/components/PanelDrawer/PanelTypeSelect.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelTypeSelect.tsx
@@ -1,0 +1,41 @@
+// Copyright 2022 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Select, SelectProps, MenuItem } from '@mui/material';
+import { useListPluginMetadata } from '@perses-dev/plugin-system';
+
+export type PanelTypeSelectProps = Omit<SelectProps<string>, 'children'>;
+
+/**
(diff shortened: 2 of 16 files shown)


saminzadeh (approved):

ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
```diff
@@ -28,7 +28,7 @@ import PanelDrawer from './PanelDrawer';
 describe('Panel Drawer', () => {
   const renderPanelDrawer = () => {
     const { addMockPlugin, pluginRegistryProps } = mockPluginRegistryProps();
-    addMockPlugin('Panel', 'FakePanel', FAKE_PANEL_PLUGIN);
+    addMockPlugin('Panel', 'LineChart', FAKE_PANEL_PLUGIN);
```
cndonovan, line 31: Yay nice

ui/dashboards/src/components/PanelDrawer/PanelTypeSelect.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Select, SelectProps, MenuItem } from '@mui/material';
+import { useListPluginMetadata } from '@perses-dev/plugin-system';
+
+export type PanelTypeSelectProps = Omit<SelectProps<string>, 'children'>;
+
+/**
+ * Displays a MUI Select input for selecing a Panel type. Queries the plugin system to get the list of all panel types.
+ */
+export function PanelTypeSelect(props: PanelTypeSelectProps) {
+  const { value: propValue, ...others } = props;
```
cndonovan, line 23: I'm down to keep calling this `value` within this component, but is there something you're trying to distinguish from?

ui/dashboards/src/components/PanelDrawer/PanelTypeSelect.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { data, isLoading } = useListPluginMetadata('Panel');
+
+  // Pass an empty value while options are still loading so MUI doesn't complain about us using an "out of range" value
+  const value = propValue !== '' && isLoading ? '' : propValue;
+
+  // TODO: Does this need a loading indicator of some kind?
```
cndonovan, line 29: Oo maybe ya 🤔  

Seems like this data always be fetched when the editor component is rendered... so I guess it's a question of how long the fetch could take

Alternative is that we could fetch this data when we load the app / load perses framework and cache it or something... it doesn't seem like the list of avaialbe plugins would change very frequently, if my mental model of customer behavior is right

ui/dashboards/src/test/render.tsx
```diff
@@ -12,13 +12,13 @@
...
-
-const queryClient = new QueryClient({ defaultOptions: { queries: { refetchOnWindowFocus: false } } });
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 
 /**
  * Test helper to render a React component with some common app-level providers wrapped around it.
  */
 export function renderWithContext(ui: React.ReactElement, options?: Omit<RenderOptions, 'queries'>) {
+  // Create a new QueryClient for each test to avoid caching issues
+  const queryClient = new QueryClient({ defaultOptions: { queries: { refetchOnWindowFocus: false, retry: false } } });
```
cndonovan, line 22: 👏

cndonovan (approved):

ui/dashboards/src/components/PanelDrawer/PanelTypeSelect.tsx
LukeTillman: Yeah, right now it's not even noticeable since loading this data is already cached in a couple places, but you always wonder about the worst case scenario of someone with a slow connection.

sjcobb (approved):

eunicorn (approved):

PR #579, opened 2022-09-29
Shan/var chaining

saminzadeh:
Video demo: https://www.loom.com/share/82d034251f8c4d628cfdc23b6cf8dce9

This includes the following:
- PromQL variables now support variable chaining. Meaning you can use another variable's value in a variable definition.
For example, lets assume you had a variable called `job`, you can use that in your definition. The `handler` variable will update whenever the `job` variable value changes.
```ts
spec: {
   label_name: 'handler',
   matchers: ['prometheus_http_requests_total{job=~"$job"}'],
},
````
- Moves variable options loading to react-query
- Modifies the plugin spec to include a new property called `dependsOn` which allows the plugin to return a list of variables a variable depends on. This optimizes loading of variables and also gives the ability to let the "framework" which variables a variable depends on to build features like a dependency graph in the future. We could potentially use this paradigm for TimeSeriesQuery plugin as well.

 

### DIFF
--- a/ui/app/src/App.tsx
+++ b/ui/app/src/App.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { useMemo } from 'react';
-import { useSearchParams, useLocation } from 'react-router-dom';
+import { useSearchParams } from 'react-router-dom';
 import { Box, useTheme } from '@mui/material';
 import {
   ChartsThemeProvider,
@@ -36,10 +36,7 @@ const ECHARTS_THEME_OVERRIDES = {};
 function App() {
   const { getInstalledPlugins, importPluginModule } = useBundledPlugins();
 
-  const [searchParams, setSearchParams] = useSearchParams();
-
-  // TODO: remove temporary location.key reload hack when routing setup properly
-  const location = useLocation();
+  const [searchParams] = useSearchParams();
 
   const muiTheme = useTheme();
   const chartsTheme: PersesChartsTheme = useMemo(() => {
@@ -67,10 +64,10 @@ function App() {
           <ChartsThemeProvider themeName="perses" chartsTheme={chartsTheme}>
             <PluginRegistry getInstalledPlugins={getInstalledPlugins} importPluginModule={importPluginModule}>
               <LegacyDataSourceRegistry>
-                <QueryStringProvider queryString={searchParams} setQueryString={setSearchParams}>
+                <QueryStringProvider queryString={searchParams}>
                   <ErrorBoundary FallbackComponent={ErrorAlert}>
                     {/* temp fix to ensure dashboard refreshes when URL changes since setQueryString not reloading as expected  */}
-                    <ViewDashboard key={location.key} />
+                    <ViewDashboard />
                   </ErrorBoundary>
                 </QueryStringProvider>
               </LegacyDataSourceRegistry>
--- a/ui/dashboards/src/components/Variables/Variable.tsx
+++ b/ui/dashboards/src/components/Variables/Variable.tsx
@@ -11,10 +11,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { useEffect, useMemo, useRef, useCallback } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import { Select, FormControl, InputLabel, MenuItem, Box, LinearProgress, TextField } from '@mui/material';
 import { VariableName, ListVariableDefinition, VariableValue } from '@perses-dev/core';
-import { usePlugin, DEFAULT_ALL_VALUE, useLegacyDatasources } from '@perses-dev/plugin-system';
+import {
+  usePlugin,
+  DEFAULT_ALL_VALUE,
+  useLegacyDatasources,
+  useTemplateVariableValues,
+  VariableStateMap,
+} from '@perses-dev/plugin-system';
+import { useQuery } from 'react-query';
 import { useTemplateVariable, useTemplateVariableActions, useTemplateVariableStore } from '../../context';
 
 type TemplateVariableProps = {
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

**Comment by Nexucis** [comment]
ooouh niiice !!

1. Does it induce a chain in the loading of the variable + panel ? 

2. Again I'm late in the game, just looking at the syntax for the variable. Does `$<variable_name>` is the one we want, or the we choose because it's one of the two used by Grafana ? 

I'm wondering if the syntax `${variable_name}` should be used instead. It would be easy to detect the variable with a syntax like that, because you have a clean begin and an end in the pattern.

**Comment by Nexucis** [comment]
ah I watched the video, and I think if I understood correctly, there is an order of loading in the variables, so cool @saminzadeh well done :)

File: ui/plugin-system/src/model/variables.ts [hunk 1]
```diff
@@ -12,19 +12,23 @@
...
 }
 
 /**
  * Plugin for handling custom VariableDefinitions.
  */
 export interface VariablePlugin<Spec = unknown> {
   getVariableOptions: GetVariableOptions<Spec>;
+
+  /** Returns a list of variables name this variable depends on. Used to optimize fetching */
+  dependsOn?: (definition: ListVariableDefinition<Spec>) => string[];
```
saminzadeh, line 31: This is a new feature on a plugin that I think it worth discussing. Basically makes its easy for plugins to still "own" how template variables are consumed in the plugin but have the framework still understand the dependencies.

saminzadeh:
> 1. Does it induce a chain in the loading of the variable + panel ?

Yep! You can see that code here, basically waits to load unless the variables it depends on are loaded.
https://github.com/perses/perses/pull/579/files#diff-0b0f0d68ff8fdcee5fd5a8fb605d7bf10fe65f34de670456976d8f0a6bee420fR71

> Again I'm late in the game, just looking at the syntax for the variable. Does $variable_name is the one we want, or the we choose because it's one of the two used by Grafana ?
> 
> I'm wondering if the syntax ${variable_name} should be used instead. It would be easy to detect the variable with a syntax like that, because you have a clean begin and an end in the pattern.

We can have the discussion. FYI it also [supports](https://cs.github.com/perses/perses/blob/54ba48affefdf14a608c90aeb94992add1b3af23/ui/prometheus-plugin/src/model/utils.ts#L45) ${variable_name}. That being said, I don't mind the syntax especially since its already used. Unless there is a valid reason to change it. Side note: I worked on a project where we changed to handlebars and people kept getting confused since they were so used to Grafana style.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -34,33 +41,53 @@ export function TemplateVariable({ name }: TemplateVariableProps) {
...
+      const resp = await variablePlugin?.getVariableOptions(definition, { datasources, variables });
+      return resp?.data ?? [];
+    },
+    { enabled: !!variablePlugin || waitToLoad }
+  );
 
   useEffect(() => {
-    loadOptions();
-  }, [loadOptions]);
+    setVariableLoading(name, variablesOptionsQuery.isFetching);
```
saminzadeh, line 86: In a follow up, I'm thinking about moving all this logic to the TemplateVariableStore so the component is more lightweight

ui/app/src/App.tsx
```diff
@@ -67,10 +64,10 @@ function App() {
           <ChartsThemeProvider themeName="perses" chartsTheme={chartsTheme}>
             <PluginRegistry getInstalledPlugins={getInstalledPlugins} importPluginModule={importPluginModule}>
               <LegacyDataSourceRegistry>
-                <QueryStringProvider queryString={searchParams} setQueryString={setSearchParams}>
+                <QueryStringProvider queryString={searchParams}>
                   <ErrorBoundary FallbackComponent={ErrorAlert}>
                     {/* temp fix to ensure dashboard refreshes when URL changes since setQueryString not reloading as expected  */}
-                    <ViewDashboard key={location.key} />
```
saminzadeh, line 73: I discussed this with @sjcobb  offline but we had to remove this because it was remounting the whole app and wiping out state. We will address the proper way of syncing URL state in a follow up PR.

LukeTillman:
Super cool demo. Need to think on the dependencies thing a bit, so more thoughts coming soon.

saminzadeh:
> Super cool demo. Need to think on the dependencies thing a bit, so more thoughts coming soon.

Thanks! The other approach I thought about was exposing a function called `queryKey` that gets passed into the [queryKey argument](https://tanstack.com/query/v4/docs/guides/query-keys) for react-query which would give you even more control. 

For example:
```ts
export const PrometheusLabelValuesVariable: VariablePlugin<PrometheusLabelValuesVariableOptions> = {
  getVariableOptions: async (definition, ctx) => {
    //... implementation
  },
  queryKey: (definition, ctx) => { 
    // return some unique string here that can consume variables, time range state, etc
    return someKey;
  }
};
```

I ultimately decided against it because I felt it was too primitive and doesn't really solve the problem of the framework knowing about dependencies.

ui/app/src/App.tsx
sjcobb: I'm working to fix this now so hopefully we'll be able to re-enable setQueryString before we publish v0.9!

ui/plugin-system/src/model/variables.ts
LukeTillman: > > Super cool demo. Need to think on the dependencies thing a bit, so more thoughts coming soon.
> 
> Thanks! The other approach I thought about was exposing a function called `queryKey` that gets passed into the [queryKey argument](https://tanstack.com/query/v4/docs/guides/query-keys) for react-query which would give you even more control.
> 
> For example:
> 
> ```ts
> export const PrometheusLabelValuesVariable: VariablePlugin<PrometheusLabelValuesVariableOptions> = {
>   getVariableOptions: async (definition, ctx) => {
>     //... implementation
>   },
>   queryKey: (definition, ctx) => { 
>     // return some unique string here that can consume variables, time range state, etc
>     return someKey;
>   }
> };
> ```
> 
> I ultimately decided against it because I felt it was too primitive and doesn't really solve the problem of the framework knowing about dependencies.

Yeah, at first glance for some reason, I thought you'd put this into the JSON spec and I was thinking to myself last night, "why isn't it just a function on the plugin". Then I came back and looked at the PR again and was like, "Oh, that's totally what I was thinking you should do". 😂

ui/app/src/App.tsx
LukeTillman: Yeah, I noticed this too locally. LMK if you need some help figuring it out.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -34,33 +41,53 @@ export function TemplateVariable({ name }: TemplateVariableProps) {
...
-    setVariableLoading(name, false);
-  }, [variablePlugin, definition, name, setVariableLoading, setVariableOptions, datasources]);
+  let waitToLoad = false;
+  if (dependsOnVariables) {
+    waitToLoad = dependsOnVariables.some((v) => variables[v]?.loading);
+  }
+
+  const variablesValueKey = getVariableValuesKey(variables);
+
+  const variablesOptionsQuery = useQuery(
```
LukeTillman, line 76: Do you think the `usePlugin` call being a separate async call is complicating this flow/logic more than it could be? If so, you could always use the plugin registry directly to combine some of the async stuff, e.g.:

```ts
const { getPlugin } = usePluginRegistry();

const pluginStuff = useQuery([ key ], async () => {
  const plugin = await getPlugin('Variable', kind);
  const options = await plugin.getVariableOptions(...);
  return { options, plugin /* etc */ };
});
```

Not sure if that would be helpful at all, but just a thought.

ui/plugin-system/src/model/variables.ts
LukeTillman: My only question here is: do you need the entire definition as an argument, or just the `Spec`?

ui/prometheus-plugin/src/plugins/prometheus-variables.ts
```diff
@@ -44,6 +46,15 @@ function getQueryOptions(ctx: GetVariableOptionsContext) {
   return queryOptions;
 }
 
+function interpolatePluginDefinition(def: object, ctx: GetVariableOptionsContext) {
+  return replaceTemplateVariablesInObject(def, ctx.variables);
+}
+
+const dependsOn: VariablePlugin['dependsOn'] = (definition) => {
+  const stringDefinition = JSON.stringify(definition.spec.plugin.spec);
```
LukeTillman, line 54: A little worried about using `JSON.stringify` here (and above in the `replace` function). Is there a way to write these without needing it? Just worried about the cost of it potentially being called frequently. I think I've just seen too many projects where JSON stringify/parse usage ended up being a performance issue later. 😂

LukeTillman (approved):
Overall, looks great. I think the dependencies approach you took is spot on. Left a few suggestions inline.

ui/prometheus-plugin/src/plugins/prometheus-variables.ts
saminzadeh: Haha yeah that's a fair argument. I will rewrite it so it looks at the specific properties that may have variables. I mostly did JSON.stringify as a quick way to iterate since the size of a definition would be relatively small vs a dynamic payload of objects.

ui/plugin-system/src/model/variables.ts
LukeTillman: Yeah, I've been thinking about changing up the other plugins too so they don't have access to the whole definition, just the spec (seems like it's not necessary anywhere I can think of), so if you wanted to do it, cool with me.

ui/dashboards/src/components/Variables/Variable.tsx
saminzadeh: Actually decided to keep it the same so I can utilize the `enabled` property on react-query. Although, I find this API really useful.

PR #581, opened 2022-09-29
Iterate on PanelDrawer state flow

LukeTillman:
This tries to make the state flow in the `PanelDrawer` component a little easier to understand. I tried to remove a few effects, handle resetting the form values between opens, as well as make add vs edit functionality a little clearer. Not sure if I succeeded, but we can definitely keep improving things.


ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -11,242 +11,70 @@
...
+import { useDashboardApp } from '../../context';
+import { usePanelDrawerModel } from './panel-editor-model';
+import { PanelEditorForm, panelEditorFormId, PanelEditorFormProps } from './PanelEditorForm';
+
+/**
+ * The Add/Edit panel drawer for editing a panel's options.
+ */
+export const PanelDrawer = () => {
+  const { closePanelDrawer } = useDashboardApp();
+  const model = usePanelDrawerModel();
```
LukeTillman, line 26: I tried to pull all the add vs edit functionality out into this new hook.

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -11,242 +11,70 @@
...
+              {/* Using the 'form' attribute lets us have a submit button like this outside the form element */}
+              <Button type="submit" variant="contained" form={panelEditorFormId}>
+                {model.submitButtonText}
+              </Button>
+              <Button variant="outlined" onClick={handleClose}>
+                Cancel
+              </Button>
+            </Stack>
+          </Box>
+          <PanelEditorForm onSubmit={handleSubmit} initialValues={model.initialValues} />
```
LukeTillman, line 75: See the comment above, but the form state is now in this component which gets unmounted when the drawer is closed, allow the form state values to be reset.

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </Grid>
+      </Grid>
+    </form>
+  );
+}
+
+/**
+ * The `id` attribute added to the `PanelEditorForm` component, allowing submit buttons to live outside the form.
+ */
+export const panelEditorFormId = 'panel-editor-form';
```
LukeTillman, line 111: Something I learned while working on this PR is that you can have submit buttons outside of the `form` element with just some vanilla HTML props. 🧐

ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts
```diff
@@ -0,0 +1,149 @@
...
+  // values intact from before the switch
+  const [specByKind, setSpecByKind] = useImmer<Record<string, unknown>>({ [panelPluginKind]: initialState });
+  const { data: plugin } = usePlugin('Panel', panelPluginKind, { enabled: panelPluginKind !== '' });
+  const pluginInitialSpec = useMemo(() => plugin?.createInitialOptions(), [plugin]);
+
+  // Use the value in specByKind or if unset, use the initial values from the plugin (which could still be undefined)
+  const spec = specByKind[panelPluginKind] ?? pluginInitialSpec;
+
+  // TODO: Do we want to expose more of a immer style API to plugin authors for managing their state, rather than the
+  // current "onChange" API?
```
sjcobb, line 138: I'm definitely interested in hearing more specifics about this! I need to play around some more with the current API in TimeSeriesChartOptionsEditor, but this TODO sounds appealing.

ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts
```diff
@@ -0,0 +1,149 @@
...
+
+// Props that vary based on whether the PanelDrawer is in add or edit mode
+interface PanelDrawerModel {
+  drawerTitle: string;
+  submitButtonText: string;
+  initialValues: PanelEditorFormValues;
+  applyChanges: (values: PanelEditorFormValues) => void;
+}
+
+/**
```
sjcobb, line 36: Nice! I love all the comments here and below, makes it really clear what's happening

sjcobb (approved):

ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  // Otherwise we don't have a panel key, so we're trying to edit an existing panel
+  const existingPanel = panels[panelKey];
+
+  // TODO: Can we better express this via the type system on the dashboard store to avoid these states?
+  if (existingPanel === undefined) {
+    throw new Error(`Cannot find existing panel '${panelKey}' to edit`);
+  }
+  if (groupIndex === undefined) {
```
eunicorn, line 90: is there a case when groupIndex is undefined for panel drawer?

LukeTillman: Yeah, the types say it could be `undefined`, so I added this check, but like my TODO says, I wonder if this types should change since maybe that shouldn't be possible?

eunicorn: i forgot why i made it optional, but i think you can change it since i can't think of a case when you don't pass in a groupIndex

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -11,242 +11,70 @@
...
+              display: 'flex',
+              alignItems: 'center',
+              marginBottom: (theme) => theme.spacing(2),
+              paddingBottom: (theme) => theme.spacing(2),
+              borderBottom: (theme) => `1px solid ${theme.palette.grey[100]}`,
+            }}
+          >
+            <Typography variant="h2">{model.drawerTitle}</Typography>
+            <Stack direction="row" spacing={1} sx={{ marginLeft: 'auto' }}>
+              {/* Using the 'form' attribute lets us have a submit button like this outside the form element */}
```
eunicorn, line 66: oh cool!

ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts
LukeTillman: I think it needs to be optional for the "Add Panel" button in the header? That was my thinking anyway.

LukeTillman: But I'll look at the store and its types next and see if there are any improvements I can make there.

ui/dashboards/src/components/PanelDrawer/PanelDrawer.tsx
```diff
@@ -11,242 +11,70 @@
...
+import { PanelEditorForm, panelEditorFormId, PanelEditorFormProps } from './PanelEditorForm';
+
+/**
+ * The Add/Edit panel drawer for editing a panel's options.
+ */
+export const PanelDrawer = () => {
+  const { closePanelDrawer } = useDashboardApp();
+  const model = usePanelDrawerModel();
+
+  // When the user clicks close, start closing but don't call the store yet to keep values stable during animtation
```
eunicorn, line 28: it's not clear to me why we need this. Is it because you can somehow stop the drawer from closing?

eunicorn (approved):

ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts
eunicorn: It seems like we only need to call setSpec/onSpecChange when user changes panel type though. If so, then the onChange api is more intuitive and simple. But if we have to call it every time some input changes, then i like the immer way!

ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts
```diff
@@ -0,0 +1,149 @@
...
+  const { panelKey, groupIndex } = panelDrawer;
+  if (panelKey === undefined) {
+    return {
+      drawerTitle: 'Add Panel',
+      submitButtonText: 'Add',
+      initialValues: {
+        name: '',
+        description: '',
+        group: groupIndex ?? 0,
+        kind: '',
```
eunicorn, line 59: should we always default it to Time Series?

PR #582, opened 2022-09-30
Find and replace copyright year

cndonovan:
It does what the PR title says :) 


sjcobb (approved):
Thanks! Will definitely clean up our PRs

Nexucis:
I'm not super familiar with the copyright year, but based on what I'm seeing in other project, it doesn't look like we need to update the year every year.
But I don't know if it's a good thing or a miss

PR #583, opened 2022-09-30
refactor the way the ui is embedded

Nexucis:
When replacing `goreleaser` by `promu`, I finally realised what was the issue that was crashing `gorelease`. It was the way that we were injecting the FE in the binary.
It was just not efficient at all.

This new way is using the `embed` of Go. It's far far more efficient in term of compilation.


Nexucis:
le build passed from 10min or more and crashing to 3min stable

AntoineThebaud (approved):

PR #593, opened 2022-10-03
Use snake_case in spec definitions

saminzadeh:
Renaming:
- `defaultValue` -> `default_value`
- `allowMultiple` -> `allow_multiple`
- `allowAllValue` -> `allow_all_value`


cndonovan (approved):

cndonovan:
If snake_case is what we be doing right now, then works 4 me

Nexucis (approved):
Ah yeah good to point to align everything with snake_case !

PR #596, opened 2022-10-04
add static panel options for TimeSeriesChart

sjcobb:
## Overview

Add hard-coded options to TimeSeriesChart panel as first step in chart options editing. Static panel can now be previewed in edit drawer and added to dashboards, but _nothing is editable yet_ (adding form controls is next).

## Screenshots:

hard-coded_panel_edit_ex

hard-coded_panel_dash_ex

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
@@ -12,12 +12,38 @@
 // limitations under the License.
 
 import { OptionsEditorProps } from '@perses-dev/plugin-system';
-import { Box } from '@mui/material';
+import { Stack, Box } from '@mui/material';
 import { TimeSeriesChartOptions } from './time-series-chart-model';
+import { TimeSeriesChartPanel, TimeSeriesChartProps } from './TimeSeriesChartPanel';
 
 export type TimeSeriesChartOptionsEditorProps = OptionsEditorProps<TimeSeriesChartOptions>;
 
 export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditorProps) {
   const { value } = props;
-  return <Box>{JSON.stringify(value)}</Box>;
+
+  const panelPreviewProps: TimeSeriesChartProps = {
+    definition: {
+      kind: 'Panel',
+      spec: {
+        display: { name: 'Single Query' },
+        plugin: {
+          kind: 'TimeSeriesChart',
+          spec: {
+            ...value,
+          },
+        },
+      },
+    },
+    contentDimensions: { width: 500, height: 250 },
+  };
+
+  return (
+    <>
+      <TimeSeriesChartPanel {...panelPreviewProps} />
+      <Stack spacing={1}>
+        <Box>{JSON.stringify(value)}</Box>
+        {/* TODO: add form controls to edit panel options */}
+      </Stack>
+    </>
+  );
 }
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
@@ -25,7 +25,8 @@ export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditor
     definition: {
       kind: 'Panel',
       spec: {
-        display: { name: 'Single Query' },
+        // TODO: how to exlude display (display.name is required but gets overriden)
+        display: { name: 'Temp Name' },
         plugin: {
           kind: 'TimeSeriesChart',
           spec: {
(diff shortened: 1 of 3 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
```diff
@@ -12,12 +12,39 @@
...
+        plugin: {
+          kind: 'TimeSeriesChart',
+          spec: {
+            ...value,
+          },
+        },
+      },
+    },
+    contentDimensions: { width: 500, height: 250 },
+  };
```
cndonovan, line 39: At one point, I played with using the generic `Panel` in the `PanelDrawer` (at the time, it was accessible in PanelDrawer, not sure if it still is), and then it would take care of rendering the preview for every type of panel. But also not sure this approach can work fully, it was more prototype code.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -17,9 +17,9 @@ import TimeSeriesQueryRunner from './TimeSeriesQueryRunner';
 import { TimeSeriesChartOptions } from './time-series-chart-model';
 import { TimeSeriesChartContainer } from './TimeSeriesChartContainer';
 
-export type LineChartProps = PanelProps<TimeSeriesChartOptions>;
+export type TimeSeriesChartProps = PanelProps<TimeSeriesChartOptions>;
 
-export function TimeSeriesChartPanel(props: LineChartProps) {
+export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
```
cndonovan, line 22: <3

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -30,6 +30,23 @@ export interface TimeSeriesChartOptions {
...
+          plugin: {
+            kind: 'PrometheusTimeSeriesQuery',
+            spec: {
+              query: 'up',
+            },
+          },
+        },
+      },
+    ],
+    // TODO: configurable legend, change from show_legend to legend.show
```
cndonovan, line 46: :+1:

I'm probably going to write down some thoughts today, but basically it was on my mind to sync with you about the schema for chart options

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -30,6 +30,23 @@ export interface TimeSeriesChartOptions {
...
   return {
-    queries: [],
+    queries: [
+      {
+        kind: 'TimeSeriesQuery',
+        spec: {
+          plugin: {
+            kind: 'PrometheusTimeSeriesQuery',
+            spec: {
+              query: 'up',
```
cndonovan, line 40: :+1:, we can remove this later

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
```diff
@@ -12,12 +12,39 @@
...
 export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditorProps) {
   const { value } = props;
-  return <Box>{JSON.stringify(value)}</Box>;
+
+  const panelPreviewProps: TimeSeriesChartProps = {
+    definition: {
+      kind: 'Panel',
+      spec: {
+        // TODO: how to exlude display (display.name is required but gets overriden)
+        display: { name: 'Temp Name' },
```
cndonovan, line 29: Yeah was thinking about this a little too!

If I'm understanding your comment right, my ideas so far were:
- We could have a prop on the panel component to show/hide title
- Build a new component using the underlying TimeSeriesChart + new chrome around it (very minimal chrome)

cndonovan (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
LukeTillman: I was thinking about maybe changing up the PanelPlugin API so that the plugin component only gets the `spec` prop and not the entire definition object (we already made this change in both query and variable plugins). Would that help solve it?

LukeTillman: My idea is for "preview" the PanelDrawer could just render the actual plugin component with the current spec state passed in after we made that change.

LukeTillman (approved):
See my question below, but looks great!

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
sjcobb: Yeah not getting the entire definition object would be great! I'll merge this for now with the assumption that using TimeSeriesChartPanel directly in TimeSeriesChartOptionsEditor will change very soon in favor of the solution you both described

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
sjcobb: I'll look more into this too, it definitely would be a much better approach!

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
sjcobb: yeah definitely, I should have left a TODO here too--let's sync up soon like you said, I imagine we're going to have a lot of merge conflicts soon

PR #598, opened 2022-10-04
Use dashboard store in view mode also

LukeTillman:
This just removes the `useDashboard` hook and some of the weird state flow where I screwed up and told Eunice we should be passing the dashboard spec around everywhere. 😂  Instead, it just uses the `layouts` and `panels` state already in the store for both view and edit modes. I haven't tried to do anything "smart" yet (like more granular selectors to help with re-rendering), just simplified the mess I made so we can start doing that.

### DIFF
--- a/ui/dashboards/src/components/GridLayout/GridLayout.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridLayout.tsx
@@ -13,17 +13,18 @@
 import { useState } from 'react';
 import { Responsive, WidthProvider } from 'react-grid-layout';
 import { Box, BoxProps, Collapse, GlobalStyles } from '@mui/material';
-import { GridDefinition, GridItemDefinition } from '@perses-dev/core';
+import { GridDefinition } from '@perses-dev/core';
+import { ErrorAlert, ErrorBoundary } from '@perses-dev/components';
 import { styles } from '../../css/styles';
 import { useEditMode } from '../../context';
 import { GridTitle } from './GridTitle';
+import { GridItemContent } from './GridItemContent';
 
 const ResponsiveGridLayout = WidthProvider(Responsive);
 
 export interface GridLayoutProps extends BoxProps {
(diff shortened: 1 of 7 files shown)


ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
```diff
@@ -42,18 +42,6 @@ export function useEditMode() {
   return useDashboardStore(({ isEditMode, setEditMode }) => ({ isEditMode, setEditMode }));
 }
 
-export function useDashboard() {
```
eunicorn, line 45: should we keep this for the export feature?

ui/dashboards/src/components/GridLayout/GridLayout.tsx
```diff
@@ -13,17 +13,18 @@
...
 import { useEditMode } from '../../context';
 import { GridTitle } from './GridTitle';
+import { GridItemContent } from './GridItemContent';
 
 const ResponsiveGridLayout = WidthProvider(Responsive);
 
 export interface GridLayoutProps extends BoxProps {
   groupIndex: number;
   definition: GridDefinition;
-  renderGridItemContent: (definition: GridItemDefinition, itemIndex: number) => React.ReactNode;
```
eunicorn, line 26: i've been wanting to remove this for so long 😆

LukeTillman: Hahahaha, that was back when I thought `GridLayout` might be more generic/reusable on its own, but good to just "embrace the store".

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
LukeTillman: I figure we should probably just build a "getDashboardJson" or something into the store for export (whenever we get around to it).

PR #599, opened 2022-10-04
Pass spec only to panel plugins, rather than definition

LukeTillman:
This just simplifies Panel plugins to only take the plugin's `spec` as props, rather than the entire `PanelDefinition` object. This is consistent with what we've already done in variable and time series query plugins where we've found that the plugin code doesn't really care about the stuff in the full-blown definition at this point. It should also make building panel previews for visual editing a little easier. (See some of the discussion in #596)

### DIFF
--- a/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
@@ -24,14 +24,7 @@ import { GaugeChartOptions } from './gauge-chart-model';
 export type GaugeChartPanelProps = PanelProps<GaugeChartOptions>;
 
 export function GaugeChartPanel(props: GaugeChartPanelProps) {
-  const {
-    definition: {
-      spec: {
-        plugin: { spec: pluginSpec },
-      },
-    },
-    contentDimensions,
-  } = props;
+  const { spec: pluginSpec, contentDimensions } = props;
   const { query, calculation, max } = pluginSpec;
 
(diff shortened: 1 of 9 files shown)


sjcobb (approved):
LGTM! I'll remove TimeSeriesChartPanel from the editor like we talked about

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
```diff
@@ -24,14 +24,7 @@ import { GaugeChartOptions } from './gauge-chart-model';
...
 
 export function GaugeChartPanel(props: GaugeChartPanelProps) {
-  const {
-    definition: {
-      spec: {
-        plugin: { spec: pluginSpec },
-      },
-    },
-    contentDimensions,
-  } = props;
```
sjcobb, line 34: so much cleaner 🎉

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
```diff
@@ -81,9 +74,10 @@ const useChartData = (data: TimeSeriesData | undefined, calculation: Calculation
     return {
       calculatedValue,
       seriesData,
-      name,
+      // TODO: How is this name used in StatChart? Do we really need it? Should it be a chart spec option?
```
sjcobb, line 77: This can be removed completely now, way [back in v0.4](https://github.com/perses/perses/blob/v0.4.2/ui/components/src/StatChart.tsx#L107) and before, I had an optional prop to show the name down beside the value, but it isn't used anymore since it's redundant with the panel header

PR #600, opened 2022-10-04
[WIP] potential legend spec

sjcobb:
Iterate on how we should structure the spec for legends

### DIFF
--- a/ui/app/sample-data/demoDashboard.ts
+++ b/ui/app/sample-data/demoDashboard.ts
@@ -160,7 +160,7 @@ const demoDashboard: DashboardResource = {
                       spec: {
                         query:
                           '1 - node_filesystem_free_bytes{job="node",instance=~"$instance",fstype!="rootfs",mountpoint!~"/(run|var).*",mountpoint!=""} / node_filesystem_size_bytes{job="node",instance=~"$instance"}',
-                        legendFormat: '{{instance}} {{job}}', // figure this syntax out with Shan / Luke
+                        legend_format: '{{instance}} {{job}}', // figure this syntax out with Shan / Luke
                       },
                     },
                   },
@@ -207,7 +207,6 @@ const demoDashboard: DashboardResource = {
                   },
                 },
               ],
-              show_legend: false,
               unit: {
                 kind: 'PercentDecimal',
                 decimal_places: 1,
@@ -256,7 +255,6 @@ const demoDashboard: DashboardResource = {
                 kind: 'Decimal',
                 decimal_places: 2,
               },
-              show_legend: true,
             },
           },
         },
--- a/ui/app/sample-data/demoDashboard.ts
+++ b/ui/app/sample-data/demoDashboard.ts
@@ -159,47 +159,19 @@ const demoDashboard: DashboardResource = {
                       kind: 'PrometheusTimeSeriesQuery',
                       spec: {
                         query:
-                          'node_memory_MemTotal_bytes{job="node",instance=~"$instance"} - node_memory_MemFree_bytes{job="node",instance=~"$instance"} - node_memory_Buffers_bytes{job="node",instance=~"$instance"} - node_memory_Cached_bytes{job="node",instance=~"$instance"}',
-                      },
-                    },
-                  },
-                },
-                {
-                  kind: 'TimeSeriesQuery',
-                  spec: {
-                    plugin: {
-                      kind: 'PrometheusTimeSeriesQuery',
-                      spec: {
-                        query: 'node_memory_Buffers_bytes{job="node",instance=~"$instance"}',
-                      },
-                    },
-                  },
-                },
-                {
-                  kind: 'TimeSeriesQuery',
-                  spec: {
-                    plugin: {
-                      kind: 'PrometheusTimeSeriesQuery',
-                      spec: {
-                        query: 'node_memory_Cached_bytes{job="node",instance=~"$instance"}',
-                      },
-                    },
-                  },
-                },
-                {
-                  kind: 'TimeSeriesQuery',
-                  spec: {
-                    plugin: {
-                      kind: 'PrometheusTimeSeriesQuery',
-                      spec: {
-                        query: 'node_memory_MemFree_bytes{job="node",instance=~"$instance"}',
+                          '1 - node_filesystem_free_bytes{job="node",instance=~"$instance",fstype!="rootfs",mountpoint!~"/(run|var).*",mountpoint!=""} / node_filesystem_size_bytes{job="node",instance=~"$instance"}',
+                        legendFormat: '{{instance}} {{job}}', // figure this syntax out with Shan / Luke
                       },
                     },
                   },
                 },
               ],
-              show_legend: true,
-              unit: { kind: 'Bytes' },
+              legend: {
+                show: true,
+                type: 'compact', // optional, eventually we'll have 'table' in addtition to default 'compact'
+                position: 'bottom', // optional, defaults to bottom
+              },
+              unit: { kind: 'Percent' },
             },
           },
         },


ui/app/sample-data/demoDashboard.ts
```diff
@@ -159,47 +159,19 @@ const demoDashboard: DashboardResource = {
...
-                },
-                {
-                  kind: 'TimeSeriesQuery',
-                  spec: {
-                    plugin: {
-                      kind: 'PrometheusTimeSeriesQuery',
-                      spec: {
-                        query: 'node_memory_MemFree_bytes{job="node",instance=~"$instance"}',
+                          '1 - node_filesystem_free_bytes{job="node",instance=~"$instance",fstype!="rootfs",mountpoint!~"/(run|var).*",mountpoint!=""} / node_filesystem_size_bytes{job="node",instance=~"$instance"}',
+                        legendFormat: '{{instance}} {{job}}', // figure this syntax out with Shan / Luke
```
cndonovan, line 163: To support showing/hiding queries on charts*, we'll need some `is_hidden` property too.

We could do:
```
spec: {
  display: {
    legend_format: ...,
    is_hidden: ...,
  },
  query: ...,
}
```
Or:
```
spec: {
  legend_format: ...,
  is_hidden: ...,
  query: ...,
}
```

* I am 95% sure that is the intention of the toggle in this design, but I'll talk to Mohit to confirm.
Screen Shot 2022-10-04 at 12 44 40 PM

ui/app/sample-data/demoDashboard.ts
```diff
@@ -159,47 +159,19 @@ const demoDashboard: DashboardResource = {
...
                 },
               ],
-              show_legend: true,
-              unit: { kind: 'Bytes' },
+              legend: {
+                show: true,
+                type: 'compact', // optional, eventually we'll have 'table' in addtition to default 'compact'
+                position: 'bottom', // optional, defaults to bottom
+              },
+              unit: { kind: 'Percent' },
```
cndonovan, line 174: This looks great to me 👍

ui/app/sample-data/demoDashboard.ts
sjcobb: Ohh gotcha, yeah I was going to say we'd just leave it out of the queries array when we don't want it to show, but forgot about this disabled state

ui/app/sample-data/demoDashboard.ts
sjcobb: Another thought, we'll need to figure out how we're going to do multiple y axes in the future (unit will need to move at that point). We could do a y_axes array, but I think it'll depend on if we want a max of 2 (primary on left, secondary on right) or will users need to be able to add more than 2?

ui/app/sample-data/demoDashboard.ts
sjcobb: I think I slightly prefer the second option where it isn't inside display, but curious what everyone else thinks

ui/app/sample-data/demoDashboard.ts
```diff
@@ -159,47 +159,19 @@ const demoDashboard: DashboardResource = {
...
-                },
-                {
-                  kind: 'TimeSeriesQuery',
-                  spec: {
-                    plugin: {
-                      kind: 'PrometheusTimeSeriesQuery',
-                      spec: {
-                        query: 'node_memory_MemFree_bytes{job="node",instance=~"$instance"}',
+                          '1 - node_filesystem_free_bytes{job="node",instance=~"$instance",fstype!="rootfs",mountpoint!~"/(run|var).*",mountpoint!=""} / node_filesystem_size_bytes{job="node",instance=~"$instance"}',
+                        legend_format: '{{instance}} {{job}}', // figure this syntax out with Shan / Luke
```
sjcobb, line 163: Wonder if just using `format` would be better or is that too generic? The thought is that this option will also be used to format the focused series names in the tooltip

cndonovan: I was even thinking just `legend`, for me `legend` indicates both the text in the box that appears outside the chart and the text that appears in the tooltip... 

And we could change below to `legend_options`? 

Or another idea, `legend_name`?

I'm mostly brainstorming...

I think `format` by itself wouldn't mean a lot to me, but I get why you're trying to think about something generic

ui/app/sample-data/demoDashboard.ts
cndonovan: Oh that is interesting...

Would be maximum flexibility to plan for more than 2 axes...

But my knee-jerk reaction says 2 axes covers the majority of use cases, and it's enough to plan for that / planning for more might be extra abstraction that doesn't help us much. 

But maybe there's some data could help us know if that's true

ui/app/sample-data/demoDashboard.ts
cndonovan: (I'd also be fine with `legend` and `legend`, i.e. `legend` field appears inside the query schema, and inside the chart schema (and has different meanings))

LukeTillman:
Just a couple general thoughts around format in particular:

1. In Grafana, format is one of the Prometheus Datasource's options (i.e. the thing that runs the query) and so the format is something that's specific to that datasource. Always felt weird to me that it's a totally different format than other "templating" things in Grafana (e.g. template variables). Looks like it uses handlebars syntax maybe?
2. I've always wanted to see some concrete use cases of how people use it to better understand what we're trying to design for here.
3. Been wondering if this should actually be some kind of option/plugin/something for the time series chart itself and not the thing running the query? Or is the plugin running the query always going to know best?
4. If this is an option in the query, what does this option mean in the context of things that don't have a legend? Feels weird to have this on the query but have it be used in something that doesn't even have a legend?

Anyway, I think my main point is, it's hard to design this "right" when I'm not sure I fully understand what we want to accomplish with formatting.

ui/app/sample-data/demoDashboard.ts
saminzadeh: `legend_format` makes sense to me. It's really formatting the name of the individual times series so something like `series_name_format` would also make sense to me. Maybe thats better since Legend seems to be more coupled with chart visualizations? This should work for Tables as well which doesn't really include legends.

sjcobb: yeah good point about tables, it's a little verbose, but I do like `series_name_format`, it makes sense for all the use-cases we've discussed so far: legends, tooltips, and tables

Nexucis:
Adding more chaos here :) : 

1. In the context of the time-series, I think the use-case about "legend formatting" is just to be able to distinct two lines. You could see it as an hash at then end. But since a hash is not super readable, you provide to the user a way to distinct two lines with information coming from the data.

As a side note, being able to identify precisely a line can help to preserve the color from a chart to another, which then help to read the different Panel and to correlate the information together quickly.

2. Since the query plugin need to respect a format so it can be read by the `LineChart`, (from my perspective) the legend format can be put at the query level like that:

```typescript
                {
                  kind: 'TimeSeriesQuery',
                  spec: {
                    hidden: false,
                    legend_format: "something",
                    plugin: {
                      kind: 'PrometheusTimeSeriesQuery',
                      spec: {
                        query:'query',
                      },
                    },
                  },
                },
```

ui/app/sample-data/demoDashboard.ts
Nexucis: I would go for the second option too. `display` seems wrong regarding the effect of `hidden`. 

By the way by `is_hidden`, do we actually mean "deactivate the query` so it's not ran? Or do we run the query and then we don't just display the result. 

and `display` could conflict with the other `display` coming from the chart itself.

saminzadeh:
> Been wondering if this should actually be some kind of option/plugin/something for the time series chart itself and not the thing running the query? Or is the plugin running the query always going to know best?

> If this is an option in the query, what does this option mean in the context of things that don't have a legend? Feels weird to have this on the query but have it be used in something that doesn't even have a legend?

I think this is the important bit - we can think of them independently. The native Prometheus query response doesn't have the concept of a time series name, it basically just returns a list of unique labels for each individual series.

image

Then it's up to a visualization library for how to format a time series name based on labels. In the above example, some UIs do the following format: `up{instance="cadvisor:8080", job="cadvisor"}` or `$__name.value{$listOfLabelKeyValuePairs}`

The reason why it may be helpful to have a formatter at a datasource level is that you can define how individual time series should be named so its re-useable across multiple visualizations , e.g. a Table column vs TimeSeriesChart series name. This would populate the [`name` property](https://github.com/perses/perses/blob/main/ui/plugin-system/src/model/time-series-queries.ts#L41) in our TimeSeries response which is equivalent to Grafana's[ `name` property of a Field](https://grafana.com/docs/grafana/latest/packages_api/data/field/#name-property) in Dataframes. That being said it's feasible that a table and timeseries chart have their own "formatters" that operate purely based on labels for an individual series for further customization / more control. Some datasources have this built into their query language natively - somewhat similar to [label_replace](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace)

ui/app/sample-data/demoDashboard.ts
sjcobb: I’ve started to prefer `series_name_format` since as it’s a property on the query, it will be to used to format more than just legends (see Shan’s point above), which also addresses one of Luke’s questions as well:

“If this is an option in the query, what does this option mean in the context of things that don't have a legend? Feels weird to have this on the query but have it be used in something that doesn't even have a legend?”

chucknthem: Does the legend property on the TimeSeriesChart also let you format the legend? or is that only for configuring other legend properties?

sjcobb: Only for configuring other legend properties initially, but I have some ideas that I want to prototype soon which would allow for more control and could require additional options.

sjcobb:
Closing since I'm about to open a PR with an initial formatting approach that we can discuss further / iterate on.

PR #601, opened 2022-10-04
Get rid of TimeSeriesQueryRunner, use hooks for running multiple queries

LukeTillman:
This gets rid of the `TimeSeriesQueryRunner` that was being used by the `TimeSeriesChart` panel and instead, just uses a hook to run multiple queries in parallel (powered by `useQueries` from `react-query`). This became possible once we switched those plugins to use a promise-based interface in #556, rather than a hooks-based one. Hopefully this is a lot more straightforward now!

@sjcobb This might mean that if you want to combine the panel/container components, you could do it now. 🎉


ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -12,37 +12,27 @@
...
-    suggestedStepMs: options?.suggestedStepMs,
-    timeRange,
-    variableState,
-    datasourceStore,
-  };
+  const context = useTimeSeriesQueryContext(options);
 
   const key = [definition, context] as const;
-  const { data, isLoading, error } = useQuery(
+  return useQuery(
```
saminzadeh, line 35: Are we committing to the react-query interface being tightly coupled with our plugin system interface? I'm not against it, more of a general question. I personally love it but wonder if we'll regret that in 2 years lol

LukeTillman: Good question--this was kinda a quick fix, but maybe I should be more selective about the return types to hide `react-query` a bit? Was definitely something I was thinking about when I did it. With all the generics involved in the return type, I basically punted. 😂

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -26,18 +25,21 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   } = props;
 
   const suggestedStepMs = useSuggestedStepMs(contentDimensions?.width);
+  const queryResults = useTimeSeriesQueries(queries, { suggestedStepMs });
 
+  if (contentDimensions === undefined) {
+    return null;
+  }
+
+  // TODO: Do we need this container component any more now that we can run multiple queries here?
```
sjcobb, line 34: We shouldn't need it now! This extra component has been bothering me for awhile so excited to get rid of it 🎉

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -54,7 +44,37 @@ export const useTimeSeriesQueryData = (definition: TimeSeriesQueryDefinition, op
...
 };
+
+/**
+ * Runs multiple time series queries using plugins and returns the results.
+ */
+export function useTimeSeriesQueries(definitions: TimeSeriesQueryDefinition[], options?: UseTimeSeriesQueryOptions) {
+  const { getPlugin } = usePluginRegistry();
+  const context = useTimeSeriesQueryContext(options);
+
+  return useQueries({
```
sjcobb, line 56: Huge improvement / way easier to follow now!

ui/plugin-system/src/runtime/time-series-queries.ts
saminzadeh: Yeah maybe selecting a few of the properties like `data` `error` `isFetching` but beats me 🤷 . Their API is so good that I want all the properties haha

PR #604, opened 2022-10-05
create initial panel preview component

sjcobb:
Follow up to #596 
- move preview out of TimeSeriesChartOptionsEditor
- add PanelPreview.tsx component
- currently it is pulled into PanelEditorForm above PanelSpecEditor
  - _I tried a few places and this seemed to work the best but open to other ideas!_
- static height was needed since by default you could only see the panel header

### TODO
- [ ] show preview component without edit action icons in panel header

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
@@ -12,10 +12,11 @@
 // limitations under the License.
 
 import { FormEventHandler, useState } from 'react';
-import { Grid, FormControl, InputLabel, Select, MenuItem, TextField, SelectProps } from '@mui/material';
+import { Grid, FormControl, InputLabel, Select, MenuItem, TextField, SelectProps, Box } from '@mui/material';
 import { ErrorAlert, ErrorBoundary } from '@perses-dev/components';
 import { useLayouts } from '../../context';
 import { PanelEditorValues } from '../../context/DashboardProvider/panel-editing';
+import { Panel, PanelProps } from '../Panel';
 import { usePanelSpecState } from './panel-editor-model';
 import { PanelTypeSelect } from './PanelTypeSelect';
 import { PanelSpecEditor } from './PanelSpecEditor';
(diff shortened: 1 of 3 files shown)


ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
```diff
@@ -96,6 +97,9 @@ export function PanelEditorForm(props: PanelEditorFormProps) {
           </FormControl>
         </Grid>
         <Grid item xs={12}>
+          {kind !== '' && (
```
cndonovan, line 100: Would `kind &&` work?

cndonovan (approved):

cndonovan:
Oops just remembered -- I've been working on adding the query input field, and changing the query will trigger fetching data, and sometimes it's a bad query that returns an error... so I wrapped the preview in an ErrorBoundary!

sjcobb:
> Oops just remembered

Yeah good call! I just moved it inside the same ErrorBoundary as PanelSpecEditor

cndonovan:
Oh wait! There is an error boundary in the new file. NVM!

ui/dashboards/src/components/PanelDrawer/PanelPreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+import { Box } from '@mui/material';
+import { PanelEditorValues } from '../../context/DashboardProvider/panel-editing';
+import { Panel, PanelProps } from '../Panel';
+
+export function PanelPreview({ name, description, kind, spec, groupIndex }: PanelEditorValues) {
+  const previewValues: PanelProps = {
+    definition: {
+      kind: 'Panel',
```
eunicorn, line 21: is it always going to be a `Panel` kind here? if so, do we need to remove `kind` from `PanelEditorValues`?

eunicorn (approved):

LukeTillman (approved):
Overall looks good to me! One minor thing I think you should probably fix inline.

ui/dashboards/src/components/PanelDrawer/PanelPreview.tsx
sjcobb: Yeah that's true, could be a good way to simplify things further!

PR #605, opened 2022-10-05
remove unused panel name in StatChartPanel

sjcobb:
Follow up to discussion in #599, name can be removed completely now. Before it was an optional prop to show the name above the calculated value-- I since refactored StatChart in #462 but forgot to remove it from StatChartPanel


LukeTillman (approved):
Thanks for following up!

PR #607, opened 2022-10-05
Allow listing datasources for a given datasource plugin kind

LukeTillman:
This adds functionality to the `DatasourceStoreProvider` to enable us to list all the datasources of a specific plugin kind (i.e. "give me all the `PrometheusDatasource` instances available"). This is so that while visually editing a time series query or variable definition, we can present the user with a list of datasources to select from. I added a hook to the plugin runtime called `useListDatasources` that should hopefully be helpful when someone is ready to build that UI.

### DIFF
--- a/ui/app/src/model/datasource-api.ts
+++ b/ui/app/src/model/datasource-api.ts
@@ -51,7 +51,7 @@ export function useDatasourceApi(): DatasourceStoreProviderProps['datasourceApi'
 function getProxyUrl(datasource: Datasource | GlobalDatasource) {
   let url = `/proxy`;
   if (datasource.kind === 'Datasource') {
-    url += `/project/${encodeURIComponent(datasource.metadata.project)}`;
+    url += `/projects/${encodeURIComponent(datasource.metadata.project)}`;
   }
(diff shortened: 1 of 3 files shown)


ui/app/src/model/datasource-api.ts
```diff
@@ -24,34 +25,57 @@ export function useDatasourceApi(): DatasourceStoreProviderProps['datasourceApi'
...
+      return [];
+    },
   };
 }
+
+// Helper function for getting a proxy URL from a datasource or global datasource
+function getProxyUrl(datasource: Datasource | GlobalDatasource) {
+  let url = `/proxy`;
+  if (datasource.kind === 'Datasource') {
+    url += `/project/${encodeURIComponent(datasource.metadata.project)}`;
```
Nexucis, line 54:
```suggestion
    url += `/projects/${encodeURIComponent(datasource.metadata.project)}`;
```

Nexucis: otherwise it won't work ;)

Nexucis (approved):
excepting the missing plural, it looks good to me

LukeTillman:
Oh man, the DCO doesn't like that I committed your suggestion from the GitHub UI. 😂

PR #612, opened 2022-10-05
Add PluginSpecEditor to support a query input for TimeSeriesChart

cndonovan:
The goal of this PR is to add a query input to the panel editor.

Originally, I tried to implement this entirely inside of TimeSeriesChartOptionsEditor alone, but that didn't work.

The reason is that the TimeSeriesChart is a panel plugin that is using other plugins -- namely, query plugins. 

Each query plugins need to tell the code that uses it how to edit it, since each query plugin can have different specs. In this PR, we're achieving this by adding `OptionsEditorComponent` and `createInitialOptions()` to the query plugins. 

Additionally, this PR:
* Converted `PanelSpecEditor` to a more generic `PluginSpecEditor` that can be used for editing panel plugins AND query plugins.
* Used `PluginSpecEditor` in both `PanelEditorForm` and `TimeSeriesChartOptionsEditor` to implement the query input. 

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
@@ -13,6 +13,7 @@
 
 import produce from 'immer';
 import { Stack, Typography } from '@mui/material';
+import { UnknownSpec } from '@perses-dev/core';
 import { PluginSpecEditor, OptionsEditorProps } from '@perses-dev/plugin-system';
 import { TimeSeriesChartOptions } from './time-series-chart-model';
 
@@ -22,10 +23,10 @@ export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditor
   const { onChange, value } = props;
   const { queries } = value;
 
-  const handleQueryPluginSpecChange = (index: number, pluginSpec: unknown) => {
+  const handleQueryPluginSpecChange = (index: number, pluginSpec: UnknownSpec) => {
     onChange(
       produce(value, (draft: TimeSeriesChartOptions) => {
-        if (draft.queries[index]?.spec?.plugin?.spec) {
+        if (draft.queries.length && draft.queries[index]?.spec?.plugin?.spec) {
           draft.queries[index].spec.plugin.spec = pluginSpec;
         }
       })
@@ -43,7 +44,7 @@ export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditor
               pluginType="TimeSeriesQuery"
               pluginKind={plugin.kind}
               value={plugin.spec}
-              onChange={(next: unknown) => handleQueryPluginSpecChange(i, next)}
+              onChange={(next: UnknownSpec) => handleQueryPluginSpecChange(i, next)}
             />
           )}
         </>
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
@@ -26,7 +26,10 @@ export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditor
   const handleQueryPluginSpecChange = (index: number, pluginSpec: UnknownSpec) => {
     onChange(
       produce(value, (draft: TimeSeriesChartOptions) => {
-        draft.queries[index].spec.plugin.spec = pluginSpec;
+        const timeSeriesQuery = draft.queries[index];
+        if (timeSeriesQuery) {
+          timeSeriesQuery.spec.plugin.spec = pluginSpec;
+        }
       })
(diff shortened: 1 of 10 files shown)


ui/plugin-system/src/runtime/index.ts
```diff
@@ -12,8 +12,9 @@
 // limitations under the License.
```
sjcobb, line 13: hmm maybe exports need to be adjusted? Seeing `import { OptionsEditorProps, usePlugin } from '@perses-dev/plugin-system';` error in CI

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
```diff
@@ -11,18 +11,43 @@
...
   return (
     <Stack spacing={1}>
-      <Box>{JSON.stringify(value)}</Box>
-      {/* TODO: add form controls to edit panel options */}
+      {/* TODO: Deal with user deleting all queries */}
+      {queries.map(({ spec: { plugin } }, i) => (
+        <>
+          <Typography variant="overline">Query {i + 1}</Typography>
+          {plugin && (
+            <PluginSpecEditor
```
sjcobb, line 42: Wow I'm so excited to play around with this that I actually finally took the time to follow through on [instructions](https://stackoverflow.com/questions/5884784/how-to-pull-remote-branch-from-somebody-elses-repo/42716331#42716331) so I can pull your changes!

ui/prometheus-plugin/src/plugins/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export type PrometheusTimeSeriesQueryEditorProps = OptionsEditorProps<PrometheusTimeSeriesQuerySpec>;
+
+export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQueryEditorProps) {
+  const { onChange, value } = props;
+  const { query } = value;
+
+  const handleQueryChange = (e: ChangeEvent<HTMLInputElement>) => {
+    onChange(
+      produce(value, (draft: PrometheusTimeSeriesQuerySpec) => {
```
saminzadeh, line 28: Is immer tied to the plugin system? or can developers decide how they want to update the state in their plugin and the plugin just think it's just some normal JS value? I think this is cool

LukeTillman: I left a TODO to figure out whether we should expose it as part of the plugin system:

https://github.com/perses/perses/blob/47bc3d1a7d3bfa06e38fd7ed6f2e1812bc89a4ab/ui/dashboards/src/components/PanelDrawer/panel-editor-model.ts#L31-L37

But this is a totally cool way to do it without exposing it. ❤️

ui/plugin-system/src/runtime/index.ts
LukeTillman: Probably some circular imports going on. You probably need to adjust the imports in `PluginSpecEditor` (the stuff it's importing from `@perses-dev/plugin-system`) to be local imports now that it's in the same package as that stuff.

LukeTillman (approved):
Awesome job! This is exactly what I was thinking, and the `immer` usage in the editor is super cool. I left a comment about what I think it going on with the CI failures and imports. LMK if I can help.

ui/plugin-system/src/runtime/index.ts
cndonovan: Ah yeah, let me see if that fixes it

ui/prometheus-plugin/src/plugins/PrometheusTimeSeriesQueryEditor.tsx
cndonovan: This is making me realize that I used `immer` without installing `immer` in either package (`panels-plugin` or `prometheus-plugin`)... but it worked?

I'm guessing it's because they both depend on `plugin-system` and that installs `immer`.

But is that the right way to do things? Normally, if I install some library, I don't expect to be able to use whatever libraries it has installed too...

If I'm doing this, should I install `immer` directly in `panels-plugin` and `prometheus-plugin`?

sjcobb: Yeah good catch, I’d add it in those package.json’s too to be safe

ui/plugin-system/src/runtime/index.ts
sjcobb: Ah yeah I’ve done this before too, we should install that package we use internally, I think: https://www.npmjs.com/package/circular-dependency-plugin

ui/prometheus-plugin/src/plugins/PrometheusTimeSeriesQueryEditor.tsx
cndonovan: Just added to `panels-plugin` and `prometheus-plugin`! (in this PR)

PR #615, opened 2022-10-06
Add visual editing props to all plugins

LukeTillman:
At this point, we know all plugins are going to have at least two props for visual editing (`OptionsEditorComponent` and `createInitialOptions`). This just moves those props into a `Plugin` type and makes all the plugins extend from that type. Variable plugins were the ones that hadn't been updated yet with placeholder values, so I also went ahead and added those placeholders in the `prometheus-plugin` package.


sjcobb (approved):
LGTM!

PR #616, opened 2022-10-06
simplify the cue definitions by hardcoding the plugin struct

Nexucis:
This is PR is here to simplify the Cue definitions following the defined struct of the plugin.

As the definition of the plugin cannot be changed, then we can totally hard coded them which simplify the Cue definition and make them less verbose.


Nexucis:
with this PR we should be good with the panel definitions regarding the Cue schema.
Thanks to #614 we should be good on the backend side.

/cc @sjcobb @LukeTillman

AntoineThebaud (approved):

PR #617, opened 2022-10-07
Delete Panel Group

eunicorn:
image

deleteGroup

### DIFF
--- a/ui/dashboards/src/components/GridLayout/GridTitle.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridTitle.tsx
@@ -40,7 +40,7 @@ export function GridTitle(props: GridTitleProps) {
   const { groupIndex, title, collapse } = props;
 
   const [isHovered, setIsHovered] = useState(false);
-  const { openPanelGroupDialog } = usePanelGroupDialog();
+  const { openPanelGroupDialog, openDeletePanelGroupDialog } = usePanelGroupDialog();
   const { addPanel } = usePanels();
   const { isEditMode } = useEditMode();
   const { layouts, swapPanelGroups } = useLayouts();
@@ -71,24 +71,27 @@ export function GridTitle(props: GridTitleProps) {
           {text}
           {isEditMode && isHovered && (
             <Stack direction="row" sx={{ marginLeft: 'auto' }}>
-              <IconButton onClick={() => addPanel(groupIndex)}>
+              <IconButton aria-label="add panel to group" onClick={() => addPanel(groupIndex)}>
                 <AddIcon />
               </IconButton>
-              <IconButton onClick={() => openPanelGroupDialog(groupIndex)}>
+              <IconButton aria-label="edit group" onClick={() => openPanelGroupDialog(groupIndex)}>
                 <PencilIcon />
               </IconButton>
-              <IconButton>
+              <IconButton aria-label="delete group" onClick={() => openDeletePanelGroupDialog(groupIndex)}>
                 <DeleteIcon />
               </IconButton>
-
               <IconButton
+                aria-label="move group down"
                 disabled={groupIndex === layouts.length - 1}
                 onClick={() => swapPanelGroups(groupIndex, groupIndex + 1)}
               >
                 <ArrowDownIcon />
               </IconButton>
-
-              <IconButton disabled={groupIndex === 0} onClick={() => swapPanelGroups(groupIndex, groupIndex - 1)}>
+              <IconButton
+                aria-label="move group up"
+                disabled={groupIndex === 0}
+                onClick={() => swapPanelGroups(groupIndex, groupIndex - 1)}
+              >
                 <ArrowUpIcon />
               </IconButton>
             </Stack>
--- a/ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
+++ b/ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
@@ -14,7 +14,6 @@
 import { createPanelRef, getPanelKeyFromRef, GridItemDefinition, LayoutDefinition } from '@perses-dev/core';
 import { StateCreator } from 'zustand';
 import { Middleware } from './common';
-import { DashboardStoreState } from './DashboardProvider';
 import { PanelEditorSlice } from './panel-editing-slice';
 
 export interface LayoutSlice {
--- a/ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
+++ b/ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
@@ -14,6 +14,7 @@
 import { createPanelRef, getPanelKeyFromRef, GridItemDefinition, LayoutDefinition } from '@perses-dev/core';
 import { StateCreator } from 'zustand';
 import { Middleware } from './common';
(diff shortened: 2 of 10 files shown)


ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
```diff
@@ -158,6 +164,32 @@ export function createLayoutSlice(layouts: LayoutDefinition[]): StateCreator<Lay
...
     },
+
+    deletePanelGroup(groupIndex) {
+      set((state) => {
+        // remove group from state.layouts
+        const deletedPanelGroup = state.layouts.splice(groupIndex, 1);
+
+        // build an object that maps each panel to the groups it belongs
+        const panels = (get() as DashboardStoreState).panels;
+        const groups = get().layouts;
```
cndonovan, line 175: Curious - why does `panels` need `as DashboardStoreState` but not `groups`?

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
```diff
@@ -190,3 +222,20 @@ function getYForNewRow(group: PanelGroupDefinition) {
   }
   return newRowY;
 }
+
+// Return an object that maps each panel to the groups it belongs
+function mapPanelGroupToPanel(groups: PanelGroupDefinition[]) {
```
cndonovan, line 227: What do you think of `mapPanelToPanelGroups`? Since the mapping is from a panel key to a list of panel group IDs.

cndonovan (approved):

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
eunicorn: because panels live in another slice `panel-editing-slice`

ui/dashboards/src/components/GridLayout/GridTitle.tsx
```diff
@@ -71,24 +71,27 @@ export function GridTitle(props: GridTitleProps) {
           {text}
           {isEditMode && isHovered && (
             <Stack direction="row" sx={{ marginLeft: 'auto' }}>
-              <IconButton onClick={() => addPanel(groupIndex)}>
+              <IconButton aria-label="add panel to group" onClick={() => addPanel(groupIndex)}>
```
sjcobb, line 74: Love that you're thinking about accessibility!

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
```diff
@@ -158,6 +164,32 @@ export function createLayoutSlice(layouts: LayoutDefinition[]): StateCreator<Lay
...
+        const map: Record<string, PanelGroupDefinition['id'][]> = mapPanelGroupToPanel(groups);
+        // for each panel in the deleted panel group, remove panel in state.panels as well
+        const panelsCopy = { ...panels };
+        deletedPanelGroup[0]?.items.forEach((panel) => {
+          const panelKey = getPanelKeyFromRef(panel.content);
+          const groups = map[panelKey];
+          if (groups === undefined) {
+            throw new Error(`Panel is undefined: ${panelKey}`);
+          }
+          // make sure panel is not referenced in another group before deleting it from state.panels
```
sjcobb, line 185: 💥 Good call on thinking about this, the dialog text makes sense to me too

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
```diff
@@ -158,6 +164,32 @@ export function createLayoutSlice(layouts: LayoutDefinition[]): StateCreator<Lay
...
       });
     },
+
+    deletePanelGroup(groupIndex) {
+      set((state) => {
+        // remove group from state.layouts
+        const deletedPanelGroup = state.layouts.splice(groupIndex, 1);
+
+        // build an object that maps each panel to the groups it belongs
+        const panels = (get() as DashboardStoreState).panels;
```
LukeTillman, line 174: You can avoid this cast by changing the `StateCreator` definition above to include other slices in the store that this one needs access to. For example, since this looks like it needs `panels` which (I think?) is in `PanelEditorSlice`:

```ts
StateCreator<LayoutSlice & PanelEditorSlice, Middleware, [], LayoutSlice>
```

If you're curious, I'm pretty sure that the 4 type arguments to `StateCreator` are basically:

1. The overall store's state type that your slice needs to be able to get/set. We could have used `DashboardStoreState` here (for example) if we wanted to give this slice access to EVERYTHING in the store.
2. Any middleware applied to the store (in order its applied). This is, for example, how Typescript knows that `set` should be an `immer` style callback in our store.
3. Store modifiers. I haven't figured out what these are used for yet, but we also don't use any right now.
4. The return type for this slice (i.e. what part of the overall store in type argument 1 is this responsible for creating).

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
```diff
@@ -158,6 +164,32 @@ export function createLayoutSlice(layouts: LayoutDefinition[]): StateCreator<Lay
...
+          const groups = map[panelKey];
+          if (groups === undefined) {
+            throw new Error(`Panel is undefined: ${panelKey}`);
+          }
+          // make sure panel is not referenced in another group before deleting it from state.panels
+          if (groups.length === 1 && groups[0] === deletedPanelGroup[0]?.id) {
+            delete panelsCopy[panelKey];
+          }
+        });
+        (state as DashboardStoreState).panels = panelsCopy;
```
LukeTillman, line 190: Just a general API suggestion: we're (probably?) going to need a `deletePanel` method on the store for deleting an individual panel, right? Why not implement that now on `PanelEditorSlice`, then use that when deleting a group to delete each panel in the group? Feels weird to me to have the `LayoutsSlice` modifying the `panels` state of the `PanelEditorSlice` directly.

ui/dashboards/src/views/ViewDashboard/DashboardApp.tsx
```diff
@@ -44,6 +45,7 @@ export const DashboardApp = (props: DashboardAppProps) => {
         </ErrorBoundary>
         <PanelDrawer />
         {panelGroupDialog && <PanelGroupDialog />}
+        {deletePanelGroupDialog && <DeletePanelGroupDialog />}
```
LukeTillman, line 48: Nit: at some point (doesn't have to be now), I think we're going to want to stop conditionally rendering these dialogs and use their `open` prop so we get proper animation transitions (i.e. conditionally rendering means the dialog just disappears rather than transitioning out).

LukeTillman (approved):
Looks good! I left one store API suggestion that I think we should consider, but if you want to punt, we can always iterate later.

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
eunicorn: I totally agree, as i'm implementing delete panel in another branch, i find myself refactoring this part of the logic. So i think i'll merge this first and address this in the next PR if that's ok!

LukeTillman: Yup, totally cool!

PR #620, opened 2022-10-10
Constrain plugin types to those that use Plugin base

LukeTillman:
This is a small follow-up to #615. In that PR, I added a `Plugin` base type that is meant to be the common props for all plugins in the `plugin-system`. The only thing missing was some kind of "enforcement mechanism" in the type system that ensured that all plugins use it. I tried out a bunch of different ways to do this, but the most straightforward way to add this enforcement mechanism was to change `PluginType` to only include plugin implementations that extend from `Plugin`.

I also realized that this basically means that consumers can now extend the plugin system with their own plugin types. For example, if in my Perses-based application I wanted to add a new plugin type, I could do something like this:

```ts
import { UnknownSpec } from '@perses-dev/core';
import { usePlugin } from '@perses-dev/plugin-system';

declare module '@perses-dev/plugin-system' {
  // Use interface augmentation to add a new plugin type and implementation
  interface SupportedPlugins {
    MyPluginType: MyPlugin;
  }
}

export interface MyPlugin<Spec = UnknownSpec> {
  doSomethingCool: () => Spec;
  createInitialOptions: () => Spec;
  OptionsEditorComponent: () => null;
}

export function useMyPlugin() {
  const { data: plugin } = usePlugin('MyPluginType', 'somekind');
  if (plugin !== undefined) {
    plugin.doSomethingCool();
  }
}
```

If the consumer were to forget to add `createInitialOptions` or `OptionsEditorComponent` (from the `Plugin` base type), then their call to `usePlugin` would fail with a compiler error because `MyPluginType` would not be considered a valid plugin type yet.

### DIFF
--- a/ui/plugin-system/src/model/plugins.ts
+++ b/ui/plugin-system/src/model/plugins.ts
@@ -11,11 +11,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Metadata } from '@perses-dev/core';
+import { Metadata, UnknownSpec } from '@perses-dev/core';
 import { TimeSeriesQueryPlugin } from './time-series-queries';
 import { PanelPlugin } from './panels';
 import { VariablePlugin } from './variables';
 import { DatasourcePlugin } from './datasource';
+import { Plugin } from './plugin-base';
 
 /**
  * Information about a module/package that contains plugins.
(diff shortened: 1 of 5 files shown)


ui/plugin-system/src/model/plugins.ts
```diff
@@ -43,12 +44,17 @@ export interface PluginMetadata {
 }
 
 /**
- * All supported plugin types.
+ * All supported plugin types. A plugin's implementation must extend from `Plugin<UnknownSpec>` to be considered a valid
+ * `PluginType`.
  */
-export type PluginType = keyof SupportedPlugins;
+export type PluginType = {
```
sjcobb, line 50: Your comment, _"I also realized that this basically means that consumers can now extend the plugin system with their own plugin types"_ sounds really cool. 

So in the hypothetical, `kind: 'MonitorTooltip'` or `kind: 'AlertTooltip'`, plugin use case, what would potentially live in our internal codebase vs. in Perses?

sjcobb (approved):
👍 👍

ui/plugin-system/src/model/plugins.ts
LukeTillman: Yeah, basically a plugin could decide "I want people to be able to write plugins for this part of my functionality" and could then define a new plugin type. So in your example, this might be a plugin (like `TimeSeriesChart`) in some other package (like `@perses-dev/panels-plugin`) deciding to make tooltips a plugin without modifying the core plugin system package.

(Note: I think for that _specific_ example we might debate whether to make those plugins part of the core plugin system since they're used by our core panel plugins, but we'd have the option to not do it.)

PR #621, opened 2022-10-10
refactor time-range query param approach

sjcobb:
Query string approach refactored for setting time-range params, summary:
- [use-query-params package](https://github.com/pbeshai/use-query-params) is now used, [use-search-params](https://reactrouter.com/en/main/hooks/use-search-params) alone was too simple
  - changing to use-query-params was proposed by Shan and will also be used for serializing template variables values to query string
- functions added to manipulate start and end params, see: useInitialTimeRange and useSyncTimeRangeParams
- set default time range start param with new useInitialTimeRange
- TimeRangeControls refactored
- selectedTimeRange, setSelectedTimeRange added in DashboardProvider.tsx
- useDefaultTimeRange that previously returned dashboard duration removed in favor of selectedTimeRange

### TODOs
- [x] discuss where QueryParamProvider should go long term 
  - currently it's in render-app.tsx since it was needed in Perses app
  - moving to dashboards package would require refactoring ViewDashboard since DashboardProvider and TimeRangeProvider rely on params when setting initialTimeRange
- [ ] fix Warning `Cannot update a component BrowserRouter while rendering a different component ViewDashboard`
- [ ] additional tests


ui/app/src/render-app.tsx
```diff
@@ -36,12 +38,14 @@ export function renderApp(container: Element | null) {
     <React.StrictMode>
       <BrowserRouter>
         <QueryClientProvider client={queryClient}>
-          <DarkModeContextProvider>
-            <SnackbarProvider anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}>
-              <CssBaseline />
-              <Router />
-            </SnackbarProvider>
-          </DarkModeContextProvider>
+          <QueryParamProvider adapter={ReactRouter6Adapter}>
```
saminzadeh, line 41: I'm wondering if it's better to add `QueryParamProvider` into the `DashboardView` component in the `dashboard` package so all the providers are self contained. I suppose this way is more consistent with `QueryClientProvider` as well. But maybe both should go in there and we can pass optional params:

```tsx
<DashboardView reactQueryClient={queryClient} urlAdapter={ReactRouter6Adapter}>
...
</DashboardView>
```

Maybe in a follow up we can think about this more.

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -22,18 +23,20 @@ const DEFAULT_DASHBOARD_ID = 'node-exporter-full';
  * The View for viewing a Dashboard.
  */
 function ViewDashboard() {
-  const dashboard = useSampleData<DashboardResource>(
-    new URLSearchParams(window.location.search).get('dashboard') || DEFAULT_DASHBOARD_ID
-  );
+  const [query] = useQueryParams({ dashboard: StringParam });
```
saminzadeh, line 26: Nice!

ui/dashboards/src/utils/time-range-params.ts
```diff
@@ -0,0 +1,111 @@
...
+  }, [start, end, dashboardDuration, setQuery]);
+}
+
+/**
+ * Set start and end URL params and update when selected time range changes
+ */
+export function useSyncTimeRangeParams() {
+  const { selectedTimeRange } = useSelectedTimeRange();
+  const [query, setQuery] = useQueryParams(timeRangeQueryConfig);
+  const lastParamSync = useRef<DecodedValueMap<typeof timeRangeQueryConfig>>();
```
saminzadeh, line 68: I'm not sure if I'm following this part. Why do we need to keep track of this?

ui/dashboards/src/utils/time-range-params.ts
```diff
@@ -0,0 +1,111 @@
...
+      };
+    }
+    return initialTimeRange;
+  }, [start, end, dashboardDuration, setQuery]);
+}
+
+/**
+ * Set start and end URL params and update when selected time range changes
+ */
+export function useSyncTimeRangeParams() {
```
saminzadeh, line 65: Wondering if there is a case for when we don't want URL params, so maybe a boolean `enabled` could be nice too. For example, if we have two instances of the DashboardView that need different time ranges

```tsx
useSyncTimeRangeParams(enabled: boolean)
```

saminzadeh (commented):
I like this refactor! Nice work - I had a few questions / comments inline

ui/app/src/render-app.tsx
sjcobb: Yeah that sounds good for follow up! I did try this briefly, but it was going to require moving DashboardProvider, TimeRangeProvider, etc. down a level from that ViewDashboard since they rely on query params to [populate initialTimeRange](https://github.com/perses/perses/blob/33a590cee37a10f1cee92d98348df6417ea268e7/ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx#L39) (and we aren't importing that ViewDashboard internally anymore anyways since we've rebuilt DashboardToolbar).

ui/dashboards/src/utils/time-range-params.ts
sjcobb: There may be a better / less complicated way to handle this, but currently it's needed for the back button to function (see the two `lastStart` conditions: [1](https://github.com/perses/perses/blob/33a590cee37a10f1cee92d98348df6417ea268e7/ui/dashboards/src/utils/time-range-params.ts#L74-L77), [2](https://github.com/perses/perses/blob/33a590cee37a10f1cee92d98348df6417ea268e7/ui/dashboards/src/utils/time-range-params.ts#L83-L88) and then useSyncActiveTimeRange). 

https://github.com/perses/perses/blob/33a590cee37a10f1cee92d98348df6417ea268e7/ui/dashboards/src/utils/time-range-params.ts#L74-L77

https://github.com/perses/perses/blob/33a590cee37a10f1cee92d98348df6417ea268e7/ui/dashboards/src/utils/time-range-params.ts#L83-L88

ui/dashboards/src/utils/time-range-params.ts
sjcobb: Yeah good point, similar to how it was possible to "disable" them before by leaving setQueryString undefined, I'll add this boolean!

ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
```diff
@@ -36,26 +35,13 @@ export interface ViewDashboardProps extends Omit<BoxProps, 'children'> {
...
-    }
-  }, [dashboardDuration, queryString, setQueryString]);
+  const initialTimeRange = useInitialTimeRange(dashboardDuration);
 
   return (
     <DatasourceStoreProvider dashboardResource={dashboardResource} datasourceApi={datasourceApi}>
-      <DashboardProvider initialState={{ dashboardSpec: spec }}>
-        <TimeRangeProvider initialTimeRange={defaultTimeRange}>
+      <DashboardProvider initialState={{ dashboardSpec: spec, selectedTimeRange: initialTimeRange }}>
+        <TimeRangeProvider initialTimeRange={initialTimeRange}>
```
eunicorn, line 44: Can we remove `initialTimeRange` here since it can get it from the store?

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
```diff
@@ -64,7 +66,8 @@ export function DashboardProvider(props: DashboardProviderProps) {
           ...createPanelGroupSlice(...args),
           ...createLayoutSlice(layouts)(...args),
           ...createPanelEditorSlice(panels)(...args),
-          defaultTimeRange: dashboardSpec.duration,
+          selectedTimeRange: selectedTimeRange ?? { pastDuration: dashboardSpec.duration },
```
eunicorn, line 69: Do we need `selectedTimeRange` and `setSelectedTimeRange` in the store, or can we have these states live in the `TimeRangeProvider`?

eunicorn: It's just that now the TimeRangeProvider depends on DashboardProvider so wondering if this is necessary

ui/app/src/render-app.tsx
LukeTillman, line 41: Already digging it has an "adapter" approach...

ui/app/src/views/ViewDashboard.tsx
LukeTillman, line 26: This API already looks so much like our internal implementation of query string params, I'm wondering where it was a couple years ago. 😂

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -39,13 +34,8 @@ export const TIME_OPTIONS: TimeOption[] = [
...
 export function TimeRangeControls() {
   const { timeRange, setTimeRange } = useTimeRange();
-  const dashboardDefaultTimeRange = useDefaultTimeRange();
-  const { queryString } = useQueryString();
-
-  const defaultTimeRange = getDefaultTimeRange(dashboardDefaultTimeRange, queryString);
-
-  // selected form value can be relative or absolute, timeRange from plugin-system is only absolute
-  const [selectedTimeRange, setSelectedTimeRange] = useState<TimeRangeValue>(defaultTimeRange);
+  const { selectedTimeRange } = useSelectedTimeRange();
```
LukeTillman, line 37: I think I mentioned this in the previous PR, but as I'm reading this, I'm a little confused as to what the difference is between `selectedTimeRange` and the `timeRange` directly above this line? `E_TOO_MANY_TIME_RANGES` 😂 

Do we actually need both? Can one just be derived from the other?

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
LukeTillman: Yeah, +1 to this question.

ui/dashboards/src/context/TimeRangeProvider.tsx
```diff
@@ -44,34 +45,24 @@ export function TimeRangeProvider(props: TimeRangeProviderProps) {
...
       if (isRelativeTimeRange(value)) {
-        if (setQueryString) {
-          queryString.set('start', value.pastDuration);
-          // end not required for relative time but may have been set by AbsoluteTimePicker or zoom
-          queryString.delete('end');
-          setQueryString(queryString);
-        } else {
-          setActiveTimeRange(toAbsoluteTimeRange(value));
-        }
+        setActiveTimeRange(toAbsoluteTimeRange(value));
```
LukeTillman, line 53: Does this really need to be a separate piece of state? Looks like it can potentially just be "derived" maybe?

ui/dashboards/src/test/render.tsx
```diff
@@ -22,3 +26,25 @@ export function renderWithContext(ui: React.ReactElement, options?: Omit<RenderO
   const queryClient = new QueryClient({ defaultOptions: { queries: { refetchOnWindowFocus: false, retry: false } } });
   return render(<QueryClientProvider client={queryClient}>{ui}</QueryClientProvider>, options);
 }
+
+/**
+ * Similar to renderWithContext test helper but to test routing / query param functionality
+ * see: https://github.com/pbeshai/use-query-params/blob/master/packages/use-query-params-adapter-react-router-6/src/__tests__/react-router-6.test.tsx
+ */
+export function renderWithHistory(ui: React.ReactElement, options?: Omit<RenderOptions, 'queries'>) {
```
LukeTillman, line 34: If you wanted to combine this into `renderWithContext`, I wouldn't be opposed. Our internal test helper has an option for initial `history` state because it's so common of a need.

ui/dashboards/src/utils/time-range-params.ts
```diff
@@ -0,0 +1,111 @@
...
+  RelativeTimeRange,
+  DurationString,
+  isAbsoluteTimeRange,
+  parseDurationString,
+} from '@perses-dev/core';
+import { useSelectedTimeRange } from '../context';
+
+const timeRangeQueryConfig = {
+  start: StringParam,
+  end: StringParam,
```
LukeTillman, line 32: Rather than `StringParam` should this be some kind of custom param type to avoid some of the conversion logic below?

ui/dashboards/src/utils/time-range-params.ts
LukeTillman: Are we talking about the browser back button? If so, does this work if I (for example) use the back button to go back two (or more) history entries rather than just one in a single click (i.e. click and hold the back button, select an entry farther back in the stack)?

Feels like we should try to figure out what's up and try to make it less complicated if we can. Maybe there's some state-syncing going on somewhere that we can refactor to avoid or something.

ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
LukeTillman, line 44: Feels weird this is getting used/stored in two providers.

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
sjcobb: So I have been gradually simplifying it, in my first PR TimeRangeControls was even harder to follow 😣. I just updated this PR and was able to remove selectedTimeRange! 

This involved refactoring the plugin runtime so the isRelativeTimeRange conversion happens in useTimeRange. Then I created an "internal" version of this API in the dashboards package (see useDashboardTimeRange) that supports relative and absolute time ranges (since the plugin-system version is for [resolved absolute values](https://github.com/perses/perses/pull/530#discussion_r972423337)).

ui/dashboards/src/context/TimeRangeProvider.tsx
sjcobb: This is fixed now 🙌 (after refactoring useTimeRange, see other [comment here](https://github.com/perses/perses/pull/621#discussion_r993889578)). 

I also adjusted TimeRangeProvider's props like we talked about: https://github.com/perses/perses/blob/1b5377230d8a9cc04796019f4c084dc3e574a0ce/ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx#L40-L45

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
sjcobb: Yeah I went back and forth on this, I think I got thrown off when I saw useDefaultTimeRange in dashboard-provider-api, but good point about not wanting to add a dependency to TimeRangeProvider. I've removed selectedTimeRange 
 and setSelectedTimeRange now and updated this PR with hopefully a much better approach!

ui/dashboards/src/utils/time-range-params.ts
sjcobb: So it did work when pressing back multiple times due to the logic in useSyncActiveTimeRange, but agreed this approach was extremely hacky. I updated this PR based on our discussion yesterday so useSyncTimeRangeParams is gone[ in favor of useSetTimeRangeParams ](https://github.com/perses/perses/pull/621/files#diff-6f7f1406b070f8ac3958be3136acc07bf54dc7f7725013108f4813ac8583b903)

ui/dashboards/src/utils/time-range-params.ts
sjcobb: useSyncTimeRangeParams is gone now (see back btn thread), but I did add [this paramsEnabled boolean](https://github.com/perses/perses/blob/1b5377230d8a9cc04796019f4c084dc3e574a0ce/ui/dashboards/src/utils/time-range-params.ts#L119) to useSetTimeRangeParams!

ui/app/src/render-app.tsx
sjcobb: To add to comment above, I wanted to avoid adding QueryParamProvider and QueryClientProvider in dashboards package's ViewDashboard (aka DashboardView) since that would mean duplicating those two providers since they're also needed in the Perses app.

ui/dashboards/src/utils/time-range-params.ts
sjcobb: Yeah good call, this is updated now (see [TimeRangeParam](https://github.com/perses/perses/blob/1b5377230d8a9cc04796019f4c084dc3e574a0ce/ui/dashboards/src/utils/time-range-params.ts#L119)) which simplified things a ton

PR #622, opened 2022-10-10
add circular-dependency-plugin to catch import issues

sjcobb:
This PR adds the circular-dependency-plugin npm package to our app build, which will help catch issues with our imports. An example is a package importing from its own barrel file or circular dependencies between multiple packages. See: https://github.com/aackerman/circular-dependency-plugin

I intentionally created a circular dependency here as an example:
 
intentional_circ_dep_ex


ui/package.json
```diff
@@ -29,12 +29,14 @@
...
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^13.5.0",
+    "@types/circular-dependency-plugin": "^5.0.5",
     "@types/jest": "^27.5.2",
     "@types/node": "^17.0.45",
     "@types/react": "^18.0.0",
     "@types/react-dom": "^18.0.0",
     "@typescript-eslint/eslint-plugin": "^5.30.7",
     "@typescript-eslint/parser": "^5.30.7",
+    "circular-dependency-plugin": "^5.2.2",
```
Nexucis, line 39: doesn't it make more sense to put it in the app package ?

Nexucis (approved):
Besides my comment, good idea to have it !

ui/package.json
LukeTillman: I had to think about it for a minute, but 👍 to this. Since `app` is the only package using `webpack` (and it has the devDependency on webpack right now), probably make sense to put it there for now. If we ever have multiple packages using webpack though, we'll want to move all those dev dependencies out here to the root.

LukeTillman (approved):
Nice. Thanks for the repro screenshot. I was curious whether or not it would be able to pick up issues in our other packages.

sjcobb:
I moved the dependencies to app package.json and merged chagnes from main branch, looks like it caught some potential circular dependencies 😕 
image

Nexucis:
mmm I'm wondering if it is supposed to analyse the circular dependencies in the `dist` folder, and if it is actually a false positive. (I didn't check if what it said is accurate)

I find it weird that we are able to build the UI with no particular issue and it is actually running. If there was an actual circular deps, it wouldn't be possible to run. Or turbo is super strong and manage to compile it anyway ?

LukeTillman:
I'll pull the branch down and check it out. Definitely possible to have circular dependencies that _don't_ break anything if you get lucky with the ordering in the files and such, but I'll see if I can move stuff around to get rid of them. I was thinking as I was working on the plugin stuff that I had some things that were possibly circular in there, so I'm not terribly surprised to see that in the output.

LukeTillman:
All right, think I moved stuff around to correct things. 🤞

PR #623, opened 2022-10-10
Use react-markdown to map HTML elements to MUI components 

cndonovan:
Looking at `0.9.0` in embedded in Chronosphere, the markdown panel looks OK, but could look better...
* Links shouldn't be green
* Table border is too dark
* Probably nicer to have border-radius on table (like other containers in the app)
* Code background is too dark

At this point, just want to be able to take some of the raw HTML elements (e.g. `<a>` tag) and map them to Material UI components (e.g. `<Link>`) because the Material UI components know how to style themselves correctly. 

So, with this PR, I'm changing the markdown-to-HTML libraries. Specifically, using `react-markdown` instead of `marked` so that we can use a feature of `react-markdown` that allows us to map HTML elements to MUI components. 

This should make the panel look better in apps that are using Material UI.

# Light mode
Screen Shot 2022-10-10 at 5 42 00 PM

# Dark mode
Screen Shot 2022-10-10 at 5 42 14 PM

### DIFF
--- a/ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
+++ b/ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
@@ -16,13 +16,13 @@ import ReactMarkdown from 'react-markdown';
 import remarkGfm from 'remark-gfm';
 import rehypeRaw from 'rehype-raw';
 import rehypeSanitize from 'rehype-sanitize';
-import { Box, Link, Paper, Table, TableHead, TableBody, TableRow, TableCell } from '@mui/material';
+import { Box, Link, Paper, Table, TableHead, TableBody, TableRow, TableCell, Theme } from '@mui/material';
 import { PanelProps } from '@perses-dev/plugin-system';
 import { MarkdownPanelOptions } from './markdown-panel-model';
 
 export type MarkdownPanelProps = PanelProps<MarkdownPanelOptions>;
 
-function createMarkdownPanelStyles() {
+function createMarkdownPanelStyles(theme: Theme) {
   return {
     // Make the content scrollable
     height: '100%',
@@ -35,12 +35,12 @@ function createMarkdownPanelStyles() {
     '& code': { fontSize: '0.85em' },
     '& :not(pre) code': {
       padding: '0.2em 0.4em',
-      backgroundColor: '#f5f5f5',
+      backgroundColor: theme.palette.mode === 'light' ? '#f5f5f5' : '#2c2c2c',
       borderRadius: '4px',
     },
     '& pre': {
       padding: '1.2em',
-      backgroundColor: '#f5f5f5',
+      backgroundColor: theme.palette.mode === 'light' ? '#f5f5f5' : '#2c2c2c',
       borderRadius: '4px',
     },
     // <table>
--- a/ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
+++ b/ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
@@ -12,69 +12,82 @@
 // limitations under the License.
 
 import { useMemo } from 'react';
-import * as DOMPurify from 'dompurify';
-import { marked } from 'marked';
-import { Box, Theme } from '@mui/material';
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+import rehypeRaw from 'rehype-raw';
+import rehypeSanitize from 'rehype-sanitize';
+import { Box, Link, Paper, Table, TableHead, TableBody, TableRow, TableCell } from '@mui/material';
 import { PanelProps } from '@perses-dev/plugin-system';
 import { MarkdownPanelOptions } from './markdown-panel-model';
 
 export type MarkdownPanelProps = PanelProps<MarkdownPanelOptions>;
 
-function createMarkdownPanelStyles(theme: Theme) {
+function createMarkdownPanelStyles() {
   return {
     // Make the content scrollable
     height: '100%',
     overflowY: 'auto',
-    // Ignore first margin
+    // Ignore initial margin
     '& :first-child': {
       marginTop: 0,
     },
-    // Styles for <code>
+    // <code>
     '& code': { fontSize: '0.85em' },
     '& :not(pre) code': {
       padding: '0.2em 0.4em',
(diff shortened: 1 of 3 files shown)


ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -12,69 +12,82 @@
...
     '& :not(pre) code': {
       padding: '0.2em 0.4em',
-      backgroundColor: theme.palette.grey[100],
+      backgroundColor: '#f5f5f5',
       borderRadius: '4px',
     },
     '& pre': {
       padding: '1.2em',
-      backgroundColor: theme.palette.grey[100],
+      backgroundColor: '#f5f5f5',
```
cndonovan, line 43: Hard-code the background-color to a nice grey.

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -12,69 +12,82 @@
...
 
-  const html = useMemo(() => convertTextToHTML(text), [text]);
-  const sanitizedHTML = useMemo(() => sanitizeHTML(html), [html]);
+  const memoizedComponent = useMemo(() => {
+    return (
+      <Box sx={createMarkdownPanelStyles}>
+        <ReactMarkdown
+          remarkPlugins={[remarkGfm]}
+          rehypePlugins={[rehypeRaw, rehypeSanitize]}
+          /* eslint-disable @typescript-eslint/no-unused-vars */
```
cndonovan, line 71: In the functions below, I don't want to pass `node` into the MUI components (or we end up with a weird `node` attribute in the actual HTML), so I'm excluding it from `...props`

ESLint complains about that, so I had to disable the linter.

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
eunicorn: does this look ok in dark mode too?

cndonovan: Foiled again by dark mode! 

Just checked, it doesn't look OK in dark mode... which I think is part of why I used `theme.palette` to begin with... but the lightest grey (i.e. grey[100]) is a bit too dark in regular mode...

What a chicken or egg problem

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -25,56 +27,67 @@ function createMarkdownPanelStyles(theme: Theme) {
...
     '& :not(pre) code': {
       padding: '0.2em 0.4em',
-      backgroundColor: theme.palette.grey[100],
+      backgroundColor: theme.palette.mode === 'light' ? '#f5f5f5' : '#2c2c2c',
       borderRadius: '4px',
     },
     '& pre': {
       padding: '1.2em',
-      backgroundColor: theme.palette.grey[100],
+      backgroundColor: theme.palette.mode === 'light' ? '#f5f5f5' : '#2c2c2c',
```
Nexucis, line 43: Isn't a better idea to use the color from the panelette ? Or if the color are not well defined, to add these colors in the palette.
My little knowledge around theming is we shouldn't have hardcoded color, because you have theme that defined for your the color you have to used.
Quite useful when you want to swap from a theme to another. Here for example from light to dark mode. But we could have multiple different dark or light themes with different color contrast. That would be useful for the accessibility for example.

LukeTillman: 👍  to this--should try to use theme tokens here rather than hardcoding colors if possible so that consumers have control over the color via the theme. Totally cool to keep the conditional for light vs dark though.

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
LukeTillman: Ahh, crap. There is some way to fix that. Came up internally and we were able to switch some lint rule config to allow it. I'll follow-up.

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -25,56 +27,67 @@ function createMarkdownPanelStyles(theme: Theme) {
...
+    return (
+      <Box sx={createMarkdownPanelStyles}>
+        <ReactMarkdown
+          remarkPlugins={[remarkGfm]}
+          rehypePlugins={[rehypeRaw, rehypeSanitize]}
+          /* eslint-disable @typescript-eslint/no-unused-vars */
+          components={{
+            a: ({ node, ...props }) => <Link {...props} />,
+            table: ({ node, ...props }) => (
+              <Paper variant="outlined" sx={{ width: 'fit-content' }}>
```
LukeTillman, line 75: MUI has a `TableContainer` element which _might_ be worth looking into to see if that's what you want here (instead of just `Paper`). See https://mui.com/material-ui/react-table/#basic-table

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -25,56 +27,67 @@ function createMarkdownPanelStyles(theme: Theme) {
...
 export function MarkdownPanel(props: MarkdownPanelProps) {
-  const {
-    spec: { text },
-  } = props;
+  const { spec } = props;
+  const { text } = spec;
 
-  const html = useMemo(() => convertTextToHTML(text), [text]);
-  const sanitizedHTML = useMemo(() => sanitizeHTML(html), [html]);
+  const memoizedComponent = useMemo(() => {
```
LukeTillman, line 65: Rather than `useMemo` this should probably be using `React.memo` since you're trying to memoize the component rendering (see https://beta.reactjs.org/apis/react/memo)

LukeTillman (approved):
Overall, love the new approach. I left a few suggestions inline that we should try and address before merge. LMK if I can help.

cndonovan:
FYI this PR will stay open for a bit as I go focus on some `cloud-ui` things, but appreciate the feedback so far. 

Will come back after to address PR feedback!

Nexucis:
any chance to know why you closed this PR. These changes were interesting :)

PR #624, opened 2022-10-10
Add some common UI components for working with the plugin system

LukeTillman:
Now that we're building visual editing components, there are some common things people are going to be doing while editing that deal with the plugin system. This tries to cover at least _some_ of the most common use cases by providing some UI components for editing parts of the JSON that deal with plugins. These initial components are:

- `PluginKindSelect`: this is basically the `PanelTypeSelect` component we already built, but now generic so you can specify the plugin type you want to display a list of options for (e.g. "display a Select with all the Panel plugins available")
- `DatasourceSelect`: displays a Select with a list of the available datasources for a particular kind of Datasource (e.g. "display a Select with all the PrometheusDatasource instances available")
- `PluginEditor`: a combination `PluginKindSelect` and `PluginSpecEditor` component for dealing with `plugin` props in specs. I exported the hook/logic powering that component as `usePluginEditor` separately so that we could wire up the same logic in the `PanelEditorForm` where those inputs are in different places.
- `TimeSeriesQueryEditor`: this is an editor for a `TimeSeriesQueryDefinition` object, which multiple chart options editors are going to need when allowing a user to edit a time series query. We might also want to consider doing a "list" version of this component if running multiple queries becomes common across multiple charts.

I went ahead and integrated these components into a couple of the existing editors that we've already gotten started on. Here's an example screenshot of the Panel editor now:

Screen Shot 2022-10-10 at 3 50 10 PM

I intend to add a few tests to these new components, but wanted to get the PR up now while I'm working on those so that folks working on options editing for individual plugins know about it.


eunicorn (approved):

ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Box, FormControl, FormHelperText, InputLabel } from '@mui/material';
+import { PluginKindSelect } from '../PluginKindSelect';
+import { PluginSpecEditor } from '../PluginSpecEditor';
+import { PluginEditorProps, usePluginEditor } from './plugin-editor-api';
+
+/**
+ * A combination PluginKindSelect and PluginSpecEditor component, meant for editing the common pattern in our specs of
```
eunicorn, line 20: maybe good to explain what the common pattern is here 😛

LukeTillman: Clearly the rest of that comment did not do a good enough job. 🤣 

> common pattern in our specs of a `plugin` property with a `Definition` object attached to it.

I'm basically trying to say "for editing the `plugin` property that's in our JSON a lot and points at a plugin definition". Any suggestions for how to word that better?

sjcobb (approved):
I’m sure I’m going to have questions once I get back into the weeds on viz option editing / legends but this looks great at first glance!

ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
LukeTillman: I gave it a shot trying to provide some more detail in the comments. Hopefully it's a little clearer!

PR #625, opened 2022-10-11
Delete panel

eunicorn:
deletePanel

### DIFF
--- a/ui/dashboards/src/views/ViewDashboard/tests/panelGroups.test.tsx
+++ b/ui/dashboards/src/views/ViewDashboard/tests/panelGroups.test.tsx
@@ -64,18 +64,30 @@ describe('Panel Groups', () => {
     userEvent.click(deleteButton);
 
     const layouts = storeApi.getState().layouts;
-    const deletedPanel = {
-      x: 0,
-      y: 0,
-      width: 12,
-      height: 4,
-      content: { $ref: '#/spec/panels/cpu' },
-    };
+    const deletedPanel = testDashboard.spec.panels['cpu'];
     expect(layouts[0]?.items).toEqual(expect.not.objectContaining(deletedPanel));
 
     const panels = storeApi.getState().panels;
     // should remove cpu from state.panels since it's not used anymore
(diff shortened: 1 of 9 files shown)


ui/dashboards/src/views/ViewDashboard/tests/panelGroups.test.tsx
```diff
@@ -43,6 +43,53 @@ describe('Panel Groups', () => {
     return storeApi;
   };
 
+  beforeEach(() => {
+    const mockIntersectionObserver = jest.fn();
```
LukeTillman, line 47: Oooh, I swear we had to do this at one point (in our internal codebase) and figured out a way to do this globally during test setup (maybe we used a polyfill for it?). We should probably try to figure that out for this package since I'll bet it's going to keep coming up.

ui/dashboards/src/views/ViewDashboard/tests/panelGroups.test.tsx
```diff
@@ -43,6 +43,53 @@ describe('Panel Groups', () => {
...
+    const layouts = storeApi.getState().layouts;
+    const deletedPanel = testDashboard.spec.panels['cpu'];
+    expect(layouts[0]?.items).toEqual(expect.not.objectContaining(deletedPanel));
+
+    const panels = storeApi.getState().panels;
+    // should remove cpu from state.panels since it's not used anymore
+    expect(panels).toEqual(expect.not.objectContaining({ cpu: testDashboard.spec.panels['cpu'] }));
+  });
+
+  it('should only delete panel from panel group if panel is referenced more than once', () => {
```
LukeTillman, line 75: Love that you tested this. We should follow-up during the next meeting to see if there's any way to "disallow" this weirdness in the spec (assuming there isn't a real use case for it) so that we don't have to keep tying ourselves in knots dealing with it.

LukeTillman (approved):
Looks like this turned out pretty good! Glad you figured out the state weirdness. Nothing I see that really needs immediate attention, just a couple "we should figure this out soon" things.

Nexucis (approved):
🥳

PR #627, opened 2022-10-11
[Variables] Visual editing pt. 1

saminzadeh:
Loom video demo: https://www.loom.com/share/98710a81e9614d20a60a52ee4c874963

This PR implements the core of visual editing for template variables. It includes the following:
- Ability to add, edit, and delete variables
- Implements a JSON spec editor for each ListVariables

TODO:
- Add validation to values
- Add testing
- Implement variables editors each each plugin
- Re-order variables 
- UX overhaul to match mocks
   - Confirm dialogs 

image
image


ui/prometheus-plugin/src/plugins/prometheus-variables.tsx
```diff
@@ -51,7 +51,7 @@ export const PrometheusLabelNamesVariable: VariablePlugin<PrometheusLabelNamesVa
     };
   },
   dependsOn: () => [],
-  OptionsEditorComponent: () => null,
+  OptionsEditorComponent: JSONSpecEditor,
```
saminzadeh, line 54: Wondering if this should be the default when a plugin doesn't provide React components

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -177,7 +178,8 @@ function TextVariable({ name }: TemplateVariableProps) {
   const ref = useRef<HTMLInputElement>(null);
   return (
     <TextField
-      ref={ref}
+      inputRef={ref}
+      key={(state?.value as string) ?? ''}
```
saminzadeh, line 182: Since we use defaultValue and ref, this makes sure the component gets the latest value prop. Alternatively, could refactor this to do something similar with useState

ui/prometheus-plugin/src/plugins/prometheus-variables.tsx
sjcobb: Yeah this makes sense to me!

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+
+  const toggleVariableVisibility = (index: number, visible: boolean) => {
+    setVariableDefinitions((draft) => {
+      const v = draft[index];
+      if (!v) {
+        return;
+      }
+      if (!v.spec.display) {
+        v.spec.display = {
+          hidden: false,
```
sjcobb, line 76: We should probably try to make these consistent since there are a few places where things can be enabled/disabled, @cndonovan what did you propose calling this property for queries again?

ui/dashboards/src/components/Variables/Variable.tsx
LukeTillman: Man, I wish MUI played better with uncontrolled components. We had this whole debate internally when figuring out the component library guidelines and ended up settling on "make everything controlled" because MUI's so inconsistent in its support. Would be good if we could try and refactor just for consistency everywhere. Gonna be weird if we're doing it differently in different places based on whether the MUI component we're using supports it.

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
```diff
@@ -0,0 +1,306 @@
+// Copyright 2022 The Perses Authors
```
LukeTillman, line 1: Just a general observation as I'm starting to look at this component: this file is getting pretty big. Do you think there's an opportunity to break it up a bit into sub-components or separate out some of the "model" stuff into its own file(s)?

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+  };
+
+  if (kind === 'TextVariable') {
+    return {
+      kind,
+      spec: {
+        ...commonSpec,
+        ...state.textVariableFields,
+      },
+    } as TextVariableDefinition;
```
LukeTillman, line 83: Do you actually need the casts in here? Just always makes me a little leery and try to avoid if possible since it can hide type errors. One possible alternative (more verbose, but avoids the cast) is to use a temp variable with a type like:

```ts
const textVariable: TextVariableDefinition = {
  // ... insert construction of value here
};
return textVariable;
```

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+  const { getPlugin } = usePluginRegistry();
+  const [state, setState] = useImmer(getInitialState(initialVariableDefinition));
+
+  const updatePluginKind = async (kind: string) => {
+    const p = await getPlugin('Variable', kind);
+    setState((draft) => {
+      draft.listVariableFields.plugin.kind = kind;
+      draft.listVariableFields.plugin.spec = p.createInitialOptions();
+    });
+  };
```
LukeTillman, line 125: Hopefully (maybe?) the components I added in #624 might get rid of some of this logic (or handle it for you). In particular [`PluginEditor`](https://github.com/perses/perses/blob/2db8a47e34b744ceea450c4ebc9fc1c97e3a79cc/ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx) or if you want the behavior with a slightly different UI layout the [`usePluginEditor`](https://github.com/perses/perses/blob/2db8a47e34b744ceea450c4ebc9fc1c97e3a79cc/ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts) hook might be what you want.

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+  onChange: (variableDefinitions: VariableDefinition[]) => void;
+  onCancel: () => void;
+}) {
+  const [variableDefinitions, setVariableDefinitions] = useImmer(props.variableDefinitions);
+  const [variableEditIdx, setVariableEditIdx] = useState<number | null>(null);
+  const isEditingVariable = variableEditIdx !== null;
+
+  const removeVariable = (index: number) => {
+    setVariableDefinitions((draft) => {
+      draft.splice(index, 1);
```
LukeTillman, line 52: It's times like these that I really appreciate `immer` 👍

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+
+  return (
+    <Box p={4}>
+      {isEditingVariable && (
+        <>
+          <Typography variant="h3" mb={2}>
+            Edit Variable
+          </Typography>
+          <VariableEditForm
+            initialVariableDefinition={variableDefinitions[variableEditIdx] as VariableDefinition}
```
LukeTillman, line 91: Same idea as my previous comment: would be cool if we could figure out what's up and if we can improve types of whatever to avoid the cast.

ui/prometheus-plugin/src/plugins/JSONSpecEditor.tsx
```diff
@@ -0,0 +1,50 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { useEffect, useRef, useState } from 'react';
+import { TextField } from '@mui/material';
+import { OptionsEditorProps } from '@perses-dev/plugin-system';
+
+export function JSONSpecEditor<T>(props: OptionsEditorProps<T>) {
```
LukeTillman, line 18: Hahaha, love the `TextField` JSON editor as a placeholder. 😗👌  Well done.

LukeTillman (approved):
Awesome. 💥  I left a few suggestions inline (nothing too major I don't think?), but let me know if you need more info.

You mentioned wanting to add validation and 👍  this is a problem I think we need to address across the framework and plugin code both. Would love if you put that implementation off for now and let's have a discussion as a group to see if there's a library or something we could introduce that people like from past experience. Once we do framework code, then we can figure out if there's a good way to wire it up so it's easy for components in plugins (e.g. options editors) to use as well.

Nexucis:
based on the screen shot in the description, I can see you need to know exactly what is the syntax of the `PrometheusLabelValuesVariable`.
Wouldn't be cooler if you have an input `label_name` instead ? Currently if you don't have the documentation about the variable syntax, it's impossible to guess what you have to define

Nexucis:
Or this is the actual plan behind the TODO point `Implement variables editors each each plugin` ?

ui/dashboards/src/components/Variables/Variable.tsx
saminzadeh: Yep agreed sounds good!

ui/dashboards/src/components/Variables/VariableList.tsx
saminzadeh: Yeah I'm loving immer. TBH I don't see too much value in Zustand alone. I feel like the real value is using Immer then a simple solution like `constate` if you need shared state.

ui/dashboards/src/components/Variables/VariableList.tsx
saminzadeh: Yeah not a bad idea. Although will probably need to do this in a follow up PR since its changing the spec schema.

ui/dashboards/src/components/Variables/VariableList.tsx
saminzadeh: I believe the reason I did this is because its possible the result `variableDefinitions[variableEditIdx] === null` if `variableEditIdx` is not a valid position in the array. However, the way this value gets set is via a button that is derived from the array of definitions. Do you think its safer to save a variable in the top of the component then use that instead of `isEditingVariable`

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
saminzadeh: Yeah good call, I'll pull out the functions

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
saminzadeh: I'll take a look at those and see if they make sense to integrate

ui/prometheus-plugin/src/plugins/JSONSpecEditor.tsx
saminzadeh: :) Further enhancements might be optional props like the ones below:

```tsx
<JSONSpecEditor 
   jsonSchema={someSchema} 
   links={[title: "Docs", url: "..." ]} 
   samples={snippets}
/>

```

saminzadeh:
> Or this is the actual plan behind the TODO point `Implement variables editors each each plugin` ?

Yes, exactly. The JSON editor is just a placeholder right now. See [this comment](https://github.com/perses/perses/pull/627#discussion_r993470920) for enhancements to that component

ui/dashboards/src/components/Variables/VariableList.tsx
Nexucis: for queries, the proposition from @cndonovan was to call it `is_enabled` because hidding a result of query make it unnecessary to run it.
That's not the same behaviour for the variable. You can hide it, but you still want to run it because you are using it in other variable that are displayed.

Nexucis: Also, as with @AntoineThebaud we are currently implementing the support of the variable in the backend, it would be nice to have the current format of the variable with the different possible field, just to be sync between the FE and the BE.
That would reduce the amount of issues we will have when connecting the FE to the BE to store and retrieve the dashboard.

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+      draft.splice(index, 1);
+    });
+  };
+
+  const addVariable = () => {
+    setVariableDefinitions((draft) => {
+      draft.push({
+        kind: 'TextVariable',
+        spec: {
+          name: 'New Variable',
```
Nexucis, line 61: The name is supposed to be the thing used to reference the variable in other variable or in the different panels.

So as it's more an ID than just a name, spaces are forbidden.
That's why like for the panel, you have a struct `Display` that contains the name to be displayed.
If `display` is not set, then the name to be displayed is the `ID` of the variable

saminzadeh: Good call - the follow up PR addresses validation

ui/dashboards/src/components/Variables/VariableList.tsx
LukeTillman: Yeah, sounds like if there's just a way to structure this so you can do a `null` check first somewhere (and handle that case, even if it's a `throw`), that would be the way to try and go.

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+    </Box>
+  );
+}
 
 export function TemplateVariableList() {
   const [isEditing, setIsEditing] = useState(false);
   const variableDefinitions = useTemplateVariableDefinitions();
   const { isEditMode } = useEditMode();
+  const [showVariablesInEditMode, setShowVariablesInEditMode] = useState(true);
+  const showVariables = isEditMode ? showVariablesInEditMode : true;
```
eunicorn, line 181: do we need `showVariables`? seems like we can derive it from `showVariablesInEditMode`

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+              }}
+            >
+              Apply Changes
+            </Button>
+          </Stack>
+          <Typography variant="h3" mb={2}>
+            Variable List
+          </Typography>
+          <Stack spacing={2}>
+            <TableContainer component={Paper}>
```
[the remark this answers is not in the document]
saminzadeh: Yeah that's my plan next. I'll pick your brain on this

ui/dashboards/src/components/Variables/VariableList.tsx
saminzadeh: The value of showVariables is always true when it's not an editMode. The `showVariablesInEditMode` is only respected when its in edit mode

eunicorn: oh if it's always true, then why do we need it?

eunicorn: oh got it, it was confusing at first glance haha

eunicorn (approved):

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -24,51 +24,197 @@ import {
...
+import { useImmer } from 'use-immer';
+import EyeIcon from 'mdi-material-ui/Eye';
+import PencilIcon from 'mdi-material-ui/Pencil';
+import TrashIcon from 'mdi-material-ui/TrashCan';
+
+import { useTemplateVariableDefinitions, useEditMode, useTemplateVariableActions } from '../../context';
 import { TemplateVariable } from './Variable';
+import { VariableEditForm } from './VariableEditorForm';
+
+function VariableEditor(props: {
```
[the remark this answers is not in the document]
saminzadeh: Cool sounds good 

@LukeTillman @eunicorn On a more general note, what do we consider long? Maybe we can just add the [max-lines eslint rule ](https://eslint.org/docs/latest/rules/max-lines) (seems to default to 300) if we have a general rule. Also, do we prefer one component per file? Personal preference is to avoid [too many files](https://twitter.com/dan_abramov/status/1220029098543788032?lang=en) (😆) but happy to adopt that as well if that's the general consensus.

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
saminzadeh: Integrated PluginEditor, it was pretty helpful. Had to make a few tweaks to styling, let me know if you think thats ok

ui/dashboards/src/components/Variables/VariableList.tsx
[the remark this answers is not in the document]
eunicorn: I personally prefer one component per file and name it the same as the component so it's easy to see/find what components there are, but I think it's ok that the file is long if the component is huge

ui/dashboards/src/components/Variables/VariableEditorForm.tsx
LukeTillman: Totally cool on styling tweaks, I just basically tried to get something functional in there.

ui/dashboards/src/components/Variables/VariableList.tsx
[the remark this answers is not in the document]
LukeTillman: Yeah, I'm the same on preferring one component for file unless the second component is pretty small (e.g. a styled component, a really short helper component that's 20ish lines, etc). I just find it easier to read and jump around to other components in new tabs without losing my place in the original one if I want to go back (I also probably suck at navigating in the IDE tricks 😂).

My "this is getting pretty long I should split this up so others can read it" alarm tends to start warning me around 150-200 lines, but that's totally subjective. 🤷‍♂️

PR #630, opened 2022-10-12
Release v0.10

eunicorn:
Signed-off-by: Eunice Wong <eunicewong@live.com>


Nexucis:
did you run the script `make bump-version` @eunicorn. Just to know if there is a bug with it

PR #632, opened 2022-10-12
Fix release title and propose a template title in the release documentation

Nexucis:
I just realised that the script we are using to generate the changelog that has to be pushed in the github release is catching everything between two title starting with two `#`.

This PR aims to aligne the different release title in the changelog, and to add a template title in the release documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-# 0.10.0 / 2022-10-12
+## 0.10.0 / 2022-10-12
 
 - [FEATURE] Implement new datasource format #570
 - [FEATURE] Add PluginSpecEditor to support a query input and static panel options for TimeSeriesChart #596, #612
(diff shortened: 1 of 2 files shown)


eunicorn (approved):

sjcobb:
So this would probably break the current script but a future idea that I think would help with getting releases out quicker is we could keep an “Unreleased” section at the top of CHANGELOG.md, then as we open PRs into main, we could gradually keep it up to date. Once it’s time to do a release, it would start with copying the items in Unreleased to the new version heading (plus some manual work to consolidate and clean it up, but overall this would save a lot of time for the person doing the release). Got this idea from here btw: https://keepachangelog.com/en/1.0.0/#effort

sjcobb (approved):
LGTM!

Nexucis:
> So this would probably break the current script but a future idea that I think would help with getting releases out quicker is we could keep an “Unreleased” section at the top of CHANGELOG.md, then as we open PRs into main, we could gradually keep it up to date. Once it’s time to do a release, it would start with copying the items in Unreleased to the new version heading (plus some manual work to consolidate and clean it up, but overall this would save a lot of time for the person doing the release). Got this idea from here btw: https://keepachangelog.com/en/1.0.0/#effort

I didn't get why it would break the script that generate the changelog :).

I'm totally fine with this approach. We could do something like Thanos: https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#unreleased

Another idea that followup could be we don't add everything to the changelog. Just the things that impact the user. (That's also what Thanos or Prometheus is doing)

I'm not entirely convinced by this approach (the one that add only what impact the user), but I'm curious about what all of you think about it.

PR #634, opened 2022-10-13
PanelGroupDialog refactor to need less state, look more like PanelDrawer

LukeTillman:
Following up on #633, this continues refactoring panel groups. This time it's the dialog so it basically follows the same pattern as the `PanelDrawer` where it exposes all the state the editor needs on open (including how to save or close without making changes). This was basically trying to make it so the editor dialog didn't need the state of _all_ panel groups just to do its work, as well as to try and be specific about what it _could_ edit (since that editor wasn't editing things like the items in the group). Hiding the "items in the group" from the editor is going to be more important as I go to split a group's items out so that drag and drop state is separate from the rest of the group's state.

Again, I left a few TODOs in here because there's more to come! 🥳 🙃 

### DIFF
--- a/ui/dashboards/src/components/PanelGroupDialog/PanelGroupDialog.test.tsx
+++ b/ui/dashboards/src/components/PanelGroupDialog/PanelGroupDialog.test.tsx
@@ -42,7 +42,7 @@ describe('Add Panel Group', () => {
     const storeApi = renderDialog();
 
     // Open the dialog for a new panel group
-    act(() => storeApi.getState().openPanelGroupDialog());
+    act(() => storeApi.getState().addPanelGroup());
 
     const nameInput = await screen.findByLabelText(/Name/);
     userEvent.type(nameInput, 'New Panel Group');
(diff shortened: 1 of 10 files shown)


ui/dashboards/src/components/PanelGroupDialog/PanelGroupDialog.test.tsx
```diff
@@ -42,7 +42,7 @@ describe('Add Panel Group', () => {
     const storeApi = renderDialog();
 
     // Open the dialog for a new panel group
-    act(() => storeApi.getState().openPanelGroupDialog());
+    act(() => storeApi.getState().addPanelGroup());
```
eunicorn, line 45: thinking we should figure out how to open the dialog without using the store api here

eunicorn (approved):

ui/dashboards/src/components/PanelGroupDialog/PanelGroupDialog.test.tsx
LukeTillman: Yeah, I was thinking we should figure out if there's a way to make these test work without manipulating the store directly.

PR #637, opened 2022-10-13
improve linter by supporting datasources schema and online validation

Nexucis:
With this PR, maybe you will more understand the need of the package `validate` @AntoineThebaud :).

I'm also wondering if we should provide an endpoint in the API to be able to download the different schemas.
WDYT @AntoineThebaud @chucknthem @LukeTillman ? 

### DIFF
--- a/internal/cli/cmd/lint/lint.go
+++ b/internal/cli/cmd/lint/lint.go
@@ -17,13 +17,15 @@ import (
 	"fmt"
 	"io"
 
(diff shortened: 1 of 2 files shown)


AntoineThebaud (approved):

Nexucis:
merging this for the moment, If you have any thought around this subject @LukeTillman @chucknthem @AntoineThebaud don't hesitate to yell :)

PR #644, opened 2022-10-14
remove extra component in TimeSeriesChart panel

sjcobb:
As a follow-up to the [discussion here](https://github.com/perses/perses/pull/601#discussion_r987899409), this PR is a refactor of our TimeSeriesChart panel to remove the unnecessary container component. 

Previously, since GraphQueryRunner passed data via context, but the chart needs it as a prop--I had to add an extra TimeSeriesChartContainer.tsx component as a workaround. This wasn't necessary in StatChart and GaugeChart since those panels only support one query.  Since then in #601, the TimeSeriesQueryRunner approach was removed in favor of react-query's [useQueries](https://tanstack.com/query/v4/docs/reference/useQueries) (which was possible after PR #556)


sjcobb:
Closing in favor of upcoming PR that also includes legend updates

Issue #643, opened 2022-10-14
On mobile, the header doesn't have enough space


Nexucis:
On mobile, the header where you have the different button doesn't have enough space to exist. 

I think we can safely remove the edit button. You likely don't want to edit a dashboard on a so small screen

![image](https://user-images.githubusercontent.com/4548045/195871467-5fe09dbc-b8d2-482d-bd5d-b0b0f4017fa3.png)

Nexucis:
I think we are good for the moment. Removing different buttons helped somehow 🤣

PR #646, opened 2022-10-14
remove edit button on mobile

Nexucis:
When you are on mobile, I think it's fair you cannot edit the dashboard.

Let me know what you think about :)

It's part of the issue #643 

### DIFF
--- a/ui/dashboards/src/components/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar.tsx
@@ -81,7 +81,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
             <Typography variant="h2">{dashboardName}</Typography>
             <Stack direction="row" spacing={2} sx={{ marginLeft: 'auto' }}>
               <TimeRangeControls />
-              {isLaptopSize ? (
+              {isLaptopSize && (
                 <Button
                   variant="outlined"
                   startIcon={<PencilIcon />}
@@ -90,8 +90,6 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
                 >
                   Edit
                 </Button>
-              ) : (
-                <></>
               )}
             </Stack>
           </Box>
--- a/ui/dashboards/src/components/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Typography, Stack, Button, Box } from '@mui/material';
+import { Typography, Stack, Button, Box, useTheme, useMediaQuery } from '@mui/material';
 import PencilIcon from 'mdi-material-ui/PencilOutline';
 import AddPanelGroupIcon from 'mdi-material-ui/PlusBoxOutline';
 import AddPanelIcon from 'mdi-material-ui/ChartBoxPlusOutline';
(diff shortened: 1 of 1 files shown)


ui/dashboards/src/components/DashboardToolbar.tsx
```diff
@@ -28,6 +28,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
 
   const { isEditMode, setEditMode } = useEditMode();
   const { addPanelGroup, addPanel } = useDashboardActions();
+  const isLaptopSize = useMediaQuery(useTheme().breakpoints.up('sm'));
```
Nexucis, line 31: I am wondering if it makes the create a wrapper around that. So it's even more easier to use it, even if it's already super easy to use !

ui/dashboards/src/components/DashboardToolbar.tsx
```diff
@@ -80,14 +81,18 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
...
               <TimeRangeControls />
-              <Button
-                variant="outlined"
-                startIcon={<PencilIcon />}
-                onClick={onEditButtonClick}
-                sx={{ marginLeft: 'auto' }}
-              >
-                Edit
-              </Button>
+              {isLaptopSize ? (
```
Nexucis, line 84: probably here there is a better way to write it. I found a bit ugly like that, specially the `else` statement that returns an empty bracket.
I'm pretty sure there is way to avoid it, but I don't remember how 😅

cndonovan (commented):
This seems reasonable to me! 

At some point, I could see us having a "long-term vision" discussion to clarify: do we believe we will always disable editing on mobile? or, do we believe we will make it possible to edit on mobile someday (or maybe make it possible for other developers to do it if they want to)?

But personally I think it makes sense to disable it for now... and maybe never enable it :) 

At my previous company, we had a banner that simply said "we don't support the screen size" when it was below a certain screen size, and that applied to both viewing and editing dashboards. 

Grafana does seem to support editing on mobile, but to do the same we would need to also invest in the extra designs + engineering work for that.

eunicorn (approved):

ui/dashboards/src/components/DashboardToolbar.tsx
eunicorn: maybe instead of completely removing the edit button, we should disable it and have a tooltip that tells users that we do not support edit mode for mobiles

Nexucis:
Thank you for your review and thoughts @cndonovan @eunicorn !

> maybe instead of completely removing the edit button, we should disable it and have a tooltip that tells users that we do not support edit mode for mobiles

Actually, it was convenient for me to entirely remove it because then the header has more space. See the screenshot in #643. 

> At some point, I could see us having a "long-term vision" discussion to clarify: do we believe we will always disable editing on mobile? or, do we believe we will make it possible to edit on mobile someday (or maybe make it possible for other developers to do it if they want to)?

Of course we can have and probably we should have this `long-term vision` discussion ! There is maybe more to say than just do we enable the edit functionality on mobile.

> But personally I think it makes sense to disable it for now... and maybe never enable it :)

Totally agree with you !
My vision around this topic is: If you are going to use Perses on your mobile, it's certainly during an incident, someone shared a dashboard and you want to look at it quickly. So you are not in the mood to edit it. If you want to edit it, you would be much more confortable to do it on a large screen !

ui/dashboards/src/components/DashboardToolbar.tsx
Nexucis: ah yeah perfect. Thank you @cndonovan !

sjcobb (approved):

PR #647, opened 2022-10-14
Add variable plugin editors

saminzadeh:
Loom demo: https://www.loom.com/share/0dfd6ccf99ce419d9e270d2ab5c7010a

This adds in the Variable plugin editors for 3 plugin types: StaticList, PromLabelValues, PromLabelNames

Eventually, the Series Selectors will be replaced with [CodeMirror](https://codemirror.net/) for syntax highlighting and autocomplete.

### DIFF
--- a/ui/prometheus-plugin/src/plugins/MatcherEditor.tsx
+++ b/ui/prometheus-plugin/src/plugins/MatcherEditor.tsx
@@ -1,0 +1,60 @@
+// Copyright 2022 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { useState } from 'react';
+import { Stack, TextField, Button, Box, IconButton } from '@mui/material';
+import TrashIcon from 'mdi-material-ui/TrashCan';
+
+export function MatcherEditor(props: { initialMatchers?: string[]; onChange: (matchers: string[]) => void }) {
+  const [matchers, setMatchers] = useState<string[]>(props.initialMatchers ?? []);
+  return (
+    <Stack spacing={1}>
+      {matchers.map((matcher, index) => (
+        <Box key={index} display="flex">
+          <TextField
+            onBlur={() => {
+              props.onChange(matchers);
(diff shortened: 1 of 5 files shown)


ui/prometheus-plugin/src/plugins/MatcherEditor.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [matchers, setMatchers] = useState<string[]>(props.initialMatchers ?? []);
+  return (
+    <Stack spacing={1}>
+      {matchers.map((matcher, index) => (
+        <Box key={index} display="flex">
+          <TextField
+            onBlur={() => {
+              props.onChange(matchers);
+            }}
+            label="Series Selector"
```
saminzadeh, line 28: Is there a good name for this?

Nexucis: Perhaps `Label Matchers`, but I'm not sure ever if it's a good one. I'm based on what is written here: https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers

@juliusv would you have any thought around this topic ? We are trying to figure out what name should be used to talk about the value you have in the query parameter `match[]`

Nexucis (approved):
minus the name used for `match[]`, it looks good to me !
And thank you for the demo ! It's cool to see it in action.

ui/prometheus-plugin/src/plugins/MatcherEditor.tsx
LukeTillman: Yeah, you matched what the API docs call it right now, so seems OK to me. But open to suggestions from Prom folks like Julius.

LukeTillman (approved):
🚀

ui/prometheus-plugin/src/plugins/MatcherEditor.tsx
juliusv: The terminology that has evolved in Prometheus is:

* The entire selector is called an "instant vector selector" / "range vector selector". But just "series selector" is fine too when it's about metadata queries (since you are not selecting actual values into an instant/range vector), I use that as well.
* The part in the curly braces after the metric name are the "label matchers" (each individual one is one "label matcher").

However, Prometheus has not always been 100% consistent on this (that happens sometimes in an organically growing OSS project with not everyone coordinating all the time super tightly around terminology, but we should still try). One example is the series metadata API: https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers. The HTTP parameter is called `match[]` here, with each `match` argument actually being a full series selector. If you wanted to be super consistent maybe `selector[]` would have been a better name.

juliusv: TL;DR: I think "series selector" is totally fine here if it's about the whole selector, not just label matchers.

sjcobb (approved):

PR #654, opened 2022-10-17
Migrate variables management to cuelang

AntoineThebaud:
As for panels, queries & datasources previously, variables are now managed with CUE & plugins can be added at runtime.


Nexucis (approved):
nice looks good to me ! 

Cool to see it's smooth to add a new plugin !

Looks like we are ready to connect the FE to the BE

/cc @LukeTillman @sjcobb @eunicorn

PR #659, opened 2022-10-18
provide kind and default query parameter to filter the list of datasources

Nexucis:
I'm adding two new query parameters `kind` and `default` when getting the list of datasources or global datasources.

For example:

```bash
curl -XGET http://localhost:8080/api/v1/globaldatasources?kind=PrometheusDatasource&default=true
```

This kind of request will return at most one element in the array.

/cc @LukeTillman 

### DIFF
--- a/pkg/model/api/v1/datasource.go
+++ b/pkg/model/api/v1/datasource.go
@@ -20,6 +20,23 @@ import (
 	"github.com/perses/perses/pkg/model/api/v1/common"
 )
 
+func FilterDatasource[T DatasourceInterface](kind string, defaultDTS *bool, list []T) []T {
+	result := make([]T, 0, len(list))
+	for _, d := range list {
+		keep := true
+		if len(kind) > 0 && kind != d.GetSpec().Plugin.Kind {
+			keep = false
+		}
+		if defaultDTS != nil && *defaultDTS != d.GetSpec().Default {
+			keep = false
+		}
+		if keep {
+			result = append(result, d)
+		}
+	}
+	return result
+}
+
 func GenerateGlobalDatasourceID(name string) string {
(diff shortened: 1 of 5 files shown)


pkg/model/api/v1/datasource.go
```diff
@@ -20,6 +20,23 @@ import (
 	"github.com/perses/perses/pkg/model/api/v1/common"
 )
 
+func FilterDatasource[T DatasourceInterface](kind string, defaultDTS *bool, list []T) []T {
+	result := make([]T, 0, len(list))
+	for _, d := range list {
+		keep := true
+		if len(kind) > 0 && kind != d.GetSpec().Plugin.Kind {
+			keep = false
+		}
```
AntoineThebaud, line 29: else if instead ? no need to go through the second one if keep is already set to `false`

Nexucis: I could even merge the conditions together.
Like 

```go
              if (len(kind) > 0 && kind != d.GetSpec().Plugin.Kind) && 
                  defaultDTS != nil && *defaultDTS != d.GetSpec().Default {
			keep = false
		}
```

WDYT ?

AntoineThebaud: What about

```go
for _, d := range list {
    if (len(kind) > 0 && kind != d.GetSpec().Plugin.Kind) || (defaultDTS != nil && *defaultDTS != d.GetSpec().Default) {
        continue
    }
    result = append(result, d)
}
```

?

Nexucis: even better. Thank you !

AntoineThebaud (approved):

PR #638, opened 2022-10-13
Initial compact legend component (sumeete fork)

sjcobb:
Add ListLegend.tsx and ListLegendItem.tsx in @perses-dev/components package for use in TimeSeriesChart panel

Original branch here: https://github.com/sumeete/perses/tree/sumeete/listlegend
- updated with signoff, sync with upstream, removed panel changes since I'll handle those in a follow up PR

PR #660, opened 2022-10-18
TimeSeriesChart panel refactor and replace echarts legend with custom

sjcobb:
Follow up to #638 with TimeSeriesChart panel updates to integrate the legend component. This involves a breaking change to the current legend options, `show_legend` is now `legend.show` and `legend.position` is also supported, ex:

```
legend: {
  show: true, // false is default
  position: 'bottom', // 'bottom' is default, 'right' also supported
}
```

PR also includes refactoring from #644 to remove the extra TimeSeriesChartContainer component in TimeSeriesChart panel

### TODOs
- [ ] add tests (_working on this now_)
- [ ] legend series name formatting
- [ ] styling / refinements to legend display and editor
- [ ] multi-select legend items on Shift-click

### Screenshots
Position option ex:
legend_screenshot

Selected legend item:
selected_legend_item

#### Recording
https://user-images.githubusercontent.com/9369625/196469482-312a9d8a-5dd3-48b2-a8c3-55e54a16d290.mp4


Nexucis:
Would be great to add these new types in Cue ;).

sjcobb:
> Would be great to add these new types in Cue ;).

Yeah definitely, I'll work on that now!

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
```diff
@@ -72,7 +72,7 @@ export function PanelEditorForm(props: PanelEditorFormProps) {
 
   return (
     <form id={panelEditorFormId} onSubmit={handleSubmit}>
-      <Grid container spacing={2}>
+      <Grid container spacing={2} sx={{ overflowY: 'scroll', maxHeight: '90vh' }}>
```
sjcobb, line 75: This is temporary, I added it so the user can scroll inside the edit panel drawer, can probably remove once I add separate tabs for 'Query' and 'Visualization' options

cndonovan: Oh yeah I just added almost the same thing in a different PR (added it on the `form` element). We will need some kind of scrolling I believe, because I was running into the case of when a user adds lots of queries. I think I like this approach better (having it on `<Grid container>`)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
```diff
@@ -42,6 +75,41 @@ export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditor
...
+            value={value.legend.position ?? DEFAULT_LEGEND.position}
+            onChange={(e) => {
+              updateLegendPosition(e.target.value as LegendPositionOptions['position']);
+            }}
+          >
+            {LEGEND_POSITIONS.map((v) => (
+              <MenuItem key={v} value={v}>
+                {v}
+              </MenuItem>
+            ))}
```
cndonovan, line 107: v!

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -11,35 +11,167 @@
...
     contentDimensions,
   } = props;
 
+  // popuate 'position' and 'show' if undefined
+  const legend = merge({}, DEFAULT_LEGEND, props.spec.legend);
+
+  const unit = props.spec.unit ?? {
+    kind: 'Decimal',
+    decimal_places: 2,
+  };
```
cndonovan, line 40: Could do DEFAULT_UNIT too (I like how you did DEFAULT_LEGEND a lot), but really whatevs / potato potato

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -11,35 +11,167 @@
...
+    });
+  };
+
+  // populate series data based on query results
+  const { graphData, loading } = useMemo(() => {
+    const timeScale = getCommonTimeScale(queryResults);
+    if (timeScale === undefined) {
+      for (const query of queryResults) {
+        // does not show error message if any query is successful (due to timeScale check)
+        if (query.error) throw query.error;
```
cndonovan, line 65: I think I'm missing something, but this reads to me that we WILL throw an error if any query is unsuccessful (even if another query is successful)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -11,35 +11,167 @@
...
 
-  // TODO: Do we need this container component any more now that we can run multiple queries here?
+  if (loading === true) {
+    return (
+      <Box
+        sx={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}
+        width={contentDimensions.width}
+        height={contentDimensions.height}
+      >
+        <Skeleton variant="text" width={contentDimensions.width - 20} height={contentDimensions.height / 2} />
```
cndonovan, line 141: Love the skeleton component

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -11,35 +11,167 @@
...
+        graphData.timeSeries.push(thresholdLineSeries);
+      });
+    }
+
+    return {
+      graphData,
+      loading: queriesFinished === 0,
+      allQueriesLoaded: queriesFinished === queryResults.length,
+    };
+  }, [queryResults, thresholds, selectedSeriesName, legend]);
```
cndonovan, line 128: If it's snappy enough, we might not need to worry about this optimization:

But I wonder if eventually we'll want to separate: processing that we do when we initially load data, and processing that we dod when we change the selectedSeriesName (because seems like we don't need to do as much math when we just change the selected series))

cndonovan (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: So I did this to fix error handling awhile back in #479, but there might be a way to refactor this after all the new useQueries changes in #601 to be more intuitive 🤔 

That `if (timeScale === undefined)` line is only reason it works since [timeScale will be defined](https://github.com/perses/perses/blob/3723d62657d9cfa83e9ce00f0d26d59284873d1d/ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts#L42) if any single query returns successful

ui/app/sample-data/demoDashboard.ts
```diff
@@ -258,7 +271,9 @@ const demoDashboard: DashboardResource = {
                   },
                 },
               ],
-              show_legend: false,
+              // legend: {
+              //   show: true,
+              // },
```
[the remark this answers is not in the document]
sjcobb: Yep that's right! Was testing a bunch of variations when I added the merge DEFAULT_LEGEND line, just cleaned up this file a little bit

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Looking at it again, I can probably pull out the thresholds logic but I'm not sure how separating out selectedSeriesName would work 🤔

The way I'm thinking about it currently, when a legend item is clicked, it does need to reprocess what gets [pushed into graphData.timeSeries](https://github.com/perses/perses/blob/a9962cbeaab6744701d0e7fefefae59d407a5aca/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx#L88-L91) as well so that only the correct selected series line shows and for the legend and tooltip to stay in sync (that's actually what is broken about the current ECharts legend, where clicking an item does not also filter the series data that gets pushed into our custom tooltip)

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -20,11 +20,22 @@ import { ThresholdOptions } from '../../model/thresholds';
...
-  show_legend?: boolean;
+  legend: LegendOptions;
   unit?: UnitOptions;
   thresholds?: ThresholdOptions;
 }
 
+export const DEFAULT_LEGEND: LegendOptions = {
+  show: false,
+  position: 'bottom',
+};
```
LukeTillman, line 31: One possible way to make this more compact (less JSON! 🎉 ): remove the `show` boolean and make `legend` optional. Basically you can infer `show` based on whether you have a legend options object like the `position`. (This assumes there's no point in keeping legend options around if it's not being shown.)

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -20,11 +20,22 @@ import { ThresholdOptions } from '../../model/thresholds';
...
+
+export const LEGEND_POSITIONS = ['bottom', 'right'] as const;
+
+export type LegendPositionOptions = {
+  position: typeof LEGEND_POSITIONS[number];
```
LukeTillman, line 36: You might consider aliasing this `typeof` type, cause I could see it being useful:

```ts
export type LegendPosition = typeof LEGEND_POSITIONS[number];
```

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
```diff
@@ -31,6 +46,22 @@ export function TimeSeriesChartOptionsEditor(props: TimeSeriesChartOptionsEditor
...
 
+  function updateLegendShow(show: boolean) {
+    onChange(
+      produce(value, (draft: TimeSeriesChartOptions) => {
+        draft.legend.show = show;
+      })
+    );
+  }
+
+  function updateLegendPosition(position: LegendPositionOptions['position']) {
```
LukeTillman, line 57: Tip: you might be able to avoid the cast below (i.e. `e.target.value as LegendPositionOptions['position']`) if you type this event handler using `SelectProps` and its generic argument. For example:

```tsx
const updateLegendPosition: SelectProps<LegendPosition>['onChange'] = (e) => {
  // Typescript will know e.target.value is of type LegendPosition or an empty string (which
  // can be ignored since you don't let people clear the value to empty string)
}
```

LukeTillman (approved):
Overall looks good, left a few suggestions inline. I know we chatted a bit offline, so LMK if you need to talk through testing at all.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Glad to hear you like the merge approach for DEFAULT_LEGEND, I'll update unit to be similar in a follow up PR when I go to make that property editable next!

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
sjcobb: Yeah this makes sense! Just updated it to remove `show` and make legend optional--the only downside I see is when you're editing legend options, if you hide and then unhide the legend, you lose the previous value that `position` was set to. So if `position` was initially 'right', it changes to 'bottom' from DEFAULT_LEGEND when legend is toggled off.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartOptionsEditor.tsx
sjcobb: hmm so I tried making [that change ](https://github.com/perses/perses/pull/660/commits/299ce400e573954d9dfb3ae0c5b484fa642206ce) and still getting `Type 'string' is not assignable to type '"bottom" | "right"'` error:
image

PR #661, opened 2022-10-18
Save and cancel dashboard visual editing changes

eunicorn:
`useDashboardSpec` will return the spec of the dashboard in its current state

When clicking save, it'll only save the current state in the dashboard ui store.  Nothing is saved in the backend, meaning all visual changes will be discarded when refreshing the page. 

When clicking cancel, it'll reset the store to the last saved state.


cndonovan (approved):

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
```diff
@@ -119,10 +137,27 @@ export function createLayoutSlice(
...
     createPanelGroupId,
 
+    savePanelGroups() {
+      set((state) => {
+        state.previousPanelGroupStates = {
+          panelGroups: state.panelGroups,
+          panelGroupIdOrder: state.panelGroupIdOrder,
+        };
+      });
+    },
```
LukeTillman, line 152: Rather than saving old states in the store, what do you think about just assuming that on save, something's going to be creating a new dashboard resource which will cause the store to get re-initialized from scratch? That would simplify all of this quite a bit because you wouldn't need to keep additional state around and "reset" could basically just use the initial state it has from the closure when the store was originally initialized.

LukeTillman (approved):
Looks great. One suggestion inline.

ui/dashboards/src/context/DashboardProvider/layout-slice.tsx
LukeTillman: I'm realizing that in order to do this, all the "initial state" would need to be getting created together (not in separate slices). I had started to centralize that logic in another branch, but I'll think about it some more and maybe we can refactor later to do this.

eunicorn:
> Just in case, the backend is now ready to store the dashboard and is aligned to what the Frontend is proposing in term of typed.
> 
> So you can now save a dashboard by calling the BE.
> 
> Also because it is painful to start ETCD and then you are not able to change the data directly in the DB, I changed that in #658. Which means we will be able to modify the JSON file directly and then see the result directly in the UI. The BE will just need to be started.
> 
> I hope you will find this way confortable and we will be able to migrate the typescript file that holds currently the demo to the json file

oh this is cool!! i'll keep this in mind :)

PR #664, opened 2022-10-19
align the format of the different ID used

Nexucis:
All resource name, variable name and panel name follow the same rule:
- more than 2 characters
- less than 75 characters
- cannot contain any space or special characters 

/cc @saminzadeh @LukeTillman 


Nexucis:
the choice of the number of characters is totally arbitrary and we can of course talk about. The idea is to avoid a reference with only one letter. A variable ID with a single letter won't be helpful for example. At least 2 characters can start to say something.

AntoineThebaud (approved):

PR #666, opened 2022-10-19
Dashboard store renames and clean-up

LukeTillman:
Was discussing with @eunicorn and the name `LayoutSlice` doesn't _really_ make sense any more now that we've moved towards a more view-centric model in the Dashboard state store. So this basically just does some renaming to better reflect the current state of our slices/actions:

1. `LayoutSlice` is now `PanelGroupSlice` since it operates on Panel Groups.
2. The old `PanelGroupSlice` is now `PanelGroupEditorSlice` since it operates on the editor state for Panel Groups.

I also renamed a few action methods to better reflect what they really do (e.g. `openAddPanel` instead of `addPanel`) and updated our hook APIs to use more targeted selectors for panels (e.g. `usePanelActions` instead of `usePanels`). None of the logic really changed in this PR, it's all just renaming of types/methods/files. As always, naming continues to be hard. 😂


internal/api/impl/v1/datasource/service.go
```diff
@@ -124,7 +124,12 @@ func (s *service) Get(parameters shared.Parameters) (interface{}, error) {
 }
 
 func (s *service) List(q etcd.Query, _ shared.Parameters) (interface{}, error) {
-	return s.dao.List(q)
```
sjcobb, line 127: Does this branch need to be rebased (seeing a lot of .go file changes)?

LukeTillman: Yup, I just screwed up a merge. Working on it.

LukeTillman: OK, should be good to go now! Git is hard. 😂

ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
```diff
@@ -24,18 +24,16 @@ export function useEditMode() {
  * Returns actions that can be performed on the current dashboard.
  */
 export function useDashboardActions() {
-  const { addPanelGroup, reset, save } = useDashboardStore(({ addPanelGroup, reset, save }) => ({
-    addPanelGroup,
-    reset,
-    save,
-  }));
-  const addPanel = useDashboardStore((store) => store.addPanel);
+  const save = useDashboardStore((store) => store.save);
```
eunicorn, line 27: how is calling `useDashboardStore` each time different than getting multiple states at once?

ui/dashboards/src/components/DashboardToolbar.tsx
```diff
@@ -27,7 +27,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
   const { dashboardName } = props;
 
   const { isEditMode, setEditMode } = useEditMode();
-  const { addPanelGroup, addPanel, reset, save } = useDashboardActions();
+  const { openAddPanelGroup, openAddPanel, reset, save } = useDashboardActions();
```
sjcobb, line 30: I'm curious if any of these count as a BREAKINGCHANGE in the changelog?

eunicorn: i was going to point this out too. i think it is a break change every time we change the dashboard-provider-api

ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
```diff
@@ -133,52 +152,38 @@ export function usePanel(panelGroupItemId: PanelGroupItemId) {
...
-      updatePanelGroup,
-      swapPanelGroups,
-      deletePanelGroup,
-    })
-  );
+/**
+ * Gets the state for the Panel Editor.
+ */
+export function usePanelEditor() {
+  return useDashboardStore((store) => store.panelEditor);
```
sjcobb, line 170: Nice! This diff looks a lot cleaner than how it was previously

sjcobb (approved):

ui/dashboards/src/components/DashboardToolbar.tsx
LukeTillman: Yeah, unfortunately it is. 😢  I think this API should be stable very soon though. I'm working on another PR at the moment to remove a bunch of "public methods" on the store API which should shrink the surface are of what's public quite a bit.

ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
LukeTillman: Good question: the only difference is avoiding the more expensive shallow comparison check. Zustand basically says "use whichever style you want" in the docs, so I've just been doing the "cheaper" one.

LukeTillman: Edit: I just realized that when you created `useDashboardStore` you added the shallow comparison check automatically 😂 🤦‍♂️  

Thinking maybe we should just remove it, but if you like the object-style selectors better, also totally cool with me to do that instead. 🤷‍♂️

PR #667, opened 2022-10-19
Ability to change variable order

saminzadeh:
Demo: https://www.loom.com/share/6478791cda074913936271a57480f16d

This PR allows you to change the variable order through the UI. Until we land on a drag and drop solution, we can use up and down arrows.

image

### DIFF
--- a/ui/dashboards/src/components/Variables/VariableEditor.tsx
+++ b/ui/dashboards/src/components/Variables/VariableEditor.tsx
@@ -31,6 +31,8 @@ import { VariableDefinition } from '@perses-dev/core';
 import { useImmer } from 'use-immer';
 import PencilIcon from 'mdi-material-ui/Pencil';
 import TrashIcon from 'mdi-material-ui/TrashCan';
+import ArrowUp from 'mdi-material-ui/ArrowUp';
+import ArrowDown from 'mdi-material-ui/ArrowDown';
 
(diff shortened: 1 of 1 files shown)


ui/dashboards/src/components/Variables/VariableEditor.tsx
```diff
@@ -77,6 +79,26 @@ export function VariableEditor(props: {
     });
   };
 
+  const changeVariableOrder = (index: number, direction: 'up' | 'down') => {
+    setVariableDefinitions((draft) => {
+      if (direction === 'up') {
+        if (index === 0) {
+          return;
+        }
+        const tmp = draft[index - 1] as VariableDefinition;
```
sjcobb, line 88: Not sure how but wonder if there's some way to avoid the `as VariableDefinition` type casting 🤔

sjcobb (approved):
💥 Looks like a good approach until we have time to build the fancy drag & drop version

ui/dashboards/src/components/Variables/VariableEditor.tsx
saminzadeh: > Not sure how but wonder if there's some way to avoid the `as VariableDefinition` type casting 🤔

Ah yeah I wrestled with the type system here but I updated it to avoid casting.

PR #668, opened 2022-10-19
Refactor Theme

eunicorn:
- refactored the way we do theme so it's easier to understand
- removed `fullWidth` from default props


sjcobb:
It does look like the query editor looks a little weird, I'm kind of confused on when to use fullWidth, seems like we always need to set it explicitly now?

image

sjcobb (approved):

PR #670, opened 2022-10-19
Compose smaller slices in dashboard store, remove "internal only" methods

LukeTillman:
A few improvements to the dashboard store organization and removes a bunch of methods from the "public API" that shouldn't really have been public:

1. Base slices with the actual state for panels and panel groups that don't depend on any other slices. Since these are the "base slices", this helps us avoid circular dependencies since these are always the "lowest level" slices being imported and having state modified.
2. Smaller action slices (e.g. split editor and delete dialog up into separate slices for both panels and panel groups) that can depend on the base slices they need to modify.
3. No more "internal only" methods on the store that were being used so actions in one slice could call actions in another slice (e.g. get rid of "movePanelToGroup", "addPanelToGroup", etc. that weren't being called from the UI components)
4. Inline the logic for those internal methods so there's only a single "set" call, even when an action needs to modify both panels and panel groups at the same time, which is super common.

Should mean no more multiple renders during those actions since only a single "set" is being used now. It also means a smaller "public API" on the DashboardStore (which should help with fewer breaking changes 🤞).


ui/dashboards/src/context/DashboardProvider/panel-group-delete-slice.ts
```diff
@@ -0,0 +1,101 @@
...
+import { getPanelKeyFromRef } from '@perses-dev/core';
+import { StateCreator } from 'zustand';
+import { Middleware } from './common';
+import { PanelGroupId, PanelGroupSlice } from './panel-group-slice';
+import { PanelSlice } from './panel-slice';
+
+/**
+ * Slice that handles the visual editor state and related actions for deleting a Panel Group.
+ */
+export interface PanelGroupDeleteSlice {
```
eunicorn, line 23: rename to DeletePanelGroupSlice?

ui/dashboards/src/context/DashboardProvider/panel-slice.ts
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+
+import { PanelDefinition } from '@perses-dev/core';
+import { StateCreator } from 'zustand';
+import { Middleware } from './common';
+
+/**
+ * Slice with the state of Panels, along with any actions that modify only the Panels state.
+ */
+export interface PanelSlice {
```
eunicorn, line 21: now i'm thinking if we really need this slice... since this is a "base slice", maybe it can just live in DashboardProvider?

ui/dashboards/src/context/DashboardProvider/panel-slice.ts
```diff
@@ -0,0 +1,56 @@
...
+  panels: Record<string, PanelDefinition>;
+  previousPanels: Record<string, PanelDefinition>;
```
cndonovan, line 23: We don't have a type for panel keys / IDs yet, I wonder if we'd want this type

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -0,0 +1,145 @@
...
+export interface PanelGroupSlice {
+  /**
+   * Panel groups indexed by their ID.
+   */
+  panelGroups: Record<PanelGroupId, PanelGroupDefinition>;
+
+  /**
+   * An array of panel group IDs, representing their order in the dashboard.
+   */
+  panelGroupIdOrder: PanelGroupId[];
```
cndonovan, line 30: I'd call this `panelGroupOrder`, but definitely fine either way

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * previous state
+   */
+  previousPanelGroupStates: {
```
cndonovan, line 35: I'd just called `previousPanelGroupState` (no plural)

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -0,0 +1,145 @@
...
+  title?: string;
+}
+
+/**
+ * Uniquely identifies an item in a PanelGroup.
+ */
+export interface PanelGroupItemId {
+  panelGroupId: PanelGroupId;
+  itemIndex: number;
+}
```
cndonovan, line 74: OK my idea here would be:

`export interface PanelGroupItem` (because this encapsulates more than an ID)
`panelGroupId: PanelGroupID`
`index: number` (since we're inside PanelGroupItem, `index` seems enough)

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+export function createPanelGroupSlice(
+  layouts: LayoutDefinition[]
+): StateCreator<PanelGroupSlice, Middleware, [], PanelGroupSlice> {
+  // Helper function for generating unique IDs for a PanelGroup
+  let id: PanelGroupId = -1;
+  function createPanelGroupId(): PanelGroupId {
+    id++;
+    return id;
+  }
```
cndonovan, line 87: We're not exposing these IDs to the user, right? In other words, we don't need to worry about always having the IDs in an order like `1, 2, 3...` even if panel groups get deleted... because instead, these are just internally used IDs

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
cndonovan: Or could even just do `previousState`!

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
cndonovan: Actually I take it back `itemIndex` seems nice too

ui/dashboards/src/context/DashboardProvider/panel-editor-slice.ts
```diff
@@ -113,33 +74,26 @@ export interface PanelEditorValues {
...
-      });
-    },
-
-    openEditPanel(item) {
-      const { panels, getPanelKey } = get();
-
-      // Ask the layout store for the panel key at that location
-      const panelKey = getPanelKey(item);
+      // Figure out the panel key at that location
+      const content = panelGroups[panelGroupItemId.panelGroupId]?.items[panelGroupItemId.itemIndex]?.content;
```
cndonovan, line 92: This kind of makes me want to destructure from `panelGroupItemId` before doing this line of code

ui/dashboards/src/context/DashboardProvider/panel-group-delete-slice.ts
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Delete panel group and all the panels within the group
+   */
+  deletePanelGroup: (panelGroupId: PanelGroupId) => void;
+
+  openDeletePanelGroupDialog: (panelGroupId: PanelGroupId) => void;
+  closeDeletePanelGroupDialog: () => void;
+}
+
+export interface DeletePanelGroupDialog {
```
cndonovan, line 35: Could rename to `DeletePanelGroupDialogState` like `DeletePanelDialogState` but it's mouthful for sure

ui/dashboards/src/context/DashboardProvider/panel-group-delete-slice.ts
```diff
@@ -0,0 +1,101 @@
...
+      // Delete the panel group which also deletes all its items
+      delete draft.panelGroups[panelGroupId];
+      draft.panelGroupIdOrder.splice(idIndex, 1);
+
+      // Get all remaining panel keys in use
+      const usedPanelKeys = getUsedPanelKeys(draft.panelGroups);
+
+      // For the panel keys of the items that were just deleted, see if they're still used and if not, also delete the
+      // panel definition
+      for (const panelKey of panelKeys) {
```
cndonovan, line 68: Does it make sense to try to use the `panel-delete-slice` `deletePanel` function within this function? In other words, calling `deletePanelGroup` means we will also call `deletePanel` under the hood? 

Or is it better / less bug-prone and easier to reason about if we don't cross those wires?

cndonovan (approved):
Overall, really like the slicing and dicing because it brings more organization, and also love the idea of trying to have "base slices" that represent the lowest level that other parts of state can depend on.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
cndonovan: I'm just spitballing now, but I wonder if we'd want to even have:

`panelGroupId: PanelGroupID`
`panelIndex: number`

Have not thought this all the way through, but maybe we want to stop saying generic "item" and just starting saying "panel"

ui/dashboards/src/context/DashboardProvider/panel-slice.ts
LukeTillman: Yeah, good idea, we probably want a type alias.

ui/dashboards/src/context/DashboardProvider/panel-slice.ts
LukeTillman: Yeah, like we talked about offline, the main reason to keep it as a "base slice" is to have imports all go in one direction and not be circular.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
LukeTillman: Yeah, I'm thinking when I split out "items" so that I can connect drag and drop, there may be some renaming here. Look for that in a PR soon.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
LukeTillman: Yeah, they're exposed to the UI components, but the fact they're numbers/in order isn't relied on at all. Just meant to be a "stable" way to uniquely identify a panel group.

ui/dashboards/src/context/DashboardProvider/panel-group-delete-slice.ts
LukeTillman: Yeah, good question. Originally it was implemented that way and this is basically "undoing" that because it was more confusing than it needed to be. For example, it meant you'd have multiple "set" calls to state (causing multiple re-renders) when each panel was deleted.

eunicorn (approved):

PR #674, opened 2022-10-20
Fix variable editor styling

saminzadeh:
The changes made in #668 caused the variable editor styles to be wonky. This was fixed by adding the `fullWidth` property on the inputs.

image

### DIFF
--- a/ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
+++ b/ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
@@ -181,20 +181,20 @@ export function VariableEditForm({
 
       <Stack direction={'row'} spacing={2} justifyContent="end">
         <Button
-          onClick={() => {
-            onCancel();
-          }}
-        >
-          Cancel
-        </Button>
-
-        <Button
           variant="contained"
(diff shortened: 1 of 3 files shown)


ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -191,7 +194,7 @@ export function VariableEditForm({
             onChange(getVariableDefinitionFromState(state));
           }}
         >
-          Update Variable
+          Update
```
eunicorn, line 197: can you switch the update and cancel button to keep it consistent with the rest of the app? i think we have primary button on the left and cancel on the right

eunicorn (approved):
thanks for fixing these!!

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
saminzadeh: Yep, will do. I'm also planning to do an overhaul of the UX once all the functionality is complete. We can build re-usable components too to help us be consistent.

cndonovan (approved):

PR #677, opened 2022-10-21
display the dashboard list

Nexucis:
This PR proposed a first version to display the list of available dashboard.

image

### DIFF
--- a/ui/app/src/views/ViewDashboard.tsx
+++ b/ui/app/src/views/ViewDashboard.tsx
@@ -41,36 +41,39 @@ function ViewDashboard() {
     return generateChartsTheme('perses', muiTheme, ECHARTS_THEME_OVERRIDES);
   }, [muiTheme]);
 
-  const projectName = useParams().projectName;
-  const dashboardName = useParams().dashboardName;
+  const { projectName, dashboardName } = useParams();
+
   if (projectName === undefined || dashboardName === undefined) {
     throw new Error('Unable to get the Dashboard or Project name');
   }
-  const { data, isLoading } = useDashboard(projectName, dashboardName);
+
   const datasourceApi = useDatasourceApi();
 
-  if (!isLoading && data !== undefined) {
-    return (
-      <Box
-        component="main"
-        sx={{
-          flexGrow: 1,
-          overflow: 'hidden',
-        }}
-      >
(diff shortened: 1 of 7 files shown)


Nexucis:
I'm moving this PR as draft since I have some trouble with the navigation. Somethings strange happens when passing from the list to the `DashboardView` it doesn't look it used the correct data.

Same thing when moving from `ViewDashboard` to the root. It doesn't get the list of dashboard. Instead the data is already filled with the value coming from the `ViewDashboard`. Which is super weird for me since `ViewDashboardList` is a different view

Nexucis:
I'm wondering if it's coming from the warning you can see when you are going on the page https://localhost:3000/projects/perses/dashboards/Benchmark?start=6h manually

image

sjcobb:
> I'm wondering if it's coming from the warning

Just pushed a commit to this branch that fixes the console error you mentioned!

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -41,36 +41,39 @@ function ViewDashboard() {
...
-                <DashboardView dashboardResource={data} datasourceApi={datasourceApi} />;
-              </ErrorBoundary>
-            </PluginRegistry>
-          </ChartsThemeProvider>
-        </ErrorBoundary>
-      </Box>
-    );
-  }
-  return null;
+  const { data, isLoading } = useDashboard(projectName, dashboardName);
```
sjcobb, line 52: Pushed a commit that adjusts this loading state and now you can navigate from the dashboard listing to an individual dashboard. 

Something is still off with useDashboard results though and routing generally (back button is broken when trying to go back to the dashboard listing), but I need to log off for the day

Nexucis: Thank you very much, everything is working now !

For the back button actually it's working, it's just the dashboard that is adding automatically the `?start=6h` so when you click on the back button, you can see it's removing the `?start=6h`, which is then added back automatically by the dashboard.

sjcobb: Ah okay gotcha, I can look into fixing this in follow up PR

sjcobb (approved):

PR #678, opened 2022-10-21
fallback to static json TextField in gauge and stat editors

sjcobb:
- copy parts of Shan's JSONSpecEditor into GaugeChartOptionsEditor and StatChartOptionsEditor
- allows user to edit panel options if they already know info about GaugeChart and StatChart specs

### Screenshots

gauge_temp_editor
stat_temp_editor


ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
```diff
@@ -11,13 +11,44 @@
...
 import { OptionsEditorProps } from '@perses-dev/plugin-system';
-import { Box } from '@mui/material';
 import { GaugeChartOptions } from './gauge-chart-model';
 
 export type GaugeChartOptionsEditorProps = OptionsEditorProps<GaugeChartOptions>;
 
 export function GaugeChartOptionsEditor(props: GaugeChartOptionsEditorProps) {
-  const { value } = props;
-  return <Box>{JSON.stringify(value)}</Box>;
+  const ref = useRef<HTMLInputElement>(null);
```
LukeTillman, line 22: Don't _think_ we need the ref in here at all?

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
```diff
@@ -11,13 +11,44 @@
...
+  const [value, setValue] = useState(JSON.stringify(props.value, null, 2));
```
LukeTillman, line 23: Might want to change this to the callback style to init the state so we're not running `JSON.stringify` on every render cycle.

LukeTillman (approved):
Looks like a good stopgap solution to me. Left a few minor suggestions.

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
saminzadeh, line 22: Rather than copy and paste, should we just create a JSONEditor component than add it to each Plugin. Similar to this PR:

https://github.com/perses/perses/pull/627/files#diff-e98abc89e67f736bd08bda2b6e011f4e555adcde880f8c5ac17b743d987411d7R18

https://github.com/perses/perses/pull/627/files#diff-95552669c95c924f6a5299409039403a40ba51baa6fe2bb1ae6580a851ea17afR57

sjcobb: Yeah definitely, just added it to the components package, I was being lazy before since this shouldn't be around very long once I get to adding actual form controls

PR #683, opened 2022-10-24
deep review of the API e2e tests and removed user endpoint

Nexucis:
This PR reviews the way we are doing the API e2e tests to factorise a bit the common test on each endpoint.
Like that it should easier to have a set of regular test per endpoint.

I also took the opportunity to remove the user endpoints that are not used for the moment, and likely we will need a complete specification to know how to put in place the authorisation/authentication.


Nexucis:
this PR should help also to fixed the tests in #680

AntoineThebaud (approved):

PR #684, opened 2022-10-24
Connect react-grid-layout changes with dashboard store

LukeTillman:
For each panel group, this PR splits the item state inside that panel group into two separate pieces of state:

- `itemLayouts`: an array of the layout information (`x`, `y`, `w`, `h`, and a unique ID `i`) which corresponds 1:1 with the model that `react-grid-layout` expects
- `itemPanelKeys`: a map of the unique ID of the item in `itemLayouts` -> the panel key referenced by that panel

By splitting this state up, I was then able to connect `react-grid-layout` with the store in a pretty straightforward manner. Its `onLayoutChange` handler can now just overwrite the `itemLayouts` property on a panel group every time a move/resize happens, while the content reference (i.e. panel key) remains stable. This also means that drag and drop actions should now be reflected in the object that's output for conversion to JSON. 🥳

### DIFF
--- a/ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
+++ b/ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
@@ -11,7 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { getPanelKeyFromRef } from '@perses-dev/core';
 import { useMemo } from 'react';
 import { useDashboardStore } from './DashboardProvider';
 import { PanelGroupItemId, PanelGroupId } from './panel-group-slice';
@@ -136,12 +135,11 @@ export function useDeletePanelGroupDialog() {
  * Gets an individual panel in the store. Throws if the panel can't be found.
  */
 export function usePanel(panelGroupItemId: PanelGroupItemId) {
-  const { panelGroupId, itemIndex } = panelGroupItemId;
+  const { panelGroupId, panelGroupLayoutId } = panelGroupItemId;
 
   const panel = useDashboardStore((store) => {
-    const panelRef = store.panelGroups[panelGroupId]?.items[itemIndex]?.content;
-    if (panelRef === undefined) return;
-    const panelKey = getPanelKeyFromRef(panelRef);
+    const panelKey = store.panelGroups[panelGroupId]?.itemPanelKeys[panelGroupLayoutId];
+    if (panelKey === undefined) return;
     return store.panels[panelKey];
   });
 
--- a/ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
+++ b/ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
@@ -13,7 +13,7 @@
 
 import { useMemo } from 'react';
 import { useDashboardStore } from './DashboardProvider';
-import { PanelGroupItemId, PanelGroupId } from './panel-group-slice';
+import { PanelGroupItemId, PanelGroupId, PanelGroupLayout } from './panel-group-slice';
 
 export function useEditMode() {
   return useDashboardStore(({ isEditMode, setEditMode }) => ({ isEditMode, setEditMode }));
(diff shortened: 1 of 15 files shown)


ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
```diff
@@ -136,12 +137,11 @@ export function useDeletePanelGroupDialog() {
...
  */
 export function usePanel(panelGroupItemId: PanelGroupItemId) {
-  const { panelGroupId, itemIndex } = panelGroupItemId;
+  const { panelGroupId, panelGroupLayoutId } = panelGroupItemId;
 
   const panel = useDashboardStore((store) => {
-    const panelRef = store.panelGroups[panelGroupId]?.items[itemIndex]?.content;
-    if (panelRef === undefined) return;
-    const panelKey = getPanelKeyFromRef(panelRef);
+    const panelKey = store.panelGroups[panelGroupId]?.itemPanelKeys[panelGroupLayoutId];
```
sjcobb, line 143: Oh this reminds me, not sure if it's still an issue, but last week I saw some weird behavior where if I gave two panels the same name, there would be a collision and subsequent changes to one would effect the other (I think I created a StatChart  and GaugeChart, then they both became GaugeCharts with duplicate names / descriptions).

sjcobb (approved):
🚀

ui/dashboards/src/context/DashboardProvider/dashboard-provider-api.ts
LukeTillman: Yeah, that's been in the back of my mind as something we need to fix for panel keys. Definitely still an issue.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -57,17 +63,24 @@ export type PanelGroupId = number;
...
   id: PanelGroupId;
-  items: GridItemDefinition[];
   isCollapsed: boolean;
   title?: string;
+  itemLayouts: PanelGroupLayout[];
+  itemPanelKeys: Record<PanelGroupLayoutId, string>;
 }
 
+export interface PanelGroupLayout extends Layout {
+  i: PanelGroupLayoutId;
```
eunicorn, line 73: can we call this `id`?

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -57,17 +63,24 @@ export type PanelGroupId = number;
...
 export interface PanelGroupDefinition {
   id: PanelGroupId;
-  items: GridItemDefinition[];
   isCollapsed: boolean;
   title?: string;
+  itemLayouts: PanelGroupLayout[];
+  itemPanelKeys: Record<PanelGroupLayoutId, string>;
 }
 
+export interface PanelGroupLayout extends Layout {
```
eunicorn, line 72: can we rename this to PanelGroupItemLayout? PanelGroupLayout makes it seems like its the layout for the whole panel group itself

eunicorn: naming is hard 😆

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -76,18 +89,36 @@ export interface PanelGroupItemId {
...
-    const id = generateId();
-    panelGroups[id] = {
-      id,
-      items: layout.spec.items,
+    const itemLayouts: PanelGroupDefinition['itemLayouts'] = [];
+    const itemPanelKeys: PanelGroupDefinition['itemPanelKeys'] = {};
+
+    // Split layout information from panel keys to make it easier to update just layouts on move/resize of panels
+    for (const item of layout.spec.items) {
+      const panelGroupLayoutId = generateId().toString();
```
eunicorn, line 101: i am thinking if we could somehow generate/enforce unique `panelRef`, then we wouldn't have to keep track of unique id and panelRef

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
LukeTillman: This comes from `react-grid-layout` (I basically just redefined the type from `string` so it was more obvious what it actually is), so unfortunately has to stay `i` unless we want to have to transform back and forth with their format.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
LukeTillman: Yup, good call.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
LukeTillman: Yeah, we're still dealing with the whole "panels can be referenced multiple times" thing, so a `panelRef` isn't guaranteed to be unique. 🙃

eunicorn (approved):

PR #685, opened 2022-10-24
add test for new legend in TimeSeriesChart panel

sjcobb:
- add unit test for legend visibility and selected state
- update ECharts wrapper component to not throw errors
- mock data for useTimeSeriesQueries results


ui/components/src/Legend/ListLegendItem.tsx
```diff
@@ -30,6 +30,7 @@ export const ListLegendItem = React.memo(function ListLegendItem({ item }: ListL
         padding: 0,
         cursor: 'pointer',
       }}
+      data-testid={`legend-item-${item.color}`}
```
sjcobb, line 33: This is temporary, I'm planning to change how item.id is generated when I make changes at the query level

pkg/model/api/v1/dashboard_test.go
```diff
@@ -25,6 +25,7 @@ import (
 )
 
 type TimeSeriesSpec struct {
+	// TODO: show_legend needs to be removed in favor of new spec
```
Nexucis, line 28: Just for the record, this is just a random struct I created for the test, so it doesn't need to follow the actual definition of TimeSeries panel :)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    unit: { kind: 'Decimal', decimal_places: 2 },
+    legend: {
+      position: 'right',
+    },
+  },
+};
+
+describe('TimeSeriesChartPanel', () => {
+  beforeEach(() => {
+    (useTimeSeriesQueries as jest.Mock).mockReturnValue(mockTimeSeriesQueryResult);
```
LukeTillman, line 60: Rather than mocking the hook, why not use the real hook and mock the `PrometheusTimeSeriesQuery` plugin instead? The plugin registry mock utility you added returns a `addMockPlugin` function for just that reason.

ui/panels-plugin/src/test/mock-query-result.ts
```diff
@@ -0,0 +1,1830 @@
+// Copyright 2022 The Perses Authors
```
LukeTillman, line 1: Holy crap this file 😂  I wonder if there's a shorthand way to generate this rather than a 1800 line file? Totally fine if not, but I saw this PR and was like "how did he write 2000 lines of tests?!"

LukeTillman (approved):
Couple minor suggestions, but LGTM. 🚀

ui/panels-plugin/src/test/mock-query-result.ts
sjcobb: haha yeah good point, forgot to clean this up, just reduced that file to ~60 lines since most of the values weren't necessary for what I'm testing here

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.test.tsx
sjcobb: We discussed this offline, the useTimeSeriesQueries hook has [other runtime dependencies](https://github.com/perses/perses/blob/v0.12.0/ui/plugin-system/src/runtime/time-series-queries.ts#L73-L75) that aren't easy to mock, I'll come back to this later after we work on adding some additional test utils

PR #686, opened 2022-10-24
Remove suggested step from time series query key

LukeTillman:
This just removes the "suggested step size" from the `react-query` key when running time series queries. Should fix the issue where queries get re-run when the browser is resized.


sjcobb (approved):
💯

PR #689, opened 2022-10-25
add a display field to the dashboards

Nexucis:

### DIFF
--- a/ui/core/src/model/dashboard.ts
+++ b/ui/core/src/model/dashboard.ts
@@ -17,6 +17,7 @@ import { PanelDefinition } from './panels';
 import { ProjectMetadata } from './resource';
 import { DurationString } from './time';
 import { VariableDefinition } from './variables';
+import { Display } from './display';
 
 export interface DashboardResource {
   kind: 'Dashboard';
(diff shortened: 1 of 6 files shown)


ui/core/src/model/dashboard.ts
```diff
@@ -25,6 +26,7 @@ export interface DashboardResource {
 }
 
 export interface DashboardSpec {
+  display?: Display;
```
sjcobb, line 29: Is this mostly for consistency? Curious if there is overlap with how `metadata` is used--for instance, when would display.name be used compared with metadata.name?

Nexucis: Yeah it's for consistency and also because in `metadata.name` you cannot put space. So I was thinking maybe you would like to put some space in the dashboard title.
Basically if `disply.name` is used, it replaced metadata.name in the Breadcrumbs.

But I'm not super sure it's a great idea.

sjcobb: Ah okay I see, sorry it took me a while to follow up here! And it's optional so this isn't a breaking change or anything right?

sjcobb (approved):

ui/core/src/model/dashboard.ts
Nexucis: yup no breaking change here. If it's not used, then `metadata.name` should be used :)

PR #690, opened 2022-10-25
Implement variable state in URL parameters

saminzadeh:
Video Demo: https://www.loom.com/share/6db09276573c4e9393fc19a8a05d5670

This implements syncing the template variable values in URL query parameters.
Each variable has a unique query parameter `var-${name}`. Multiple values for a query parameter are comma separated.

So for the following variable state:
```json
{ "foo": ["val1", "val2"], "bar": "val1" }
```

The URL query parameters would be:
```
?var-foo=val1,val2&var-bar=val1
```

### DIFF
--- a/ui/dashboards/src/components/Variables/Variable.tsx
+++ b/ui/dashboards/src/components/Variables/Variable.tsx
@@ -113,23 +113,10 @@ function ListVariable({ name }: TemplateVariableProps) {
 
   useEffect(() => {
     const firstOption = finalOptions?.[0];
-    const valueIsInOptions = Boolean(
-      finalOptions.find((v) => {
-        if (allowMultiple) {
-          return (value as string[]).includes(v.value);
-        }
-        return value === v.value;
-      })
-    );
 
     // If there is no value but there are options, set the value to the first option.
     if (!value && firstOption) {
       setVariableValue(name, firstOption.value);
-    }
-
-    // If there is a value but it's not in the options, select the first value or set to null
-    if (value && !valueIsInOptions && !definition.spec.default_value) {
-      setVariableValue(name, firstOption?.value || null);
     }
   }, [finalOptions, setVariableValue, value, name, allowMultiple]);
 
@@ -181,12 +168,14 @@ function TextVariable({ name }: TemplateVariableProps) {
   }, [state?.value]);
 
   return (
-    <TextField
-      value={tempValue}
-      onChange={(e) => setTempValue(e.target.value)}
-      onBlur={() => setVariableValue(name, tempValue)}
-      placeholder={name}
-      label={name}
-    />
+    <>
+      <TextField
+        value={tempValue}
+        onChange={(e) => setTempValue(e.target.value)}
+        onBlur={() => setVariableValue(name, tempValue)}
+        placeholder={name}
+        label={name}
+      />
+    </>
   );
 }
--- a/ui/dashboards/src/components/Variables/Variable.tsx
+++ b/ui/dashboards/src/components/Variables/Variable.tsx
@@ -168,14 +168,12 @@ function TextVariable({ name }: TemplateVariableProps) {
   }, [state?.value]);
 
   return (
-    <>
-      <TextField
-        value={tempValue}
-        onChange={(e) => setTempValue(e.target.value)}
-        onBlur={() => setVariableValue(name, tempValue)}
-        placeholder={name}
-        label={name}
-      />
-    </>
+    <TextField
+      value={tempValue}
+      onChange={(e) => setTempValue(e.target.value)}
+      onBlur={() => setVariableValue(name, tempValue)}
+      placeholder={name}
+      label={name}
+    />
   );
 }
--- a/ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
+++ b/ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
@@ -15,6 +15,7 @@ import { createContext, useContext, useMemo, useState } from 'react';
 import { createStore, useStore } from 'zustand';
 import { immer } from 'zustand/middleware/immer';
 import { devtools } from 'zustand/middleware';
+
 import {
   TemplateVariableContext,
   VariableStateMap,
@@ -23,6 +24,7 @@ import {
   DEFAULT_ALL_VALUE as ALL_VALUE,
 } from '@perses-dev/plugin-system';
 import { VariableName, VariableValue, VariableDefinition } from '@perses-dev/core';
+import { useVariableQueryParams, getInitalValuesFromQueryParameters, getURLQueryParamName } from './query-params';
 
 type TemplateVariableStore = {
   variableDefinitions: VariableDefinition[];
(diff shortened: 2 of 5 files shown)


ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -113,24 +113,11 @@ function ListVariable({ name }: TemplateVariableProps) {
...
-      })
-    );
 
     // If there is no value but there are options, set the value to the first option.
     if (!value && firstOption) {
       setVariableValue(name, firstOption.value);
     }
-
-    // If there is a value but it's not in the options, select the first value or set to null
-    if (value && !valueIsInOptions && !definition.spec.default_value) {
```
saminzadeh, line 131: Removing this logic for now because this becomes a bit more tricky when you have URL parameters. I think we'll have to think about this more. Should values be allowed if the values are not in the options? My gut is yes, since sometimes options are slow to load / fail.

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
```diff
@@ -174,6 +178,10 @@ function createTemplateVariableSrvStore({ initialVariableDefinitions = [] }: Tem
                 val = val.filter((v) => v !== ALL_VALUE);
               }
             }
+            if (queryParams) {
+              const setQueryParams = queryParams[1];
+              setQueryParams({ [getURLQueryParamName(name)]: val });
```
sjcobb, line 183: Pulled your branch locally and seemed to work well! I like that our URLs are actually readable so glad we didn't go with the JSON encoded params option.

In my implementation, one downside I just noticed of getting rid of 'end=now' when 'start' is relative, is once the time range switches to absolute, the end param gets appended instead of being beside 'start' ([ex1](https://localhost:3000/projects/perses/dashboards/Demo?var-instance=%24__all&start=1666721610000&var-interval=1m&var-job=grafana%2Cnode&end=1666721700000), [ex2](https://localhost:3000/projects/perses/dashboards/DemoDashboard?start=1666717080000&var-instance=demo.do.prometheus.io%3A3000&var-interval=1m&var-job=node&end=1666717125000)). 

Definitely a nit, but I wonder if there is some way we could ensure time range params and template variable params stay together, another example is when I load a dashboard without any params, it defaulted to: `?var-instance=%24__all&start=5m&var-interval=1m`

sjcobb:
Seeing some console warnings (but could just be an issue with my dashboards locally)

`MUI: You have provided an out-of-range value '1m' for the select component.
Consider providing a value that matches one of the available options or ''. The available values are loading.`

`MUI: You have provided an out-of-range value 'demo.do.prometheus.io:9100' for the select component.
Consider providing a value that matches one of the available options or ''. The available values are loading.`

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
saminzadeh: Hmm yeah. My guess is because when a variable loads, it only updates its value and not all the values at once. I believe we would need to group the updates together to get that functionality

saminzadeh:
> Seeing some console warnings (but could just be an issue with my dashboards locally)
> 
> `MUI: You have provided an out-of-range value '1m' for the select component. Consider providing a value that matches one of the available options or ''. The available values are loading.`
> 
> `MUI: You have provided an out-of-range value 'demo.do.prometheus.io:9100' for the select component. Consider providing a value that matches one of the available options or ''. The available values are loading.`

The reason why is the value can be provided but the option is not there or not loaded for the MUI Select. I think we will need to switch to [Autocomplete](https://mui.com/material-ui/react-autocomplete/) instead of MUI Select to avoid this.

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
```diff
@@ -23,6 +24,7 @@ import {
   DEFAULT_ALL_VALUE as ALL_VALUE,
 } from '@perses-dev/plugin-system';
 import { VariableName, VariableValue, VariableDefinition } from '@perses-dev/core';
+import { useVariableQueryParams, getInitalValuesFromQueryParameters, getURLQueryParamName } from './query-params';
```
sjcobb, line 27: I like how you organized this, I put a lot of the time-range implementation inside of utils, which I'm going to change now. For disabling params, we can do it in QueryParamProvider like you mentioned, but is the plan for TemplateVariableProvider to also have a new prop for this?

saminzadeh: Yeah exactly. I didn't implement it yet, but it would basically be:
```tsx
<TemplateVariableProvider enabledURLParams={false} {...rest} />
```

The top level Dashboard component can control that.

sjcobb (approved):

LukeTillman:
> > Seeing some console warnings (but could just be an issue with my dashboards locally)
> > `MUI: You have provided an out-of-range value '1m' for the select component. Consider providing a value that matches one of the available options or ''. The available values are loading.`
> > `MUI: You have provided an out-of-range value 'demo.do.prometheus.io:9100' for the select component. Consider providing a value that matches one of the available options or ''. The available values are loading.`
> 
> The reason why is the value can be provided but the option is not there or not loaded for the MUI Select. I think we will need to switch to [Autocomplete](https://mui.com/material-ui/react-autocomplete/) instead of MUI Select to avoid this.

I went back and forth on whether to switch to `Autocomplete` also to avoid this (you also get a built-in loading indicator which is nice), but as a temporary workaround you can do something like this:

https://github.com/perses/perses/blob/f33741e43c01615296fc9a101364c9e7f7bff085/ui/plugin-system/src/components/PluginKindSelect.tsx#L30-L31

Did that in at least a couple places to avoid that warning.

ui/dashboards/src/context/TemplateVariableProvider/query-params.ts
```diff
@@ -0,0 +1,73 @@
...
+export function useVariableQueryParams(defs: VariableDefinition[]) {
+  const config: Record<string, typeof VariableValueParam> = {};
+  defs.forEach((def) => {
+    const name = getURLQueryParamName(def.spec.name);
+    config[name] = VariableValueParam;
+  });
+  return useQueryParams(config);
+}
+
+export function getInitalValuesFromQueryParameters(
```
LukeTillman, line 60: Might not be a TON of value (so feel free to ignore), but some of these little "util" functions might be a good candidate for unit tests.

LukeTillman (approved):
Looks great. If you want to avoid the MUI warnings with the `Select`, see my comment above.

Nexucis:
Interesting ! I'm just wondering what would happen if we hit the maximum size of the URL (2048 characters)

ui/dashboards/src/context/TemplateVariableProvider/query-params.ts
saminzadeh: Yeah definitely, I'll add those.

ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -111,26 +111,27 @@ function ListVariable({ name }: TemplateVariableProps) {
...
+        finalOptions.find((v) => {
+          if (allowMultiple) {
+            return (value as string[]).includes(v.value);
+          }
+          return value === v.value;
+        })
+      ),
+    [finalOptions, value, allowMultiple]
+  );
+  const selectValue = valueIsInOptions ? value ?? '' : '';
```
saminzadeh, line 126: I had to use this over `loading` since the value can be set even before the options start loading (via query parameter or default value)

PR #692, opened 2022-10-25
Add time range context to variables plugin + prom plugins

saminzadeh:
- [core plugin] Variable plugin now has the timeRange context so variables can consume 
- [prom] The labelValues and labelNames variables now include the time range in the API request (speeds on the request)


ui/dashboards/src/components/Variables/Variable.tsx
```diff
@@ -75,11 +76,12 @@ function ListVariable({ name }: TemplateVariableProps) {
   }
 
   const variablesValueKey = getVariableValuesKey(variables);
+  const { timeRange } = useTimeRange();
 
   const variablesOptionsQuery = useQuery(
     [name, definition, variablesValueKey],
     async () => {
-      const resp = await variablePlugin?.getVariableOptions(spec, { datasourceStore, variables });
+      const resp = await variablePlugin?.getVariableOptions(spec, { datasourceStore, variables, timeRange });
```
saminzadeh, line 84: We'll have to think about giving users the ability to control when these variables reload based on time range. Right now it will only use the time range on the initial page load. So there may be cases where 5m has different data than 6h and this could be problematic. 

See [Refresh](https://grafana.com/docs/grafana/v9.0/variables/variable-types/add-query-variable/#enter-query-options) for Grafana

LukeTillman: Do you think this should be part of the query key so we _always_ reload on time range change? Seems like that's maybe the behavior we want initially? (We could then add the option to refresh less frequently if that's an issue.)

LukeTillman (approved):
💥 Gonna be a big help with metadata query performance. One question inline.

ui/dashboards/src/components/Variables/Variable.tsx
saminzadeh: Ah yeah I thought about this. I suppose it makes sense because then the values would always be accurate based on the time range, but this means all variables would reload on each time range change even if the query backend doesn't support it. My gut is that 95% of use cases don't need the time range so I feel like it could reduce the overall performance of the app for majority of use cases but it could be worth trying.

PR #694, opened 2022-10-26
Decouple PanelPreview from edit state

LukeTillman:
This decouples the `Panel` from dashboard store's edit state, allowing `PanelPreview` to be rendered in "view mode" when used in the panel editor. Here's what it looks like in action now (you'll see the "edit buttons" are gone in the preview component now):

https://user-images.githubusercontent.com/428023/198093684-87765333-8743-416f-a51b-1c799459051c.mov

I also tried to add a bunch of tests to `Panel` and improved some of the ARIA attributes in the process. This should start to make dashboard tests a little easier to write and (for example) look for individual panels.

### DIFF
--- a/ui/dashboards/src/components/GridLayout/GridItemContent.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridItemContent.tsx
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { PanelGroupItemId, usePanel } from '../../context';
-import { Panel } from '../Panel/Panel';
+import { PanelGroupItemId, useEditMode, usePanel, usePanelActions } from '../../context';
+import { Panel, PanelProps } from '../Panel/Panel';
 
 export interface GridItemContentProps {
   panelGroupItemId: PanelGroupItemId;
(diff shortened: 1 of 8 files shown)


ui/dashboards/src/components/GridLayout/GridItemContent.tsx
```diff
@@ -24,5 +24,17 @@ export interface GridItemContentProps {
...
+  // Provide actions to the panel when in edit mode
+  let editHandlers: PanelProps['editHandlers'] = undefined;
+  if (isEditMode) {
+    editHandlers = {
+      onEditPanelClick: openEditPanel,
+      onDeletePanelClick: openDeletePanelDialog,
+    };
+  }
+
+  return <Panel definition={panelDefinition} editHandlers={editHandlers} />;
```
cndonovan, line 39: I wonder if we want to call these `actionHandlers` for a future when we might have other actions to handle (duplicate?)

cndonovan: But also this seems good for now, we always do that if/when that day comes

cndonovan (approved):

ui/dashboards/src/components/GridLayout/GridItemContent.tsx
eunicorn: in the future, it would also be cool if users can customize what icons they want there

eunicorn (approved):

PR #697, opened 2022-10-27
add a project view with dasbhoard list

Nexucis:


ui/app/src/components/Header.tsx
```diff
@@ -48,45 +51,55 @@ function ProjectMenu(): JSX.Element {
...
+          vertical: 'bottom',
+          horizontal: 'left',
+        }}
+        keepMounted
+        open={anchorEl !== null}
+        onClose={handleCloseMenu}
+        PaperProps={{
+          style: {
+            maxHeight: ITEM_HEIGHT * 4.5,
+            width: '20ch',
```
sjcobb, line 83: Maybe an inline comment here could be useful on why using 'ch' here is needed? Not saying it isn't justified, just isn't something you see too often

ui/app/src/views/ViewDashboardList.tsx
```diff
@@ -31,9 +31,34 @@ import { ErrorAlert, ErrorBoundary } from '@perses-dev/components';
...
+  dashboardList: DashboardResource[];
+}
+
+export function DashboardList(props: DashboardListProperties) {
+  const navigate = useNavigate();
+  return (
+    <List>
+      {props.dashboardList.map((dashboard, i) => {
+        return (
+          <Box sx={{ backgroundColor: (theme) => theme.palette.primary.main + '10' }} key={dashboard.metadata.name}>
```
sjcobb, line 44: Could a <ListItem> be used here instead of Box?

ui/app/src/views/ViewProject.tsx
```diff
@@ -0,0 +1,61 @@
...
+  if (projectName === undefined) {
+    throw new Error('Unable to get the Project name');
+  }
+  return (
+    <Container maxWidth="md">
+      <Stack direction="row" alignItems="center" gap={1} mb={2}>
+        <FolderPound fontSize={'large'} />
+        <Typography variant="h1">{projectName}</Typography>
+      </Stack>
+      <RenderDashboardInProject projectName={projectName} />
```
sjcobb, line 56: This component name is a little awkward, would something like DashboardPage.tsx make sense? I don't think we have one of those yet, this is tough since there are already so many components with some form of 'Dashboard' in their names 😂

sjcobb (approved):
Couple minor comments but overall LGTM! 🚀

ui/app/src/views/ViewDashboardList.tsx
Nexucis: ah yeah that would make more sense. I will try it. Thanks for your eyes ! 🙏

ui/app/src/views/ViewProject.tsx
Nexucis: yeah naming around dashboard is horrible :(

ui/app/src/components/Header.tsx
Nexucis: I put `10em` instead

PR #698, opened 2022-10-27
remove not used doc page

Nexucis:
If one day we have a website that will display the doc, perhaps we could just put a link in the footer instead. It will be tinier but it will take less place :)

For the moment I'm removing it, because it doesn't show anything.


cndonovan (approved):

sjcobb (approved):
👍 👍

PR #699, opened 2022-10-27
Unsaved Changes Confirmation Dialog

eunicorn:
Clicking cancel button will show an unsaved changes confirmation dialog if user makes any changes

TO DO:

- show the unsaved changes confirmation dialog when user navigates away from dashboard 

unsavedChanges


Nexucis:
I'm trying to trigger this confirmation dialog and so far I didn't get how to do it.

1. I click on the `Edit` button
2. I added a new element in the dashboard
3. I clicked on the `Perses` title to come back to the root page, or I'm clicking on the cancel button.

I thought one of the actions I did in 3 would trigger it.

Could you explain when this dialog should appear ? 

As a side note, I would prefer to reduce the number of dialogs. I found it a bit intrusive and increase the complexity of the UI and the code in general. So the less we have, the happier I am (even if it doesn't count that much :))

For example, I really like the PromLens UI because Julius managed to avoid the usage of the Dialog. So even when adding a new stuf, I'm able to navigate through the whole view. It would help me for example to have an example about a variable configuration at the same time. Which is not possible at this moment

Nexucis:
Being NOT able to see another variable configuration at the same time I'm adding or editing another one was something that super annoy me in Grafana.
Generally speaking not being able to the whole dashboard value at once always bothered me. Because many parts are hidden, and you can only see some piece of it by navigate to another view or to another dialog.

Instead we could have some buttons that expand new boxes that display configuration of the panel, variable ...etc. Like a desk drawers. You can open every drawers and access to all of then at once.

I know it's not the topic, but we never talked about the actual design of the dashboard. And after some usage about the edition of a dashboard, I wanted to share a bit my opinion around that.

Sorry for highjacking your PR @eunicorn

eunicorn:
> Could you explain when this dialog should appear ?

Nothing triggers this dialog to appear in this PR yet, it's one of the TO DOs.

eunicorn:
> Being NOT able to see another variable configuration at the same time I'm adding or editing another one was something that super annoy me in Grafana.

Are you specifically referring to how editing a template variable opens up a new drawer? This is definitely good feedback, i'll bring it up to our designer and let @saminzadeh know

ui/components/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export interface DialogButtonProps {
+  name: string;
+  onClick: () => void;
+}
+
+const DIALOG_DEFAULT_WIDTH = 500;
+
+export const Dialog = ({ isOpen, title, children, onClose, primaryButton, secondaryButton }: DialogProps) => {
```
LukeTillman, line 33: Just a general thought, but I know we're debating a bit whether we want to really be building a "component library" for general UI components and this kinda feels like that. Should we hold off on making this a component for now until we think about that some more?

For this component in particular, every time we've tried to create a single generic "Dialog" component in our internal codebase, we've basically ended up getting rid of it because it's really hard to build one that covers all use cases. We've typically ended up going back to something more composable like Dialog, DialogTitle, DialogActions, etc. (i.e. what's in MUI or some slightly custom version of it). I think that might be the other reason to hold off on doing this now.

Anyway, just a couple general things I was thinking about when I skimmed this real quick. Happy to chat more if you want to talk it through!

Nexucis:
> Are you specifically referring to how editing a template variable opens up a new drawer?

I think the variables are the easier to modify. But I'm actually talking about the whole dashboard. Editing the variable is an easy example.

I could see editing the panel directly in the dashboard view without passing through a dialog too. It's just harder in term of design to make it smooth and beautiful.

ui/components/src/Dialog/Dialog.tsx
eunicorn: makes sense, I removed the generic dialog component

eunicorn:
@Nexucis  I updated the PR so now the unsaved changes dialog will show up when clicking cancel button, but still need to figure out how to make it show when navigating away by clicking the Perses button in another PR

Nexucis:
ah cool, thank you for doing that @eunicorn :)

Nexucis:
> but still need to figure out how to make it show when navigating away by clicking the Perses button in another PR

Perhaps an easier way would be to add a sort of URL Watching that would trigger this confirmation dialogue when the URL is going to change.

Currently on the main branch, you can change the page by clicking on the `Perses` title, but if #702 or #697 are merged, then you have other way to move to a different page.
And probably in the future we would have other pages too.

So It would be better to have a generic way to trigger this confirmation dialogue. WDYT @eunicorn ?

ui/dashboards/src/components/UnsavedChangesConfirmationDialog/UnsavedChangesConfirmationDialog.tsx
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
```
sjcobb, line 1: Is this eslint-disable line still needed?

LukeTillman (approved):

sjcobb (approved):
🚢

ui/dashboards/src/components/UnsavedChangesConfirmationDialog/UnsavedChangesConfirmationDialog.tsx
eunicorn: oh good catch thanks

PR #702, opened 2022-10-28
inject a breadcrumbs in the dashboardToolbar

Nexucis:
After playing for a couple of time with the `DashboardToolbar` and trying to split it between the app and the dashboard module, I'm proposing a simpler solution: 

`DashboardToolbar` can take as a property a component that would be used for the title. It's an optional property and if the component is not present, it's just replaced by a regular title with the name of the dashboard.

Like that I'm able to inject a simple Breadcrumbs so we can navigate from a dashboard to the project.
Note that the link won't work until #697 is merged.

Perhaps later, the breadcrumbs would be more sophisticated. For example, we could list the different dashboards in the Breadcrumbs so we can navigate from one dashboard to another without passing first by the project view. This is just an idea, I'm not really sure it's interesting.


ui/app/src/views/ViewDashboard.tsx
```diff
@@ -31,6 +31,23 @@ import { useDatasourceApi } from '../model/datasource-api';
 // https://apache.github.io/echarts-handbook/en/concepts/style/#theme
 const ECHARTS_THEME_OVERRIDES = {};
 
+interface DashboardBreadcrumbsProps {
+  dashboardName: string;
+  dashboardProject: string;
+}
+
+function DashboardBreadcrumbs(props: DashboardBreadcrumbsProps) {
```
eunicorn, line 39: maybe move this to app/src/components?

eunicorn:
i like this approach, much simpler!

ui/app/src/views/ViewDashboard.tsx
sjcobb: Yeah agreed, moving this out of ViewDashboard makes sense, I do like how you used the MUI Breadcrumbs component!

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -31,6 +31,23 @@ import { useDatasourceApi } from '../model/datasource-api';
...
+  const { dashboardName, dashboardProject } = props;
+  return (
+    <Breadcrumbs sx={{ fontSize: 'large' }}>
+      <Link underline={'hover'} variant={'h2'} component={RouterLink} to={`/projects/${dashboardProject}`}>
+        {dashboardProject}
+      </Link>
+      <Typography variant={'h2'}>{dashboardName}</Typography>
```
sjcobb, line 46: Should this actually be the H1? Currently, we have 'Perses' in the future logo spot as the H1, which isn't really correct ([ex](https://webmasters.stackexchange.com/a/4523)) since it isn't unique / descriptive about the current page like using the dashboardName here would be.

sjcobb (approved):

eunicorn (approved):

PR #703, opened 2022-10-28
put the variable in a sticky header when you scroll down

### LINKED ISSUE #576: Dashboard variables as a sticky header
It would be nice if variables at the top were sticky when you scroll down the dashboard.

Firstly, it would allow users to modify their variables/filter without having to scroll back to top in a dashboard with a lot of rows and panels.
Secondly, it would ensure that users always see in the same screen the graphs and the variable values leading to this result/visualization. In a way, a graph without the variable values visible is similar to a graph without a proper legend...

WDYT?

N.B. This feature has never been implemented in Grafana, so people are sometime displaying the Prometheus label values as part of the legend ( `{{myLabel}}`) just for that. It fixes the second part of the problem, but it must be done 'manually' on each panel and target and it becomes quite verbose, repetitive or even fuzzy with a lot of targets and variables.

Nexucis:
This PR fixed #576.

I slightly modified the variable component to made it. I hope you won't mind if it's modified like that.
Perhaps if you don't want to have this behaviour by default we could add a boolean to activate it.

So if someone want to use this component outside of perses it won't be bothered by this behaviour.

https://user-images.githubusercontent.com/4548045/198618810-053ad10c-26c7-42cf-bb30-8150433a9368.mov

/cc @saminzadeh 


Nexucis:
As a side note, when you change a variable value it re-draw everything and close every row opened. Perhaps there is something to fix around that. WDYT @saminzadeh ?

saminzadeh:
I think this looks great! My only question is how does this look when the Dashboard is embedded in another app where it's not full screen. I assume it should still work ok. We could always add a prop like `sticky={true}` if its an issue.

saminzadeh:
> As a side note, when you change a variable value it re-draw everything and close every row opened. Perhaps there is something to fix around that. WDYT @saminzadeh ?

Yeah this is a known issue. I think we need to write some optimizations around when to render.

Nexucis:
> I think this looks great! My only question is how does this look when the Dashboard is embedded in another app where it's not full screen. I assume it should still work ok. We could always add a prop like `sticky={true}` if its an issue.

ah that's a good question. I have no idea, but yeah definitely adding a properties like you suggest should fix the issue if there is one.

Nexucis:
For that matters, I have been told there is an even simpler way to have a sticky header. We just need to use the css directive `position: 'sticky'` (based on https://www.w3schools.com/howto/howto_css_sticky_element.asp) but somehow it didn't work. I don't know why, maybe there is a conflict with React MUI somehow. 

Here what I tried : 

```typescript
      <Box display={'flex'} justifyContent="space-between" position={'sticky'} top={0}>
        <Stack direction={'row'} spacing={2}>
          {showVariables &&
            variableDefinitions.map((v) => (
              <Box key={v.spec.name} display={v.spec.display?.hidden ? 'none' : undefined}>
                <TemplateVariable key={v.spec.name} name={v.spec.name} />
              </Box>
            ))}
        </Stack>
      </Box>
```

Nexucis:
even if I put `position={'sticky'} top={0}` in the previous box, it doesn't work too.

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -52,16 +53,19 @@ export function TemplateVariableList() {
...
-          {showVariables &&
-            variableDefinitions.map((v) => (
-              <Box key={v.spec.name} display={v.spec.display?.hidden ? 'none' : undefined}>
-                <TemplateVariable key={v.spec.name} name={v.spec.name} />
-              </Box>
-            ))}
-        </Stack>
-      </Box>
+
+      <AppBar color={'inherit'} position={trigger ? 'fixed' : 'static'} elevation={trigger ? 4 : 0}>
```
sjcobb, line 57: Yeah +1 to Shan's comment, I really like this idea and it makes perfect sense in our app here, but with how we embed Perses internally, it will cause some problems since variables are further down in the component tree. Could an optional prop be added to turn it off just in case?

Nexucis: Yeah I like more and more this idea. In terme of API exposed Iike the Idea you can customize the behaviour of the component with simple boolean properties.
I will do it !

Nexucis:
@sjcobb @saminzadeh I added the optional props `variableIsSticky`. By default, the variable list is not sticky. And it is explicitly set at `true` in the app.

Should be good to merge right ?

I noticed in others PRs, we started to complete the Changelog with `unrelease` section. Should I do the same ?

Nexucis:
I added a changelog entry, we should be good to merge @sjcobb @saminzadeh

sjcobb (approved):
Looks good! 🚢

PR #704, opened 2022-10-28
Add PluginLoader type, export test util for mocking PluginRegistry

LukeTillman:
This is a breaking change to the `PluginRegistry` props. It introduces a `PluginLoader` interface which encapsulates the two things needed by the `PluginRegistry` in order to load plugins (`getInstalledPlugins` and `importPluginModule`). I added a helper function called `dynamicImportPluginLoader` which should reduce some of the repetition we have in the codebase where we eagerly load a plugin's JSON, but then dynamically import the plugin module code itself.

I also added a shared test helper in the `plugin-system` package (`mockPluginRegistry`) to make it easier for people writing/consuming plugins to mock out the plugins available in tests. I changed up our `dashboards` package tests to start using this when rendering tests and included one mock `TimeSeriesChart` plugin to start with. That list of mock plugins should be pretty straightforward to add to if we need more in the dashboards tests.

Next on the test helpers list is probably adding some more helpers to `plugin-system` to make mocking some of the runtime easier (e.g. variable state, time range state, etc.).

### DIFF
--- a/ui/app/src/model/bundled-plugins.ts
+++ b/ui/app/src/model/bundled-plugins.ts
@@ -10,44 +10,20 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
-// Eagerly load the metadata for the bundled plugins, but lazy-load the plugins
 import prometheusResource from '@perses-dev/prometheus-plugin/plugin.json';
 import panelsResource from '@perses-dev/panels-plugin/plugin.json';
-import { PluginRegistryProps, PluginModuleResource } from '@perses-dev/plugin-system';
-import { useCallback } from 'react';
-
-interface BundledPlugin {
-  importPluginModule: () => Promise<unknown>;
-}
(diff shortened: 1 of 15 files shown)


ui/app/src/model/bundled-plugins.ts
```diff
@@ -10,44 +10,20 @@
...
-  importPluginModule: () => import('@perses-dev/prometheus-plugin'),
-});
-BUNDLED_PLUGINS.set(panelsResource as PluginModuleResource, {
-  importPluginModule: () => import('@perses-dev/panels-plugin'),
-});
-
-/**
- * Returns props for the PluginRegistry allowing it to load plugins that are bundled with the app.
- */
-export function useBundledPlugins(): Pick<PluginRegistryProps, 'getInstalledPlugins' | 'importPluginModule'> {
```
sjcobb, line 38: Nice, looks like you were able to remove a lot of code in this PR!

sjcobb:
LGTM 🎉 I'm excited for the future runtime test utils too!

sjcobb (approved):

PR #705, opened 2022-10-28
Run queries in editor on input blur

LukeTillman:
This fixes an issue in the panel editor where every time a user types into the query editor, the state change propagates up to the `PanelPreview` and it re-runs the Prom query. With this change, we keep a local copy of the state in the editor and then only propagate those changes once the query editor input is blurred. Here's what that looks like in action:

https://user-images.githubusercontent.com/428023/198675185-150c89b4-32b8-4987-aa83-8099004a8767.mov

Once we get a "proper" query editor in (i.e. CodeMirror), we may want to consider adding something like a "Run Query" button to make the interaction more obvious to the user.

### DIFF
--- a/ui/prometheus-plugin/src/index.ts
+++ b/ui/prometheus-plugin/src/index.ts
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { PrometheusTimeSeriesQuery } from './plugins/time-series-query';
+import { PrometheusTimeSeriesQuery } from './plugins/prometheus-time-series-query';
 // @TODO: Move this to a more generic location;
 import { StaticListVariable } from './plugins/variable';
 import { PrometheusLabelNamesVariable, PrometheusLabelValuesVariable } from './plugins/prometheus-variables';
--- a/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
+++ b/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
@@ -11,26 +11,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { ChangeEvent } from 'react';
-import produce from 'immer';
+import { produce } from 'immer';
 import { Box, TextField, FormControl, InputLabel } from '@mui/material';
-import { OptionsEditorProps, DatasourceSelect, DatasourceSelectProps } from '@perses-dev/plugin-system';
-import { DEFAULT_PROM, isDefaultPromSelector, isPrometheusDatasourceSelector } from '../model';
-import { PrometheusTimeSeriesQuerySpec } from './time-series-query';
+import { DatasourceSelect, DatasourceSelectProps } from '@perses-dev/plugin-system';
+import { DEFAULT_PROM, isDefaultPromSelector, isPrometheusDatasourceSelector } from '../../model';
+import { PrometheusTimeSeriesQueryEditorProps, useQueryState } from './query-editor-model';
 
-export type PrometheusTimeSeriesQueryEditorProps = OptionsEditorProps<PrometheusTimeSeriesQuerySpec>;
-
+/**
+ * The options editor component for editing a PrometheusTimeSeriesQuery's spec.
+ */
 export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQueryEditorProps) {
   const { onChange, value } = props;
-  const { query, datasource } = value;
+  const { datasource } = value;
 
-  const handleQueryChange = (e: ChangeEvent<HTMLInputElement>) => {
-    onChange(
-      produce(value, (draft) => {
-        draft.query = e.target.value;
-      })
-    );
-  };
+  const { query, handleQueryChange, handleQueryBlur } = useQueryState(props);
 
   const handleDatasourceChange: DatasourceSelectProps['onChange'] = (next) => {
     if (isPrometheusDatasourceSelector(next)) {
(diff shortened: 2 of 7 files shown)


ui/prometheus-plugin/src/plugins/prometheus-time-series-query/query-editor-model.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export type PrometheusTimeSeriesQueryEditorProps = OptionsEditorProps<PrometheusTimeSeriesQuerySpec>;
+
+/**
+ * A hook for managing the `query` state in PrometheusTimeSeriesQuerySpec. Returns the `query` value, along with
+ * `onChange` and `onBlur` event handlers to the input. Keeps a local copy of the user's input and only syncs those
+ * changes with the overall spec value once the input is blurred to prevent re-running queries in the panel's preview
+ * every time the user types.
+ */
+export function useQueryState(props: PrometheusTimeSeriesQueryEditorProps) {
```
LukeTillman, line 28: This file has the really the interesting changes in it. The rest of the PR changes were mostly just moving the `PrometheusTimeSeriesQuery` plugin into its own folder.

ui/prometheus-plugin/src/index.ts
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { PrometheusTimeSeriesQuery } from './plugins/time-series-query';
+import { PrometheusTimeSeriesQuery } from './plugins/prometheus-time-series-query';
```
cndonovan, line 14: noice!

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/query-editor-model.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { onChange, value } = props;
+
+  // Local copy of the query's value
+  const [query, setQuery] = useState(value.query);
+
+  // This is basically "getDerivedStateFromProps" to make sure if spec's value changes external to this component,
+  // we render with the latest value
```
cndonovan, line 35: What kind of external change might cause this?

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/query-editor-model.ts
```diff
@@ -0,0 +1,59 @@
...
+  const [lastSyncedQuery, setLastSyncedQuery] = useState(value.query);
+  if (value.query !== lastSyncedQuery) {
+    setQuery(value.query);
+    setLastSyncedQuery(value.query);
+  }
+
+  // Update our local state's copy as the user types
+  const handleQueryChange: TextFieldProps['onChange'] = (e) => {
+    setQuery(e.target.value);
```
cndonovan, line 44: setLastSyncedQuery here too? Though still trying to understand how it comes up

cndonovan (approved):
huzzah!

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/query-editor-model.ts
LukeTillman: Any component that's higher up in the tree also has access to getting/setting the spec state (going all the way up to the `PanelDrawer` itself). So the prop value that's getting passed in _could_ get changed higher up in the tree and become out of sync with our local state. In practice, hopefully that never happens (cause it's gonna make the state flows really weird).

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/query-editor-model.ts
LukeTillman: Not sure if the comment above helps explain at all, but here's an example from the new React docs that's doing the same thing: https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -11,26 +11,20 @@
...
+  const { datasource } = value;
 
-  const handleQueryChange = (e: ChangeEvent<HTMLInputElement>) => {
-    onChange(
-      produce(value, (draft) => {
-        draft.query = e.target.value;
-      })
-    );
-  };
+  const { query, handleQueryChange, handleQueryBlur } = useQueryState(props);
```
cndonovan, line 27: Just realized, we could get verbose and call it `usePrometheusTimeSeriesQueryState(props)`, but I'll leave that deicison up to you (to verbose or not to verbose... to be or not to be)

LukeTillman: 😂  If the hook was getting exported for use elsewhere, I'd probably go verbose. But since it's just for internal use by this one component, I think the shorter version works all right.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/query-editor-model.ts
LukeTillman: Yeah, hooks definitely made what used to be `getDerivedStateFromProps` a lot uglier than it was in the class-based component world. 😂  But I'm pretty firmly in the camp of "if I'm doing `useEffect`, I should see if there's some way to avoid it" because they often lead to weirdness.

The old React docs have an example of this too, it's just buried in the hooks FAQ: https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

sjcobb (approved):
🙏 🚀

PR #706, opened 2022-10-28
Fix file watching that was no longer detecting plugin addition

AntoineThebaud:
For some reason https://github.com/perses/perses/commit/44282beb66ca14735ecf9754ce8428886a51f1ba#diff-88358bf82ecc8c2771379b628ade8c9769b02abc1f8f39f3b81112e7c0854c35 broke the detection of plugin addition, as adding a new file/folder in the watched folder produces a CREATE event and not a WRITE.

### DIFF
--- a/internal/api/shared/schemas/loader.go
+++ b/internal/api/shared/schemas/loader.go
@@ -207,8 +207,8 @@ func (w *watcher) Execute(ctx context.Context, cancel context.CancelFunc) error 
 				return fmt.Errorf("schemas watcher channel has been closed unexpectedly")
 			}
 			// NB room for improvement: the event fsnotify.Remove could be used to actually remove the CUE schema from the map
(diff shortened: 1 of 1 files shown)


Nexucis (approved):
That's comprehensive, but I'm wondering why we didn't need to do it before.

Thanks for the fix, do you mind to do it also [here](https://github.com/perses/common/blob/main/file/watch.go#L43) ?

Probably we should double check internally in Amadeus we might have the same issue.

AntoineThebaud:
> That's comprehensive, but I'm wondering why we didn't need to do it before.
> 
> Thanks for the fix, do you mind to do it also [here](https://github.com/perses/common/blob/main/file/watch.go#L43) ?
> 
> Probably we should double check internally in Amadeus we might have the same issue.

Yes sure, I can do the PR & the checks on monday.

PR #707, opened 2022-10-28
Fix select multi value bug when options arent present

saminzadeh:
This fixes some regression that happened in #690. This ensures the default value is an empty array when the value allows for multiple. This is a good reminder to come up with some good comprehensive tests after the autocomplete refactor.


LukeTillman (approved):
Thanks for fixing this! 🚀

PR #709, opened 2022-10-31
legend / tooltip custom labels using series_name_format in query editor

sjcobb:
TimeSeriesChart panel queries can now be customized using a new optional property, `series_name_format`.  This allows users to specify how legend, tooltip, and (in the future) table labels will display using `{{example}}` syntax. See discussion on #600 for additional context.

Styles for the legend are also updated so labels wrap without setting a static maxWidth on legend items and to not wrap to multiple lines when there are more than 5 series (see hover idea TODO below).
 
### Future TODOs
- Add additional panel-level options to customize legends
  - one idea is to pass the full unformatted series name into the legend and show it on hover
- Update TimeSeriesChart panel test after #704 and runtime dependency test utils are ready
  - This PR does include a unit test for formatSeriesName in time-series-queries.test.ts

### Screen recording

https://user-images.githubusercontent.com/9369625/199014914-98995ad4-a1f5-4522-bc2b-866fb2938663.mp4

query-series-format-scrnsht


ui/plugin-system/src/model/time-series-queries.ts
```diff
@@ -41,6 +41,18 @@ export interface TimeSeriesData {
...
 
 export type TimeSeriesValueTuple = [timestamp: UnixTimeMs, value: number];
+
+export type SeriesLabels = Record<string, string>;
+
+/*
+ * Formatter used for series name display in legends and tooltips
+ * Regex replaces label {{ name }} with resolved label value
+ */
+export function formatSeriesName(inputFormat: string, seriesLabels: SeriesLabels): string {
```
sjcobb, line 55: I added this function in plugin-system since it isn't specific to PrometheusTimeSeriesQuery, but interested in feedback on where exactly it should go (core is another option). 

Also, this regex can be updated pretty easily if we want to go with something different than the handlebars `{{example}}` syntax, any alternatives that people prefer?

eunicorn (approved):

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -51,6 +52,14 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
...
         margin="dense"
       />
+      <TextField
+        fullWidth
+        label="Series Name Format"
+        value={format ?? ''}
+        onChange={handleFormatChange}
+        onBlur={handleFormatBlur}
+        margin="dense"
+      />
```
cndonovan, line 62: I think I remember in the designs that Mohit called the field "Legend" and then there was an info tooltip to explain more about the formatting syntax

cndonovan: We could always add the info tooltip in a follow-up PR

cndonovan (approved):

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
cndonovan: Ok, looking at the screenshot ... part of me also wants to move this input to the section that's titled "Legend"

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/time-series-query-model.ts
```diff
@@ -19,6 +19,7 @@ import { PrometheusDatasourceSelector, TemplateString } from '../../model';
  */
 export interface PrometheusTimeSeriesQuerySpec {
   query: TemplateString;
+  series_name_format?: string;
```
Nexucis, line 22: Could you update as well the corresponding Cue schema available here: https://github.com/perses/perses/blob/main/schemas/queries/prometheus/query.cue#L21-L31

Nexucis (commented):
Nice PR :).

I'm wondering if it would make sense to update as well the file `dev/data/dashboard.json` as well.

ui/plugin-system/src/model/time-series-queries.ts
sjcobb: Yeah that makes sense about keeping here, but moving it out of model folder--we due sometimes put functions in model files, but agreed this one makes more sense in a `utils` folder. 

Also, @LukeTillman if we decide to go with the potential new `runtime` lib would this formatSeriesName function go there?

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
sjcobb: We discussed this more offline but to summarize, the screenshot in the description only has one query but when there are multiple, each query needs to specify how series names should be formatted (some good discussion on #600 about this). There is potential that further customization is possible from the chart-level legend in the future though! Also, I'm following up about what should go in the TextField label / info tooltip

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/time-series-query-model.ts
sjcobb: Yeah definitely, I'll update this soon!

ui/plugin-system/src/model/time-series-queries.ts
LukeTillman: 🤔  Probably not? I think it would probably stay here. I was thinking `runtime` would be more for things that are related to the "runtime state of the app", but definitely something to think about some more.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/time-series-query-model.ts
LukeTillman: So the one part that feels weird here is that it sounds like maybe we want this to be an option on all time series queries, not just Prometheus ones? Or am I reading that wrong?

ui/plugin-system/src/model/time-series-queries.ts
LukeTillman: I scrolled down and see that you put this option into the Prometheus-specific types. If that's the case, this should _probably_ live in the `prometheus-plugin` package for now. If we move the option to the spec for ALL time series queries, regardless of plugin, it should probably live here. If we want it to just be a "utility" function that individual plugins could opt into using in their own options, we could move it to `core`.

LukeTillman (approved):
Overall, this looks awesome. I have a few questions on where we really intend for this to live, but nothing blocking.

ui/plugin-system/src/model/time-series-queries.ts
sjcobb: Okay that makes sense about keeping it consistent! I ended up moving it to utils.ts in ui/prometheus-plugin

It's possible it could go up a level on `kind: 'TimeSeriesQuery'`, the regex in the `formatSeriesName` function isn't really specific to Prom, but it wasn't clear to me how to make that happen as far as the editor components go? Maybe something we can follow up on

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/time-series-query-model.ts
sjcobb: Updated the cue schema and see [comment here](https://github.com/perses/perses/pull/709#discussion_r1016733954) about the TimeSeriesQuery vs. PrometheusTimeSeriesQuery part, going to chat you with some questions about how moving it to TimeSeriesQuery would potentially work

PR #710, opened 2022-10-31
Remove animation from grid items on first render

LukeTillman:
This just removes the initial animation happening with grid items on the first render of a dashboard. Here's what it looked like before:

https://user-images.githubusercontent.com/428023/199026467-2e8493db-b259-4e04-83c3-b6d9cffa8988.mov

And here's what it looks like after:

https://user-images.githubusercontent.com/428023/199026549-e830da88-5649-4b20-933b-e98f290809da.mov

This was being added by some styling from the `react-grid-layout` stylesheet, so I tried to encapsulate all the styles/logic in a `GridContainer` component.


ui/dashboards/src/components/GridLayout/GridContainer.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+import { useEffect, useState } from 'react';
+import { styled } from '@mui/material';
+
+export interface GridContainerProps {
+  children: React.ReactNode;
+}
+
+export function GridContainer(props: GridContainerProps) {
+  const [isFirstRender, setIsFirstRender] = useState(true);
```
eunicorn, line 22: what about using `useRef` here so that GridContainer doesn't render again when we set it back to false?

eunicorn: or use [this](https://usehooks-ts.com/react-hook/use-is-first-render)?

eunicorn (approved):

sjcobb (approved):

ui/dashboards/src/components/GridLayout/GridContainer.tsx
LukeTillman: Yeah, the problem is, we _do_ want it to render twice (which a ref will prevent). Otherwise animation/CSS transform stays disabled until something else triggers it to render again.

LukeTillman: Hahaha, I thought we wanted to disable animation when the grid first mounts? Right now because of some of that weird state stuff (which maybe you fixed in #711 ?) changing template variables recreates the store and so the grids get remounted, causing them to do that "grow" animation from the top-left.

The reason we don't want the CSS transforms disabled completely (which would kill all the animation I think), is because of this from the docs:

Screen Shot 2022-10-31 at 9 38 41 AM

When someone is dragging/resizing (after the initial render), I think we probably want that perf boost.

eunicorn: Right, my PR fixes the "grow" animation issue. Sorry i'm getting confused haha. Why do we want to disable animation on first mount though?

LukeTillman: All right, if your PR fixed it, then maybe this isn't necessary? I'm gonna go try out your PR. I thought we didn't want that animation transition at all when grid items first render.

ui/dashboards/src/components/GridLayout/GridContainer.tsx
```diff
@@ -0,0 +1,221 @@
...
+      {props.children}
+    </ReactGridLayoutContainer>
+  );
+}
+
+/**
+ * These are the classes needed by react-grid-layout from their CSS stylesheet.
+ */
+const ReactGridLayoutContainer = styled('section')(({ theme }) => ({
+  '& .react-grid-layout': {
```
saminzadeh, line 48: Out of curiosity, are these classes part of the "public css API" of react-grid-layout?

LukeTillman: Yup, Eunice basically copied them from the CSS stylesheet that ships with the package a couple months ago.

LukeTillman:
The initial animation still happens, but we've decided to keep it for now since animation generally is happening a lot less frequently after #711. If we want to revisit, happy to reopen this down the road.

LukeTillman:
OK, reopening this after a bunch of folks reported this during user testing.

PR #711, opened 2022-10-31
Refactor DashboardProvider and useDashboardSpec

eunicorn:
- useDashboardSpec (renamed to useDashboardResource) now returns the dashboard resource, which makes it easier for save/export to work.  
- rather than saving previous state in the store, reset function now takes in dashboard resource (similar to how setVariables works)
- prevent calling createDashboardStore every time DashboardProvider renders

### DIFF
--- a/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
@@ -30,11 +30,10 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
   const { dashboardName, onEditButtonClick, onCancelButtonClick } = props;
 
   const { isEditMode, setEditMode } = useEditMode();
-  const { openAddPanelGroup, openAddPanel, save } = useDashboardActions();
+  const { openAddPanelGroup, openAddPanel } = useDashboardActions();
   const isLaptopSize = useMediaQuery(useTheme().breakpoints.up('sm'));
 
   const onSave = () => {
-    save();
     setEditMode(false);
   };
 
--- a/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
@@ -23,7 +23,7 @@ describe('Panel Drawer', () => {
     const { store, DashboardProviderSpy } = createDashboardProviderSpy();
 
     renderWithContext(
-      <DashboardProvider initialState={{ dashboardSpec: getTestDashboard().spec, isEditMode: true }}>
+      <DashboardProvider initialState={{ dashboardResource: getTestDashboard(), isEditMode: true }}>
         <DashboardProviderSpy />
         <PanelDrawer />
       </DashboardProvider>
--- a/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
+++ b/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
@@ -17,7 +17,7 @@ import { devtools } from 'zustand/middleware';
 import { immer } from 'zustand/middleware/immer';
 import shallow from 'zustand/shallow';
 import { createContext, useContext, useState } from 'react';
-import { DashboardSpec, RelativeTimeRange } from '@perses-dev/core';
+import { DashboardResource, ProjectMetadata, RelativeTimeRange } from '@perses-dev/core';
 import { createPanelGroupEditorSlice, PanelGroupEditorSlice } from './panel-group-editor-slice';
 import { createPanelGroupSlice, PanelGroupSlice } from './panel-group-slice';
 import { createPanelEditorSlice, PanelEditorSlice } from './panel-editor-slice';
@@ -35,12 +35,12 @@ export interface DashboardStoreState
   isEditMode: boolean;
   setEditMode: (isEditMode: boolean) => void;
   defaultTimeRange: RelativeTimeRange;
-  reset: () => void;
-  save: () => void;
+  reset: (dashboardResource: DashboardResource) => void;
+  metadata: ProjectMetadata;
 }
 
 export interface DashboardStoreProps {
-  dashboardSpec: DashboardSpec;
+  dashboardResource: DashboardResource;
   isEditMode?: boolean;
 }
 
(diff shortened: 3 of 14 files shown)


ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
```diff
@@ -23,7 +23,7 @@ describe('Panel Drawer', () => {
     const { store, DashboardProviderSpy } = createDashboardProviderSpy();
 
     renderWithContext(
-      <DashboardProvider initialState={{ dashboardSpec: getTestDashboard().spec, isEditMode: true }}>
+      <DashboardProvider initialState={{ dashboardResource: getTestDashboard(), isEditMode: true }}>
```
cndonovan, line 26: 🎉

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
```diff
@@ -78,27 +89,19 @@ export function DashboardProvider(props: DashboardProviderProps) {
...
-            const { resetPanels, resetPanelGroups } = get();
-            resetPanels();
-            resetPanelGroups();
-          },
-          save: () => {
-            const { savePanels, savePanelGroups } = get();
-            savePanels();
-            savePanelGroups();
+          reset: ({ spec: { panels, layouts } }) => {
+            const { setPanels, setPanelGroups } = get();
```
LukeTillman, line 97: Do we actually need these as "public" methods on the store? Feels like reset here should just be able to call `set` on whatever state it wants since it's composed of the other slices?

Basically, if these `set` methods aren't actually called by any UI components directly, would be nice if they weren't part of the public API of the store.

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
```diff
@@ -60,14 +60,25 @@ export function useDashboardStore<T>(selector: (state: DashboardStoreState) => T
 }
 
 export function DashboardProvider(props: DashboardProviderProps) {
+  const [store] = useState(createDashboardStore(props)); // prevent calling createDashboardStore every time it rerenders
```
LukeTillman, line 63: Yay! 👏 

One small suggestion: use the callback form here to avoid `createDashboardStore` getting run every time this renders.

ui/dashboards/src/context/DashboardProvider/panel-group-slice.ts
```diff
@@ -171,3 +118,43 @@ export function createPanelGroupSlice(
     },
   });
 }
+
+function convertLayoutsToPanelGroups(
```
LukeTillman, line 122: If you end up being able to remove the `setPanels` and `setPanelGroups` methods from the public API, you might just need to export this for use in `reset`.

cndonovan, line 122: Nice

LukeTillman (approved):
Couple of minor suggestions, but overall looks great! So glad you finally tackled the store creation issue.

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
LukeTillman: Oh, also: what if the prop values change? Right now it seems like we'd just ignore it?

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -30,26 +30,25 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
...
-    if (originalSpec) {
-      resetSpec(originalSpec);
+    if (originalDashboard) {
+      resetDashboardResource(originalDashboard);
     }
     setEditMode(false);
   };
 
   const onSave = () => {
-    save();
```
cndonovan, line 52: I assume the state for panels / panel groups is still being set at some point in the editing process, so it's OK we don't call `save` here?

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -30,26 +30,25 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
...
+    setOriginalDashboard(dashboardResource);
     setEditMode(true);
   };
 
   const onCancelButtonClick = () => {
     // Reset to the original spec and exit edit mode
-    if (originalSpec) {
-      resetSpec(originalSpec);
+    if (originalDashboard) {
+      resetDashboardResource(originalDashboard);
```
cndonovan, line 46: I'm pumped for these changes, thank you for doing this.

Some overall thoughts:
- I'm not opposed to calling `useDashboard` (vs. `useDashboardResource`). I think the caller might appreciate the simplicity.
- I'm also not opposed to calling it `setDashboard`, (vs. `resetDashboard`), because I think that's essentially what it's doing? Taking whatever is given to it and setting it. 
- I think this has been planned already, but I think it would be cool if the management for the `originalDashboard` eventually moves into the provider/store, so that the caller doesn't have to do any of that. Then there could be functions like `reset` that internally take care of setting it back to the "original dashboard" (the dashboard at the time that edit mode begins).

cndonovan (approved):

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
eunicorn: > Oh, also: what if the prop values change?

hmm how about using a `key` for now to remount DashboardProvider?

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
eunicorn: yes every time you click the edit button, the state is saved in `originalDashboard`

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
eunicorn: i was debating the two names too, and i think i like `useDashboard` better :)

> * I think it would be cool if the management for the `originalDashboard` eventually moves into the provider/store

the issue right now is that template variables and dashboard are two different stores, so essentially we will need another provider to wrap both template variables and dashboard stores if we don't want the caller to worry about `originalDashboard`

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
eunicorn: oh good point we can set the panels and groups here directly

PR #714, opened 2022-10-31
Adjust gauge chart radius to avoid clipping

juliepagano:
Prior to this fix, the gauge chart would sometimes be clipped on the horizontal edges depending on the size of the chart.

## Notes for reviewers

- This change was made previously to try to fit more tightly in small screens, but led to inconsistencies with clipping.
- This PR provides a quick fix to address the gauges looking broken in some cases.
- A better long-term fix for the tighter fit on small screens will be to change how padding within panel components is managed to move it from a constant in `Panel` to being handled by individual panel components, so we can potentially do things like hide the padding on gauges based on the size. This will be done in a follow-up by the team at some point in the future.

### Before fix

Gauges looked fine at some sizes:
Screen Shot 2022-10-31 at 1 52 04 PM

But would be clipped at the horizontal edges at other sizes:
Screen Shot 2022-10-31 at 1 52 11 PM

### After fix

Gauges are not on the horizontal edges, no matter the size:
Screen Shot 2022-10-31 at 1 55 28 PM
Screen Shot 2022-10-31 at 1 55 42 PM

### DIFF
--- a/ui/components/src/GaugeChart/GaugeChart.tsx
+++ b/ui/components/src/GaugeChart/GaugeChart.tsx
@@ -53,7 +53,7 @@ export function GaugeChart(props: GaugeChartProps) {
         {
           type: 'gauge',
           center: ['50%', '65%'],
-          radius: '100%',
+          radius: '86%',
           startAngle: 200,
           endAngle: -20,
           min: 0,
@@ -106,7 +106,7 @@ export function GaugeChart(props: GaugeChartProps) {
         {
           type: 'gauge',
           center: ['50%', '65%'],
-          radius: '114%',
+          radius: '100%',
           startAngle: 200,
           endAngle: -20,
           min: 0,


ui/components/src/GaugeChart/GaugeChart.tsx
```diff
@@ -106,7 +106,7 @@ export function GaugeChart(props: GaugeChartProps) {
         {
           type: 'gauge',
           center: ['50%', '65%'],
-          radius: '114%',
+          radius: '100%',
```
juliepagano, line 109: A maximum of 100% should avoid clipping.

ui/components/src/GaugeChart/GaugeChart.tsx
```diff
@@ -53,7 +53,7 @@ export function GaugeChart(props: GaugeChartProps) {
         {
           type: 'gauge',
           center: ['50%', '65%'],
-          radius: '100%',
+          radius: '86%',
```
juliepagano, line 56: Setting this to a value that maintains the ratio that existed previously. Let me know if you think that should be tweaked.

LukeTillman (approved):
I will defer to @sjcobb as our resident e-charts expert, but this looks great to me. Congrats on your first Perses PR! 🥳

sjcobb (approved):
LGTM thanks for fixing this!

cndonovan:
yay fixing the clipping!

Nexucis:
oooh so cool (I missed this one). Thank you so much for this one @juliepagano !

PR #718, opened 2022-11-01
Make it possible to expand/collapse queries

cndonovan:
This PR makes it possible for a user to expand/collapse queries in the panel editor.

## Notes / TODOs
Things that could make this implementation easier:
- if we had IDs on queries
- if we wanted to persist the "collapsed/not collapsed" state to the backend, because then I could just attach a field to the queries themselves

Curious if anyone has thoughts on either of these things (i.e. does that sound nice / do you think it makes sense) 

## Screenshot
Screen Shot 2022-11-01 at 3 21 30 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [FEATURE] Add a breadcrumbs as a dashboard title #702
 - [FEATURE] Add a Project view with a dashboard list section #697
+- [FEATURE] User can collapse / expand queries in the panel editor #718
 - [ENHANCEMENT] Dashboard variables list is displayed as a sticky header #703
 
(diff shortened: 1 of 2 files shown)


saminzadeh (approved):

saminzadeh:
> if we wanted to persist the "collapsed/not collapsed" state to the backend, because then I could just attach a field to the queries themselves

IMO I feel like "enabled/disabled" would be more valuable in regards to saving in the backend.

cndonovan:
> > if we wanted to persist the "collapsed/not collapsed" state to the backend, because then I could just attach a field to the queries themselves
> 
> IMO I feel like "enabled/disabled" would be more valuable in regards to saving in the backend.

Yeah that does make a lot of sense

PR #719, opened 2022-11-01
Only load TS queries if consumed variables change or are loaded

saminzadeh:
This addresses the problem that TimeSeries queries are run every time _any_ variable value state is changed. This does the following:

- Only load TS queries when the variables it depends on are loaded or changed.
- This requires an additional hook into the plugin `dependsOn` to define which variables a TS query depends on


ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -56,6 +87,7 @@ export function useTimeSeriesQueries(definitions: TimeSeriesQueryDefinition[], o
 
   return useQueries({
     queries: definitions.map((definition) => ({
+      // TODO: Need a way to load multiple plugins at once to call plugin.dependsOn for the queryKey
```
saminzadeh, line 90: @LukeTillman I think we may need a `usePlugins` here so we can load multiple plugins at once. We need to filter the variableState to only the variables that a query uses to get a queryKey but this is an async operation.

LukeTillman:
Was thinking about this and isn't this a problem that Grafana also has? Wondering if we should punt on it for now. In some ways, we're kinda recreating the "dependency array" from hooks and I wonder if we're gonna do this if we need to think about it more holistically (e.g. should you be able to specify more than just variables here)

saminzadeh:
> Was thinking about this and isn't this a problem that Grafana also has? Wondering if we should punt on it for now. In some ways, we're kinda recreating the "dependency array" from hooks and I wonder if we're gonna do this if we need to think about it more holistically (e.g. should you be able to specify more than just variables here)

@LukeTillman The main difference between Perses and Grafana is that Grafana waits until all variables are loaded to render panels and fire TS queries. IMO I believe what Perses does is better since one slow variables doesn't block the loading of all other panels. However, the downside is that queries fire before variables are "ready", resulting in a bunch of invalid and/or duplicate requests.  After variables are loaded, it fires the `query()` function on any change and leaves caching up to the plugin. 

I'm ok punting for now if we come up with a solution to solve the initial load problem (maybe return an empty array instead of firing the request in the plugin itself?). That being said, the `dependsOn` hook will have more use cases beyond query caching like determining how many queries use a particular variable (e.g. building good UX around what data changes when you update a variable, deleting variables, etc.)

As far as the recreating the dependency array, that is basically what it does since plugins no longer rely on React runtime (with the exception of visual editors).

saminzadeh:
>  I thought Grafana didn't block on variables loading because they had default/selected values baked into the JSON

It was kind of a half baked solution (not documented) that's behavior changed from version to version. In the latest version, they completely ripped it out and stopped saving the options in the JSON (see [PR](https://github.com/grafana/grafana/pull/33533)). They do however allow you to set default values when saving the dashboard but still wait for all variables to load before rendering panels. Here is a quick demo in version 9 where I changed my Chrome to mimic a slow internet connection to demonstrate the problem: https://www.loom.com/share/1f31d8dcba28496494611354b36ff76b

> So agreed: definitely should solve the invalid queries thing somehow. Just a question of the best way to do it.

If we want to avoid the `dependsOn` route (framework level caching / query skipping), then it's probably best to avoid running the query in the plugin (or even caching to avoid duplicate requests). It might look something like this:

```tsx
export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQuerySpec>['getTimeSeriesData'] = async (
  spec,
  context
) => {

   /** Return empty if variables the query needs are not "ready" yet */
    const parseTemplateVariables = parseTemplateVariables(spec);
    const dependentVariablesAreNotResolved = parseTemplateVariables.some((v) => !v.value);
    if (dependentVariablesAreNotResolved) {
    // return empty result to avoid sending an invalid query
        return [];
    }

    /** cache use-case here as well to avoid sending a request every time a variable loads.
    This will avoid the multiple query problem.
    */
    ... // TODO: implement

    //... do what we usually do
};
```
Pros: 
- Don't need to modify the plugin framework yet

Cons: 
- More complexity for the plugins
- The framework will still treat this as a new data fetch, hard for the framework to tell between query skipping/caching vs normal results.

LukeTillman:
Yeah. If we go the "dependsOn" route, I wonder if it should basically just be it telling us all the things in the context it depends on somehow? Basically we'd let it give us a query key based on the context or something like that.

saminzadeh:
> Yeah. If we go the "dependsOn" route, I wonder if it should basically just be it telling us all the things in the context it depends on somehow? Basically we'd let it give us a query key based on the context or something like that.

Yeah originally I had implemented `queryKey` (basically same thing as react-query) instead of `dependsOn` for template variables but I felt like it was too low level and didn't really solve the problem of "What variables does this thing depend on?" and only solved the memoization / caching use-case.

Are you thinking `dependsOn` returns something like `{ variables, ...otherValues }`? I would think we need a more semantic queryKey than just string or array of objects.

Nexucis:
wouldn't be simpler to do the following: 

Each time a variable is changing, every query is recompiling the query using the new values. Using a hash or JSON.stringify you can compare with the previous compiled query to see if something change. If it is the case, then you call the datasource.

Nexucis:
and also it's a generic behaviour that can be applied for any kind of query

saminzadeh:
> Each time a variable is changing, every query is recompiling the query using the new values. Using a hash or JSON.stringify you can compare with the previous compiled query to see if something change. If it is the case, then you call the datasource.

Yes, in theory this is what it's doing but the compilation of the query happens at the plugin level and not the framework so the plugins need to tell the framework if it should update or not, or if its waiting on a dependency (e.g. a variable value). This was a decision we made to avoid having the framework know how to interpolate template variables and put that responsibility on the plugin.

> and also it's a generic behaviour that can be applied for any kind of query

Yes, Grafana does it this way.

LukeTillman:
> > Yeah. If we go the "dependsOn" route, I wonder if it should basically just be it telling us all the things in the context it depends on somehow? Basically we'd let it give us a query key based on the context or something like that.
> 
> Yeah originally I had implemented `queryKey` (basically same thing as react-query) instead of `dependsOn` for template variables but I felt like it was too low level and didn't really solve the problem of "What variables does this thing depend on?" and only solved the memoization / caching use-case.
> 
> Are you thinking `dependsOn` returns something like `{ variables, ...otherValues }`? I would think we need a more semantic queryKey than just string or array of objects.

Yeah, like for the template variables example, some of the plugins now depend on the current time range (since you added that optimization to include it in the label API endpoint calls). If the `dependsOn` in template variables knew about that dependency, we could get "refresh values on time range change" for free basically. I'm wondering if time series queries are similar in that they _might_ (probably do) depend on other runtime things like the current time range state.

saminzadeh:
> Yeah, like for the template variables example, some of the plugins now depend on the current time range (since you added that optimization to include it in the label API endpoint calls). If the dependsOn in template variables knew about that dependency, we could get "refresh values on time range change" for free basically. I'm wondering if time series queries are similar in that they might (probably do) depend on other runtime things like the current time range state.

Yeah that's not a bad idea. I suppose for the template variable time range example its easier to do at the framework level since the variable might have a definition property (`reload: "ON_LOAD" | "ON_TIME_CHANGE"`) so you don't really need the plugin to even tell you (but I suppose it could).

Even so, we can introduce the API incrementally when we need it:

```tsx
const myPlugin = {
  // .. other
  dependsOnVariables: (ctx) => (['var1']), // alias to `dependsOn` for now
  updateOnTimeRange: (ctx) => false,
}

// ^ The above is basically the same as 
const myPlugin = {
  // .. other
  dependsOn: ctx => ({
      variables: ['var1'],
      timeRange: true, // inspect ctx to set true / false
  })
}

```

Not to go on a tangent, I feel like we should implement some kind of prefix standard for stuff we are not sure about but want to try out.
```tsx
const myPlugin = {
  // .. other
  dependsOnVariables: (ctx) => (['var1']), // alias to `dependsOn` for now
  __EXPERIMENTAL__updateOnTimeRange: (ctx) => false,
}
```

saminzadeh:
Something got corrupt while try to rebase this to get DCO. I'm going to close and create a fresh PR wit the changes.

PR #720, opened 2022-11-02
removing Julius Volz from the Perses org

Nexucis:
With the agreement of @juliusv I'm removing him from the Perses org. Thank you for your help at the beginning of Perses Julius :).
Hope you will be around to try it :).

Signed-off-by: Augustin Husson <augustin.husson@amadeus.com>


juliusv:
Thanks :) :+1:

PR #721, opened 2022-11-02
Clean up imports and exports to help with package and bundle size

LukeTillman:
While looking into some bundle size issues internally, I noticed that there were a lot of third party libraries coming from Perses packages that probably shouldn't be. This made me go an take a look at our JavaScript bundles for the Perses app (i.e. `npm run analyze -w app`). This was what it looked like when I started:

Screen Shot 2022-11-02 at 11 57 17 AM

You can see we have 2.7 MB of Javascript total (before gzipping which would happen on the wire). After doing some digging around and cleaning up a bunch of things, including adding a bit of code splitting to the app, this is what things look like now:

Screen Shot 2022-11-02 at 11 58 38 AM

This ended up cutting almost 800 KB of JavaScript from our bundles. This should also help folks using the Perses packages in their own bundles. I'll add a few comments inline about a few of the more interesting changes.

### DIFF
--- a/ui/app/src/Router.tsx
+++ b/ui/app/src/Router.tsx
@@ -16,10 +16,10 @@ import { Routes, Route } from 'react-router-dom';
 import { ErrorBoundary, ErrorAlert } from '@perses-dev/components';
 // Default route is eagerly loaded
 import ViewDashboardList from './views/ViewDashboardList';
-import ViewProject from './views/ViewProject';
 
 // Other routes are lazy-loaded for code-splitting
 const ViewDashboard = lazy(() => import('./views/ViewDashboard'));
+const ViewProject = lazy(() => import('./views/ViewProject'));
 
 function Router() {
   return (
--- a/ui/app/src/Router.tsx
+++ b/ui/app/src/Router.tsx
@@ -11,16 +11,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { Suspense, lazy } from 'react';
 import { Routes, Route } from 'react-router-dom';
-import ViewDashboard from './views/ViewDashboard';
+import { ErrorBoundary, ErrorAlert } from '@perses-dev/components';
+// Default route is eagerly loaded
 import ViewDashboardList from './views/ViewDashboardList';
+
+// Other routes are lazy-loaded for code-splitting
+const ViewDashboard = lazy(() => import('./views/ViewDashboard'));
 
 function Router() {
   return (
-    <Routes>
-      <Route path="/projects/:projectName/dashboards/:dashboardName" element={<ViewDashboard />} />
-      <Route path="/" element={<ViewDashboardList />} />
-    </Routes>
+    <ErrorBoundary FallbackComponent={ErrorAlert}>
+      {/* TODO: What sort of loading fallback to we want? */}
+      <Suspense>
+        <Routes>
+          <Route path="/projects/:projectName/dashboards/:dashboardName" element={<ViewDashboard />} />
+          <Route path="/" element={<ViewDashboardList />} />
+        </Routes>
+      </Suspense>
+    </ErrorBoundary>
   );
 }
 
--- a/ui/app/src/context/DarkMode.tsx
+++ b/ui/app/src/context/DarkMode.tsx
@@ -13,8 +13,13 @@
 
 import React, { createContext, useContext, useMemo } from 'react';
 import { CssBaseline, ThemeProvider, useMediaQuery } from '@mui/material';
+import { ChartsThemeProvider, generateChartsTheme, PersesChartsTheme } from '@perses-dev/components';
 import { useLocalStorage } from '../utils/browser-storage';
 import { getTheme } from '../theme/theme';
+
+// app specific echarts option overrides, empty since perses uses default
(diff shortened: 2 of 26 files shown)


ui/app/src/Router.tsx
```diff
@@ -11,16 +11,26 @@
...
 
+import { Suspense, lazy } from 'react';
 import { Routes, Route } from 'react-router-dom';
-import ViewDashboard from './views/ViewDashboard';
+import { ErrorBoundary, ErrorAlert } from '@perses-dev/components';
+// Default route is eagerly loaded
 import ViewDashboardList from './views/ViewDashboardList';
 
+// Other routes are lazy-loaded for code-splitting
+const ViewDashboard = lazy(() => import('./views/ViewDashboard'));
```
LukeTillman, line 21: Here's where I started adding some code splitting.

ui/app/src/context/DarkMode.tsx
```diff
@@ -46,12 +51,16 @@ export function DarkModeContextProvider(props: { children: React.ReactNode }) {
...
+  const chartsTheme: PersesChartsTheme = useMemo(() => {
+    return generateChartsTheme('perses', theme, ECHARTS_THEME_OVERRIDES);
+  }, [theme]);
 
   return (
     <ThemeProvider theme={theme}>
       <CssBaseline />
-
-      <DarkModeContext.Provider value={darkModeContext}>{props.children}</DarkModeContext.Provider>
+      <ChartsThemeProvider themeName="perses" chartsTheme={chartsTheme}>
```
LukeTillman, line 61: I moved `ChartsThemeProvider` up to the app-level, rather than keeping it in a view since it's basically a global provider.

ui/components/package.json
```diff
@@ -14,6 +14,11 @@
   "module": "dist/index.js",
   "main": "dist/cjs/index.js",
   "types": "dist/index.d.ts",
+  "sideEffects": [
```
LukeTillman, line 17: These files have side-effects in them because of what we're doing with `echarts.use` to keep echarts bundles small.

ui/components/src/index.ts
```diff
@@ -25,4 +25,4 @@ export * from './TimeRangeSelector';
 export * from './context/ChartsThemeProvider';
 export * from './utils';
 export * from './model';
-export * from './test';
+export * from './test-utils';
```
LukeTillman, line 28: This was exporting the entire `tests` folder before into our public API, which was causing test-only libraries like `@testing-library/react` to appear in output bundles.

ui/components/src/context/ChartsThemeProvider.tsx
```diff
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import React, { createContext, useContext } from 'react';
-import { registerTheme } from 'echarts';
+import { registerTheme } from 'echarts/core';
```
LukeTillman, line 15: This was causing all of `echarts` to get pulled in when you used the `ChartThemeProvider` 😂

ui/components/src/utils/mathjs.ts
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { roundDependencies, create } from 'mathjs';
+
+// This ensures we get a minimal mathjs bundle for just what we need (see https://mathjs.org/docs/custom_bundling.html)
+const { round } = create({ roundDependencies });
```
LukeTillman, line 17: The `mathjs` import was also a source of a ton of unused javascript. Similar to `echarts`, they provide a way to get "just what you need".

ui/dashboards/src/context/DashboardProvider/common.ts
```diff
@@ -21,13 +21,12 @@ declare global {
...
 
-if (globalThis.dashboardStoreId === undefined) {
-  globalThis.dashboardStoreId = 0;
-}
-
 /**
  * Helper function to generate unique IDs for things in the dashboard store that don't have a "natural" ID.
  */
 export function generateId() {
+  if (globalThis.dashboardStoreId === undefined) {
```
LukeTillman, line 28: I just removed some side effects here so that I could mark the overall package as side-effect free.

LukeTillman:
If anyone's interested, the webpack docs explain the `sideEffects` stuff here: https://webpack.js.org/guides/tree-shaking/

ui/app/src/context/DarkMode.tsx
sjcobb: This still works with using the MUI theme as a param to generate the ECharts theme right?

sjcobb: ohh nvm I see now, this makes sense and there's a unit test for this too

sjcobb (approved):
Dang didn't realize how messed up the bundle was 😣 awesome you figured out how to fix it! 🎉

Nexucis:
Really impressive. Kudo @LukeTillman for doing it !

PR #722, opened 2022-11-02
perf fix to pass echarts theme directly into chart init

sjcobb:
Performance bugfix to remove using registerTheme to share an ECharts theme between charts and instead pass the generated theme obj directly into our echarts wrapper component.


LukeTillman (approved):
Awesome. Thanks for the quick turnaround on this. I want to run the bundle analyzer again once this is in, but we should do a release soon.

LukeTillman:
Update: this ended up cutting the main entrypoint bundle down by 400 KB. Nice work!

PR #723, opened 2022-11-02
Memoize selectors to prevent unnecessary computations each render

eunicorn:
https://docs.pmnd.rs/zustand/recipes/recipes#memoizing-selectors


LukeTillman (approved):
Thanks!

PR #726, opened 2022-11-02
fix link to the perses team and sort members

Nexucis:
Signed-off-by: Augustin Husson <augustin.husson@amadeus.com>


LukeTillman (approved):
🙌

PR #729, opened 2022-11-03
add a button to stick the variable header

Nexucis:
I'm adding a little button to be able to remove the sticky header at runtime. I'm thinking sometimes you want the full space to see the dashboard.

https://user-images.githubusercontent.com/4548045/199757533-4442f495-5952-4ee0-a346-5d2b741a1123.mov

### DIFF
--- a/ui/dashboards/src/components/Variables/VariableList.tsx
+++ b/ui/dashboards/src/components/Variables/VariableList.tsx
@@ -12,10 +12,11 @@
 // limitations under the License.
 
 import React, { useState } from 'react';
-import { Button, Stack, Box, Drawer, AppBar, useScrollTrigger } from '@mui/material';
+import { Button, Stack, Box, Drawer, AppBar, useScrollTrigger, IconButton } from '@mui/material';
 import EyeIcon from 'mdi-material-ui/Eye';
 import PencilIcon from 'mdi-material-ui/Pencil';
-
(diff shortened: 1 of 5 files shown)


ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -26,13 +27,14 @@ interface TemplateVariableListProps {
 
 export function TemplateVariableList(props: TemplateVariableListProps) {
   const [isEditing, setIsEditing] = useState(false);
+  const [isPin, setIsPin] = useState(props.variableIsSticky);
```
saminzadeh, line 30: Super nit: It's generally good practice to name props in a way that makes it clear that the component is controlled vs [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html).

For example, the native input uses `default` prefix when the `value` of the input is uncontrolled.
```tsx
<input value="foo" /> // controlled
<input defaultValue="foo" /> // uncontrolled
```

Maybe `variableIsSticky` might be better as `initialVariableIsSticky`

saminzadeh (approved):
This is sweet UX! I also like to remove it sometimes

ui/dashboards/src/components/Variables/VariableList.tsx
Nexucis: interesting ! Makes sense I will change that.  Thanks !

PR #730, opened 2022-11-03
Get rid of deprecation warning from webpack dev server

LukeTillman:
Just gets rid of the deprecation warning from `webpack-dev-server` about the old configuration property we were using.


saminzadeh (approved):

Nexucis (approved):
nice, less warning is something that makes me happy haha

PR #731, opened 2022-11-04
Read config to set UI in readonly mode

Nexucis:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [FEATURE] Add ability to refresh dashboard #777
 - [FEATURE] Add a migration page #774
+- [ENHANCEMENT] Dashboard view can be configured to be readonly #731
 - [BREAKINGCHANGE] bump peer-dependencies @mui/material to v5.10.0 #782
 - [BREAKINGCHANGE] useTimeRange now returns timeRange and absoluteTimeRange #777
 
--- a/ui/app/src/model/config-client.ts
+++ b/ui/app/src/model/config-client.ts
@@ -1,0 +1,54 @@
+// Copyright 2022 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


LukeTillman:
Thanks for taking a stab at this, @Nexucis. I think I want to understand the use case here a bit more. Why do you see someone making Perses read-only? Is this for folks who want to use source control to manage changes to dashboards?

Nexucis:
@LukeTillman yes exactly. It would happen in case you have the dashboard in Git and then you deploy them in a storage volume read by Perses.

The deployment process could use the CRD (Custom Resource Document) in kubernetes when it will be available.

Nexucis:
So in that configuration, you don't want to see someone overriding your dashboard without passing through a PR first

eunicorn:
@Nexucis Rather than removing the edit button completely, what do you think about disabling save button instead? Users might still want to do some visual editing to update the json and then they can copy the json to open a PR in git

Nexucis:
ah yeah that could happen indeed. But then we would need an export button no ? 

Also we might want to disable the dialog that is opened when we move out of the dashboard view and we have on going changes not saved. 

Just thinking widely, but having an `edit` button and then a disable `save` button would be weird no ?

eunicorn:
Yeah we can implement an export button. This should be very easy to implement since @cndonovan already implemented it in our chronosphere product. Eventually we also want to add a json editor and users can copy the json from there as well.

> Just thinking widely, but having an `edit` button and then a disable `save` button would be weird no ?

agreed, we can show a banner that distinguishes that this is a different editing mode

eunicorn:
@Nexucis What do you think about this? We can disable the `save` button and have a message next to it saying `Dashboard managed via code only. Download JSON and commit changes to Save.`

image

Nexucis:
that sounds really cool @eunicorn !!

Nexucis:
In the current code, the `export` button would export the non changed dashboard. Because you would need to save the dashboard to change the state of the current dashboard.

So we can't currently use the `export` button to get the modified dashboard :(

eunicorn:
> In the current code, the `export` button would export the non changed dashboard. Because you would need to save the dashboard to change the state of the current dashboard.
> 
> So we can't currently use the `export` button to get the modified dashboard :(

Oops I forgot to also add the download button when in edit mode! This PR should fix that -https://github.com/perses/perses/pull/801

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [FEATURE] Add ability to refresh dashboard #777
 - [FEATURE] Add a migration page #774
+- [ENHANCEMENT] Dashboard view can be configured to be readonly
```
eunicorn, line 7: missing the pull request number here

eunicorn (approved):
awesome thanks! :)

PR #732, opened 2022-11-04
Update TimeSeriesQuery plugins to load based on variable state

saminzadeh:
This is a follow up to the draft PR #719 

> This addresses the problem that TimeSeries queries are run every time any variable value state is changed. This does the following:
> 
> Only load TS queries when the variables it depends on are loaded or changed.
> This requires an additional hook into the plugin dependsOn to define which variables a TS query depends on

Video demo: https://www.loom.com/share/211f502fef3c44f5977888dff44d5292

### DIFF
--- a/ui/plugin-system/src/runtime/time-series-queries.ts
+++ b/ui/plugin-system/src/runtime/time-series-queries.ts
@@ -63,10 +63,10 @@ function getQueryOptions({
   if (variableDependencies) {
     waitToLoad = variableDependencies.some((v) => variableState[v]?.loading);
   }
-  const pluginEnabled = plugin !== undefined && !waitToLoad;
+  const queryEnabled = plugin !== undefined && !waitToLoad;
   return {
     queryKey,
-    pluginEnabled,
+    queryEnabled,
   };
 }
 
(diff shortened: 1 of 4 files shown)


saminzadeh:
I will also do a follow up PR to make the VariablesPlugin API similar

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -14,36 +14,83 @@
...
 export interface UseTimeSeriesQueryOptions {
   suggestedStepMs?: number;
 }
 
+/**
+ * Returns a serialized string of the current state of variable values.
+ */
+function getVariableValuesKey(v: VariableStateMap) {
+  return Object.values(v)
+    .map((v) => JSON.stringify(v.value))
```
LukeTillman, line 33: Curious: is this because variables could be an array value?

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -54,17 +101,28 @@ export function useTimeSeriesQueries(definitions: TimeSeriesQueryDefinition[], o
   const { getPlugin } = usePluginRegistry();
   const context = useTimeSeriesQueryContext();
 
+  const pluginLoaderResponse = usePlugins(
```
LukeTillman, line 104: Nice job untangling this async flow. Very easy to follow!

LukeTillman (approved):
Awesome work. Gonna be a huge plus to usability.

ui/plugin-system/src/runtime/time-series-queries.ts
saminzadeh: Yes exactly

PR #735, opened 2022-11-04
add paramsEnabled prop to TimeRangeProvider and fix back btn

sjcobb:
This PR fixes the broken back button so users can return to the dashboard listing page. Props are changed on TimeRangeProvider and an optional`paramsEnabled` is added that allows users to turn off query param functionality. Folder structure is updated to be more consistent with TemplateVariableProvider.


ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.test.tsx
```diff
@@ -37,7 +37,7 @@ describe('TimeRangeControls', () => {
   const renderTimeRangeControls = () => {
     renderWithContext(
       <DashboardProvider initialState={initialState}>
-        <TimeRangeProvider timeRange={testDefaultTimeRange}>
+        <TimeRangeProvider initialTimeRange={testDefaultTimeRange} paramsEnabled={false}>
```
saminzadeh, line 40: Maybe a bit more descriptive for the prop? `enabledURLParams` (or `urlParamsEnabled` in this case)?

saminzadeh (approved):

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.test.tsx
saminzadeh: Maybe at some point, if it's a library - users might want to map the URL query params to avoid collision but we can probably punt on that for now until we figure out the public framework APIs

```tsx
<TimeRangeProvider urlParamsOptions={{ start: "from",  end: "until", prefix: "dashboard_"  }}  />
```

juliepagano (approved):

LukeTillman (approved):

PR #740, opened 2022-11-08
Move OptionsEditorTabs to plugin-system

juliepagano:

Done per feedback from @sjcobb -- thank you!


ui/plugin-system/src/components/index.ts
```diff
@@ -11,9 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
```
juliepagano, line 13: I alpha sorted these while I was adding a new entry here for ease of skimming. Let me know if there's a preference for leaving it as is and adding the new entry to the end.

sjcobb (approved):
LGTM!

PR #743, opened 2022-11-08
Update setDashboard to also set metadata

eunicorn:
Right now `setDashboard` does not actually reset metadata, so there is no way to edit dashboard name/version/others fields.

This also fixes an issue where the panel icons are missing if panel is too small

### DIFF
--- a/ui/dashboards/src/components/Panel/PanelHeader.tsx
+++ b/ui/dashboards/src/components/Panel/PanelHeader.tsx
@@ -77,11 +77,7 @@ export function PanelHeader({ id, title, description, editHandlers, isHovered, s
       aria-describedby={descriptionTooltipId}
       disableTypography
       title={
-        <Typography
-          id={titleElementId}
-          variant="subtitle1"
-          sx={{ whiteSpace: 'nowrap', overflow: 'hidden', textOverflow: 'ellipsis' }}
-        >
+        <Typography id={titleElementId} variant="subtitle1">
           {title}
         </Typography>
       }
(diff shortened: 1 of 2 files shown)


saminzadeh (approved):

Nexucis:
`metadata.name` is used to calculate the ID in the database. Which means changing this field will modify the ID which result in actually updating a different document that might or might not exist in the database.

I would suggest you don't allow the modification of the name. Or if you do so, then you need to delete the dashboard with the previous name and then create the one with the new name.

Nexucis:
Regarding the `metadata.version`, it's not a field that can be modified by the UI. This field is automatically incremented by the server after each update.

Nexucis:
Basically there is no fields in the `metadata` that be modified by the UI. There are unique or automatically calculated / generated by the backend.

So I don't think you need to do that.

eunicorn:
> Basically there is no fields in the `metadata` that be modified by the UI. There are unique or automatically calculated / generated by the backend.
> 
> So I don't think you need to do that.

I should have given more context, but this hook `setDashboard` is not used anywhere right now in the app. The use case is more for people with their own backend (like Chronosphere) to be able to configure the dashboard through the json

Nexucis:
ah ok, that's interesting. Thanks for adding a bit more context @eunicorn :).

Wouldn't it make sense that `metadata` in this method is optional ?

eunicorn:
> Wouldn't it make sense that `metadata` in this method is optional ?

The hook `useDashboard` also allows you to retrieve the latest dashboard json and the type is `DashboardResource` so I think it's better to keep the type for both getter and setter consistent

PR #746, opened 2022-11-09
Grafana migration based on CUE + New API endpoint /migrate

AntoineThebaud:
This PR adds a new way to convert grafana dashboards into Perses format, this time on the backend side.

(big PR but more than half of the added lines actually come from the JSON files used in the unit tests ;))

It consists in a new API endpoint `/migrate` that receives by POST a Grafana dashbaord in JSON format and returns the corresponding Perses dashboard JSON.
The translation is based on a remapping logic written in CUE:
- The core part is in the `mapping.cuepart`* file in the migrate endpoint package, that handles the main structure of the dashboard.
- Then the translation of all the variables, panels & queries are managed through migration files called `mig.cuepart` present in each plugin, that get merged with the core file to build the final "remapping engine". This is explained in more details the documentation file that comes with this PR.

\* *the .cuepart extension is a homemade one. I didn't put .cue because these files may contain some placeholders for string replacements, that are not valid CUE syntax.*

The conversion currently works only for the few basic types today supported in Perses (TimeseriesChart, PrometheusLabelValuesVariable etc.). If a given Grafana object has no corresponding model in Perses, it will get converted to a dummy placeholder object (see code).

Remains to do for later:
- multi grafana version management
- `migrate` command in the CLI
- (when implemented in Perses) manage the translation of row repeat & panel repeat.
- validation of the migration file when a plugin is loaded
- preload the mapping engine instead of building it at each and every request on /api/migrate (watch the plugins to trigger a refresh when needed).
- TBD: how to manage the different types of queries?
- TBD: factorize in the common package the mapping tables that are use in several plugins? (e.g units)


schemas/panels/stat/stat.cue
```diff
@@ -21,7 +21,7 @@ kind: "StatChart"
 spec: close({
 	query:       #ts_query
 	calculation: common.#calculation
-	unit:        common.#unit
```
AntoineThebaud, line 24: @sjcobb I think there is (was?) a misalignment between FE and BE here, is `unit` a mandatory or optional attribute on your end for StatChart? Same question applies to GaugeChart actually

sjcobb: In FE it will default to 'Decimal' if it isn't set so it should be optional, I think I had it wrong before, my mistake!

schemas/panels/time-series/mig.cuepart
```diff
@@ -0,0 +1,49 @@
...
+            decbytes: "Bytes"
+        }
+        if #panel.fieldConfig.defaults.unit != _|_ if #unitMapping["\(#panel.fieldConfig.defaults.unit)"] != _|_ {
+            unit: {
+                kind: #unitMapping["\(#panel.fieldConfig.defaults.unit)"]
+            }
+        }
+        if #panel.fieldConfig.defaults.thresholds != _|_ {
+            thresholds: {
+                //default_color:                                                       // TODO how to fill this field?
```
sjcobb, line 36: I'm not really sure how this would be known in the BE, but on the FE side default colors are inferred from the MUI theme of the app, I guess in BE we could hard-code a default color once we decide on a logo and color palette for the project?

schemas/panels/stat/stat.cue
AntoineThebaud: no worries! thanks for the confirmation :)

horovits:
We recently released a utility that converts Grafana dashboards from InfluxQL to PromQL. 
Although it's a different conversion here, I thought some of the parsing logic could come in handy here, so I thought of pointing it out: 
https://github.com/logzio/influxql-to-promql-converter

schemas/panels/time-series/mig.cuepart
sjcobb: Actually nvm the equivalent Grafana property to what I called default_color would be [here](https://github.com/perses/perses/blob/main/ui/app/sample-data/node-exporter-full_rev23.json#L121), I didn’t like how they had the first item as null to define the default threshold color

sjcobb: Was our [FE conversion code](https://github.com/perses/perses/tree/main/ui/app/sample-data/convert) helpful at all when you were building this out btw? Wondering what the best way for us to collaborate on this will be in the future, I’ll keep cue up to date as we add more options, but is there some way I can also specify the equivalent Grafana property?

AntoineThebaud: I must admit I didn't check it, I'll take a look tomorrow.
About your question, I added explanations in the [CUE documentation file](https://github.com/perses/perses/blob/antoinethebaud/grafana-migrate/docs/cue.md#migration-from-grafana), the entrypoint to access the grafana properties is either `#panel`, `#var` or `#target` depending on which kind of object is considered, e.g
```cue
legend: {
    position: #panel.options.legend.placement
}
```
Still, it's not always a simple remapping like this, e.g for the units as you might have seen we have to use an intermediary mapping table & some conditionals to avoid errors:
```cue
#unitMapping: {
    ms: "Milliseconds"
    s: "Seconds"
    m: "Minutes"
    h: "Hours"
    d: "Days"
    percent: "Percent"
    percentunit: "PercentDecimal"
    bytes: "Bytes"
    decbytes: "Bytes"
}
if #panel.fieldConfig.defaults.unit != _|_ if #unitMapping["\(#panel.fieldConfig.defaults.unit)"] != _|_ {
    unit: {
        kind: #unitMapping["\(#panel.fieldConfig.defaults.unit)"]
    }
}
```

sjcobb: Oh I see, this is really cool! Yeah not sure how helpful some of the FE code would be, it was mostly focused on that older node exporter dashboard, but for ex the mapping for calculations looks similar: [convert-transformations](https://github.com/perses/perses/blob/main/ui/app/sample-data/convert/convert-transformations.ts)

sjcobb (approved):

schemas/panels/markdown/mig.cuepart
```diff
@@ -0,0 +1,6 @@
+if #panel.type == "text" if #panel.options.mode == "markdown" { // the second if has to be nested, otherwise getting "undefined field: mode" error everytime the panel doesn't match the first condition
+    kind: "Markdown"
+    spec: {
+        text: #panel.options.content
+    }
+}
```
Nexucis, line 6: in every `cuepart` file, it missed an empty end of line.

Nexucis (commented):
Thanks a lot for this amazing PR @AntoineThebaud. That was hard to do it, but hopefully the future evolution of the migration won't be that hard.

I added a couple of comments, but nothing too serious.

Nexucis:
> We recently released a utility that converts Grafana dashboards from InfluxQL to PromQL. Although it's a different conversion here, I thought some of the parsing logic could come in handy here, so I thought of pointing it out: https://github.com/logzio/influxql-to-promql-converter

Thanks for sharing @horovits :). That might be useful for later.
Do you manage different grafana version in this tool ? We are currently wondering how we should manage the different version of Grafana. Or even if we should care about the version.

We could consider to support only the last version of Grafana for the migration script. A user with an old version of Grafana would have first to upgrade his dashboard to the last version before performing the conversion.

It's not decided at all, just my thought around the subject. To Be discussed.

docs/cue.md
```diff
@@ -60,5 +78,89 @@ it should contain:
...
+if #var.type == "custom" || #var.type == "interval" {
+    kind: "StaticListVariable"
+    spec: {
+        values: [ for _, option in #var.options if option.value !~ "^\\$__.*$" { // the if filters out the grafana global vars that'd be causing validation issues later (e.g "__auto_interval_sampling is used but not defined")
+            option.value
+        }]
+    }
+}
+```
+- The file has to be named `mig.cuepart`.
```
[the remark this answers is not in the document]
AntoineThebaud: Yes that was just to push some kind of governance let's say, since we expect a single file with a specific extension, but indeed we can remove the constraint on the file name.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/perses/perses
 go 1.18
 
 require (
-	cuelang.org/go v0.4.2
+	cuelang.org/go v0.4.3
```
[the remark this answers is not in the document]
AntoineThebaud: Ah I didnt pay puch attention actually, I will check again if we still have the issue from last time.

internal/api/impl/migrate/mapping.cuepart
```diff
@@ -0,0 +1,110 @@
+import (
+    "regexp"
+)
+
+kind: "Dashboard",
+metadata: {
+    name: #grafanaDashboard.title
+},
+spec: {
```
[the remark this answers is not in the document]
AntoineThebaud: yep it would make sense here indeed

schemas/panels/gauge/mig.cuepart
```diff
@@ -0,0 +1,67 @@
+if #panel.type == "gauge" {
```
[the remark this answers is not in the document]
AntoineThebaud: Yes that's sth I was thinking to tackle later, see above:

_TBD: validation of the migration file when a plugin is loaded?_

I think we can move it from TBD to TODO 😄

internal/api/impl/migrate/migrate.go
```diff
@@ -0,0 +1,199 @@
...
+
+	var listOfConditions strings.Builder
+
+	// first append an empty kind field in order to be able to append a default, fallback case for not covered plugins at the end of the conditions list (see below)
+	listOfConditions.WriteString("kind: _\n")
+
+	// process each schema plugin to convert it into a CUE Value
+	for _, file := range files {
+		migFilePath := filepath.Join(schemasPath, file.Name(), "mig.cuepart")
+		contentStr, err := os.ReadFile(migFilePath)
```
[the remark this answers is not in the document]
AntoineThebaud: 👍 makes sense

internal/api/impl/migrate/migrate.go
```diff
@@ -0,0 +1,199 @@
...
+	mappingString := string(mappingFileBytes)
+
+	// generate the blocks of conditionals from the plugins mig.cuepart files & replace the placeholders with them
+	for _, conf := range e.migrateConfs {
+		conditionals, err := getListOfConditions(conf.schemasPath, conf.defaultValue)
+		if err != nil {
+			logrus.WithError(err).Errorf("Unable to get list of conditions for %s", e.schemasConf.VariablesPath)
+		}
+		mappingString = strings.Replace(mappingString, conf.placeholderText, conditionals, -1)
+	}
```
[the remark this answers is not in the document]
AntoineThebaud: ok but in this case we need to trigger a refresh on some kind of file watch, if ever a plugin is added/modified while Perses is still running.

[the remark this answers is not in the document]
Nexucis: yeah probably. We tackle this in another PR. That would be more digest :)

Nexucis:
Closing and Reopening the PR hopping it will re trigger the different github actions

internal/api/impl/migrate/migrate.go
```diff
@@ -0,0 +1,201 @@
...
+	logrus.Tracef("mappingString after placeholders replacement: %s", mappingString)
+
+	// Build a CUE value that is just the received grafana dashboard wrapped in a definition block* & append this def to the mapping string.
+	// *By doing this we isolate the fields from the Grafana dashboard together in a common namespace, so that they are not mixed with the
```
Nexucis, line 134: you have a `*` here. That's a bit weird to have a bullet list with one point.

docs/cue.md
```diff
@@ -60,5 +78,92 @@ it should contain:
...
-		- any other field you want for this query plugin.
+    - a `datasource` field that holds the `kind` of datasource corresponding to this query type.
+    - any other field you want for this query plugin.
+
+# Migration from Grafana
+
+A Perses plugin can optionally embed a `mig.cuepart`* file that is basically describing in CUE language how to convert a given Grafana object into an instance of this plugin. In such case your plugin is considered as the Perses equivalent of this Grafana object type, and it will be used as part of the translation process when a Grafana dashboard is received on the `/migrate` endpoint.
+
+*\* this is mandatory to have it named that way. We put in place this constraint because it makes sense to have a single file containing the remapping logic, with the benefit of making the backend logic easier (no need to search for the file). It's also easier to check the migration logic of the different plugins, because you know which file to look for.* 
+*\* the .cuepart extension is a homemade one. It is used for CUE files that may contain some placeholders for string replacements, that are not valid CUE syntax.*
```
Nexucis, line 89: Probably you need to add an empty line, otherwise Markdown is considering it's the same paragraph and it will concat both on the same line.

Nexucis (approved):
just some nits. But otherall LGTM, thank you for doing it !

PR #747, opened 2022-11-09
fix decimal_places bug to show correct number of digits

sjcobb:
Partially fixs a bug Julie found when implementing #744 that has been around for a while (Thanks Julie!)

_It works in TimeSeriesChart when unit is `kind: 'Decimal'` now but other units like 'Bytes' are still broken so I'll need to dig more into this later._

See [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#parameters) about fix in Intl.NumberFormatOptions


sjcobb:
Oh I see now, I missed this:

image

juliepagano (approved):
Thank you!

PR #748, opened 2022-11-09
Always show the edit icons in edit mode (instead of on hover)

cndonovan:
When the user enters edit mode, we want to always show the panel edit icons, instead of only showing them when the user is hovering over the panel. 

In view mode, nothing changes (the user only sees the description icon when they hover over a panel). 

# Edit mode
Screen Shot 2022-11-09 at 8 10 52 AM

# View mode
Screen Shot 2022-11-09 at 8 11 01 AM
Screen Shot 2022-11-09 at 8 11 07 AM

### DIFF
--- a/ui/dashboards/src/components/GridLayout/GridTitle.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridTitle.tsx
@@ -11,11 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { useState } from 'react';
 import { Box, IconButton, Stack, Typography } from '@mui/material';
 import ExpandedIcon from 'mdi-material-ui/ChevronUp';
 import CollapsedIcon from 'mdi-material-ui/ChevronDown';
-import AddIcon from 'mdi-material-ui/Plus';
+import AddPanelIcon from 'mdi-material-ui/ChartBoxPlusOutline';
 import PencilIcon from 'mdi-material-ui/PencilOutline';
 import ArrowUpIcon from 'mdi-material-ui/ArrowUp';
 import ArrowDownIcon from 'mdi-material-ui/ArrowDown';
@@ -38,7 +37,6 @@ export interface GridTitleProps {
 export function GridTitle(props: GridTitleProps) {
   const { panelGroupId, title, collapse } = props;
 
-  const [isHovered, setIsHovered] = useState(false);
   const { openAddPanel, openEditPanelGroup, moveUp, moveDown } = usePanelGroupActions(panelGroupId);
   const { openDeletePanelGroupDialog } = useDeletePanelGroupDialog();
   const { isEditMode } = useEditMode();
@@ -58,8 +56,6 @@ export function GridTitle(props: GridTitleProps) {
         padding: (theme) => theme.spacing(1),
         backgroundColor: (theme) => theme.palette.background.default,
       }}
-      onMouseEnter={() => setIsHovered(true)}
-      onMouseLeave={() => setIsHovered(false)}
     >
       {collapse ? (
         <>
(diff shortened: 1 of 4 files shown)


eunicorn:
also we should have the same behavior for panel group?

cndonovan:
> also we should have the same behavior for editing panel group?

yes we should! (the tests reminded me of the panel group icons too, thank you for tests)

cndonovan:
> Should we do something different when panel is small? Looks a bit crowded and the panel name is cut off image

Oh yeah, I guess that's different from this one that you fixed: "Cannot see edit/delete/move icons when panel is too small"

Do you know if Mohit has had said anything / shared any designs for this?

ui/dashboards/src/components/GridLayout/GridTitle.tsx
```diff
@@ -58,30 +56,28 @@ export function GridTitle(props: GridTitleProps) {
...
           <IconButton onClick={collapse.onToggleOpen}>
             {collapse.isOpen ? <ExpandedIcon /> : <CollapsedIcon />}
           </IconButton>
           {text}
-          {isEditMode && isHovered && (
+          {isEditMode && (
             <Stack direction="row" sx={{ marginLeft: 'auto' }}>
-              <IconButton aria-label="add panel to group" onClick={openAddPanel}>
-                <AddIcon />
+              <IconButton aria-label={`add panel to group ${title}`} onClick={openAddPanel}>
```
juliepagano, line 68: Thanks for improving the a11y of these!

juliepagano (approved):
Generally looks good to me! 

I agree with @eunicorn about the tiny panels looking a little odd, but it's not a huge deal to fix immediately because it's just in edit mode. Maybe a "fit & finish" follow-up with Mohit?

cndonovan:
> Generally looks good to me!
> 
> I agree with @eunicorn about the tiny panels looking a little odd, but it's not a huge deal to fix immediately because it's just in edit mode. Maybe a "fit & finish" follow-up with Mohit?

Yeah good call. I improved it for now by adding "ellipses behavior", I gathered from [this thread](https://chronosphereio.slack.com/archives/C03M6FNCENQ/p1667603067334649) that we at least did intend to have `...` ellipses when the panel is small

ui/dashboards/src/components/GridLayout/GridTitle.tsx
```diff
@@ -58,30 +56,28 @@ export function GridTitle(props: GridTitleProps) {
...
+                <AddPanelIcon />
               </IconButton>
-              <IconButton aria-label="edit group" onClick={openEditPanelGroup}>
+              <IconButton aria-label={`edit group ${title}`} onClick={openEditPanelGroup}>
```
cndonovan, line 71: Added the `title` so that (a) I could query for these buttons in the tests and (b) because seems to me that it doesn't hurt the accessibility to say the name of the panel

ui/dashboards/src/components/GridLayout/GridTitle.tsx
```diff
@@ -58,30 +56,28 @@ export function GridTitle(props: GridTitleProps) {
...
             {collapse.isOpen ? <ExpandedIcon /> : <CollapsedIcon />}
           </IconButton>
           {text}
-          {isEditMode && isHovered && (
+          {isEditMode && (
             <Stack direction="row" sx={{ marginLeft: 'auto' }}>
-              <IconButton aria-label="add panel to group" onClick={openAddPanel}>
-                <AddIcon />
+              <IconButton aria-label={`add panel to group ${title}`} onClick={openAddPanel}>
+                <AddPanelIcon />
```
cndonovan, line 69: Updated this to match the other icon (at the top of edit mode). I'm checking with Mohit about this change too, but I think he'll be cool with it

ui/dashboards/src/components/Panel/PanelHeader.tsx
```diff
@@ -77,7 +74,15 @@ export function PanelHeader({ id, title, description, editHandlers, isHovered, s
...
-        <Typography id={titleElementId} variant="subtitle1">
+        <Typography
+          id={titleElementId}
+          variant="subtitle1"
+          sx={{
+            whiteSpace: 'nowrap',
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+          }}
+        >
```
cndonovan, line 85: This ensures that the panel name becomes "ellipsified" when it's cut off (when the panel is resized to be small)

eunicorn (approved):

PR #751, opened 2022-11-09
Wrap browser navigation in act

juliepagano:
This fixes a test unnecessarily emitting a console error like this: "Warning: An update to unstable_HistoryRouter inside a test was not wrapped in act"

## Notes for reviewers

Fixing this while I was poking at the jest console errors/warnings I mentioned in the matrix chat. This is the one that is fixable. I think the rest are related to https://github.com/remix-run/react-router/issues/8809.


cndonovan (approved):

Nexucis (approved):

sjcobb (approved):
Ah whoops, thanks for fixing this!

PR #737, opened 2022-11-08
Release v0.14.0 (#725)

Nexucis:
Signed-off-by: Steven Cobb <sjcobb.dev@gmail.com>

PR #753, opened 2022-11-10
Sync Release v0.15 with main

sjcobb:
Sync the minor release branch back into main (release/v0.15)


sjcobb:
Similar to Augustin's [comment here ](https://github.com/perses/perses/pull/737#issuecomment-1306981956), I'm going to force merge this so the changelog doesn't get too out of sync (we definitely need to come up with a better process, having the Unreleased section was super helpful for this release but we need a strategy to minimize merge conflicts)

PR #754, opened 2022-11-10
add line_width and point_radius visual options in time series panel

sjcobb:
## Overview
This PR experiments with using MUI Sliders for two new visual options in our TimeSeriesChart panel. Currently, the new properties are grouped inside 'visual' and called `line_width` and `point_radius`. These control line size and symbol size (on hover).

In the future, other options will be added inside visual, such as visual.type, 'line', 'bar', or 'scatter'. When these options are added, the ECharts series type will be customizable and point_radius will also control scatterplot [symbolSize](https://echarts.apache.org/en/option.html#series-scatter.symbolSize).

## Screenshots
image

https://user-images.githubusercontent.com/9369625/201696472-83b001e6-1106-41c6-8e2d-d3b907ba2c72.mp4


ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -23,8 +23,15 @@ export interface TimeSeriesChartOptions {
   legend?: LegendOptions;
   unit?: UnitOptions;
   thresholds?: ThresholdOptions;
+  visual?: VisualOptions;
 }
 
+export type VisualOptions = {
+  // type: 'line' | 'bar' | 'scatter'; // TODO: new option to change series type
```
sjcobb, line 30: Ability to customize series type (switch to bar and scatterplot similar to Grafana's custom.drawStyle property) will come in a future PR, I left this commented line as an example in case anyone has thoughts / questions

Nexucis:
Ah that's a really cool feature !

I think, we might want to share this kind of config across all TimeSeriesChart don't you think @sjcobb ?

schemas/panels/time-series/time-series.cue
```diff
@@ -21,12 +21,18 @@ import (
 	position?: "bottom" | "right"
 }
 
+#visual: {
+	line_width?:   number
+	point_radius?: number
```
Nexucis, line 26: does these numbers have a range ? Like they cannot be negative and they cannot be superior than 100 for example ?

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -60,7 +76,6 @@ export function createInitialTimeSeriesChartOptions(): TimeSeriesChartOptions {
         },
       },
     ],
-    legend: DEFAULT_LEGEND,
```
juliepagano, line 63: Why did this line get removed? Guessing I missed something.

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+      />
+      <Typography variant="h5">Line Width</Typography>
+      <Slider
+        aria-label="Line Width"
+        data-testid="slider-line-width"
+        defaultValue={DEFAULT_LINE_WIDTH}
+        valueLabelDisplay="auto"
+        step={0.5}
+        marks
+        min={0}
```
juliepagano, line 64: Should the min be `0.5`? At `0` the line isn't visible, which would be weird for a line chart.

schemas/panels/time-series/time-series.cue
Nexucis: just asking, because I think it's possible to put some constraints about a range in Cue.
@AntoineThebaud relying on you as our Cue expert :)

AntoineThebaud: Yep you can write sth like
```cue
line_width?: number & >0 & <100
```
or more simply
```cue
line_width?: >0 & <100
```

and to avoid repetition:
```cue
#short_positive: >=0 & <=100
line_width?: #short_positive
point_radius?: #short_positive
```
(btw if we're going to have the same constraint in other cases it could make sense to move it to the common lib).

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
sjcobb: Oh whoops I meant to call this out or I probably should have done it in a separate PR, this was an oversight I made when responding to feedback on the original legend options PR [here](https://github.com/perses/perses/pull/660#discussion_r998787231). 

Since I got rid of the `legend.show` property that means if legend is populated then it will be on by default, but we want legend to be off by default

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: Yeah good call, I'll update this to `0.5`

schemas/panels/time-series/time-series.cue
sjcobb: That's really cool, yeah I had no idea how to do that in Cue, but will update time-series.cue based on props I set in MUI Slider

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
sjcobb: @Nexucis to your comment about sharing `visual` outside of panels/time-series/time-series.cue, my plan is to add this `visual.type` option in the future to switch the series type to 'bar' and 'scatter'. This would be implemented in `kind: 'TimeSeriesChart'`, so at that point I don't see this version of 'visual' being reused anywhere else (other future panels will have their own visual options though)

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Slider, Stack, Typography } from '@mui/material';
+import { VisualOptions, DEFAULT_LINE_WIDTH, DEFAULT_POINT_RADIUS } from './time-series-chart-model';
+
+export interface VisualOptionsEditorProps {
+  value: VisualOptions;
+  onChange: (visual: VisualOptions) => void;
+}
+
+export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProps) {
+  const handlePointRadiusChange = (_: unknown, sliderValue: number | number[]) => {
```
eunicorn, line 23: should type be `Event` instead of `unknown`?

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { VisualOptions } from './time-series-chart-model';
+import { VisualOptionsEditor } from './VisualOptionsEditor';
+
+describe('VisualOptionsEditor', () => {
+  const renderVisualOptionsEditor = (value: VisualOptions, onChange = jest.fn()) => {
+    render(<VisualOptionsEditor value={value} onChange={onChange} />);
+  };
+
+  const getLineWidthSlider = () => {
+    return screen.getByTestId('slider-line-width');
```
eunicorn, line 24: should we have constants for all test id so if we ever need to update it, we don't have to update in all places?

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const newValue = Array.isArray(sliderValue) ? sliderValue[0] : sliderValue;
```
eunicorn, line 24: i think it should always return `number` since `number[]` is for range slider?

eunicorn (approved):
🚀

juliepagano (approved):
🎉

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: So I fought with this for a bit and I might just be missing something, but when I remove number[] it currently throws:

image

PR #755, opened 2022-11-14
fix dependsOn for labelValues and labelNames variables

Nexucis:


saminzadeh (approved):
Nice!

PR #758, opened 2022-11-14
add new variable plugin: PrometheusPromQLVariable

Nexucis:
This PR is adding a new Variable plugin named `PrometheusPromQLVariable`. It allows user to enter a PromQL expression and then to select a `label_name` that will finally give the value for the variable.

https://user-images.githubusercontent.com/4548045/201677724-d31b951b-2329-4678-abd4-6b6e227e0ed4.mov

Side Note: Pretty cool to see it tooks me only 2h to implement a new simple variable plugin from scratch (aka without not so much knowledge about how it works).

### DIFF
--- a/ui/prometheus-plugin/src/plugins/prometheus-variables.tsx
+++ b/ui/prometheus-plugin/src/plugins/prometheus-variables.tsx
@@ -169,11 +169,12 @@ export const PrometheusPromQLVariable: VariablePlugin<PrometheusPromQLVariableOp
     const { data: options } = await client.instantQuery({
       query: replaceTemplateVariables(spec.expr, ctx.variables),
     });
+    const labelName = replaceTemplateVariables(spec.label_name, ctx.variables);
     let values: string[] = [];
     if (options?.resultType === 'matrix') {
-      values = capturingMatrix(options, spec.label_name);
+      values = capturingMatrix(options, labelName);
     } else if (options?.resultType === 'vector') {
-      values = capturingVector(options, spec.label_name);
+      values = capturingVector(options, labelName);
     }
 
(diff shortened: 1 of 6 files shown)


ui/prometheus-plugin/src/plugins/prometheus-variables.tsx
```diff
@@ -110,3 +161,29 @@ export const PrometheusLabelValuesVariable: VariablePlugin<PrometheusLabelValues
   OptionsEditorComponent: PrometheusLabelValuesVariableEditor,
   createInitialOptions: () => ({ label_name: '' }),
 };
+
+export const PrometheusPromQLVariable: VariablePlugin<PrometheusPromQLVariableOptions> = {
+  getVariableOptions: async (spec, ctx) => {
+    const client: PrometheusClient = await ctx.datasourceStore.getDatasourceClient(spec.datasource ?? DEFAULT_PROM);
+    // TODO we may want to manage a range query as well.
```
AntoineThebaud, line 168: I think it's ok to target `/api/v1/query` only, at least that's what grafana is doing with `query_result()`. For volatile stuff you can just use an <aggr>_over_time function in your instant query (like avg_over_time(up[1h])), and this should do the trick.

AntoineThebaud (approved):

ui/prometheus-plugin/src/plugins/prometheus-variables.tsx
saminzadeh: Yeah maybe we can start simple then add it in.

saminzadeh (approved):
LGTM!

PR #760, opened 2022-11-14
Add query editor to stat and gauge charts

LukeTillman:
This just adds the query editor to both stat and gauge charts while leaving the raw JSON editor in a separate tab for now to allow editing the other options we haven't built editors for yet. For example, in Stat Chart:

Screen Shot 2022-11-14 at 8 08 19 AM

And in Gauge Chart:

Screen Shot 2022-11-14 at 8 09 32 AM


sjcobb (approved):
LGTM!

PR #762, opened 2022-11-15
Add UI for configuring stat panel options

juliepagano:

## Notes for reviewers

- This PR follows from https://github.com/perses/perses/pull/761
- I think we might need some adjustment to the font sizing in the stat chart when the sparkline isn't enabled. This pre-exists this change, so I think it should get fixed in a separate PR. I added it to the list of issues.
- The units not always behaving as expected is a known issue that was found when building the `UnitSelector` and is being handled separately.

## Screenshots

image
image
image

### Example of the font sizing issue

image


saminzadeh (approved):
Awesome!

PR #764, opened 2022-11-15
Export json button for dashboard 

eunicorn:
downloadJson

- Creates a blob url each time the dashboard changes (and deletes the previous blob url)
- Renders a hidden link
- When user clicks on the download button, we programmatically click on the hidden link with the download attribute and the correct blob url (the download attribute is the API that triggers putting the file into the downloads folder)

credits to @cndonovan for original implementation :)

### DIFF
--- a/ui/dashboards/src/components/DownloadButton/DownloadButton.tsx
+++ b/ui/dashboards/src/components/DownloadButton/DownloadButton.tsx
@@ -1,0 +1,44 @@
+import React, { useRef } from 'react';
+import DownloadIcon from 'mdi-material-ui/DownloadOutline';
+import { IconButton, styled } from '@mui/material';
+import { useDashboard } from '../../context';
+
+// Button to download the dashboard as a JSON file.
+export function DownloadButton() {
+  const { dashboard } = useDashboard();
+  const hiddenLinkRef = useRef<HTMLAnchorElement>(null);
+
+  const onDownloadButtonClick = () => {
+    if (!hiddenLinkRef || !hiddenLinkRef.current) return;
+    // Create blob URL
+    const hiddenLinkUrl = URL.createObjectURL(
+      new Blob([JSON.stringify(dashboard)], {
+        type: 'application/json',
+      })
+    );
+    // Simulate click
+    hiddenLinkRef.current.href = hiddenLinkUrl;
+    hiddenLinkRef.current.click();
+    // Remove blob URL (for memory management)
+    URL.revokeObjectURL(hiddenLinkUrl);
+  };
+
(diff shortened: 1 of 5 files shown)


ui/dashboards/src/components/DownloadButton/DownloadButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      {/* eslint-disable jsx-a11y/anchor-is-valid  */}
+      <a ref={hiddenLinkRef} style={{ display: 'none' }} download={`${dashboard.metadata.name}.json`} />
+    </>
+  );
+}
+
+const DownloadIconButton = styled(IconButton)(({ theme }) => ({
+  border: `1px solid ${theme.palette.grey[300]}`,
+  borderRadius: theme.shape.borderRadius,
+  padding: '4px',
```
saminzadeh, line 55: Nit: use a design token instead?
```
padding: theme.spacing[1],
```

saminzadeh (approved):
LGTM! I also feel like the JSON editor inline would be really nice too and will knock out 2 birds with 1 stone.

eunicorn:
> LGTM! I also feel like the JSON editor inline would be really nice too and will knock out 2 birds with 1 stone.

I think we still need to decide what code editor we want to use, right?

saminzadeh:
> I think we still need to decide what code editor we want to use, right?

We can just do the same thing we do for panel JSON right now using the TextInput.

That being said, I think we more or less have agreed on CodeMirror. That being said, if we change it I don't think its a big deal since its basically just a component that can be swapped out.

sjcobb (approved):
🙌

Nexucis (approved):
nice, thanks for doing it @eunicorn.
I'm wondering if the edit button can have the same display. Meaning, it could be just an `Icon` button with no text. It's quite straight forward to know it's an edit button with just the icon from my opinion.
Besides it would be coherent with the download button and probably with the next one : a refresh button

eunicorn:
> I'm wondering if the edit button can have the same display. Meaning, it could be just an `Icon` button with no text. It's quite straight forward to know it's an edit button with just the icon from my opinion.
> Besides it would be coherent with the download button and probably with the next one : a refresh button

I think we have the edit button bigger because it's an action that will be used more often than download or refresh buttons but let me see what our designer thinks

cndonovan:
> @Nexucis we will look into shrinking the edit button to just an icon down the line. Since it is a new workflow (as compared to grafana etc) we are introducing and I want it to be obvious.

Mohit is that you?? 😃

PR #765, opened 2022-11-15
Fix duplicate panel keys

eunicorn:
**Bug**
If you add two panels with the same name, then they reference the same panel. When you edit one panel, you'll also see the changes in the other panel

**Fix**
Create a unique panel key for each panel

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
@@ -64,6 +64,31 @@ describe('Panel Drawer', () => {
     });
   });
 
+  it('should add panel with duplicate panel name', async () => {
+    const storeApi = renderPanelDrawer();
+
+    act(() => storeApi.getState().openAddPanel());
+
+    const nameInput = await screen.findByLabelText(/Name/);
+    userEvent.type(nameInput, 'cpu');
+    userEvent.click(screen.getByText('Add'));
+
+    const panels = storeApi.getState().panels;
+    expect(panels).toMatchObject({
+      // make sure we don't have duplicate panel key by appending "-1"
+      'cpu-1': {
+        kind: 'Panel',
+        spec: {
+          display: { name: 'cpu' },
+          plugin: {
+            kind: '',
+            spec: {},
+          },
+        },
+      },
+    });
+  });
+
   it('should edit an existing panel', async () => {
     const storeApi = renderPanelDrawer();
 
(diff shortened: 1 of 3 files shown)


ui/dashboards/src/context/DashboardProvider/panel-editor-slice.ts
```diff
@@ -195,7 +195,7 @@ export function createPanelEditorSlice(): StateCreator<
         },
         applyChanges: (next) => {
           const panelDef = createPanelDefinitionFromEditorValues(next);
-          const panelKey = removeWhiteSpacesAndSpecialCharacters(next.name);
+          const panelKey = createUniquePanelKey(next.name);
```
LukeTillman, line 198: Just a suggestion (feel free to ignore), but I really like the fact that the panel keys don't have numbers by default under the old implementation. Would be really nice to only add a number if a given panel key is already taken (and then _ideally_ I think you'd add whatever the next number was starting from `2` and counting up that gets you a unique key, but maybe that's overkill).

Anyway, just a thought that I think leads to nicer/easier to read JSON.

LukeTillman (approved):

ui/dashboards/src/context/DashboardProvider/panel-editor-slice.ts
eunicorn: good point, i'll change it since the number from `generateId` will look random

LukeTillman:
One other random quick thought: be sure to update the `CHANGELOG` before merging.

ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
```diff
@@ -64,6 +64,31 @@ describe('Panel Drawer', () => {
...
+        spec: {
+          display: { name: 'cpu' },
+          plugin: {
+            kind: '',
+            spec: {},
+          },
+        },
+      },
+    });
+  });
```
cndonovan, line 90: Nice call on the test

ui/dashboards/src/context/DashboardProvider/panel-editor-slice.ts
```diff
@@ -266,3 +271,19 @@ function getYForNewRow(group: PanelGroupDefinition) {
   }
   return newRowY;
 }
+
+// Find all the unique panel keys
+// ex: cpu, cpu-1, cpu-2 count as the same panel key since these panels have the same name
+function getUniquePanelKeys(panels: Record<string, PanelDefinition>): Record<string, number> {
```
cndonovan, line 277: This is also totally optional, but I would think of this function as more `getPanelNameCounts` or `getPanelNameToCount`

(versus getting the unique panel keys, because it doesn't exactly do that)

But I can totally follow the code and see what's going on, so I think it's fine either way

cndonovan (approved):
✨

PR #767, opened 2022-11-15
add props to support custom y axis label, min, and max

sjcobb:
Make TimeSeriesChart panel / chart-level updates to support future visual editing options (form controls will come in a follow up PR), currently proposed new y_axis properties are:
```
y_axis?: {
  unit?: UnitOptions
  label?: string;
  min?: number;
  max?: number;
}
```
These properties are all optional. I'm currently debating whether to move unit inside y_axis or to keep it where it is currently. ~~I'm leaning toward leaving it as is since that top level `unit` controls other values inside the tooltip (once I fix a bug).~~ If we add a y_axis_alt in the future that shows on the right side, it will need a designated `unit` to override the default.

_unit is now inside y_axis and will override the unit set in the top level panel spec_

### Screenshot

image


ui/components/src/LineChart/LineChart.tsx
```diff
@@ -20,6 +21,7 @@ import type {
   LineSeriesOption,
   LegendComponentOption,
   VisualMapComponentOption,
+  YAXisComponentOption,
```
sjcobb, line 24: This is a typo in one of the types that echarts exports, I'm probably going to submit a PR in their repo to fix it

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -194,7 +210,7 @@ export function LineChart({ height, data, unit, grid, legend, visualMap, onDataZ
     };
 
     return option;
-  }, [data, grid, legend, visualMap]);
+  }, [data, yAxis, grid, legend, visualMap]);
```
sjcobb, line 213: I'm planning to refactor this in the future since the number of dependencies has grown, I also might remove the ability to customize [visualMap](https://echarts.apache.org/en/option.html#visualMap) since we didn't end up using it internally

saminzadeh:
> I'm currently debating whether to move unit inside y_axis or to keep it where it is currently.

I lean towards inside since the units are specific to the y_axis. However, it's not the end of the world if not

If we add another y_axis in the future, I would assume it would contain something like a "series matcher" to figure out which series apply to the second y-axis so the tooltip can consume the metadata from there

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -34,19 +34,32 @@ export type TimeSeriesChartProps = PanelProps<TimeSeriesChartOptions>;
...
   } = props;
 
   // popuate default 'position' and other future properties
   const legend = props.spec.legend ? merge({}, DEFAULT_LEGEND, props.spec.legend) : undefined;
 
-  const unit = props.spec.unit ?? DEFAULT_UNIT;
+  // TODO: eventually remove props.spec.unit, add support for y_axis_alt.unit
+  let unit = DEFAULT_UNIT;
+  if (props.spec.y_axis?.unit) {
+    unit = props.spec.y_axis.unit;
```
sjcobb, line 47: @saminzadeh good points about unit / how the tooltip should work in the future, I ended up adding `unit` inside y_axis here. Currently this will override the top level spec.unit, but eventually I'll probably get rid of spec.unit completely

schemas/common/axis.cue
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+#y_axis: {
```
Nexucis, line 16: ah I'm wondering if it makes sense it is not related to the timeseries chart actually. Or do you foresee we will use it for another chart ? 

@AntoineThebaud I think we should use `close` here right ? Just so it won't be possible to add more unknown field there.
Also maybe the migration script needs to be updated. WDYT @AntoineThebaud ?

sjcobb: SGTM I just moved it to time-series.cue, if we need it in another chart later, we can always move it to common then

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -150,6 +163,17 @@ export function LineChart({ height, data, unit, grid, legend, visualMap, onDataZ
 
     const rangeMs = data.rangeMs ?? getDateRange(data.xAxis);
 
+    const yAxisDefault = {
```
juliepagano, line 166: Nit: I think we've mostly been using SCREAMING_SNAKE for constants, so this might be more consistent as `Y_AXIS_DEFAULT` or something like that.

Also, since this doesn't depend on any of the props, could it get pulled outside of the function to avoid being redefined on every render cycle?

juliepagano (approved):
One small minor comment, but otherwise looks good to me. 🚀

schemas/common/axis.cue
AntoineThebaud: No it's ok like this @Nexucis, actually to have this closed struct restriction you have to options:

1. use the close() function
```cue
a: close({
	a1: number
	a2?: string
})
```
2. use a definition
```cue
#a: {
	a1: number
	a2?: string
}
a: #a
```

I used close() in some places because I found it less cumbersome for those, but most of the time you should use the definitions syntax (especially if you reuse it in different places)

PR #769, opened 2022-11-16
Improve panel options editing UI

juliepagano:
- Add shared components that manage the layout, spacing, and other common user interface considerations for panel options editing.
- Use those components in time series, gauge, state panel option editors.

## Notes for reviewers

- Highly recommend checking the "hide whitespace" option when reviewing this PR. It's much easier to make sense of that way.
- This change could cause some merge conflict frustrations if anyone is touching the time series. Happy to hold off on something merging and fixing that here if that's easier on folks.
- There's likely a lot more we could do to customize these components. I'm starting with the bare minimum to do the following. We can iterate more in the future.
  - Introduce a UI that looks nicer and is closer to the mocks.
  - Refactor a bunch of boilerplate into common components for consistency and to hopefully reduce churn on adjusting the UI going forward.
- As a side effect, this fixes a few a11y cases where the label and control weren't associated with each other.

### Time series panel
The improvement is most noticeable here because this panel has the most options enabled.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<th>
image

</th>
<th>
image
</th>
</tr>
</table>

### Gauge panel
<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<th>
image

</th>
<th>
image

</th>
</tr>
</table>

### Stats panel
<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<th>
image
</th>
<th>
image

</th>
</tr>
</table>


ui/components/src/OptionsEditorLayout/OptionsEditorColumn.tsx
```diff
@@ -0,0 +1,33 @@
...
+   */
+  children: React.ReactNode;
+};
+
+/**
+ * Lay out content in a column within panel options.
+ */
+export const OptionsEditorColumn = ({ children }: OptionsEditorColumnProps) => {
+  return (
+    <Grid item xs={4}>
```
juliepagano, line 29: Starting with the assumption that the content is laid out in a 3 column grid based on mocks. I imagine this will evolve over time and need some props to adjust, but not accounting for it until we need it.

ui/components/src/OptionsEditorLayout/OptionsEditorControl.tsx
```diff
@@ -0,0 +1,38 @@
...
+import React from 'react';
+import { useId } from '../utils';
+
+export type OptionsEditorControlProps = Pick<FormControlLabelProps, 'label' | 'control'>;
+
+export const OptionsEditorControl = ({ label, control }: OptionsEditorControlProps) => {
+  // Make sure we have a unique ID we can use for associating labels and
+  // controls for a11y.
+  const generatedControlId = useId('EditorSectionControl');
+  const controlId = `${generatedControlId}-control`;
```
juliepagano, line 24: This also will remove the need to make sure we're giving things unique ids by hand, which could get annoying as start reusing some of the editor components (e.g. unit).

ui/components/src/OptionsEditorLayout/OptionsEditorControl.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const controlProps = {
+    id: controlId,
+  };
+
+  return (
+    <FormControl>
+      <Stack direction="row" alignItems="center" justifyContent="space-between">
+        <FormLabel htmlFor={controlId}>{label}</FormLabel>
+        <Box sx={{ width: '160px', textAlign: 'right' }}> {React.cloneElement(control, controlProps)}</Box>
```
juliepagano, line 34: This `cloneElement` bit allows us to add the `id` prop to the passed in component. This is what MUI does under the hood in a bunch of their components.

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -37,32 +38,37 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
...
-      <Slider
-        data-testid={VISUAL_CONFIG.point_radius.testId}
-        defaultValue={DEFAULT_POINT_RADIUS}
-        valueLabelDisplay="auto"
-        step={0.5}
-        marks
-        min={0}
-        max={20}
-        onChange={handlePointRadiusChange}
+    <OptionsEditorGroup title="Visual">
```
juliepagano, line 41: These are the ones I'm least sure about the look/feel because we don't have mocks for the sliders. I think it looks ok to start, and we may want to tweak it based on follow-ups with a designer.

ui/components/src/OptionsEditorLayout/OptionsEditorGroup.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children: React.ReactNode;
+};
+
+/**
+ * Group similar content within panel options.
+ */
+export const OptionsEditorGroup = ({ title, children }: OptionsEditorGroupProps) => {
+  return (
+    <Box>
+      <Typography
```
sjcobb, line 30: Looks so much better than how it was previously 🙌

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: Yeah I can follow up on this to get feedback, I do really like the adjustments you made though at first glance

ui/components/src/OptionsEditorLayout/OptionsEditorControl.tsx
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { FormControl, FormLabel, FormControlLabelProps, Stack, Box } from '@mui/material';
+import React from 'react';
+import { useId } from '../utils';
+
+export type OptionsEditorControlProps = Pick<FormControlLabelProps, 'label' | 'control'>;
+
+export const OptionsEditorControl = ({ label, control }: OptionsEditorControlProps) => {
```
sjcobb, line 20: Would there ever be a situation where we wouldn't want to use OptionsEditorControl? It looks really flexible, but for an option like thresholds in the future, I guess we'll still be able to build that completely custom right? 
See screenshot here for ex: 

image

sjcobb (approved):
This is great and I love the before and after pics 🎉
I'm going to go ahead and merge and start using it in my y_axis options branch

ui/components/src/OptionsEditorLayout/OptionsEditorControl.tsx
juliepagano: I think we'll need to build some additional components for more complex cases like that, but hopefully we can leverage some of the same techniques.

PR #770, opened 2022-11-16
Do not send empty queries 

cndonovan:
## Problem

When a query is empty, and we send a request to the metrics database, we get a 400 error. 

## Possible solution
I took the point of view that we don't want to send empty queries to the metrics database. Ways to look at it:
- An empty query indicates that a user hasn't yet input a query. They haven't told us the query that they want to run. 
- The server is telling us that this is a client-side error, and we shouldn't send this kind of request. 
- Why ping the server when we don't need to?

Feedback is welcome. 

## ~Different problem~
~I also looked into the fact that the panel gets stuck if the backend returns an error. I'm still not sure what to do about this one.~

~What seems to be going on:~
- ~When the backend returns an error (like a 400 for a bad request), `react-query` does not catch the error like I would expect... instead it throws the error and the `ErrorBoundary` for a panel has to catch it.~
- ~Even when I removed `useErrorBoundary: true` from the `useQuery` options, it still does not catch the error. It is confounding my expectations!~
- ~When the `ErrorBoundary` kicks in, the panel cannot be reset without saving a good query and refreshing the whole page. This is because the logic to run the query, as well as the logic to render a panel, has errored.~

~Ideally, I think we want `react-query` to catch the error for us, so that the panel doesn't error out.~

~We could also add a try-catch, but that feels like it's defeating the purpose of using `react-query`.~

^ Shan adding `resetKeys` to the `ErrorBoundary` should fix this for now.

### DIFF
--- a/ui/dashboards/src/components/Panel/PanelHeader.tsx
+++ b/ui/dashboards/src/components/Panel/PanelHeader.tsx
@@ -78,6 +78,10 @@ export function PanelHeader({ id, title, description, editHandlers, isHovered, s
           id={titleElementId}
           variant="subtitle1"
           sx={{
+            // `minHeight` guarantees that the header has the correct height
+            // when there is no title (i.e. in the preview)
+            lineHeight: '24px',
+            minHeight: '24px',
             whiteSpace: 'nowrap',
             overflow: 'hidden',
             textOverflow: 'ellipsis',
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -48,6 +48,8 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
   const suggestedStepMs = useSuggestedStepMs(contentDimensions?.width);
   const queryResults = useTimeSeriesQueries(queries, { suggestedStepMs });
+  const loading = queryResults.some((result) => result.isLoading);
+  const hasData = queryResults.some((result) => result.data && Array.from(result.data.series).length > 0);
 
   const { setTimeRange } = useTimeRange();
 
@@ -60,16 +62,11 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     });
   };
 
-  // populate series data based on query results
-  const { graphData, loading } = useMemo(() => {
+  // Populate series data based on query results
+  const { graphData } = useMemo(() => {
     const timeScale = getCommonTimeScale(queryResults);
     if (timeScale === undefined) {
-      for (const query of queryResults) {
-        // If any query is successful, we will not show error (due to timeScale check)
-        if (query.error) throw query.error;
-      }
       return {
-        loading: queryResults.every((result) => result.isLoading),
         graphData: EMPTY_GRAPH_DATA,
       };
     }
@@ -82,12 +79,11 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     };
     const xAxisData = [...getXValues(timeScale)];
 
-    let queriesFinished = 0;
-    for (const query of queryResults) {
+    for (const result of queryResults) {
       // Skip queries that are still loading and don't have data
-      if (query.isLoading || query.data === undefined) continue;
+      if (result.isLoading || result.data === undefined) continue;
 
-      for (const timeSeries of query.data.series) {
+      for (const timeSeries of result.data.series) {
         const formattedSeriesName = timeSeries.formattedName ?? timeSeries.name;
         const yValues = getYValues(timeSeries, timeScale);
         const lineSeries = getLineSeries(timeSeries.name, formattedSeriesName, yValues, selectedSeriesName);
@@ -104,7 +100,6 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
           });
         }
       }
-      queriesFinished++;
     }
     graphData.xAxis = xAxisData;
 
@@ -127,7 +122,6 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
     return {
       graphData,
-      loading: queriesFinished === 0,
     };
   }, [queryResults, thresholds, selectedSeriesName, legend]);
 
@@ -145,6 +139,17 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
         <Skeleton variant="text" width={contentDimensions.width - 20} height={contentDimensions.height / 2} />
       </Box>
     );
+  }
+
+  // At this point, we have a response from the backend for all queries. (We're past loading === true).
+  // If we don't data from any of the queries, then check if we want to show an error.
+  // Put another way: If any queries have data, even if other queries failed, we will show the data (not the error).
+  // Unfortunately, partial errors get swallowed in this case.
+  // This could be refactored when someone takes a look at validation and error-handling.
+  if (!hasData) {
+    for (const result of queryResults) {
+      if (result.error) throw result.error;
+    }
   }
 
   const legendWidth = legend && legend.position === 'right' ? 200 : contentDimensions.width;
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -38,6 +38,8 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     contentDimensions,
   } = props;
 
+  const hasOnlyEmptyQueries = queries.every((q) => q.spec.plugin.spec.query === '');
+
   // popuate default 'position' and other future properties
   const legend = props.spec.legend ? merge({}, DEFAULT_LEGEND, props.spec.legend) : undefined;
 
@@ -127,12 +129,15 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
     return {
       graphData,
-      loading: queriesFinished === 0,
-      allQueriesLoaded: queriesFinished === queryResults.length,
+      loading: hasOnlyEmptyQueries ? false : queriesFinished === 0,
     };
-  }, [queryResults, thresholds, selectedSeriesName, legend]);
+  }, [queryResults, thresholds, selectedSeriesName, legend, hasOnlyEmptyQueries]);
 
   if (contentDimensions === undefined) {
+    return null;
+  }
+
+  if (hasOnlyEmptyQueries) {
     return null;
   }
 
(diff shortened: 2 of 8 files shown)


ui/dashboards/src/components/Panel/PanelHeader.tsx
```diff
@@ -78,6 +78,10 @@ export function PanelHeader({ id, title, description, editHandlers, isHovered, s
           id={titleElementId}
           variant="subtitle1"
           sx={{
+            // `minHeight` guarantees that the header has the correct height
+            // when there is no title (i.e. in the preview)
+            lineHeight: '24px',
+            minHeight: '24px',
```
cndonovan, line 84: This is a different bug that I just fixed while I was in here

saminzadeh:
> It's hard for the framework to tell when a query is empty/not set since the `spec` is a generic interface.

Actually this comment is not entirely true, we do know that its a `TimeSeriesQueryDefinition` but we can't assume the semantics of what `query: ""` means

cndonovan:
> > It's hard for the framework to tell when a query is empty/not set since the `spec` is a generic interface.
> 
> Actually this comment is not entirely true, we do know that its a `TimeSeriesQueryDefinition` but we can't assume the semantics of what `query: ""` means

Yeah that makes a ton of sense

cndonovan:
> So I actually think we don't want to handle this in the framework visualizations but push this down to the TimeSeriesQuery plugin (in this case [PrometheusTimeSeriesQuery](https://github.com/perses/perses/blob/main/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts#L28)). It's hard for the framework to tell when a query is empty/not set since the `spec` is a generic interface. For example, when we add a logging datasource, it's perfectly fine that a query config is `query: ""` (meaning give me all the logs)
> 
> Instead, a simple solution might be to return empty results or an error [here](https://github.com/perses/perses/blob/main/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts#L32) (before hitting the query backend):
> 
> ```ts
> export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQuerySpec>['getTimeSeriesData'] = async (
>   spec,
>   context
> ) => {
> 
> if(spec.query === null || spec.query === '') {
>    // Option 1: return empty time series result, the graph will show no time series
>    return { series: [] }
> 
>    // Option 2: or throw an error saying you need to define a query and skip the request
>    throw new Error("Invalid a query")
> }
> 
> //...rest of the normal logic flow
> ```
> 
> Our panel and visualizations can handle showing error messages to the user.
> 
> > When the backend returns an error (like a 400 for a bad request), react-query does not catch the error like I would expect... instead it throws the error and the ErrorBoundary for a panel has to catch it.
> 
> As far as error recovery goes, we probably need to fix this so it doesn't throw but handles gracefully and displays the message to the user without an ErrorBoundary

Woo thank you for these thoughts

saminzadeh:
> OK, found out what's going on... there's another place where we throw an error (after running the query), and that's what causes the errors to be thrown.
> 
> WOO

Yep! We can still use the ErrorBoundary to display the Error message if we want but need to reset it. I made that fix here: 
https://github.com/perses/perses/pull/771

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
```diff
@@ -29,6 +29,11 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
   spec,
   context
 ) => {
+  if (spec.query === undefined || spec.query === null || spec.query === '') {
+    // Do not make a request to the backend, instead return an empty TimeSeriesData
+    return { timeRange: { start: new Date(), end: new Date() }, stepMs: 0, series: [] };
```
saminzadeh, line 34: Nice! On a side note, I wonder if we change the plugin implementation to optionally allow these fields? I feel like they don't make sense to be required. cc: @LukeTillman

saminzadeh (approved):

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
cndonovan: Yeah I'm still playing with this, not sure this is best yet...

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
```diff
@@ -29,6 +29,11 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
   spec,
   context
 ) => {
+  if (spec.query === undefined || spec.query === null || spec.query === '') {
+    // Do not make a request to the backend, instead return an empty TimeSeriesData
+    return { series: [] };
+  }
```
cndonovan, line 35: I could also move this down and use the timeRange + stepMs that are calculated below, and then we don't need to make `timeRange` and `stepMs` optional... 

Open to thoughts

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
LukeTillman: That's an interesting question. I guess my question would be: if someone were writing a plugin that actually returned series and left those two fields empty/undefined, would the chart still work correctly? If so, totally cool to make them optional. If not, feels like we need some other solution here.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
LukeTillman: This turned out to be pretty straightforward. 👍 

I think the only downside here is we're not differentiating between "blocked because the query is empty" and "empty results cause there's no data". I wonder if that's an important distinction? When I was thinking through this initially, I was thinking this might be another case like @saminzadeh had with variables where `dependsOn` might come into play.

Maybe this is good enough? Or maybe it's good short-term but we need to differentiate long-term? Not sure.

sjcobb: Played around with it a little bit and noticed one weird case where when you multiple queries and one is empty, you can occasionally see 'No data' flash as the loading state, see recording here:

https://user-images.githubusercontent.com/9369625/202294746-b22df68b-c97e-4622-9045-66b79418f94a.mp4

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
saminzadeh: @cndonovan we're talking offline but I believe the answer is yes.

`timeRange` should be consumed via the TimeRangeProvider context for better consistency between datasources and panels.
`stepMs` not sure why need this but this can also be computed if not provided.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
saminzadeh: My 2 cents: I don't really think there is a need to differentiate between "Empty results" and "You must provide a query". FWIW Grafana does it this way as well and I personally don't mind that experience. 

If we go the `dependsOn` route and use our current implementation, it would basically be stuck in a loading state which is a bad experience IMO. Nice thing is that this is relatively an easy thing to change so maybe we should move forward with this and revisit if the feedback is negative.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
LukeTillman: The reason we added it originally was because the time range that really gets used might be different from the system state that was passed in, see:

https://github.com/perses/perses/blob/f0eb9e8ca896b6b158a0f9baac49ddac39f61c10/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts#L34-L43

saminzadeh: I see. I think thats fine as metadata but I believe the min-x and max-y in the Chart should still be derived from the Dashboard TimeRange provider

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -82,12 +79,11 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
 
-    let queriesFinished = 0;
-    for (const query of queryResults) {
+    for (const result of queryResults) {
       // Skip queries that are still loading and don't have data
-      if (query.isLoading || query.data === undefined) continue;
+      if (result.isLoading || result.data === undefined) continue;
 
-      for (const timeSeries of query.data.series) {
+      for (const timeSeries of result.data.series) {
```
cndonovan, line 86: Renamed `query` => `result` for consistency

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -48,6 +48,8 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
   const suggestedStepMs = useSuggestedStepMs(contentDimensions?.width);
   const queryResults = useTimeSeriesQueries(queries, { suggestedStepMs });
+  const loading = queryResults.some((result) => result.isLoading);
+  const hasData = queryResults.some((result) => result.data && Array.from(result.data.series).length > 0);
```
cndonovan, line 52: Changes:
- `useMemo` below will always attempt to crunch the data for the chart (i.e. `graphData`)
- `loading` and `hasData` will be used to decide when to show the loading state, and when to show an error state (see below)
- Errors are being caught by ErrorBoundary, which is a different discussion

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
saminzadeh: ^ That makes a lot of sense to me. I think we can think about the "1 of the 4 queries" fail use-case in the next iteration.

sjcobb: I think we should come back to the loading state too though, if one or multiple queries take a lot longer than others, it was nice how they rendered as each finished, but something we can look into later, this is a lot more clear and fixes the main issue 🙌

sjcobb (approved):

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
saminzadeh: > I think we should come back to the loading state too though, if one or multiple queries take a lot longer than others, it was nice how they rendered as each finished, but something we can look into later, this is a lot more clear and fixes the main issue

If we do that, getting the UX "right" is going to be crucial. It should be clear that there is more data inbound (almost like streaming results). 

I would caution on the side of showing the data once everything is ready to prevent confusion and simulate a batch request to a backend, almost as if they are one query returning multiple results. But agreed, definitely something to look into.

PR #771, opened 2022-11-16
Reset error boundary on spec change

saminzadeh:
This should address the issue mentioned in #770 where a 400 (or any JS error via the TimeSeriesQuery plugin) causes the Panel to be stuck in an error state. 

Since some of our panel plugins [throw uncaught errors,](https://github.com/perses/perses/blob/9b5263a29b07a3bc1a537d095c740d9a68eaebbb/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx#L48) we need a way to reset the wrapping ErrorBoundary when the spec (query) changes.

Demo: https://www.loom.com/share/02b0651e45e340d89bd02edbd2ca557a

### DIFF
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -108,7 +108,7 @@ export function Panel(props: PanelProps) {
         }}
         ref={setContentElement}
       >
-        <ErrorBoundary FallbackComponent={ErrorAlert}>
+        <ErrorBoundary FallbackComponent={ErrorAlert} resetKeys={[definition.spec.plugin.spec]}>
           {inView === true && (
             <PanelContent
               panelPluginKind={definition.spec.plugin.kind}


LukeTillman (approved):
Yay! I wonder if we need to audit all the ErrorBoundaries for this kinda thing... 🤔

cndonovan:
I also found out about this this morning: https://tanstack.com/query/v4/docs/reference/QueryErrorResetBoundary, if it interests you

saminzadeh:
> I also found out about this this morning: https://tanstack.com/query/v4/docs/reference/QueryErrorResetBoundary, if it interests you

Oh sweet - I haven't been throwing errors with react-query but I find this strategy useful if we do more of this. I feel like a place where this could be beneficial is when the Dashboard fails to render (aka throws an uncaught error) due to bad JSON, we can have something like "Open JSON to Edit"

PR #774, opened 2022-11-16
Add a migration page

Nexucis:
This PR is providing a page that will help the user to migrate a Grafana dashboard to a perses one.

~I didn't find (yet) how to contain the `JSONEditor` it has an infinite height and in this context I would prefer it's not. I tried to wrap it in a `Box` with a limited height but it didn't work. I will find the way tomorrow.~

I took the opportunity to remove the folder `sample_data`, hope nobody minds about this cleanup.

* ~It also misses the way to add the dashboard to a project as well.~
* I'm wondering if it would be interesting to have the preview of the dashboard also.

And thanks @AntoineThebaud for the fix around cue.

https://user-images.githubusercontent.com/4548045/202242821-342f2a30-3035-4e02-864a-e48c8382713f.mov


AntoineThebaud (approved):

ui/app/src/views/ViewMigrate.tsx
```diff
@@ -0,0 +1,95 @@
...
+        >
+          Upload JSON file
+          <input type="file" onChange={fileUploadOnChange} hidden />
+        </Button>
+        <TextareaAutosize
+          value={grafanaDashboard}
+          onChange={(e) => setGrafanaDashboard(e.target.value)}
+          minRows={10}
+          maxRows={20}
+          placeholder="Past your Grafana dashboard"
```
sjcobb, line 73: Typo, should be “Paste” 😄

sjcobb (approved):
Really cool! Yeah showing a preview of the converted dashboard would be neat. Also, when we have a Table panel maybe we can use that as a fallback for a lot of the unsupported panel types

Nexucis:
> Yeah showing a preview of the converted dashboard would be neat.

Alright :) I think I will adresse that in another PR.

I added a way to save a migrated dashboard.
It looks like this now:
image

LukeTillman (approved):
Looks good! The only thing I noticed is that the `package-lock.json` changed? I wouldn't expect there to be changes to that since none of the `package.json` files changed. 🤔

Nexucis:
> Looks good! The only thing I noticed is that the `package-lock.json` changed? I wouldn't expect there to be changes to that since none of the `package.json` files changed. 🤔

yeah I think I messed up this file when I added the dependency `@mui/labs` and then rollback. It's now fixed there is no change in the file `package-lock.json`.

Good catch thanks @LukeTillman

PR #775, opened 2022-11-16
bugfix tooltip in LineChart should use unit passed from panel

sjcobb:
This PR addresses a bug with the current TimeSeriesChart tooltip implementation where formatValue was not being called, (thanks for raising this @juliepagano) now `unit` is passed to the getNearbySeries function to fix this.

Fixed tooltip screenshot:
image


juliepagano (approved):
Thanks for fixing this! 🎉

PR #776, opened 2022-11-16
Ability to preview ListVariable values while editing

saminzadeh:
This adds the ability to preview the ListVariable values while editing. 
Notes:
- You currently need to click on the refresh button. The main reason was that any changes to the forms in the editors trigger a change. I can change that in the future to change only onBlur

Demo: https://www.loom.com/share/57b226c6eaed4a0c8a23b3cb99f2371d

image


ui/dashboards/src/components/Variables/variable-model.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  usePlugin,
+  useTemplateVariableValues,
+  useDatasourceStore,
+  useTimeRange,
+  VariableStateMap,
+} from '@perses-dev/plugin-system';
+import { useQuery } from '@tanstack/react-query';
+
+export function useListVariablePluginValues(definition: ListVariableDefinition) {
```
saminzadeh, line 24: This was refactored out of the <ListVariable /> component to be a standalone re-usable hook so it can be used in the edit form as well. This may even be better to put in the `plugin-system` package.

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -159,6 +209,11 @@ export function VariableEditForm({
           <SectionHeader>List Options</SectionHeader>
           <Grid container spacing={2} mb={2}>
             <Grid item xs={6}>
+              {/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
```
saminzadeh, line 212: This can be used to reload the preview whenever something is clicked outside of the Inputs
More details here: https://www.loom.com/share/9b978d45503042c1a55dd34d7b182762

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -56,6 +65,36 @@ const SectionHeader = ({ children }: React.PropsWithChildren) => (
...
 
+function VariableListPreview({ definition }: { definition: ListVariableDefinition }) {
+  const { data, isFetching, error } = useListVariablePluginValues(definition);
+  const [maxValues, setMaxValues] = useState<number | undefined>(DEFAULT_MAX_PREVIEW_VALUES);
+  const showAll = () => {
+    setMaxValues(undefined);
+  };
+  let notShown = 0;
+
+  if (data && data?.length > 0 && maxValues) {
```
juliepagano, line 76: Out of curiosity, how big can `data.length` be? Is there a world where clicking "show more" could be a HUGE number that you'd want to be able to hide again or is it usually a dozen or so more items?

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -56,6 +65,36 @@ const SectionHeader = ({ children }: React.PropsWithChildren) => (
...
+    notShown = data.length - maxValues;
+  }
```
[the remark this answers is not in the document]
saminzadeh: In the conditional, I use the `notShown` as a number in the `<Chip />` label
```tsx
label={`+${notShown} more`} // "+20 more"
```

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
saminzadeh: It could be huge and it's hard to reason about the data since its arbitrary to the individual datasource instance (eg. Company A has 40k hosts, Company B has 10 hosts). 

There are cases where they may be `10k+` values and so it might make sense to hide again or redo the query. Although, that may lock up the browser anyway which is why I set the sane default to "50" and anything after that is "here is the total number of values that are not shown, click to see more but use at your own risk" 😆

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -68,6 +107,17 @@ export function VariableEditForm({
...
 
+  const [previewKey, setPreviewKey] = React.useState(0);
+
+  const refreshPreview = () => {
+    setPreviewKey((prev) => prev + 1);
+  };
+
+  const previewSpec = useMemo(() => {
+    return getVariableDefinitionFromState(state) as ListVariableDefinition;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
saminzadeh: Yeah that's a good point, I'll add that.

The general idea is that `previewKey` needs to be explicitly updated either through UI action (refresh button) or onBlur (ClickAwayListener) so the spec gets computed and passed to the preview component to refetch the values. This prevents trying to preview the config on each keystroke change and firing a bunch of unnecessary or invalid requests while editing.

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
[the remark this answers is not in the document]
juliepagano: Ahhhh, totally missed that. Sorry about that!

ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
juliepagano: Makes sense! I could see iterating on it in the future if it causes issues for people, but not something to prematurely optimize for right now. :D

juliepagano (approved):
Looks great! 🎉

PR #777, opened 2022-11-17
Implement Refresh Dashboard

eunicorn:
Clicking refresh button should refresh the dashboard and rerun all queries.

refresh


ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -41,5 +43,26 @@ export function TimeRangeControls() {
...
+  };
+
+  return (
+    <>
+      <DateTimeRangePicker timeOptions={TIME_OPTIONS} value={timeRange} onChange={setTimeRange} />
+      <RefreshIconButton aria-label="Refresh Dashboard" onClick={handleRefresh}>
+        <RefreshIcon />
+      </RefreshIconButton>
+    </>
+  );
```
cndonovan, line 60: Do you think we should only show the button if `isRelativeTimeRange` is true? Or have it disabled and show a tooltip explaining that it can only be used for relative time ranges?

Might be a question for Mohit too

cndonovan: Oh oops this says draft, I may have jumped the gun

ui/dashboards/src/context/TimeRangeProvider/TimeRangeProvider.tsx
```diff
@@ -28,15 +28,27 @@ export interface TimeRangeProviderProps {
 export function TimeRangeProvider(props: TimeRangeProviderProps) {
   const { initialTimeRange, enabledURLParams, children } = props;
```
sjcobb, line 29: Curious can you check if setting enabledURLParams to false still works?

eunicorn: yep still works!

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -41,5 +43,26 @@ export function TimeRangeControls() {
       });
     }
   }
-  return <DateTimeRangePicker timeOptions={TIME_OPTIONS} value={timeRange} onChange={setTimeRange} />;
+
+  const handleRefresh = () => {
```
saminzadeh, line 47: Hmm, do you think we should add a method to the `useTimeRange` hook call `refresh` instead? That way, any component can easily call refresh without having this be specific to our TimeRangeControls

```tsx
const { timeRange, refresh, setTimeRange } = useTimeRange()

return <Button onClick={refresh}>Refresh</Button>
```

Calling refresh would increment a counter known as the `refreshKey` to signal signal components to refetch their queries.

I implemented this concept in the [framework prototype](https://codesandbox.io/s/dashboard-framework-concept-v2-g0qtc7?file=/src/framework/core/time.tsx:527-565) to get a general idea

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
saminzadeh: IMO I think refresh shouldn't be specific to relative time ranges. Refresh should just mean - "force queries to refetch data and recompute the time range". So even if you have an AbsoluteTimeRange, this should force queries to refetch. There are situations where queries fail, or load only partial data and want to refetch the query. 

Grafana does this as well, and I think it's a useful feature.

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
eunicorn: ooh thanks i like this idea, and it makes sense if we also want to refresh for absolute time range, i'll try it!

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
eunicorn: yep i also talked to Mohit and he agreed that it's fine to let users refresh even if it's absolute time range

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
```diff
@@ -0,0 +1,76 @@
+// Copyright 2022 The Perses Authors
```
eunicorn, line 1: Moved all time range related code from 'dashboard' to `plugin-system`

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -29,7 +31,7 @@ export const TIME_OPTIONS: TimeOption[] = [
 ];
 
 export function TimeRangeControls() {
-  const { timeRange, setTimeRange } = useDashboardTimeRange();
```
eunicorn, line 32: removed useDashboardTimeRange so we only have one hook `useTimeRange` :)

sjcobb: Yeah I definitely prefer having one hook, it simplifies the implementation a ton! We talked about this offline but for context, I added this internal version to not expose extra properties in the plugin runtime. 

@saminzadeh @LukeTillman previously I got feedback from you both / this [discussion](https://github.com/perses/perses/discussions/91#discussioncomment-1425254) that plugin system should ideally only expose absolute time, thoughts on adding the [separate absoluteTimeRange](https://github.com/perses/perses/blob/c8e8239939a175516dc03d6a4c4fcb80895762b6/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx#L24-L30) back to plugin-system? I think it's fine since TimeRangeControls also needs to use the hook and needs access to both since you can’t really derive relative time from absolute time.

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
sjcobb: Awesome that this worked without creating a circular dependency 🎉  This does mean that use-query-params become a required dependency for plugin-system, but I think that's worthwhile to not split the implementation b/w packages, can you update the plugin-system package.json?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- [FEATURE] Add ability to refresh dashboard #777
```
sjcobb, line 5: There is also a breaking change since the plugin-system timeRange type is TimeRangeValue now, for ex our internal topology view plugin expects useTimeRange's timeRange to always be absolute. It will be a quick fix though, I love how much this refactor simplifies things 💥

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
LukeTillman: Yeah, I don't love that plugin authors now have to take that dependency through `plugin-system`, but I think we're gonna be figuring out how to split this stuff up differently pretty soon with the testing convo.

LukeTillman (approved):
Looks great. The refresh key is a nice way to get the refresh functionality for absolute time ranges too 👍

sjcobb (approved):
💯

Issue #642, opened 2022-10-14
Moving from Mobile screen size to tablet laptop screen size doesn't redraw the different graph


Nexucis:
It appears when you are in a mobile screen and you increase the size of the screen to the tablet size (min), the different charts don't re draw and so they keep a really small size.

See the video above that describes what I'm talking about.

https://user-images.githubusercontent.com/4548045/195869892-b5b7300c-6fff-4d18-8ac7-724218c510c7.mov

PR #778, opened 2022-11-17
Fix bug with grid layout responsive behavior

juliepagano:
Before this fix, the contents of the grid would stay small and look strange after making the browser window small and then large again.

After the fix, the grid behaves responsively, as expected.

## Steps to reproduce the bug
- Open a dashboard.
- Make your browser window small to trigger the responsive behavior.
- Maker your browser window wider to go back to the larger responsive size. 
- See a dashboard with very scrunched up charts that looks odd and is hard to use.

This appears to be related to https://github.com/react-grid-layout/react-grid-layout/issues/1663.

### Before the fix

After following the steps to reproduce the bug:
image

### After the fix

After following the steps to reproduce the bug:
image

### DIFF
--- a/ui/dashboards/src/components/GridLayout/GridLayout.tsx
+++ b/ui/dashboards/src/components/GridLayout/GridLayout.tsx
@@ -25,7 +25,7 @@ export interface GridLayoutProps {
   panelGroupId: PanelGroupId;
 }
 
-const STORED_LAYOUT_BREAKPOINT = 'sm' as const;
+const SMALL_LAYOUT_BREAKPOINT = 'sm' as const;
 
 /**
  * Layout component that arranges children in a Grid based on the definition.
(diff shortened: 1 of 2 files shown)


ui/dashboards/src/components/GridLayout/GridLayout.tsx
```diff
@@ -37,6 +39,17 @@ export function GridLayout(props: GridLayoutProps) {
...
   const { isEditMode } = useEditMode();
 
+  const handleLayoutChange = (currentLayout: Layout[], allLayouts: Layouts) => {
+    // Using the value from `allLayouts` instead of `currentLayout` because of
+    // a bug in react-layout-grid where `currentLayout` does not adjust properly
+    // when going to a smaller breakpoint and then back to a larger breakpoint.
+    // https://github.com/react-grid-layout/react-grid-layout/issues/1663
+    const smallLayout = allLayouts[STORED_LAYOUT_BREAKPOINT];
+    if (smallLayout) {
+      updatePanelGroupLayouts(smallLayout);
```
juliepagano, line 49: I _think_ always ensuring we store the `sm` breakpoint should also avoid weirdness where someone tries to edit and save their dashboard on an extra small screen size.

Nexucis:
oh this is so cool ! Thank you so much @juliepagano !
This fix #642 :)

Nexucis (approved):
🥳

eunicorn (approved):
Yay thanks 🚀

ui/dashboards/src/components/GridLayout/GridLayout.tsx
cndonovan: Not a big deal, but you could maybe called `smallLayout` => `storedLayout`, or you could call `STORED_LAYOUT_BREAKPOINT` => `SMALL_LAYOUT_BREAKPOINT`

Just calling out because if `STORED_LAYOUT_BREAKPOINT` value did change for some reason, it would no longer be the `smallLayout`

PR #779, opened 2022-11-17
Even out the spacing between query options

cndonovan:
## Before
Screen Shot 2022-11-16 at 4 37 41 PM

## After
Screen Shot 2022-11-16 at 4 37 24 PM


sjcobb (approved):
👍👍 Much better

PR #782, opened 2022-11-17
bump @mui/material to v5.10.14

Nexucis:


sjcobb:
This is probably fine but we’ll need to be careful about this update, I need to confirm which version of MUI we’re on internally and make sure it doesn’t cause issues with some other dependencies (it’d be great if we had some way to do an alpha release that maybe just publishes to npm so we can quickly spot check this kind of thing)

Nexucis:
at some point it would make point to create a release candidate like `v0.17.0-rc0` indeed. That would mean it's a pre-released under testing :)

As a side note, it would be sad that we don't upgrade the version of a dependency on Perses because a company (here chronosphere but later another one could have the issue as well) cannot do the update :/

sjcobb:
> at some point it would make point to create a release candidate like `v0.17.0-rc0` indeed. That would mean it's a pre-released under testing :)
> 
That's exciting to hear, let's discuss this more at the sync meeting on Mon!
> As a side note, it would be sad that we don't upgrade the version of a dependency on Perses because a company (here chronosphere but later another one could have the issue as well) cannot do the update :/
> 
Yeah definitely, I agree--so it's not the change in the Perses app package.json that worries me, it's the updates in peerDependencies. But to be honest that's mostly because I don't 100% understand how peer dependencies work 😆, I've just seen weird things happen when they get changed. It does look like we're on v5.10.1 of @mui/material internally though so that's good news! Would reducing the peer dependency requirements slightly to `^5.9.0` make sense?

Nexucis:
> That's exciting to hear, let's discuss this more at the sync meeting on Mon!

Of course :). 

>  It does look like we're on v5.10.1 of @mui/material internally though so that's good news! Would reducing the peer dependency requirements slightly to ^5.9.0 make sense?

I'm not sure actually it makes sense. Perhaps `5.10.0` would be more coherent. 
But I'm not an expert ever with per-dependency. Maybe @LukeTillman, could you tell us what makes sense to put for the peer-dependencies here ?

sjcobb:
> I'm not sure actually it makes sense. Perhaps `5.10.0` would be more coherent.

Yeah that should be fine, `^5.10.0` sounds good to me!

PR #785, opened 2022-11-17
Update README.md

jordanbertasso:
It looks like this was fixed in https://github.com/perses/perses/pull/745


Nexucis (approved):
ah yes with v0.16.0 out, this message is not necessary anymore.
Thanks @jordanbertasso !

PR #786, opened 2022-11-17
Update spacing to match designs and do cleanup

cndonovan:
- Updated some spacing to match designs
- Also did some code consistency updates while I was looking at this: `direction={'row'}` => `direction="row"`, `sx={{ marginLeft: 'auto' }}` => `marginLeft="auto"`

Screen Shot 2022-11-17 at 3 54 12 PM

### DIFF
--- a/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
@@ -53,7 +53,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
           <Box sx={{ backgroundColor: (theme) => theme.palette.primary.light + '20' }}>
             <Box padding={2} display="flex">
               {dashboardTitle}
-              <Stack direction="row" spacing={1} sx={{ marginLeft: 'auto' }}>
+              <Stack direction="row" spacing={1} marginLeft="auto">
                 <Button variant="contained" onClick={onSave}>
                   Save
                 </Button>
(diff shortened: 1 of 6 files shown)


ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -86,10 +86,10 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
           </Box>
         </Stack>
       ) : (
-        <Stack spacing={2} padding={2}>
+        <Stack spacing={1} padding={2}>
```
cndonovan, line 89: Actual value change

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -86,10 +86,10 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
...
           <Box sx={{ display: 'flex', width: '100%' }}>
             {dashboardTitle}
-            <Stack direction="row" spacing={2} sx={{ marginLeft: 'auto' }}>
+            <Stack direction="row" spacing={1} marginLeft="auto">
```
cndonovan, line 92: Actual value change

ui/dashboards/src/components/Variables/VariableList.tsx
```diff
@@ -62,7 +62,7 @@ export function TemplateVariableList(props: TemplateVariableListProps) {
 
       <AppBar color={'inherit'} position={isSticky ? 'fixed' : 'static'} elevation={isSticky ? 4 : 0}>
         <Box display={'flex'} justifyContent="space-between" my={isSticky ? 2 : 0} ml={isSticky ? 2 : 0}>
-          <Stack direction={'row'} spacing={2}>
+          <Stack direction="row" spacing={1}>
```
cndonovan, line 65: Actual value change

Nexucis:
I'm using a bit the `Stack` component in this PR https://github.com/perses/perses/pull/774. Should I do apply the same policy regarding the sapce and the way to set the `direction` field ?

cndonovan:
> I'm using a bit the `Stack` component in this PR #774. Should I apply the same policy regarding the sapce and the way to set the `direction` field ?

Yeah! I think having spacing={1} between any buttons/inputs would be really good / would result in a consistent look that matches the designs. The way to set the direction field is just nicer / simpler IMO, not a big deal but just something I changed.

saminzadeh (approved):

PR #788, opened 2022-11-18
Reduce the number of retries when sending time series queries

cndonovan:


cndonovan:
> I was thinking about this and it might make sense for the application that embeds Perses to decide this. In order to use Perses, you need to provide a react-query client and that allows you to set the defaults, including `retries`
> 
> https://github.com/perses/perses/blob/83cfeab49fcc4c57bff2eeb3783652b5f2ee734c/ui/app/src/render-app.tsx#L32

Oh yeah that sounds great... I wonder if the QueryClient value would override it, that way we could have a default AND the application developer could override it. I'll test!

cndonovan:
@saminzadeh Ok, so here's kind of a rub: 

If we (or any developer) set `retry` in `QueryClient`, that applies to _all_ requests, not just time-series requests. So it would apply to requests for the plugin metadata, or requests for the variable data. 

Also, it looks like the behavior for `react-query` is that the `useQuery` `retry` will override the `QueryClient` `retry` (instead of my hope that `QueryClient` `retry` would override `useQuery` `retry`).

I have some code working where when we're about to run a time series request, we will check for `QueryClient` `retry` and default to that value. However, I'm not sure if we actually want to do it this way. I think if we want to expose this to developers, we would want to expose something like `timeSeriesRetry`.

saminzadeh:
> If we (or any developer) set `retry` in `QueryClient`, that applies to _all_ requests, not just time-series requests. So it would apply to requests for the plugin metadata, or requests for the variable data.

I would think that's the intent? I couldn't think of a reason of why something should retry unless it's explicitly being asked to. Basically, making it opt in instead of opt out. IMO Variable queries, plugin loading should all get this change as well. Do we have a valid reason why those should get retries? 

> Also, it looks like the behavior for react-query is that the useQuery retry will override the QueryClient retry (instead of my hope that QueryClient retry would override useQuery retry).

Ah yeah, figured that would be the case. Thanks for checking!

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -91,8 +93,7 @@ export const useTimeSeriesQuery = (definition: TimeSeriesQueryDefinition, option
       const ctx: TimeSeriesQueryContext = { ...context, suggestedStepMs: options?.suggestedStepMs };
       return plugin.getTimeSeriesData(definition.spec.plugin.spec, ctx);
     },
-    { enabled: pluginEnabled }
-  );
+  });
```
cndonovan, line 96: This makes `useTimeSeriesQuery` and `useTimeSeriesQueries` have the same format

saminzadeh (approved):

PR #789, opened 2022-11-18
Use Perses Drawer component for variable editor 

cndonovan:
Perses Drawer:
- has default right position
- has a default width
- requires a prop that tells it how to close when someone clicks outside the drawer 


saminzadeh (approved):
You're on fire with all the little UX clean up! 🔥

PR #790, opened 2022-11-18
form controls for new y_axis options

sjcobb:
Follow up to PR #767 with visual editing of new TimeSeriesChart y_axis options, `label`, `min`, `max`

### TODOS 
- adding more tests, validation / helperText, option to hide y_axis

### Demo video

https://user-images.githubusercontent.com/9369625/202747996-e671cb9c-2800-4f7d-b845-5415850be5c8.mp4

### DIFF
--- a/ui/components/src/AxisLabel.tsx
+++ b/ui/components/src/AxisLabel.tsx
@@ -1,0 +1,47 @@
+// Copyright 2022 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Box, Typography } from '@mui/material';
+
+interface AxisLabelProps {
+  name: string;
+  height: number;
+}
+
+export function AxisLabel({ name, height }: AxisLabelProps) {
+  return (
+    <Box
+      sx={{
+        display: 'inline-block',
+        maxWidth: height, // allows rotated text to truncate instead of causing overlap
+        position: 'absolute',
+        top: '45%',
+        transform: 'translateX(-50%) rotate(-90deg)',
+        transformOrigin: 'top',
+        textAlign: 'center',
+        zIndex: 1,
+      }}
+    >
+      <Typography
+        variant="body1"
+        sx={{
+          whiteSpace: 'nowrap',
+          overflow: 'hidden',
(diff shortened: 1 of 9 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -55,7 +55,6 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
   // convert Perses dashboard format to be ECharts compatible
   const yAxis = {
-    name: y_axis?.label ?? '',
```
sjcobb, line 58: Decided to change the axis label to a regular component instead of using the out of the box ECharts one, it gives us more control and doesn't have the limitations of being rendered to canvas (aka will be easier to test)

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -89,8 +103,5 @@ export function createInitialTimeSeriesChartOptions(): TimeSeriesChartOptions {
         },
       },
     ],
-    unit: {
```
sjcobb, line 92: `unit` moves inside `y_axis`, I removed it from initial options here since all the new properties are optional (most people can rely on defaults and less JSON for GitOps)

ui/components/src/AxisLabel.tsx
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+import { Box, Typography } from '@mui/material';
+
+interface AxisLabelProps {
+  name: string;
+  height: number;
+}
+
+export function AxisLabel({ name, height }: AxisLabelProps) {
```
cndonovan, line 21: Not sure, but if this is always rotated to the left, should we call it `YAxisLabel` (it would only apply to y-axis?)

ui/panels-plugin/src/plugins/time-series-chart/YAxisOptionsEditor.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onChange: (yAxis: YAxisOptions) => void;
+}
+
+export function YAxisOptionsEditor({ value, onChange }: YAxisOptionsEditorProps) {
+  const handleUnitChange: UnitSelectorProps['onChange'] = (newUnit) => {
+    onChange({
+      ...value,
+      unit: newUnit,
+    });
+  };
```
cndonovan, line 29: Not opposed to putting this inline like the other `onChange` handlers (but really it's fine however)

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
cndonovan: Oh yeah that sounds nice

cndonovan (approved):

ui/components/src/AxisLabel.tsx
sjcobb: Yeah good call, I just changed the component name!

PR #798, opened 2022-11-21
add option to hide y_axis and reset to defaults

sjcobb:
### Overview
This PR includes the following updates to TimeSeriesChart panel:
- add `y_axis.show` as supported time-series panel option
- in TimeSeriesChartPanel.tsx add support for ECharts show property mapping from dash definition
- adjust grid.left depending on if axis is hidden and if y_axis.label is populated
- Add reset button that removes all optional time-series panel settings (less JSON for GitOps)
- Reduce point_radius and line_width allowed values

### Screenshots

image

image

image


schemas/panels/time-series/time-series.cue
```diff
@@ -22,11 +22,12 @@ import (
 }
 
 #visual: {
-	line_width?:   number & >=0.5 & <=10
-	point_radius?: number & >=0 & <=20
+	line_width?:   number & >=0.5 & <=4
+	point_radius?: number & >=0 & <=5
```
juliepagano, line 26: Out of curiosity, any reasons for these specific numbers? I wonder if `point_radius` should be a little bigger, so you can have a more distinct point if you have the largest `line_width`.

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -44,12 +44,12 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
...
           <Slider
             data-testid={VISUAL_CONFIG.point_radius.testId}
-            defaultValue={DEFAULT_POINT_RADIUS}
+            value={value.point_radius ?? DEFAULT_POINT_RADIUS}
             valueLabelDisplay="auto"
             step={0.5}
             marks
             min={0}
-            max={20}
+            max={5}
```
juliepagano, line 52: Optional suggestion: It could be nice to pull this into a constant in the model like we do for some other values that are quasi-synced up with what's in cue. Maybe something like `MAX_POINT_RADIUS` or an object like `POINT_RADIUS = { min: 0, max: 5}` or an array with two numbers like `POINT_RADIUS = [0, 5]`.

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -59,12 +59,12 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
...
           <Slider
             data-testid={VISUAL_CONFIG.line_width.testId}
-            defaultValue={DEFAULT_LINE_WIDTH}
+            value={value.line_width ?? DEFAULT_LINE_WIDTH}
             valueLabelDisplay="auto"
             step={0.5}
             marks
             min={0.5}
-            max={10}
+            max={4}
```
juliepagano, line 67: Similar optional suggestion as above about pulling these out into constants. This PR is already doing a lot, so totally ok to do these in a follow-up or the future if you don't want to mess with it right now.

juliepagano (approved):
Generally looks good to me! One small question and a few suggestions. Also don't forget to update the CHANGELOG (I always forget about this until I'm about to merge!), especially since you've got a few things happening in here.

eunicorn (approved):

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: Yeah good call, I just [added this in similar way](https://github.com/perses/perses/pull/798/commits/3e22f061eff55e029a8799521decc4fd0e779995) to your second example

schemas/panels/time-series/time-series.cue
sjcobb: I got some feedback from @foxbat07 about reducing them since I had the max set way too high originally, but I agree when playing around with it, you almost can't even see the point when line_width is larger. I just increased the point_radius max a little bit to account for this.

PR #803, opened 2022-11-21
Fix button wrapping when too many vars

juliepagano:

## Notes for reviewers
There is probably some more work we need to do with design on this around the behavior of variables when we have a lot of them and how they should wrap. I'm not digging into that this week because Mohit is on vacation, so this is a small incremental improvement.

## Screenshots

<table>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>


eunicorn (approved):

eunicorn:
I know @foxbat07 wants the template variables to be its own div that takes up the whole app width, but i think we will have to break apart Template Variable List component to do that 🤔

saminzadeh (approved):

juliepagano:
> I know @foxbat07 wants the template variables to be its own div that takes up the whole app width, but i think we will have to break apart Template Variable List component to do that 🤔

I was wondering if that was the case, but didn't dig into it since he was out this week. That sounds like it'll be a good follow-up to tackle, and we can also dig into the wrapping considerations then. This is a baby step to make it look less broken in the mean time.

PR #809, opened 2022-11-22
Remove empty chart

juliepagano:
This chart was created early in the project to help with development, but is no longer useful now that there are many other chart types.

Removing it since it is odd for users to see it as an option when adding a new panel, and it is not adding value for users in a dashboard.


LukeTillman (approved):
🎉

PR #811, opened 2022-11-22
Fix pressing back button removes query param individually

eunicorn:
Pressing browser back button removes each query param individually instead actually navigating to the previous page

**Before**
before-back-button

**After**
after-back-button

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - [ENHANCEMENT] Dashboard view can be configured to be readonly #731
 - [ENHANCEMENT] reduce line_width and point_radius max allowed values #798
 - [BUGFIX] Fix button wrapping when too many vars #803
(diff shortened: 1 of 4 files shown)


saminzadeh:
I wonder if there is a way to fix this experience just on the first render. Ideally, after you change a time range or variable, you can press "Back" and it will go back to previous state you had selected. At least that's the expected behavior in Grafana, but not sure how we feel about it.

Currently, this feels worse since the URL is updated each time a component (time range or variable) is rendered. So if we have 3 variables, you have to press the back button 4 times (3 for each variable, and 1 for the time range) to go back to the previous page. A few ideas to fix this:
- Batch the URL updates at the same time 
- Don't update the URL if the default values are selected

Feel free to ping me offline if this response isn't clear

saminzadeh (approved):

PR #813, opened 2022-11-23
set abbreviate in default Decimal unit

sjcobb:
Change default Decimal to always abbreviate large number (15K, 15M, etc.) in y_axis labels. Show full unabbreviated value in LineChart's Tooltip.

May add  more control for tooltip display in the future once we add tooltip-specific panel options.

### Screenshot
Revert changes to first commit: 
image

y_axis example:
image

~~Tooltip unabbreviated ex~~:


ui/components/src/Tooltip/utils.ts
```diff
@@ -58,3 +59,17 @@ export function assembleTransform(
...
+/**
+ * Default to abbreviated value in y_axis but full value in tooltip
+ */
+export function getTooltipUnit(unit?: UnitOptions) {
+  if (!unit) {
+    return undefined;
+  }
+  const tooltipUnit = { ...unit };
+  if (tooltipUnit.kind === 'Decimal' && tooltipUnit.abbreviate) {
+    tooltipUnit.abbreviate = false;
```
sjcobb, line 72: I thought about using immer, but decided not to since it isn't a required dependency in @perses-dev/components (at least not yet)

saminzadeh (approved):
My personal preference is to keep the Y-axis and tooltip consistent, so abbreviate both or none. 

When in doubt, it's worth looking at Grafana's issue history [[1]](https://github.com/grafana/grafana/issues?q=abbreviate) [[2]](https://github.com/grafana/grafana/issues?q=units++tooltip) to see if anyone has discussed/requested that feature before.

eunicorn (approved):

sjcobb:
> My personal preference is to keep the Y-axis and tooltip consistent, so abbreviate both or none.

Yeah I went back and forth on this, ultimately went with Karan's suggestion, but discussed with @foxbat07 yesterday and he agreed with you about consistency, going to revert the tooltip changes and show `abbreviate: true` in both places!

PR #815, opened 2022-11-23
Add ability to set default plugin kinds

juliepagano:
The optional `defaultPluginKinds` prop on `PluginRegistry` enables developers to configure the default selected plugin kind by type. If this prop is not set or does not have a value for a given type, no plugin kind is selected by default.

## Notes for reviewers

- I'm still getting up-to-speed on how all of the plugin plumbing works, so I'm not sure if this is the best approach to implementing this. I'm open to alternative suggestions!

## Screenshots

image


ui/dashboards/src/context/DashboardProvider/panel-editor-slice.ts
```diff
@@ -190,11 +195,10 @@ export function createPanelEditorSlice(): StateCreator<
           name: '',
           description: '',
           groupId: panelGroupId,
-          // TODO: If we knew what plugins were available (and how to create the initial spec), we might be able to
-          // set a smarter default here?
```
juliepagano, line 194: Instead of trying to duplicate the plugin loading logic in `panel-editor-slice` to get the default spec, I pass `defaultPanelKind` down to where that logic is, so it can be used as the initializing value when available.

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -66,7 +67,7 @@ export function usePluginEditor(props: UsePluginEditorProps) {
 
   // When kind changes and we haven't loaded that plugin before, we will need to enter a "pending" state so that we
   // can generate proper initial spec values that match the new plugin kind
-  const [pendingKind, setPendingKind] = useState('');
+  const [pendingKind, setPendingKind] = useState(props.defaultPanelKind || '');
```
juliepagano, line 70: Implementing this here lets us leverage the loading logic that's already set up instead of refactoring it a ton or duplicating it.

cndonovan: Whoa amazing

cndonovan (approved):

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
cndonovan: Love this idea, I was thinking about trying to load a plugin in `panel-editor-slice`, this feels better

LukeTillman (commented):
This is awesome! ❤️ 

Funny, I was thinking about this problem right before the break and this is very similar to what I was thinking about doing. What do you think about maybe baking in "default plugin kinds" to the `PluginRegistry` itself (probably as props like you've done here)? Since that's the component where the user provides the available plugins, feels like maybe that's also where they could set defaults. Child components could then access the defaults on the registry's API/context.

juliepagano:
> This is awesome! ❤️
> 
> Funny, I was thinking about this problem right before the break and this is very similar to what I was thinking about doing. What do you think about maybe baking in "default plugin kinds" to the `PluginRegistry` itself (probably as props like you've done here)? Since that's the component where the user provides the available plugins, feels like maybe that's also where they could set defaults. Child components could then access the defaults on the registry's API/context.

Good call! I'll take a stab at that this afternoon.

juliepagano:
Ok, I think I've got this working with the plugin registry instead. Let me know what you think, @LukeTillman!

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -48,7 +48,7 @@ function ViewDashboard() {
       }}
     >
       <ErrorBoundary FallbackComponent={ErrorAlert}>
-        <PluginRegistry pluginLoader={bundledPluginLoader}>
+        <PluginRegistry pluginLoader={bundledPluginLoader} defaultPanelKind="TimeSeriesChart">
```
LukeTillman, line 51: Nice! I think in the future, we may want to expand this to be more generic (e.g. `defaultPluginKinds` or something) since I think other places where you need to select a plugin have a similar problem. (Other places would be variable kind selection when editing a variable and query kind selection when editing a query in a panel.)

LukeTillman (approved):
Looks great, thanks for knocking this out!

ui/app/src/views/ViewDashboard.tsx
juliepagano: Good call. Let me take a quick timeboxed look at if that's an easy refactor now since that could avoid a breaking change in the future.

juliepagano: That was an easy refactor! Tagging you to give this one more pass before merging since it was a significant code change.

LukeTillman (approved):
Refactor looks good, nice work!

PR #821, opened 2022-11-28
show axisPointer line in time series panel

sjcobb:
Show [axisPointer](https://echarts.apache.org/en/option.html#tooltip.axisPointer.type) dashed line on hover when not in optimized mode. Adding the OPTIMIZED_MODE_SERIES_LIMIT condition is important or else moving your cursor when there are tons of time series can get sluggish due to how many symbols need to render.

### Screenshot

image

image


ui/components/src/LineChart/LineChart.tsx
```diff
@@ -177,11 +178,11 @@ export function LineChart({
       yAxis: getYAxes(yAxis, unit),
       animation: false,
       tooltip: {
-        show: showPointsOnHover,
+        show: showHoverEmphasis,
         trigger: 'axis',
         showContent: false,
         axisPointer: {
-          type: 'none',
+          type: showHoverEmphasis ? 'line' : 'none',
```
juliepagano, line 185: Optional suggestion: I _think_ this bit of code might be a little easier to read if the variable was named after the status that informs these conditionals (e.g. `isOptimizedMode` or `isStandardMode` if you want the reverse) instead of a description of the outcome of that state (`showHoverEmphasis`).  For example:

```suggestion
          type: isOptimizedMode ? 'none' : 'line',
```

makes it a little clearer what's happening and why.

ui/components/src/model/graph.ts
```diff
@@ -14,7 +14,7 @@
 import { MouseEventHandler } from 'react';
 import { LineSeriesOption } from 'echarts/charts';
 
-export const PROGRESSIVE_MODE_SERIES_LIMIT = 500;
+export const OPTIMIZED_MODE_SERIES_LIMIT = 500;
```
juliepagano, line 17: Do we have a note elsewhere in the code explaining what this mode is, why it's needed, and how we came up with 500? If not, it might be worth adding a very short summary in a comment here for future us.

juliepagano (approved):
A few optional questions and suggestions. Otherwise, looks good to me. 🎉

ui/components/src/LineChart/LineChart.tsx
sjcobb: Yeah I definitely like your suggested name better! I actually ended up removing these conditions from LineChart upon further testing ([see other comment](https://github.com/perses/perses/pull/821#discussion_r1036168943))

PR #825, opened 2022-11-29
Introduce TimeZoneProvider for specifying a timezone other than browser/local

sjcrank:
**Motivation**
Applications embedding Perses dashboards or components may need to format date/time using a timezone other than the browser/local setting. For example, a host application may allow end users to toggle between UTC and browser time.

**Proposal**
Introduce a `TimeZoneProvider` with similar semantics and usage to the `ChartsThemeProvider`, through which Perses clients may provide a timezone selection. The timezone may be "local" or "browser" to specify usage of the default browser timezone, "utc" for UTC timezone, or any browser-supported IANA timezone name such as "America/New_York". The list of supported IANA timezone names may be found using [Intl.supportedValuesOf('timeZone')](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf).

**Implementation Notes**
- introduced dependency on [date-fns-tz](https://github.com/marnusw/date-fns-tz) which adds timezone support to the existing `date-fns` dependency
- introduced utility function `formatWithTimeZone` which formats a Date using a provided format string and timezone
- introduced utility function `dateFormatOptionsWithTimeZone` which adds timeZone to the provided date format options
- updated usages of `date-fns format` with `formatWithTimeZone`, passing in the timeZone from TimeZoneProvider
- updated usages of `Intl.DateTimeFormat`, providing the timeZone from TimeZoneProvider to the format options using `dateFormatOptionsWithTimeZone`
- `TimeZoneProvider.useTimeZone()` falls back to browser/local timezone in case the context provider is not in the React tree (existing Perses clients will NOT need to add TimeZoneProvider unless they want to specify the timezone)

**Screenshots**

local

utc


ui/components/src/utils/format.ts
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { formatInTimeZone } from 'date-fns-tz';
+import { format } from 'date-fns';
+
+export function dateFormatOptionsWithTimeZone(dateFormatOptions: Intl.DateTimeFormatOptions, timeZone?: string) {
```
saminzadeh, line 17: Might be worth adding a few unit tests to these utils

ui/components/src/DateTimeRangePicker/AbsoluteTimePicker.tsx
```diff
@@ -29,6 +30,7 @@ interface AbsoluteTimeFormProps {
 export const AbsoluteTimePicker = ({ initialTimeRange, onChange }: AbsoluteTimeFormProps) => {
   const [timeRange, setTimeRange] = useState<AbsoluteTimeRange>(initialTimeRange);
   const [showStartCalendar, setShowStartCalendar] = useState<boolean>(true);
+  const timeZone = useTimeZone();
```
saminzadeh, line 33: This makes sense, however do you think its worth also providing the `formatWithTimeZone()` as injected utility as well?

```tsx
  const { timeZone, formatWithTimeZone } = useTimeZone();
  
  // the timezone (last parameter) is optional since it will take the injected context
  formatWithTimeZone(timeRange.start, DATE_TIME_FORMAT)
 ```

**Review by saminzadeh — COMMENTED** [review]
This looks great Sam, thanks for contributing! I left a few comments

File: ui/app/src/context/UtcTimeZone.tsx [hunk 3]
```diff
@@ -0,0 +1,41 @@
...
+export const UtcTimeZoneContext = createContext<UtcTimeZoneContext | undefined>(undefined);
+
+export function UtcTimeZoneProvider(props: { children: React.ReactNode }) {
+  const [isUtcTimeZone, setUtcTimeZone] = useState(false);
+  return (
+    <TimeZoneProvider timeZone={isUtcTimeZone ? 'utc' : 'local'}>
+      <UtcTimeZoneContext.Provider value={{ isUtcTimeZone, setUtcTimeZone }}>
+        {props.children}
+      </UtcTimeZoneContext.Provider>
+    </TimeZoneProvider>
```
LukeTillman, line 31: This threw me for a loop a little bit at first when reading it (with there being two providers). Did you consider doing this as one provider? Feels like maybe you didn't want to expose set functionality to all the consumers?

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -213,7 +215,14 @@ export function LineChart({
...
     >
       {showTooltip === true && (
-        <Tooltip chartRef={chartRef} chartData={data} wrapLabels={true} pinTooltip={pinTooltip} unit={unit}></Tooltip>
+        <Tooltip
+          chartRef={chartRef}
+          chartData={data}
+          wrapLabels={true}
+          pinTooltip={pinTooltip}
+          unit={unit}
+          timeZone={timeZone}
```
LukeTillman, line 224: Suggestion (feel free to ignore): looks like maybe you could avoid passing this value as a prop through a couple levels of components if `TooltipContent` also called `useTimeZone` instead.

LukeTillman (approved):
Overall, this looks great. Left a couple comments/questions inline. Thanks and congrats on your first Perses contribution! 🙌

ui/components/src/utils/format.ts
sjcrank: makes sense, will do

ui/components/src/DateTimeRangePicker/AbsoluteTimePicker.tsx
sjcrank: Sure, will explore this approach

ui/app/src/context/UtcTimeZone.tsx
sjcrank: You will notice this closely mirrors the app/src/context/DarkMode implementation.

The demonstration App can be implemented in various ways, supporting the ability to show and modify the timezone in the app header, as long as that timezone is made available via TimeZoneProvider to the embedded Perses dashboards. I thought it best to keep the Perses TimeZoneProvider minimal (read-only timezone state), and to implement the App support for view/edit timezone similar to its theme switcher for consistency.

ui/components/src/LineChart/LineChart.tsx
sjcrank: I will explore this option

juliepagano:
The double toggles without labels in the UI feels a little confusing to me. It's also not clear what will happen when you click the time toggle until after you do it and look at the tooltip. Maybe it would make sense to add a small text label next to each (e.g. "time zone" and "theme" respectively)?

image

juliepagano (commented):
One bit of UI feedback I left as a direct comment and one tiny nit in the code proper. 

Could you also add an updated entry to the [`CHANGELOG` for this in the "unreleased" section](https://github.com/perses/perses/blob/main/CHANGELOG.md#unreleased) as part of the PR? We're hoping to eventually make this a little more automated with some nice documentation, but we're not there quite yet.

sjcrank:
> The double toggles without labels in the UI feels a little confusing to me. It's also not clear what will happen when you click the time toggle until after you do it and look at the tooltip. Maybe it would make sense to add a small text label next to each (e.g. "time zone" and "theme" respectively)?
> 
> image

Done:
Screen Shot 2022-12-05 at 9 06 03 AM

sjcrank:
> One bit of UI feedback I left as a direct comment and one tiny nit in the code proper.
> 
> Could you also add an updated entry to the [`CHANGELOG` for this in the "unreleased" section](https://github.com/perses/perses/blob/main/CHANGELOG.md#unreleased) as part of the PR? We're hoping to eventually make this a little more automated with some nice documentation, but we're not there quite yet.

Added the CHANGELOG entry

Nexucis:
Maybe a drop down instead of a toogle could be better. The value of the dropdown would be UTC or browser timezone. Like that you can see without passing the cursor what is the current timezone used.

sjcrank:
> Maybe a drop down instead of a toogle could be better. The value of the dropdown would be UTC or browser timezone. Like that you can see without passing the cursor what is the current timezone used.

Prefer to keep the toggle for expediency. This is a simple demonstration app, and the toggle follows the pattern already in place for toggling the dark/light theme. If we want to optimize the UX, it would be a larger process of bringing in UX design, adding the appropriate theme updates to MUI select component, etc. This seems like overkill when we simply want a demonstration of how to change time zones.

Nexucis:
Actually it's not a demonstration app. This is the actual Perses UI, even if it's not that bright for the moment, we don't build an app just for testing purpose or demonstration.
This app is going to be used in production :).

sjcrank:
> Actually it's not a demonstration app. This is the actual Perses UI, even if it's not that bright for the moment, we don't build an app just for testing purpose or demonstration. This app is going to be used in production :).

OK got it, makes sense. Who is responsible for UX design / review? Perhaps we can ask for input on whether a UTC/local timezone selector is desirable, and what form it should take?

sjcrank:
> > Actually it's not a demonstration app. This is the actual Perses UI, even if it's not that bright for the moment, we don't build an app just for testing purpose or demonstration. This app is going to be used in production :).

I'm wondering if the best approach is to simply remove the timezone toggle from the App. This will leave the App functionality unchanged, but this PR exposes the provider that can be used by any app that wants to embed Perses dashboards and specify time zone. Thoughts?

sjcobb:
> best approach is to simply remove the timezone toggle from the App. This will leave the App functionality unchanged

Yeah this makes sense to me, we can figure out the selector UX in a separate PR later!

sjcrank:
@Nexucis @juliepagano @sjcobb please review - I have removed the timezone toggle from the App. This PR is now limited to adding support for TimeZoneProvider which will allow apps embedding Perses dashboards to override the default browser/local timezone.

We can add a PR later if it is determined that a timezone selector should be added to the App.

juliepagano (approved):
LGTM! Thanks for the contribution. 🚀

sjcobb (approved):

PR #827, opened 2022-11-30
[FEATURE] Persist dashboard modification when clicking on save button

### LINKED ISSUE #807: Persist modification of the dashboard
When clicking on the `save` button available after editing a dashboard, the dashboard is not persisted in the database. So if you reload the page you lost every changes you did.

Nexucis:
This PR allowed the UI to send a request to the backend to persist the modified dashboard.

Success or error message are displayed using the `Snackbar`

This PR fixes #807 

### DIFF
--- a/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
@@ -45,7 +45,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
   } = props;
 
   const { isEditMode, setEditMode } = useEditMode();
-  const [isSaveDashboard, setSaveDashboard] = useState<boolean>(false);
+  const [isSavingDashboard, setSavingDashboard] = useState<boolean>(false);
   const dashboard = useDashboard();
   const { openAddPanelGroup, openAddPanel } = useDashboardActions();
   const isLaptopSize = useMediaQuery(useTheme().breakpoints.up('sm'));
@@ -56,15 +56,15 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
   );
 
   const onSaveButtonClick = () => {
-    if (onSave != undefined) {
-      setSaveDashboard(true);
+    if (onSave !== undefined) {
+      setSavingDashboard(true);
       onSave(dashboard.dashboard)
         .then(() => {
-          setSaveDashboard(false);
+          setSavingDashboard(false);
           setEditMode(false);
         })
         .catch(() => {
-          setSaveDashboard(false);
+          setSavingDashboard(false);
         });
     } else {
       setEditMode(false);
@@ -83,7 +83,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
                   Dashboard managed via code only. Download JSON and commit changes to save.
                 </Alert>
               )}
-              <Button variant="contained" onClick={onSaveButtonClick} disabled={isReadonly || isSaveDashboard}>
+              <Button variant="contained" onClick={onSaveButtonClick} disabled={isReadonly || isSavingDashboard}>
                 Save
               </Button>
               <Button variant="outlined" onClick={onCancelButtonClick}>
--- a/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
@@ -16,35 +16,36 @@ import PencilIcon from 'mdi-material-ui/PencilOutline';
 import AddPanelGroupIcon from 'mdi-material-ui/PlusBoxOutline';
 import AddPanelIcon from 'mdi-material-ui/ChartBoxPlusOutline';
 import { ErrorBoundary, ErrorAlert } from '@perses-dev/components';
-import { UseMutationResult } from '@tanstack/react-query';
 import { DashboardResource } from '@perses-dev/core';
 import { useDashboard, useDashboardActions, useEditMode } from '../../context';
 import { TemplateVariableList } from '../Variables';
 import { TimeRangeControls } from '../TimeRangeControls';
 import { DownloadButton } from '../DownloadButton';
+import { useState } from 'react';
 
 export interface DashboardToolbarProps {
   dashboardName: string;
   dashboardTitleComponent?: JSX.Element;
-  dashboardMutation?: UseMutationResult<DashboardResource, Error, DashboardResource>;
   initialVariableIsSticky?: boolean;
   isReadonly: boolean;
   onEditButtonClick: () => void;
   onCancelButtonClick: () => void;
+  onSave?: (entity: DashboardResource) => Promise<DashboardResource>;
 }
 
 export const DashboardToolbar = (props: DashboardToolbarProps) => {
   const {
     dashboardName,
     dashboardTitleComponent,
-    dashboardMutation,
     initialVariableIsSticky,
     isReadonly,
     onEditButtonClick,
     onCancelButtonClick,
+    onSave,
   } = props;
 
   const { isEditMode, setEditMode } = useEditMode();
+  const [isSaveDashboard, setSaveDashboard] = useState<boolean>(false);
   const dashboard = useDashboard();
   const { openAddPanelGroup, openAddPanel } = useDashboardActions();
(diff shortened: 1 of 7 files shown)


ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -16,14 +16,17 @@ import PencilIcon from 'mdi-material-ui/PencilOutline';
...
+import { DashboardResource } from '@perses-dev/core';
+import { useDashboard, useDashboardActions, useEditMode } from '../../context';
 import { TemplateVariableList } from '../Variables';
 import { TimeRangeControls } from '../TimeRangeControls';
 import { DownloadButton } from '../DownloadButton';
 
 export interface DashboardToolbarProps {
   dashboardName: string;
   dashboardTitleComponent?: JSX.Element;
+  dashboardMutation?: UseMutationResult<DashboardResource, Error, DashboardResource>;
```
eunicorn, line 29: Instead of adding a dashboardMutation prop here, I think it's better if we add an onSave prop that returns a promise so this component doesn't need to know anything about react-query and we can also handle loading/error/success state in here as well.

For example:

```
const { dashboard } = useDashboard();
const onSaveButtonClick = async () => {
    if (props.onSave) {
      try {
        await props.onSave(dashboard);
      } catch (err) {
        exceptionSnackbar(
          err,
          'There was an error while trying to save the dashboard. Please try again.'
        );
      }
    }
  };
```

saminzadeh: +1 `onSaveClick` to keep it generic. @eunicorn As far as the promise goes, are you thinking that when the Promise is in pending state, the button would be disabled and show a loading icon?

Nexucis: while I agree with this statement (dashboard shouldn't depend on a third party library like react-query), I just would like to raise the point `@perses-dev/dashboard` has a direct dependency to `@tanstack/react-query`.

So from my point of view: 
1. Yes it makes sense to avoid having a property that is using the type from `@tanstack/react-query`
2. Point 1 really makes sense if the package `@perses-dev/dashboard` doesn't depend at all of `@tanstack/react-query`

saminzadeh: I suppose it's less about dependencies but more about interfaces. A package can transparently depend on another package but its interface is still decoupled (not saying we are prefect at this everywhere). I think this is an area where it can be decoupled.

Nexucis: > exceptionSnackbar(
          err,
          'There was an error while trying to save the dashboard. Please try again.'
        );

Currently the snackbar is only available in the app.
I'm not sure if it makes sense to move it to the package `@perses-dev/component`. That would makes my life easier for sure, but then that would mean to be able to used the dashboard you need to use the `SnackbarProvider`. Maybe we don't want that ?

And if we don't want that, do you have an idea on how I could manage the error or the success message ?
I really like the snackbar because it's perfect to display a volatile message

saminzadeh: It would look something like this:

```tsx
<DashboardToolbar onSave={async (dashboardJSON) => {
    showLoadingSnackbar();
    try {
        const result = await saveDashboard() // via backend
        showSuccessSnackbar();
    } catch(e) {
        showErrorSnackbar(e);
    }
}} />
```

Nexucis:
thanks for your help @eunicorn @saminzadeh . This PR is ready to review now :)

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -65,7 +83,7 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
                   Dashboard managed via code only. Download JSON and commit changes to save.
                 </Alert>
               )}
-              <Button variant="contained" onClick={onSave} disabled={isReadonly}>
+              <Button variant="contained" onClick={onSaveButtonClick} disabled={isReadonly || isSaveDashboard}>
```
eunicorn, line 86: should we hide `Save` button if `onSave` is undefined?

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -38,9 +41,12 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
     isReadonly,
     onEditButtonClick,
     onCancelButtonClick,
+    onSave,
   } = props;
 
   const { isEditMode, setEditMode } = useEditMode();
+  const [isSaveDashboard, setSaveDashboard] = useState<boolean>(false);
```
eunicorn, line 48: i think calling this `isSavingDashboard` might be more clear. I was confused if this meant is dashboard saved at first

eunicorn (approved):
this is awesome!!

saminzadeh (approved):

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -49,8 +55,20 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
     <Typography variant="h2">{dashboardName}</Typography>
   );
 
-  const onSave = () => {
-    setEditMode(false);
+  const onSaveButtonClick = () => {
+    if (onSave != undefined) {
```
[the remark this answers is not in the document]
Nexucis: yeah I know but I followed the UI best practices available here: https://github.com/perses/perses/blob/main/ui/ui-guidelines.md#frontend-best-practices :)

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
Nexucis: > should we hide the save button if `onSave` is undefined?

mmm that's an interesting question. But does it make sense then that you can still edit the dashboard.
I think for the moment we should keep the `save` button and at some point we should revisit the edit/save flow to be sure it's consistent depending of the different context (readonly UI, onSave not defined ...etc)

Nexucis: > I wish MUI button had a `loading` prop :( and not this [experimental one ](https://mui.com/material-ui/api/loading-button/)

yeah that would be great, it would simplify so much work. We could still depend on `@mui/lab` if it's a wish we all share.

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -49,8 +55,20 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
...
+      setSaveDashboard(true);
+      onSave(dashboard.dashboard)
```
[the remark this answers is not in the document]
Nexucis: mmm actually in this context, I feel it's simpler to use `then().catch()` instead.

sjcobb (approved):
This is great, I think it counts as a FEATURE instead of ENHANCEMENT 😄 

I played around with it some and was really cool to see it catch some invalid properties I wasn't aware of in my dashboards locally 🙌

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -49,8 +55,20 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
     <Typography variant="h2">{dashboardName}</Typography>
   );
 
-  const onSave = () => {
-    setEditMode(false);
+  const onSaveButtonClick = () => {
```
sjcobb, line 58: I was just playing around some more and I think I found a bug. So my dashboard locally had JSON that was formatted in VSCode, when I made changes in the UI and then saved the dashboard, the raw JSON was saved, but all remaining lines after the first are still there, see screenshot:
image
 
 Here is the data too after I clicked save: 
[DemoDashboardAlt.json.zip](https://github.com/perses/perses/files/10187687/DemoDashboardAlt.json.zip)

Nexucis: this is fixed with #856

Issue #819, opened 2022-11-28
Browser freeze when editing panel `Legend Example` in dashboard `Benchmark`


Nexucis:
## Steps to reproduce 
1. Go on the `Benchmark` dashboard
2. Click on the `edit` button to activate the edit mode
3. Click on the `edit` button of the panel `Legend Example`

## Result

After a few seconds, the page is frozen and you can just kill the browser tab. I didn't notice an error in the console when it's happening.

cndonovan:
@Nexucis Would you like to confirm that this has been fixed and close this issue if it is?

PR #828, opened 2022-11-30
Fix panel editor crash by reducing number of expensive loops

cndonovan:
This gets us to a better place (no crashes), but there are probably ways to improve.

## Issue
We we have been seeing that the browser can freeze/crash when we open the panel editor AND the panel has multiple queries: https://github.com/perses/perses/issues/819

## What's going on?
When we render a time series chart, we load data from the backend, then we use this data to compute `graphData`, and then pass `graphData` into `LineChart` for rendering. But computing `graphData` involves looping over every data point -- so this is the part where we can cause the browser to freeze.

The computation of `graphData` is memoized, with the main dependency being the results from the backend. 

When we open the editor, it looks like we trigger fetching data, which changes the `fetchStatus` of each query result. When there are multiple queries, this can lead to `graphData` being recomputed each time any of the query results changes from `fetchStatus: fetching` to `fetchStatus: idle`. On top of this, it also looks like `stepMs` can go through changes when we're in this fetching stage, which also causes `graphData` to be recomputed too.

## Fix
For now, I was able to stop browser crashes by simply not running the loops if we are either loading or fetching data.

The downside is that we'll render the loading indicator when we're fetching new data, instead of rendering the old data until we're ready to render the new data. It would be nicer to render the old data until we're ready, but I figure this is a good step for now (better than a browser crash).

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -11,8 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 import { merge } from 'lodash-es';
+import { useDeepMemo } from '@perses-dev/core';
 import { PanelProps, useTimeSeriesQueries, useTimeRange } from '@perses-dev/plugin-system';
 import type { GridComponentOption } from 'echarts';
 import { Box, Skeleton } from '@mui/material';
@@ -70,6 +71,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
   const suggestedStepMs = useSuggestedStepMs(contentDimensions?.width);
   const queryResults = useTimeSeriesQueries(queries, { suggestedStepMs });
+  const fetching = queryResults.some((result) => result.isFetching);
   const loading = queryResults.some((result) => result.isLoading);
   const hasData = queryResults.some((result) => result.data && Array.from(result.data.series).length > 0);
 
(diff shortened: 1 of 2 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -120,8 +130,8 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     const xAxisData = [...getXValues(timeScale)];
 
     for (const result of queryResults) {
-      // Skip queries that are still loading and don't have data
-      if (result.isLoading || result.data === undefined) continue;
+      // Skip queries that are still loading or don't have data
+      if (result.isLoading || result.isFetching || result.data === undefined) continue;
```
cndonovan, line 134: We shouldn't actually hit this, but I left it in case we want it in the future

sjcobb (approved):
🎉 awesome we need to get this published ASAP

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -103,7 +105,15 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   };
 
   // Populate series data based on query results
-  const { graphData } = useMemo(() => {
+  const { graphData } = useDeepMemo(() => {
```
LukeTillman, line 108: Just curious: why deep memo now instead of regular memo? Asking since that calculation is more expensive. (Although as you pointed out, definitely better than a browser crash 😂 )

LukeTillman (approved):
🚀

PR #829, opened 2022-12-01
[ENHANCEMENT] Make it possible to adjust the height of the time range controls

cndonovan:
### Problem

In our app, the time range controls (the time range picker and the refresh button) appear next to some buttons, but these controls don't have the same height as the buttons. It looks broken.

Screen Shot 2022-12-01 at 8 24 15 AM

### Updates
Looked into the custom variants idea (julie), the make components grow to fill the container idea (luke), and the pass along specific styles to components idea (luke).

It would be cool to be able to do the fill-the-container idea, but `MUI Select` gave me a hard time on this. I wasn't able to find a way to apply styles to the underlying elements that would make them grow to fill the container. 

So, in the end, changed the external prop to `height` and use the `height` prop to (1) set the `height` of the refresh button and (2) set the vertical padding on the element that represents the time range selector. 

~### Solution~

~In our app, buttons are meant to be one of three heights: small = 28px, medium = 32px (default), and large = 40px.~

~I added a way to adjust the height of the time range controls that matches these values.~

~### Notes~

~I'm not actually going to use the `size` prop in Perses! (Because it looks fine in Perses.) I'll be waiting for integration with cloud to use the `size` prop.~

~### Pitfalls~

~I took a simple approach and didn't make a `type` / `enum` for the `size` prop. I wasn't sure where this `type` / `enum` would live so that it could be accessed by `perses/dashboards` and `perses/components` (though, as I type this, I think the answer is `perses/core`). I also thought I might be getting a little ahead of myself if I enshrined `size` as an idea in wider parts of the codebase.~

~The other pitfall with this approach is that it's hard-coded to 28px/32px/40px, instead of reading values from `theme`. I looked at `theme` and didn't see any fields that seemed appropriate -- MUI theme doesn't seem to have a concept of global `size`.~

Open to ideas on any of this!

### DIFF
--- a/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
+++ b/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
@@ -19,6 +19,12 @@ import { formatAbsoluteRange } from './utils';
 
 const DATE_TIME_FORMAT = 'yyyy-MM-dd HH:mm:ss';
 
+const SIZE_TO_PADDING_Y = {
+  small: '4px',
+  medium: '6px',
+  large: '10px',
+};
+
 export interface TimeOption {
   value: RelativeTimeRange;
   display: string;
@@ -29,10 +35,11 @@ interface TimeRangeSelectorProps {
   timeOptions: TimeOption[];
   onSelectChange: (event: SelectChangeEvent<string>) => void;
   onCustomClick: () => void;
+  size?: 'small' | 'medium' | 'large';
 }
 
 export function TimeRangeSelector(props: TimeRangeSelectorProps) {
-  const { value, timeOptions, onSelectChange, onCustomClick } = props;
+  const { value, timeOptions, onSelectChange, onCustomClick, size } = props;
   const { timeZone } = useTimeZone();
   const formattedValue = !isRelativeTimeRange(value)
     ? formatAbsoluteRange(value, DATE_TIME_FORMAT, timeZone)
@@ -46,6 +53,7 @@ export function TimeRangeSelector(props: TimeRangeSelectorProps) {
         '.MuiSelect-icon': {
           marginTop: '1px',
         },
+        '.MuiSelect-select': size ? { paddingY: SIZE_TO_PADDING_Y[size] } : {},
       }}
     >
       {timeOptions.map((item, idx) => (
--- a/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
+++ b/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
@@ -19,16 +19,6 @@ import { formatAbsoluteRange } from './utils';
 
 const DATE_TIME_FORMAT = 'yyyy-MM-dd HH:mm:ss';
 
-function heightToPaddingY(height: string): string {
-  const heightNum = parseFloat(height);
-
-  if (heightNum >= 20) {
-    return `${(heightNum - 20) / 2}px`;
-  } else {
-    return '0px';
-  }
-}
-
 export interface TimeOption {
   value: RelativeTimeRange;
   display: string;
@@ -57,7 +47,7 @@ export function TimeRangeSelector(props: TimeRangeSelectorProps) {
         '.MuiSelect-icon': {
           marginTop: '1px',
         },
-        '.MuiSelect-select': height ? { paddingY: heightToPaddingY(height) } : {},
+        '.MuiSelect-select': height ? { lineHeight: height, paddingY: 0 } : {},
       }}
     >
       {timeOptions.map((item, idx) => (
--- a/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
+++ b/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
@@ -19,11 +19,15 @@ import { formatAbsoluteRange } from './utils';
 
 const DATE_TIME_FORMAT = 'yyyy-MM-dd HH:mm:ss';
 
-const SIZE_TO_PADDING_Y = {
-  small: '4px',
-  medium: '6px',
-  large: '10px',
-};
+function heightToPaddingY(height: string): string {
+  const heightNum = parseFloat(height);
+
+  if (heightNum >= 20) {
+    return `${(heightNum - 20) / 2}px`;
+  } else {
+    return '0px';
+  }
+}
 
 export interface TimeOption {
   value: RelativeTimeRange;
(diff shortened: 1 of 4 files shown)


juliepagano:
The other way I could think of to _maybe_ approach this would be to add some custom variants for the impacted components to the MUI theme (e.g. `perses-toolbar` or something like that that wouldn't conflict with other internal styles) and override using those. That would give other consumers the ability the customize these button sizes. I'm curious what some folks with more MUI experience think about the tradeoffs between the two approaches since I'm still really new to it.

https://mui.com/material-ui/customization/theme-components/#creating-new-component-variants

juliepagano (commented):
The code changes generally look ok to me, but, as noted in my other comment, I want to wait to hear from folks with a little more experience with the project and MUI before giving a 👍 to merge this.

LukeTillman (commented):
This approach feels a little "brittle" to me since it's very tied to our integration's very specific height needs. Just spitballing, but a couple other possible options:

1. Allow the consumer to pass sx props to the nested components via some props on the overall TimeRangeControls (so in our integration, we could pass styles to make things look correct).
2. Add a container element to TimeRangeControls, let the user set the height there, then use styling on the sub-components to make it stretch/fill the overall container height.

MUI itself kinda uses a form of option 1 to allow folks to customize nested elements in more complicated components. For example, in [AutoCcomplete](https://mui.com/material-ui/api/autocomplete/) they expose things like `ChipProps`, `ListboxProps`, etc.

cndonovan:
Good ideas to mull over, thank you both!

saminzadeh:
> Allow the consumer to pass sx props to the nested components via some props on the overall TimeRangeControls (so in our integration, we could pass styles to make things look correct).

This makes sense to me as well. In general, this brings up more discussion about the "public" perses components that should bd exported for apps to consume and what type of customization they will allow.

sjcobb (approved):

ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
```diff
@@ -19,6 +19,16 @@ import { formatAbsoluteRange } from './utils';
 
 const DATE_TIME_FORMAT = 'yyyy-MM-dd HH:mm:ss';
 
+function heightToPaddingY(height: string): string {
```
juliepagano, line 22: It looks like you assume this is a string with a number and `px`, but I could see someone passing in other units (e.g. `rem`, `em`), which could break this functionality. I wonder if there's a slightly different way to write the CSS to better handle this -- let me poke at it a teeny bit to see if I can come up with a suggestion.

Alternatively (and much easier), maybe the prop can have a name (e.g. `heightPx` or some jsdoc over it to communicate that it has to be in pixels.

ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
```diff
@@ -19,6 +19,16 @@ import { formatAbsoluteRange } from './utils';
...
+  const heightNum = parseFloat(height);
+
+  if (heightNum >= 20) {
```
juliepagano, line 25: Why `20` here? Is this number dependent on something that can change in a MUI theme?

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -46,8 +50,8 @@ export function TimeRangeControls() {
 
   return (
     <>
-      <DateTimeRangePicker timeOptions={TIME_OPTIONS} value={timeRange} onChange={setTimeRange} />
-      <RefreshIconButton aria-label="Refresh Dashboard" onClick={refresh}>
+      <DateTimeRangePicker timeOptions={TIME_OPTIONS} value={timeRange} onChange={setTimeRange} height={height} />
+      <RefreshIconButton aria-label="Refresh Dashboard" onClick={refresh} sx={height ? { height } : {}}>
```
juliepagano, line 54: optional nit: Do you need this ternary here? I _think_ if height is undefined, it won't set it, but I'm not sure if there's anything underlying this that would get clobbered by that `undefined`.

juliepagano (commented):
Thanks for taking another run at this. I've got a few questions/nits and one bit I'm going to dig into to see if I can give you more actionable feedback.

ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
juliepagano: See follow-up comment later down for a suggestion to adjust.

ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
juliepagano: Feel free to ignore this if the alternative css suggestion removes the need for it.

ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
```diff
@@ -46,6 +57,7 @@ export function TimeRangeSelector(props: TimeRangeSelectorProps) {
         '.MuiSelect-icon': {
           marginTop: '1px',
         },
+        '.MuiSelect-select': height ? { paddingY: heightToPaddingY(height) } : {},
```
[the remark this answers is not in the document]
cndonovan: Love it, less code and less error-prone

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
cndonovan: ah yeah, seems like we can get away without it, so let's do that (testing it, it doesn't seem to add `height: undefined` to the CSS)

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -30,10 +30,20 @@ export const TIME_OPTIONS: TimeOption[] = [
   { value: { pastDuration: '14d' }, display: 'Last 14 days' },
 ];
 
-export function TimeRangeControls() {
+interface TimeRangeControlsProps {
+  // Height of the controls in pixels.
+  // The controls look best at heights >= 28 pixels.
+  // You can use values less than 28, but it won't look great.
```
eunicorn, line 36: 😆

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -30,10 +30,20 @@ export const TIME_OPTIONS: TimeOption[] = [
...
+  heightPx?: number;
+}
+
+export function TimeRangeControls({ heightPx }: TimeRangeControlsProps) {
   const { timeRange, setTimeRange, refresh } = useTimeRange();
   const defaultTimeRange = useDefaultTimeRange();
 
+  // Convert height as a number to height as a string, then use this value for styling
+  const height = heightPx === undefined ? undefined : `${heightPx}px`;
```
eunicorn, line 45: should we default height to 28px here?

eunicorn (approved):

juliepagano (approved):
This looks great! Thanks for working through all the rounds of feedback. 🚀

PR #834, opened 2022-12-02
[FEATURE] Discard Changes Confirmation Dialog

eunicorn:
Previously we had unsaved changes dialog and we got feedback that the dialog was confusing. Instead we will have a discard changes confirmation dialog. 

image

We also want to warn users if they have any unsaved changes when trying to cancel out of edit panel drawer or variable editor.

image

image


ui/dashboards/src/components/DeletePanelDialog/DeletePanelDialog.tsx
```diff
@@ -53,15 +53,17 @@ const DeletePanelForm = ({ deletePanelDialog }: DeletePanelFormProps) => {
   };
   return (
     <form onSubmit={handleDelete}>
-      <DialogContent sx={{ width: '500px' }}>
+      <DialogContent dividers sx={{ width: '500px' }}>
```
eunicorn, line 56: dialogs should have dividers and secondary button should have `outlined` variant

PR #840, opened 2022-12-05
Use 24hr time formatting in tooltip

sjcrank:
**Motivation**
Perses renders time using 24-hour formatting in the time selector and x-axis, but the chart tooltip uses 12-hour formatting resulting in an inconsistency that may be confusing to users.

Screen Shot 2022-12-05 at 3 43 47 PM

Screen Shot 2022-12-05 at 3 42 13 PM

**Proposal**
Update the Tooltip to also use 24-hour format, which will be consistent with the time selector and x-axis. In the future we may consider supporting a user-preference for customizable time formatting.

Screen Shot 2022-12-05 at 3 43 23 PM

Signed-off-by: Sam Crank <samuel.crank@chronosphere.io>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- [ENHANCEMENT] Use consistent formatting for time (24 hour) #840
 - [ENHANCEMENT] Add ability to specify time zone via TimeZoneProvider #825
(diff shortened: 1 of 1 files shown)


saminzadeh (approved):

sjcobb (approved):
Thanks for fixing this!

PR #844, opened 2022-12-06
Support snapshot UI releases

juliepagano:
Adds the ability to generate "snapshot" UI releases after pushing to branches prefixed with "snapshot/" to help with complex testing use cases.

## Notes for reviewers

- I recommend starting with the `RELEASE.md` change to get some context from the docs before reviewing the code.
- My bash is **very rusty**, so please let me know if any of that is ugly/suboptimal/etc.

### Versioning

In our Monday discussion, we talked about using a release based on the most recent version instead of my suggested approach of using `0.0.0`. I tried it out, but was concerned that it bumped the snapshots to the top of the versions list in the npm UI (see screenshot). I think this could be confusing to consumers of the project and might be misconstrued as prereleases. 

image

Based on this, I think it would be safer to stick with the `0.0.0` versioning. It will keep the snapshots below `latest` (the actual version most consumers will want). That versioning also communicates that it's probably something internal to the project not for external consumption and is unlikely to be confused with prereleases. Project developers can look up information about the snapshot by going to the associated branch in github, which is included in the tag and version name.

### Testing out the functionality

I created a simple test project that uses a barebones version of the scripts from perses, so I could safely play with this without breaking anything in the real perses project.

- https://github.com/juliepagano/julie-test-changesets
- https://www.npmjs.com/package/@julie-test-changesets/core?activeTab=versions
- https://www.npmjs.com/package/@julie-test-changesets/components?activeTab=versions

#### Screenshots

Create branch named `snapshot/test-refactor`
image

Note the latest git sha is `c66a944`
image

Snapshot releases made to npm for packages with version that includes the branch name and the git sha and a tag that includes the branch name.
image

image

Push a new commit (`d9cef4f`) to `snapshot/test-refactor`
image

Note a new snapshot version was published to the same tag.

image
image

After deleting the branch in github, the tags will automatically be deleted from npm. The versions will still remain.
image
image


scripts/ui_release.sh
```diff
@@ -9,6 +9,12 @@ cd ui/
 
 files=("../LICENSE" "../CHANGELOG.md")
 workspaces=$(jq -r '.workspaces[]' < package.json)
+publish_workspaces=$(for workspace in ${workspaces}; do 
+  # package "app" is private so we shouldn't try to publish it.
+  if [[ "${workspace}" != "app" ]]; then
+    echo $workspace;
+  fi 
+done)
```
juliepagano, line 17: This value is now used in multiple places, so pulling it out into a variable.

scripts/ui_release.sh
```diff
@@ -74,13 +77,81 @@ function bumpVersion() {
   fi
   # upgrade the @perses-dev/* dependencies on all packages
   for workspace in ${workspaces}; do
-    sed -E -i "" "s|(\"@perses-dev/.+\": )\".+\"|\1\"\^${version}\"|" "${workspace}"/package.json
+    # sed -i syntax is different on mac and linux
+    if [[ "$OSTYPE" == "darwin"* ]]; then
```
juliepagano, line 81: I ran into this trying to automate this in github. Previously, I think it was only run on people's machines.

scripts/ui_release.sh
```diff
@@ -74,13 +77,81 @@ function bumpVersion() {
...
 
   # increase the version on all packages
   npm version "${version}" --workspaces
 }
 
+# Validates branch name and extracts snapshot name from it.
+function getBranchSnapshotName() {
+  branch="${1}"
+
+  snapshotRegex="^snapshot\/[-a-zA-Z0-9]+$"
```
juliepagano, line 96: Forcing kebab-case here for consistency and because that's a safe syntax for npm tags/versions. Let me know if anyone has strong feelings about supporting something more complex.

scripts/ui_release.sh
```diff
@@ -74,13 +77,81 @@ function bumpVersion() {
...
+  fi
+
+  # Replace "/"" with "-"
+  echo "$(echo "${branch}" | sed -E  's/\//-/')"
+}
+
+function snapshotVersion() {
+  # Use version 0.0.0 to keep snapshots at the bottom of the npm versions UI to 
+  # avoid confusion for consumers of the package. This also helps differentiate
+  # snapshots from the concept of prereleases.
```
juliepagano, line 110: See "notes for reviewers" for more details on why I reverted back to 0.0.0.

RELEASE.md
```diff
@@ -1,5 +1,4 @@
-Releases
```
juliepagano, line 1: I think prettier in vscode updated this automatically on me. The visual output is the same in github, but let me know if you'd rather I switch it back.

.github/workflows/ci.yaml
```diff
@@ -69,6 +74,22 @@ jobs:
...
+      - name: Use snapshot version
+        ## This step only runs for "snapshot" branches starting with "snapshot/"
+        ## We have to snapshot-version again in this job because the version
+        ## change in build-frontend only impacts the compiled assets. We need
+        ## the correct version again here for the publish to work properly.
+        if: ${{ github.event_name == 'push' && startsWith(github.ref_name, 'snapshot/') }}
+        run: ./scripts/ui_release.sh --snapshot-version "${{ github.ref_name }}" "${{ github.sha }}" && cd ./ui && npm install
+      - name: Publish snapshot libraries
+        ## This step only runs for "snapshot" branches starting with "snapshot/"
+        if: ${{ github.event_name == 'push' && startsWith(github.ref_name, 'snapshot/') }}
```
juliepagano, line 86: Just to be extra-sure because I'm still new-ish to github actions, can a reviewer validate for me that this will only be able to run when a maintainer does this (i.e. we can't have any shenanigans with someone external trying to do something malicious with it).

scripts/ui_release.sh
```diff
@@ -74,13 +77,81 @@ function bumpVersion() {
...
+
+function removeSnapshot() {
+  branch="${1}"
+  tagName=$(getBranchSnapshotName $branch)
+
+  echo "Removing snapshot for tag ${tagName}"
+  cmd="npm publish --access public --tag ${tagName}"
+  for workspace in ${publish_workspaces}; do
+    cd "${workspace}"
+    eval "npm dist-tag rm @julie-test-changesets/${workspace} ${tagName}"
```
Nexucis, line 150: i guess it's not `@julie-test-changesets` anymore :)

scripts/ui_release.sh
```diff
@@ -74,13 +77,81 @@ function bumpVersion() {
...
+    cd ../
+  done
+}
+
+function removeSnapshot() {
+  branch="${1}"
+  tagName=$(getBranchSnapshotName $branch)
+
+  echo "Removing snapshot for tag ${tagName}"
+  cmd="npm publish --access public --tag ${tagName}"
```
Nexucis, line 147: it doesn't seem you are using this variable. Probably a copy/past issue ;)

RELEASE.md
Nexucis: that's fine. We should probably put in place a markdown formateur like for the rest of the code.

.github/workflows/ci.yaml
```diff
@@ -69,6 +74,22 @@ jobs:
           # as the placeholder for the auth token
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      - name: Use snapshot version
```
Nexucis, line 77: I would merge the steps `Use snapshot version` and `publish snapshot libraries` since there are correlated

.github/workflows/ci.yaml
```diff
@@ -69,6 +74,22 @@ jobs:
...
+        ## This step only runs for "snapshot" branches starting with "snapshot/"
+        ## We have to snapshot-version again in this job because the version
+        ## change in build-frontend only impacts the compiled assets. We need
+        ## the correct version again here for the publish to work properly.
```
Nexucis, line 81: it took me a while to understand exactly what was the issue.
So I would suggest this comment instead: 

```
The previous job that is building the UI assets doesn't persist the different files `package.json` that has been modified by the script.
So in order to be able to publish the snapshot, we need to ensure the different files `package.json` contain the correct version. That's why we are running again the script to bump a snapshot version.

.github/workflows/ci.yaml
Nexucis: yeah that should work

Nexucis (approved):
I added a couple of comments, but other all it looks ok to me :)

eunicorn (approved):
can't wait to try this out! thanks julie! 🚀

.github/workflows/ci.yaml
juliepagano: Thanks for the rewording! I was struggling with how to communicate this well.

juliepagano:
I did one final test on my test repo before merging this and found an edge case I hadn't hit before. The use of carets in the cross-linked internal dependencies (e.g. `components` depending on `core`) could lead to some unexpected snapshot versions being used because npm seems to sort versions reverse alphabetically for the text after the semver content (e.g. `0.0.0-zzzz` will get used over `0.0.0-aaaa`). I think I didn't hit this with my previous testing because of some luck with branch naming.

Based on talking with @LukeTillman  and @sjcobb, there's not a compelling reason why we're using carets for our internal dependencies, so I created https://github.com/perses/perses/pull/846 to adjust that (see more details in the PR). I thought it was good to introduce that change in a separate PR for clarity in the repo. I've also updated this PR to account for the changes.

scripts/ui_release.sh
LukeTillman: Oh wow, TIL!

LukeTillman (approved):
Looks good to me. 🚀

PR #845, opened 2022-12-06
[ENHANCEMENT][WIP] legend selector UX improvements

sjcobb:
## Overview
- Add display names for each legend position property to capitalize 'bottom' and 'right'
- Disable position control when legend is hidden
- Refactor to use MUI Autocomplete instead of Select to be consistent with calculations, unit form controls
- Create separate LegendSelector.tsx component, move to @perses-dev/components package

## Screenshots
Autocomplete open state:
image

Disabled position when legend hidden:
image


ui/components/src/LegendOptionsEditor/LegendOptionsEditor.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export interface LegendOptionsEditorProps {
+  value?: LegendOptions;
+  onChange: (legend?: LegendOptions) => void;
+}
+
+export function LegendOptionsEditor({ value, onChange }: LegendOptionsEditorProps) {
+  const handleLegendShowChange: SwitchProps['onChange'] = (_: unknown, checked: boolean) => {
+    // legend is hidden when legend obj is undefined
+    const legendValue = checked === true ? {} : undefined;
```
LukeTillman, line 48: Nit: this is really about the overall `LegendOptions` type I think, but seems weird to have an empty object be a valid value here (I think I expected `position` to be required).

LukeTillman (approved):
Nice work!

ui/components/src/LegendOptionsEditor/LegendOptionsEditor.tsx
sjcobb: Yeah this makes sense! It was pretty quick to change, I did it as a [separate PR here](https://github.com/perses/perses/pull/848) to be safe.

ui/components/src/LegendOptionsEditor/LegendOptionsEditor.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+  it('can change legend visibility by clicking', () => {
+    const onChange = jest.fn();
+    renderLegendOptionsEditor(undefined, onChange);
+
+    userEvent.click(getLegendShowSwitch());
+
+    expect(onChange).toHaveBeenCalledWith({});
+  });
+
+  // TODO: additional test for legend position change, disabled state
```
juliepagano, line 41: Is this something you intend to do as part of this PR or is it a note for a follow-up?

sjcobb: A follow up! There is still a glitch when you switch legend.position from 'right' to 'bottom' that I'm planning to fix (due to lineChartHeight condition in TimeSeriesChartPanel), so planning to add more tests then (I'm also keeping a look out for Luke's PR to make writing integration tests easier since we really need more of those).

PR #847, opened 2022-12-07
[ENHANCEMENT] Bytes unit default changed to unabbreviated

sjcobb:
## Overview
We have gotten some feedback that `unit: { kind: 'Bytes' }` is counterintuitive since it defaults to the abbreviated version (KB, MB, GB, etc.). This PR, changes this default to always show "Bytes" after the value, unless `unit: { kind: 'Bytes', abbreviate: true }` is set explicitly. Other commits include:
- formatBytes refactoring to support `abbreviate` (if abbreviate is undefined set to false)
- add condition that checks for out of range decimal_places
- add additional tests for formatValue function, update previous tests

## Screenshots
abbreviate set to true
image

abbreviate set to false (TimeSeriesChart)
image

abbreviate set to false (StatChart)
image


ui/components/src/model/units/bytes.ts
```diff
@@ -31,16 +34,37 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
+  const options = unitOptions.abbreviate === undefined ? { ...unitOptions, abbreviate: false } : unitOptions;
+
   if (bytes === 0) return '0 Bytes';
 
+  let decimals = options.decimal_places ?? DEFAULT_DECIMAL_PLACES;
+  // avoids minimumFractionDigits value is out of range error
+  if (decimals < 0) {
+    decimals = 0;
+  } else if (decimals > 100) {
+    decimals = 100;
```
LukeTillman, line 48: Just curious, is `100` just an arbitrary number or is that a real limitation of some code in here?

ui/components/src/model/units/bytes.ts
```diff
@@ -31,16 +34,37 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
   const k = 1024;
-  const dm = decimals < 0 ? 0 : decimals;
   const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
 
   // Math.max(0, ...) ensures that we don't return -1 as a value for the index.
   // Why? When the number of bytes are between -1 and 1, Math.floor(Math.log(bytes)/Math.log(1024)) returns -1.
   const i = Math.max(0, Math.floor(Math.log(bytes) / Math.log(k)));
 
-  return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
+  return parseFloat((bytes / Math.pow(k, i)).toFixed(decimals)) + ' ' + sizes[i];
```
LukeTillman, line 69: Is it weird that above, we're using the `Intl.NumberFormat` API but down here, we're using `toFixed`? Would this benefit at all from using the same thing in either case?

ui/components/src/model/units/units.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      unit: { kind: 'Years' },
+      expected: '300.00 years',
+    },
+  ];
+
+  tests.forEach(({ value, unit, expected }) => {
+    it(`formats ${value} as ${JSON.stringify(unit)}`, () => {
+      expect(formatValue(value, unit as UnitOptions)).toEqual(expected);
+    });
+  });
```
LukeTillman, line 109: Suggestion: maybe look at converting this into using the built-in `test.each` format in Jest? See:

https://jestjs.io/docs/27.x/api#testeachtablename-fn-timeout

If you want an example, I used it when I was playing around with using Zustand for form state recently [here](https://github.com/LukeTillman/zustand-form-state/blob/main/src/__tests__/selector.ts#L40).

LukeTillman (approved):
Just a couple questions/minor things inline, but looks great. Love the new tests.

ui/components/src/model/units/bytes.ts
sjcobb: So at one point, I saw this [limit of 100](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#exceptions) for toFixed (or else a RangeError is thrown), but turns out the limit [should actually be 20](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#minimumfractiondigits) since I switched to use Intl.NumberFormat so I'll update this line. This avoids highly unlikely edge case of:
image

sjcobb: Also, to note we already limit it to 4 when editing a panel in the UI:

image

ui/components/src/model/units/bytes.ts
sjcobb: Yeah good point, I was a little scared to change this logic, but agreed it'll be simpler to update it above to just use `toFixed` as well

ui/components/src/model/units/bytes.ts
sjcobb: nvm saw your comment below, I'm going to change this to toFixed so 100 will be correct 😆

ui/components/src/UnitSelector/UnitSelector.test.tsx
```diff
@@ -182,7 +181,7 @@ describe('UnitSelector', () => {
 
     it('does not allow the user to set abbreviate', () => {
       renderUnitSelector({ kind: 'Bytes' });
-      expect(getAbbreviateSwitch()).toBeDisabled();
+      expect(getAbbreviateSwitch()).toBeEnabled();
```
cndonovan, line 184: Change the test title too? (remove "not")?

ui/components/src/model/units/bytes.ts
```diff
@@ -31,16 +34,37 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
 };
 
 // https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript/18650828#18650828
-export function formatBytes(bytes: number, decimals = 2) {
+export function formatBytes(bytes: number, unitOptions: BytesUnitOptions) {
+  // default to full 'Bytes' formatting
+  const options = unitOptions.abbreviate === undefined ? { ...unitOptions, abbreviate: false } : unitOptions;
+
   if (bytes === 0) return '0 Bytes';
```
cndonovan, line 41: If you wanted, this line could still go first

ui/components/src/model/units/bytes.ts
```diff
@@ -31,16 +34,37 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
 };
 
 // https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript/18650828#18650828
-export function formatBytes(bytes: number, decimals = 2) {
+export function formatBytes(bytes: number, unitOptions: BytesUnitOptions) {
+  // default to full 'Bytes' formatting
+  const options = unitOptions.abbreviate === undefined ? { ...unitOptions, abbreviate: false } : unitOptions;
```
cndonovan, line 39: I wonder if it would be nice if this default was built into the form itself? 

So like, I imagining a radio button with only two choies, and it defaults to false. Then, the value is always either true or false, so we don't need to do this.

But I dunno, kinda gets into "what's best way to handle defaults", and I think it's nice that in this code both defaults are being handled together in this same function

ui/components/src/model/units/units.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    },
+    {
+      value: 0.1,
+      unit: { kind: 'PercentDecimal' },
+      expected: '10.00%',
+    },
+    {
+      value: 100,
+      unit: { kind: 'Bytes', decimal_places: 0, abbreviate: false },
+      expected: '100 Bytes',
```
cndonovan, line 46: I kinda want to lowercase `bytes` :) but not enough to actually go find out what everyone thinks, `Bytes` is good too

cndonovan (approved):

ui/components/src/model/units/bytes.ts
sjcobb: I might be misunderstanding, but I think you have to handle it in the code since users could just edit the JSON and aren't necessarily using form controls in panel edit drawer? There might be a better way to do this though so talking through the best way to handle defaults sounds great since it isn't consistent currently.

cndonovan: Ahhhhh very good point, I hadn't thought of that

ui/components/src/model/units/units.test.ts
sjcobb: Checked with Karan and he agrees too, just made this update!

ui/components/src/model/units/units.test.ts
sjcobb: I'm planning to come back to this later! Want to try and get this release out this afternoon

PR #848, opened 2022-12-07
[BREAKINGCHANGE] legend.position now required in time series panel

sjcobb:
For context, see [Luke feedback](https://github.com/perses/perses/pull/845#discussion_r1042324995) on #845 about how allowing an empty object as a valid legend property is counterintuitive, this PR makes changes in legend options so now `position` is required. Cue schema and existing legend tests are also updated.

_After this PR, using `{}` as a legend value technically won't throw an error in the UI since DEFAULT_LEGEND is already merged using lodash when `legend` is [defined in the panel here](https://github.com/perses/perses/blob/dd2cd966efd6468cf366d925531708eb4568120c/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx#L50), but to be safe I'm labeling this a BREAKINGCHANGE._


schemas/panels/time-series/time-series.cue
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 #legend: {
-	position?: "bottom" | "right"
+	position: "bottom" | "right"
```
sjcobb, line 21: @AntoineThebaud is this still all I need to do to make this change, will the new migration code automatically work?

LukeTillman (approved):
Nice, thanks for following up.

AntoineThebaud (approved):

PR #855, opened 2022-12-09
[FEATURE] Add tooltips to icons in panel and panel group header

eunicorn:
These tooltips will show up after 0.5 sec

image

image

### DIFF
--- a/ui/components/src/InfoTooltip/InfoTooltip.tsx
+++ b/ui/components/src/InfoTooltip/InfoTooltip.tsx
@@ -33,15 +33,27 @@ interface InfoTooltipProps {
   id?: string;
   title?: string;
   placement?: TooltipPlacement;
+  enterDelay?: number; // default is 500ms
+  enterNextDelay?: number; // default is 500ms
 }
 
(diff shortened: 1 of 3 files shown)


ui/components/src/InfoTooltip/InfoTooltip.tsx
```diff
@@ -33,15 +33,27 @@ interface InfoTooltipProps {
...
+  enterNextDelay,
+}: InfoTooltipProps) => {
   return (
     <StyledTooltip
       arrow
       id={id}
-      placement={placement}
+      placement={placement ?? TooltipPlacement.Top}
       title={<TooltipContent title={title} description={description} />}
+      enterDelay={enterDelay ?? 500}
```
sjcobb, line 55: Curious, why are these delays needed?

sjcobb (approved):
LGTM!

ui/components/src/InfoTooltip/InfoTooltip.tsx
eunicorn: once you learn what the icons mean, the tooltips will be very distracting if they show up every time.

discussed with @foxbat07 and agreed that 500ms seems acceptable

PR #860, opened 2022-12-09
[FEATURE] Add new endpoints to validate dashboard and datasource

Nexucis:
This PR add three new endpoints : 
- `/api/validate/dashboards`
- `/api/validate/datasources`
- `/api/validate/globaldatasources`

these endpoints are used by the CLI to validate dashboard and datasources using the remote server.

### DIFF
--- a/internal/api/core/server.go
+++ b/internal/api/core/server.go
@@ -25,6 +25,7 @@ import (
 	"github.com/perses/perses/internal/api/impl/v1/globaldatasource"
 	"github.com/perses/perses/internal/api/impl/v1/health"
 	"github.com/perses/perses/internal/api/impl/v1/project"
+	validateendpoint "github.com/perses/perses/internal/api/impl/validate"
 	"github.com/perses/perses/internal/api/shared"
 	"github.com/perses/perses/internal/api/shared/dependency"
 )
@@ -52,6 +53,7 @@ func NewPersesAPI(serviceManager dependency.ServiceManager, cfg config.Config) e
 	apiEndpoints := []endpoint{
 		configendpoint.New(cfg),
 		migrateendpoint.New(cfg.Schemas),
+		validateendpoint.New(serviceManager.GetSchemas()),
 	}
 	return &api{
 		apiV1Endpoints: apiV1Endpoints,
--- a/internal/api/impl/validate/validate.go
+++ b/internal/api/impl/validate/validate.go
@@ -51,22 +51,18 @@ func (e *Endpoint) ValidateDashboard(ctx echo.Context) error {
 }
 
 func (e *Endpoint) ValidateDatasource(ctx echo.Context) error {
-	entity := &v1.Datasource{}
-	if err := ctx.Bind(entity); err != nil {
-		return shared.HandleError(fmt.Errorf("%w: %s", shared.BadRequestError, err))
-	}
-	if err := validate.Datasource(entity, nil, e.sch); err != nil {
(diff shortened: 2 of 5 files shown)


internal/api/core/server.go
```diff
@@ -52,6 +53,7 @@ func NewPersesAPI(serviceManager dependency.ServiceManager, cfg config.Config) e
 	apiEndpoints := []endpoint{
 		configendpoint.New(cfg),
 		migrateendpoint.New(cfg.Schemas),
+		validateendpoint.New(serviceManager.GetSchemas()),
```
AntoineThebaud, line 56: I think we should be consistent here so either have
```go
migrateendpoint.New(cfg.Schemas),
validateendpoint.New(cfg.Schemas),
```
or
```go
migrateendpoint.New(serviceManager.GetSchemas()),
validateendpoint.New(serviceManager.GetSchemas()),
```

internal/api/impl/validate/validate.go
```diff
@@ -0,0 +1,73 @@
...
+	if err := ctx.Bind(entity); err != nil {
+		return shared.HandleError(fmt.Errorf("%w: %s", shared.BadRequestError, err))
+	}
+	if err := validate.Dashboard(entity, e.sch); err != nil {
+		return fmt.Errorf("%w: %s", shared.BadRequestError, err)
+	}
+	return ctx.NoContent(http.StatusOK)
+}
+
+func (e *Endpoint) ValidateDatasource(ctx echo.Context) error {
```
AntoineThebaud, line 53: What about something like this ?

```go
func (e *Endpoint) ValidateDatasource(ctx echo.Context) error {
	return e.validateDatasource(&v1.Datasource{}, ctx)
}

func (e *Endpoint) ValidateGlobalDatasource(ctx echo.Context) error {
	return e.validateDatasource(&v1.GlobalDatasource{}, ctx)
}

func (e *Endpoint) validateDatasource[T modelV1.DatasourceInterface](entity T, ctx echo.Context) error {
	if err := ctx.Bind(entity); err != nil {
		return shared.HandleError(fmt.Errorf("%w: %s", shared.BadRequestError, err))
	}
	if err := validate.Datasource(entity, nil, e.sch); err != nil {
		return shared.HandleError(fmt.Errorf("%w: %s", shared.BadRequestError, err))
	}
	return ctx.NoContent(http.StatusOK)
}
```

sjcobb (approved):
🚀 Wow you're fast! Can't wait to try these out

internal/api/core/server.go
Nexucis: I don't get why we need consistency here. the Migrate endpoint doesn't use the service `schemas` while it is used by the Validate endpoints.

internal/api/impl/validate/validate.go
Nexucis: oh yes you are right. We should do this.

Nexucis: ah actually that's not possible to do that today. See https://github.com/golang/go/issues/49085

Nexucis: I defined the function `validateDatasource` which isn't a method of the endpoint anymore and like that it works

internal/api/core/server.go
AntoineThebaud: Ah nevermind I thought serviceManager.GetSchemas() was returning the schemas..

AntoineThebaud (approved):

PR #842, opened 2022-12-05
[ENHANCEMENT] Variable UX fixes

saminzadeh:
- Changing some variable copy in the editor (e.g. drop the "Variable" prefix)
image
- Better names of the Variable plugins
image
- Tooltip to tell user to press enter in the Static List (otherwise the values will get erased)
image
- Ability to copy the variables values:
image
- Allow pasting comma separated values into the Static List options Input 
- Remove the "Hide variables" option in Edit mode
image
- [BugFix] Custom Label is now used for TextVariables

Signed-off-by: Shan Aminzadeh <shan.aminzadeh@chronosphere.io>

PR #866, opened 2022-12-12
[BUGFIX] show variables when not in edit mode

saminzadeh:
Fixes a bug in https://github.com/perses/perses/pull/842


saminzadeh:
Omitting from changelog since it hasn't been released yet

sjcobb (approved):

PR #870, opened 2022-12-13
[BUGFIX] Fix variable collisions with promQL function

### LINKED ISSUE #859: label_replace() replacement argument is wrongly rejected as unknown variable
See [label_replace()](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace) documentation. The `replacement` argument can use capturing groups like `$1`, `$2` that get wrongly interpreted as dashboard variables by the backend validation, which makes it fail.

```json
{
    "message": "bad request: variable \"1\" is used in the variable \"extlabels_prometheus\" but not defined"
}
```

Nexucis:
This PR fixed #859 

We are going to forbid the possibility to name a variable with just a number. It's not a meaningful name anyway, so it's not that bad to do that.

That should be a way to mitigate the issue #859 and to avoid to detect additional variable coming from different language.

### DIFF
--- a/pkg/model/api/v1/dashboard/variable.go
+++ b/pkg/model/api/v1/dashboard/variable.go
@@ -16,6 +16,7 @@ package dashboard
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	"github.com/perses/perses/pkg/model/api/v1/common"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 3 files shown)


pkg/model/api/v1/dashboard/variable.go
```diff
@@ -127,6 +128,9 @@ func (v *TextVariableSpec) validate() error {
 	if err := common.ValidateID(v.Name); err != nil {
 		return err
 	}
+	if _, err := strconv.Atoi(v.Name); err == nil {
+		return fmt.Errorf("variable name cannot contain only number. That's not a meaningfull name for a variable")
```
AntoineThebaud, line 132: _only a number_
or
_only digits_

meaningfu**l**

PR #879, opened 2022-12-15
Wrape codemirror around a box and set a default height

Nexucis:
By default, codemirror doesn't have a maximum height. That can be ok for tiny JSON. It's less ok when you have thousand line (like in the migration page).

Also Codemirror doesn't seem to be well constraint by the width of the container like `Stack`. So I wrapped it around a `Box` and allow user to give a pre-determine `width` if necessary

### DIFF
--- a/ui/components/src/JSONEditor.tsx
+++ b/ui/components/src/JSONEditor.tsx
@@ -15,20 +15,15 @@ import { useEffect, useState } from 'react';
 import CodeMirror from '@uiw/react-codemirror';
 import { json, jsonParseLinter } from '@codemirror/lang-json';
 import { linter, lintGutter } from '@codemirror/lint';
-import { Box, useTheme } from '@mui/material';
+import { useTheme } from '@mui/material';
+import { ReactCodeMirrorProps } from '@uiw/react-codemirror/src';
 
-interface JSONEditorProps<Spec> {
-  value: Spec;
-  onChange?: (next: Spec) => void;
-  maxHeight?: string;
-  height?: string;
-  width?: string;
-  maxWidth?: string;
-}
+type JSONEditorProps<T> = Omit<ReactCodeMirrorProps, 'onBlur' | 'theme' | 'extensions' | 'onChange' | 'value'> & {
+  value: T;
+  onChange?: (next: T) => void;
+};
 
 export function JSONEditor<T>(props: JSONEditorProps<T>) {
-  const { onChange, height, width, maxWidth } = props;
-  const maxHeight = props.maxHeight ?? '50rem';
   const theme = useTheme();
   const isDarkMode = theme.palette.mode === 'dark';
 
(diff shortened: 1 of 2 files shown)


juliepagano:
When the json editor in the panel gets large, you can end up with double-scrollbars, which can be a confusing or awkward user experience. I couldn't reproduce this with what's on `main`, so guessing something about the box wrapper. I don't have time to poke at this to see if there's an easy adjustment right now, but will try to circle back later in the day when I free from meetings. Shan might know offhand since he's got more experience with codemirror.

image

saminzadeh:
Not sure if I'm fully following the issue (I also can't seem to use Migrate on my local dev?). Getting a 500

image

Can you record a Loom or add some screenshots to clarify? There may be a better way to do this.

Nexucis:
Ah yes so sorry about that, I should have recorded an example of the issue.

So before this PR here what we have:

Screen Recording 2022-12-16 at 09 04 48

With this PR, it becomes like that:

Screen Recording 2022-12-16 at 09 10 14

Regarding the `500` error, I rebased the PR and it fixed the issue. You might need to rebuild the backend.

Is it clearer like that ?

ui/components/src/JSONEditor.tsx
```diff
@@ -33,24 +39,29 @@ export function JSONEditor<T>(props: JSONEditorProps<T>) {
...
-      value={value}
-      onChange={(newValue) => {
-        setValue(newValue);
-      }}
-      onBlur={() => {
-        try {
-          const json = JSON.parse(value ?? '{}');
-          if (props.onChange !== undefined) {
-            props.onChange(json);
+    <Box maxWidth={maxWidth} width={width}>
```
saminzadeh, line 42: The `Box` is not needed, you can just pass these props directly to CodeMirror.

It might be even better to have JSONEditor extend CodeMirror and only override the properties that make it a json specific.

Rough idea
```tsx
type JSONEditorProps = Omit<CodeMirrorProps, 'onBlur' | 'theme' | 'extensions'>

function JSONEditor(props: JSONEditorProps) {
  return (
     <CodeMirror 
         {...props}
         extensions={[json(), linter(jsonParseLinter()), lintGutter()]}
      />
}
```

saminzadeh:
> Ah yes so sorry about that, I should have recorded an example of the issue.
> 
> So before this PR here what we have:
> 
> Screen Recording 2022-12-16 at 09 04 48 [ Screen Recording 2022-12-16 at 09 04 48 ](https://user-images.githubusercontent.com/4548045/208053747-ed1b78e4-456f-4a38-b2c0-87471c00b56d.gif) [ ](https://user-images.githubusercontent.com/4548045/208053747-ed1b78e4-456f-4a38-b2c0-87471c00b56d.gif)
> 
> With this PR, it becomes like that:
> 
> Screen Recording 2022-12-16 at 09 10 14 [ Screen Recording 2022-12-16 at 09 10 14 ](https://user-images.githubusercontent.com/4548045/208054006-d3ca437f-92dd-4fc0-a68c-6a5b50f219e7.gif) [ ](https://user-images.githubusercontent.com/4548045/208054006-d3ca437f-92dd-4fc0-a68c-6a5b50f219e7.gif)
> 
> Regarding the `500` error, I rebased the PR and it fixed the issue. You might need to rebuild the backend.
> 
> Is it clearer like that ?

Yep thanks! Made a comment on the line. If anything, you can just wrap the JSONEditor with a Box in the "Migrate" section.

ui/components/src/JSONEditor.tsx
Nexucis: ah yeah that's definitely better like that. Thanks @saminzadeh !

Nexucis:
should be good for another review now :)

saminzadeh (approved):
LGTM! Thanks for those changes

PR #868, opened 2022-12-12
[IGNORE] Use the new name for the extracted changelog

juliepagano:
Signed-off-by: Julie Pagano <julie.pagano@chronosphere.io>

PR #884, opened 2022-12-16
[IGNORE] Reduce duplicate github actions

juliepagano:
Prior to this change, the github actions in `ci.yaml` would run on pull request and push events. This means that when maintainers work in branches, they trigger these actions twice when working in pull requests. This is an unnecessary use of actions resources and can slow down getting work done (e.g. slowed down getting #868 done to unblock a release).

After this change, the github actions in `ci.yaml` will continue to run on pull request events. It will modify the behavior to only run on push events when they match one of the following:
- `main` branch: default project branch
- `release/*` branch: branches used to prepare releases
- `snapshot/*` branch: branches used to release one-off snapshots for testing
- `v*` tag: release tags.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,5 @@
 name: ci
 on:
-  pull_request:
   push:
     branches:
       - main
@@ -8,6 +7,7 @@ on:
       - snapshot/*
     tags:
       - v*
+  pull_request:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,6 +2,12 @@ name: ci
 on:
   pull_request:
   push:
+    branches:
+      - main
+      - release/*
+      - snapshot/*
+    tags:
+      - v*
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -2,6 +2,12 @@ name: ci
 on:
   pull_request:
   push:
+    branches:
+      - main
+      - release/*
+      - snapshot/*
+    tags:
+      - v*
```
juliepagano, line 10: Are there any other special case branches or tags we'd want to run on "push"?

Nexucis: I don't see other cases for the moment.

But I think it will be interesting to do the same for other ci files. So we have a consistent way to run the different job regardless their purpose.

juliepagano: Good call! The other files were set up to only run on push to `main`. I'll adjust them to include these other items too.

.github/workflows/go.yml
```diff
@@ -22,59 +26,58 @@ jobs:
     name: "check code format"
     runs-on: ubuntu-latest
     steps:
```
juliepagano, line 28: vscode auto-formatted the yaml a bit when I touched this. I think this is more consistent with the format in `ci.yaml` and a bit more human-readable, so I'm inclined to keep it. Let me know if you'd rather not, and I can take the whitespace adjustments out.

.github/workflows/ci.yaml
juliepagano: Done!

Nexucis (approved):
nice :). Thank you @juliepagano for this improvement !

PR #885, opened 2022-12-19
Add CLI documentation

Nexucis:


docs/cli.md
```diff
@@ -0,0 +1,309 @@
...
+Note that it doesn't necessary mean you won't face any issues when applying them.
+
+```bash
+$ percli lint -f ./resource.json
+```
+
+By default, the command doesn't require any remote server. We are providing a flag `--online` that will tell the CLI to
+use a remote Perses server for additional validation. For example, when it will have to validate a dashboard, it will
+use the endpoint `/api/validate/dashboards`. That can be useful if you want to be sure that your dashboard is compatible
+with the server (because it will use the remote CUE schemas instead of the local one)
```
AntoineThebaud, line 256: I'd say sth more high level like "because it will match the plugins known to the server instead of the local ones"

docs/cli.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Migrate from Grafana dashboard to Perses format
+
+The command `migrate` is for the moment only use to translate a Grafana dashboard to the Perses format. This command
```
AntoineThebaud, line 260: use**d**

docs/cli.md
```diff
@@ -0,0 +1,309 @@
+Perses CLI (percli)
+==================
+
+On top of the application, we also provide a CLI named `percli`. This tool can be used to interact with the backend REST
+API to manage the resources such as dashboard, datasource, project etc.
```
AntoineThebaud, line 5: dashboard**s**, datasource**s**, project**s**

docs/cli.md
Nexucis: ah yes good idea. Thanks !

AntoineThebaud (approved):

PR #888, opened 2022-12-19
[IGNORE] Add script to automate api backend for dev

juliepagano:

Hoping this will speed up a few steps during local dev for UI-focused devs. This may also be helpful for starting things up quickly in CI if we end up setting up end-to-end tests against the app in the future.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,17 +1,16 @@
-Contributing
-============
+# Contributing
 
 As Perses is still a work in progress, the contribution process is still evolving.
 
 We are using GitHub as our main development and discussion forum.
 
-* All PRs should go there.
-* We use pull requests and issues for tracking the development of features that are either uncontroversial and/or small
+- All PRs should go there.
+- We use pull requests and issues for tracking the development of features that are either uncontroversial and/or small
   and don't need much up-front discussion.
-* If you are thinking about contributing something more involved, you can use
+- If you are thinking about contributing something more involved, you can use
   the [GitHub discussions](https://github.com/perses/perses/discussions) feature for design discussions before sending a
   pull request or creating a feature request issue.
-* Be sure to add [DCO signoffs](https://github.com/probot/dco#how-it-works) to all of your commits.
+- Be sure to add [DCO signoffs](https://github.com/probot/dco#how-it-works) to all of your commits.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,17 +1,16 @@
-Contributing
-============
+# Contributing
```
juliepagano, line 1: Prettier auto-formatted the markdown again. 😞

scripts/api_backend_dev.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+## /!\ This file must be used at the root of the perses project
+## This script automates the steps for running the api backend server during
+## development.
+
```
juliepagano, line 6: Keeping this super-simple to start. We can always add more later.

scripts/api_backend_dev.sh
```diff
@@ -0,0 +1,19 @@
...
+# Populate database
+echo ">> populate the local database"
+cd dev 
+bash populate.sh
```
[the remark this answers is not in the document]
Nexucis:
```suggestion
rm -rf local_db
./populate.sh
```

CONTRIBUTING.md
Nexucis: perhaps we should have a linter for markdown, so we have a consistent markdown formatter and we would avoid this kind of change

juliepagano: I think my vscode is using `prettier`, which we also use for our UI code, so that might be the easiest answer. Out of curiosity, what text editor are you using? It'd be nice to pick something that integrates easily for everyone. I'll create an issue with some notes, so we can dig into this in a follow-up.

Nexucis: I'm using IntelliJ for everything

scripts/api_backend_dev.sh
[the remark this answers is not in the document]
juliepagano: Good call. Done!

CONTRIBUTING.md
juliepagano: 👍  created https://github.com/perses/perses/issues/889 to track looking into this for when one of us has time.

Nexucis (approved):
👍

PR #891, opened 2022-12-21
[ENHANCEMENT] GaugeChart panel ability to visually edit `max`

sjcobb:
## Overview

Currently, our GaugeChart panel does not have a way to visually edit the `max` property, this PR adds a TextField form control to change the default. 

_**Explicitly defining a max is only required for units other than Percent and PercentDecimal**_ 

In the future, we should offer additional customization options for thresholds, min, and ability to switch 'Threshold Mode' between 'Absolute' and 'Percentage' (see TODOs below)

## Future TODOs
- [ ] form controls to edit thresholds / step value and color properties in GaugeChart and TimeSeriesChart (this will be a larger future project to build an intuitive UX for this)
- [ ] threshold mode support, example:
  - image

## Screenshot

image


ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
```diff
@@ -47,14 +47,14 @@ export function createInitialGaugeChartOptions(): GaugeChartOptions {
       },
     },
     calculation: 'LastNumber',
-    unit: { kind: 'Percent' },
+    unit: { kind: 'PercentDecimal' },
```
sjcobb, line 50: I updated this default unit when adding a new GaugeChart panel since I've noticed that `PercentDecimal` gets used more often, but I'm open to reverting this change

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditorSettings.tsx
```diff
@@ -53,6 +55,26 @@ export function GaugeChartOptionsEditorSettings(props: GaugeChartOptionsEditorPr
...
+                  // max only needs to be set explicitly for units other than Percent and PercentDecimal
+                  const newValue = e.target.value ? Number(e.target.value) : undefined;
+                  onChange(
+                    produce(value, (draft: GaugeChartOptions) => {
+                      draft.max = newValue;
+                    })
+                  );
+                }}
+                placeholder="Default"
+              />
```
juliepagano, line 75: A few UX questions:
- Would it make sense to show the defaults in the text box for the cases where this isn't required?
- Should we have some UI to communicate when this is required (the not-percent case)?

These are both incremental improvements that could be done in a follow-up after consultation with design.

juliepagano (approved):
This looks good to me as an incremental improvement. A few questions about potential UX follow-ups inline.

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditorSettings.tsx
sjcobb: Yeah good idea! I updated the max placeholder so it shows the default value depending on unit.kind, see example here:

https://user-images.githubusercontent.com/9369625/210578538-e9193457-b17b-4591-a3b9-489742ea4c18.mp4

PR #895, opened 2022-12-23
[BREAKINGCHANGE] tooltip format and PrometheusTimeSeriesQuery series name overhaul

sjcobb:
## Overview
We got feedback that there are too many jarring transitions when moving your cursor on a TimeSeriesChart panel. One reason for this is how series name display changes when there are greater than 1 total series ('value' moves above and labels are shown on separate lines). This PR refactors tooltip formatting so if metric `__name__` is populated, it shows directly beside the formatted y_axis value. The yBuffer and yAxisInterval params used in getFocusedSeriesData are adjusted so more series are found  when your cursor is further away.

Breaking changes are made to PrometheusTimeSeriesQuery to make series name formatting easier to reason about and less one-off formatting has to happen inside TooltipContent, SeriesInfo components. Valid PromQL is also output when a custom series_name_format is not set (legend data benefits from this change too).

### Before 
Before_tooltip_value_label

image

### After
image
replace_value_with_metric_name
multi_series_tooltip_content


ui/components/src/Tooltip/SeriesInfo.tsx
```diff
@@ -109,20 +117,21 @@ export function SeriesInfo(props: SeriesInfoProps) {
...
+            minWidth: 150, // TODO: use clamp instead
             maxWidth: TOOLTIP_LABELS_MAX_WIDTH,
             overflow: 'hidden',
             textOverflow: 'ellipsis',
             whiteSpace: wrapLabels ? 'normal' : 'nowrap',
-            width: 'calc(100% - 20px)',
           })}
         >
-          {inlineSeriesLabels}
+          <Typography variant="body2">{inlineSeriesLabels}</Typography>
```
sjcobb, line 134: Noticed that fonts aren't quite right so I'm planning to fix this (I need to figure out what variant to use for the formattedY value), but the rest of this PR is ready for review.

ui/components/src/Tooltip/SeriesInfo.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+      y: 0.28,
+      formattedY: '28.0%',
+      markerColor: 'hsla(1016196180,50%,50%,0.8)',
+      totalSeries: 1,
+      wrapLabels: true,
+    };
+    renderComponent(seriesInfo);
+    expect(screen.getByText('node_load1:')).toBeInTheDocument();
+    expect(screen.getByText('28.0%')).toBeInTheDocument();
+    expect(screen.getByText('instance="demo.do.prometheus.io:')).toBeInTheDocument();
```
cndonovan, line 35: would `instance="demo.do.prometheus.io:9100"` not be in document?

ui/components/src/Tooltip/TooltipContent.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+          y: 0.1,
+          formattedY: '0.1',
+          markerColor: 'hsla(19838016,50%,50%,0.8)',
+        },
+      ],
+      wrapLabels: true,
+    };
+    renderComponent(tooltipContent);
+    expect(screen.getByText('Test node demo.do.prometheus.io:9100')).toBeInTheDocument();
+    expect(screen.getByText('Dec 23, 2022 -')).toBeInTheDocument();
```
cndonovan, line 41: skipping the time because we might have to deal with timezone differences?

cndonovan: should it check for the value too?

ui/components/src/Tooltip/TooltipContent.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+          x: 1671821040000,
+          y: 33771520,
+          formattedY: '33.77M',
+          markerColor: 'hsla(158479636,50%,50%,0.8)',
+        },
+      ],
+      wrapLabels: true,
+    };
+    renderComponent(tooltipContent);
+    expect(screen.getByText('33.77M')).toBeInTheDocument();
```
cndonovan, line 71: check for other y value too?

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
```diff
@@ -76,12 +75,11 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
...
       const { metric, values } = value;
 
-      // Name the series after the metric labels or if no metric, just use the
-      // overall query
-      let name = Object.entries(metric)
-        .map(([labelName, labelValue]) => `${labelName}="${labelValue}"`)
-        .join(', ');
-      if (name === '') name = query;
+      // Name the series after the metric labels or if no metric, use the query
+      let name = getUniqueKeyForPrometheusResult(metric, false);
```
cndonovan, line 79: You could maybe leave out `false` since `undefined` will also evaluate as false

ui/prometheus-plugin/src/utils/utils.ts
```diff
@@ -74,3 +74,53 @@ export function formatSeriesName(inputFormat: string, seriesLabels: SeriesLabels
...
+/*
+ * Stringifies object of labels into valid PromQL for querying metric by label
+ */
+function stringifyPrometheusMetricLabels(labels: { [key: string]: unknown }, removeExprWrap?: boolean) {
+  const labelStrings: string[] = [];
+  Object.keys(labels)
+    .sort()
+    .forEach((labelName) => {
+      const labelValue = labels[labelName];
+      if (typeof labelValue === 'string') {
```
cndonovan, line 87: Does this mean we'd skip values that are numbers?

ui/prometheus-plugin/src/utils/utils.test.ts
```diff
@@ -107,3 +113,44 @@ describe('formatSeriesName', () => {
...
+
+  it('should be a formatted prometheus string', () => {
+    labels = { ['foo']: 'bar' };
+    const result = getUniqueKeyForPrometheusResult(labels);
+    expect(result).toEqual('{foo="bar"}');
+  });
+
+  it('should be formatted with "', () => {
+    labels = { ['foo']: 'bar' };
+    const result = getUniqueKeyForPrometheusResult(labels, true);
```
cndonovan, line 131: Seeing it like this kind of makes me want named arguments, like `getUniqueKeyForPrometheusResult(labels, { removeExprWrap: true })`

cndonovan (approved):

ui/components/src/Tooltip/SeriesInfo.test.tsx
sjcobb: Whoops I pushed [a commit](https://github.com/perses/perses/pull/895/commits/f27c636f0) earlier that fixed this, but didn't update the test (at one point `9100` was being show separately)

ui/prometheus-plugin/src/utils/utils.test.ts
saminzadeh: Yeah I kinda like this better too. Not sure if we need more options, but it also opens the door to more in the future. No strong opinion though

ui/components/src/Tooltip/SeriesInfo.tsx
```diff
@@ -27,11 +27,17 @@ interface SeriesInfoProps {
 export function SeriesInfo(props: SeriesInfoProps) {
   const { seriesName, formattedY, markerColor, totalSeries, wrapLabels } = props;
 
-  // TODO (sjcobb): regex to remove __name__, improve series labels
-  const formattedSeriesLabels = seriesName.replace(/[{}"]/g, '');
+  // metric __name__ comes before opening curly brace, ignore if not populated
+  const nameSplit = seriesName.split('{');
```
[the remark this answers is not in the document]
sjcobb: Just added a comment with examples too!

ui/components/src/Tooltip/TooltipContent.test.tsx
sjcobb: Yeah good call, I added this and should be fine to check for time as well

ui/prometheus-plugin/src/utils/utils.ts
sjcobb: So that shouldn't ever happen with what Prom supports in metric label values, but changing this to check whether labelValue is undefined seems to work fine too so just updated this!

PR #900, opened 2023-01-03
[ENHANCEMENT] Move edit variables button and make variables wrap to next line

cndonovan:
## Changes
* Created the `EditVariablesButton` component, so that it's possible to move the button to a different location, as designs call for.  `EditVariablesButton` contains both the button itself and the drawer for editing variables. 

## Other changes
* When we have many variables, they will stack on top of each other nicely (and the margins will be correct)
* The buttons saying "Add panel" and "Add panel group" are shortened to "Panel" and "Panel group" to save space (the icon and tooltip still tell the user that these buttons are for adding things). 

Screen Shot 2023-01-03 at 1 21 05 PM
Screen Shot 2023-01-03 at 1 20 44 PM

### DIFF
--- a/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
+++ b/ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
@@ -15,13 +15,16 @@ import { Typography, Stack, Button, Box, useTheme, useMediaQuery, Alert } from '
 import PencilIcon from 'mdi-material-ui/PencilOutline';
 import AddPanelGroupIcon from 'mdi-material-ui/PlusBoxOutline';
 import AddPanelIcon from 'mdi-material-ui/ChartBoxPlusOutline';
-import { ErrorBoundary, ErrorAlert } from '@perses-dev/components';
+import { ErrorBoundary, ErrorAlert, InfoTooltip, TooltipPlacement } from '@perses-dev/components';
 import { DashboardResource } from '@perses-dev/core';
 import { useState } from 'react';
 import { useDashboard, useDashboardActions, useEditMode } from '../../context';
-import { TemplateVariableList } from '../Variables';
+import { TemplateVariableList, VariableEditButton } from '../Variables';
 import { TimeRangeControls } from '../TimeRangeControls';
 import { DownloadButton } from '../DownloadButton';
+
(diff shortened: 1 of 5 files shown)


ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -15,14 +15,17 @@ import { Typography, Stack, Button, Box, useTheme, useMediaQuery, Alert } from '
...
+import { ErrorBoundary, ErrorAlert, InfoTooltip, TooltipPlacement } from '@perses-dev/components';
 import { DashboardResource } from '@perses-dev/core';
 import { useState } from 'react';
 import { useDashboard, useDashboardActions, useEditMode } from '../../context';
-import { TemplateVariableList } from '../Variables';
+import { TemplateVariableList, EditVariablesButton } from '../Variables';
 import { TimeRangeControls } from '../TimeRangeControls';
 import { DownloadButton } from '../DownloadButton';
 
+const ADD_PANEL_BUTTON_DESCRIPTION = 'Add panel';
```
saminzadeh, line 26: Curious, what's the rational of making these constants instead of defining the values inline where it's used?

saminzadeh (approved):
LGTM - assuming we're good on the UX design review

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
cndonovan: it's cuz I was using the same copy for the tooltip AND the aria-label, but open to changing it back 

I have another PR coming related to all the copy for all the tooltips on all the buttons, I wonder what you'll think (I wonder if my upcoming PR seems like over-engineering or actually helpful)

cndonovan:
> LGTM - assuming we're good on the UX design review

Yes, Mohit/Karan and I sycned up right at the end of break, and I'm moving us toward that design

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
saminzadeh: Ah my bad, missed that!

PR #905, opened 2023-01-05
[IGNORE] Split playwright config by env

juliepagano:

Follow-up request from https://github.com/perses/perses/pull/890


ui/e2e/src/config/base.playwright.config.ts
```diff
@@ -31,16 +30,9 @@ const config: PlaywrightTestConfig = {
...
-  fullyParallel: true,
-  /* Fail the build on CI if you accidentally left test.only in the source code. */
-  forbidOnly: !!IS_CI,
-  /* Retry on CI only */
-  retries: IS_CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: IS_CI ? 1 : undefined,
+
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: IS_CI ? [['html']] : [['html'], ['list', { printSteps: true }]],
```
juliepagano, line 43: I think these being different was a result of testing a few things out in my previous PR, but there's not a good reason for it. Keeping them the same going forward.

ui/package.json
```diff
@@ -13,7 +13,8 @@
     "reinstall": "npm run clean && rm -rf **/node_modules && npm install",
     "clear-turbo-cache": "rimraf node_modules/.cache/turbo",
     "e2e": "turbo run e2e",
-    "e2e:debug": "turbo run e2e -- --debug"
+    "e2e:debug": "turbo run e2e -- --debug",
```
cndonovan, line 16: Really not married to this idea, but wonder if we want to be able to write this as `e2e: turbo run e2e:local`? So it's easy to tell that one command is running local configs and one command is running CI configs? 

Still have the convenience of just `npm e2e`, but when peering at the details in package.json it might be more obvious what the difference is?

Again, I think it's fine as is, and this might be a strange idea, b/c I don't do as much test config work.

cndonovan (approved):
It really does get super clean having it split out into two files ⭐

ui/e2e/src/config/base.playwright.config.ts
```diff
@@ -31,16 +30,9 @@ const config: PlaywrightTestConfig = {
      */
     timeout: 5000,
   },
-  /* Run tests in files in parallel */
-  fullyParallel: true,
```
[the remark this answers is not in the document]
juliepagano: Good catch. I think that one accidentally got lost when moving things around. Adding it back.

ui/package.json
juliepagano: I think that's a good idea to increase clarity! Will fix it up.

PR #907, opened 2023-01-05
[ENHANCEMENT] Add prop for adjusting the height of the DownloadButton

cndonovan:

Similar to the `TimeRangeControls`, we want to be able to adjust the height of the `DownloadButton`.

This change will enable us to use the `DownloadButton` from the perses package, instead of using the local version of the `DownloadButton` in our app.


juliepagano (approved):
LGTM! 🚀

sjcobb (approved):
💯

Issue #806, opened 2022-11-22
Getting datasources is spamming the backend


Nexucis:
Currently, every panels and variables are requesting the datasources, which ended up by making an HTTP request to the backend.

Perhaps the `GetDatasources` in the app is not good and missed a cache usage through `react-query`. 
It would be great to fix this issue, because when you are loading a big dashboard like the `node-exporter`, it makes the load of the dashboard really slow comparing to the number of info displayed.

PR #908, opened 2023-01-06
[WIP] attempt to fix datasource spam issue

sjcobb:
See Issue #806, this branch has a potential quick fix but it's hacky and I'm not 100% sure it works / doesn't break other functionality. 

I also noticed that `useProjectQuery` seems to run too frequently, which also may play a factor but I didn't dig deeper on that hunch.

### DIFF
--- a/ui/dashboards/src/context/DatasourceStoreProvider.tsx
+++ b/ui/dashboards/src/context/DatasourceStoreProvider.tsx
@@ -97,15 +97,17 @@ export function DatasourceStoreProvider(props: DatasourceStoreProviderProps) {
   // Given a Datasource selector, finds the spec for it and then uses its corresponding plugin the create a client
   const getDatasourceClient = useCallback(
     async function getClient<Client>(selector: DatasourceSelector): Promise<Client> {
-      if (activeDatasourceClient !== undefined) {
-        return activeDatasourceClient;
+      if (activeDatasourceClient && activeDatasourceClient.client) {
+        if (activeDatasourceClient.name === selector.name) {
+          return activeDatasourceClient.client;
+        }
       }
       const { kind } = selector;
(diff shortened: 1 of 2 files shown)


cndonovan:
@sjcobb Is this a PR we could close for the time being? (OK if the answer is no, it helps to have it open)

sjcobb:
> Is this a PR we could close for the time being?

I might keep it open a little longer for some questions I meant to ask but forgot about. This was a quick hack that I meant to follow up with @Nexucis on. The approach here isn't right, I think the proper way to fix the issue with too many redundant datasource requests (#806) is to change the queryKey in time-series-queries.ts

I'm a little nervous to do this though, do you happen to know why the entire datasourceStore is [passed in the queryKey here](https://github.com/perses/perses/blob/v0.24.0/ui/plugin-system/src/runtime/time-series-queries.ts#L61)? (@Nexucis @cndonovan @saminzadeh) It seems like it'd be better to be more specific.

Nexucis:
> I'm a little nervous to do this though, do you happen to know why the entire datasourceStore is [passed in the queryKey here](https://github.com/perses/perses/blob/v0.24.0/ui/plugin-system/src/runtime/time-series-queries.ts#L61)?

Ah no I don't know on my side :(. Perhaps @LukeTillman will have a better idea here.

>I might keep it open a little longer for some questions I meant to ask but forgot about. This was a quick hack that I meant to follow up with @Nexucis on. The approach here isn't right, I think the proper way to fix the issue with too many redundant datasource requests (https://github.com/perses/perses/issues/806) is to change the queryKey in time-series-queries.ts

Sorry @sjcobb I forgot about this PR. I will try it now :)

sjcobb:
Didn't realize the datasource requests aren't actually using react-query, I think the next step should be to call useQuery or useQueries for those too instead of using fetch directly.

PR #909, opened 2023-01-06
[ENHANCEMENT] Toolbar uses two rows for small screens

cndonovan:

With this PR: 
* When the screen size is tablet or bigger, the toolbar buttons will take up one row (existing behavior). 
* When the screen size is smaller, the toolbar buttons will be split into two rows to save space

## Tablet or bigger screens
Screen Shot 2023-01-06 at 11 08 38 AM

## Small screens
Screen Shot 2023-01-06 at 11 08 10 AM

### DIFF
--- a/ui/dashboards/src/components/AddGroupButton/AddGroupButton.tsx
+++ b/ui/dashboards/src/components/AddGroupButton/AddGroupButton.tsx
@@ -11,18 +11,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export * from './AddGroupButton';
-export * from './AddPanelButton';
-export * from './Dashboard';
-export * from './DashboardToolbar';
-export * from './DeletePanelDialog';
-export * from './DeletePanelGroupDialog';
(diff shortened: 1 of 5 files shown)


juliepagano (approved):
Code change looks good to me. Out of curiosity, is there any push behind this change related to supporting smaller screens or is this just a one-off nice-to-have?

Nexucis:
mmm I thought we said it won't be possible to edit things on mobile phone (small screen). So we should be able to gain space here.

> Out of curiosity, is there any push behind this change related to supporting smaller screens or is this just a one-off nice-to-have?

Yes I'm a pushing it for that. It is something I think that will be useful when you have an incident and a link to the dashboard is linked. A high manager join the conversation from a phone and see the dashboard directly in its phone. We should have something decent to display.

Nexucis:
> This is definitely an edge case (I don't think users do this often), but at the least, it doesn't hurt to assure that the layout looks good at phone-sized screen sizes.

Totally inline with that. Thank you @cndonovan !!

PR #911, opened 2023-01-06
[BUGFIX] reduce calculated min step interval in TimeSeriesPanel

sjcobb:
Multiple queries crash bug - try setting lowest possible min step interval to 10 ms (this is what Grafana uses in their rounding function). Not sure this will actually fix the issue we're seeing with the page crashing after new query is Applied, but maybe it will help somewhat until we can refactor the gcd approach.

Future TODO: make “Min Interval” configurable from query editor fields


eunicorn (approved):
thanks :)

Issue #838, opened 2022-12-05
Add functionality to assist with generating changelog


juliepagano:
Creating this based on a discussion in the meeting this morning. The current process for manually updating the changelog creates a bit of toil around things like merge conflicts and forgetting to update the changelog for PRs. It would be helpful to have some tooling to assist with generating the changelog when creating a release based on commits and pull requests. 

Below is a summary of my understanding of the problem and some notes from our discussions.

## Script to autogenerate changelog

**Output:**
- A script that a maintainer can run when creating a release to assist with generating the changelog.

**Notes from discussion:**
- We should not rely solely on the autogenerated changelog. It should generate a rough draft that is then reviewed and adjusted (as needed) by the person making the release.
- We likely need to improve our approach to PRs and merging to `main` to make it more feasible to generate the changelog. See next section for notes on this.
- It would be helpful if there was a prefix people can put on commits that tells the changelog to skip them. 
- We do not care about keeping an "unreleased" section of our changelog. It exists now to help generate the changelog, but can go away when we add the script.

**Ideas:**
- Looking at some popular open source projects, many of them have written their own scripts for this, usually with it focused on parsing git commits. We *might* be able to leverage an open source tool, but many of them are really focused on detailed commit semantics or aren't aligned with our approach. `auto-changelog` might work, but looks like it would need a lot of customization -- it might be easier just to write our own script.
- If we require people start using the prefixes we want (`[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`, `[BREAKINGCHANGE]`) in the changelog in commit titles, that would make it relatively easy to parse commits.
- Could suggest a prefix like `[NOCHANGELOG]` or `[IGNORE]` or something like that for the case where we know a commit doesn't need to be in the changelog.

## Improvements to commits merged to `main`

For automation to be helpful, the commits that make it to `main` should ideally be organized and named in a way that aligns with the changelog. This would also be helpful for anyone working in the repo looking through the [commits]( https://github.com/perses/perses/commits/main) and trying to make sense of them. 

**Output:**
- Guidelines about title naming conventions and how they translate to the changelog.
- Guidelines about merging PRs to `main`.

**Notes from discussion:**
- We should not require a full squash and merge because sometimes a pull request contains multiple entries for a changelog. 
- On the flip side, we should not merge ALL THE COMMITS in a PR, which can include a lot of noise (things like "responded to PR feedback", "fixed linter error"). 

**Ideas:**
[...]

PR #914, opened 2023-01-09
[IGNORE] Add PR template

juliepagano:
And clarify use of the `IGNORE` catalog entry in the docs

This will finish up the last remaining task for https://github.com/perses/perses/issues/838 so we can close it.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+<!--
+  See the contributing guide for detailed guidance about contributing.
+  https://github.com/perses/perses/blob/main/CONTRIBUTING.md
+ -->
+
+# Checklist
+
+- [ ] Pull request has a descriptive title and context useful to a reviewer.
+- [ ] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
+- [ ] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
+- [ ] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,10 +25,11 @@ To make it works correctly, commit or PR's title should follow the following nam
 
 where `catalog_entry` can be :
 
-- FEATURE
-- ENHANCEMENT
-- BUGFIX
-- BREAKINGCHANGE
+- `FEATURE`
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+<!--
+  See the contributing guide for detailed guidance about contributing.
+  https://github.com/perses/perses/blob/main/CONTRIBUTING.md
+ -->
+
+# Checklist
+
+- [ ] Pull request has a descriptive title and context useful to a reviewer.
+- [ ] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
+- [ ] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
```
juliepagano, line 10: Not sure if we really need this here since there's a github check, but thought it might be helpful as a pointer for people who are new and may not be sure what that github check is doing.

.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+<!--
+  See the contributing guide for detailed guidance about contributing.
+  https://github.com/perses/perses/blob/main/CONTRIBUTING.md
+ -->
+
+# Checklist
```
juliepagano, line 6: Anything else I'm missing here? I skimmed the contributing guide and didn't see a ton else to add. 

We may want to think about putting together some additional guidance on commit messages and other things as we bring on more contributors, but that work seems separate from this initial addition of a PR template.

CONTRIBUTING.md
```diff
@@ -25,10 +25,11 @@ To make it works correctly, commit or PR's title should follow the following nam
...
 
-- FEATURE
-- ENHANCEMENT
-- BUGFIX
-- BREAKINGCHANGE
+- `FEATURE`
+- `ENHANCEMENT`
+- `BUGFIX`
+- `BREAKINGCHANGE`
+- `IGNORE` - Changes that should not generate entries in the changelog. Primarily used for internal tooling changes that do not impact consumers.
```
juliepagano, line 32: Only added an explanation for ignore because the others are relatively straightforward.

.github/pull_request_template.md
cndonovan: I think this is a good call -- it is one of the main things to remember and if I was a new contributor, I would much rather see this first than run into the DCO warning (which can be harder to debug)

cndonovan (approved):
I like the short-and-sweetness of this one! 

I think our current proprietary one is too long, my eyes glaze over, and never read it.

PR #915, opened 2023-01-09
[ENHANCEMENT] Remove variable list background color

cndonovan:
Two style tweaks:

* AppBar has a background color that shows up in dark mode. It's subtle, but different from the other background color (see screenshot).
* Added a negative margin to make sure the variables don't have unnecessary padding (I added the margins in previous PR so that the variables will have margins when they wrap to next line). 

## Background color (see the division between variables and buttons)
Screen Shot 2023-01-09 at 3 17 45 PM


juliepagano (approved):
Good catch! 🚀

saminzadeh (approved):

PR #922, opened 2023-01-13
[BREAKINGCHANGE] Add a consistent JSON tab for every panel plugin

eunicorn:
Currently, some dashboard panels (e.g. time series, gauge) include a "JSON" tab where the user can edit raw JSON for the panel configuration. Other panels (e.g. markdown) do not.

We should make this behavior consistent, meaning all panels should have a consistent JSON tab. This should include both the plugins defined in perses and other custom panel plugins.

PanelPlugin will now take in two additional fields: 

1. `PanelQueryEditorComponent`: this component will be displayed under the "Query" tab
2. `panelOptionsEditorComponents`: any other custom options editors (e.g. settings, markdown)

**Before**
image

**After**
image

### DIFF
--- a/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
+++ b/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
@@ -11,16 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { JSONEditor } from '@perses-dev/components';
-import {
-  OptionsEditorProps,
-  OptionsEditorTabs,
-  TimeSeriesQueryEditor,
-  TimeSeriesQueryEditorProps,
-} from '@perses-dev/plugin-system';
+import { OptionsEditorProps, TimeSeriesQueryEditor, TimeSeriesQueryEditorProps } from '@perses-dev/plugin-system';
 import { produce } from 'immer';
 import { GaugeChartOptions } from './gauge-chart-model';
-import { GaugeChartOptionsEditorSettings } from './GaugeChartOptionsEditorSettings';
 
 export type GaugeChartOptionsEditorProps = OptionsEditorProps<GaugeChartOptions>;
 
@@ -39,19 +32,5 @@ export function GaugeChartOptionsEditor(props: GaugeChartOptionsEditorProps) {
     );
   };
 
-  return (
-    <OptionsEditorTabs
-      tabs={{
-        query: {
-          content: <TimeSeriesQueryEditor value={query} onChange={handleQueryChange} />,
-        },
-        settings: {
-          content: <GaugeChartOptionsEditorSettings {...props} />,
-        },
-        json: {
-          content: <JSONEditor {...props} />,
-        },
-      }}
-    />
-  );
+  return <TimeSeriesQueryEditor value={query} onChange={handleQueryChange} />;
 }
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
@@ -14,6 +14,7 @@
 import { PanelPlugin } from '@perses-dev/plugin-system';
 import { createInitialTimeSeriesChartOptions, TimeSeriesChartOptions } from './time-series-chart-model';
 import { TimeSeriesChartOptionsEditor } from './TimeSeriesChartOptionsEditor';
+import { TimeSeriesChartOptionsEditorSettings } from './TimeSeriesChartOptionsEditorSettings';
 import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
 
 /**
@@ -21,6 +22,7 @@ import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
  */
 export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
   PanelComponent: TimeSeriesChartPanel,
-  OptionsEditorComponent: TimeSeriesChartOptionsEditor,
+  PanelQueryEditorComponent: TimeSeriesChartOptionsEditor,
+  panelOptionsEditorComponents: [{ id: 'settings', label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
   createInitialOptions: createInitialTimeSeriesChartOptions,
 };
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
@@ -13,7 +13,7 @@
 
 import { PanelPlugin } from '@perses-dev/plugin-system';
 import { createInitialTimeSeriesChartOptions, TimeSeriesChartOptions } from './time-series-chart-model';
-import { TimeSeriesChartOptionsEditor } from './TimeSeriesChartOptionsEditor';
+import { TimeSeriesChartQueryEditor } from './TimeSeriesChartQueryEditor';
 import { TimeSeriesChartOptionsEditorSettings } from './TimeSeriesChartOptionsEditorSettings';
 import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
 
@@ -22,7 +22,7 @@ import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
  */
 export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
   PanelComponent: TimeSeriesChartPanel,
-  PanelQueryEditorComponent: TimeSeriesChartOptionsEditor,
+  PanelQueryEditorComponent: TimeSeriesChartQueryEditor,
   panelOptionsEditorComponents: [{ id: 'settings', label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
   createInitialOptions: createInitialTimeSeriesChartOptions,
 };
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
@@ -23,6 +23,6 @@ import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
 export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
   PanelComponent: TimeSeriesChartPanel,
   PanelQueryEditorComponent: TimeSeriesChartQueryEditor,
-  panelOptionsEditorComponents: [{ id: 'settings', label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
+  panelOptionsEditorComponents: [{ label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
   createInitialOptions: createInitialTimeSeriesChartOptions,
 };
--- a/ui/plugin-system/src/components/OptionsEditorTabs/OptionsEditorTabs.test.tsx
+++ b/ui/plugin-system/src/components/OptionsEditorTabs/OptionsEditorTabs.test.tsx
@@ -19,17 +19,14 @@ import { OptionsEditorTabs, OptionsEditorTabsProps } from './OptionsEditorTabs';
 describe('OptionsEditorTabs', () => {
   const mockTabs: OptionsEditorTabsProps['tabs'] = [
     {
-      id: 'query',
       label: 'Query',
       content: <div>Edit query configuration</div>,
     },
     {
-      id: 'settings',
       label: 'Settings',
       content: <div>Edit settings configuration</div>,
     },
     {
-      id: 'json',
       label: 'JSON',
       content: <div>JSON editor</div>,
     },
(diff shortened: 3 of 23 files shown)


ui/plugin-system/src/model/plugin-base.ts
```diff
@@ -20,7 +20,7 @@ export interface Plugin<Spec> {
   /**
    * React component for editing the plugin's options in the UI.
    */
-  OptionsEditorComponent: React.ComponentType<OptionsEditorProps<Spec>>;
+  OptionsEditorComponent?: React.ComponentType<OptionsEditorProps<Spec>>;
```
eunicorn, line 23: should this still be part of the base type for all plugin implementations if panel plugin doesn't use this?

ui/plugin-system/src/components/OptionsEditorTabs/OptionsEditorTabs.tsx
```diff
@@ -15,88 +15,31 @@ import { Tab, TabProps, Tabs, TabsProps, Box } from '@mui/material';
...
-] as const;
-
 export const OptionsEditorTabs = ({ tabs }: OptionsEditorTabsProps) => {
   const [activeTab, setActiveTab] = useState(0);
 
   const handleChange: TabsProps['onChange'] = (_, newValue) => {
     setActiveTab(newValue);
   };
 
-  // Normalize the common tabs that are managed via constants in this file
```
eunicorn, line 70: This logic is now handled in PluginSpecEditor.tsx

ui/plugin-system/src/components/OptionsEditorTabs/OptionsEditorTabs.test.tsx
```diff
@@ -103,59 +91,14 @@ describe('OptionsEditorTabs', () => {
...
-    expect(tabs).toHaveLength(5);
+    expect(tabs).toHaveLength(3);
     expect(tabs[0]).toHaveTextContent('Query');
     expect(tabs[1]).toHaveTextContent('Settings');
-    expect(tabs[2]).toHaveTextContent('Table column');
-    expect(tabs[3]).toHaveTextContent('Table options');
-    expect(tabs[4]).toHaveTextContent('JSON');
-  });
-
-  it('shows the correct content when selecting a custom tab', () => {
```
eunicorn, line 119: these unit tests are not needed anymore now that the component doesn't have different types of tabs

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
```diff
@@ -39,19 +32,5 @@ export function GaugeChartOptionsEditor(props: GaugeChartOptionsEditorProps) {
...
-        settings: {
-          content: <GaugeChartOptionsEditorSettings {...props} />,
-        },
-        json: {
-          content: <JSONEditor {...props} />,
-        },
-      }}
-    />
-  );
+  return <TimeSeriesQueryEditor value={query} onChange={handleQueryChange} />;
```
sjcobb, line 35: The naming of these components is kind of confusing now, for instance this is still called GaugeChartOptionsEditor but it only returns the query editor

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
```diff
@@ -14,13 +14,15 @@
...
 import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
 
 /**
  * The core TimeSeriesChart panel plugin for Perses.
  */
 export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
   PanelComponent: TimeSeriesChartPanel,
-  OptionsEditorComponent: TimeSeriesChartOptionsEditor,
+  PanelQueryEditorComponent: TimeSeriesChartOptionsEditor,
+  panelOptionsEditorComponents: [{ id: 'settings', label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
```
sjcobb, line 26: Should panelOptionsEditorComponents be capitalized like the others and if it is an array, what is an example of when you’d need to add more than one component inside the array?

eunicorn: > Should panelOptionsEditorComponents be capitalized like the others

I didn't capitalize it because panelOptionsEditorComponents is an array whereas the others indicate that it should be React.ComponentType.

> what is an example of when you’d need to add more than one component inside the array?

I don't have a specific example but this gives the flexibility to plugin authors if they ever want to have multiple tabs

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditor.tsx
eunicorn: renamed these components to xxxChartQueryEditor instead

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
juliepagano: We haven't built it yet, but I remember there being a hypothetical example for a table plugin that might split general settings and column settings into two tabs in a case like this.

saminzadeh:
I like the approach you took here. Makes sense with the two options now in the panel that are both optional so at the very least you can always edit the panel JSON.

ui/plugin-system/src/model/panels.ts
```diff
@@ -12,13 +12,26 @@
...
 export interface PanelPlugin<Spec = UnknownSpec> extends Plugin<Spec> {
   PanelComponent: React.ComponentType<PanelProps<Spec>>;
+  /**
+   * React component for the content of "Query" tab
+   */
+  PanelQueryEditorComponent?: React.ComponentType<OptionsEditorProps<Spec>>;
+  /**
+   * React components for custom tabs
+   */
+  panelOptionsEditorComponents?: Array<PanelOptionsEditorComponent<Spec>>;
```
cndonovan, line 34: Ok zooming out to bit to think about the schema changes...

So, the requirement we know we want to implement is that every plugin that is a panel type shows a `JSON` tab (PS I think the tabs look so good)

Currently, we require `OptionsEditorComponent` for every plugin (not just PanelPlugins).

If I'm following, it seems like we moved the "awareness" about the tabs out of the `OptionsEditorComponent` components, and up into the `PluginSpecEditor` so that we'd be able to always append the JSON tab to the end of the list of tabs.

And that makes sense to me -- we need to be able to append that JSON tab inside the framework code, instead of inside of the developer/custom code. 

So, I'm thinking... what if the only change we made to schema was that `Plugin` takes `optionsEditorComponents` instead of `OptionEditorComponent`

In this case, all plugins now can accept an array of editor components, instead of just one. So, we would have the same API for all plugins. 

Then, for each of the panel plugins we currently have (stat, time-series, gauge, markdown), we could implement both the SomePanelQueryEditorComponent, SomePanelOptionsEditorComponent, and we could include these in the array that we feed into `optionsEditorComponents`. 

In this case, we're not yet adding an expectation that every panel plugin has to have a query editor... I could be totally wrong, but it seems like it could be nice to leave that open, since we came to this wanting to make sure to always have a JSON tab, but not necessarily always have a Query tab. 

So overall change would be:

```
interface Plugin<Spec> {
  optionsEditorComponents: Array<React.ComponentType<OptionsEditorProps<Spec>> (or whatever TS works)
}
```

cndonovan: Open to thoughts...

The main things making me think about this approach: 
- hm, wish we didn't have to have a different API for editing panel plugins vs. all other plugins
- do we definitely want to make PanelQueryEditorComponent a dedicated/named field in the PanelPlugin schema?

ui/plugin-system/src/model/plugin-base.ts
eunicorn: yeah I'm thinking we can just add `OptionsEditorComponent` to  these types `VariablePlugin` and `TimeSeriesQueryPlugin`  instead having it in the base `Plugin` type since not every plugin will require an options editor component

ui/plugin-system/src/model/panels.ts
eunicorn: > what if the only change we made to schema was that Plugin takes optionsEditorComponents instead of OptionEditorComponent

The only concern is that only panel plugin needs to be aware of `tabs`. Having tabs require panel plugin authors to also pass in a label and content for each option editor component. 

> wish we didn't have to have a different API for editing panel plugins vs. all other plugins

i think this is ok since each plugin type is very different. For example, `PrometheusDatasource` currently does not even have an option editor component 

> do we definitely want to make PanelQueryEditorComponent a dedicated/named field in the PanelPlugin schema?

I added this to enforce consistent ui/ux since having a query editor is very common for a panel plugin. `PanelQueryEditorComponent` will always be displayed under the `Query` tab, and `Query` tab will always be displayed first before other custom tabs.

ui/plugin-system/src/components/OptionsEditorTabs/OptionsEditorTabs.tsx
```diff
@@ -15,88 +15,31 @@ import { Tab, TabProps, Tabs, TabsProps, Box } from '@mui/material';
 import { useState } from 'react';
 import { TabPanel } from './TabPanel';
 
-interface BaseTabConfig {
+export type OptionsEditorTab = {
+  id: string;
```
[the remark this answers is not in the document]
eunicorn: good call, i'll remove `id` and use `label` as the key

ui/plugin-system/src/model/panels.ts
eunicorn: but yeah these are all valid points, we can discuss this with rest of the team at the next meeting!

cndonovan: Nah, that makes sense to me -- I wasn't thinking about the tabs, and how we don't use tabs for other types of plugins (queries, datasource)

Yeah you're right, it's nice to be setting up the panel plugins for success -- many panels do have queries, guarantee that the query is always displayed in Query tab and Query tab always goes first, UI will look good ✨ 

So yeah my mind is changed, I think doing special stuff for the panel plugins is good / enabling us to be  more successful at building an awesome UI -- panel plugins have the most complex editing needs, and we want the editing experience that helps tame that complexity (nicely organized and consistent, which this does). I'm in ✅

saminzadeh: So there is another way of thinking about this (not necessarily for this PR):

Don't have a `PanelQueryEditorComponent` at all. Instead, the Panel plugin could specify which DataResponse formats it supports and that way we can automatically render the correct Query components. This would also help us de-couple query editing and the panel options (similar to the approach Grafana takes).
 
Grafana assumes every panel will also receive a DataFrame but gives you the ability to disable query editing with a declarative config option `skipDataQuery` if you don't use it. For example, see [Markdown panel config](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/text/plugin.json#L6) for example

Our interface might look something like this:

```ts
  supportedQueryInterfaces?: Array<'TimeSeriesQuery' | 'LogQuery' | 'TraceQuery'>;
}
```

So Plugins would look like this:

```ts
// No editor will show `supportedQueryInterfaces` is not present
const markdownPanel: PanelPlugin<MarkdownPanelOptions> = {
  PanelComponent: MarkdownPanel,
  panelOptionsEditorComponents: [{ label: 'Markdown', component: MarkdownOptionsEditor}],
}

// Only editors that support TimeSeriesQueries
const timeSeriesPanel: PanelPlugin<TimeSeriesPanelOptions> = {
  PanelComponent: TimeSeriesPanel,
  supportedQueryInterfaces: ['TimeSeriesQuery'],
  panelOptionsEditorComponents: [{ label: 'Time series', component: TimeSeriesOptionsEditor}],
}

// Only editors that support LogQueries
const logPanel: PanelPlugin<LogPanelOptions> = {
  PanelComponent: LogPanel,
  supportedQueryInterfaces: ['LogQuery'],
  panelOptionsEditorComponents: [{ label: 'Logs', component: LogOptionsEditor}],
}

// Will show editors that support 'TimeSeriesQuery', 'LogQuery', 'TraceQuery'
const tablePanel: PanelPlugin<TablePanelOptions> = {
  PanelComponent: TablePanel,
  supportedQueryInterfaces: ['TimeSeriesQuery', 'LogQuery', 'TraceQuery'],
  panelOptionsEditorComponents: [{ label: 'Table', component: TableOptionsEditor}],
}
```

cndonovan:
Accidentally closed this when running a script to remove my local copy of the branch -- googling if there are ways to go back.

cndonovan:
Eunice was able to restore, now reopening!

juliepagano:
Is this ready for another pass now, @eunicorn, or are there some outstanding changes you want to make first?

eunicorn:
> Is this ready for another pass now, @eunicorn, or are there some outstanding changes you want to make first?

Yes ready for review again!

ui/plugin-system/src/model/panels.ts
juliepagano: Are you thinking you'll pick up with the additional changes in a fast follow PR or that it's something we'll come back to farther in the future? If the latter, can these additional thoughts get captured in a github discussion or something else that will be a little more durable than a PR after this gets merged?

juliepagano (approved):
Looks good to me assuming the follow-ups will either be documented somewhere more durable when this is merged and/or they're being handled soon after! 🚀

ui/plugin-system/src/model/panels.ts
eunicorn: i think i'll pick up the changes in a fast follow PR

sjcobb (approved):
💯

saminzadeh (approved):

PR #923, opened 2023-01-16
[BUGFIX] Fix API error EOF that happens when decoding multiple times the body from a middleware

Nexucis:
When calling the endpoint `/api/v1/dashboards`, it will trigger a middleware that is decoding the body to find the what project shall be used to perform the request.

It appears it's not possible to decode multiple times the body when we are using the method `bind` from the framework `echo`.

This PR is fixing that, assuring we don't modify the request body so it can be decoded later by the business code.

### DIFF
--- a/internal/api/core/middleware/verification.go
+++ b/internal/api/core/middleware/verification.go
@@ -14,8 +14,11 @@
 package middleware
 
 import (
+	"bytes"
+	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


Nexucis:
@juliepagano this PR should help to fix your current issue with the e2e tests.

AntoineThebaud (approved):

PR #929, opened 2023-01-18
[BUGFIX] Fix discard changes confirmation dialog showing up even when there is no change

eunicorn:
**BEFORE**
The discard changes confirmation dialog appears even when you don't make any changes 
BEFORE

**AFTER** 
Jan-18-2023 18-15-25

### DIFF
--- a/ui/e2e/src/pages/DashboardPage.ts
+++ b/ui/e2e/src/pages/DashboardPage.ts
@@ -51,7 +51,6 @@ export class DashboardPage {
 
   readonly panelEditor: Locator;
   readonly variableEditor: Locator;
-  readonly discardChangesConfirmationDialog: Locator;
 
   readonly alert: Locator;
 
@@ -81,7 +80,6 @@ export class DashboardPage {
 
     this.panelEditor = page.getByTestId('panel-editor');
     this.variableEditor = page.getByTestId('variable-editor');
-    this.discardChangesConfirmationDialog = page.getByTestId('discard-changes-confirmation-dialog');
 
     this.alert = page.getByRole('alert');
   }
@@ -101,6 +99,14 @@ export class DashboardPage {
     return this.page.getByRole('dialog', {
       name: name,
     });
+  }
+
+  /**
+   * PANEL EDITOR HELPERS
+   */
+
+  getPanelEditor() {
+    return new PanelEditor(this.panelEditor);
   }
 
   /**
--- a/ui/e2e/src/pages/DashboardPage.ts
+++ b/ui/e2e/src/pages/DashboardPage.ts
@@ -51,6 +51,7 @@ export class DashboardPage {
 
   readonly panelEditor: Locator;
   readonly variableEditor: Locator;
+  readonly discardChangesConfirmationDialog: Locator;
 
   readonly alert: Locator;
 
@@ -80,6 +81,7 @@ export class DashboardPage {
 
     this.panelEditor = page.getByTestId('panel-editor');
     this.variableEditor = page.getByTestId('variable-editor');
+    this.discardChangesConfirmationDialog = page.getByTestId('discard-changes-confirmation-dialog');
 
     this.alert = page.getByRole('alert');
   }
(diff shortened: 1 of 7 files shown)


ui/e2e/src/pages/DashboardPage.ts
```diff
@@ -80,6 +81,7 @@ export class DashboardPage {
 
     this.panelEditor = page.getByTestId('panel-editor');
     this.variableEditor = page.getByTestId('variable-editor');
+    this.discardChangesConfirmationDialog = page.getByTestId('discard-changes-confirmation-dialog');
```
juliepagano, line 84: You can usually get away with just looking for a dialog with the `getDialog` helper in this file and passing in the name to lean more on looking things up accessibly. Is it possible to do that for this test? If you ran into issues, the `data-testid` is an ok fallback.

ui/e2e/src/tests/panelEditor.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+test.use({
+  dashboardName: 'Panels',
+});
+
+test.describe('Dashboard: Panel Editor', () => {
+  test('should show discard changes confirmation dialog', async ({ dashboardPage }) => {
+    await dashboardPage.startEditing();
+    await dashboardPage.addPanel();
+    await dashboardPage.panelEditor.isVisible();
+    const panelEditor = new PanelEditor(dashboardPage.panelEditor);
```
juliepagano, line 26: Could you use the `isVisible` on the panel editor class here. It's a bit smarter about waiting for all the animations to finish on the panel sliding in.

Related, it might be nice to create a `getPanelEditor` helper in `DashboardPage` that simplifies the `new PanelEditor` bit for you. I added a similar helper like that for `getVariableEditor`, but didn't have a chance to go back and refactor a similar change for the panel editor. If you don't feel like doing this, no worries, and I'll come back to it at some point.

ui/e2e/src/tests/panelEditor.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    // clicking "Cancel" should do nothing and keep current changes
+    await dashboardPage.discardChangesConfirmationDialog.getByRole('button', { name: 'Cancel' }).click();
+    await expect(dashboardPage.panelEditor).toBeVisible();
+    await expect(dashboardPage.panelEditor.getByLabel(/^Type/)).toContainText('Markdown');
+
+    // clicking "Discard Changes" should discard changes
+    await panelEditor.cancelButton.click();
+    await dashboardPage.discardChangesConfirmationDialog.getByRole('button', { name: 'Discard Changes' }).click();
+    await expect(dashboardPage.panelEditor).toBeHidden();
```
juliepagano, line 40: Could you use the `panelEditor.isClosed` helper here. Similar to the `isVisible` helper I added, it's smart about waiting for the panel animations to finish up.

ui/e2e/src/tests/panelEditor.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { test, expect } from '../fixtures/dashboardTest';
+import { PanelEditor } from '../pages/PanelEditor';
+
+test.use({
+  dashboardName: 'Panels',
+});
+
+test.describe('Dashboard: Panel Editor', () => {
+  test('should show discard changes confirmation dialog', async ({ dashboardPage }) => {
```
juliepagano, line 22: Could you clarify the use case a little bit. Maybe something like "requires confirmation to discard changes", so the full test reads "Panel editor requires confirmation to discard changes."

Related: could you also add a test case for the reason you're making this PR. Maybe something like "can be closed without confirmation if nothing has changed"?

juliepagano (approved):
A handful of mostly optional feedback on the tests. Otherwise, the code changes look good to me. Thanks for working through adding tests!

eunicorn:
> A handful of mostly optional feedback on the tests. Otherwise, the code changes look good to me. Thanks for working through adding tests!

@juliepagano all your helper functions are so useful! 😄

ui/e2e/src/pages/DashboardPage.ts
eunicorn: oh this makes a lot of sense since there should only be one dialog appearing at a time :)

cndonovan (approved):
lgtm!

PR #930, opened 2023-01-18
[IGNORE] Fix npm e2e commands

eunicorn:


juliepagano (approved):
Thanks for catching this and fixing it up! 🚀

PR #938, opened 2023-01-25
[IGNORE] Add example visual tests for chart panels

juliepagano:
This is not adding comprehensive visual tests for panels. Instead, it is
adding some examples and test helpers to smooth over the process, so it
is easier for other contributors to add more visual tests in follow-ups.

# Notes for reviewers

- Click the "details" link next to the Happo check in the status checks section of the PR to see the output of the visual tests in Happo. Let me know if you need an invite to the account.
- This code to be reviewed isn't quite as big as the diff makes it look. A good chunk of it is adding more test dashboard json.

# Checklist


ui/e2e/package.json
```diff
@@ -14,10 +14,11 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "devDependencies": {
+    "@perses-dev/prometheus-plugin": "0.21.1",
     "@playwright/test": "^1.29.1",
     "chokidar-cli": "^3.0.0",
     "eslint-plugin-playwright": "^0.11.2",
-    "happo-e2e": "^2.1.0",
+    "happo-e2e": "^2.1.1",
```
juliepagano, line 21: We need this new version because there was a bug in the canvas handling. In positive news, they had it fixed within about an hour of me emailing support, so it's really nice to see how responsive they are to us, even in a "free" use case.

ui/e2e/package.json
```diff
@@ -14,10 +14,11 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "devDependencies": {
+    "@perses-dev/prometheus-plugin": "0.21.1",
```
juliepagano, line 17: Needed to grab some types for the data mocking.

ui/e2e/src/config/base.playwright.config.ts
```diff
@@ -45,6 +45,22 @@ const config: PlaywrightTestConfig = {
...
+     * Disabled per playwright documentation, so we can reliably mock network
+     * requests.
+     * https://playwright.dev/docs/network#missing-network-events-and-service-workers
+     */
+    serviceWorkers: 'block',
+
+    // Keep this in sync with the setting in `.happo.js` for consistency for
+    // canvas elements, which are converted into images when run in playwright
+    // and sent to happo.
+    viewport: { width: 1200, height: 800 },
```
juliepagano, line 59: I was debating putting this out into a constant to share across the two config files, but the happo one needs to be in js, and it seemed like a small thing we'll rarely touch to keep in sync. Let me know if you'd prefer I do pull it out into a constant.

ui/e2e/src/pages/DashboardPage.ts
```diff
@@ -230,4 +242,28 @@ export class DashboardPage {
...
+
+      if (mockQuery) {
+        route.fulfill(mockQuery.response);
+      } else {
+        route.continue();
+      }
+    });
+  }
+
+  async cleanupMockRequests() {
```
juliepagano, line 266: Intentionally baking this in and calling it in the fixture to avoid having to remember to always call it in `beforeEach`. If you forget the cleanup, the browser behaves weirdly when running headed tests using vscode, and I _think_ it's possible they'd mess up queries in other tests, but I haven't tested it.

ui/e2e/src/tests/gaugeChartPanel.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    await happoPlaywright.init(context);
+  });
+
+  test.afterEach(async () => {
+    await happoPlaywright.finish();
+  });
+
+  // TODO: figure out waiting for the gauge animation to finish before taking screenshots
+  test(`displays single line as expected`, async ({ page, dashboardPage, mockNow }) => {
+    // Mock data response, so we can make assertions on consistent response data.
```
juliepagano, line 34: If we don't mock the data, these tests will be inherently flaky because the demo data changes.

ui/e2e/src/tests/gaugeChartPanel.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+          },
+        },
+      ],
+    });
+
+    await dashboardPage.isLightMode();
+
+    const panel = dashboardPage.getPanel('Single Gauge');
+    await panel.isLoaded();
+    await waitForStableCanvas(panel.canvas);
```
juliepagano, line 66: The gauge chart has an animation. We want to take a screenshot when it is finished to avoid flaky snapshots.

ui/e2e/src/utils/mockTimeSeries.ts
```diff
@@ -0,0 +1,63 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { PrometheusDatasource } from '@perses-dev/prometheus-plugin';
+
+type PrometheusClient = ReturnType<typeof PrometheusDatasource['createClient']>;
+type RangeQuery = Awaited<ReturnType<PrometheusClient['rangeQuery']>>;
```
juliepagano, line 17: Inferring these types instead of adding them to the export from `@perses-dev/prometheus-plugin` because I wasn't sure we wanted to expose them generally just to get access to them for this test util. Let me know if you think those types should be exposed for use by consumers.

ui/e2e/src/utils/mockTimeSeries.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+type MockStableValueConfig = {
+  metrics: MockStableValueResultConfig[];
+  startTimeMs: number;
+  endTimeMs: number;
+  count?: number;
+};
+
+export function mockTimeSeriesResponseWithStableValue({
```
juliepagano, line 31: This is a very simple flat line as an example for how to mock responses. We'll probably want to have more complex examples in the future, but I wanted to keep this PR as a simple starting point.

ui/e2e/src/fixtures/dashboardTest.ts
```diff
@@ -28,6 +31,13 @@ type DashboardTestFixtures = {
    * tests because it will unnecessarily slow them down.
    */
   modifiesDashboard: boolean;
+
+  /**
+   * Time in unix milliseconds to be returned when `Date.now` is called on the
+   * page. Useful for stabilizing tests that depend on the current time.
+   */
+  mockNow: number;
```
eunicorn, line 39: Is it simpler if we make this optional and default it to `0`?

ui/e2e/src/tests/timeSeriesChartPanel.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+                startTimeMs: mockNow - 6 * 60 * 60 * 1000,
+                endTimeMs: mockNow,
+              })
+            ),
+          },
+        },
+      ],
+    });
+
+    await dashboardPage.isLightMode();
```
eunicorn, line 61: Hmm i wonder if there is a way to run the tests all in light mode first then dark mode. So we don't have to write this logic (toggling theme) for every test 🤔

eunicorn (approved):
💯

ui/e2e/src/fixtures/dashboardTest.ts
juliepagano: It gets made optional at the point we need to use it in tests like the example below. Something deeper in the `testBase.extend` generic type transforms it. It's defaulted to `0` in the initial setup of the fixture.

```js
test.use({
  dashboardName: 'Panels',
});
```

ui/e2e/src/tests/timeSeriesChartPanel.spec.ts
juliepagano: I was debating the same thing, but I wasn't sure if that'd obfuscate too much. Let me take a quick stab at a helper function to do this to see if I can come up with something reasonable.

juliepagano: Thanks for pushing me on that! I just pushed up a change with it, and it looks a lot cleaner, while hopefully still being readable. 🎉

ui/e2e/src/fixtures/dashboardTest.ts
juliepagano: In the `0` default case, the date is not mocked. I'm inclined to avoid mocking it all the time unless we really need it for the majority of our tests.

PR #940, opened 2023-01-26
[IGNORE] minor visual test improvements

juliepagano:
I caught a few minor issues with the visual tests while moving on to the more complicated case of creating tests for charts. I wanted to get these in first as a smaller, separate PR, and will follow-up with a PR for the chart tests later.

- Adjust check for loading to finish.
- Rename markdown test file to clarify it's for the panel.
- Use parent of panel to ensure the height/width are maintained.

# Checklist

### DIFF
--- a/ui/e2e/src/pages/Panel.ts
+++ b/ui/e2e/src/pages/Panel.ts
@@ -23,6 +23,7 @@ type resizePanelOptions = {
  */
 export class Panel {
   readonly container: Locator;
+  readonly parent: Locator;
 
   readonly editButton: Locator;
   readonly deleteButton: Locator;
@@ -34,6 +35,10 @@ export class Panel {
 
   constructor(container: Locator) {
     this.container = container;
+
+    // Useful for visual testing because the parent contains the height and
+    // width setting.
+    this.parent = this.container.locator('..');
 
     this.deleteButton = this.container.getByRole('button', {
       name: 'delete panel',
@@ -48,17 +53,18 @@ export class Panel {
     this.resizeHandle = this.container.locator('..').locator('.react-resizable-handle');
 
     this.figure = this.container.getByRole('figure');
-    this.loader = this.container.getByLabel('Loading');
+    this.loader = this.container.locator('[aria-label*=Loading]');
   }
 
   async isLoaded() {
     // Wait for the figure to have at least one visible child that is not the loader.
     await expect(async () => {
-      expect(this.figure).not.toContain(this.loader);
+      await expect(this.loader).toHaveCount(0);
 
       const figureChildren = this.figure.locator('*:visible');
-
       expect(await figureChildren.count()).toBeGreaterThan(0);
+
+      await expect(this.loader).toHaveCount(0);
     }).toPass();
 
     // Trial a click to wait for the figure to be stable. Replace this with a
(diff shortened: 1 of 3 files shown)


ui/e2e/src/pages/Panel.ts
```diff
@@ -48,17 +53,18 @@ export class Panel {
     this.resizeHandle = this.container.locator('..').locator('.react-resizable-handle');
 
     this.figure = this.container.getByRole('figure');
-    this.loader = this.container.getByLabel('Loading');
+    this.loader = this.container.locator('[aria-label*=Loading]');
```
juliepagano, line 56: This wasn't working quite how I expected -- I think it's focused on input labels. Moving to a more explicit locator.

ui/e2e/src/pages/Panel.ts
```diff
@@ -48,17 +53,18 @@ export class Panel {
...
     // Wait for the figure to have at least one visible child that is not the loader.
     await expect(async () => {
-      expect(this.figure).not.toContain(this.loader);
+      await expect(this.loader).toHaveCount(0);
 
       const figureChildren = this.figure.locator('*:visible');
-
       expect(await figureChildren.count()).toBeGreaterThan(0);
+
+      await expect(this.loader).toHaveCount(0);
```
juliepagano, line 67: Checking both before and after looking for children just to be extra sure and avoid a small edge case with timing that might lead to a flaky test. There might be some complex regex or something I could do with a single check, but this was a bit simpler to write.

ui/e2e/src/tests/markdownPanel.spec.ts
```diff
@@ -34,7 +34,7 @@ test.describe('Dashboard: Markdown Panel', () => {
       const markdownPanel = dashboardPage.getPanel(panelName);
       await markdownPanel.isLoaded();
 
-      await happoPlaywright.screenshot(page, markdownPanel.container, {
+      await happoPlaywright.screenshot(page, markdownPanel.parent, {
```
juliepagano, line 37: The width/height live on the container, so previously the screenshots were filling the whole page. This was way more noticeable when working on the canvas visualizations because happo turns them into images, so they weren't stretching. Moving to the parent, keeps them the size we intend to test.

juliepagano: I swapped them to a 3 across size, so they're a little bigger now and probably at a common size people might use them in a dashboard. Does that look better to you?

cndonovan: Oh where should I go to look, it looks like the Happo link above still points to old ones

juliepagano: Oops, sorry about there. Here's the new one. https://happo.io/a/1009/p/1488/compare/9e577592bcdc75699bd55ad53ab0026e5e80ecf6/58c5e98ef29df87f73f159015136533ce65ea65c

You can also find it by clicking the "details" link on the happo check 
image

cndonovan (approved):

ui/e2e/src/tests/markdownPanel.spec.ts
cndonovan: Looks great, I like it!

PR #942, opened 2023-01-27
[BREAKINGCHANGE] Move panel padding to plugins

juliepagano:
Previously, the panel content always added padding. This would lead to some annoying workarounds in plugins where the visualization wanted to have less or no padding.

After this fix, the panel content adds no padding and any spacing needs have moved down into the individual plugins.

This commit also adjusts some minor spacing and alignment issues in the plugins.

BREAKINGCHANGE: Panel plugins external to the perses repository can no longer depend on the container providing padding and will likely need to make some styling adjustments after this change.

# Checklist

# Screenshots

Click the "details" link for happo in the status checks to see our visual tests in action! Note that there are some issues related to font loading that are being addressed in https://github.com/perses/perses/pull/943.
image

## Time series chart panel

Should look and behave identically to before.

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>

## Gauge chart panel

I left off the padding in the gauge chart and rely on the underlying padding in stack. This is mostly noticeable in the "multiple gauges" because the single gauge is center aligned.

### Single gauge

While in here, I made some adjustments to the use case with a single gauge to center it, which looked nicer in that use case (to my eyes, anyways -- let me know if you think it was better the other way).

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>

### Multiple gauges

Note a slight spacing difference here based on the adjustments to padding.

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>

## Start chart panel

Note that the stat chart is a teeny bit lower than previously. This is intentional to avoid the chart potentially overlapping with the text value above when refactoring some of the styling.

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>

## Markdown panel

Should look and behave identically to before. 

Note that the diffs for markdown in the happo output are the result of a font loading issue that is being resolved in https://github.com/perses/perses/pull/943.

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>

### DIFF
--- a/ui/components/src/StatChart/StatChart.tsx
+++ b/ui/components/src/StatChart/StatChart.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import React, { useMemo } from 'react';
-import { Box, Typography, Stack } from '@mui/material';
+import { Box, Typography } from '@mui/material';
 import { merge } from 'lodash-es';
 import { use, EChartsCoreOption } from 'echarts/core';
 import { GaugeChart as EChartsGaugeChart, GaugeSeriesOption } from 'echarts/charts';
@@ -115,13 +115,13 @@ export function StatChart(props: StatChartProps) {
   const containerPadding = `${chartsTheme.container.padding.default}px`;
 
   return (
-    <Stack sx={{ height: '100%', width: '100%' }} spacing={0}>
+    <Box sx={{ height: '100%', width: '100%', display: 'flex', flexDirection: 'column' }}>
       <Typography
         variant="h3"
         sx={(theme) => ({
           color: theme.palette.text.primary,
           fontSize: `clamp(${MIN_VALUE_SIZE}px, ${valueSize}px, ${MAX_VALUE_SIZE}px)`,
-          padding: [containerPadding, containerPadding, 0, containerPadding],
+          padding: `${containerPadding} ${containerPadding} 0 ${containerPadding}`,
         })}
       >
         {formattedValue}
@@ -139,6 +139,6 @@ export function StatChart(props: StatChartProps) {
           />
         </Box>
       )}
-    </Stack>
+    </Box>
   );
 }
--- a/ui/components/src/StatChart/StatChart.tsx
+++ b/ui/components/src/StatChart/StatChart.tsx
@@ -112,6 +112,8 @@ export function StatChart(props: StatChartProps) {
   const charactersAdjust = formattedValue.length;
   const valueSize = isLargePanel === true ? MAX_VALUE_SIZE : Math.min(width, height) / charactersAdjust;
 
+  const containerPadding = `${chartsTheme.container.padding.default}px`;
+
   return (
     <Stack sx={{ height: '100%', width: '100%' }} spacing={0}>
       <Typography
@@ -119,7 +121,7 @@ export function StatChart(props: StatChartProps) {
         sx={(theme) => ({
           color: theme.palette.text.primary,
           fontSize: `clamp(${MIN_VALUE_SIZE}px, ${valueSize}px, ${MAX_VALUE_SIZE}px)`,
-          padding: (theme) => theme.spacing(1.5, 1.5, 0, 1.5),
+          padding: [containerPadding, containerPadding, 0, containerPadding],
         })}
       >
         {formattedValue}
--- a/ui/components/src/StatChart/StatChart.tsx
+++ b/ui/components/src/StatChart/StatChart.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import React, { useMemo } from 'react';
-import { Box, Typography } from '@mui/material';
+import { Box, Typography, Stack } from '@mui/material';
 import { merge } from 'lodash-es';
 import { use, EChartsCoreOption } from 'echarts/core';
 import { GaugeChart as EChartsGaugeChart, GaugeSeriesOption } from 'echarts/charts';
@@ -34,7 +34,6 @@ use([
   CanvasRenderer,
 ]);
 
-const PANEL_PADDING = 32;
 const MIN_VALUE_SIZE = 12;
 const MAX_VALUE_SIZE = 36;
 
@@ -114,30 +113,30 @@ export function StatChart(props: StatChartProps) {
   const valueSize = isLargePanel === true ? MAX_VALUE_SIZE : Math.min(width, height) / charactersAdjust;
 
   return (
-    <Box>
+    <Stack sx={{ height: '100%', width: '100%' }} spacing={0}>
       <Typography
         variant="h3"
         sx={(theme) => ({
           color: theme.palette.text.primary,
           fontSize: `clamp(${MIN_VALUE_SIZE}px, ${valueSize}px, ${MAX_VALUE_SIZE}px)`,
+          padding: (theme) => theme.spacing(1.5, 1.5, 0, 1.5),
         })}
       >
         {formattedValue}
       </Typography>
       {sparkline !== undefined && (
-        <EChart
-          sx={{
-            width: width + PANEL_PADDING, // allows sparkline to extend to edge of panel
-            height: height,
-            position: 'absolute',
-            bottom: 0,
-            left: 0,
-          }}
-          option={option}
-          theme={chartsTheme.echartsTheme}
-          renderer="svg"
-        />
+        <Box sx={{ flex: 1 }}>
+          <EChart
+            sx={{
+              width: '100%',
+              height: '100%',
+            }}
+            option={option}
+            theme={chartsTheme.echartsTheme}
+            renderer="svg"
+          />
+        </Box>
       )}
-    </Box>
+    </Stack>
   );
 }
--- a/ui/components/src/model/theme.ts
+++ b/ui/components/src/model/theme.ts
@@ -20,6 +20,17 @@ export interface PersesChartsTheme {
     width: number;
     color: string;
   };
+  /**
+   * Theming for the container that wraps a chart.
+   */
+  container: {
+    /**
+     * Padding in pixels.
+     */
+    padding: {
+      default: number;
+    };
+  };
 }
 
 // https://github.com/apache/echarts/issues/12489#issuecomment-643185207
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -14,7 +14,7 @@
 import { useState, useMemo } from 'react';
 import useResizeObserver from 'use-resize-observer';
 import { useInView } from 'react-intersection-observer';
-import { ErrorBoundary, ErrorAlert, combineSx, useId } from '@perses-dev/components';
+import { ErrorBoundary, ErrorAlert, combineSx, useId, useChartsTheme } from '@perses-dev/components';
 import { PanelDefinition } from '@perses-dev/core';
 import { Card, CardProps, CardContent } from '@mui/material';
 import { PanelHeader, PanelHeaderProps } from './PanelHeader';
@@ -51,8 +51,7 @@ export function Panel(props: PanelProps) {
     triggerOnce: true,
   });
 
-  // TODO: adjust padding for small panels, consistent way to determine isLargePanel here and in StatChart
-  const panelPadding = 1.5;
+  const chartsTheme = useChartsTheme();
 
   const handleMouseEnter: CardProps['onMouseEnter'] = (e) => {
     setIsHovered(true);
@@ -91,7 +90,7 @@ export function Panel(props: PanelProps) {
         description={definition.spec.display.description}
         editHandlers={editHandlers}
         isHovered={isHovered}
-        sx={{ paddingX: (theme) => theme.spacing(panelPadding) }}
+        sx={{ paddingX: `${chartsTheme.container.padding.default}px` }}
       />
       <CardContent
         component="figure"
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -100,10 +100,10 @@ export function Panel(props: PanelProps) {
           overflow: 'hidden',
           flexGrow: 1,
           margin: 0,
-          padding: (theme) => theme.spacing(panelPadding),
+          padding: 0,
           // Override MUI default style for last-child
           ':last-child': {
-            padding: (theme) => theme.spacing(panelPadding),
+            padding: 0,
           },
         }}
         ref={setContentElement}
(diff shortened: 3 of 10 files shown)


ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
```diff
@@ -110,11 +110,13 @@ export function GaugeChartPanel(props: GaugeChartPanelProps) {
...
 
+  const hasMultipleCharts = gaugeData.length > 1;
+
   return (
     <Stack
       direction="row"
-      spacing={2}
-      justifyContent="left"
+      spacing={hasMultipleCharts ? 2 : 0}
+      justifyContent={hasMultipleCharts ? 'left' : 'center'}
```
juliepagano, line 119: I noticed while I was in there that gauge charts look a little off when they're left aligned when there's a single gauge chart. Adjusting the behavior to center it when there's just one. If folks like it better the other way, let me know, and I can undo this bit.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -46,6 +46,15 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
   } = props;
+  const theme = useTheme();
+
+  const contentPadding = parseInt(theme.spacing(1.5), 10);
+  const adjustedContentDimensions: typeof contentDimensions = contentDimensions
+    ? {
+        width: contentDimensions.width - contentPadding * 2,
+        height: contentDimensions.height - contentPadding * 2,
+      }
+    : undefined;
```
juliepagano, line 57: This feels a little yucky, and the time series chart could probably be significantly refactored to do something different. However, I was inclined not to shave that yak as part of this PR and leave it for the next time we're putting more significant attention into the time series panel and need to refactor anyways (e.g. adjusting the legend to be more in line with the mocks from design).

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartPanel.tsx
sjcobb: SGTM, thanks for fixing this!

sjcobb (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -46,6 +46,18 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     spec: { queries, thresholds, y_axis },
     contentDimensions,
   } = props;
+  const theme = useTheme();
```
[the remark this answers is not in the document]
juliepagano: Good call! I'll fix that up and ask you for a second pass on the changes.

ui/components/src/model/theme.ts
```diff
@@ -20,6 +20,17 @@ export interface PersesChartsTheme {
...
   };
+  /**
+   * Theming for the container that wraps a chart.
+   */
+  container: {
+    /**
+     * Padding in pixels.
+     */
+    padding: {
+      default: number;
```
juliepagano, line 31: Intentionally including a `default` to start here, so we can potentially specify other spacing in the future without introducing a breaking change in external plugins that depend on the value. e.g. I could imagine us specifying a "small" value we want to use for charts that fit a little more tightly.

ui/components/src/model/theme.ts
```diff
@@ -20,6 +20,17 @@ export interface PersesChartsTheme {
     width: number;
     color: string;
   };
+  /**
+   * Theming for the container that wraps a chart.
+   */
+  container: {
```
juliepagano, line 26: Does this general structure/naming of the theme entry look good to you? Very open to feedback now since changing it later can cause breaking changes.

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
```diff
@@ -16,12 +16,14 @@ import * as DOMPurify from 'dompurify';
...
 import { PanelProps } from '@perses-dev/plugin-system';
+import { PersesChartsTheme, useChartsTheme } from '@perses-dev/components';
 import { MarkdownPanelOptions } from './markdown-panel-model';
 
 export type MarkdownPanelProps = PanelProps<MarkdownPanelOptions>;
 
-function createMarkdownPanelStyles(theme: Theme) {
+function createMarkdownPanelStyles(theme: Theme, chartsTheme: PersesChartsTheme) {
   return {
+    padding: `${chartsTheme.container.padding.default}px`,
```
cndonovan, line 26: I had this thought that I might remove the CSS rule below that says to remove the margin on the first element in the panel (see comment below)

The reason for this rule was that most HTML elements have built-in margins (h1, h2, p, etc) and their margins were combining themselves with the panel padding and creating too much space

I was thinking that once we had the ability to adjust padding, we could NOT set the vertical padding, and allow elements to use their built-in margins

However, now I'm thinking that I'm not sure all the HTML elements have built-in margins (tables? links?), so it's probably safer to have padding, and continue to remove the first element's top margin

All of which to say, will you delete that TODO comment that I have below, and maybe just say "Ignore the top margin on the first element because it creates too much space when combined with panel padding" 

(Unless what I'm saying makes sense to you, and you want to take the approach)

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -100,10 +100,10 @@ export function Panel(props: PanelProps) {
           overflow: 'hidden',
           flexGrow: 1,
           margin: 0,
-          padding: (theme) => theme.spacing(panelPadding),
+          padding: 0,
```
cndonovan, line 103: Looks like we might want to rename `panelPadding` variable / get rid of the comment above -- looks like it's still used for PanelHeader, mahybe we just inline it

juliepagano: Good catch! I can use the value I moved up into the theme.

ui/panels-plugin/src/plugins/markdown/MarkdownPanel.tsx
juliepagano: I agree that it's safer to keep the `:first-child` top padding removal for exactly the reason you mentioned about inconsistent padding on different elements. I'll take out the TODO.

ui/components/src/model/theme.ts
cndonovan: I just took a peek at EChartsTheme's JSON to see what they do: https://echarts.apache.org/en/theme-builder.html

The only thing coming to mind right now is maybe we want to call this `containerPadding` instead of `container.padding`

In ECharts, it seems like most of the top-level fields in "theme" are the individual panels, which makes sense to me... each panel has its own options

But then there are some global fields, like `color` and `backgroundColor`

It does seem nice to have a default padding size for all panels, but I can't think of anything else we'll put inside `container`.

Thus, collapsing the fields `container` and `padding` into just `containerPadding`

But I'm not 100% of this idea, just something to think about. I can also see that reasons to have `container` mean that if we do come up with another property of container that we want global styles for, we can put it here. 

Just not sure if global styles for all panels will be common

cndonovan (approved):
lgtm 

Looking at the PR as a whole, part of me thought it could be cool if the padding was set either: (a) always in the "Panel" component, e.g. GaugeChartPanel, TimeSeriesChartPanel, or (b) always in the lower level component, e.g. GaugeChart, LineChart.

But it didn't peek at the code to see if that's actually feasible. My guess is that it's not, and I don't think that kind of symmetry is actually that necessary

juliepagano:
> lgtm!
> 
> Looking at the PR as a whole, part of me thought it could be cool if the padding was set either: (a) always in the "Panel" component, e.g. GaugeChartPanel, TimeSeriesChartPanel, or (b) always in the lower level component, e.g. GaugeChart, LineChart.
> 
> But it didn't peek at the code to see if that's actually feasible. My guess is that it's not, and I don't think that kind of symmetry is actually that necessary

I think it should be that way longer-term, but would have required some significant refactoring, especially for the time series because of how the legends work (maybe they should also move down to the chart?). I think we should fix that in the future, but didn't want to shave that particular yak in this PR.

ui/components/src/model/theme.ts
juliepagano: I was also struggling with that. I opted for the object to leave room to grow without requiring breaking changes, but it is a little awkward as empty as it is right now, and it's hard to predict what we'll need in there in the future. @sjcobb -- you have any strong feels one way or the other to sway us?

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -113,31 +112,33 @@ export function StatChart(props: StatChartProps) {
...
+
   return (
-    <Box>
+    <Stack sx={{ height: '100%', width: '100%' }} spacing={0}>
       <Typography
         variant="h3"
         sx={(theme) => ({
           color: theme.palette.text.primary,
           fontSize: `clamp(${MIN_VALUE_SIZE}px, ${valueSize}px, ${MAX_VALUE_SIZE}px)`,
+          padding: [containerPadding, containerPadding, 0, containerPadding],
```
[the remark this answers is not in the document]
juliepagano: This allows the metric in the stat chart to fit snugly with the line chart below it. With the previous CSS, they actually had some overlap, which could cause UX issues in some cases.

eunicorn:
>> part of me thought it could be cool if the padding was set either: (a) always in the "Panel" component, e.g. GaugeChartPanel, TimeSeriesChartPanel, or (b) always in the lower level component, e.g. GaugeChart, LineChart.

maybe it's fine to have the padding in plugins because we don't know what the padding should be for other custom panels?

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -46,6 +47,18 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     spec: { queries, thresholds, y_axis },
     contentDimensions,
   } = props;
+  const chartsTheme = useChartsTheme();
+
+  // TODO: consider refactoring how the layout/spacing/alignment are calculated
+  // the next time significant changes are made the the time series panel (e.g.
```
eunicorn, line 53: nit:
```suggestion
  // the next time significant changes are made to the time series panel (e.g.
```

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -113,31 +112,33 @@ export function StatChart(props: StatChartProps) {
   const charactersAdjust = formattedValue.length;
   const valueSize = isLargePanel === true ? MAX_VALUE_SIZE : Math.min(width, height) / charactersAdjust;
 
+  const containerPadding = `${chartsTheme.container.padding.default}px`;
+
   return (
-    <Box>
+    <Stack sx={{ height: '100%', width: '100%' }} spacing={0}>
```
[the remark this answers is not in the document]
juliepagano: It does some flexbox things for us, so the two values sit nicely together to fill the space. I could also set the flex values on `Box` in `sx` if that'd be cleaner. I'm still getting used to working with MUI. Let me know if you think that'd be better.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
eunicorn: hmm seems like it makes sense to move the legend to the TimeSeriesChart level then  🤔

juliepagano:
> > > part of me thought it could be cool if the padding was set either: (a) always in the "Panel" component, e.g. GaugeChartPanel, TimeSeriesChartPanel, or (b) always in the lower level component, e.g. GaugeChart, LineChart.
> 
> maybe it's fine to have the padding in plugins because we don't know what the padding should be for other custom panels?

I could see that too. Definitely feels like we need to revisit where we want those decisions to be made around the embedded use case. Maybe something it will be easier to sort out when we add the storybook and can more easily play around with that use case?

ui/components/src/StatChart/StatChart.tsx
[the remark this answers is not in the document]
eunicorn: Personally I like setting `flex` on `Box` or just `div` better because `Stack` usually means you want even spacing between elements and i've heard there's performance issue with using too many `Box` and `Stack`. But i don't feel strongly about this 🤷‍♀️

eunicorn (approved):

ui/components/src/StatChart/StatChart.tsx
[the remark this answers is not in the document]
juliepagano: That's good to know! I wasn't aware of the perf bits since I'm still learning MUI. Let me take a quick stab at just doing the flex on the `Box`.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
juliepagano: Yeah, I think there's a strong argument to move it down the next time we make any meaningful changes to the legend. I'll add a bit about that in this comment too.

ui/components/src/model/theme.ts
sjcobb: This is all outside the `echartsTheme` so we can really name it whatever we want, `sparkline` for instance is also a custom property that I made up.  Normally with ECharts you customize the padding using the [grid properties](https://echarts.apache.org/en/option.html#grid), but since we've built our own Legend and YAxisLabel components with React, that only helps us for the inner portions rendered to canvas. 

I'm good with `container.padding` though! Maybe we'll think of other properties to add inside container later like you said (other apps might want to change the border or something)

sjcobb (approved):
🎉

ui/components/src/StatChart/StatChart.tsx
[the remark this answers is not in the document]
cndonovan: Ah nice

PR #943, opened 2023-01-27
[IGNORE] Short-term fix for visual test fonts

juliepagano:
Prior to this fix, the visual tests were not using the correct font because the happo asset uploader only caught relative asset links and our fonts use a full link including localhost.

The addition of the `HAPPO_DOWNLOAD_ALL=true` env var forces the happo asset uploader to grab all assets, which will grab the fonts. This is the easiest short-term fix. Their support team will take a look at adjusting the script to support non-relative asset links with localhost in them, and we can hopefully pull down and get a fix in the near future. We have very few external assets, so the side effects of using this env var should be minimal even if we need it for a longer period of time.

# Checklist

### DIFF
--- a/ui/e2e/package.json
+++ b/ui/e2e/package.json
@@ -9,7 +9,7 @@
     "e2e:debug": "npm run e2e:local -- --debug",
     "e2e:report": "playwright show-report",
     "e2e:watch": "chokidar src/tests/**/**.ts -c 'playwright test {path}'",
-    "e2e:ci": "npx happo-e2e -- playwright test --config=src/config/ci.playwright.config.ts",
+    "e2e:ci": "HAPPO_DOWNLOAD_ALL=true npx happo-e2e -- playwright test --config=src/config/ci.playwright.config.ts",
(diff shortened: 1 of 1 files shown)


juliepagano:
The failed go job seems completely unrelated to my change -- something about having trouble finding go releaser. Re-running and hoping it's a blip.

juliepagano:
> The failed go job seems completely unrelated to my change -- something about having trouble finding go releaser. Re-running and hoping it's a blip.

Still broken. I'll try again on Monday, and dig into what's going on if it's not fixed then.

juliepagano:
> > The failed go job seems completely unrelated to my change -- something about having trouble finding go releaser. Re-running and hoping it's a blip.
> 
> Still broken. I'll try again on Monday, and dig into what's going on if it's not fixed then.

It's fixed now, so this should be good to go. Augustin mentioned it might've been some unfortunate timing around a new release of the package that was throwing an error.

sjcobb (approved):

PR #946, opened 2023-01-30
[BUGFIX] round small decimal values when abbreviate is set

sjcobb:
## Overview

In certain cases, when a custom unit such as `"unit": { "kind": "Decimal", "decimal_places": 2, "abbreviate": true }` is used, the value is not rounded. This is most noticeable for smaller values in the tooltip. Now the abbreviateLargeNumber function is only called when a value is 1000 or greater, otherwise decimalFormatter.format is called like normal.

## Before Screenshot
tooltip_rounding_bug_before

## After Screenshot
tooltip_rounding_bug_after_fixed

# Checklist

### DIFF
--- a/ui/components/src/model/units/units.test.ts
+++ b/ui/components/src/model/units/units.test.ts
@@ -32,6 +32,26 @@ describe('formatValue', () => {
       expected: '155,900.0000',
     },
     {
+      value: 1000,
+      unit: { kind: 'Decimal', decimal_places: 2, abbreviate: true },
+      expected: '1K',
+    },
+    {
+      value: 1590.878787,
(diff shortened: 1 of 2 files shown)


ui/components/src/model/units/units.test.ts
```diff
@@ -31,6 +31,36 @@ describe('formatValue', () => {
       unit: { kind: 'Decimal', decimal_places: 4 },
       expected: '155,900.0000',
     },
+    {
```
juliepagano, line 34: Yay, more tests! 😍

juliepagano (approved):
LGTM. Thanks for fixing this up! 🥇

PR #916, opened 2023-01-10
[FEATURE] Initiate project creation/deletion in the frontend

celian-garcia:
Signed-off-by: Célian Garcia <celian.garcia@amadeus.com>

This PR initiates the project CRUD. It includes : 
- [x] An "ADD PROJECT" button in the home page (dialog redirecting to the project once submitted)
- [x] A delete icon button on each project of the home page  (with confirmation dialog)
- [x] A "DELETE PROJECT" button in the project page  (with confirmation dialog)

> Regarding the second point, to be discussed with designer about the clarity of having the dashboards stored by project in the home page view. 
> Despite the bug that removing a project does not remove dashboard (will be soon fixed), I would find clearer to have directly a list of dashboards (with project mention on the line).

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.

# Screenshots 

## VIEWS
image
image

## CREATION
image
image
image
image

## DELETION
image

PR #948, opened 2023-01-30
[FEATURE] Generic Dialog Components

eunicorn:
https://github.com/perses/perses/discussions/935

Based on the discussion, we want generic dialog components that include common UX patterns:

1. Close icon button at the top right corner
2. Dividers for dialog content
3. For dialog actions, we always want primary button with contained variant on the left and secondary button (usually Cancel) with outlined variant on the right

### DIFF
--- a/ui/components/src/Dialog/Dialog.tsx
+++ b/ui/components/src/Dialog/Dialog.tsx
@@ -14,13 +14,16 @@
 import React from 'react';
 import {
   Button,
+  ButtonProps,
   Dialog as MuiDialog,
   DialogActions,
   DialogContent,
   DialogContentProps as MuiDialogContentProps,
+  DialogProps,
   DialogTitle,
   DialogTitleProps,
   IconButton,
+  Theme,
 } from '@mui/material';
 import CloseIcon from 'mdi-material-ui/Close';
 import { combineSx } from '../utils';
@@ -32,7 +35,7 @@ export interface DialogHeaderProps extends DialogTitleProps {
   onClose?: (e: React.MouseEvent<HTMLElement>) => void;
 }
 
-export interface DialogButtonProps {
+export interface DialogButtonProps extends Omit<ButtonProps, 'variant' | 'color'> {
   children: React.ReactNode;
   onClick?: (e: React.MouseEvent<HTMLElement>) => void;
   /**
@@ -54,15 +57,7 @@ const Header = ({ children, onClose, ...props }: DialogHeaderProps) => {
     <>
       <DialogTitle {...props}>{children}</DialogTitle>
       {onClose && (
-        <IconButton
-          aria-label="Close"
-          onClick={onClose}
-          sx={(theme) => ({
-            position: 'absolute',
-            top: theme.spacing(0.5),
-            right: theme.spacing(0.5),
-          })}
-        >
+        <IconButton aria-label="Close" onClick={onClose} sx={dialogCloseIconButtonStyle}>
           <CloseIcon />
         </IconButton>
       )}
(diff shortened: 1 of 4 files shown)


ui/components/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { combineSx } from '../utils';
+
+export interface DialogHeaderProps extends DialogTitleProps {
+  /**
+   * Callback fired when close button is clicked. If undefined, close button will not appear in header.
+   */
+  onClose?: (e: React.MouseEvent<HTMLElement>) => void;
+}
+
+export interface DialogButtonProps {
```
celian-garcia, line 35: You should probably extend the material ButtonProps here as well, to have all the power of a regular button. 

If you don't want the ``variant`` and ``type`` properties  you can use Omit

```ts
/**
 * following properties are removed from the DialogButtonProps
 */
interface UnwantedButtonProps {
  variant: unknown;
  type: unknown;
}

export interface DialogButtonProps extends Omit<ButtonProps, keyof UnwantedButtonProps> {}
```

ui/components/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const Header = ({ children, onClose, ...props }: DialogHeaderProps) => {
+  return (
+    <>
+      <DialogTitle {...props}>{children}</DialogTitle>
+      {onClose && (
+        <IconButton
+          aria-label="Close"
+          onClick={onClose}
+          sx={(theme) => ({
```
celian-garcia, line 60: In the Project CRUD PR, I wrapped this code in a function to optimize rendering performance (an advice from sonarlint). 

I'm not sure it's optimizing that much performances, but I found it clearer at the end. You can go and see if you find it interesting https://github.com/perses/perses/pull/916/files#diff-ce2596b6d1065537010136370c2489af315269aaafa78564f55f9bfd19cd80eeR20-R26

celian-garcia (changes requested):
Great! That would remove already the main duplication issues in the dialogs.

ui/components/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+);
+
+const SecondaryButton = ({ children, onClick }: DialogButtonProps) => (
+  <Button variant="outlined" color="secondary" onClick={onClick}>
+    {children}
+  </Button>
+);
+
+export const Dialog = {
+  Dialog: MuiDialog,
```
saminzadeh, line 92: `Dialog.Dialog` feels bit weird. A few ideas:

```
Dialog.Wrapper
Dialog.Base
```
or just expose this as `Dialog` and only add the sub components.
```tsx
export function Dialog() {...}
Dialog.Header = Header;
```

```tsx
<Dialog>
    <Dialog.Header>My Title</Dialog.Header>
</Dialog>

File: ui/components/src/Dialog/Dialog.tsx [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+import { combineSx } from '../utils';
+
+export interface DialogHeaderProps extends DialogTitleProps {
+  /**
+   * Callback fired when close button is clicked. If undefined, close button will not appear in header.
+   */
+  onClose?: (e: React.MouseEvent<HTMLElement>) => void;
+}
+
+export interface DialogButtonProps extends Omit<ButtonProps, 'variant' | 'color'> {
```
celian-garcia, line 38: If you don't put ``type`` in the excluded properties, somebody using it will have some issues no ?

celian-garcia (approved):

eunicorn:
@saminzadeh @cndonovan can one of you take another look when you have the chance? thanks!

saminzadeh (approved):

PR #954, opened 2023-02-07
[ENHANCEMENT] Modify placement of duplicate panels

juliepagano:
Prior to this change, duplicate panels would always be placed on the
far left in a new row at the bottom of the dashboard.

After this change, duplicate panels are placed as follows:
- If there is space directly to the right of the panel being duplicated,
  the duplicate will be placed there.
- Otherwise, the duplicate panel is placed directly below the panel
  being duplicated, pushing down any panels below it.

This approach to placement of duplicate panels balances the following
considerations:
- Making it easy for the user to find a duplicate by initially placing
  it close to the original panel.
- Minimizing rework for the user that may be caused if we move other
  panels to make place for the duplicate panel. The grid compacts
  vertically, so moving other panels downward while retaining their
  horizontal positioning is less disruptive than changing horizontal
  positioning.

# Checklist

# Screenshots

See [added examples in Happo screenshots](https://happo.io/a/1009/p/1488/compare/d22943c5c7cab3726bcf00f4b6b8174c0ba9a2e7/04987906de88563ba5534e6448e5e53a44357f25?t=added).

### DIFF
--- a/ui/app/src/components/Footer.tsx
+++ b/ui/app/src/components/Footer.tsx
@@ -35,7 +35,7 @@ export default function Footer(): JSX.Element {
             <Github sx={{ verticalAlign: 'bottom' }} />
           </a>
         </li>
-        <li>
+        <li data-happo-hide>
           {isLoading ? (
             <CircularProgress size="1rem" />
           ) : data !== undefined && data.version !== '' ? (
--- a/ui/dashboards/src/utils/panelUtils.ts
+++ b/ui/dashboards/src/utils/panelUtils.ts
@@ -81,12 +81,11 @@ export function insertPanelInLayout(
 
   const referenceBounds = getPanelBounds(referenceLayout);
 
+  // Organize layouts based on vertical relation to the item being inserted
+  // after.
   const aboveInsertRow: PanelGroupItemLayout[] = [];
   const insertRow: PanelGroupItemLayout[] = [];
   const belowInsertRow: PanelGroupItemLayout[] = [];
-
-  // Organize layouts based on vertical relation to the item being inserted
-  // after.
   itemLayouts.forEach((itemLayout) => {
     const itemBounds = getPanelBounds(itemLayout);
 
--- a/ui/dashboards/src/utils/panelUtils.ts
+++ b/ui/dashboards/src/utils/panelUtils.ts
@@ -135,7 +135,7 @@ export function insertPanelInLayout(
     }
   }
 
-  // Inserted the new item below the original and shift the items below the
+  // Insert the new item below the original and shift the items below the
   // row where the reference is located.
   return [
     ...aboveInsertRow,
--- a/ui/dashboards/src/utils/panelUtils.ts
+++ b/ui/dashboards/src/utils/panelUtils.ts
@@ -55,7 +55,7 @@ function getPanelBounds({ x, y, w, h }: PanelGroupItemLayout): PanelGroupItemBou
   };
 }
 
-export type PartialPanelGroupItemLayout = Omit<PanelGroupItemLayout, 'x' | 'y'>;
+export type UnpositionedPanelGroupItemLayout = Omit<PanelGroupItemLayout, 'x' | 'y'>;
 
 /**
  * Inserts a new panel into the layout with placement determined by a specified
@@ -73,7 +73,7 @@ export type PartialPanelGroupItemLayout = Omit<PanelGroupItemLayout, 'x' | 'y'>;
  * @returns - Item layouts modified to insert the new panel.
  */
 export function insertPanelInLayout(
-  newLayout: PartialPanelGroupItemLayout,
+  newLayout: UnpositionedPanelGroupItemLayout,
   referenceLayout: PanelGroupItemLayout,
   itemLayouts: PanelGroupItemLayout[]
 ): PanelGroupItemLayout[] {
--- a/ui/dashboards/src/utils/panelUtils.ts
+++ b/ui/dashboards/src/utils/panelUtils.ts
@@ -119,9 +119,7 @@ export function insertPanelInLayout(
         ...belowInsertRow,
       ];
     }
-  }
-
-  if (insertAfterIndex >= 0) {
+  } else if (insertAfterIndex >= 0) {
     const nextItem = insertRow[insertAfterIndex + 1];
 
     if (nextItem && getPanelBounds(nextItem).x1 - referenceBounds.x2 >= newLayout.w) {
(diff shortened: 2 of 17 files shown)


ui/app/src/components/Footer.tsx
```diff
@@ -35,7 +35,7 @@ export default function Footer(): JSX.Element {
             <Github sx={{ verticalAlign: 'bottom' }} />
           </a>
         </li>
-        <li>
+        <li data-happo-hide>
```
juliepagano, line 38: Hiding the version number in the footer from happo because it is something that will change somewhat regularly, leading to flaky snapshots.

ui/dashboards/src/utils/panelUtils.ts
```diff
@@ -26,6 +27,133 @@ export function getYForNewRow(group: PanelGroupDefinition) {
...
+ *   this location, they will also shift downward because the grid uses
+ *   vertical-based compacting.
+ *
+ * @param newLayout - Layout for new panel to insert into the grid.
+ * @param referenceLayout - Layout for reference panel used to determine the
+ *   placement of the new panel.
+ * @param itemLayouts - Full grid layout.
+ * @returns - Item layouts modified to insert the new panel.
+ */
+export function insertPanelInLayout(
```
juliepagano, line 75: I left this logic relatively generic around panel insertion (instead of very focused on duplication) because we may want to use it in the future for other functionality like copy-paste.

ui/dashboards/src/utils/panelUtils.ts
```diff
@@ -26,6 +27,133 @@ export function getYForNewRow(group: PanelGroupDefinition) {
...
+
+/**
+ * Inserts a new panel into the layout with placement determined by a specified
+ * reference panel. The new panel is placed:
+ * - To the right of the reference panel if there is space available without
+ *   moving other panels.
+ * - Otherwise, directly below the reference panel. If other panels are below
+ *   this location, they will also shift downward because the grid uses
+ *   vertical-based compacting.
+ *
```
juliepagano, line 68: I left this logic relatively generic around panel insertion (instead of very focused on duplication) because we may want to use it in the future for other functionality like copy-paste.

ui/dashboards/src/utils/panelUtils.ts
```diff
@@ -26,6 +27,132 @@ export function getYForNewRow(group: PanelGroupDefinition) {
...
+ * @param newLayout - Layout for new panel to insert into the grid.
+ * @param referenceLayout - Layout for reference panel used to determine the
+ *   placement of the new panel.
+ * @param itemLayouts - Full grid layout.
+ * @returns - Item layouts modified to insert the new panel.
+ */
+export function insertPanelInLayout(
```
juliepagano, line 75: I wrote this function focused more on readability/maintainability than performance at scale based on the assumption that a dashboard is unlikely to have a massive scale of panels that would cause issues with the iteration & duplication of panels below. Let me know if you disagree with that order of priorities and think it should optimize more for perf.

ui/e2e/src/pages/AppHomePage.ts
```diff
@@ -52,6 +52,7 @@ export class AppHomePage {
   async clickDashboardItem(dashboardName: string) {
     const dashboardButton = this.page.getByRole('button', {
       name: dashboardName,
+      exact: true,
```
juliepagano, line 55: Needed because there are multiple dashboards with the name "panels" in them now.

ui/e2e/src/pages/DashboardPage.ts
```diff
@@ -56,9 +77,6 @@ export class DashboardPage {
   readonly panelGroups: Locator;
   readonly panelGroupHeadings: Locator;
 
-  readonly panels: Locator;
```
juliepagano, line 59: Removed these in favor of `getPanels` and `getPanelHeadings` methods that have more flexibility.

ui/e2e/src/tests/panels.spec.ts
```diff
@@ -141,60 +141,4 @@ test.describe('Dashboard: Panels', () => {
       expect(largePanelPercentSize.width).toBeCloseTo(0.25, 1);
     }
   });
-
-  test('can be duplicated', async ({ dashboardPage }) => {
```
juliepagano, line 145: This test is moved to the separate `duplicatePanels` tests.

ui/dashboards/src/utils/panelUtils.ts
```diff
@@ -26,6 +27,130 @@ export function getYForNewRow(group: PanelGroupDefinition) {
...
+        y: referenceBounds.y1,
+      });
+
+      return [...aboveInsertRow, ...insertRow, ...belowInsertRow];
+    }
+  }
+
+  // Insert the new item below the original and shift the items below the
+  // row where the reference is located.
+  return [
```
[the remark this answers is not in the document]
sjcobb: While I'm bringing up edge cases, this is completely unrelated to your PR, but I wonder if we should be setting a min width and min height? Currently, you can resize the panel down to a tiny, unusable size: 
image

ui/app/src/components/Footer.tsx
sjcobb: Nice, TIL how to hide elements that could break our visual regression tests

ui/dashboards/src/utils/panelUtils.ts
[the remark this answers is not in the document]
juliepagano: > One edge case I found is in the attached recording, where there is room to the right, but the panel is placed underneath. I don't think it's important to address though, I was intentionally trying weird layouts where panels weren't quite aligned--when using the dashboard normally, it works as expected.

In that case, it assumes the panel to the right of it is in the same row because it has vertical overlap, so it doesn't see space to the right and falls back to going below. I agree with your assessment that it's probably not worth doing something different there. It would increase the complexity of the algorithm without delivering a significantly better UX.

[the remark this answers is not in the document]
juliepagano: > While I'm bringing up edge cases, this is completely unrelated to your PR, but I wonder if we should be setting a min width and min height? Currently, you can resize the panel down to a tiny, unusable size: image

Good call. I'll make a shortcut ticket for this, so we can prioritize fixing it.

ui/dashboards/src/utils/panelUtils.ts
sjcobb: Yeah I like this, another possible use case is if we add custom actions like 'Add to Native Dashboard' from other parts of our product internally.

sjcobb (approved):
LGTM! 🚀

ui/dashboards/src/utils/panelUtils.ts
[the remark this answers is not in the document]
juliepagano: Created https://app.shortcut.com/chronosphere/story/43552/implement-min-width-height-for-panels-in-dashboard?stories_sort_by=priority&stories_group_by=workflow_state_id

PR #955, opened 2023-02-07
[IGNORE] Use specified playwright viewport

juliepagano:
Previously, a global viewport was set, but would be overridden in the chrome config, so we weren't using the intended value. After this fix, the intended viewport is used.

# Checklist

### DIFF
--- a/ui/e2e/src/config/base.playwright.config.ts
+++ b/ui/e2e/src/config/base.playwright.config.ts
@@ -53,11 +53,6 @@ const config: PlaywrightTestConfig = {
      */
     serviceWorkers: 'block',
 
-    // Keep this in sync with the setting in `.happo.js` for consistency for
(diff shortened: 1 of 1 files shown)


ui/e2e/src/config/base.playwright.config.ts
```diff
@@ -69,6 +64,10 @@ const config: PlaywrightTestConfig = {
       name: 'chromium',
       use: {
         ...devices['Desktop Chrome'],
```
juliepagano, line 66: Caught this while working on another PR. This spread was overriding the global defaults, which wasn't obvious -- there were a few github issues on playwright where people hit the same issue.

For now, inlining the correct value in the chrome config. If we ever expand to multiple browsers, we'll likely set it as a constant that's used across them.

sjcobb (approved):

PR #961, opened 2023-02-14
[FEATURE] Support mysql for the main DB

Nexucis:
This PR is adding the code required to support a MySQL database to store the dashboard.

It also introduced an init stage in the `DAO` interface. This method is called in the `main` and should be used to initialise the database. With MySQL the init step is the creation of the necessary tables. Like that the SRE that deployed `Perses` doesn't have to run a `sql` script to prepare the database.

Here we really have a basic usage of mysql. We have one table per object. Each object have a column `id`, `name`, `project` when it's a project resource and a column `doc` that contains the full JSON document.

For the moment there is no relation between the different tables. That might change in the future. 

Since I don't want to write manually every SQL request to insert, update, get data from the DB, I'm using the lib [go-sqlbuilder](https://github.com/huandu/go-sqlbuilder) that contains different builder for the different kind of request we need.

I'm slightly modifying the integration tests that used the File DB to run. Now the integrations tests are running twice, once with the File DB, and another time with MySQL running.


ui/app/src/views/ViewDashboardList.tsx
```diff
@@ -57,7 +57,7 @@ function RenderDashboardList() {
     return <CircularProgress />;
   }
 
-  if (data === undefined) return null;
+  if (data === undefined || data === null) return null;
```
sjcobb, line 60: Ah is this what was breaking your local dev environment?

Nexucis: yeah :). But it was because the API didn't return an empty array, but returned `null` instead. It's fixed now.
I kept these modifications for safety.

internal/api/shared/database/sql/query.go
```diff
@@ -0,0 +1,124 @@
...
+	sql, args := builder.Build()
+	return sql, args, nil
+}
+
+func generateProjectResourceSelectQuery(tableName string, project string, name string) (string, []interface{}) {
+	queryBuilder := sqlbuilder.NewSelectBuilder().
+		Select(colDoc).
+		From(tableName)
+	if len(name) > 0 {
+		queryBuilder.Where(queryBuilder.Like(colName, fmt.Sprintf("%s%%", name)))
```
AntoineThebaud, line 88: Maybe I'm missing context here but shouldn't it be `%%s%` instead ? to generate something like `WHERE col LIKE '%My Dashboard%'`, otherwise here it would give `WHERE col LIKE 'My Dashboard%%'`

internal/api/shared/database/sql/query.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return queryBuilder.Build()
+}
+
+func generateResourceSelectQuery(tableName string, name string) (string, []interface{}) {
+	queryBuilder := sqlbuilder.NewSelectBuilder().
+		Select(colDoc).
+		From(tableName)
+	if len(name) > 0 {
+		queryBuilder.Where(queryBuilder.Like(colName, fmt.Sprintf("%s%%", name)))
```
AntoineThebaud, line 101: same as above

AntoineThebaud (approved):

dev/docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
+version: '3.1'
+
+services:
```
sjcobb, line 3: Any documentation need to be updated? Mostly thinking about local dev, is it as simple as `docker-compose up -d` to use MySQL, but the file system is used like it’s currently if there isn’t a container running?

sjcobb (approved):
🚀

dev/docker-compose.yaml
Nexucis: so I didn't change anything regarding the usual dev environment.

In case you would like to use the mysql database, yeah you will have to run the `docker-compose` command and then to use the config file `config-mariadb.yaml` when running the `perses` binary. 

Do you think it should be documented somewhere @sjcobb ?

sjcobb: Gotcha, that makes sense! Could maybe update this [docs/configuration.md](https://github.com/perses/perses/blob/main/docs/configuration.md) file if it's referring to the same config

PR #968, opened 2023-02-17
[FEATURE] Ability to edit thresholds in panel options for gauge chart

eunicorn:
thresholds-editor

**CHANGES**
- You can now define default thresholds colors and add color palette in charts theme. 
- You can add, update, and delete thresholds
- Implemented a color picker using [react-colorful](https://omgovich.github.io/react-colorful/)
- Added `percentage` mode 

**TO DO**
- tests
- add thresholds editor to other charts
- add support for `min` ?


ui/components/package.json
```diff
@@ -43,6 +43,7 @@
     "lodash-es": "^4.17.21",
     "mathjs": "^10.6.4",
     "mdi-material-ui": "^7.4.0",
+    "react-colorful": "^5.6.1",
```
eunicorn, line 46: I was choosing between either [react-color](https://github.com/casesandberg/react-color) or  [react-colorful](https://github.com/omgovich/react-colorful), the two most popular color picker libraries.

I ended up choosing `react-colorful` because the design is simple (@foxbat07 approved), package is smaller, and it is built with react hooks and functional components. It ended up being very easy to use :)

ui/panels-plugin/src/model/thresholds.ts
```diff
@@ -11,53 +11,38 @@
...
-
-export const ThresholdColorsPalette = [ThresholdColors.ORANGE, ThresholdColors.RED];
-
-export type ThresholdColorsType = keyof typeof ThresholdColors | string;
-
 export type GaugeColorStop = [number, string];
 
 export type EChartsAxisLineColors = GaugeColorStop[];
 
-export interface StepOptions {
```
eunicorn, line 33: moved `StepOptions` and `ThresholdOptions` to plugin-system since I needed these types in `ThresholdsEditor` which lives in plugin-system/components

ui/panels-plugin/src/model/thresholds.ts
```diff
@@ -11,53 +11,38 @@
...
+  const defaultThresholdColor = thresholds.default_color ?? palette.defaultColor;
   const defaultThresholdSteps: EChartsAxisLineColors = [[0, defaultThresholdColor]];
 
   if (thresholds.steps !== undefined) {
     // https://echarts.apache.org/en/option.html#series-gauge.axisLine.lineStyle.color
     // color segments must be decimal between 0 and 1
     const segmentMax = 1;
     const valuesArr: number[] = thresholds.steps.map((step: StepOptions) => {
+      if (thresholds.mode === 'percentage') {
+        return step.value / 100;
```
eunicorn, line 39: we will have to update this once we support `min`

ui/plugin-system/src/model/thresholds.ts
```diff
@@ -0,0 +1,12 @@
+export interface StepOptions {
+  value: number;
+  color?: string;
+  name?: string;
+}
+
+export interface ThresholdOptions {
+  mode?: 'percentage' | 'absolute';
```
eunicorn, line 8: Debating if i should make this a boolean instead. 

For now the default is 'absolute'

ui/panels-plugin/src/model/thresholds.ts
sjcobb: Another option could be to move ThresholdsEditor to @perses-dev/components and the types to @perses-dev/core.

We definitely need to decide on long term rules for what goes in each package since currently it isn't consistent (see [discussion here](https://github.com/perses/perses/pull/761#discussion_r1022875646): LegendOptionsEditor, UnitSelector, unit types are in @perses-dev/components for instance). Probably fine to wait until @juliepagano looks into package reorganization though too

schemas/common/threshold.cue
```diff
@@ -20,6 +20,7 @@ package common
 }
 
 #thresholds: {
+	mode?:          string
```
sjcobb, line 23: Could be good to define the supported modes here (similar to [calculation.cue](https://github.com/perses/perses/blob/main/schemas/common/calculation.cue)), so something like: `mode?:  "Absolute" | "Percent"`

ui/plugin-system/src/model/thresholds.ts
sjcobb: I think not doing boolean in this case is okay since the strings are more descriptive. Main small thing is we've been using PascalCase for enum values like this and maybe just 'Percent' instead of 'Percentage'?

ui/panels-plugin/src/model/thresholds.ts
sjcobb: I think ThresholdsEditor being in @perses-dev/components will also make it easier to have context-aware components in the future. If it lives in plugin-system, then ColorPicker.tsx would need to move to avoid a circular dependency for instance

ui/panels-plugin/src/model/thresholds.ts
```diff
@@ -11,53 +11,38 @@
...
 
-// TODO (sjcobb): pull threshold colors from perses charts theme
-export const ThresholdColors = {
-  GREEN: 'rgba(47, 191, 114, 1)', // green.500
-  YELLOW: 'rgba(255, 193, 7, 1)',
-  ORANGE: 'rgba(255, 159, 28, 0.9)', // orange.500
-  RED: 'rgba(234, 71, 71, 1)', // red.500
-};
-
-export const ThresholdColorsPalette = [ThresholdColors.ORANGE, ThresholdColors.RED];
```
sjcobb, line 25: I really like that threshold colors moved to the chartsTheme, but one difference I'm seeing is previously orange and red were the defaults if a step color wasn't set explicitly, now it shows as blue for ex: 

image

ui/panels-plugin/src/model/thresholds.ts
eunicorn: i think i was debating to move it to @perses-dev/components but got confused since some components like CalculationSelector are in plugin-system. Can't wait to get this figure out lol.

Anyways moved ThresholdsEditor to @perses-dev/component 😄

ui/components/src/ThresholdsEditor/ThresholdColorPicker.tsx
```diff
@@ -0,0 +1,77 @@
...
+  color,
+  onColorChange,
+  label,
+}: Pick<ThresholdInputProps, 'color' | 'onColorChange' | 'label'>) {
+  const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(null);
+  const isOpen = Boolean(anchorEl);
+
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
```
cndonovan, line 31: If felt like it, could do something like `openColorPicker` and `closeColorPicker` (and both internally call the `setAnchorEl`) (if I'm reading right, I think that's the main outcome for `setAnchorEl`)

ui/components/src/ThresholdsEditor/ThresholdColorPicker.tsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+}
+
+const ColorIconButton = styled(IconButton, {
+  shouldForwardProp: (props) => props !== 'isSelected' && props !== 'iconColor',
+})<{
+  iconColor?: string;
+  isSelected?: boolean;
+}>(({ iconColor, isSelected }) => ({
+  backgroundColor: isSelected && iconColor ? `${iconColor}3F` : 'undefined', // 3F represents 25% opacity
```
cndonovan, line 75: Pretttty

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,236 @@
...
+export function ThresholdsEditor({ thresholds, onChange }: ThresholdsEditorProps) {
+  const {
+    thresholds: { defaultColor, palette },
+  } = useChartsTheme();
+  const defaultThresholdColor = thresholds?.default_color ?? defaultColor;
+
+  const [steps, setSteps] = useState(thresholds?.steps);
+  useEffect(() => {
+    setSteps(thresholds?.steps);
+  }, [thresholds?.steps]);
```
cndonovan, line 40: Noticed that we're using `palette` in other places (vs `steps`). Is `palette` like a simplified version of the same data in steps?

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  // every time a new threshold is added, we want to focus the recently added input
+  const recentlyAddedInputRef = useRef<HTMLInputElement | null>(null);
+  const focusRef = useRef(false);
+  useEffect(() => {
+    if (!recentlyAddedInputRef.current || !focusRef.current) return;
+    recentlyAddedInputRef.current?.focus();
+    focusRef.current = false;
+  }, [steps?.length]);
```
cndonovan, line 49: Might not be possible, but is there a way to call `.focus()` from inside the `addThresholdChange` handler and get it to work?

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,236 @@
...
+    } else {
+      onChange(
+        produce(thresholds, (draft) => {
+          const steps = draft.steps;
+          if (steps?.length) {
+            const lastStep = steps[steps.length - 1];
+            const color = palette[steps.length] ?? getRandomColor(); // we will assign color from the palette first, then generate random color
+            steps.push({ color, value: (lastStep?.value ?? 0) + 10 });
+          } else if (steps) {
+            steps.push({ value: 10 });
```
cndonovan, line 128: Is this saying that any newly added threshold defaults to a value of `10`?

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,236 @@
...
+          if (draft.steps !== undefined) {
+            const step = draft.steps[i];
+            if (step) {
+              step.color = color;
+            }
+          }
+        })
+      );
+    }
+  };
```
cndonovan, line 75: I was wondering what the difference between `handleThresholdChange` vs `handleThresholdColorChange` was... I thought maybe `handleThresholdChange` was changing the entire object. Could do `handleThresholdValueChange` if that seems like an improvement 🤷‍♀️

cndonovan (commented):
This is so nice, yay for our users

ui/components/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,85 @@
...
+    setColor(color);
+    setValue(color);
+    onColorChange && onColorChange(color);
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const inputValue = e.target.value.replace(/([^0-9A-F]+)/gi, '').substring(0, 8);
+    setValue(`#${inputValue}`);
+    // only set color if input value is a valid hex color
+    if (isValidHex(e.target.value)) {
```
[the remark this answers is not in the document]
eunicorn: you're correct - i'll add comments in the code to make this more clear :)

ui/components/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,85 @@
...
+              <CircleIcon />
+            </IconButton>
+          ))}
+      </Stack>
+      <TextField
+        inputProps={{ 'aria-label': 'enter hex color' }}
+        fullWidth
+        value={value}
+        onChange={handleInputChange}
+      />
```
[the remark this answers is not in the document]
eunicorn: right now it always prefixes the input with a pound sign to indicate that it's hex format and i don't think it should let you enter parentheses and commas

ui/components/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,85 @@
...
+    </Stack>
+  );
+};
+
+const isValidHex = (value: string, alpha?: boolean): boolean => {
```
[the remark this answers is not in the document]
eunicorn: not yet but i actually got this function from the library. It currently only exports an input component with the validation logic, but there is an open issue to support custom input components https://github.com/omgovich/react-colorful/issues/157

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: `palette` is the color palette we'll use in the chart and `steps` are the actual values.  

for example, if steps[0] is 10 and palette[0] is yellow, we'll show yellow for values greater than or equal to 10

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: if there is no previous threshold set, then the first threshold you add will default to `10`

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: Hmm i don't think so? We don't have a ref to the text input yet since it hasn't been rendered until `steps` is updated

juliepagano:
UX feedback for a potential follow-up with @foxbat07. I find the top bar that communicates the ranges a little overwhelming to process quickly in some cases (e.g. it's a lot when there's a bunch of gauges right next to each other) because there's so many colors. I wonder if it might be helpful to add an option where you can disable them if you like?

image

juliepagano:
I noticed that the button toggle highlights the "%" when it's selected...
image

...but not the "#." Is this a bug?
image

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,238 @@
...
+      title="Thresholds"
+      icon={
+        <IconButton size="small" aria-label="add threshold" onClick={addThresholdInput}>
+          <PlusIcon />
+        </IconButton>
+      }
+    >
+      <ToggleButtonGroup
+        exclusive
+        value={thresholds?.mode ?? 'absolute'}
```
juliepagano, line 158: Should this be `Absolute` to match the value below? I _think_ that might be why the button doesn't show selected in the screenshots I included as a top-level comment.

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,238 @@
...
+    if (thresholds !== undefined) {
+      const updatedThresholds = produce(thresholds, (draft) => {
+        if (draft.steps) {
+          draft.steps.splice(i, 1);
+        }
+      });
+      onChange(updatedThresholds);
+    }
+  };
+
```
juliepagano, line 114: Optional suggestion -- You've got a decent amount of logic around adjusting these values that's probably a bit of a pain to test in the context of a react component. If you pulled the core logic out into a few utility functions that you call from the component, you could more easily write some unit tests on those functions outside the react context. That said, you've already written a bunch of tests for the components, so may not be worth refactoring now.

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -0,0 +1,238 @@
...
+          ))
+          .reverse()}
+      <Stack flex={1} direction="row" alignItems="center" spacing={1}>
+        <ThresholdColorPicker label="default" color={defaultThresholdColor} onColorChange={handleDefaultColorChange} />
+        <Typography>Default</Typography>
+      </Stack>
+    </OptionsEditorGroup>
+  );
+}
+
```
juliepagano, line 196: This is already a relatively big file. Is it work pulling the `ThresholdInput` component out into a separate file?

juliepagano:
If you have the time, I think this would be a good case to add some additional visual tests for the gauge chart. Let me know if you want help doing so.

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: thanks for catching this!

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: oh nice i'll keep that in mind for next time, and writing those tests in react context actually helped me uncover a few accessibility issues

eunicorn:
> If you have the time, I think this would be a good case to add some additional visual tests for the gauge chart. Let me know if you want help doing so.

yes definitely! i am hoping to do this as one of the follow ups since this PR is already so big

juliepagano (approved):
LGTM! 🚀

juliepagano:
FYI, in case you haven't done it any more, you can remove the ❌  next to "happo" by clicking "details" and selecting "approve" from the review dropdown.

image

eunicorn:
> FYI, in case you haven't done it before, you can remove the ❌ next to "happo" by clicking "details" and selecting "approve" from the review dropdown.
> 

probably a dumb question 😅 , but where is this "happo" button?

ui/components/src/OptionsEditorLayout/OptionsEditorGroup.tsx
```diff
@@ -12,31 +12,30 @@
...
-      <Typography
-        variant="overline"
-        component="h4"
-        borderBottom={1}
-        borderColor="grey.300"
-        marginBottom={(theme) => theme.spacing(1)}
-      >
-        {title}
-      </Typography>
+    <>
```
saminzadeh, line 31: Can we still wrap this in a `<Box>` to group the components together? It can cause unintended side-effects when using standalone similar to our [TimeRangeControls](https://github.com/perses/perses/pull/974).

ui/components/src/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,88 @@
...
+// limitations under the License.
+
+import { IconButton, Stack, TextField } from '@mui/material';
+import React, { useState } from 'react';
+import { HexColorPicker } from 'react-colorful';
+import CircleIcon from 'mdi-material-ui/Circle';
+
+interface ColorPickerProps {
+  initialColor?: string;
+  onColorChange?: (color: string) => void;
```
saminzadeh, line 21: What is the purpose of `onColorChange`? Seems like `initialColor` means that component is uncontrolled so I wonder if this can be misleading. Otherwise, it might make sense to use the controlled component paradigm with `value` and `onChange`

ui/components/src/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
cndonovan: sweet, sg

ui/components/src/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
cndonovan: oo yes the `#` prefix does exactly that!

ui/components/src/ColorPicker/ColorPicker.tsx
eunicorn: good call, replaced `initialColor` with `color and `onColorChange` to `onChange`

sjcobb (approved):
🚀

PR #974, opened 2023-02-21
[ENHANCEMENT ] Fix styling of TimeRangeControls

saminzadeh:
This fixes the TimeRangeControls to wrap the component in a flex component with a spacing of 1. Without a wrapper, the component did not look great standalone. It also adds the ability to hide the refresh button.

TODO:
There is still some work to make this TimeRangeControls component composable for a majority of usec cases. Some examples:
- ability to pass custom time ranges
- Remove dependency of DashboardContext

# Checklist

### DIFF
--- a/ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
+++ b/ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 import RefreshIcon from 'mdi-material-ui/Refresh';
+import { Stack } from '@mui/material';
 import { DateTimeRangePicker, InfoTooltip, TimeOption } from '@perses-dev/components';
 import { useTimeRange } from '@perses-dev/plugin-system';
 import { isDurationString } from '@perses-dev/core';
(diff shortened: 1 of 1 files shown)


eunicorn (approved):

ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -36,10 +37,14 @@ const DEFAULT_HEIGHT = '34px';
...
   heightPx?: number;
+
+  // Whether to show the refresh button or not
+  showRefresh?: boolean;
 }
 
-export function TimeRangeControls({ heightPx }: TimeRangeControlsProps) {
+export function TimeRangeControls({ heightPx, showRefresh = true }: TimeRangeControlsProps) {
   const { timeRange, setTimeRange, refresh } = useTimeRange();
+  // TODO: Remove this since it couples to the dashboard context
```
saminzadeh, line 47: I see that we use this to add to add a timeOption if the dashboard spec has a pastDuration that is not available in our list of options. Ideally, we can remove this and find a better solution than plugging into the dashboard context.

sjcobb: I remember adding this now, it is important though since I know Mohit wanted to trim the list of options and there is a high chance a duration isn’t in our default options. But agreed, this could be accounted for outside of TimeRangeControls and passed in

sjcobb (approved):
LGTM thanks for fix this!

PR #976, opened 2023-02-22
[BUGFIX] Do not include tests in compiled output

juliepagano:
- Also adds separate `type-check` script and job in CI.

# Notes for reviewers

- Prior to this fix, we'd include jest tests (`.test.` files) in our compiled output, which adds unnecessary extra content for consumers that isn't doing anything useful for them. Modifying our process to exclude them.
- This does not comprehensively remove test-related files. I think there are a few utils that still get compiled. I'm intentionally not trying to remove them right now.
- Doing this ahead of my storybook work, where we'll also want to exclude `.stories.` files.
- I think this should be a noop for consumers (besides having a few less bytes to download), but _please_ let me know if I missed anything.

## Validating the files were excluded
- `cd ui`
- `npm run build`
- `find */dist  | grep '\.test\.'` to list test files
- `find */dist  | grep '\.test\.' | wc -l` to count test files

### Before
image

### After
image

# Checklist


ui/components/package.json
```diff
@@ -24,7 +24,7 @@
     "build": "concurrently \"npm:build:*\"",
     "build:cjs": "swc ./src -d dist/cjs --config-file ../.cjs.swcrc",
     "build:esm": "swc ./src -d dist --config-file ../.swcrc",
-    "build:types": "tsc --emitDeclarationOnly --declaration --preserveWatchOutput",
+    "build:types": "tsc --project tsconfig.build.json",
```
juliepagano, line 27: You [can't exclude from the CLI](https://github.com/microsoft/TypeScript/issues/46005) and we don't want to exclude typescripting these files during dev, so I needed to make a separate `tsconfig` file for the build step. Pulling all the build config into that file so it's in one place.

ui/components/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.json",
```
juliepagano, line 2: You can't extend from multiple tsconfig files, so erring on the side of importing from the local tsconfig and adding the build options (instead of making a global build options tsconfig and extending from that).

juliepagano:
Tagging @LukeTillman on this one because I think he was who originally set up the build system and might be most likely to catch anything I'm doing wrong here.

LukeTillman (approved):
Nice! This has been on my `TODO` list forever, so thanks for fixing it. 🥳 

FWIW, I _think_ the `tsconfig.json` file in each project is only used by `tsc` for the type declarations output (no other tooling, including Jest uses it), so I think you could probably make this change in some sort of central place rather than having to duplicate it everywhere. But this is totally cool too.

juliepagano:
> Nice! This has been on my `TODO` list forever, so thanks for fixing it. 🥳
> 
> FWIW, I _think_ the `tsconfig.json` file in each project is only used by `tsc` for the type declarations output (no other tooling, including Jest uses it), so I think you could probably make this change in some sort of central place rather than having to duplicate it everywhere. But this is totally cool too.

If I recall correctly, VSCode will also use the closest parent tsconfig.json file, so it's likely using them. I'll leave it as is for now, and maybe take a careful pass at seeing if some of it can be consolidated in the future.

LukeTillman:
One other random thing I just thought of: I think a side-effect of this change is that test files will no longer be type-checked. Not sure if that matters to folks, but just thought I should call it out.

juliepagano:
> 

I've been wondering if we should add a separate "type-check" task that just type-checks everything without compiling out the types. Lemme see if that would be easy to set up while I'm in here.

juliepagano:
> > 
> 
> I've been wondering if we should add a separate "type-check" task that just type-checks everything without compiling out the types. Lemme see if that would be easy to set up while I'm in here.

Done! @LukeTillman could you give the changes to add the `type-check` a quick pass to make sure that looks ok.

LukeTillman (approved):
LGTM. One minor nit: if you want to omit `type-check` from the `app` workspace, you probably could since the webpack build takes care of it in a separate process. But fine either way.

PR #977, opened 2023-02-22
Sync Release v0.23 with main 

sjcobb:
Sync the minor release branch back into main includes tags for v0.23.0 and v0.23.1

- Patch release included a fix in #972 for console warnings we found during integration


sjcobb:
@Nexucis I went ahead and force merged while I was thinking about it since it was green / no merge conflicts

Nexucis:
yeah I think it's fair to do that @sjcobb since it's quite straight forward PR :)

PR #973, opened 2023-02-21
WIP: Initial storybook setup - v6.5

juliepagano:
# Checklist

- [ ] Pull request has a descriptive title and context useful to a reviewer.
- [ ] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [ ] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
- [ ] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.

# TODO
- [ ] Extract changes to exclude tests from output into a separate PR. Then this one can just add stories to that setup. https://github.com/perses/perses/pull/976

PR #979, opened 2023-02-23
[IGNORE] Initial Storybook v7 beta setup

juliepagano:
This is some basic storybook framework setup. See the TODO section for follow-ups. Sorry this took so long to set up -- lots of complexity with the combo of the monorepo, our theming, and looking at the v7 beta.

# Notes for reviewers

- Please kick the tires on this on your local machine by trying to run the storybook locally. I _think_ all the setup I've done works durably because storybook builds in CI, but I want to be extra sure some of the more complex bits work locally on more machines than mine before we merge this. Some things to check:
  - You can run storybook
  - Stories load
  - Stories for components include a populated docs table (see screenshot for what it should look like)
- I left a bunch of inline comments explaining the approach and pointing out areas where I am most interested in feedback.
# Checklist

# TODO
- [x] Look into why ui-libs artifact from CI includes storybook. Probably something to adjust slightly.

Things I'm going to do in follow-ups:
- [ ] Figuring out the right setup of `react-docgen-typescript` config options. Leaving the defaults for now, but it may be apparent we need to adjust as we add more complex stories with more complex types.
- [ ] Figure out getting Happo working with storybook. I did a quick spike to make sure I could get it doing some basics with v7, but there's a bunch of other config/setup needed that should get done separate from this PR.
- [ ] More fully fleshed out example stories. Right now, this is suuuuper bare bones to make sure things work.
- [ ] Publishing storybook to github pages. It'll make sense to do this in a while when the storybook is more fleshed out.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,7 @@ jobs:
           path: |
             ui/**/dist
             !ui/node_modules
+            !ui/storybook
 
   build-storybook:
     name: "build-storybook"
@@ -69,8 +70,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: storybook
-          path: |
-            ui/storybook/storybook-static
+          path: ui/storybook/storybook-static
   
   libs-release:
     name: "libs-release"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,6 +52,26 @@ jobs:
             ui/**/dist
             !ui/node_modules
 
+  build-storybook:
+    name: "build-storybook"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - uses: ./.github/actions/setup_environment
+        with:
+          enable_npm: true
+      - name: Install UI deps
+        run: cd ./ui && npm install
+      - name: Build the storybook
+        run: cd ./ui && npm run storybook:build
+      - name: store storybook build
+        uses: actions/upload-artifact@v3
+        with:
+          name: storybook
+          path: |
+            ui/storybook/storybook-static
+  
   libs-release:
     name: "libs-release"
     needs: "build-frontend"
--- a/scripts/ui_release.sh
+++ b/scripts/ui_release.sh
@@ -10,8 +10,10 @@ cd ui/
 files=("../LICENSE" "../CHANGELOG.md")
 workspaces=$(jq -r '.workspaces[]' < package.json)
 publish_workspaces=$(for workspace in ${workspaces}; do 
-  # package "app" is private so we shouldn't try to publish it.
-  if [[ "${workspace}" != "app" ]] && [[ "${workspace}" != "e2e" ]]; then
+  # packages are private so we shouldn't try to publish them.
+  # TODO: see if we can do something smarter here by checking for "private" set
+  # to true in a given package's package.json.
+  if [[ "${workspace}" != "app" ]] && [[ "${workspace}" != "e2e" ]] && [[ "${workspace}" != "storybook" ]]; then
     echo $workspace;
   fi 
 done)
--- a/ui/.eslintrc.base.js
+++ b/ui/.eslintrc.base.js
@@ -20,6 +20,9 @@ module.exports = {
     'plugin:jsx-a11y/recommended',
     'plugin:@typescript-eslint/recommended',
     'plugin:prettier/recommended',
+
+    // Only impacts stories following *.stories.* naming by default.
+    'plugin:storybook/recommended',
   ],
 
   plugins: ['import'],
--- a/ui/components/src/EChart.tsx
+++ b/ui/components/src/EChart.tsx
@@ -50,7 +50,7 @@ const mouseEvents = [
   'contextmenu',
 ] as const;
 
-export type MouseEventName = typeof mouseEvents[number];
+export type MouseEventName = (typeof mouseEvents)[number];
 
 // batch event types
 export interface DataZoomPayloadBatchItem {
@@ -84,7 +84,7 @@ type OnBatchEventFunction = (params: BatchEventsParameters) => void;
 
 const batchEvents = ['datazoom', 'downplay', 'highlight'] as const;
 
-export type BatchEventName = typeof batchEvents[number];
+export type BatchEventName = (typeof batchEvents)[number];
 
 type ChartEventName = 'finished';
 
--- a/ui/components/src/LineChart/LineChart.stories.tsx
+++ b/ui/components/src/LineChart/LineChart.stories.tsx
@@ -11,14 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React from 'react';
-
-import { StoryFn, Meta } from '@storybook/react';
+import { StoryObj, Meta } from '@storybook/react';
 import { LineChart } from '@perses-dev/components';
 
-// More on default export: https://storybook.js.org/docs/react/writing-stories/introduction#default-export
-export default {
-  title: 'LineChart',
+const meta: Meta<typeof LineChart> = {
   component: LineChart,
   args: {
     height: 200,
@@ -54,11 +50,10 @@ export default {
       bottom: 0,
     },
   },
-} as Meta<typeof LineChart>;
-
-// More on component templates: https://storybook.js.org/docs/react/writing-stories/introduction#using-args
-const Template: StoryFn<typeof LineChart> = (args) => {
-  return <LineChart {...args} />;
 };
 
-export const Primary = Template.bind({});
+export default meta;
+
+type Story = StoryObj<typeof LineChart>;
+
+export const Primary: Story = {};
--- a/ui/components/src/LineChart/LineChart.stories.tsx
+++ b/ui/components/src/LineChart/LineChart.stories.tsx
@@ -1,0 +1,64 @@
+// Copyright 2023 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react';
+
+import { StoryFn, Meta } from '@storybook/react';
+import { LineChart } from '@perses-dev/components';
+
+// More on default export: https://storybook.js.org/docs/react/writing-stories/introduction#default-export
+export default {
+  title: 'LineChart',
+  component: LineChart,
+  args: {
+    height: 200,
+    data: {
+      timeSeries: [
+        {
+          type: 'line' as const,
+          name: 'up{instance="demo.do.prometheus.io:3000",job="grafana"}',
+          data: [1, 1, 1],
+          color: 'hsla(158782136,50%,50%,0.8)',
+          sampling: 'lttb' as const,
+          progressiveThreshold: 1000,
+          symbolSize: 4,
+          lineStyle: { width: 1.5 },
+          emphasis: { lineStyle: { width: 2.5 } },
+        },
+      ],
+      xAxis: [1673784000000, 1673784060000, 1673784120000],
+      legendItems: [],
+      rangeMs: 21600000,
+    },
+    yAxis: {
+      show: true,
+    },
+    unit: {
+      kind: 'Decimal' as const,
+      decimal_places: 2,
+      abbreviate: true,
+    },
+    grid: {
+      left: 20,
+      right: 20,
+      bottom: 0,
+    },
+  },
+} as Meta<typeof LineChart>;
+
+// More on component templates: https://storybook.js.org/docs/react/writing-stories/introduction#using-args
+const Template: StoryFn<typeof LineChart> = (args) => {
+  return <LineChart {...args} />;
+};
+
+export const Primary = Template.bind({});
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { MouseEvent, useMemo, useRef, useState } from 'react';
+import React, { MouseEvent, useMemo, useRef, useState } from 'react';
 import { Box } from '@mui/material';
 import type {
   EChartsCoreOption,
@@ -56,7 +56,10 @@ use([
   CanvasRenderer,
 ]);
 
-interface LineChartProps {
+export interface LineChartProps {
+  /**
+   * Height of the chart
+   */
   height: number;
   data: EChartsDataFormat;
   yAxis?: YAXisComponentOption;
--- a/ui/components/src/stories/About.stories.mdx
+++ b/ui/components/src/stories/About.stories.mdx
@@ -1,0 +1,7 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="About" />
+
+# Components
+
+This Storybook is currently a work-in-progress! Stay tuned for more information soon.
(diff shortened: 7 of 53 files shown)


.github/workflows/ci.yaml
```diff
@@ -52,6 +52,26 @@ jobs:
             ui/**/dist
             !ui/node_modules
 
+  build-storybook:
```
juliepagano, line 55: Currently, this builds storybook and outputs it as an artifact. It'll be nice to publish it to github pages in the future, but I want to tackle that in a follow-up, ideally after we've fleshed out more stories.

scripts/ui_release.sh
```diff
@@ -10,8 +10,10 @@ cd ui/
 files=("../LICENSE" "../CHANGELOG.md")
 workspaces=$(jq -r '.workspaces[]' < package.json)
 publish_workspaces=$(for workspace in ${workspaces}; do 
-  # package "app" is private so we shouldn't try to publish it.
-  if [[ "${workspace}" != "app" ]] && [[ "${workspace}" != "e2e" ]]; then
+  # packages are private so we shouldn't try to publish them.
+  # TODO: see if we can do something smarter here by checking for "private" set
+  # to true in a given package's package.json.
```
juliepagano, line 15: Part of me is really tempted to do this and rewrite this as a node script because it would be a lot easier for UI folks to maintain than bash, but I'm not digging into it as part of this specific effort.

ui/.eslintrc.base.js
```diff
@@ -20,6 +20,9 @@ module.exports = {
     'plugin:jsx-a11y/recommended',
     'plugin:@typescript-eslint/recommended',
     'plugin:prettier/recommended',
+
+    // Only impacts stories following *.stories.* naming by default.
+    'plugin:storybook/recommended',
```
juliepagano, line 25: Starting out with their recommended rules, and we can adjust as needed.

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Not like this:
+
+```ts
+// Importing directly from the source file will make it easier to forget to
+// export components for consumption from the main export for the package.
+import { LineChart } from '.';
+```
+
+#### Long version
```
juliepagano, line 64: This section describes one of the most controversial things I'm doing in this setup. Please review and let me know if you really dislike these tradeoffs and have suggestions for other approaches.

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
+# Storybook
+
+This package is used to generate documentation for Perses UI components using [Storybook](https://storybook.js.org/).
+
+> This project currently uses the [v7 beta](https://storybook.js.org/docs/7.0/react/get-started/introduction) of Storybook. Be sure to reference the associated [v7 beta documentation](https://storybook.js.org/docs/7.0/react/get-started/introduction).
```
juliepagano, line 5: The v7 build is supposed to land soonish, it fixes a handful of issues I noticed when setting up v6.5 (current stable release) in https://github.com/perses/perses/pull/973, and it has better TS support. Considering all of this, I think it's worth us using the beta to start, so we don't have to deal with migrating a ton of stories to the new format. Let me know if you disagree with this.

See the migration doc if you're curious about all the changes:
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700

ui/components/src/EChart.tsx
```diff
@@ -50,7 +50,7 @@ const mouseEvents = [
   'contextmenu',
 ] as const;
 
-export type MouseEventName = typeof mouseEvents[number];
+export type MouseEventName = (typeof mouseEvents)[number];
```
juliepagano, line 53: One of the other deps installed led to an update to prettier, so there's a few minor changes like this in here. Seems harmless to me, but let me know if you think we need to pin prettier at a particular version.

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -56,7 +56,10 @@ use([
   CanvasRenderer,
 ]);
 
-interface LineChartProps {
+export interface LineChartProps {
+  /**
```
juliepagano, line 60: Added this to test out the docs table showing jsdoc.

ui/components/src/stories/About.stories.mdx
```diff
@@ -0,0 +1,7 @@
+import { Meta } from '@storybook/addon-docs';
+
```
juliepagano, line 2: These top-level "about" docs stories should be used to explain what each of the packages are for and how to use them. Not trying to fill these docs out as part of this PR, but putting them in place to show how to use them.

ui/components/tsconfig.build.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["**/*.test.*"],
+  "exclude": ["**/*.stories.*", "**/*.test.*"],
```
juliepagano, line 3: Do not publish stories with our compiled code.

ui/components/tsconfig.json
```diff
@@ -3,6 +3,9 @@
   "include": ["src"],
   "compilerOptions": {
     "outDir": "./dist",
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "paths": {
+      "@perses-dev/components": ["./src"]
```
juliepagano, line 8: Allows embedded stories to reference source code at the top level without thinking too hard about it. See the storybook README for a big explanation of why we're doing this.

ui/package.json
```diff
@@ -10,6 +10,10 @@
     "lint:fix": "turbo run lint -- --fix",
     "type-check": "turbo run type-check",
     "start": "turbo run start",
+    "start-with-storybook": "turbo run start storybook:start",
+    "storybook": "turbo run storybook:start",
+    "storybook:build": "turbo run storybook:build",
+    "storybook:serve": "turbo run storybook:serve",
```
juliepagano, line 16: Trying to think about some of this. Do folks want to always start both the storybook and the app or will they mostly want to start them separately?

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+- Run storybook in development mode: `npm run storybook`
+
+### Building storybook
+
+- Build only: `npm run storybook:build`
+- Build & serve static assets from build: `npm run storybook:serve`
+
+## Creating stories
+
+### Where should stories live?
```
juliepagano, line 20: How do folks feel about this? I think this approach makes it easier to remember to add stories because they're colocated like unit tests.

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+- [Dark Mode](https://storybook.js.org/addons/storybook-dark-mode/) - Adds a toolbar toggle that switches between light and dark mode.
+- [Docs](https://storybook.js.org/addons/@storybook/addon-docs/) - Document component usage and properties in Markdown.
+- [Links](https://storybook.js.org/addons/@storybook/addon-links/) - Use to create links that navigate between stories.
+- [Measure](https://storybook.js.org/addons/@storybook/addon-measure/) - Helpful for inspecting layouts by visualizing the box model.
+- [Outline](https://storybook.js.org/addons/@storybook/addon-outline/) - Outline all elements with CSS to help with layout placement and alignment.
+- [Storysource](https://storybook.js.org/addons/@storybook/addon-storysource/) - Used to show stories source in the addon panel
+- [Viewport](https://storybook.js.org/addons/@storybook/addon-viewport/) - Allows stories to be displayed in different sizes and layouts.
+
+We primarily use first party addons maintained by Storybook to avoid pain with upgrades and interoperability with the relatively complex Storybook ecosystem.
```
juliepagano, line 88: Thoughts?

ui/storybook/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "version": "0.22.0",
+  "private": true,
+  "scripts": {
+    "storybook:start": "storybook dev --config-dir src/config --port 6006",
+    "storybook:clean": "rimraf storybook-static/",
+    "storybook:build": "npm run storybook:clean && storybook build --config-dir src/config",
+    "storybook:serve": "npx http-server ./storybook-static",
+    "lint": "eslint src --ext .ts,.tsx",
+    "lint:fix": "npm run lint -- --fix",
+    "type-check": "tsc --noEmit"
```
juliepagano, line 12: Prefixing everything because the storybook is really different from our source code, and we don't want to run all of these steps in most cases where we're running things like "build" and it would be tedious to always remember to filter out the storybook workspace.

ui/components/src/LineChart/LineChart.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { StoryFn, Meta } from '@storybook/react';
+import { LineChart } from '@perses-dev/components';
+
+// More on default export: https://storybook.js.org/docs/react/writing-stories/introduction#default-export
+export default {
+  title: 'LineChart',
+  component: LineChart,
+  args: {
+    height: 200,
+    data: {
```
juliepagano, line 25: Doing a quick and dirty example here to get a chart working. Will likely want to extract this into something cleaner in follow-ups.

juliepagano:
The go mysql failure is unrelated to my changes. There seem to be some issues with that test being flaky.

ui/components/tsconfig.json
saminzadeh: This makes sense, basically bypass the build and go directly to the files as if it were part of the project.

ui/package.json
saminzadeh: Separately works now

ui/storybook/README.md
saminzadeh: I think we should move forward with it. I support using v7

ui/storybook/README.md
saminzadeh: +1 I think colocated makes sense for now

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Creating stories
+
+### Where should stories live?
+
+Stories **SHOULD** live in the first entry in the list below that matches their purpose:
+
+- Stories that are tightly coupled with a component should be colocated with the component code in the associated package (e.g. the stories for the `LineChart` component live at `ui/components/src/LineChart/LineChart.stories.tsx` as a sibling of `LineChart.tsx`).
+- Package-specific stories should live in the `src/stories` directory for that package.
+- Stories that provide project-wide documentation or involve components from multiple packages should live in the storybook package (`ui/storybook/stories`).
```
saminzadeh, line 26: +1 this makes sense

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+#### TLDR;
+
+Stories should import package code by referencing the package, not by making a relative reference. Under the hood, this will reference source code (not compiled code), so we can benefit from some Storybook features like autogenerating documentation from types.
+
+Import like this in a story:
+
+```ts
+// Import from the package. Storybook's webpack config will alias this to the
+// associated source code.
+import { LineChart } from '@perses-dev/components';
```
saminzadeh, line 53: +1

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+```ts
+// Importing directly from the source file will make it easier to forget to
+// export components for consumption from the main export for the package.
+import { LineChart } from '.';
+```
+
+#### Long version
+
+Stories should import from source code (NOT compiled code) because this enables us to automatically generate documentation tables for component props from types. Stories should import from the top level export for that package to help ensure the component was properly exported for consumption from the package when published. The same versions of code need to be used across the storybook to avoid issues when using `context` and singletons.
```
saminzadeh, line 66: It's a bummer we need to do this, but I think having the docs outweighs the downside by a lot. So I agree.

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+The project enforces the above contraints using the following tooling:
+
+- **aliases to source for all @perses-dev packages in storybook's webpack config** - E.g. `@perses-dev/components` will alias to `ui/components/src` instead of pointing to `ui/components/dist`). This makes it easy to import from packages in stories in a consistent way while pointing at the top level `src` export. It also ensures that components referenced indirectly in stories (usually a child of the component used directly in the story) point to source code to avoid accidentally mixing and matching source and compiled code, which can cause issues when using `context` and singletons.
```
saminzadeh, line 70: Somewhat unrelated but this is an approach I was thinking about as well when deciding how to do the codemirror packages for the prom fork. If we did this aliasing in the codemirror extension + lezer grammar, there would be no need to publish a separate package but it could still live in its own package folder. Still thinking about it. But for storybook, I think this makes sense.

saminzadeh (approved):
This is amazing, thank you for doing this Julie! 🚀 
Also, kudos to the detailed documentation and thorough thought you put into this

ui/storybook/README.md
cndonovan: I forget what the downside to doing it this way are (just that we're not testing our compiled code?)

cndonovan: Is it also sort of more abnormal / not the usual way to do it? (it's more common to import compiled code)?

ui/storybook/README.md
cndonovan: Seems good, I don't know of any non-first-party-addons that I even need (seems like we have the things we need)

ui/storybook/README.md
```diff
@@ -0,0 +1,122 @@
...
+- [Docs](https://storybook.js.org/addons/@storybook/addon-docs/) - Document component usage and properties in Markdown.
+- [Links](https://storybook.js.org/addons/@storybook/addon-links/) - Use to create links that navigate between stories.
+- [Measure](https://storybook.js.org/addons/@storybook/addon-measure/) - Helpful for inspecting layouts by visualizing the box model.
+- [Outline](https://storybook.js.org/addons/@storybook/addon-outline/) - Outline all elements with CSS to help with layout placement and alignment.
+- [Storysource](https://storybook.js.org/addons/@storybook/addon-storysource/) - Used to show stories source in the addon panel
+- [Viewport](https://storybook.js.org/addons/@storybook/addon-viewport/) - Allows stories to be displayed in different sizes and layouts.
+
+We primarily use first party addons maintained by Storybook to avoid pain with upgrades and interoperability with the relatively complex Storybook ecosystem.
+
+## Configuring Storybook
```
cndonovan, line 90: Dunno if it matters, but saw that sometimes we're using lowercase `storybook` and sometimes uppercase `Storybook` 🤷‍♀️  I feel like I'm getting the point, so I dunno if it matters (`storybook` might be more about the CLI tool)

ui/package.json
cndonovan: Separately is how I'd think about it... I'd run storybook when I need it (because I'm developing locally and updating stories)

The other way I use Storybook is for reference, and in that usage, I like the idea of having a permanent URL to login to and see the current state of Storybook

cndonovan (approved):

ui/storybook/README.md
juliepagano: Primarily that we're not testing our compiled code. There's also a bit of complexity in the webpack/ts config to do the aliasing. Because of this limitation of storybook, most people who write typescript point at their source code. I ran into a few discussions/issues/etc. asking about supporting compiled code with `.d.ts` files from a while back and there was little-to-no response, so I don't think there's a ton of appetite to support it.

ui/package.json
juliepagano: > The other way I use Storybook is for reference, and in that usage, I like the idea of having a permanent URL to login to and see the current state of Storybook

When we get to a place where we're happy with the results, I'm hoping we'll build and publish a static version of the storybook to github pages on `merge-to-master` for this purpose.

ui/components/src/LineChart/LineChart.stories.tsx
juliepagano: There's def some weirdness with some of the chart stuff that probably needs some adjusting. Something we'll want to sort out in follow-ups as we work more on the embeddability use cases.

ui/storybook/README.md
juliepagano: I _think_ you may also be able to do some magic with private packages in turborepo, but I haven't used it before. Definitely something we can play with a bit when we're ready to do that work.

ui/storybook/README.md
juliepagano: I think the naming was just me being lazy...lemme see if I can normalize it.

PR #981, opened 2023-02-23
[IGNORE] Bump go to v1.20.x

Nexucis:
I'm just bumping the Golang version to the last one


Nexucis:
the linter error is really weird. I need to take a deeper look at that ...

cndonovan:
@Nexucis is this a PR we could close for the time being while it's in draft mode? (It's OK if the answer is "no, it helps to have it open")

Nexucis:
yeah we can close it @cndonovan no issue on that :)

PR #993, opened 2023-02-28
[IGNORE] fix datasource docs examples

jgbernalp:
This PR fixes the example values to match the current API definitions


Nexucis (approved):
nice, thanks !

PR #962, opened 2023-02-15
[FEATURE] New `area_opacity` option in time series panel

sjcobb:
## Overview

Add optional `area_opacity` option to TimeSeriesChart panel. This includes a MUI Slider to customize the ECharts property [series-line.areaStyle.opacity](https://echarts.apache.org/en/option.html#series-line.areaStyle.opacity)

## TODOs For Follow Up PRs

- I started to add a visual regression test, but I'm planning to come back to it since I have some questions about mocking the range query response, see [branch here](https://github.com/perses/perses/compare/sjcobb/time-series-area-opacity-option...sjcobb/time-series-area-opacity-option-test) and [screenshot](https://github.com/perses/perses/pull/962#discussion_r1109928450)
- [Stacked area](https://echarts.apache.org/examples/en/editor.html?c=area-stack) support by adding new `stacking` visual option
  - schema will most likely be: 
```cue
#visual: {
	line_width?:   number & >=0.25 & <=3
	point_radius?: number & >=0 & <=6
	area_opacity?: number & >=0 & <=1
	stacking?: "Normal" | "Percent"
        // TBD: another option will be needed to control ordering
}
```

## Screenshots

panel_edit_area_opacity_slider

https://user-images.githubusercontent.com/9369625/219121680-07c0c621-4838-4d47-aac6-053eba292070.mp4

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.

PR #997, opened 2023-02-28
[FEATURE] Stack option support in time series panel

sjcobb:
Follow up to #962, this PR adds stacked area support to our time series panel. Proposed panel option is `stack` and currently only 'Normal' is supported, 'Percent' will be supported in the future (which is why Autocomplete is used)

## TODOs for future PR
- add `stack: 'Percent'` support
- visual regression tests (will need a new mockTimeSeriesResponse function that returns multiple series with pre-defined curve)

## Screenshots

image

# Checklist


ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -109,6 +113,22 @@ export const Y_AXIS_CONFIG = {
   max: { label: 'Max' },
 };
 
+// TODO: add Percent mode support for stacked area charts
+export type StackOptions = 'None' | 'Normal';
```
juliepagano, line 117: I found the "Normal" name kind of confusing, but I think it'll make more sense once the "Percent" is in there.

juliepagano (approved):
LGTM! Let me know if you need help with the visual testing follow-up bits.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
sjcobb: Yeah that’s true, would changing to all “All”

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -99,6 +105,28 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
           />
         }
       />
+      <OptionsEditorControl
+        label={VISUAL_CONFIG.stack.label}
+        control={
+          <Autocomplete
```
eunicorn, line 111: why not use `ToggleButtons`? 😛

eunicorn (approved):

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: 😂  So my [first commit](https://github.com/perses/perses/pull/997/commits/c53fb85fd5b6f02d666fd1b97938c442d2d99dee#diff-217014ce88a4481e76f779f48d62dbb61b6aae3be28f42fdb7ade48afcb6ed5a) actually used ToggleButtons, since there are going to be more than 2 options (adding 'Percent') Mohit recommended switching to a dropdown so it's consistent with the other fields

normal_stack_area_ex

schemas/panels/time-series/time-series.cue
```diff
@@ -26,6 +26,7 @@ import (
 	area_opacity?:  number & >=0 & <=1
 	show_points?:   "Auto" | "Always"
 	point_radius?:  number & >=0 & <=6
+	stack?:         "None" | "Normal" // TODO: add "Percent" option support
```
Nexucis, line 29: `"None"` isn't a bit redundant since `stack` is an optional field ?

sjcobb: Yeah agreed I'll fix this, planning to leave it as an option in the dropdown for visual editing, but selecting 'None' will reset `stack` so it's undefined.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
sjcobb: Ended up changing to use 'All' instead of 'Normal'

schemas/panels/time-series/time-series.cue
sjcobb: This is updated now

PR #999, opened 2023-03-02
[BUGFIX] Add null checks for layouts and panels in DashboardProvider

cndonovan:
If a dashboard passed into the DashboardProvider has `layouts` or `panels` that are `undefined`/`null` for some reason, this guarantees that they are initialized to `[]` or `{}`.

# Checklist

### DIFF
--- a/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
+++ b/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
@@ -103,7 +103,7 @@ function initStore(props: DashboardProviderProps) {
     spec: { layouts, panels },
   } = dashboardResource;
 
-  // Set fallbacks
+  // Set fallbacks in case the frontend is used with a non-Perses backend
   layouts = layouts ?? [];
   panels = panels ?? {};
 
--- a/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
+++ b/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
@@ -128,12 +128,12 @@ function initStore(props: DashboardProviderProps) {
           defaultTimeRange: { pastDuration: duration },
           isEditMode: !!isEditMode,
           setEditMode: (isEditMode: boolean) => set({ isEditMode }),
-          setDashboard: ({ metadata, spec: { display, panels, layouts } }) => {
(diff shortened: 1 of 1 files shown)


eunicorn:
We should also add a check to [setDashboard](https://github.com/perses/perses/blob/f66510e9a436efe45b83b648bad1b634b2c1b364/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx#L119) which is used to update the json schema when people use the json editor

eunicorn (approved):

sjcobb (approved):

saminzadeh (approved):
I wonder if we make the properties in the spec optional as well?

https://github.com/perses/perses/blob/f66510e9a436efe45b83b648bad1b634b2c1b364/ui/core/src/model/dashboard.ts#L32-L33

cndonovan:
> 

I am not sure actually! I was mostly doing this to protect against what we get from the backend (we had a bug because the backend was returning `null` in some cases). Related to that, I was imagining a future where Perses is widespread, and many different backends are being used with it, and those backends could return whatever kind of data they want. 

However, I'm also doing it because I want the ability to create a new dashboard on the frontend, and I don't want to have to type out the full object when I do that -- I want to just type something like `{ metadata: { name: 'New Dashboard' }, spec: {}}` when I create a new dashboard. Of course, I can use a helper function to make it easier (which it's currently doing), but still does seem nice to say `{ metadata: {}, spec: {}}` represents an empty, fresh dashboard, and the store can take care of the rest. 

So, maybe? I'll leave it as-is for now and think about it.

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
```diff
@@ -95,33 +95,45 @@ function initStore(props: DashboardProviderProps) {
...
-    spec: { display, layouts, panels, duration },
+    spec: { display, duration },
     metadata,
   } = dashboardResource;
+
+  let {
+    spec: { layouts, panels },
+  } = dashboardResource;
+
+  // Set fallbacks
```
juliepagano, line 106: It might be worth adding a little clarification about the generic (i.e. not chronosphere-specific) "why" for this for future contributors. I could see someone with less context being slightly confused because I think the perses Go backend prevents you from having these values be `null`.

juliepagano (approved):
One optional suggestion to add some additional context in the comment. Otherwise, LGTM! 🎉

PR #1010, opened 2023-03-08
[IGNORE] Logo added to app header

sjcobb:
# Checklist


Nexucis (approved):
🥳 
thank you for doing it @sjcobb !! Looks so cool :)

eunicorn:
Should we also update all the pngs to white?

eunicorn (approved):

cndonovan (approved):
This might be a silly question, but why do we need the assets too?

I see the `PersesLogo` seems to contain all the data about how to draw the logo in SVG format, but I feel like I'm forgetting something obvious here.

Nexucis:
we should add also the favicon

sjcobb:
> I see the `PersesLogo` seems to contain all the data about how to draw the logo in SVG format, but I feel like I'm forgetting something obvious here.

Nah you're right @cndonovan! When I first exported out the logos from Figma I put them in that folder, then I decided to go with the inline SVG approach. Then was thinking that maybe it could be useful to commit the logos anyways so we have easy access to them later for other purposes. There probably is a better spot to put them though long term so I just removed that folder

sjcobb:
> Should we also update all the pngs to white?

So I ended up removing that folder! There was some discussion about this though with Mohit and he agreed about changing to white in the app header, but outside of the product, it's okay to use the logo with color

sjcobb:
> Perhaps in this PR you could add the logo to the Readme too

> we should add also the favicon

@Nexucis The README update was surprisingly tricky since having the text in the logo didn't look good in both dark and light mode. So I needed to strip out the text and use a regular H1. Also I was going to need to play around with positioning. 

Planning to do a separate PR for the README and favicon! I'm going to go ahead merge this while it's still green

PR #1011, opened 2023-03-09
[IGNORE] Fix flaky e2e test

juliepagano:
Occasionally this test would fail because the selector would find two labels with "max" in them if the tooltip that described the "#" button was open. Modified the selector to be more specific to only grab the max numeric input.

# Checklist

# How I debugged this

Writing this out to hopefully help others debug similar issues in the future.

I looked at the [error output from a flaky test where this happened](https://github.com/perses/perses/actions/runs/4369773233/jobs/7643946154).
image

The error output noted that there were two selectors. I was able to use that to poke around at the source code and see what the conflict was. I then adjusted the selector to something more specific that would not overlap with another item.

I ran the e2e tests twice in CI to verify it wasn't flaky after this.


eunicorn (approved):
thank you!!! i was also seeing this test failing in my branch and was so happy to see that you already got a fix out :D

PR #1012, opened 2023-03-09
[FEATURE] Implement Edit JSON Dialog

eunicorn:
Now you can edit json directly :) 

image

**FUTURE WORK:**
- missing validation and error handling


ui/e2e/src/pages/DashboardPage.ts
```diff
@@ -95,7 +95,7 @@ export class DashboardPage {
 
     this.toolbar = page.getByTestId('dashboard-toolbar');
     this.timePicker = page.getByRole('button', { name: 'Select time range' });
-    this.editButton = this.toolbar.getByRole('button', { name: 'Edit' });
+    this.editButton = this.toolbar.getByRole('button', { name: /Edit$/ });
```
eunicorn, line 98: need exact match now that I've added `Edit JSON` button

juliepagano:
I think you might need to make some styling adjustments, so that the "apply" button and header are always in view.

image

It also might be nice to add a min height based on a chunk of the viewport. Otherwise, it looks a little silly when there isn't a lot in there or the sections are collapsed.

image

juliepagano:
It looks like modifying the `metadata` (e.g. dashboard name, project) does not work properly in the perses app. Is that part of the planned follow-up validation/error handling work, or is that a different issue that needs to be sorted out?

Nexucis:
> It looks like modifying the `metadata` (e.g. dashboard name, project) does not work properly in the perses app. Is that part of the planned follow-up validation/error handling work, or is that a different issue that needs to be sorted out?

well these fields cannot be modified once the dashboard is created. Because what is contained in the `metadata` are used to generate the unique key for the database.
So these fields shouldn't change to be able to find back what entry in the database we need to update

juliepagano:
> > It looks like modifying the `metadata` (e.g. dashboard name, project) does not work properly in the perses app. Is that part of the planned follow-up validation/error handling work, or is that a different issue that needs to be sorted out?
> 
> well these fields cannot be modified once the dashboard is created. Because what is contained in the `metadata` are used to generate the unique key for the database. So these fields shouldn't change to be able to find back what entry in the database we need to update

I wonder if there's a reasonable UX where we just make the other content editable instead of the whole JSON, since those bits can't be modified this way.

eunicorn:
> I wonder if there's a reasonable UX where we just make the other content editable instead of the whole JSON, since those bits can't be modified this way.

Do you think maybe we shouldn't show the `metadata` in the json editor? And only get the full dashboard spec when you download the json?

 I think this also relates to how we shouldn't need to [pass the whole `DashboardResource` to `DashboardProvider`](https://github.com/perses/perses/blob/main/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx#L49) since `DashboardProvider` doesn't care about the metadata

Nexucis:
> Do you think maybe we shouldn't show the metadata in the json editor? And only get the full dashboard spec when you download the json?

I think it's ok to display the field `kind` and `metadata` even if they are not editable. We could add a tooltip explaining that don't you think ?

cndonovan:
@Nexucis  If we can't change the `metadata.name`, how do we change the name of a dashboard? Is that stored somewhere else?

Nexucis:
> @Nexucis If we can't change the `metadata.name`, how do we change the name of a dashboard? Is that stored somewhere else?

in `spec.display.name` you can set a different title with spaces and special characters.

ui/dashboards/src/context/DashboardProvider/edit-json-dialog-slice.tsx
```diff
@@ -0,0 +1,45 @@
...
+  editJsonDialog?: EditJsonDialogState;
+  openEditJsonDialog: () => void;
+  closeEditJsonDialog: () => void;
+}
+
+export interface EditJsonDialogState {
+  isOpen: boolean;
+}
+
+export const createEditJsonDialogSlice: StateCreator<EditJsonDialogSlice, Middleware, [], EditJsonDialogSlice> = (
```
sjcobb, line 27: This is pretty neat, will we leverage this in the internal implementation too?

sjcobb (approved):
+1 on Julie’s comments, going ahead and approving, can’t wait to use this and not have to download to see the full JSON! 🎉

eunicorn:
@juliepagano 

added min height: 
image

always show header and apply button:
image

juliepagano:
> > Do you think maybe we shouldn't show the metadata in the json editor? And only get the full dashboard spec when you download the json?
> 
> I think it's ok to display the field `kind` and `metadata` even if they are not editable. We could add a tooltip explaining that don't you think ?

I think it would be better to just show the content that is editable if that's not a terrible amount of extra work. Otherwise, I think the UX would be really confusing. A tooltip might help a little, but tooltips generally should only include supplemental information because users often ignore them, so I'd worry about relying on it to explain that you can't edit part of the content in a text editor.

juliepagano (approved):
🎉

eunicorn:
> I think it's ok to display the field kind and metadata even if they are not editable. We could add a tooltip explaining that don't you think ?

added alert warning users metadata cannot be modified or saved for now until @foxbat07 comes up with a better design :)

image

PR #1015, opened 2023-03-10
[ENHANCEMENT] Fetch updates to show full error message

sjcobb:
Update fetch to throw UserFriendlyError, tested with invalid panel properties

## Screenshots

fetch_update_unit_invalid

invalid_legend

# Checklist


ui/core/src/utils/fetch.ts
```diff
@@ -42,3 +49,13 @@ export class FetchError extends Error {
     Object.setPrototypeOf(this, FetchError.prototype);
   }
 }
+
+/**
+ * General error type for an error that has a message that is OK to show to the end user.
+ */
+export class UserFriendlyError extends Error {
```
juliepagano, line 56: The error messages from the backend often aren't particularly user friendly right now. Is this aspirational assuming that the backend will be updated eventually to make them more friendly? I'm cool with the change either way to surface errors more helpfully, just curious about the thoughts about the UX going forward.

juliepagano (approved):
One context question. Otherwise looks good to me assuming @Nexucis is cool with it. 🎉

Nexucis:
yeah I'm totally fine with this change !

For sure we can try to improve the error message but sometimes it's quite harder unfortunately :(.
When it's related to the `cue`, hopefully the next version of `cue` will help around that.

Nexucis (approved):

ui/core/src/utils/fetch.ts
```diff
@@ -17,6 +17,13 @@
 export async function fetch(...args: Parameters<typeof global.fetch>) {
```
cndonovan, line 17: Could update the comment if it seems helpful

cndonovan (approved):
I don't mess around with `class` stuff a lot, I'm wondering if we could just throw a general `Error`?

It seems like the `FetchError` does special things with the `response.status` and `response.statusText` (and assumes that it has a `response` argument with this fields), so that's why we call it on a failed response

Seems like `UserFriendlyError` does what `Error` might normally do, and only assumes it's being given a string?

ui/core/src/utils/fetch.ts
sjcobb: Yeah definitely, also you're probably right about how using `Error` would be sufficient for this PR. I do think there is functionality we can build on top of this though, where having UserFriendlyError will be helpful. If you search for `getUserFriendlyErrors` in our internal codebase, there are ideas there that can we possibly leverage in the future.

PR #1019, opened 2023-03-14
[ENHANCEMENT] Extend datasource with onCreate prop, return warnings

sjcobb:
DatasourceStoreProvider updates to extend createClient using a new `onCreate` prop. This allows embed users to make adjustments to the core client.

Optional TimeSeriesMetadata is added to TimeSeriesData, currently this allows setting `notices` from response data, which will be used to display warnings and errors in panel headers.

# Checklist


ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
```diff
@@ -64,6 +65,19 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
   // TODO: What about error responses from Prom that have a response body?
   const result = response.data?.result ?? [];
 
+  // Custom display for response header warnings, configurable error responses display coming next
```
juliepagano, line 68: Is this another bit of code that could be tested?

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -140,3 +144,33 @@ function useTimeSeriesQueryContext(): TimeSeriesQueryContext {
...
+  const queryClient = useQueryClient();
+  const queryCache = queryClient.getQueryCache();
+  const timeSeriesQueries = getActiveTimeSeriesQueries(queryCache);
+  return timeSeriesQueries;
+}
+
+/**
+ * Filter all cached queries down to only active time series queries
+ */
+export function getActiveTimeSeriesQueries(cache: QueryCache) {
```
juliepagano, line 161: Would it be beneficial to add a unit test for this bit of logic?

ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -96,9 +99,18 @@ export function DatasourceStoreProvider(props: DatasourceStoreProviderProps) {
...
       const { kind } = selector;
       const [{ spec, proxyUrl }, plugin] = await Promise.all([findDatasource(selector), getPlugin('Datasource', kind)]);
-      return plugin.createClient(spec.plugin.spec, { proxyUrl }) as Client;
+
+      // allows extending client
+      const datasourceSpec = { ...spec.plugin.spec };
+      const client = plugin.createClient(datasourceSpec, { proxyUrl }) as Client;
+      if (onCreate !== undefined) {
+        if (isDatasourceClient(client)) {
+          return onCreate(client) as unknown as Client;
```
juliepagano, line 108: Why the double `as`?

ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -96,9 +99,18 @@ export function DatasourceStoreProvider(props: DatasourceStoreProviderProps) {
     async function getClient<Client>(selector: DatasourceSelector): Promise<Client> {
       const { kind } = selector;
       const [{ spec, proxyUrl }, plugin] = await Promise.all([findDatasource(selector), getPlugin('Datasource', kind)]);
-      return plugin.createClient(spec.plugin.spec, { proxyUrl }) as Client;
+
+      // allows extending client
+      const datasourceSpec = { ...spec.plugin.spec };
```
juliepagano, line 104: Is this trying to do a copy to avoid issues with mutation from the code calling `onCreate` or something else? If the former, is this deep copying enough to do that?

juliepagano (commented):
A handful of small questions. I wasn't sure how to test out that this works -- let me know if you've got a recommendation, but adding tests would also cover that.

saminzadeh (approved):
Glad to see we were able to make this work without modifying too much of the dashboard spec configuration. Some minor comments but I think we can iterate on those ideas.

ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -26,12 +26,15 @@ import {
...
   DatasourceMetadata,
+  DatasourceClient,
+  isDatasourceClient,
 } from '@perses-dev/plugin-system';
 
 export interface DatasourceStoreProviderProps {
   dashboardResource: DashboardResource;
   datasourceApi: DatasourceApi;
   children?: React.ReactNode;
+  onCreate?: (client: DatasourceClient) => DatasourceClient;
```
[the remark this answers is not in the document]
LukeTillman: That's another good possibility for hooking in (i.e. a "plugin factory"). Definitely worth thinking about. 👍

ui/plugin-system/src/model/datasource.ts
```diff
@@ -24,3 +24,22 @@ export interface DatasourcePlugin<Spec = UnknownSpec, Client = unknown> extends
...
   proxyUrl?: string;
 }
+
+/**
+ * Common properties for all clients
+ */
+export interface DatasourceClient {
+  // TODO: define common client properties
+  options: {
+    datasourceUrl: string;
```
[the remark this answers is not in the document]
LukeTillman: Yeah, we talked about this and up until now, "Client" has just been `unknown` I think. Not clear to me what (if anything) should go on a common interface. I was thinking maybe a `healthCheck` method might make sense for the framework code to handle, but that's all I got right now.

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -140,3 +144,33 @@ function useTimeSeriesQueryContext(): TimeSeriesQueryContext {
...
+    .findAll({ type: 'active' })
+    .filter((query) => {
+      const firstPart = query.queryKey?.[0] as UnknownSpec;
+      if (firstPart?.kind) {
+        return (firstPart?.kind as string).startsWith(TIME_SERIES_QUERY_KEY);
+      }
+      return false;
+    })
+    .filter((query) => query.isActive)
+    .map((query) => {
```
LukeTillman, line 172: Just a general thought here: this is a lot of loops happening with the multiple `filter`/`map`/etc. calls. Wonder if it would be simpler/faster to just write this as a single for...of loop?

LukeTillman (approved):
Overall, looks good. I think Shan definitely has some good suggestions to think about for a "v2". And we should think about how to iterate on the types as well. One thought I had as a possibility for the generic "DatasourceClient" type:

```ts
interface DatasourceClient {
  // Framework code sets this to match the plugin's kind
  kind: string;
  // Framework can check to see if Datasource is working/healthy when configured
  healthCheck(): Promise<boolean>;
}
```

Then maybe the `kind` can be overridden by the specific client type and used as a sort of "discriminator" for casting to the correct type.

ui/plugin-system/src/model/datasource.ts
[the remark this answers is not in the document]
saminzadeh: +1 `healthCheck()` makes sense

ui/plugin-system/src/runtime/time-series-queries.ts
sjcobb: This one is kind of difficult to write a unit test for, I tried mocking what is returned in queryCache, but it's probably better to use msw so started on that, but it's becoming a bigger effort since we haven't used msw yet in Perses. 

I'll probably come back to this later when we go to make the validation / warnings banner more generic--at that point, I'll make an e2e test as well.

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -140,3 +144,33 @@ function useTimeSeriesQueryContext(): TimeSeriesQueryContext {
     refreshKey,
   };
 }
+
+/**
+ * Get active time series queries for query results summary
+ */
+export function useActiveTimeSeriesQueries() {
```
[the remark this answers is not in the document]
sjcobb: Gotcha yeah I adjusted some things internally to get it to work in our custom banner

ui/dashboards/src/context/DatasourceStoreProvider.tsx
sjcobb: Fixed this, I was having trouble with the types since DatasourceClient was essentially unknown before, but adding healthCheck and adjusting getClient helped

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
sjcobb: Yeah definitely, I'm planning to add tests internally that will be helpful and when Eunice's DataQueryProvider updates are in, I'm going to come back to this, build out the warnings tags that will show in a Panel's header, and add tests with Playwright / Happo

ui/plugin-system/src/runtime/time-series-queries.ts
sjcobb: Yeah good call, I [refactored this](https://github.com/perses/perses/pull/1019/commits/42566aa6983981768ed9da33ca2f73c37fe7eb9c) so it's much simpler/faster now

ui/plugin-system/src/model/datasource.ts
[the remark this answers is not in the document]
sjcobb: Yeah this makes sense, I added it as optional for now and left a TODO for next steps

ui/dashboards/src/context/DatasourceStoreProvider.tsx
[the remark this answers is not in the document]
sjcobb: +1 I like this idea too for the future, planning to look more at how it works in CodeMirror

juliepagano (approved):

PR #1020, opened 2023-03-15
[IGNORE] Bump go to v1.20.x

Nexucis:


internal/api/shared/migrate/loader.go
```diff
@@ -56,7 +56,6 @@ type loader interface {
 }
 
 type migCuePart struct {
-	loader
```
Nexucis, line 59: I removed it because the linter was saying it was unused.

```bash
make checkstyle
>> checking code style
golangci-lint run --timeout 5m
internal/api/shared/migrate/loader.go:59:2: field `loader` is unused (unused)
	loader
	^
```

I guess that's because the interface / field in the struct wasn't public

AntoineThebaud (approved):

PR #1023, opened 2023-03-15
[FEATURE] Add __experimentalEChartsOptionsOverride prop to time series charts

stephtaylor:
Allows users to set a `__experimentalEChartsOptionsOverride` prop to override EChart options for the LineChart component.

# Checklist

### DIFF
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -68,7 +68,7 @@ export interface LineChartProps {
   legend?: LegendComponentOption;
   onDataZoom?: (e: ZoomEventData) => void;
   onDoubleClick?: (e: MouseEvent) => void;
-  __experimentalEChartsOptionsOverride?: EChartsCoreOption;
+  __experimentalEChartsOptionsOverride?: (options: EChartsCoreOption) => void;
 }
 
 export function LineChart({
@@ -177,9 +177,11 @@ export function LineChart({
       },
       grid,
       legend,
-      ...__experimentalEChartsOptionsOverride,
     };
 
+    if (__experimentalEChartsOptionsOverride) {
+      return __experimentalEChartsOptionsOverride(option);
+    }
     return option;
   }, [data, yAxis, unit, grid, legend, noDataOption, timeZone, __experimentalEChartsOptionsOverride]);
 
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -68,7 +68,7 @@ export interface LineChartProps {
   legend?: LegendComponentOption;
   onDataZoom?: (e: ZoomEventData) => void;
   onDoubleClick?: (e: MouseEvent) => void;
-  __experimentalEChartsOptionsOverride?: (options: EChartsCoreOption) => void;
+  __experimentalEChartsOptionsOverride?: (options: EChartsCoreOption) => EChartsCoreOption;
 }
 
 export function LineChart({
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -19,6 +19,7 @@ import type {
   LineSeriesOption,
   LegendComponentOption,
   YAXisComponentOption,
+  TooltipComponentOption,
 } from 'echarts';
 import { ECharts as EChartsInstance, use } from 'echarts/core';
 import { LineChart as EChartsLineChart } from 'echarts/charts';
@@ -212,7 +213,7 @@ export function LineChart({
       }}
       onDoubleClick={handleOnDoubleClick}
     >
-      {showTooltip === true && (
+      {showTooltip === true && (option.tooltip as TooltipComponentOption).showContent === false && (
         <TimeSeriesTooltip chartRef={chartRef} chartData={data} wrapLabels={true} pinTooltip={pinTooltip} unit={unit} />
       )}
 
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -68,9 +68,20 @@ export interface LineChartProps {
   legend?: LegendComponentOption;
   onDataZoom?: (e: ZoomEventData) => void;
   onDoubleClick?: (e: MouseEvent) => void;
+  __experimentalEChartsOptionsOverride?: EChartsCoreOption;
 }
 
-export function LineChart({ height, data, yAxis, unit, grid, legend, onDataZoom, onDoubleClick }: LineChartProps) {
+export function LineChart({
+  height,
+  data,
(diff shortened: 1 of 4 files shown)


Nexucis:
Thanks for opening this PR and welcome in Perses :).

I supposed the real goal of this PR is to support multiple y axis right ?
If it's what you would like to achieve here, I think a proper integration would be better.
Also because I guess there is no connexion between a query and which y axis it should be correlated and so the y axis cannot be adapted depending of the value of the timeseries.

I think we discussed sometimes ago about if we should integrate the management of two y axis. So perhaps here we have the good opportunity to know why it's important to do so :). Would you mind to explain your needs ? :)

/cc @sjcobb @LukeTillman @juliepagano @saminzadeh @eunicorn @cndonovan

sjcobb:
> I supposed the real goal of this PR is to support multiple y axis right ?

Hey @Nexucis, this is a change Steph and I are going to collaborate on (she is from Chronosphere). For this PR, the actual goal is to provide an "escape hatch" for embed users that are using a single panel and/or component. This was allow access to other ECharts option properties than what we don't support explicitly on LineChart.tsx [here](https://github.com/perses/perses/blob/v0.24.0/ui/components/src/LineChart/LineChart.tsx#L59-L71). There is a different way to do this though by exporting out a separate wrapper component and setting the new prop there. That way we aren't changing the actual panel spec / CUE schema. I'm going to make an example and follow up with Steph shortly.

As far as proper support for a second y axis, I definitely agree that this will be bigger decision. Like you said, it will require adding the concept of series matchers / overrides so we know which time series goes with which axis and which unit to show in places like the tooltip. Going to write up some more thoughts on this as well

saminzadeh:
I think it would be cool if this was a function instead:

```tsx

function Example() {
   // Do anything custom in this function
   const transformConfig = (options: EChartsCoreOption) => {
        // Adds a title to graph
        options.title = {
            title: { show: true }
       }
        return options;
   }
   
   return <LineChart __experimentalEChartsOptionsOverride={transformConfig} />
}
```

`__experimentalEChartsOptionsOverride` would be a function that receives all the current [eCharts options ](https://github.com/perses/perses/blob/b0d0c88bccf3fa7fa0da4da0c0527954f0c94265/ui/components/src/EChart/EChart.tsx#LL102C22-L102C22) and allows the consumer to modify the object before it passed to the chart. That way it's more clear what will be overwritten.

stephtaylor:
Thanks @sjcobb for filling in the context for everyone! I totally slipped my mind to include it after chatting yesterday.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -25,6 +26,7 @@ export interface TimeSeriesChartOptions {
   unit?: UnitOptions;
   thresholds?: ThresholdOptions;
   visual?: VisualOptions;
+  __experimentalEChartsOptionsOverride?: (options: EChartsCoreOption) => void;
```
saminzadeh, line 29: Should this return `EChartsCoreOption` ?

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -52,7 +52,7 @@ export type TimeSeriesChartProps = PanelProps<TimeSeriesChartOptions>;
 
 export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   const {
-    spec: { queries, thresholds, y_axis },
+    spec: { queries, thresholds, y_axis, __experimentalEChartsOptionsOverride },
```
sjcobb, line 55: Hey @stephtaylor! I talked to @saminzadeh about this PR and he's planning to look into the `TimeSeriesChartViz` export part of this task when he's back next week. In the meantime, for this PR it'd be great to focus on `__experimentalEChartsOptionsOverride` support for LineChart.tsx and ability to hide TimeSeriesTooltip.

Can you remove the changes in time-series-chart-model.ts and TimeSeriesChartPanel.tsx? And then in LineChart.tsx can you update [these lines](https://github.com/perses/perses/blob/v0.25.0/ui/components/src/LineChart/LineChart.tsx#L201-L203) to:


```typescript
{showTooltip === true && option.tooltip.showContent === false && (
  <TimeSeriesTooltip chartRef={chartRef} chartData={data} wrapLabels={true} pinTooltip={pinTooltip} unit={unit} />
)}
```

Then when you're testing this, make sure __experimentalEChartsOptionsOverride.tooltip.showContent is true. I think this will hopefully be enough short term for setting the correct unit for each series in the tooltip since it will allow you to customize the tooltip [formatter](https://echarts.apache.org/en/option.html#tooltip.formatter) (there is a good example internally of this in timeline's BarChart.tsx)

stephtaylor: That sounds good to me! 👍🏼

sjcobb: 😣 So I tested this and some additional changes will be needed, currently the ECharts tooltip gets cut off:

https://user-images.githubusercontent.com/9369625/228897585-7d51a579-a4cd-442f-826c-9821cf772d86.mp4

Our custom TimeSeriesTooltip uses [Portal](https://mui.com/material-ui/react-portal/) so doesn't run into this, I remember adding `overflow: hidden` to Panel.tsx at some point, but need to see what happens when it is removed

sjcobb: Adding `tooltip.appendToBody = true` fixes this!
https://echarts.apache.org/en/option.html#tooltip.appendToBody

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -198,7 +213,7 @@ export function LineChart({ height, data, yAxis, unit, grid, legend, onDataZoom,
       }}
       onDoubleClick={handleOnDoubleClick}
     >
-      {showTooltip === true && (
+      {showTooltip === true && (option.tooltip as TooltipComponentOption).showContent === false && (
```
sjcobb, line 216: I'm going to play around with [echarts.ComposeOption](https://apache.github.io/echarts-handbook/en/basics/import/#creating-an-option-type-in-typescript) later to see if I can make it so the `as TooltipComponentOption` isn't necessary but this looks good for now!

sjcobb (approved):
🎉 LGTM thanks for taking this on!

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
stephtaylor: Oh nice find!

PR #1026, opened 2023-03-16
[IGNORE] Validate all dashboards and dts contain in dev/data

Nexucis:
I just figured out we are starting to have plenty of dashboard in `dev/data` and maybe having something that is validating the dashboards will ensure the `dev/data` are compatible with the current version of the application.

### DIFF
--- a/scripts/cue-test/cue-test.go
+++ b/scripts/cue-test/cue-test.go
@@ -17,10 +17,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 
 	"github.com/perses/perses/internal/api/config"
 	"github.com/perses/perses/internal/api/shared/schemas"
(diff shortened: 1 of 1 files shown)


AntoineThebaud (approved):

scripts/cue-test/cue-test.go
```diff
@@ -65,3 +62,67 @@ func main() {
 		return sch.ValidateVariable(plugin, name)
 	})
 }
+
+func validateAllDashboards(sch schemas.Schemas) {
+	logrus.Info("validate all dashboard in dev/data")
+	data, err := os.ReadFile(path.Join("dev", "data", "dashboard.json"))
```
sjcobb, line 68: Looking at this again it, it makes sense, I see we're validating the initial data that gets populated, I think I'd be able to iterate on panel properties in local_db still?

sjcobb (approved):

PR #1030, opened 2023-03-20
[BUGFIX] Fix the deletion of a project that didn't remove the related dashboards

Nexucis:
Before this PR, when you remove a project, the related dashboards, folders and datasources were not removed as well. So when you re-created the project with the same name, you were seeing again the old dashboard / folders / datasources

### DIFF
--- a/internal/api/shared/database/sql/query.go
+++ b/internal/api/shared/database/sql/query.go
@@ -80,7 +80,7 @@ func (d *DAO) generateUpdateQuery(entity modelAPI.Entity) (string, []interface{}
 	return sql, args, nil
 }
 
-func generateProjectResourceSelectQuery(tableName string, project string, name string) (string, []interface{}) {
+func generatSelectQuery(tableName string, project string, name string) (string, []interface{}) {
 	queryBuilder := sqlbuilder.NewSelectBuilder().
 		Select(colDoc).
 		From(tableName)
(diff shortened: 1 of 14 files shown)


internal/api/shared/database/sql/query.go
```diff
@@ -122,3 +122,44 @@ func (d *DAO) buildQuery(query databaseModel.Query) (string, []interface{}, erro
 	}
 	return sqlQuery, args, nil
 }
+
+func generateProjectResourceDeleteQuery(tableName string, project string, name string) (string, []interface{}) {
```
AntoineThebaud, line 126: What about this to avoid some duplication ?

```go
func generateResourceDeleteQuery(tableName string, project string, name string) (string, []interface{}) {
        queryBuilder := sqlbuilder.NewDeleteBuilder().DeleteFrom(tableName)
	if len(name) > 0 {
		queryBuilder.Where(queryBuilder.Like(colName, fmt.Sprintf("%s%%", name)))
	}
	if len(project) > 0 {
		queryBuilder.Where(queryBuilder.Equal(colProject, project))
	}
	return queryBuilder.Build()
}

func (d *DAO) buildDeleteQuery(query databaseModel.Query) (string, []interface{}, error) {
	var sqlQuery string
	var args []interface{}
	switch qt := query.(type) {
	case *dashboard.Query:
		sqlQuery, args = generateResourceDeleteQuery(d.generateCompleteTableName(tableDashboard), qt.Project, qt.NamePrefix)
	case *datasource.Query:
		sqlQuery, args = generateResourceDeleteQuery(d.generateCompleteTableName(tableDatasource), qt.Project, qt.NamePrefix)
	case *folder.Query:
		sqlQuery, args = generateResourceDeleteQuery(d.generateCompleteTableName(tableFolder), qt.Project, qt.NamePrefix)
	case *globaldatasource.Query:
		sqlQuery, args = generateResourceDeleteQuery(d.generateCompleteTableName(tableGlobalDatasource), "", qt.NamePrefix)
	case *project.Query:
		sqlQuery, args = generateResourceDeleteQuery(d.generateCompleteTableName(tableProject), "", qt.NamePrefix)
	default:
		return "", nil, fmt.Errorf("this type of query '%T' is not managed", qt)
	}
	return sqlQuery, args, nil
}
```

Nexucis: ah yes you are right. I don't what I was thinking🤦🏼

AntoineThebaud (approved):

PR #1031, opened 2023-03-20
[FEATURE] Categorical color palette support in time series panel spec (JSON only)

sjcobb:
## Overview

Add option for `Categorical` color palette that can be set to override `Auto` generative palette in TimeSeriesChart panel.

I also changed getRandomColor so the formatted series name is used as a seed in the generative palette instead of the raw unformatted series name. This gives users some control over colors since it's tied to how `series_name_format` is used.

_This PR does not include visual editing of our color palette spec, it is only supported by editing JSON directly_

### Proposed schema:
```cue
#palette: {
	kind: "Auto" | "Categorical"
        // colors: [string] // future option to override ECharts theme palette
}

#visual: {
	palette?: #palette
	...
}
```

## Screenshot

image

image

# Checklist


Nexucis:
I supposed we could have multiple different strategy about the coloration.

"manual", "auto", "linear" (which is the `Categorical` today), "labels"

Nexucis:
probably depending of the `kind`, it would change the properties. So based on our datamodel, we would have a `spec` section

```json
{
  "kind" : "manual"
  "spec": {
    colors: [....]
  }
}
```

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -23,3 +27,26 @@ export function getRandomColor(identifier: string): string {
...
+ * Get line color as well as color for tooltip and legend, account for whether palette is 'Cateogrical' or 'Auto' (generative)
+ */
+export function getSeriesColor(
+  name: string,
+  seriesCount: number,
+  palette: string[],
+  paletteKind: PaletteOptions['kind'] = 'Auto'
+): string {
+  if (paletteKind === 'Auto') {
+    return getRandomColor(name);
```
eunicorn, line 41: should we just make `Categorical` the default since the current `getRandomColor` doesn't seem the best?

eunicorn (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -154,25 +155,43 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
+        if (timeSeries === undefined) {
+          return { graphData };
+        }
+
+        // Format is determined by series_name_format in query spec
         const formattedSeriesName = timeSeries.formattedName ?? timeSeries.name;
+
+        const seriesColor = getSeriesColor(
+          formattedSeriesName,
+          seriesCount,
```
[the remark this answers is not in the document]
sjcobb: So I tried that at first, but if you just use index, then the color gets reset too frequently. For instance, if a panel has more than 1 query, then the color doesn't cycle through the entire palette before going back to the first color--keeping track of total series regardless of the query was the workaround

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
sjcobb: We are planning to make `getRandomColor` a lot better, the SO article we took the algorithm from actually has a few improvements we can change to that will increase color contrast: https://stackoverflow.com/questions/3426404/create-a-hexadecimal-colour-based-on-a-string-with-javascript/31037383#31037383

sjcobb: Downside of `Categorical` is the colors start to repeat and they aren't consistent across panels. With `Auto` it works better with many time series and the colors will be the same across panels as long as the series names are the same.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -37,10 +37,16 @@ export interface YAxisOptions {
   max?: number;
 }
 
+export interface PaletteOptions {
+  kind: 'Auto' | 'Categorical';
+  // colors: string []; // TODO: add colors to override ECharts theme
```
sjcobb, line 42: @Nexucis so I was thinking about having a colors array (see TODO) to override the colors set in the ECharts [theme here](https://github.com/perses/perses/blob/v0.24.0/ui/components/src/utils/theme-gen.ts#L42). Do you think, we'd need a separate `kind: 'Manual'` for that? 

I was thinking, it be similar to how `abbreviate` is disabled for certain units. Basically, the ability to manually define a palette would be disabled when 'Auto' is selected, but if you pick 'Categorical', the optional color picker UI is available (but if you don't change anything it defaults to the palette set by ChartsThemeProvider).

I think the main change you are proposing is renaming 'Categorical' to 'Linear'?

Nexucis: I was thinking we could have different strategy to set the color.
So here based on what you implemented I had the impression that the selection of the color is "linear" since the color number `n` is always the same.

And then maybe one day we could have a different strategy based on the label like I was explaining above.

But probably in this particular case, independently of the strategy chosen, you could define a set of colors that would be used for the `x` numbers of timeseries available (`x` is the number of colors defined).

And instead of `Auto` I would probably put `Random` if I understood correctly what it is doing.

sjcobb: Okay I’m realizing from Eunice’s comment too that `getRandomColor` is a terrible function name 😂

The colors from that function aren’t actually random, they take the series name as a param, and use that to seed the color so if the same series names are used across panels, the colors should be consistent. I did fix a gap in this PR where the raw series names are used instead of the formatted series names, but otherwise I think the palette we’ve had since the beginning does meet your first comment: “same color for timeseries that are matching the same labels or the same name” 

I do think the colors that get generated aren’t great and don’t have enough contrast, but there is an adjustment we can make to improve that soon

Nexucis: oh interesting ! 

What do you think about the possibility to override the first `x` timeseries color whatever the mode / strategy chosen ? 

Maybe instead of `Auto | Categorical ` we could have `Name | Linear`.
But even `Name` is not that explicite :(. It's quite hard to find a good name that summarise what it is doing 🤣. I'm so bad at naming haha

cndonovan (commented):
One general question I have is -- should we loop Mohit/Karan into this? I feel like we're extending the feature set, but I'm not actually sure if this is the designed/intended way to do it.

Maybe you've already talked to him and I missed it.

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.test.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { getSeriesColor } from './palette-gen';
+
+describe('getSeriesColor', () => {
```
cndonovan, line 16: noiiice, these are super easy to read

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -23,3 +27,26 @@ export function getRandomColor(identifier: string): string {
...
   return color;
 }
+
+/**
+ * Get line color as well as color for tooltip and legend, account for whether palette is 'Cateogrical' or 'Auto' (generative)
+ */
+export function getSeriesColor(
+  name: string,
+  seriesCount: number,
+  palette: string[],
```
cndonovan, line 37: Actually, here would be the fun place to fallback to a default palette of colors!

cndonovan (approved):

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -23,3 +27,26 @@ export function getRandomColor(identifier: string): string {
...
+  paletteKind: PaletteOptions['kind'] = 'Auto'
+): string {
+  if (paletteKind === 'Auto') {
+    return getRandomColor(name);
+  }
+
+  if (paletteKind === 'Categorical' && Array.isArray(palette)) {
```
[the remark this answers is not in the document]
sjcobb: So the default palette is what is defined in the EChart theme [here](https://github.com/perses/perses/blob/v0.25.0/ui/components/src/utils/theme-gen.ts#L42), since that theme gets passed into our ECharts wrapper component, it automatically takes effect. If we used the ECharts legend and tooltip and it would just work there too, but we built those custom so had to loop through the palette explicitly to populate graphData.legendItems

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
sjcobb: I'll add a comment describing more about the ECharts theme! The palette being passed in here is our default palette from `useChartsTheme` [here](https://github.com/perses/perses/blob/6a9432648cb642d48e4c7086a40b2d0bcfafdbf3/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx#L59). Internally we have a different palette defined in our ECHARTS_THEME_OVERRIDES which lets us use different colors for Chronosphere vs. Perses (RedHat also wants to leverage this to use their design system colors)

The FALLBACK_PALETTE_COLOR above isn't really necessary since the ECharts theme palette will always be set, planning to play around with the types to see if I can remove it

Nexucis: @jgbernalp maybe you can be interested by that :)

PR #1039, opened 2023-03-24
[IGNORE] don't generate source map when building the ui

Nexucis:
When we are building the UI for the production, we don't need anymore to generate the source map.
The build is also faster when it is disabled.


Nexucis:
ah wait actually this doesn't work. I thought it was, but no the maps back

PR #1042, opened 2023-03-27
[BUGFIX] Fix datasource spamming by using a cache

### LINKED ISSUE #806: Getting datasources is spamming the backend

Nexucis:
This PR is another proposition to fix #806.

The idea here is to proposed two implementations of the interface `DatasourceAPI`. 

* The first one `HTTPDatasourceAPI` is the regular implementation that is simply contacting the backend like it is done currently on the `main` branch.

* The second one is using a LRU cache in order to save temporary the list of the datasources used by the dashboard.

Finally, the cache is warming up before the dashboard is loaded, so each panel is actually using the cache to get their datasource and doesn't contact the backend anymore.

With this PR for the dashboard `Benchmark`, it contacts the backend 6 times instead of 20.

### DIFF
--- a/ui/app/src/model/datasource-api.ts
+++ b/ui/app/src/model/datasource-api.ts
@@ -11,45 +11,203 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Datasource, GlobalDatasource } from '@perses-dev/core';
+import { Datasource, DatasourceSelector, GlobalDatasource } from '@perses-dev/core';
 import { DatasourceApi } from '@perses-dev/dashboards';
+import LRUCache from 'lru-cache';
 import { fetchDatasourceList, fetchGlobalDatasourceList } from './datasource-client';
 
-export function useDatasourceApi(): DatasourceApi {
-  return {
-    getDatasource: async (project, selector) => {
-      return fetchDatasourceList(project, selector.kind, selector.name ? undefined : true, selector.name).then(
-        (list) => {
-          // hopefully it should return at most one element
-          if (list[0] !== undefined) {
-            return {
-              resource: list[0],
-              proxyUrl: getProxyUrl(list[0]),
-            };
-          }
-        }
-      );
-    },
-    getGlobalDatasource: async (selector) => {
-      return fetchGlobalDatasourceList(selector.kind, selector.name ? undefined : true, selector.name).then((list) => {
-        // hopefully it should return at most one element
-        if (list[0] !== undefined) {
-          return {
-            resource: list[0],
-            proxyUrl: getProxyUrl(list[0]),
-          };
-        }
-      });
-    },
+export class HTTPDatasourceAPI implements DatasourceApi {
+  getDatasource(
+    project: string,
+    selector: DatasourceSelector
+  ): Promise<{ resource: Datasource; proxyUrl: string } | undefined> {
+    return fetchDatasourceList(project, selector.kind, selector.name ? undefined : true, selector.name).then((list) => {
+      // hopefully it should return at most one element
+      if (list[0] !== undefined) {
+        return {
+          resource: list[0],
+          proxyUrl: getProxyUrl(list[0]),
+        };
+      }
+    });
+  }
 
-    listDatasources: async (project, pluginKind) => {
-      return fetchDatasourceList(project, pluginKind);
-    },
+  getGlobalDatasource(
+    selector: DatasourceSelector
+  ): Promise<{ resource: GlobalDatasource; proxyUrl: string } | undefined> {
+    return fetchGlobalDatasourceList(selector.kind, selector.name ? undefined : true, selector.name).then((list) => {
+      // hopefully it should return at most one element
+      if (list[0] !== undefined) {
+        return {
+          resource: list[0],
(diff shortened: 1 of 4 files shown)


ui/app/src/model/datasource-api.ts
```diff
@@ -11,45 +11,203 @@
...
+  }
+}
+
+class Cache {
+  private datasources: LRUCache<string, Datasource>;
+  private emptyDatasources: LRUCache<string, boolean>;
+  private globalDatasources: LRUCache<string, GlobalDatasource>;
+
+  constructor() {
+    const option = { ttl: 5 * 60 * 1000, ttlAutopurge: true };
```
sjcobb, line 64: Can you add a quick inline comment about what `ttl` is / why these values were picked?

ui/app/src/model/datasource-api.ts
```diff
@@ -11,45 +11,203 @@
...
+    const kind = dts.spec.plugin.kind;
+    if (dts.spec.default) {
+      // in case it's the default datasource for the given kind, we store it twice
+      // because we might get the datasource with the name or because we want the default one.
+      this.globalDatasources.set(this.generateKey({ kind: kind }), dts);
+    }
+    this.globalDatasources.set(this.generateKey({ kind: kind, name: dts.metadata.name }), dts);
+  }
+
+  setUndefinedGlobalDatasource(selector: DatasourceSelector) {
```
sjcobb, line 117: I'm not 100% sure it would help here but we've started using [immer](https://immerjs.github.io/immer/) a lot in panels-plugin ([example](https://github.com/perses/perses/blob/main/ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditorSettings.tsx)) for visual editing. It could be worth trying it out in the app too in the future

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -52,7 +52,12 @@ function ViewDashboard() {
 
   const navigate = useNavigate();
   const { successSnackbar, exceptionSnackbar, warningSnackbar } = useSnackbar();
-  const datasourceApi = useDatasourceApi();
+  const [datasourceApi] = useState(new CachedDatasourceAPI(new HTTPDatasourceAPI()));
+  useEffect(() => {
+    // warm up the caching of the datasources
+    datasourceApi.listDatasources(projectName, undefined);
```
sjcobb, line 58: If pluginKind is optional, would just not passing anything work instead of setting `undefined`?

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -52,7 +52,12 @@ function ViewDashboard() {
 
   const navigate = useNavigate();
   const { successSnackbar, exceptionSnackbar, warningSnackbar } = useSnackbar();
-  const datasourceApi = useDatasourceApi();
+  const [datasourceApi] = useState(new CachedDatasourceAPI(new HTTPDatasourceAPI()));
```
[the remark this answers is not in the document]
Nexucis: oh so cool, I didn't know it was possible ! Thank you so much @sjcobb !

ui/app/src/views/ViewDashboard.tsx
Nexucis: yeah you are totally right. I think the auto implementation of the `DatasourceAPI` of IntelliJ messed up.

ui/app/src/model/datasource-api.ts
Nexucis: mmm that's interesting indeed. Not sure as well it will work here

ui/app/src/model/datasource-api.ts
```diff
@@ -11,45 +11,206 @@
...
+class Cache {
+  private datasources: LRUCache<string, Datasource>;
+  private emptyDatasources: LRUCache<string, boolean>;
+  private globalDatasources: LRUCache<string, GlobalDatasource>;
+
+  constructor() {
+    // We want to have a cache expiration set to 5min and removed automatically the old values.
+    // The option below is setting that.
+    // Note: TTL (Time To Leave) is in millisecond.
+    const option = { ttl: 5 * 60 * 1000, ttlAutopurge: true };
```
juliepagano, line 67: Are there any weird edge cases someone could end up in if these values are cached for 5 minutes and something changes in the backend? I'm guessing it's fine, but I have a little less context on this bit of the code, so wanted to double-check.

juliepagano (approved):
Generally LGTM besides the comments Steven already left and a small curiosity question.

ui/app/src/model/datasource-api.ts
Nexucis: Yeah that should be fine. The datasource configuration shouldn't change that often.
And if it's the case then simply reloading the page will refresh the cache. That's not so hurtful I think.

Issue #894, opened 2022-12-23
Fix CSS import errors when Perses components are embedded


sjcobb:
Recently, we moved a lot of our MUI theme into the @perses-dev/components package to make extending types easier, but this causes new issues when embedding Perses in apps built with certain frameworks like Next.js.

The newest versions of Next.js have [a workaround](https://github.com/vercel/next.js/discussions/27953#discussioncomment-3978605) for this, but it is bad practice for libraries like our's to import CSS files this way (more details: https://nextjs.org/docs/messages/css-npm)

These [Lato imports ](https://github.com/perses/perses/blob/main/ui/components/src/theme/typography.ts#L14-L17)specifically are causing problems.

## Screenshot

<img width="877" alt="image" src="https://user-images.githubusercontent.com/9369625/209268497-1eb48e79-b2d0-44bc-b976-853ebefc04ef.png">

sjcobb:
Note: Errors are also being thrown related to lodash-es
```
error - Error [ERR_REQUIRE_ESM]: require() of ES Module /sjcobb/perses-embed-example/node_modules/lodash-es/lodash.js from /sjcobb/perses-embed-example/node_modules/@perses-dev/core/dist/cjs/utils/memo.js not supported.
Instead change the require of lodash.js in /sjcobb/perses-embed-example/node_modules/@perses-dev/core/dist/cjs/utils/memo.js to a dynamic import() which is available in all CommonJS modules.
```

juliepagano:
In general, I don't know that we've been especially good about making sure Perses is SSR-friendly. Do you think that's something we should spend a little time investing in going forward?

sjcobb:
> In general, I don't know that we've been especially good about making sure Perses is SSR-friendly. Do you think that's something we should spend a little time investing in going forward?

Yeah definitely, at least a spike to see where the gaps are and how much work it’ll be (do you have some ideas already?)

PR #1045, opened 2023-03-27
[BUGFIX] Fallback to query when metric labels are empty

sjcobb:
Fallback to query name when metric labels are empty

### Before
image

### After
image

# Checklist


ui/core/src/utils/is-empty-object.ts
```diff
@@ -0,0 +1,19 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Determines if an object is an empty object
+ */
+export function isEmptyObject(obj: object) {
```
saminzadeh, line 17: Even though this is small, my vote would be not to rewrite utils and start using lodash here. 
https://lodash.com/docs/4.17.15#isEmpty

We can treeshake the library to only use utils that are consumed in the code.

saminzadeh (approved):

saminzadeh:
The tooltip has a strange name?
image

Shouldn't it be the full query name?

sjcobb:
> The tooltip has a strange name? image
> 
> Shouldn't it be the full query name?

Yeah it should, will be a quick fix, thanks for catching that!

ui/components/src/TimeSeriesTooltip/SeriesInfo.tsx
```diff
@@ -23,10 +23,11 @@ export interface SeriesInfoProps {
   markerColor: string;
   totalSeries: number;
   wrapLabels?: boolean;
+  showQuery?: boolean;
 }
 
 export function SeriesInfo(props: SeriesInfoProps) {
-  const { seriesName, formattedY, markerColor, totalSeries, wrapLabels } = props;
+  const { seriesName, formattedY, markerColor, totalSeries, wrapLabels, showQuery = true } = props;
```
sjcobb, line 30: Added this `showQuery` prop for additional customization of series name display in the tooltip, but I need to expose this on LineChart itself. 

Ideal behavior in the future will probably be to only show the query automatically when more than 1 query was used in the panel.

ui/core/src/utils/is-empty-object.ts
sjcobb: Yeah this makes sense, our lodash usage is a mess though, I'm planning to fix this up when I tackle Issue #894 (see error in comment [here](https://github.com/perses/perses/issues/894#issuecomment-1363612315))

Lodash also isn't a dependency in prometheus-plugin yet, so I want to make sure we set it up properly this time

sjcobb: cc @juliepagano we might want to add this to our list of technical debt / deciding long term what we want to do about lodash usage, I'll add it as a discussion topic too

ui/components/src/TimeSeriesTooltip/SeriesInfo.tsx
sjcobb: Fixed this up and added a new optional tooltipConfig prop to customize other TimeSeriesTooltip behavior

PR #1046, opened 2023-03-28
[ENHANCEMENT] Option to allow y_axis.min to auto adjust to data

sjcobb:
Changes so origin is not always zero. Will now scale with the min datapoint, can still be overridden from y_axis.min in panel spec

## Screenshots

image

image

image

# Checklist


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -88,7 +88,8 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   // convert Perses dashboard format to be ECharts compatible
   const yAxis = {
     show: y_axis?.show ?? DEFAULT_Y_AXIS.show,
-    min: y_axis?.min, // leaves min and max undefined by default to let ECharts calcualate
+    // dataMin sets minimum axis label relative to data instead of zero
+    min: y_axis?.min === 'ScaleToData' ? 'dataMin' : y_axis?.min, // leaves min and max undefined by default to let ECharts calcualate
```
cndonovan, line 92: We might want to add some padding (i.e. make the y-axis minimum lower than the data minimum) 

But I think this is fine for starting / might be fine in general

schemas/panels/time-series/time-series.cue
```diff
@@ -34,7 +34,7 @@ import (
 	show?:  bool
 	label?: string
 	unit?:  common.#unit
-	min?:   number
+	min?:   number | "ScaleToData"
```
cndonovan, line 37: I like "ScaleToData", makes sense to me

The word that I initially thought of was "ZoomToData", but I think "ScaleToData" works just as well (if not better)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -88,7 +88,8 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   // convert Perses dashboard format to be ECharts compatible
   const yAxis = {
     show: y_axis?.show ?? DEFAULT_Y_AXIS.show,
-    min: y_axis?.min, // leaves min and max undefined by default to let ECharts calcualate
+    // dataMin sets minimum axis label relative to data instead of zero
```
cndonovan, line 91: I might say "dataMin uses the minimum value of the data being plotted"

cndonovan (commented):
I like it! I like keeping it as a union type for the `min` field 

Yeah I'm still jury out on if it should be the default behavior or not 🤔 Will ponder a bit more

cndonovan:
Ok I gave it a bit more thought, and my thinking is:

I think it would be good as the "default" behavior

If someone wants `0` to be the y-axis minimum, that's what the `Min` property is for 

If they don't specify `0`, then the default behavior kicks in. Default behavior zooms into the data.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Yeah good point! Updated the approach to account for this

schemas/panels/time-series/time-series.cue
sjcobb: Ended up removing 'ScaleToData' since I heard from a few other people that it probably should be the default (and min is optional so I set the min relative to the data when the spec `y_axis.min` is undefined)

sjcobb:
> If someone wants `0` to be the y-axis minimum, that's what the `Min` property is for. They can specify `Min: 0`.

That makes sense a lot of sense, PR is updated now!

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -86,12 +86,24 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
+    min: y_axis?.min,
     max: y_axis?.max,
   };
 
+  if (yAxis.min === undefined) {
+    // sets minimum axis label relative to data instead of zero
+    yAxis.min = (value) => {
+      // https://echarts.apache.org/en/option.html#yAxis.min
+      if (value.min <= 1) {
+        return 0;
```
sjcobb, line 100: Open to removing this, but before I added this condition, the origin looked odd for PercentDecimal cases or when the min value was 1, ex of before I added it:
### Before
image

### After
image

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -207,7 +219,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
           color: thresholdLineColor,
           value:
             thresholds.mode === 'Percent'
-              ? convertPercentThreshold(step.value, graphData.timeSeries, yAxis.max, yAxis.min)
+              ? convertPercentThreshold(step.value, graphData.timeSeries, y_axis?.max, y_axis?.min)
```
sjcobb, line 222: To fix param types, I changed to use min and max from panel spec here since they will always be a number (if defined)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -86,12 +86,24 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
 
+  if (yAxis.min === undefined) {
+    // sets minimum axis label relative to data instead of zero
+    yAxis.min = (value) => {
+      // https://echarts.apache.org/en/option.html#yAxis.min
+      if (value.min <= 1) {
+        return 0;
+      }
+      // allows for padding between origin and first series
+      return value.min * 0.8;
```
sjcobb, line 103: Picked `0.8` arbitrarily to give just enough padding, but we'll probably need to tweak this

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -86,12 +86,24 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   const visual = merge({}, DEFAULT_VISUAL, props.spec.visual);
 
   // convert Perses dashboard format to be ECharts compatible
-  const yAxis = {
+  const yAxis: YAXisComponentOption = {
     show: y_axis?.show ?? DEFAULT_Y_AXIS.show,
-    min: y_axis?.min, // leaves min and max undefined by default to let ECharts calcualate
+    min: y_axis?.min,
     max: y_axis?.max,
   };
```
juliepagano, line 94: This really would be a nice case to add a few stories in storybook, so we could get visual diffs for this behavior, but the time series data mocking util I wrote is still super-simple and locked into the e2e package. I'm going to take a little time today to see if I can pull those out into something we'd be able to use across both storybook and playwright.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
juliepagano: Seems reasonable to me!

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
juliepagano: Could you add that context to the comment, so future us knows this isn't some special number, and it's safe to tweak.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -86,12 +86,24 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
+    yAxis.min = (value) => {
+      // https://echarts.apache.org/en/option.html#yAxis.min
+      if (value.min <= 1) {
+        return 0;
+      }
+      // allows for padding between origin and first series
+      return value.min * 0.8;
+    };
+  }
+
```
juliepagano, line 106: [meandering thought about this component] There's a lot of utilities inside this component that make decisions about processing data and setting config. I'm wondering if it might be helpful to pull some of these out into utility functions that the component calls. It'd probably make the code a little more readable, it would make it easier to write some simple tests for the logic, and it might make it easier for us to eventually make some of that logic more accessible for embeddability with the `LineChart`. 

[specific, optional ask for this PR] Not asking you to pull them all out in this PR, but maybe it would be useful to pull out the bits you touched, so you can add a simple unit test. Thoughts?

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
juliepagano: It took me a little looking at `convertPercentThreshold` to understand why you'd use `y_axis` here instead of `yAxis`. It might be worth putting a comment about it, because I could totally see someone seeing it in the future and thinking it was a mistake and trying to switch them back.

juliepagano (approved):
A few bits of largely optional feedback, but generally looks good to me!

eunicorn (approved):
I just watched Sam's interview and he literally mentioned this! Awesome 💯

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: We also don’t have any panel stories yet right, mostly just component-level? I wasn’t sure if there was some other blocker to getting a full panel working, I can give it a shot though!

eunicorn: I added a panel story in storybook in one of my feature branches, not merged yet though

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
cndonovan: Could be nice to make a named constant too? For easier tweaking + knowing what it means

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
cndonovan: Yeah good call. These two charts definitely look better with the adjustment. We can plan to keep an eye out when we see this on charts in the wild (to continue to confirm it looks good). 

In general, I feel like it's best when we aim for the best for the overall chart viewing experience without the user having to do anything, so I like the call.

cndonovan (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Added a story with examples of customizing a panel's y_axis spec and used a query that showed how the new origin logic works.

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
```diff
@@ -130,3 +134,28 @@ function findMax(timeSeries: EChartsTimeSeries[]) {
...
+    max: inputAxis?.max,
+  };
+
+  if (inputAxis?.min === undefined) {
+    // Sets minimum axis label relative to data instead of zero.
+    yAxis.min = (value) => {
+      // https://echarts.apache.org/en/option.html#yAxis.min
+      if (value.min <= 1) {
+        return 0;
+      }
```
cndonovan, line 154: Oh just had a thought that this could be weird if the chart goes negative!! 🐛 ?

cndonovan: (In other words, if the minimum is -1000 or something, we wouldn't want to chart to cut off at zero...)

sjcobb: Oh wow, yeah you're right, thanks for catching this! It's fixed now:

image

PR #1048, opened 2023-03-29
[IGNORE] bumps @prometheus-io/codemirror-promql to v0.43.0

Nexucis:
This PR update the dependence `@prometheus-io/codemirror-promql` to the last version and remove the dependency to `@prometheus-dev/lezer-promql` which is already brought by `codemirror-promql`

### DIFF
--- a/ui/prometheus-plugin/package.json
+++ b/ui/prometheus-plugin/package.json
@@ -34,8 +34,7 @@
     "@perses-dev/components": "0.25.0",
     "@perses-dev/core": "0.25.0",
     "@perses-dev/plugin-system": "0.25.0",
-    "@prometheus-io/codemirror-promql": "^0.40.5",
-    "@prometheus-io/lezer-promql": "^0.37.0",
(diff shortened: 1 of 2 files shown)


ui/prometheus-plugin/package.json
```diff
@@ -34,8 +34,7 @@
     "@perses-dev/components": "0.25.0",
     "@perses-dev/core": "0.25.0",
     "@perses-dev/plugin-system": "0.25.0",
-    "@prometheus-io/codemirror-promql": "^0.40.5",
-    "@prometheus-io/lezer-promql": "^0.37.0",
```
saminzadeh, line 38: Do we not use lezer-promql anymore?

saminzadeh (commented):
Great i was just going to do that.

ui/prometheus-plugin/package.json
Nexucis: not from what I find in the code at least :). We only use codemirror-promql.

saminzadeh (approved):
Thanks for doing this!

PR #1050, opened 2023-03-29
[DRAFT] Annotations 

saminzadeh:
Below is a draft PR and proposal for adding Annotation support to Perses. It's design is very similar to Variables.

**Context**:
Annotations are used to add contextual information to a graph or panel. They allow users to highlight specific events, trends, or anomalies in their data by overlaying markers on a graph or panel.

**Types of Annotations:**
Annotations can be single-point-in-time or have a start time and end time. Here is an example of what they look like:
image

In terms of Perses, there will be 2 core types of annotation definitions: **Static** and **Query**

**StaticAnnotations** 
These annotations are defined directly in the definition. They are useful in the framework use-case or if the data is coming from a datasource external to the plugin ecosystem.

```tsx
{
   kind: "StaticAnnotation",
   spec: {
      annotations: [{ timestamp: 1680126896025, color: "red", title: "deployment" }]
   }
}
```

**QueryAnnotations**
Query annotations allow developers to define a plugin that can return an async list of annotations from an external source. For example, a [LaunchDarkly](https://launchdarkly.com/) annotation plugin might look something like this.
```tsx
{
  kind: 'QueryAnnotation',
  spec: {
    plugin: {
      kind: "LaunchDarklyAnnotationQuery",
      spec: {
        filters: [{ namespace: 'my-app'}]
      }
    }
  },
}
```

Plugins will take the query definition and return a list of `AnnotationEvent[]`.  
```tsx 
type AnnotationEvent = {
  timestamp: number;
  endTimestamp?: number;
  color?: string;
  title?: string;
};
```

To define annotations, we will update the dashboard spec to include a top level `annotations` property that will accept a list of `AnnotationDefinition[]`. Here is an example

```
{
   
   panels: {...}
   variables: { ... }
   annotations: [{  
     kind: "QueryAnnotation",
     spec: { 
        plugin: {
           kind: "LaunchDarklyAnnotationQuery",
           spec: { ... }
       },
     }      
   }
   ...more
}
```

# Checklist

### DIFF
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -67,7 +67,7 @@ export interface LineChartProps {
   unit?: UnitOptions;
   grid?: GridComponentOption;
   legend?: LegendComponentOption;
-  annotations: AnnotationEvent[];
+  annotations?: AnnotationEvent[];
   onDataZoom?: (e: ZoomEventData) => void;
   onDoubleClick?: (e: MouseEvent) => void;
 }
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -250,7 +250,6 @@ export function LineChart({
       grid,
       legend,
     };
-    console.log(option);
 
     return option;
   }, [data, yAxis, unit, grid, legend, noDataOption, timeZone]);
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -34,8 +34,9 @@ import {
   LegendComponent,
 } from 'echarts/components';
 import { CanvasRenderer } from 'echarts/renderers';
+import { AnnotationEvent } from '@perses-dev/core';
 import { EChart, OnEventsType } from '../EChart';
-import { EChartsDataFormat, OPTIMIZED_MODE_SERIES_LIMIT } from '../model/graph';
+import { EChartsDataFormat, EChartsTimeSeries, OPTIMIZED_MODE_SERIES_LIMIT } from '../model/graph';
 import { UnitOptions } from '../model/units';
 import { useChartsTheme } from '../context/ChartsThemeProvider';
 import { TimeSeriesTooltip } from '../TimeSeriesTooltip';
@@ -66,11 +67,22 @@ export interface LineChartProps {
   unit?: UnitOptions;
   grid?: GridComponentOption;
   legend?: LegendComponentOption;
+  annotations: AnnotationEvent[];
   onDataZoom?: (e: ZoomEventData) => void;
   onDoubleClick?: (e: MouseEvent) => void;
 }
 
-export function LineChart({ height, data, yAxis, unit, grid, legend, onDataZoom, onDoubleClick }: LineChartProps) {
+export function LineChart({
+  height,
+  data,
+  yAxis,
+  unit,
+  grid,
+  legend,
+  onDataZoom,
+  onDoubleClick,
+  annotations = [],
+}: LineChartProps) {
   const chartsTheme = useChartsTheme();
   const chartRef = useRef<EChartsInstance>();
   const [showTooltip, setShowTooltip] = useState<boolean>(true);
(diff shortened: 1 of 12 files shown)


ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,52 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Definition } from './definitions';
+import { Display } from './display';
+
+export type AnnotationEvent = {
+  timestamp: number;
+  endTimestamp?: number;
```
sjcobb, line 19: Will they always be Unix timestamps--I guess, if not, it would have been converted already in the plugin?

Also, is the equivalent for Grafana, just 'from' and 'to'? If so, another option is 'start' and 'end' kind of like how we did the AbsoluteTimeRange query params in TimeRangeProvider

ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,52 @@
...
+  color?: string;
+  title?: string;
+};
+
+export interface AnnotationSpec {
+  name: string;
+  display?: Display & {
```
sjcobb, line 26: Nice, so this is importing the same `Display` type that variables use right?

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -133,18 +145,89 @@ export function LineChart({ height, data, yAxis, unit, grid, legend, onDataZoom,
 
     const rangeMs = data.rangeMs ?? getDateRange(data.xAxis);
 
+    let series = data.timeSeries;
+
+    const startTime = data.xAxis[0];
+    const endTime = data.xAxis[data.xAxis.length - 1];
+
+    if (annotations.length > 0) {
```
sjcobb, line 153: Yeah I need to just move the thresholds mapping into the chart this way too

ui/plugin-system/src/runtime/annotations.ts
```diff
@@ -0,0 +1,31 @@
...
+export type AnnotationSrv = {
+  annotations: AnnotationEvent[];
+};
+
+export const AnnotationContext = createContext<AnnotationSrv | undefined>(undefined);
+
+export function useAnnotationContext() {
+  const ctx = useContext(AnnotationContext);
+  if (ctx === undefined) {
+    throw new Error('No found AnnotationContext. Did you forget the AnnotationsProvider?');
```
sjcobb, line 26: I know we've done this with most of the others providers so far, but does AnnotationsProvider need need to be required or is there some way to fallback gracefully without the error?

ui/components/src/LineChart/LineChart.tsx
saminzadeh: Yeah basically this component should do most of the work. The dashboard level component should just plugin in the correct context

ui/core/src/model/annotations.ts
saminzadeh: https://github.com/sampoder/extensions/blob/e9290fd962be2392489505b76e42558e12683145/extensions/grafana/src/components/annotation/interface/index.ts#L1-L4

Grafana uses `timeEnd` - happy to change this as well. Although `start` `end` is a little vague without the context of time. I was going to use unix ms, so maybe a type to make that more clear.

ui/core/src/model/annotations.ts
saminzadeh: yup

ui/plugin-system/src/runtime/annotations.ts
saminzadeh: Yeah I guess I go back and forth on this. If we end up creating one provider that has all the presets I think this should be ok. If not, I think we should make it optional.

`<PersesProvider />` should setup the core providers

ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Definition } from './definitions';
+import { Display } from './display';
+
+export type AnnotationEvent = {
+  timestamp: number;
+  endTimestamp?: number;
+  color?: string;
```
eunicorn, line 20: is `color` here used for highlighting the chart where the annotation describes?  if it can be undefined, does that mean we need to add some color tokens to the chart theme?

ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,52 @@
...
+  title?: string;
```
eunicorn, line 21: nit: personally i think `description` makes more sense than `title`.  

Or can you have both `title` and `description`?

ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,52 @@
...
+  display?: Display & {
+    hidden?: boolean;
+  };
+}
+
+export interface StaticAnnotationDefinition extends Definition<StaticAnnotationSpec> {
+  kind: 'StaticAnnotation';
+}
+export interface StaticAnnotationSpec extends AnnotationSpec {
+  annotations: AnnotationEvent[];
```
eunicorn, line 35: maybe call this `annotationEvents` to be more clear if it's referring to event vs definition

ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// export interface QueryAnnotationDefinition<PluginSpec = UnknownSpec>
```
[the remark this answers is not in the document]
eunicorn: does that mean we will need a new plugin type for QueryAnnotation?

ui/core/src/model/annotations.ts
```diff
@@ -0,0 +1,52 @@
...
+  timestamp: number;
+  endTimestamp?: number;
+  color?: string;
+  title?: string;
+};
+
+export interface AnnotationSpec {
+  name: string;
+  display?: Display & {
+    hidden?: boolean;
```
eunicorn, line 27: can you hide annotations for a specific panel?

ui/core/src/model/annotations.ts
saminzadeh: Good call

ui/core/src/model/annotations.ts
saminzadeh: At the moment, just thinking about hiding for all vs showing for all. I think we can iterate and add this feature in down the road if it makes sense.

ui/core/src/model/annotations.ts
saminzadeh: Oh yeah not a bad idea, I think both make sense.

ui/core/src/model/annotations.ts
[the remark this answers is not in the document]
saminzadeh: Yeah exactly, so in my LaunchDarkly example `LaunchDarklyAnnotationQuery` would be a `AnnotationQuery` plugin kind.

ui/core/src/model/annotations.ts
saminzadeh: Yeah I was thinking we could have a default annotation color

saminzadeh:
In spirit of not leaving old PRs open too long, will close this and revisit

PR #1054, opened 2023-03-30
[IGNORE] add eslint rule to avoid duplicate imports

juliepagano:
Commit also fixes any duplicate imports caught by the rule.

# Notes for reviewers

I ran into this while working on the refactor for https://github.com/perses/perses/pull/1053. Removing duplicate imports makes these sorts of refactors easier and makes the code a little cleaner. I'm hoping this is a non-controversial eslint rule, but please let me know if you disagree with adding it!

# Checklist

### DIFF
--- a/ui/e2e/src/config/base.playwright.config.ts
+++ b/ui/e2e/src/config/base.playwright.config.ts
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { PlaywrightTestConfig, devices } from '@playwright/test';
+import { PlaywrightTestConfig, devices } from '@playwright/test';
 
 /**
  * This is the base playwright configuration that contains common settings.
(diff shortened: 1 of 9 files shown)


ui/e2e/src/config/base.playwright.config.ts
```diff
@@ -11,8 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { PlaywrightTestConfig } from '@playwright/test';
-import { devices } from '@playwright/test';
+import { PlaywrightTestConfig, devices } from '@playwright/test';
```
juliepagano, line 14: I don't think we get any benefit from the separate `import type` for playwright use cases. Guessing this happened as a side effect of clicking the "fix" in vscode and whatever it generated.

sjcobb (approved):
Oh wow, these were mostly me 😬
Thanks for fixing and adding the rule!

juliepagano:
> Oh wow, these were mostly me 😬 Thanks for fixing and adding the rule!

No worries! I think the other half were me. I'm used to having an eslint rule catch it for me, and it's really easy to accidentally do this if you have vscode "fix" a missing import for you.

PR #1056, opened 2023-03-30
[IGNORE] Add simple typedoc setup

juliepagano:
I have been struggling to keep track of what Perses exports from the various UI packages, which can make it tricky to figure out breaking changes when things are refactored. Typedoc does a nice job evaluating the entry point and creating documentation, so that seemed like a low effort way to get this information.

In the short term, storing these as artifacts in github actions runs. Longer-term, we could potentially publish these in github pages alongside storybook (which is only good at documenting React components) to provide documentation for the non-component utilities, types, etc. that Perses provides.

It is notable that typedoc isn't great at handling cross-package dependencies at this time, so there are some limitations to what we're getting. We can watch this issue to see when this is improved. https://github.com/TypeStrong/typedoc/issues/1835

# Checklist

# Screenshots

image
image
image

### DIFF
--- a/ui/components/package.json
+++ b/ui/components/package.json
@@ -57,5 +57,8 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "typedoc": {
+    "entryPoint": "./src/index.ts"
+  }
 }
--- a/ui/typedoc.json
+++ b/ui/typedoc.json
@@ -1,0 +1,7 @@
+{
+  "entryPointStrategy": "packages",
+  "entryPoints": [
+    "."
+  ],
+  "out": "typedoc"
+}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,6 +77,25 @@ jobs:
           name: storybook
           path: ui/storybook/storybook-static
   
+  build-typedoc:
+    name: "build-typedoc"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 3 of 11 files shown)


ui/components/package.json
```diff
@@ -57,5 +57,8 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "typedoc": {
+    "entryPoint": "./src/index.ts"
+  }
```
juliepagano, line 63: This annotation tells typedoc to look at the TypeScript entry point. By default, it will try to look at whatever is defined in `main`, which is the built code. It will throw this error if it uses that.

> error Entry point "/Users/juliepagano/src/perses/perses/ui/components/dist/cjs/index.js" does not appear to be built by/included in the tsconfig found at "/Users/juliepagano/src/perses/perses/ui/components/tsconfig.json"

ui/typedoc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "entryPointStrategy": "packages",
+  "entryPoints": [
+    "."
+  ],
```
juliepagano, line 5: This tells it to treat this project like a monorepo and look at the workspaces in the `package.json` in the current directory. That means we don't have to explicitly list all the packages in here. (I'd annotate this in the file, but JSON doesn't support comments).

juliepagano:
That failing mysql test is a flaky test, not anything broken by my PR.

eunicorn (approved):

PR #1057, opened 2023-03-31
[BUGFIX] Add missing datasource selector in the variable schemas

Nexucis:
It looks like the schemas of the variables didn't contain the datasource selector.

As a side note, there is no way to select a datasource in the UI for the Prometheus Variables. I guess this is a miss.


schemas/queries/prometheus/query.cue
```diff
@@ -21,8 +21,9 @@ spec: {
 	plugin: {
 		kind: "PrometheusTimeSeriesQuery"
 		spec: close({
-			datasource: {
+			datasource?: {
 				kind: prometheus.kind
+				name?: string
```
AntoineThebaud, line 26: this doesnt add anything new compared to base def

Nexucis: ah yeah you are right. I forgot about the bas def. My bad

AntoineThebaud (approved):

PR #1059, opened 2023-03-31
[ENHANCEMENT] always show info icon when panel has a description

juliepagano:
Prior to this change, the info icon was only shown when the user was hovering over the panel, which significantly reduces discoverability of the description information, especially on larger dashboards.

After this change, the info icon is always shown in "view" mode when a panel has a description. The hover over the icon to show the description in a tooltip continues to behave as before.

Making this change based on a request from @foxbat07  to improve usability.

# Checklist

# Screenshots

## Before 

image

## After

image

### DIFF
--- a/ui/dashboards/src/components/Panel/PanelHeader.tsx
+++ b/ui/dashboards/src/components/Panel/PanelHeader.tsx
@@ -31,10 +31,9 @@ export interface PanelHeaderProps extends Omit<CardHeaderProps, OmittedProps> {
     onDuplicatePanelClick: () => void;
     onDeletePanelClick: () => void;
   };
-  isHovered: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


ui/dashboards/src/components/Panel/PanelHeader.tsx
```diff
@@ -31,10 +31,9 @@ export interface PanelHeaderProps extends Omit<CardHeaderProps, OmittedProps> {
     onDuplicatePanelClick: () => void;
     onDeletePanelClick: () => void;
   };
-  isHovered: boolean;
```
juliepagano, line 34: Taking away this prop since this was the only reason we needed it. We don't export `PanelHeader`, so this should not be a breaking change.

eunicorn (approved):

PR #1060, opened 2023-03-31
[BREAKINGCHANGE] Consistent query editor for all panel types & display full panel json in panel editor

eunicorn:
This PR depends on https://github.com/perses/perses/pull/1032

**CHANGES**

- Queries stay the same when switching between different panel types
- Display full panel definition in the json editor

https://user-images.githubusercontent.com/9817826/229634344-6132b6ea-7b9a-465a-a230-2f6c1806c210.mov

- Update PanelPlugin type
  -  remove `PanelQueryEditorComponent`
  -  add `hideQueryEditor` field 
 ```
  /**
   * React components for custom tabs
   */
  /**
   * If true, query editor will be hidden for panel plugin
   * @default false
   */
  hideQueryEditor?: boolean;
}
```
**KNOWN ISSUES**
- Although you can add multiple queries for stat and gauge charts now, it is only displaying the first query result

### CODE REVIEW

File: ui/plugin-system/src/components/TimeSeriesQueryEditor.tsx [hunk 1]
```diff
@@ -11,24 +11,137 @@
...
 import { PluginEditor, PluginEditorProps } from './PluginEditor';
 
+const DEFAULT_QUERY_PLUGIN_TYPE = 'TimeSeriesQuery';
+
+export interface TimeSeriesQueryEditorProps {
+  queries: TimeSeriesQueryDefinition[];
+  onChange: (queries: QueryDefinition[]) => void;
+}
+
+export function TimeSeriesQueryEditor(props: TimeSeriesQueryEditorProps) {
```
eunicorn, line 32: This used to be `TimeSeriesChartQueryEditor`

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
```diff
@@ -24,37 +24,41 @@ import {
...
   const panelGroups = useListPanelGroups();
-
-  const [name, setName] = useState(initialValues.name);
-  const [description, setDescription] = useState(initialValues.description);
-  const [groupId, setGroupId] = useState(initialValues.groupId);
-  const [kind, setKind] = useState(initialValues.kind);
-  const [spec, setSpec] = useState(initialValues.spec);
-  const [queries, setQueries] = useState(initialValues.queries);
+  const [groupId, setGroupId] = useState(initialGroupId);
+  const { panelDefinition, setName, setDescription, setQueries, setPlugin, setPanelDefinition } =
```
eunicorn, line 49: added `usePanelEditor` to help keep track of all the states we need since most child components need to know the full panel definition now

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  const { panelOptionsEditorComponents, hideQueryEditor } = plugin as PanelPlugin;
+  let tabs: OptionsEditorTabsProps['tabs'] = [];
+
+  if (!hideQueryEditor && panelDefinition.spec.queries !== undefined) {
+    // Since we only support TimeSeriesQuery for now, we will always show a TimeSeriesQueryEditor
+    tabs.push({
+      label: 'Query',
+      content: <TimeSeriesQueryEditor queries={panelDefinition.spec.queries} onChange={onQueriesChange} />,
```
eunicorn, line 56: Decided to add `PanelSpecEditor` instead modifying `PluginSpecEditor` directly because otherwise there is a circular dependency between `PluginSpecEditor` and `TimeSeriesQueryEditor`

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -32,11 +32,14 @@ export interface PluginEditorProps extends Omit<BoxProps, OmittedMuiProps> {
...
 
 type PreviousSpecState = Record<string, Record<string, UnknownSpec>>;
+type HideQueryEditorState = Record<string, boolean>;
 
 /**
  * Props needed by the usePluginEditor hook.
  */
-export type UsePluginEditorProps = Pick<PluginEditorProps, 'pluginType' | 'value' | 'onChange'>;
+export type UsePluginEditorProps = Pick<PluginEditorProps, 'pluginType' | 'value' | 'onChange'> & {
+  onHideQueryEditorChange?: (isHidden: boolean) => void;
```
eunicorn, line 41: This change is needed so the `PanelSpecEditor` knows when to show/hide the query editor

pkg/model/api/v1/dashboard.go
```diff
@@ -27,7 +27,7 @@ import (
 type PanelSpec struct {
 	Display common.Display `json:"display" yaml:"display"`
 	Plugin  common.Plugin  `json:"plugin" yaml:"plugin"`
-	Queries []Query  `json:"queries" yaml:"queries"`
+	Queries []Query        `json:"queries,omitempty" yaml:"queries, omitempty"`
```
juliepagano, line 30: My go is super weak -- does this exclude queries entirely if they're not set, turn them into an empty array, or some other thing?

saminzadeh, line 30: Is there a semantic difference between `queries: []` vs no queries? Double checking to make sure `omitempty` is what we expect

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
juliepagano: I've got a handful of questions on this bit to make sure I'm following how all these pieces fit together...

With this change, does `PluginSpecEditor` handle non-panel plugin editor rendering and `PanelSpecEditor` handles panel plugin editor rendering or is it more complex than that?

Would it makes sense to put `PanelSpecEditor` in `plugin-system` where the other plugin rendering components live?

Is there a potential future where we need to support non-timeseries queries? If so, how will that fit with how this is currently being organized? I'm guessing we could add an additional value in the spec that we could use to conditionally render the right thing here, but was curious how you're thinking about it.

ui/plugin-system/src/components/PluginSpecEditor.tsx
```diff
@@ -42,11 +40,9 @@ export function PluginSpecEditor(props: PluginSpecEditorProps) {
   }
 
   if (pluginType === 'Panel') {
```
juliepagano, line 42: Does the code in this `Panel` conditional still get used at all or is it all handled by `PanelSpecEditor` now?

ui/plugin-system/src/components/TimeSeriesQueryEditor.tsx
saminzadeh: I think this is ok for now, but eventually this will probably just change to `QueryEditor`. Once we have multiple types, you may want to have one query with logs and one with metrics so it should support more than `TimeSeriesQuery`

ui/dashboards/src/components/PanelDrawer/usePanelEditor.tsx
```diff
@@ -0,0 +1,82 @@
...
+      setQueriesState(queries);
+    },
+    [setName, setDescription, setPlugin, setQueriesState]
+  );
+
+  /**
+   * If hideQueryEditor is true, set panelDefinition.spec.queries to undefined.
+   * If hideQueryEditor is false and query is undefined, set panelDefinition.spec.queries to previous queries.
+   */
+  const setQueries = useCallback(
```
juliepagano, line 47: [optional suggestion] Is it worth writing a small unit test for this logic?

ui/dashboards/src/context/DashboardProvider/common.ts
```diff
@@ -30,3 +32,37 @@ export function generateId() {
...
+    spec: {
+      display: {
+        name: '',
+        description: undefined,
+      },
+      plugin: {
+        kind: defaultPanelKind ?? '',
+        spec: defaultPanelSpec ?? {},
+      },
+      queries: [
```
juliepagano, line 53: This is another bit that could get interesting if we ever support not-time-series queries.

ui/plugin-system/src/model/panels.ts
```diff
@@ -24,14 +24,15 @@ export type PanelOptionsEditorComponent<T> = Pick<OptionsEditorTab, 'label'> & {
...
-  PanelQueryEditorComponent?: React.ComponentType<QueryEditorProps<Spec>>;
   /**
    * React components for custom tabs
    */
   panelOptionsEditorComponents?: Array<PanelOptionsEditorComponent<Spec>>;
+  /**
+   * If true, query editor will be hidden for panel plugin
+   * @default false
+   */
+  hideQueryEditor?: boolean;
```
saminzadeh, line 35: Nice. I suppose this is more explicit than `supportedQueryInterfaces`

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
```diff
@@ -41,18 +38,6 @@ export interface GaugeChartOptions {
  */
 export function createInitialGaugeChartOptions(): GaugeChartOptions {
   return {
-    // TODO: How do you represent an initially empty/unset graph query?
```
juliepagano, line 44: Yay, deleted code! 🎉

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -45,16 +48,17 @@ export type UsePluginEditorProps = Pick<PluginEditorProps, 'pluginType' | 'value
  * kind back.
  */
 export function usePluginEditor(props: UsePluginEditorProps) {
-  const { pluginType, value } = props;
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const { pluginType, value, onHideQueryEditorChange = () => {} } = props;
```
juliepagano, line 52: Could you include a short comment explaining why you have an empty function here and need the eslint-disable? I find that's usually helpful for future us as the code grows.

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -104,11 +122,19 @@ export function usePluginEditor(props: UsePluginEditorProps) {
         kind: nextKind,
         spec: previousState,
       });
-      return;
+      // return;
```
juliepagano, line 125: Is this `return` intentionally commented out?

ui/plugin-system/src/components/TimeSeriesQueryEditor.tsx
```diff
@@ -11,24 +11,137 @@
...
+  const { defaultPluginKinds } = usePluginRegistry();
+  const defaultTimeSeriesQueryKind = defaultPluginKinds?.[DEFAULT_QUERY_PLUGIN_TYPE] ?? '';
+
+  const { data: defaultQueryPlugin } = usePlugin(DEFAULT_QUERY_PLUGIN_TYPE, defaultTimeSeriesQueryKind, {
+    useErrorBoundary: true,
+    enabled: true,
+  });
+
+  // State for which queries are collapsed
+  // TODO: Would be easier if we had IDs for queries.
```
juliepagano, line 44: I could see it being nice to add identifiers at some point for a variety of use cases, but you're already doing so much in these refactors that it would be nice not to have to fix that too. 😄

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  if (plugin === undefined) {
+    throw new Error(`Missing implementation for panel plugin with kind '${kind}'`);
+  }
+
+  const { panelOptionsEditorComponents, hideQueryEditor } = plugin as PanelPlugin;
+  let tabs: OptionsEditorTabsProps['tabs'] = [];
+
+  if (!hideQueryEditor && panelDefinition.spec.queries !== undefined) {
```
saminzadeh, line 52: If a user accidentally forgets to add `queries` property, is the editor hidden? I think it should just create an empty `[]` in that case. Since people edit dashboard config by hand, there will be use cases like this.

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
```diff
@@ -0,0 +1,73 @@
...
+  usePlugin,
+  PanelPlugin,
+  OptionsEditorTabsProps,
+  TimeSeriesQueryEditor,
+  OptionsEditorTabs,
+} from '@perses-dev/plugin-system';
+
+export interface PanelSpecEditorProps {
+  panelDefinition: PanelDefinition;
+  onQueriesChange: (queries: QueryDefinition[]) => void;
```
saminzadeh, line 26: I kinda wonder if there should just be one:
`onPanelDefinitionChange` ?

ui/plugin-system/src/components/TimeSeriesQueryEditor.tsx
```diff
@@ -11,24 +11,137 @@
...
+    });
+  };
+
+  return (
+    <Stack spacing={1}>
+      <Button variant="contained" startIcon={<AddIcon />} sx={{ marginLeft: 'auto' }} onClick={handleQueryAdd}>
+        Add Query
+      </Button>
+      {queries.map((query: TimeSeriesQueryDefinition, i: number) => (
+        <Stack key={i} spacing={1}>
```
juliepagano, line 108: [optional suggestion] Would it make sense to turn the content of this loop into a reusable component? This file is pretty big & complex and that might help with maintenance/readability.

juliepagano (commented):
First pass with a bunch of questions.

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
eunicorn: all valid questions!! 

> With this change, does `PluginSpecEditor` handle non-panel plugin editor rendering and `PanelSpecEditor` handles panel plugin editor rendering or is it more complex than that?

I think it makes sense for `PluginSpecEditor` to handle non-panel plugins while `PanelSpecEditor` handles only panel plugin because `PanelPlugin` type is already different from other plugin types. For instance `PanelPlugin` requires `panelOptionsEditorComponents` which is an array of react components with labels where as other plugins takes in `OptionsEditorComponent`.

However, for now i only removed the part of the code in `PluginSpecEditor` that introduced the circular dependency because I wasn't sure if I was missing some custom panel plugin use case.

This is definitely open for discussion!

>  Would it makes sense to put PanelSpecEditor in plugin-system where the other plugin rendering components live?

Yeah i can move it to plugin-system. The reason why i put it in dashboards package was because I wasn't sure if its confusing to have both PanelSpecEditor and PluginSpecEditor. I think it'll be more clear once we know the answer to the question above!

> Is there a potential future where we need to support non-timeseries queries? If so, how will that fit with how this is currently being organized? I'm guessing we could add an additional value in the spec that we could use to conditionally render the right thing here, but was curious how you're thinking about it.

I think eventually we need to update TimeSeriesQueryEditor to be a generic QueryEditor that supports other metrics like what Shan said in another comment. Also since the framework code knows what queries are being run, i think we can conditionally render the different query editors based on the query type.

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
eunicorn: So that was my initial approach, but then I was running into issues with syncing the different changes happening in `PanelEditorForm`. For instance, showing/hiding the query editor when switching the panel types requires additional logic to save the last used queries. Overall i find this approach much more clear

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
eunicorn: yes! i should probably delete this because now it needs to continue and check if it should trigger `onHideQuery`

pkg/model/api/v1/dashboard.go
eunicorn: `omitempty` will remove the `queries` field from object if it's an empty array/undefined. Anyways I removed this because i realized it's not correct. Good call

ui/dashboards/src/components/PanelDrawer/usePanelEditor.tsx
eunicorn: done!

ui/dashboards/src/context/DashboardProvider/common.ts
eunicorn: i updated this so `queries` can be an empty array

ui/plugin-system/src/components/TimeSeriesQueryEditor.tsx
eunicorn: converted this to `TimeSeriesQueryInput`

ui/plugin-system/src/components/TimeSeriesQueryEditor.tsx
eunicorn: +1

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
eunicorn: I moved `PanelSpecEditor` back to plugin-system because it doesn't really make sense to have it inside dashboards/components/PanelDrawer now that i think about it more...

ui/dashboards/src/components/PanelDrawer/PanelSpecEditor.tsx
eunicorn: good call, updated this so we only hide query editor if `hideQueryEditor` is true

pkg/model/api/v1/dashboard.go
```diff
@@ -27,7 +27,7 @@ import (
 type PanelSpec struct {
 	Display common.Display `json:"display" yaml:"display"`
 	Plugin  common.Plugin  `json:"plugin" yaml:"plugin"`
-	Queries []Query  `json:"queries" yaml:"queries"`
+	Queries []Query        `json:"queries" yaml:"queries"`
```
Nexucis, line 30: since it's an optional field, it should be marked by the tag "omitempty"
```suggestion
	Queries []Query        `json:"queries,omitempty" yaml:"queries,omitempty"`
```

juliepagano:
Is this ready for another pass or are you still working on feedback?

eunicorn:
> Is this ready for another pass or are you still working on feedback?

ready for another pass! :)

pkg/model/api/v1/dashboard.go
juliepagano: Gotcha! That seems reasonable to me, but I'll defer to @Nexucis to 👍  the backend-y bits, since it's less my area of expertise.

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -45,16 +48,17 @@ export type UsePluginEditorProps = Pick<PluginEditorProps, 'pluginType' | 'value
  * kind back.
  */
 export function usePluginEditor(props: UsePluginEditorProps) {
-  const { pluginType, value } = props;
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const { pluginType, value, onHideQueryEditorChange = () => {} } = props; // setting onHideQueryEditorChange to empty functio here because useEvent requires a function
```
juliepagano, line 52:
```suggestion
  const { pluginType, value, onHideQueryEditorChange = () => {} } = props; // setting onHideQueryEditorChange to empty function here because useEvent requires a function
```

typo

ui/plugin-system/src/components/PluginSpecEditor.tsx
juliepagano: Bump on this question. Guessing it got lost in the sea of other feedback.

ui/plugin-system/src/model/panels.ts
juliepagano: Something like `supportedQueryInterfaces` would be potentially more future-proof for if we support additional types of queries. 🤔 

Although I think this is fine for the sake of Eunice being able to move the refactor forward. We can cross that bridge when we get to it, especially knowing it could be pretty far away, since we're pretty focused on time series data from prometheus.

juliepagano (commented):
A few small bits of feedback and a bump for one that I think got lost in the giant amount of comments in the last round. Also noticed there's a broken test in the backend code. Wondering if one of the specs didn't get adjusted quite right or something like that.

ui/plugin-system/src/components/PluginSpecEditor.tsx
eunicorn: oops, updated this to throw an error instead since it should be handled by `PanelSpecEditor` now

eunicorn:
> Also noticed there's a broken test in the backend code. Wondering if one of the specs didn't get adjusted quite right or something like that.

These tests should be fixed with the changes in these two PRs: 
- https://github.com/perses/perses/pull/1072
- https://github.com/perses/perses/pull/1077

eunicorn:
@juliepagano @saminzadeh if one of you could take a final look, that would be great 🙏  thanks!

juliepagano (approved):
LGTM! 🎉

pkg/model/api/v1/dashboard.go
Nexucis: There is different thing behind the tag `omitempty` in Go.

Technically speaking about what does it do:
- when the struct will be marshalled in JSON, if the array is empty then it won't appear in the final JSON.
- when the JSON is unmarshalled in the struct, if the key doesn't appear then the array is not initialised in the struct.

Sementically speaking, and because of the behaviour described above, I think that's a nice way to indicate this field is optional.

PR #1063, opened 2023-04-03
[WIP] Date format adjustment to round to nearest five minutes 

sjcobb:
**DONOTMERGE** 
_(TBD on approach, what's described below needs to change)_

Adjust xAxis date formatter to show whole values, rounded to nearest five minutes

## Before 
image

image

## After

image

image

# Checklist

### DIFF
--- a/ui/components/src/LineChart/utils.ts
+++ b/ui/components/src/LineChart/utils.ts
@@ -67,7 +67,7 @@ export function getDateRange(data: number[]) {
 /*
  * Determines time granularity for axis labels, defaults to hh:mm
  */
-export function getFormattedDate(value: number, rangeMs: number, timeZone?: string) {
+export function getFormattedDate(value: string, rangeMs: number, timeZone?: string) {
   const dateFormatOptions: Intl.DateTimeFormatOptions = dateFormatOptionsWithTimeZone(
     {
       hour: 'numeric',
@@ -95,7 +95,7 @@ export function getFormattedDate(value: number, rangeMs: number, timeZone?: stri
   }
 
   // remove comma when month / day present
-  return DATE_FORMAT.format(value).replace(/, /g, ' ');
+  return DATE_FORMAT.format(Number(value)).replace(/, /g, ' ');
 }
 
 /*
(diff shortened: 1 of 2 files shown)


ui/components/src/LineChart/utils.ts
```diff
@@ -75,17 +76,26 @@ export function getFormattedDate(value: number, rangeMs: number, timeZone?: stri
...
+
+  if (roundToNearestFiveMinutes === true) {
+    // https://date-fns.org/v2.29.3/docs/roundToNearestMinutes
+    const roundedValue = roundToNearestMinutes(Number(value), { nearestTo: 5, roundingMethod: 'round' });
+    return DATE_FORMAT.format(roundedValue).replace(/, /g, ' ');
+  }
+
   // remove comma when month / day present
-  return DATE_FORMAT.format(value).replace(/, /g, ' ');
+  return DATE_FORMAT.format(Number(value)).replace(/, /g, ' ');
```
cndonovan, line 98: Pulled down the PR and played with this branch a little

One thing I realized is that this means that the underlying, actual timestamp will be different from the x-axis label, because of the rounding 

So, when I hover on a vertical x-axis tick, the x-axis label says "4:45", but then the cursor reveals that the actual timestamp is "4:42:45"

I think if I was a user, and the time reflected by the x-axis label and the time reflected by the cursor were not in sync, I would think something was off 

It seems like we would want to try to show the actual data at the actual time, though I'm not sure how feasible that is 🤔 In other words, if the x-axis label is "4:45", then try to show the data that corresponds to 4:45...

Or.... just update the tooltip with the same rounding, but that seems like could really lead to some bad issues for users who are debugging and need actual, truthful timestamps...

cndonovan: Anyway, excited for this improvement, thanks for grabbing it

sjcobb: All good points, yeah I wasn’t 100% sure about this solution and you articulated it really well why, need to keep thinking on this 🫤

sjcobb:
Going to try fixing our suggested step interval first, I think this might be a symptom that it needs to be adjusted

PR #1064, opened 2023-04-03
[BUGFIX] Fix tooltip position on long dashboards

sjcobb:
Fixes the following tooltip positioning issue where on longer pages with lots of panels, there are certain cases where the tooltip shows way too high on the page, this PR adjust which x and y mouse coordinates are used to position the tooltip

## Before 

tooltip_pos_issue

## After

image

# Checklist

### DIFF
--- a/ui/components/src/TimeSeriesTooltip/tooltip-model.ts
+++ b/ui/components/src/TimeSeriesTooltip/tooltip-model.ts
@@ -32,10 +32,6 @@ export const TOOLTIP_DATE_FORMAT = new Intl.DateTimeFormat(undefined, {
 
 export const defaultCursorData = {
   coords: {
-    viewport: {
-      x: 0,
-      y: 0,
-    },
     plotCanvas: {
       x: 0,
       y: 0,
@@ -55,10 +51,6 @@ export const emptyTooltipData = {
 };
 
 export interface CursorCoordinates {
-  viewport: {
-    x: number;
-    y: number;
-  };
   page: {
     x: number;
     y: number;
@@ -100,10 +92,6 @@ export const useMousePosition = (): CursorData['coords'] => {
   useEffect(() => {
     const setFromEvent = (e: ZRRawMouseEvent) => {
       return setCoords({
-        viewport: {
-          x: e.clientX,
-          y: e.clientY,
-        },
         page: {
           x: e.pageX,
           y: e.pageY,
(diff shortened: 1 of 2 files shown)


ui/components/src/TimeSeriesTooltip/utils.ts
```diff
@@ -33,20 +33,20 @@ export function assembleTransform(
 
   const cursorPaddingX = 32;
   const cursorPaddingY = 16;
-  const x = mousePos.viewport.x;
-  let y = mousePos.viewport.y + cursorPaddingY;
+  const x = mousePos.page.x;
+  let y = mousePos.page.y + cursorPaddingY;
```
juliepagano, line 38: Summarizing to make sure I understand the problem: The issue is that the tooltip is in a portal that's absolutely positioned relative to the `body`. Previously, we were setting the positioning based on the viewport position of a panel, which will not work as expected when scrolled down the page. The fix positions the tooltip based on the position of the panel within the page. 

It might be worth putting a tiny comment somewhere in here for future us about this. Maybe something quick like "Tooltip is located in a `Portal` attached to the `body` and absolutely positioned."

ui/components/src/TimeSeriesTooltip/tooltip-model.ts
```diff
@@ -59,6 +59,10 @@ export interface CursorCoordinates {
     x: number;
     y: number;
   };
+  page: {
+    x: number;
+    y: number;
+  };
```
cndonovan, line 65: Do we still need the `viewport` data (answer might be yes, I don't know a ton about this area of the code)

sjcobb: yeah good catch, I was thinking we may possibly need it later but coming back to this, I'm not really sure when we ever would so went ahead and removed it

sjcobb:
> Just me being curious: How do we end up with the mouse positioning data? is it an echarts thing? does something query the DOM APIs at some point?

Yeah good questions! So it's this mousemove listener in useMousePosition [here](https://github.com/perses/perses/blob/v0.26.1/ui/components/src/TimeSeriesTooltip/tooltip-model.ts#L117), but the key ECharts feature that makes it all possible is in [getFocusedSeriesData](https://github.com/perses/perses/blob/v0.26.1/ui/components/src/TimeSeriesTooltip/focused-series.ts#L131-L137):
```javascript
const pointInGrid = chart.convertFromPixel('grid', pointInPixel);
```

More info about `convertFromPixel` is here: https://echarts.apache.org/en/api.html#echartsInstance.convertFromPixel

PR #1067, opened 2023-04-03
[BUGFIX] Add undef check to fix add panel showContent error

sjcobb:
Follow up to PR #1023 to add an undefined check when using __experimentalEChartsOptionsOverride to hide the custom TimeSeriesTooltip

## Before

image

## After

image

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.

PR #1068, opened 2023-04-04
[BUGFIX] miscellaneous fixes to address console errors

juliepagano:
Adding e2e test failures on console errors because it would have caught the bug found in https://github.com/perses/perses/pull/1067.

# Description for squash 

*Saving this for future me when I squash this PR*

This commit introduces functionality that fails Playwright-based e2e tests if console errors
are seen during the test because this is often a sign of a subtle bug we would not catch
otherwise. 

As part of introducing this, we need to fix some of the underlying console errors from 
libraries we use (e.g. MUI, React). In one case, we ignore an emotion error because it is
tricky to fix and focused on a use case we are not currently focused on (SSR). Fixes include:

- Fixing React errors related to a text input changing from uncontrolled to controlled because a
  text input could sometimes be `undefined`. These now fall back to `''` instead.
- Fixing a use case where `li` elements were not properly parented to an `ul` or `ol`.
- Fixing a React `key` error by setting `key` value on children in a list.
- Wrapping tooltip children in a `span` to avoid MUI errors when the child is a disabled button.

# Checklist

# Screenshots 

Example of what the error output looks like -- I temporarily turned back on erroring on that emotion error to produce this.
image

### DIFF
--- a/ui/app/src/components/DashboardList.tsx
+++ b/ui/app/src/components/DashboardList.tsx
@@ -16,7 +16,7 @@ import { useNavigate } from 'react-router-dom';
 import { Box, Divider, IconButton, List, ListItem, ListItemButton, ListItemText } from '@mui/material';
 import DeleteIcon from 'mdi-material-ui/DeleteOutline';
 import PencilIcon from 'mdi-material-ui/Pencil';
-import { useState } from 'react';
+import { useState, Fragment } from 'react';
 import { dashboardDisplayName } from '@perses-dev/core/dist/utils/text';
 import { DeleteDashboardDialog } from './DeleteDashboardDialog/DeleteDashboardDialog';
 import { RenameDashboardDialog } from './RenameDashboardDialog/RenameDashboardDialog';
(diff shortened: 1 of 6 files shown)


ui/app/src/components/DashboardList.tsx
```diff
@@ -49,31 +49,33 @@ function DashboardList(props: DashboardListProperties) {
       <List>
         {dashboardList.map((dashboard, i) => {
           return (
-            <>
+            <Fragment key={dashboard.metadata.name}>
```
juliepagano, line 52: Is this a safe value to use for a key? I _think_ they're unique?

ui/app/src/components/DashboardList.tsx
```diff
@@ -49,31 +49,33 @@ function DashboardList(props: DashboardListProperties) {
...
-                    <IconButton
-                      edge="end"
-                      aria-label="delete"
-                      onClick={() => onDeleteButtonClick(dashboard)}
-                      disabled={isRenameDashboardDialogStateOpened || isDeleteDashboardDialogStateOpened}
-                    >
-                      <DeleteIcon />
-                    </IconButton>
-                  </ListItem>
+                  <List>
```
juliepagano, line 59: Reparenting these so they `li` ends up in a list parent.

ui/components/src/InfoTooltip/InfoTooltip.tsx
```diff
@@ -41,10 +41,13 @@ export const InfoTooltip = ({
...
-  // support text with a wrapper div while avoiding breaking css on element
-  // children by unnecessarily wrapping them.
-  const formattedChildren = React.isValidElement(children) ? children : <div>{children}</div>;
+  // Wrap children in a span to cover the following use cases:
+  //  - Disabled buttons. MUI console.errors on putting these inside a tooltip.
+  //  - Non-element tooltip children (e.g. text). The tooltip needs something that
+  //   can have a ref as a child.
+  // We wrap in a `span` and not a `div` to minimize the impact on page layout
+  // and styles.
+  const wrappedChildren = <span>{children}</span>;
```
juliepagano, line 50: If you put a disabled button inside a MUI tooltip, it throws an error. Someone on the team tried to fix this previously by wrapping with a `div`, which I removed because it caused CSS issues not realizing it was fixing this other issue. They are now wrapped in a `span` which should avoid both the MUI error and the css issues. Also leaving a comment, so future me doesn't cleverly try to "fix" this again.

ui/components/src/ThresholdsEditor/ThresholdInput.tsx
```diff
@@ -47,7 +47,7 @@ export function ThresholdInput({
         key={key}
         inputRef={inputRef}
         type="number"
-        value={value === 0 ? undefined : value}
+        value={value === 0 ? '' : value}
```
juliepagano, line 50: Setting the input to `undefined` makes React think it's an uncontrolled component. It then throws an error about switching from uncontrolled to controlled. Instead falling back to an empty string, which should accomplish the same UX without the react error.

ui/e2e/src/fixtures/dashboardTest.ts
```diff
@@ -86,6 +86,18 @@ async function deleteDashboard(projectName: string, dashboardName: string) {
   return result;
 }
 
+// We want to throw on console errors that may mean the app is broken.
+// Some of the libraries we use (e.g. emotion) throw console errors we do not
+// care about at this time. We track those here, so they can be ignored.
+const IGNORE_CONSOLE_ERRORS = [
+  // See https://github.com/emotion-js/emotion/issues/1105
+  'potentially unsafe when doing server-side rendering',
```
juliepagano, line 94: We should probably fix this at some point, but I don't want to fix it as part of this PR because it's not a trivial fix and we're not focused on supporting SSR right now.

eunicorn (approved):
thanks 🎉

ui/app/src/components/DashboardList.tsx
Nexucis: The combination `metadata.project` + `metadata.name` gives you a unique key

Nexucis: `metadata.name` is unique in a given project

PR #1070, opened 2023-04-04
[BUGFIX] Dashboard validation: fix issue with variable parsing when the query embeds a regexp

AntoineThebaud:
When migrating/validating a dashboard with such variable query expr:
`thanos_build_info{stack=~"$iaas_loc.*"}`
we would encounter such validation issue (notice the trailing dot):
`bad request: variable "iaas_loc." is used in the variable "stack" but not defined`

So the easiest & pragmatic solution for now is simply to not authorize characters like `.` in variables name that could conflict with regexp syntax, which was basically the case before [last change](https://github.com/perses/perses/commit/ba17e4d2e073f766b41d7d9ae7819c36bf424d0f#diff-5fee94bdc9b5e8bf134768e6f062f3379a78da68cefa3f3ecf9690c72b3ed9efL24).

⚠️ Also there was a misalignment with the regexp used on the frontend side
https://github.com/perses/perses/blob/main/ui/dashboards/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx#L53, which is thus solved by this PR, but there are room for future improvements here: if ever we decide to (continue to) not have plugin-based variables syntax, it would be good to have a common source for this kind of regexp for both the frontend & backend, to avoid such misalignment to happen again.

EDIT: In the end this PR also embeds quite some refactor around unit tests for multiple cases. 

# Checklist

### DIFF
--- a/internal/api/shared/migrate/migrate_test.go
+++ b/internal/api/shared/migrate/migrate_test.go
@@ -11,63 +11,76 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build integration
+
 package migrate
 
 import (
 	"encoding/json"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/perses/perses/internal/api/config"
+	e2eframework "github.com/perses/perses/internal/api/e2e/framework"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+const testDataFolder = "testdata"
+
 func TestMigrate(t *testing.T) {
-	simpleGrafanaDashboard, _ := os.ReadFile("testdata/simple_grafana_dashboard.json")
-	simplePersesDashboard, _ := os.ReadFile("testdata/simple_perses_dashboard.json")
-	oldFormatGrafanaDashboard, _ := os.ReadFile("testdata/old_grafana_panels_grafana_dashboard.json")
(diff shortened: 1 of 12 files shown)


internal/api/shared/migrate/migrate_test.go
```diff
@@ -16,41 +16,43 @@ package migrate
 import (
```
AntoineThebaud, line 16: I first started to add a new test in the migration part, with some refactor at the same time, before figuring out that it was a validation issue and not a migration issue. So in the end I moved the test in the `validate` package but still kept the refactoring here

internal/api/shared/validate/validate_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Run(test.title, func(t *testing.T) {
+			persesDashboardRaw, readErr := os.ReadFile(filepath.Join(testDataFolder, test.dashboardFile))
+			if readErr != nil {
+				t.Fatal(readErr)
+			}
+			schemasService, schErr := schemas.New(config.Schemas{
+				// use the real schemas for these tests
+				VariablesPath: "../../../../schemas/variables",
+				PanelsPath:    "../../../../schemas/panels",
+				QueriesPath:   "../../../../schemas/queries",
```
Nexucis, line 54: you could do something like that: https://github.com/perses/perses/blob/main/internal/api/e2e/framework/server.go#L66-L69

using `GetRepositoryPath` will avoid to count the number of `../` required :)

Nexucis (approved):

PR #1071, opened 2023-04-04
[ENHANCEMENT] Option to configure `show_query` in time series tooltip

sjcobb:
Follow up to #1045, now that LineChart's tooltipConfig is available, it makes sense to have certain properties like `tooltip.show_query` to be configurable from the TimeSeriesChart panel spec. This PR just adds support in the JSON.

See CUE schema [here](https://github.com/perses/perses/blob/033d8049bdad21bb5f7901dabbf4cbf7ac737115/schemas/panels/time-series/time-series.cue#L24-L26) 

# Checklist

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -81,7 +81,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
 
   // control whether query is visible in TimeSeriesTooltip SeriesInfo component
   const tooltipConfig: TooltipConfig = {
-    showQuery: props.spec.tooltip?.show_query ?? true,
+    showQuery: props.spec.tooltip?.show_query ?? false,
     wrapLabels: true,
   };
 
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -26,6 +26,7 @@ import {
   YAxisLabel,
   ZoomEventData,
   useChartsTheme,
+  TooltipConfig,
 } from '@perses-dev/components';
 import { useSuggestedStepMs } from '../../model/time';
 import {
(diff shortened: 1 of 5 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -78,6 +79,12 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
       ? merge({}, DEFAULT_LEGEND, props.spec.legend)
       : undefined;
 
+  // control whether query is visible in TimeSeriesTooltip SeriesInfo component
+  const tooltipConfig: TooltipConfig = {
+    showQuery: props.spec.tooltip?.show_query ?? true,
```
sjcobb, line 84: I'm thinking the ideal behavior is to automatically set `show_query` to false when there is only 1 query, but set it to true when there is more than 1 (assuming series_name_format wasn't set since that overrides the series name anyways and will still work)

saminzadeh:
I'm thinking showQuery doesn't really make sense in the context of TimeSeriesChart. It operates on the [TimeSeries](https://github.com/perses/perses/blob/356408755a0e4a94f55fb3d476cc3914803d1c02/ui/core/src/model/time-series-data.ts#L31-L35) interface which has a property called `name`. This behavior might make more sense:

- By default, use `formatedName` if available
- If not, use `name` (which is set by a plugin)
- If `showLabels` is enabled, it should encode the `labels` in the response.

`query` is a concept for PromQL and not all datasources will have this. It might more sense to tell the PromQL TimeSeriesQuery to set the `name` to a query in that case

sjcobb:
Yeah good points. Thankfully, we do most of that, as far as name vs formattedName and setting `name` to the query when labels are empty now ([ex](https://github.com/perses/perses/blob/v0.26.1/ui/prometheus-plugin/src/utils/utils.ts#L132-L147))

Working on refactor the logic in SeriesInfo.tsx to be more generic and making it so showQuery isn't necessary, [this line](https://github.com/perses/perses/blob/v0.25.0/ui/components/src/TimeSeriesTooltip/SeriesInfo.tsx#L34) in particular is causing problems

PR #1074, opened 2023-04-05
[ENHANCEMENT] Virtualize legend to fix perf issues

juliepagano:
Prior to this fix, the entire legend would be rendered. This causes browser performance issues when there are a lot of items.

This fix modifies the legend to do the following:
- Virtualize the list variation of the legend.
- The `Right` positioned legend will always be virtualized because it always uses the list variation of the legend. The UX is consistent with prior behavior.
- The `Bottom` positioned legend will continue to use the compact, non-virtualized varation of the legend when it has a relative small (<500) items. When it has a larger (>= 500) items, it will instead be rendered with the list variation of the legend, so it can be virtualized. This is necessary because it would be very difficult to virtualize the compact view. The UX of skimming hundreds of items is much harder with the compact view, so this change also may assist with UX even if we are primarily doing it for performance reasons.
- Fixes broken ellipsizing of truncated labels.

# Notes for reviewers

- Please pull down the changes and test this out. The legend has a bunch of complex behavior, and I want to be extra-sure I didn't break any edge cases while trying to fix the performance.
- CCing @saminzadeh  on this because I _think_ some of his work was impacted by the perf issues, so he might be interested in the outcomes.

# Checklist

# TODOs

- [X] Fix ellipsizing.
- [x] Some cross-browser testing. I'll make sure I do this before merging to be extra-sure all the behaviors work as expected.

# Screenshots 

*The loom video was made before I fixed the ellipsizing, but should still give you a sense of the major virtualization changes.*
[Loom video to describe the dynamic behavior](https://www.loom.com/share/7c25f2d3e5c64490991e612906083c7e)

*Screenshots are updated to the latest code with the ellipsizing fixes.*

image

image

image
image

## Ellipsizing fix
<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image

</td>
		<td>
image

</td>
	</tr>
</table>

### DIFF
--- a/ui/components/src/Legend/Legend.stories.tsx
+++ b/ui/components/src/Legend/Legend.stories.tsx
@@ -49,6 +49,11 @@ const LegendWrapper = (props: LegendProps) => {
     options: { position },
   } = props;
 
+  // The legend does not look very interesting by itself in stories, especially
+  // when considering the positioning. This wrapper puts a box with a border
+  // and some additional height/width (depending on the positioning of the
+  // legend) to make it easier to see how the legend would look in context
+  // alongside other content.
   return (
     <Box
       sx={{
--- a/ui/components/src/Legend/Legend.stories.tsx
+++ b/ui/components/src/Legend/Legend.stories.tsx
@@ -14,7 +14,7 @@
 import type { Meta, StoryObj } from '@storybook/react';
 import { Legend, LegendProps } from '@perses-dev/components';
 import { action } from '@storybook/addon-actions';
-import { Box } from '@mui/material';
+import { Box, Stack, Typography } from '@mui/material';
 import { red, orange, yellow, green, blue, indigo, purple } from '@mui/material/colors';
 
 const COLOR_SHADES = ['400', '800'] as const;
@@ -28,12 +28,12 @@ const MOCK_COLORS = COLOR_SHADES.reduce((results, colorShade) => {
   return results;
 }, [] as string[]);
 
-function generateMockLegendData(count: number): LegendProps['data'] {
+function generateMockLegendData(count: number, labelPrefix = 'legend item'): LegendProps['data'] {
   const data: LegendProps['data'] = [];
   for (let i = 0; i < count; i++) {
     data.push({
       id: `${i}`,
-      label: `legend item ${i}`,
+      label: `${labelPrefix} ${i}`,
       isSelected: false,
       color: MOCK_COLORS[i % MOCK_COLORS.length] as string,
       onClick: action(`onClick legendItem ${i}`),
@@ -45,12 +45,16 @@ function generateMockLegendData(count: number): LegendProps['data'] {
 // Simple wrapper to try to help visualize that the legend is positioned absolutely
 // inside a relative ancestor.
 const LegendWrapper = (props: LegendProps) => {
+  const {
+    options: { position },
+  } = props;
+
   return (
     <Box
       sx={{
         position: 'relative',
-        width: props.width + 100,
-        height: props.height + 100,
+        width: position === 'Right' ? props.width + 100 : props.width,
+        height: position === 'Right' ? props.height : props.height + 100,
         border: (theme) => `solid 1px ${theme.palette.divider}`,
       }}
     >
@@ -89,8 +93,8 @@ export const Bottom: Story = {
 
 export const Right: Story = {
   args: {
-    width: 100,
-    height: 200,
+    width: 200,
+    height: 300,
     options: {
       position: 'Right',
     },
@@ -98,24 +102,124 @@ export const Right: Story = {
 };
 
 /**
- * The legend currently is not virtualized, so it can have performance issues
- * with larger amounts of data.
+ * When the legend is positioned on the right, items with long labels will be
+ * displayed in full when there are a small number of items.
+ *
+ * When there are a larger number of items, longer labels will be truncated to
+ * fit within the width. On hover, the full label will be displayed.
+ */
+export const RightWithLongLabels: Story = {
+  args: {
+    width: 200,
+    height: 300,
+  },
+  argTypes: {
+    data: {
+      table: {
+        disable: true,
+      },
+    },
+    options: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+  render: (args) => {
+    const labelPrefix = 'long_legend_label{env="demo", namespace="prometheus"}';
+
+    return (
+      <Stack spacing={3}>
+        <div>
+          <Typography variant="h3" gutterBottom>
+            Small number of items
+          </Typography>
+          <LegendWrapper {...args} options={{ position: 'Right' }} data={generateMockLegendData(4, labelPrefix)} />
+        </div>
+        <div>
+          <Typography variant="h3" gutterBottom>
+            Large number of items
+          </Typography>
+          <LegendWrapper {...args} options={{ position: 'Right' }} data={generateMockLegendData(1000, labelPrefix)} />
+        </div>
+      </Stack>
+    );
+
+    return <LegendWrapper {...args} data={generateMockLegendData(3)} />;
+  },
+};
+
+/**
+ * The legend uses virtualization to avoid performance issues when there are a
+ * large number of items to display.
+ *
+ * When the legend is positioned on the right, it is always rendered in a
+ * virtualized list with a single item per row.
+ *
+ * When the legend is positioned on the bottom with a small number of items,
+ * it is rendered in a compact, non-virtualized inline list with a variable number
+ * of items per row. When the number of items is large enough to cause performance,
+ * issues, it is rendered in a virtualized list with a single item per row.
  */
 export const Scalability: StoryObj<LegendProps & { legendItemsCount: number }> = {
   argTypes: {
     data: {
-      control: false,
+      table: {
+        disable: true,
+      },
+    },
+    width: {
+      table: {
+        disable: true,
+      },
+    },
+    height: {
+      table: {
+        disable: true,
+      },
+    },
+    options: {
+      table: {
+        disable: true,
+      },
     },
   },
   args: {
     // Custom arg for just this story to easily control how many items are rendered
     // to test performance.
-    legendItemsCount: 100,
+    legendItemsCount: 1000,
   },
   parameters: {
     happo: false,
   },
   render: (args) => {
-    return <LegendWrapper {...args} data={generateMockLegendData(args.legendItemsCount)} />;
+    return (
+      <Stack spacing={3}>
+        <div>
+          <Typography variant="h3" gutterBottom>
+            Position: right
+          </Typography>
+          <LegendWrapper
+            {...args}
+            width={400}
+            height={200}
+            options={{ position: 'Right' }}
+            data={generateMockLegendData(args.legendItemsCount)}
+          />
+        </div>
+        <div>
+          <Typography variant="h3" gutterBottom>
+            Position: bottom
+          </Typography>
+          <LegendWrapper
+            {...args}
+            width={500}
+            height={100}
+            options={{ position: 'Bottom' }}
+            data={generateMockLegendData(args.legendItemsCount)}
+          />
+        </div>
+      </Stack>
+    );
   },
 };
--- a/ui/components/src/Legend/Legend.stories.tsx
+++ b/ui/components/src/Legend/Legend.stories.tsx
@@ -53,17 +53,17 @@ const LegendWrapper = (props: LegendProps) => {
     <Box
       sx={{
         border: (theme) => `solid 1px ${theme.palette.divider}`,
+        position: 'relative',
+        width: position === 'Right' ? props.width + 100 : props.width,
+        height: position === 'Right' ? props.height : props.height + 100,
+
+        // This is a rare case where content-box is helpful because we want to
+        // have the border be additive, so we don't have to do any special
+        // math with the height/width or end up with an off-by-2px issue.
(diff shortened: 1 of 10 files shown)


ui/components/src/Legend/ListLegend.tsx
```diff
@@ -11,28 +11,110 @@
...
+import { useTheme } from '@mui/material';
+import { VariableSizeList, ListChildComponentProps } from 'react-window';
+import AutoSizer from 'react-virtualized-auto-sizer';
+import { useRef, useEffect, forwardRef } from 'react';
 import { LegendItem } from '../model';
 import { ListLegendItem } from './ListLegendItem';
 
 interface ListLegendProps {
   items: LegendItem[];
 }
```
juliepagano, line 24: This code is a little gnarlier than I would like because we need to:
- Virtualize
- Simulate padding -- if you use real padding/margin, the scrollbar ends up offset too, which looks yucky.
- Handle dynamic heights because the items change on hover

The implementation is based on guidance from react-window, so I think it's a reasonable approach, but not super-easy to follow, so I added a lot of comments to try to help understand.

ui/components/src/Legend/ListLegend.tsx
```diff
@@ -11,28 +11,110 @@
...
+  // adjust to size changes.
+  function setRowHeight(index: number, size: number) {
+    // Tell the virtualized list that items changed size and need to be
+    // re-evaluated.
+    listRef.current?.resetAfterIndex(0);
+    rowHeights.current = { ...rowHeights.current, [index]: size };
+  }
+
+  // Renderer for virtualized rows in `VariableSizeList`.
+  function ListLegendRow({ index, style }: ListChildComponentProps) {
```
juliepagano, line 71: I don't normally like nesting component definitions but it's necessary for this virtualized list use case because we need access to some shared values from the parent (`items`, `LIST_PADDING`) that we can't easily pass around because of how the `react-window` apis work.

ui/components/src/Legend/ListLegendItem.tsx
```diff
@@ -11,20 +11,24 @@
...
 
-import React from 'react';
+import React, { forwardRef } from 'react';
 import { Box, ListItemText, ListItem, ListItemProps } from '@mui/material';
 import { LegendItem } from '../model';
 import { combineSx } from '../utils';
 import { LegendColorBadge } from './LegendColorBadge';
 
-interface ListLegendItemProps extends ListItemProps {
+interface ListLegendItemProps extends ListItemProps<'div'> {
```
juliepagano, line 20: `li` elements MUST have a `ul` or `ol` as a parent, which isn't possible with the virtualization, so changing these to `div`.

ui/components/src/Legend/ListLegendItem.tsx
```diff
@@ -11,20 +11,24 @@
...
 
-export const ListLegendItem = React.memo(function ListLegendItem({ item, sx, ...others }: ListLegendItemProps) {
+const ListLegendItemBase = forwardRef<HTMLDivElement, ListLegendItemProps>(function ListLegendItem(
+  { item, sx, ...others },
+  ref
+) {
   return (
     <ListItem
       {...others}
+      component="div"
```
juliepagano, line 31: Same comment as above about the li vs. div

ui/components/src/Legend/ListLegendItem.tsx
```diff
@@ -44,3 +49,5 @@ export const ListLegendItem = React.memo(function ListLegendItem({ item, sx, ...
     </ListItem>
   );
 });
+
+export const ListLegendItem = React.memo(ListLegendItemBase);
```
juliepagano, line 53: I do the `memo` separately, so we can have a function name associated with the output of the `forwardRef`. Otherwise, eslint complains about the component not having a `displayName`.

ui/components/src/model/legend.ts
```diff
@@ -26,7 +26,7 @@ export interface LegendItem {
   label: string;
   isSelected: boolean;
   color: string;
-  onClick: MouseEventHandler<HTMLLIElement>;
+  onClick: MouseEventHandler<HTMLElement>;
```
juliepagano, line 29: Making this more generic, so it can handle the li and div depending on the virtualization or not.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.test.tsx
```diff
@@ -23,9 +23,19 @@ import {
   mockPluginRegistry,
   MockPlugin,
 } from '@perses-dev/plugin-system';
+import { Props as AutoSizerProps } from 'react-virtualized-auto-sizer';
 import { MOCK_TIME_SERIES_QUERY_RESULT, MOCK_TIME_SERIES_DATA } from '../../test';
 import { TimeSeriesChartPanel, TimeSeriesChartProps } from './TimeSeriesChartPanel';
 
+// Mock AutoSizer because it does not play well with jsdom.
```
juliepagano, line 30: It might make sense to switch these to e2e tests at some point, but for now, just mocking some basics because we're not trying to test that AutoSizer works here.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.test.tsx
```diff
@@ -129,51 +139,52 @@ describe('TimeSeriesChartPanel', () => {
     renderPanel();
```
juliepagano, line 140: Throughout this test, these legend list item lookups need to be done at the time they're being checked (instead of once and reused) because they are virtualized and the underlying element can change.

juliepagano:
Converting this back to a draft because I realized one of my adjustments quasi-broke the ellipsizing of the text. I have a sense of how to refactor a little to fix it and will un-draft when that's done.

ui/components/src/Legend/Legend.stories.tsx
```diff
@@ -45,16 +45,25 @@ function generateMockLegendData(count: number): LegendProps['data'] {
...
+    options: { position },
+  } = props;
+
   return (
     <Box
       sx={{
-        position: 'relative',
-        width: props.width + 100,
-        height: props.height + 100,
         border: (theme) => `solid 1px ${theme.palette.divider}`,
```
juliepagano, line 55: Separated out the box with the border and the box with the height/width to avoid some issues with being off by a few pixels  for the legend because of a combo of the border and `box-sizing: border-box`.

ui/components/src/Legend/ListLegend.tsx
```diff
@@ -11,40 +11,155 @@
...
+          truncateLabel={truncateLabels}
+          onLayoutChange={handleRowLayoutChange}
           sx={{
-            width: 190,
-            textOverflow: 'ellipsis',
+            // Having an explicit width is important for the ellipsizing to
+            // work correctly. Subtract padding to simulate padding.
+            width: width - LIST_PADDING,
             wordBreak: 'break-word',
-            overflow: truncateLabels ? 'hidden' : 'visible',
```
juliepagano, line 38: I realized the truncation was broken while I was in here. With this html layout, it needs to happen farther down in the code, which was a little uglier to do, but works as intended.

ui/components/src/Legend/ListLegendItem.tsx
```diff
@@ -11,20 +11,59 @@
...
+   *
+   * When `false` or unset, will show the full label.
+   */
+  truncateLabel?: boolean;
+
+  /**
+   * Called when the layout of the legend item changes as a result of the hover
+   * behavior when `truncateLabel` is `true`.
+   */
+  onLayoutChange?: () => void;
```
juliepagano, line 35: Need these props to get the ellipsized truncation + the virtualization working properly together.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -260,7 +260,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   // TODO: account for number of time series returned when adjusting legend spacing
   let legendHeight = LEGEND_HEIGHT_SM;
   if (legend && legend.position === 'Right') {
-    legendHeight = adjustedContentDimensions.height;
+    legendHeight = contentDimensions?.height || adjustedContentDimensions.height;
```
juliepagano, line 263: Adjusting this to make the legend fill the full height of the panel because the padding is handled internally. Falling back to `adjustedContentDimensions` because TS thinks `contentDimensions` can be undefined.

ui/components/src/Legend/Legend.tsx
```diff
@@ -23,25 +23,35 @@ export interface LegendProps {
...
+const NEED_VIRTUALIZATION_LIMIT = 500;
+
 export function Legend({ width, height, options, data }: LegendProps) {
   if (options.position === 'Right') {
     return (
       <Box
         sx={{
           width: width,
-          height: '100%',
+          height: height,
```
juliepagano, line 39: Using the specified height is probably more "correct" from an expectation perspective AND, more importantly, enables us to use the height to do calculations for the virtualization. Otherwise, we'd need something like autosizer, which adds even more complexity to this already complex code.

ui/components/src/Legend/Legend.tsx
```diff
@@ -23,25 +23,35 @@ export interface LegendProps {
...
           position: 'absolute',
           top: 0,
           right: 0,
-          overflowX: 'hidden',
-          overflowY: 'scroll',
```
juliepagano, line 37: Scrolling is handled by the virtualized list.

juliepagano:
> Converting this back to a draft because I realized one of my adjustments quasi-broke the ellipsizing of the text. I have a sense of how to refactor a little to fix it and will un-draft when that's done.

Actually, the ellipsis was broken _before_ did this, but it looked differently broken because of some slightly different widths. In any case, I fixed it while I was already in here, so it would play nicely with the virtualization.  

This PR should now be ready to review. Pretty please pull this one down and do some stress testing because there's a lot of complex patterns in here, and I want to make sure I don't break a behavior I wasn't aware of.

ui/components/src/Legend/ListLegend.tsx
```diff
@@ -11,40 +11,156 @@
...
 
 interface ListLegendProps {
   items: LegendItem[];
+  height: number;
+  width: number;
 }
 
+// Default height used to start while virtualizing the list before a true
+// height is determined.
+const DEFAULT_ROW_HEIGHT = 26;
```
eunicorn, line 28: should we make this an optional prop?

eunicorn: is this used to determine the height of the skeleton row when it's loading?

juliepagano: > is this used to determine the height of the skeleton row when it's loading if there is a loading state?

The virtualized list needs to know heights to lay out the list, but we can't be sure of the height of items until we try to render them. This value is used to initially lay out the items in the virtualized list on the first render. After that happens,  we're able to look at the rendered height of items and adjust appropriately (the bits where we put refs on the row and then set the height).

> should we make this an optional prop?

I was debating if that was useful, but we don't expose `ListLegend` directly. Let me take a look at if it could be a helpful prop on `Legend` that gets passed down.

ui/components/src/Legend/ListLegend.tsx
```diff
@@ -11,40 +11,156 @@
...
+  // Renderer for virtualized rows in `VariableSizeList`.
+  function ListLegendRow({ index, style }: ListChildComponentProps) {
+    // Storing a ref to the row's `ListLegendItem`, so we can get the "real"
+    // height and adjust the height of the row based on it, enabling the dynamic
+    // heights on hover.
+    const rowRef = useRef<HTMLDivElement | null>(null);
+
+    // Adjust row heights when the row being rendered changes.
+    useEffect(() => {
+      if (rowRef.current) {
```
eunicorn, line 80: nit:  can call `handleRowLayoutChange` here

ui/components/src/Legend/Legend.stories.tsx
```diff
@@ -45,13 +45,22 @@ function generateMockLegendData(count: number): LegendProps['data'] {
...
+        position: 'relative',
+        width: position === 'Right' ? props.width + 100 : props.width,
```
eunicorn, line 57: curious - why do we need to add 100 to the width here?

eunicorn (approved):

ui/components/src/Legend/ListLegend.tsx
juliepagano: Added an `initialRowHeight` prop on `ListLegend` and exposed that on `Legend` as `listProps` (trying to follow that pattern MUI uses a bunch about passing props to a nested part of a top-level component).

sjcobb (approved):
🚀

ui/components/src/Legend/Legend.stories.tsx
juliepagano: Good callout. I should put a comment above here explaining what I'm doing. The Legend doesn't look very interesting by itself, especially when you're trying to look at how the right/bottom positioning works. I've put a little wrapper div with a border around it that's bigger than the legend to try to simulate the legend sitting next to a chart and make it easier to understand what the positioning looks like. I do this by adding some spacing to the width or height depending on the positioning.

I'll add a short comment summarizing this above the code for clarification.

juliepagano: Done!

PR #1075, opened 2023-04-05
Remove minimumFractionDigits

cndonovan:
When a user specifies `decimal_places`, we only want to show the decimal places if the measurement has that degree of precision. 

In other words, if the user specifies `decimal_places` as 2, and if the number is 7000, we don't want to show 7000.00. But if the number is 7000.12, we do want to show 7000.12. 

Right now, if a user picks `decimal_places` as 2, and if the number is 7000, we will show 7000.00. To fix this, we remove `minimumFractionDigits`.

`minimumFractionDigits: X` forces a number to have at least X decimal places. When we remove that option, we are only relying on `maximumFractionDigits`. This produces the behavior that we want.


ui/components/src/model/units/bytes.ts
```diff
@@ -43,19 +43,18 @@ export function formatBytes(bytes: number, unitOptions: BytesUnitOptions) {
   // default to full 'Bytes' formatting
   const options = unitOptions.abbreviate === undefined ? { ...unitOptions, abbreviate: false } : unitOptions;
 
-  let decimals = options.decimal_places ?? DEFAULT_DECIMAL_PLACES;
+  let decimalPlaces = options.decimal_places ?? DEFAULT_DECIMAL_PLACES;
```
cndonovan, line 46: @sjcobb I renamed this, let me know if you dislike

saminzadeh (approved):
This looks good. A few observations unrelated directly to this PR. 

1) The `unit` property seems strange. `kind: Decimal` doesn't make sense in the context of a unit. It should just be `null` meaning that there are no units. Also, `decimal_places` is irrespective of the units so it probably shouldn't be scoped there.
2) @juliepagano mentioned this but its probably time to bring in a [number formatting library](https://numbrojs.com/format.html) so we don't have to build this. It handles commas and unit formatting.

 I think we'll have to take another pass at some point.

sjcobb (approved):
🙌

juliepagano (approved):
LGTM! 🚀

PR #1072, opened 2023-04-04
[IGNORE] Update dashboard migration logic 

eunicorn:
Due to the breaking changes in https://github.com/perses/perses/pull/1032, we need to update the dashboard migration and validation logic.

PR #1077, opened 2023-04-06
[IGNORE] Query schemas: update load logic + validation logic + unit tests according to query datamodel breaking change

AntoineThebaud:
This query datamodel breaking change actually involved to rework quite some logic on the backend side regarding schemas load and schemas validation. Actually it simplified a lot of things, that we had to originally put in place to cope with the former "open" datamodel for queries ('could be in a single object, an array or even a map, located anywhere at any depth within the panel plugin spec). Now that the backend code is "aware" of where the queries are located, the logic for those could be realigned on what we have for all the other types of plugins (i.e panels, variables & datasources) for which this was already the case.

_NB: cue doc to be updated accordingly in another PR - dont want to create merge conflict because of this for now._

# Checklist


AntoineThebaud:
@eunicorn I opened this PR in top of your branch
@Nexucis I was assuming eunice's [BREAKINGCHANGE] tag on her "main" branch for this whole change is enough, hence why I put IGNORE here, but let me know if this is wrong

Nexucis:
> @Nexucis I was assuming eunice's [BREAKINGCHANGE] tag on her "main" branch for this change is enough, hence why I put IGNORE here, but let me know if this is wrong

That's a good questions, I'm even wondering if we should squash her PR or merge it actually to preserve the different commit

Nexucis (approved):

internal/api/shared/migrate/testdata/old_grafana_panels_perses_dashboard.json
```diff
@@ -47,7 +47,8 @@
             "spec": {
               "text": "<p>This dashboard provides information on OTF Front-End queues.<br/>Minimum OTF FE version required: 18.0.0.68 / 19.0.0.29<br/>\nMore information on FE metrics used here: <a href=\"https://rndwww.nce.amadeus.net/confluence/display/OTFTeam/OTF+Monitoring+-+FE+metrics\">Confluence link</a>\n</p>"
             }
-          }
+          },
+          "queries": []
```
Nexucis, line 51: mmm this should not appear if it is empty. But I guess it's coming from the other branch. I will look there

AntoineThebaud: Hmm if you really want it to not generate an empty array when there's no query, there's most probably a new check to add in the cue logic here https://github.com/perses/perses/pull/1072/files#diff-d4979002465bf1e3768d854a8ba3c0caa4802753bf17e356b3c3a6015b3460ea

eunicorn:
> Actually it simplified a lot of things, that we had to originally put in place to cope with the former "open" datamodel for queries ('could be in a single object, an array or even a map, located anywhere at any depth within the panel plugin spec). 

love to hear that this change not only helps UI but also simplifies BE logic 🎉

internal/api/shared/migrate/testdata/old_grafana_panels_perses_dashboard.json
eunicorn: I updated my branch to add `omitempty` for `queries`. Can you please pull the new changes?

pkg/model/api/v1/dashboard_test.go
```diff
@@ -285,7 +286,8 @@ func TestMarshalDashboard(t *testing.T) {
                 "up"
               ]
             }
-          }
+          },
+          "queries": null
```
eunicorn, line 290: if you pull the new changes where i add `omitempty`, it should undo this change, right?

eunicorn (approved):
looks good to me! `eunicorn/update-cue-for-panel-spec` is the correct branch!

pkg/model/api/v1/dashboard_test.go
Nexucis: yeah totally, this should disappear !

AntoineThebaud: fixed :)

PR #1079, opened 2023-04-07
[WIP] Try string-to-color npm package for palette generation

sjcobb:
Not crazy about some of the dependencies it adds in package-lock, but here is one potential option to change our default generative palette to: 

https://user-images.githubusercontent.com/9369625/230629095-91edfc45-5121-4947-bfa1-b46afb9e725f.mov

Screenshot 2023-04-07 at 10 41 42 AM

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -11643,6 +11643,11 @@
       "integrity": "sha512-3tlv/dIP7FWvj3BsbHrGLJ6l/oKh1O3TcgBqMn+yyCagOxc23fyzDS6HypQbgxWbkpDnf52p1LuR4eWDQ/K9WQ==",
       "dev": true
     },
+    "node_modules/colornames": {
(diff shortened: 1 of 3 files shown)


sjcobb:
Colors still don't look great in certain cases in this branch and I don't like that the [string-to-color](https://github.com/Gustu/string-to-color) dependency seems abandoned and adds a bunch of dependencies to our package-lock.json. Moving on to a different approach using approach described here: https://stackoverflow.com/questions/3426404/create-a-hexadecimal-colour-based-on-a-string-with-javascript/31037383#31037383

Contrast palette will reference following versions:
- https://colorbrewer2.org/#type=qualitative&scheme=Paired&n=12
- https://stackoverflow.com/questions/309149/generate-distinctly-different-rgb-colors-in-graphs/12224359#12224359

PR #1080, opened 2023-04-07
[WIP] Categorical contrast color palette that is consistent across panels

sjcobb:
Follow up to #1079, this branch has a demo of another approach: 
 * Color conversion using predefined palette for contrast
 * Overall approach adapted from: https://stackoverflow.com/a/31037383/17575201
 * Contrast colors started from: https://stackoverflow.com/a/12224359/17575201

_Downside here is the colors have a lot of contrast, but they start to repeat after 38 timeseries_

https://user-images.githubusercontent.com/9369625/230654974-641f980e-7233-4e25-84c0-60bc6574d75b.mov

image


sjcobb:
This could be helpful for our `Categorical` color palette later, but it's not going to work when we have hundreds or thousands of series 😞

PR #1081, opened 2023-04-07
[BUGFIX] Stat chart value to use threshold color

eunicorn:
Stat chart value wasn't using threshold color, so if sparkline was disabled, you could not tell if the chart met the threshold or not.

**BEFORE**
image

**AFTER**
image

Also updated the look of Thresholds Editor:

image

### DIFF
--- a/ui/components/src/ThresholdsEditor/ThresholdInput.tsx
+++ b/ui/components/src/ThresholdsEditor/ThresholdInput.tsx
@@ -12,8 +12,9 @@
 // limitations under the License.
 
 import { RefObject, useState } from 'react';
-import { Stack, FormLabel, TextField, IconButton } from '@mui/material';
+import { Stack, FormLabel, TextField, IconButton, Box } from '@mui/material';
 import DeleteIcon from 'mdi-material-ui/DeleteOutline';
+import { ThresholdOptions } from '@perses-dev/core';
 import { ThresholdColorPicker } from './ThresholdColorPicker';
 
 export interface ThresholdInputProps {
@@ -25,6 +26,7 @@ export interface ThresholdInputProps {
   onBlur: () => void;
   onDelete: () => void;
   inputRef?: RefObject<HTMLInputElement | null>;
+  mode?: ThresholdOptions['mode'];
 }
 
 export function ThresholdInput({
@@ -32,6 +34,7 @@ export function ThresholdInput({
   label,
   color,
   value,
+  mode,
   onChange,
   onColorChange,
   onBlur,
@@ -56,6 +59,9 @@ export function ThresholdInput({
             onBlur();
             setKey(key + 1);
           }
+        }}
+        InputProps={{
+          endAdornment: mode === 'Percent' ? <Box paddingX={1}>%</Box> : undefined,
         }}
       />
       <IconButton aria-label={`delete threshold ${label}`} size="small" onClick={onDelete}>
--- a/ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
+++ b/ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
@@ -17,9 +17,8 @@ import { IconButton, ToggleButton, ToggleButtonGroup, Typography } from '@mui/ma
 import PlusIcon from 'mdi-material-ui/Plus';
 import { Stack } from '@mui/system';
 import { ThresholdOptions } from '@perses-dev/core';
-import { InfoTooltip } from '../InfoTooltip';
 import { useChartsTheme } from '../context/ChartsThemeProvider';
-import { OptionsEditorGroup } from '../OptionsEditorLayout';
+import { OptionsEditorControl, OptionsEditorGroup } from '../OptionsEditorLayout';
 import { ThresholdColorPicker } from './ThresholdColorPicker';
 import { ThresholdInput } from './ThresholdInput';
 
@@ -85,6 +84,10 @@ export function ThresholdsEditor({ thresholds, onChange, hideDefault, disablePer
           draft.default_color = color;
         })
       );
+    } else {
+      onChange({
+        default_color: color,
+      });
     }
   };
 
@@ -116,7 +119,7 @@ export function ThresholdsEditor({ thresholds, onChange, hideDefault, disablePer
 
   const addThresholdInput = (): void => {
     focusRef.current = true;
-    if (thresholds === undefined) {
+    if (thresholds?.steps === undefined) {
       onChange({
         steps: [{ value: DEFAULT_STEP }],
       });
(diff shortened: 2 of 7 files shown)


ui/panels-plugin/src/plugins/stat-chart/utils/data-transform.ts
```diff
@@ -38,6 +38,20 @@ export function convertSparkline(
...
+export function convertSparkline(
+  chartsTheme: PersesChartsTheme,
+  sparkline?: SparklineOptions,
+  thresholds?: ThresholdOptions,
+  value?: number
+): LineSeriesOption | undefined {
+  if (sparkline === undefined) return;
+
+  // TO DO: add option for color scheme? Should default color derive from threshold.defaultColor or sparkline.color?
+  const defaultColor = chartsTheme.thresholds.defaultColor ?? chartsTheme.sparkline.color;
```
sjcobb, line 53: Not sure I 100% understand the question in the TODO, I guess it depends if thresholds in panel spec is defined or not right? If it’s undefined, always go with chartsTheme.sparkline.color for default?

sjcobb (approved):
🎉

ui/panels-plugin/src/plugins/stat-chart/utils/data-transform.ts
juliepagano: I noticed this case when testing and found it kind of confusing. Guessing that's what Eunice is describing here. 

image

eunicorn: In grafana there is an option for color scheme. I think if you choose "from thresholds", default color is using chartsTheme.threshold.defaultColor, or panel spec `thresholds.default_color`  even when you don't have threshold defined.

image

juliepagano: This might be a UX that's worth talking through with one of the designers. It is a little weird to set the default line color by setting the default threshold, but it's also really weird to have the line and the text be different colors in this case.

eunicorn: @juliepagano ooh how did you get that? are you using `sparkline.color`? 

there are too many different ways you can set sparkline colors lol 🤯

juliepagano: I also think I maybe found a bug while I was in here. It looks like editing the color of the default threshold doesn't work if you haven't added other threshold values. I wonder if we shouldn't show/use the default value at all unless a threshold is set?

juliepagano: I was also able to hit a case where I modified the default value, but the text was still using the "default" green default threshold value 

image

juliepagano: > @juliepagano ooh how did you get that? are you using `sparkline.color`?
> 
> there are too many different ways you can set sparkline colors lol 🤯

That was the default state of one of the perses app panels when I ran your PR. It's one of the stat charts in the first row if you go to http://localhost:3000/projects/perses/dashboards/Demo?start=5m&var-instance=demo.do.prometheus.io%3A9100&var-interval=1m

eunicorn: oh my LOL oops it definitely wasn't like that earlier, i must have broke something 😆

eunicorn:
@juliepagano @sjcobb ready for another pass!

Discussed with @foxbat07 and decided that thresholds colors take priority over other colors, so even if json spec defines `sparkline.color`, it will not override `threshold.default_color`.

We also updated the look of thresholds mode.
image

juliepagano:
> @juliepagano @sjcobb ready for another pass!
> 
> Discussed with @foxbat07 and decided that thresholds colors take priority over other colors, so even if json spec defines `sparkline.color`, it will not override `threshold.default_color`.
> 
> We also updated the look of thresholds mode. image

I find the UX of a new stat chart without any configured thresholds being green a little weird/confusing. Not a blocker for your PR moving forward, but something for @foxbat07 to maybe think about if there are ways we can improve the UX a little.

juliepagano (approved):
LGTM! 🎉

ui/components/src/ThresholdsEditor/ThresholdInput.tsx
```diff
@@ -57,6 +60,9 @@ export function ThresholdInput({
             setKey(key + 1);
           }
         }}
+        InputProps={{
+          endAdornment: mode === 'Percent' ? <Box paddingX={1}>%</Box> : undefined,
+        }}
```
cndonovan, line 65: Nice!

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -85,6 +84,10 @@ export function ThresholdsEditor({ thresholds, onChange, hideDefault, disablePer
           draft.default_color = color;
         })
       );
+    } else {
+      onChange({
+        default_color: color,
+      });
```
cndonovan, line 90: Would it help to add this to `handleThresholdColorChange` too?

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
```diff
@@ -116,7 +119,7 @@ export function ThresholdsEditor({ thresholds, onChange, hideDefault, disablePer
 
   const addThresholdInput = (): void => {
     focusRef.current = true;
-    if (thresholds === undefined) {
+    if (thresholds?.steps === undefined) {
       onChange({
         steps: [{ value: DEFAULT_STEP }],
       });
```
cndonovan, line 125: Could this overwrite other values on `thresholds`? Like if `thresholds = { default_color: blah-blah }`, would this overwrite / lose the `default_color` field?

But I'm not sure if that would happen

cndonovan (approved):
✨

cndonovan:
Some UX things after playing around a bit:

Could the Mode + info ico + [ Percent | Absolute ] all be aligned? The info icon looks unaligned: 
Screenshot 2023-04-10 at 12 16 06 PM

Not for this PR, but could be nice to add a tooltip that says "Add threshold"
Screenshot 2023-04-10 at 12 16 18 PM

The default values for thresholds are a little confusing to me:
Screenshot 2023-04-10 at 12 16 32 PM

Also, some kind of bug happening here:

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: I don't think we need to because `handleThresholdColorChange` should only be called when `thresholds.steps` is already defined

ui/components/src/ThresholdsEditor/ThresholdsEditor.tsx
eunicorn: good catch!!

eunicorn:
> The default values for thresholds are a little confusing to me:

Hmm yeah i don't know if there is a good way to automatically determine a default value that will makes sense. We can add some config for users to set their own default value?

> Also, some kind of bug happening here:

it automatically sorts the thresholds in ascending order

PR #1082, opened 2023-04-07
[IGNORE] Refactor test utils for reuse in storybook

juliepagano:
This commit:
- Creates an `internal-utilities` package where we can store utilities we need for development purposes across multiple packages (e.g. e2e tests and storybook).
- Moves the time series response generation utils from `e2e` to `internal-utilities`, so we can also use them in storybook.
- Adds [mock service worker (MSW)](https://mswjs.io/) and the associated storybook addon, so we can easily mock api requests in stories.
- Created an example story that mocks queries to validate that the above all works end-to-end.

This should make it much easier for us to build up stories that we want to have static data, so that we can safely take visual testing snapshots of them without worrying about flakiness.

# Checklist

## UI Changes

# Screenshots

You can find the new example at http://localhost:6006/?path=/story/dashboards-components-dashboard--example-with-time-series-panels when running storybook.

image

image


scripts/ui_release.sh
```diff
@@ -13,7 +13,7 @@ publish_workspaces=$(for workspace in ${workspaces}; do
   # packages are private so we shouldn't try to publish them.
   # TODO: see if we can do something smarter here by checking for "private" set
   # to true in a given package's package.json.
-  if [[ "${workspace}" != "app" ]] && [[ "${workspace}" != "e2e" ]] && [[ "${workspace}" != "storybook" ]]; then
+  if [[ "${workspace}" != "app" ]] && [[ "${workspace}" != "e2e" ]] && [[ "${workspace}" != "storybook" ]]; && [[ "${workspace}" != "internal-utilities" ]]; then
```
juliepagano, line 16: I feel bad every time I add another check in here, but I don't know a quick way to do this in a more clever way in bash. 😢

ui/internal-utilities/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@perses-dev/internal-utilities",
```
juliepagano, line 2: I don't have strong feels about this package name. Let me know if you'd like something else.

This is set up as an [internal package](https://turbo.build/repo/docs/handbook/sharing-code/internal-packages)

ui/internal-utilities/src/utils/time-series/mockTimeSeriesResponse.ts
```diff
@@ -28,6 +28,16 @@ type MockStableValueConfig = {
   count?: number;
 };
```
juliepagano, line 29: Mostly this file was moved around. The changes (which look a little weird in the diff) are refactoring a small amount of duplication across two of the functions.

ui/storybook/src/config/preview.ts
```diff
@@ -101,4 +102,6 @@ setThemeSwitcher(async (theme, channel) => {
   });
 });
 
-export const decorators = [WithTimeZone, WithBackground, WithThemes];
+initializeMSW({ onUnhandledRequest: 'bypass' });
```
juliepagano, line 105: This `bypass` option disables a default behavior where msw does an annoying console.warn for every query it sees that isn't mocked.

ui/storybook/src/public/mockServiceWorker.js
```diff
@@ -0,0 +1,303 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker (1.2.1).
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
```
juliepagano, line 8: This file is auto-generated by mws.

ui/storybook/src/utils/requests/queryRangeRequests.ts
```diff
@@ -0,0 +1,44 @@
...
+
+type MockQueryRangeConfig = {
+  queries: MockQueryRangeQueryConfig[];
+};
+
+/**
+ * Mock responses from '/query_range' by the query parameter in the
+ * request. Useful for stabilizing charts when taking screenshots.
+ */
+export function mockQueryRangeRequests({ queries }: MockQueryRangeConfig): RequestHandler[] {
```
juliepagano, line 20: This is very similar to a util we have in the e2e tests but written to work with MSW (what we're using in storybook to mock api requests) instead of the playwright routing api.

ui/storybook/package.json
```diff
@@ -35,9 +35,15 @@
     "babel-plugin-react-require": "^3.1.3",
     "happo-plugin-storybook": "^3.2.0",
     "happo.io": "^8.3.1",
+    "msw": "^1.2.1",
+    "msw-storybook-addon": "^1.8.0",
```
juliepagano, line 39: This is not an internal storybook addon, but it is the most popular addon for mocking data fetching and seems to be kept up-to-date.

ui/internal-utilities/src/utils/time-series/mockTimeSeriesResponse.ts
```diff
@@ -11,10 +11,12 @@
...
 
-import { PrometheusDatasource } from '@perses-dev/prometheus-plugin';
+import { MatrixData } from '@perses-dev/prometheus-plugin';
 
-type PrometheusClient = ReturnType<(typeof PrometheusDatasource)['createClient']>;
-type RangeQuery = Awaited<ReturnType<PrometheusClient['rangeQuery']>>;
+type MockResponse = {
+  status: string;
+  data: Record<string, unknown>;
+};
```
juliepagano, line 19: I simplified the typing here because I was having a really hard time getting all the types to play nicely with msw and we don't need quite this level of typing for the mocks. We still get the basics from using `MatrixData` for the results and it removes the need to infer those extra types that aren't exported.

ui/dashboards/src/components/Dashboard/Dashboard.stories.tsx
```diff
@@ -213,3 +218,241 @@ export const CustomEmptyState: Story = {
     },
   },
 };
+
```
juliepagano, line 221: This data matches what we use in http://localhost:3000/projects/testing/dashboards/TimeSeriesChartPanel?start=6h for e2e tests as an easy to set up starter case for comparison.

scripts/ui_release.sh
sjcobb: Yeahh looks fine for now, not like this gets called a ton or anything

ui/internal-utilities/package.json
sjcobb: My gut reaction was `@perses-dev/internal-utils` might be nice since it is shorter, but I don’t feel that strongly about it either 🤷‍♂️

sjcobb (approved):

ui/internal-utilities/package.json
juliepagano: That's an easy change. Will adjust that before merging.

PR #1083, opened 2023-04-07
[BUGFIX] List variable input label should match fieldset

juliepagano:
Prior to this fix, the list variable used the variable's display name for the accessible input label and the variable's name for the `label` field on the select, which is turned into a `fieldset` that is used to leave space for the label in MUI. When these values have mismatched sizes, this can lead to unexpected rendering of the label.

# Checklist

## UI Changes

# Screenshots

You can reproduce this bug by giving a list-based template variable a short name and a long display name. This bug does not happen in text-based template variables.

## Before the fix

image

## After the fix

image

### DIFF
--- a/ui/dashboards/src/components/Variables/TemplateVariable.tsx
+++ b/ui/dashboards/src/components/Variables/TemplateVariable.tsx
@@ -104,7 +104,7 @@ function ListVariable({ name }: TemplateVariableProps) {
         <Select
           sx={{ minWidth: 100, maxWidth: 250 }}
           id={name}
-          label={name}
+          label={title}
           value={selectValue}
           onChange={(e) => {
             // Must be selected
--- a/ui/dashboards/src/components/Variables/VariableList.tsx
+++ b/ui/dashboards/src/components/Variables/VariableList.tsx
@@ -17,7 +17,7 @@ import PinOutline from 'mdi-material-ui/PinOutline';
 import PinOffOutline from 'mdi-material-ui/PinOffOutline';
 import { VariableDefinition } from '@perses-dev/core';
 import { useTemplateVariableDefinitions } from '../../context';
-import { TemplateVariable } from './Variable';
+import { TemplateVariable } from './TemplateVariable';
 
 const VARIABLE_INPUT_MIN_WIDTH = '120px';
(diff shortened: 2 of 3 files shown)


ui/dashboards/src/components/Variables/TemplateVariable.tsx
```diff
@@ -104,7 +104,7 @@ function ListVariable({ name }: TemplateVariableProps) {
         <Select
           sx={{ minWidth: 100, maxWidth: 250 }}
           id={name}
-          label={name}
+          label={title}
```
juliepagano, line 107: This is a really subtle UI bug that seemed like it might not really be worth setting up visual tests for. Let me know if you disagree, and I'll add some.

ui/dashboards/src/components/Variables/TemplateVariable.tsx
```diff
@@ -104,7 +104,7 @@ function ListVariable({ name }: TemplateVariableProps) {
         <Select
```
juliepagano, line 104: I renamed this file while I was in here because this is the third or fourth time I've been confused because the file name doesn't match the exported component.

saminzadeh: Ah that was my bad during implementation. 

We meant for everything to be "Variable" and not "TemplateVariable". The idea was that Variables are more first class in Perses. Meaning they are explicitly supported in certain parts of the application rather than anywhere in a templatized string (unlike Grafana).

I think they're basically interchangeable but for sake of consistency with components like `VariableList` and the Dashboard spec, maybe it's best to remove anything that prefixes "Template" and renaming it to "Variable". However, I realize that this will result in a few breaking changes, such as the `TemplateVariableProvider` and `TemplateVariable` components. (Not for this PR)

saminzadeh (approved):

ui/dashboards/src/components/Variables/TemplateVariable.tsx
juliepagano: Ah gotcha. I'll leave this for file/component consistency for now, and we can take a pass at s/TemplateVariable/Variable in a future PR.

eunicorn (approved):

PR #1084, opened 2023-04-08
[ENHANCEMENT] Format bytes with units that are powers of 1000

cndonovan:
## Changes
* Add `numbro` package
* Use `numbro` with `base: decimal` to get units that are powers of 1000
* Use `Intl.NumberFormat` with `{ style: 'unit', unit: 'byte' }` instead of `{ style: 'decimal' }`
* Add a bunch of unit tests

### DIFF
--- a/ui/components/src/model/units/bytes.ts
+++ b/ui/components/src/model/units/bytes.ts
@@ -11,6 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import numbro from 'numbro';
+
 import { DEFAULT_DECIMAL_PLACES } from './constants';
 import { UnitGroupConfig, UnitConfig } from './types';
 
@@ -27,6 +29,7 @@ export const BYTES_GROUP_CONFIG: UnitGroupConfig = {
   abbreviate: true,
 };
 export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
+  // Using units that are powers of 1000. In other words, 1KB = 1000 bytes.
   Bytes: {
     group: 'Bytes',
     label: 'Bytes',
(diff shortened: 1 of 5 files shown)


ui/components/src/model/units/bytes.ts
```diff
@@ -27,48 +29,45 @@ export const BYTES_GROUP_CONFIG: UnitGroupConfig = {
...
-    style: 'decimal',
-    maximumFractionDigits: decimalPlaces,
-    useGrouping: true,
-  };
-  const formatter = new Intl.NumberFormat('en-US', formatParams);
+  const showFullNumber = abbreviate == false || bytes < 1000;
 
-  if (options.abbreviate === false) {
-    return `${formatter.format(bytes)} bytes`;
+  if (showFullNumber) {
```
juliepagano, line 54: Out of curiosity, why use `numbro` in one case and `Intl.NumberFormat` in the other? Do we have a rule of thumb where we'd want one over the other or can we standardize on one for most cases?

juliepagano: Related: anything worth adding to the newly updated `ui-guidelines` on this if we have some guidelines? At a minimum, it might be worth adding an entry for number formatting to the [list of common libs we use](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md#notable-tooling-choices).

juliepagano (approved):
A couple small questions, but generally LGTM. 🎉

ui/components/src/model/units/bytes.ts
cndonovan: Oo good call, will do that

Yeah, there is a limitation in terms of what each can do. `numbro` handles formatting bytes as KB, MB, GB, etc. nicely. As far as I could tell, `numbro` cannot do something like "don't use MB, GB, etc., just write out the full number."

`Intl.NumberFormat`, on the other hand, can easily write out the full number. As far as I can tell, it does not have a way to format bytes as KB, MB, GB, etc. nicely. It has an option called `notation: compact`, but this will use units that correspond to millions and billions, so we would end up with KB, MB, BB 😄

Overall, it's looking like `numbro` does the formatting that we really need with a pretty readable API, and so I'm sensing that ideally we'd move toward consolidation on `numbro`. I want to test this with the decimal formatting and time formatting to confirm, but that's my sense so far. 

Also, based on discussion today, I'm seeing a possible future where we don't need to worry about having the "full number" in the y-axis. In that possible future, we can remove `Intl.NumberFormat` ✂️

juliepagano: Ah gotcha, that makes sense!

saminzadeh (approved):
❤️ numbro

PR #1085, opened 2023-04-10
[BUGFIX] Respect `series_name_format` in legend and tooltip when metric labels are empty

sjcobb:
Fixes a regression from my PR #1045, in that PR I fixed an issue with the series name showing as `{}` when a metric's labels were empty, but I caused an issue when `series_name_format` is set, where it was no longer respected when labels are empty (still worked normally in all other cases). I've added additional unit tests to catch this.

## Broken test case before bugfix:
series_name_format_broken_test_case

## After Fix
image

# Checklist

## UI Changes

### DIFF
--- a/ui/prometheus-plugin/src/utils/utils.test.ts
+++ b/ui/prometheus-plugin/src/utils/utils.test.ts
@@ -160,8 +160,16 @@ describe('getFormattedPrometheusSeriesName', () => {
   it('should resolve empty metric to instead show query', () => {
     const query = 'node_load15{instance=~"(demo.do.prometheus.io:9100)"';
     const metric = {};
-    const output = { name: query, formattedName: 'node_load15{instance=~"(demo.do.prometheus.io:9100)"' };
+    const output = { name: query, formattedName: query };
     expect(getFormattedPrometheusSeriesName(query, metric)).toEqual(output);
+  });
+
+  it('should resolve empty metric with formatted series name', () => {
+    const query = 'up';
+    const metric = {};
+    const series_name_format = 'Custom series name';
+    const output = { name: query, formattedName: 'Custom series name' };
+    expect(getFormattedPrometheusSeriesName(query, metric, series_name_format)).toEqual(output);
   });
 
   it('should show correct formatted series name', () => {
@@ -178,5 +186,20 @@ describe('getFormattedPrometheusSeriesName', () => {
       name: 'node_memory_Buffers_bytes{env="demo",instance="demo.do.prometheus.io:9100",job="node"}',
     };
     expect(getFormattedPrometheusSeriesName(query, metric, series_name_format)).toEqual(output);
+  });
+
+  it('should show correct raw series name', () => {
+    const query = 'node_load15{instance=~"(demo.do.prometheus.io:9100)"';
+    const metric = {
+      __name__: 'node_memory_Buffers_bytes',
+      env: 'demo',
+      instance: 'demo.do.prometheus.io:9100',
+      job: 'node',
+    };
+    const output = {
+      name: 'node_memory_Buffers_bytes{env="demo",instance="demo.do.prometheus.io:9100",job="node"}',
+      formattedName: 'node_memory_Buffers_bytes{env="demo",instance="demo.do.prometheus.io:9100",job="node"}',
+    };
+    expect(getFormattedPrometheusSeriesName(query, metric)).toEqual(output);
   });
 });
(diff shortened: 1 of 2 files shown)


ui/prometheus-plugin/src/utils/utils.test.ts
```diff
@@ -160,10 +160,18 @@ describe('getFormattedPrometheusSeriesName', () => {
...
-    const output = { name: query };
+    const output = { name: query, formattedName: query };
     expect(getFormattedPrometheusSeriesName(query, metric)).toEqual(output);
   });
 
+  it('should resolve empty metric with formatted series name', () => {
+    const query = 'up';
+    const metric = {};
+    const series_name_format = 'Custom series name';
+    const output = { name: query, formattedName: 'Custom series name' };
```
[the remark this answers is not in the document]
juliepagano: I wonder if it's worth breaking out just the basics of that panel story into a separate PR from the query refactor, so we can get it out for use faster. No strong feels about it, but might be something you could collaborate with Eunice on since it seems related to a couple of your PRs.

juliepagano (approved):
LGTM! Thanks for adding those helpful unit tests. 🚀

ui/prometheus-plugin/src/utils/utils.test.ts
```diff
@@ -160,10 +160,18 @@ describe('getFormattedPrometheusSeriesName', () => {
...
     const metric = {};
-    const output = { name: query };
+    const output = { name: query, formattedName: query };
     expect(getFormattedPrometheusSeriesName(query, metric)).toEqual(output);
   });
 
+  it('should resolve empty metric with formatted series name', () => {
+    const query = 'up';
+    const metric = {};
+    const series_name_format = 'Custom series name';
```
cndonovan, line 170: Should this kick in regardless of whether there are labels? 

Maybe that's how the logic works, I haven't checked logic yet, I'm just reading the test and thinking: If someone types `my custom series name` into the input, that should show up regardless of whether the metric has labels or not, right?

ui/prometheus-plugin/src/utils/utils.test.ts
```diff
@@ -179,4 +187,19 @@ describe('getFormattedPrometheusSeriesName', () => {
...
+
+  it('should show correct raw series name', () => {
+    const query = 'node_load15{instance=~"(demo.do.prometheus.io:9100)"';
+    const metric = {
+      __name__: 'node_memory_Buffers_bytes',
+      env: 'demo',
+      instance: 'demo.do.prometheus.io:9100',
+      job: 'node',
+    };
+    const output = {
```
cndonovan, line 199: minorrr, could be worth throwing in `series_name_format: undefined`, just to clear that this is the test case where `series_name_format` is undefined

Or adding that to name of test ("when series_name_format is undefined")

cndonovan (approved):
Double-checking: Does formattedName get priority over name? In other words, if someone writes "Errors" as the `series_name_format`, that will always get priority, right? 

And/or I am describing the desired behavior?

ui/prometheus-plugin/src/utils/utils.test.ts
sjcobb: Yeah me double check the tests, but that is what this is doing right? If `series_name_format` is 'Custom series name' then the output has 'Custom series name' as the formattedName.

formattedName is then what's displayed, but I kept around the unformatted name just in case we want to show it on hover or some other use case in the future

cndonovan: Ok sweet that sounds good -- I think I just wasn't sure because the test is mentioning an empty metric, and I wasn't sure if that was a requirement for this behavior

sjcobb: Oh yeahh the wording is confusing, that is expected behavior for anytime series_name_format is defined. I changed the test cases around to make it more clear

ui/prometheus-plugin/src/utils/utils.test.ts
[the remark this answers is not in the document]
sjcobb: I did end up adding a panel story in a different branch, I think this bug has decent coverage from the unit tests now so going to go ahead and merge. Planning to add visual regression tests for more of the tooltip and legend behavior later

sjcobb:
> Does formattedName get priority over name? In other words, if someone writes "Errors" as the `series_name_format`, that will always get priority, right?

Missed this before, yes that’s correct!

PR #1088, opened 2023-04-10
[ENHANCEMENT] Auto generated color palette with contrast 

sjcobb:
_UPDATE: default palette has been changed to 'Categorical' in PR #1124 and visual editing is removed (planning to merge that PR shortly after this one after all review comments are fixed)_ 

## Overview

**Goal**: Fix current default 'Auto' palette so that series names that are really similar, result in colors that aren't almost identical
- Result: this has been fixed in this PR using the [color-hash](https://github.com/zenozeng/color-hash) library!

Future goals: 
- Should we change our default palette?
- Can we adjust our Auto palette to be better (since color-hash gives full control over generated hues)?
- Proper color palette visual editing

New `palette.kind: 'Auto'` generated color palette hash algorithm that produces colors with more contrast than the previous approach. This uses series names as a seed when generating colors and is the default color palette. 

We are now using this npm package: https://github.com/zenozeng/color-hash

Customizations have been made to adjust the generated hues and a separate red hue generator is used whenever a series name has the word "error" in it.

Colors are consistent across panels (alias lookup happens before hsl generator function is called)

## Screenshots
<table>
<tr>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>image</td>
<td>image<td>
</tr>
<tr>
<td>image</td>
<td>image<td>
</tr>
</table>

After:
many_auto_colors_ex

After - [New Auto palette with consistent colors across panels and Categorical palette example screenshot](https://user-images.githubusercontent.com/9369625/230931695-36a64ced-b236-4f77-88f4-2ac248e8de3d.png)

After - Errors automatically generated as reddish hue:
errors_red_generated_ex

After - Basic palette visual editing (Categorical selected)
image

After - Basic palette visual editing (Auto selected)
image

# Checklist

## UI Changes


ui/components/src/utils/theme-gen.ts
```diff
@@ -39,7 +39,21 @@ export function generateChartsTheme(muiTheme: MuiTheme, echartsTheme: EChartsThe
       left: 20,
       containLabel: true,
     },
-    color: ['#8dd3c7', '#fb8072', '#80b1d3', '#bebada', '#fdb462'],
+    // Accessible categorical palette from: https://colorbrewer2.org/#type=qualitative&scheme=Paired&n=12
+    color: [
```
sjcobb, line 43: These static accessible colors are used for alternate `visual.palette.kind: 'Categorical'` and the default `visual.palette.kind: 'Auto'` is still generated from series names (so they are consistent across panels and can work with many series so colors don't repeat)

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -105,6 +105,27 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
           />
         }
       />
+      <OptionsEditorControl
+        label={'Palette'}
+        control={
+          <ToggleButtonGroup
```
sjcobb, line 111: Added palette visual editing quickly so we can switch back and forth between 'Categorical' and 'Auto' palettes, this UX is going to change drastically and the hard-coded ToggleButtonGroup was quickest (Mohit seemed okay with it now, but he prefers that it becomes a dropdown later) 

I can also spend a bit more time on this if people think it is worth it now?

juliepagano:
From a UX perspective (will look at the code next), I really like the categorical palette! It's visually distinct to me (I don't have color blindness) and, while definitely not perfect, usually has sufficient contrast when emulating the various forms of color blindness using the chrome dev tools (with it performing worst at achromatopsia).

In the examples I've looked at, I find the auto palette to still be challenging to differentiate in some cases. I'll defer to design's opinions about what the best thing is to do wit this one.

Example `up` query -- the chart is hard to look at because they're all the same value, but the legends provide a nice comparison. The left is the "auto" palette and the right is the "categorical" one.

## no emulation 
- auto: several of the green & blue colors look similar enough that I have trouble distinguishing them.
- categorical: all of the colors look distinct to me

image

## emulating protanopia (no red) 
image

## emulating deutaranopia (no green)

image

## emulating tritanopia (no blue)

image

## emulating achromatopsia (no color)

image

ui/components/src/utils/theme-gen.ts
```diff
@@ -39,7 +39,21 @@ export function generateChartsTheme(muiTheme: MuiTheme, echartsTheme: EChartsThe
       left: 20,
       containLabel: true,
     },
-    color: ['#8dd3c7', '#fb8072', '#80b1d3', '#bebada', '#fdb462'],
+    // Accessible categorical palette from: https://colorbrewer2.org/#type=qualitative&scheme=Paired&n=12
```
juliepagano, line 42: It's worth noting that they don't list that value as colorblind safe. If you check this checkbox, they'll filter it down.

image

It looks like the most the can do safely is:

11 diverging
image

9 sequential
image

4 categorical
image

juliepagano: This person has some recommendations that might also be useful.

https://personal.sron.nl/~pault/#sec:qualitative

sjcobb: Oh wow, yeah I missed that checkbox! So the only option for colorblind safe has four colors 😬

I really worry that repeating on the 5th color is a dealbreaker with the amount of data that we need to be able to show (so many queries are going to return series where many lines are exactly the same color), to me colorblind safe is less about every single color having the maximum amount of contrast and more that certain hues are favored since certain hues are easier for me to distinguish (people with other types of colorblindness could potentially opt-in to a different palette or maybe we give some control over generating hues in panel options later). The whole errors === 'red' hues has a great side effect where those hues are off limits to other metrics, meaning the 'Auto' palette favors the cooler side of the ROYGBIV spectrum. 

The other piece to emphasize is we have _complete_ control to adjust this based on feedback (the [color-hash library](https://github.com/zenozeng/color-hash#custom-hue) is so powerful and seems like a core architectural piece in Perses that we can keep iterating on).

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -105,6 +105,27 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
...
+            color="primary"
+            exclusive
+            value={value.palette?.kind ?? 'Auto'}
+            onChange={(__, newValue) => {
+              const palette: VisualOptions['palette'] =
```
juliepagano, line 116: Optional nit: it might be nice to pull this handler logic into a function. I find it's a little easier to read/maintain.

juliepagano (commented):
A handful of smaller code comments.

sjcobb:
> I really like the categorical palette!
> I'll defer to design's opinions about what the best thing is to do with this one.

I'm hoping we can address the "default" color palette question in a follow up PR once @foxbat07 has a chance to weigh in. For this PR, the goal is to fix the issue where our Auto palette doesn't work well when series names are really similar.

I also want to note that the 5-color Categorical palette we've chosen internally in our ChartsThemeProvider overrides is not nearly as good as the 12-color Categorical palette that I added in this PR for the Perses app. So again, if we do decide to change the current default, that'd give UX time to get back to me with new colors

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -105,6 +105,27 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
           />
         }
       />
+      <OptionsEditorControl
+        label={'Palette'}
+        control={
+          <ToggleButtonGroup
+            color="primary"
```
eunicorn, line 112: can you also fix the color in `ThresholdEditor` so the two toggle button groups look consistent?

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -11,39 +11,57 @@
...
+  const seriesColor = palette[colorIndex] ?? fallbackColor;
+  return seriesColor;
 }
+
+// Valid hue values are 0 to 360 and can be adjusted to control the generated colors.
+// More info: https://github.com/zenozeng/color-hash#custom-hue
+// Picked min of 30 and max of 360 to exclude common threshold colors (reddish).
+// Series names with "error" in them will always be generated as red.
+const ERROR_HUE_CUTOFF = 30;
+const colorGenerator = new ColorHash({ hue: { min: ERROR_HUE_CUTOFF, max: 360 } });
```
eunicorn, line 41: Does setting min to `ERROR_HUE_CUTOFF` mean we are excluding most/all red hue from color generator? 

I'm thinking if we should just set `ERROR_HUE_CUTOFF` to a smaller number or dedicate like 5-10 different shades of red for the `error` case,  so we get more of a variety of colors. In your screenshot where all series names include errors, I see some colors that i don't think it's very red.

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -11,39 +11,57 @@
...
-    if (seriesColor !== undefined) {
-      return seriesColor;
-    }
-  }
-
-  // corresponds to 'Auto' in palette.kind
-  return getRandomColor(name);
+export function getCategoricalPaletteColor(palette: string[], paletteIndex: number, fallbackColor: string): string {
+  // Loop through predefined static color palette
+  const colorIndex = paletteIndex % palette.length;
```
eunicorn, line 30: I like the idea where we reduce the opacity if it loops again, what do you think?

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
sjcobb: Yeah that's right, I wrote a little bit about this in my other comment [here](https://github.com/perses/perses/pull/1088#discussion_r1171696993) about colorblind friendliness. This is helpful though, I didn't realize some of the error colors weren't red, I reduced the ERROR_HUE_CUTOFF from 30 to 10!

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
sjcobb: Update: I'm thinking I'll merge this PR first and then follow up with the new approach in this PR #1124 based on Mohit's feedback. In that PR, the palette is automatically changed based on total series so it wouldn't ever need to loop through the new default Categorical palette a second time

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: This will be deleted in follow-up PR #1124 based on feedback from Mohit

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
sjcobb: This will be deleted in follow up PR #1124 based on feedback from Mohit

ui/internal-utils/src/utils/time-series/mockTimeSeriesResponse.ts
```diff
@@ -103,3 +103,73 @@ export function mockTimeSeriesResponseWithNullValues({
...
+  const exampleMetric = ['rpc_durations_histogram_seconds_bucket', 'caddy_http_request_duration_seconds_bucket'];
+  const exampleInstances = ['demo.do.prometheus.io:8996', 'rc-demo-environment/data/prom'];
+
+  const GAP_BETWEEN_SERIES_MULTIPLIER = totalSeries * 1000;
+  const STATIC_GAP = 10;
+
+  for (let i = 0; i < totalSeries; i++) {
+    const seriesIsEven = i % 2;
+    const metric: ExampleMetric = {
+      __name__: exampleMetric[seriesIsEven] ?? 'up',
```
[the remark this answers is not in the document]
sjcobb: Yeah it was to work around typing issues, I adjusted the ExampleMetric type so the fallbacks aren't needed anymore

ui/dashboards/src/components/Dashboard/Dashboard.stories.tsx
```diff
@@ -113,17 +150,27 @@ const DEFAULT_ALL_DASHBOARD: DashboardResource = {
             {
               x: 0,
               y: 0,
-              width: 10,
+              width: 12,
```
cndonovan, line 153: ooc, why did this one get wider?

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -163,13 +163,19 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
         const formattedSeriesName = timeSeries.formattedName ?? timeSeries.name;
 
-        const seriesColor = getSeriesColor(
-          formattedSeriesName,
-          seriesIndex,
-          echartsPalette as string[],
-          visual.palette?.kind
-        );
-        seriesIndex++; // Used for repeating colors in Categorical palette
+        // Fallback is unlikely to set unless echarts theme palette in charts theme provider is undefined.
```
cndonovan, line 166: Do you mean that the fallback is unlikely to be used unless `echartsPalette` is undefined?

I thought RedHat was the only one that was using the eCharts palette?

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -54,7 +54,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   } = props;
   const chartsTheme = useChartsTheme();
   const muiTheme = useTheme();
-  const echartsPalette = chartsTheme.echartsTheme.color ?? [muiTheme.palette.primary];
+  const echartsPalette = chartsTheme.echartsTheme.color;
```
cndonovan, line 57: What if we called this variable `categoricalPalette`? Do you think that would be more clear / does that accurately represent this?

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.test.ts
```diff
@@ -11,26 +11,50 @@
...
-    expect(value).toEqual('#111');
+    const paletteColor = getCategoricalPaletteColor(['#fff', '000', '#111', '#222', '#333'], 2, fallbackColor);
+    expect(paletteColor).toEqual('#111');
+  });
+
+  it('should repeat color after looping through entire palette', () => {
+    const paletteColor = getCategoricalPaletteColor(['#fff', '000', '#111', '#222', '#333'], 5, fallbackColor);
+    expect(paletteColor).toEqual('#fff');
   });
+});
```
cndonovan, line 33: Could add a test for when it returns `fallback` color? (if the array is undefined, I'm guessing)

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.test.ts
```diff
@@ -11,26 +11,50 @@
...
-    const value = getSeriesColor('p90 test api subdomain', 5, ['#fff', '000', '#111', '#222', '#333'], 'Categorical');
-    expect(value).toEqual('#fff');
+describe('getConsistentSeriesNameColor', () => {
+  it('should generate a consistent custom hsla color', () => {
+    const color = getConsistentSeriesNameColor('test');
+    const colorAlt = getConsistentSeriesNameColor('test');
+    const firstResult = 'hsla(287.30,35%,50%,0.9)';
+    // ensures generated color does not change on subsequent calls with same series name
+    expect(color).toEqual(firstResult);
+    expect(colorAlt).toEqual(firstResult);
```
cndonovan, line 58: <3

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -11,39 +11,57 @@
...
-    hash = identifier.charCodeAt(index) + ((hash << 5) - hash);
-  }
-  // Use HSLA to only get random "bright" colors from this
-  const color = `hsla(${~~(180 * hash)},50%,50%,0.8)`;
-  return color;
+export function getAutoPaletteColor(name: string, fallbackColor: string): string {
+  // corresponds to 'Auto' in palette.kind for generative color palette
+  const generatedColor = getConsistentSeriesNameColor(name);
+  return generatedColor ?? fallbackColor;
 }
```
cndonovan, line 23: I love this wrapper around getConsistentSeriesNameColor

cndonovan (approved):

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
sjcobb: This makes sense to me! I'm going to go ahead and merge this PR since CI is green and update variables based on this feedback in #1124 (I'm touching the getCategoricalPaletteColor function there too)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Just talked with Mohit and we're going to change the hex values to be more colorblind friendly, I'll make this change as part of that PR!

ui/dashboards/src/components/Dashboard/Dashboard.stories.tsx
sjcobb: Oh sorry missed this comment! This probably happened at some point during rebasing since this branch started before the queries breaking changes. I'll update the other panel to be same width since it's weird that one is 10 and one is 12 now

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.test.ts
cndonovan: <3 it

PR #1095, opened 2023-04-11
[ENHANCEMENT] Add more unit tests for formatDecimal

cndonovan:
## Changes

* Add more units tests for formatDecimal (take a look and let me know if you see more cases)
* Use `Intl.NumberFormat` `{ notation: 'compact' }`. This seems to cover what we care about, and means less code to maintain.


ui/components/src/model/units/decimal.ts
```diff
@@ -33,50 +32,24 @@ export const DECIMAL_UNIT_CONFIG: Readonly<Record<DecimalUnitKind, UnitConfig>>
...
-  const decimalFormatter = new Intl.NumberFormat('en-US', formatParams);
-  return decimalFormatter.format(value);
-}
-
-/**
- * Takes large numbers and abbreviates them with the appropriate suffix
- * 10123 -> 10.123k
- * 1000000 -> 1M
- */
-export function abbreviateLargeNumber(num: number, decimalPlaces = 2) {
```
sjcobb, line 61: Trying to remember, but we might be using this function somewhere internally btw

sjcobb: Nvm just checked, we aren't importing this one anywhere! There is a similar `abbreviateLargeNumber` function in that repo that I was thinking of

ui/components/src/model/units/decimal.test.ts
```diff
@@ -0,0 +1,294 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { formatValue } from './units';
+import { UnitTestCase } from './units.test';
+
+const decimalTests: UnitTestCase[] = [
```
sjcobb, line 17: 🥳

ui/components/src/model/units/decimal.ts
cndonovan: I just deleted them without thinking if they were used outside of this file :)

PR #1062, opened 2023-04-03
[FEATURE] Revamp project page

Gladorme:
# Description

Revamping project page to the [design selected](https://www.figma.com/file/D9VznEcxMYrhyVwSb6fLuf/Perses-Design?node-id=144-6403). The dashboard listing has been replaced by MUI DataGrid component allowing user to search, to filter and to sort items easily.

# Images

## Project page
image

## Home page
image

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).
- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.

PR #1097, opened 2023-04-12
[FEATURE] Landing page revamp

Gladorme:
# Description
Revamping project page to the [design selected](https://www.figma.com/file/D9VznEcxMYrhyVwSb6fLuf/Perses-Design?node-id=144-6403).
The home page has now 4 main sections:
- Recent Dashboards: show the latest dashboard viewed
- Projects & Dashboard: same listing as before with a new search bar in order to filter on project name, dashboard name and dashboard display name
- Important Dashboards: show "important" dashboard setup in the config
- Information: basic information section provided by the back (config), it's rendering Markdown content

I added a loader on all dashboard lists too 😄 

# Image
## Home Page / Landing Page
image

## Project page
image
Small change on button appearance

# Checklist

## UI Changes


Gladorme:
Hola @sjcobb & @Nexucis, the PR is now ready to be reviewed 😄

ui/app/src/components/DashboardCard/DashboardCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface DashboardCardProps {
+  dashboard: DashboardResource;
+}
+
+export function DashboardCard(props: DashboardCardProps) {
+  const naviguate = useNavigate();
+
+  return (
+    <Button
```
eunicorn, line 29: The color of these buttons seem a little too bright. Maybe we should use the same color as the top navbar?

ui/app/src/components/DashboardCard/DashboardCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { Button, Stack, Typography } from '@mui/material';
+import ViewDashboardIcon from 'mdi-material-ui/ViewDashboard';
+import { DashboardResource } from '@perses-dev/core';
+import { dashboardDisplayName } from '@perses-dev/core/dist/utils/text';
+import FolderPoundIcon from 'mdi-material-ui/FolderPound';
+import { useNavigate } from 'react-router-dom';
+
+interface DashboardCardProps {
+  dashboard: DashboardResource;
```
eunicorn, line 22: How about just passing the `dashboard.metadata` only?

ui/app/src/components/CreateDashboardDialog/CreateDashboardInProjectDialog.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Dialog used to create a dashboard.
+ * @param props.open Define if the dialog should be opened or not.
+ * @param props.closeDialog Provides the function to close itself.
+ * @param props.onConfirm Action to perform when user confirmed.
+ * @param props.projectName The project where the dashboard will be created.
+ * @constructor
+ */
+export const CreateDashboardInProjectDialog = (props: CreateDashboardInProjectProps) => {
```
eunicorn, line 34: We already have a `CreateDashboardDialog`. Is it possible to refactor that dialog so it works here as well?

ui/app/src/components/DashboardCard/DashboardCard.tsx
Gladorme: It will look like this, is it better?
image
Using the same background color (`theme.palette.designSystem.blue[700]`)

Gladorme:
Thank you @eunicorn for your review, PR has been updated 😃

ui/app/src/components/CreateDashboardDialog/CreateDashboardDialog.tsx
```diff
@@ -68,21 +84,46 @@ export const CreateDashboardDialog = (props: CreateDashboardDialogProps) => {
...
-          fullWidth
-          onChange={handleChange}
-          value={name}
-          error={!!error}
-          helperText={error}
-        />
+        <Stack gap={1}>
+          {projectOptions && projectOptions.length > 1 && (
+            <FormControl size="small" fullWidth>
+              <InputLabel>Project name</InputLabel>
```
eunicorn, line 90:
```suggestion
              <InputLabel id="project-name-id">Project name</InputLabel>
              <Select labelId="project-name-id">
```

would be good to  add `id` and `labelId` here for accessibility reasons

ui/app/src/components/DashboardList/DashboardList.tsx
```diff
@@ -73,6 +83,28 @@ export function DashboardList(props: DashboardListProperties) {
     [getDashboard]
   );
 
+  // Merging default initial state with the props initial state (props initial state will overwrite properties)
+  const mergedInitialState = useMemo(() => {
+    return {
+      ...{
```
eunicorn, line 89: nit: pulling this out into a constant would make this easier to read

ui/app/src/model/dashboard-client.ts
```diff
@@ -91,6 +92,32 @@ export function useRecentDashboardList(project?: string) {
 
     return result;
   }, [data, history]);
+
+  return { data: result, isLoading: isLoading };
+}
+
+/**
+ * Used to get dashboards seen recently by the user.
```
eunicorn, line 100:
```suggestion
 * Used to get important dashboards
```

ui/app/src/views/home/HomeView.tsx
```diff
@@ -0,0 +1,148 @@
...
+    (dashboardSelector: DashboardSelector) =>
+      navigate(`/projects/${dashboardSelector.project}/dashboards/${dashboardSelector.dashboard}/create`),
+    [navigate]
+  );
+
+  // Open/Close management for dialogs
+  const [isAddProjectDialogOpen, setIsAddProjectDialogOpen] = useState(false);
+  const [isAddDashboardDialogOpen, setIsAddDashboardDialogOpen] = useState(false);
+
+  const handleAddProjectDialogOpen = useCallback(() => {
```
eunicorn, line 56: IMO i think it's ok to skip `useCallback` for simple functions like these. I don't think it'll be too much of a performance impact

ui/app/src/views/home/HomeView.tsx
```diff
@@ -0,0 +1,148 @@
...
+            <CreateDashboardDialog
+              open={isAddDashboardDialogOpen}
+              projectOptions={projectOptions}
+              onClose={handleAddDashboardDialogClose}
+              onSuccess={handleAddDashboardDialogSubmit}
+            />
+          </Stack>
+        </Stack>
+      </Box>
+      <Grid container spacing={8}>
```
eunicorn, line 110: What do you think about making each grid into its own component?

```suggestion
      <RecentDashboards />
      <ProjectsAndDashboards />
      <ImportantDashboards/>
```

ui/app/src/views/home/ImportantDashboards.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { Box } from '@mui/material';
+import { useImportantDashboardList } from '../../model/dashboard-client';
+import { DashboardList } from '../../components/DashboardList/DashboardList';
+
+interface ImportantDashboardProps {
+  id?: string;
+}
+
+export function ImportantDashboard(props: ImportantDashboardProps) {
```
eunicorn, line 22: Maybe rename this to `ImportantDashboardList`?

eunicorn:
I think i found a bug where the newly created project wasn't showing up on the Home View. When I try adding it again with the same name, i get an error `document already exists` 

bug-add-project

ui/e2e/src/tests/homeView.spec.ts
eunicorn, line 1: Yay thanks for adding tests!

eunicorn:
didn't see it anywhere in the UI, but how do you add a dashboard to `Important Dashboards`?

ui/app/src/views/home/ImportantDashboards.tsx
Gladorme: I will move it at the same place as RecentDashboardList & DashboardList components

ui/app/src/views/home/RecentDashboardsMosaic.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return (
+      <Stack sx={{ alignItems: 'center', justifyContent: 'center' }} width="100%" height="50">
+        <Typography variant={'subtitle1'}>No dashboards recently viewed</Typography>
+      </Stack>
+    );
+  }
+
+  return (
+    <>
+      {data.slice(0, 6).map((datedDashboard) => (
```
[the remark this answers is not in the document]
Gladorme: Sure

Gladorme:
> 



> I think i found a bug where the newly created project wasn't showing up on the Home View. When I try adding it again with the same name, i get an error `document already exists`
> 
> Looks like there needs to be a saved dashboard in the project in order for it to show up here?
> 
> bug-add-project [ bug-add-project ](https://user-images.githubusercontent.com/9817826/234093276-30a87b2f-dbbb-4d0e-be45-fc36abe8b664.gif) [ ](https://user-images.githubusercontent.com/9817826/234093276-30a87b2f-dbbb-4d0e-be45-fc36abe8b664.gif)
> 
> Also after deleting a project, it doesn't look like the project & dashboard list is updated.

Yes, you need to have dashboards in a project in order to be displayed here. I didn't change the logic of the component, except with the search.
When you create a project, you should be redirected to the project page. I don't understand why is it not working for you. Could you try again in order to confirm? 
And about the project deletion, I confirm there is an issue, I will check

Gladorme:
Thanks for the review @eunicorn! I added/fix all your feedback, let me know if what I did for home components is fine for you. 
Nice catch on the delete project. Now when a project is deleted, dashboard query cache is invalidated too.
About important dashboards, they are retrieved from the backend config. You can check in the dev config how to mark a dashboard as important 😄

ui/app/src/views/home/HomeView.tsx
eunicorn: This is great, much easier to read. Thanks!

ui/app/src/model/config-client.ts
```diff
@@ -52,3 +58,29 @@ export function useIsReadonly() {
   }
   return data.readonly;
 }
+
+export function useDashboardLists() {
```
Nexucis, line 62: to avoid a confusion with the function coming from `dashboard-client.ts` I would rename it `useConfigDashboardLists`

dev/config.yaml
```diff
@@ -10,4 +12,15 @@ schemas:
   variables_path: "schemas/variables"
   interval: "5m"
 
-readonly: false
+dashboard_lists:
+  importants:
```
Nexucis, line 16: by looking at the code in the frontend, it looks like this keyword is actually required to use it properly.

The config shouldn't let the user decide which name it can use if we are looking for a special keyword. The config should reflect it.
Finally by looking at the UI currently it's not possible to display other sections than the "important" one.

Did you want to provide a way to configure multiple list of important dashboard? In that case the key would be the title of the section and then you are looping other the map and create multiple card with the different dashboards.
Like we have in Grafana in Amadeus I guess

Nexucis: Probably because in the current UX design we can just display a single list. Maybe we should just have list of dashboards instead of a map.
And then later we could think about moving to a map + having a proper design for that

Nexucis (commented):
looks nice and I will definitely rely on UI expert here (thank you @eunicorn for taking care about it)

I just have some reserve about the config.

eunicorn:
> When you create a project, you should be redirected to the project page. I don't understand why is it not working for you. Could you try again in order to confirm?

It did take me to the project page after creating a project, but then I clicked "home" in the breadcrumb to go back to the home page.

dev/config.yaml
Gladorme: Removed the multi list and moved important dashboards in a "basic" list 😄

ui/app/package.json
```diff
@@ -17,6 +17,7 @@
     "@loadable/component": "^5.15.2",
     "@mui/material": "^5.10.14",
     "@mui/x-data-grid": "^6.0.3",
+    "@nexucis/kvsearch": "^0.8.1",
```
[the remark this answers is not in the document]
Gladorme: It's a lib for filtering items in an array on specific fields or with a query (feature not used). About the second part, same for me never heard of him 😋

sjcobb:
Random comment, it looks like we ended up with a package-lock.json file in the [root of our repo](https://github.com/perses/perses/blob/main/package-lock.json) (possibly after #1062), can that be deleted?

Nexucis:
> Random comment, it looks like we ended up with a package-lock.json file in the [root of our repo](https://github.com/perses/perses/blob/main/package-lock.json) (possibly after #1062), can that be deleted?

yeah it is totally safe to remove it !

dev/config.yaml
```diff
@@ -10,4 +12,14 @@ schemas:
   variables_path: "schemas/variables"
   interval: "5m"
 
-readonly: false
+important_dashboards:
```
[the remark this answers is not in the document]
Gladorme: Yes, it's based on the order in the config.

Nexucis (approved):
looks really great. Thank you for all these changes !

As a side note, yeah I agree the search is a bit weird when you are typing `Demo` and it is still displays every dashboards. Probably we can make perform two filtering (one for the project, one for the dashboard) and then merge both. Pretty sure it will work and the result will be much more accurate.
We could even highlight what is matching because behind the scene it is using a fuzzy algorithm so searching `Demo` could match `DaEkdsMsdO`

eunicorn (approved):
looks good to me 💯

PR #1101, opened 2023-04-12
[ENHANCEMENT] Center value when sparkline is undefined 

eunicorn:
When disabling sparkline, the value will be automatically centered.

image

### DIFF
--- a/ui/components/src/StatChart/StatChart.tsx
+++ b/ui/components/src/StatChart/StatChart.tsx
@@ -109,7 +109,7 @@ export function StatChart(props: StatChartProps) {
   const textStyles = {
     display: 'flex',
     flexDirection: 'column',
-    justifyContent: textAlignment,
+    justifyContent: sparkline ? 'auto' : textAlignment,
     alignItems: textAlignment,
   };
 
@@ -118,12 +118,12 @@ export function StatChart(props: StatChartProps) {
       <Box sx={{ flexGrow: 1, ...textStyles }}>
         <Typography
           variant="h3"
-          sx={(theme) => ({
-            color: theme.palette.text.primary,
+          sx={{
+            color,
             fontSize: `clamp(${MIN_VALUE_SIZE}px, ${valueSize}px, ${MAX_VALUE_SIZE}px)`,
             padding: `${containerPadding} ${containerPadding} 0 ${containerPadding}`,
             zIndex: '1', // set zIndex to 1 so text goes on top of graph
-          })}
+          }}
         >
           {formattedValue}
         </Typography>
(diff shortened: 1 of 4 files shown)


ui/components/src/StatChart/StatChart.tsx
```diff
@@ -104,24 +104,43 @@ export function StatChart(props: StatChartProps) {
...
+            fontSize: `clamp(${MIN_VALUE_SIZE}px, ${valueSize}px, ${MAX_VALUE_SIZE}px)`,
+            padding: `${containerPadding} ${containerPadding} 0 ${containerPadding}`,
+          }}
+        >
+          {formattedValue}
+        </Typography>
+      </Box>
       {sparkline !== undefined && (
-        <Box sx={{ flex: 1 }}>
+        <Box
```
juliepagano, line 130: If the stat chart fills the entire content of the panel, isn't there a risk of the line and the text overlapping, making the text hard to read?

sjcobb: +1 I'm curious about the wrapping behavior

sjcobb: Sorry to add scope creep to this discussion, this is separate from this PR, but while I'm thinking about it--this [Node Exporter dashboard](https://grafana.com/grafana/dashboards/1860-node-exporter-full/), is one of the most ubiquitous Grafana dashboards out there. It has a ton of Stat charts all displayed together at really small dimensions (which really throws off our panel header too and Gauge chart), here is how it looks when using the Perses app migration feature: 

image

sjcobb (approved):
Much better! Added a question but nothing blocking, hopefully the clamp fontSize approach scales the text down so it's unlikely to need to wrap very often 🤞

ui/components/src/StatChart/StatChart.tsx
eunicorn: Yeah i notice grafana does a lot of auto formatting with stat charts. Here it automatically hides the sparkline when chart gets too small.  The problem also gets way more complicated once we support multiple series for stat charts

stat chart

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -117,17 +125,15 @@ export function StatChart(props: StatChartProps) {
         {formattedValue}
       </Typography>
       {sparkline !== undefined && (
-        <Box sx={{ flex: 1 }}>
```
eunicorn, line 120: @sjcobb curious if there was a reason you added `flex: 1` here... i removed it because i didn't notice any difference but maybe i'm missing something

ui/components/src/StatChart/StatChart.tsx
eunicorn: > If the stat chart fills the entire content of the panel, isn't there a risk of the line and the text overlapping, making the text hard to read?

Fixed

ui/components/src/StatChart/StatChart.tsx
juliepagano: I think I did this ages ago to fix some css issues. With the `EChart` component having height set to `100%`, I think you're getting the same effect, so it should be ok to adjust it.

juliepagano (approved):
LGTM!

PR #1104, opened 2023-04-13
[ENHANCEMENT] Add help text and placeholder for series_name_format control

sjcobb:
Add context when using our Prom query editor so that users know how to use the `series_name_format` input, open to feedback on the wording!

## Screenshot
image

## Recording

https://user-images.githubusercontent.com/9369625/231636902-d2e2a5d4-abd7-4de7-a76d-9784d392347a.mp4

# Checklist

## UI Changes

### DIFF
--- a/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
+++ b/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
@@ -57,9 +57,9 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
       />
       <TextField
         fullWidth
-        label="Series Name Format"
-        placeholder="Templating ex: {{instance}} will be replaced with label value for instance"
-        helperText="Controls legend and tooltip formatting"
+        label="Name"
+        placeholder="Tip: Use {{variable_name}}. Example: {{instance}} will be replaced with values such as 'webserver-123' and 'webserver-456'."
+        helperText="Set the name for each series in the legend and the tooltip."
         value={format ?? ''}
         onChange={(e) => handleFormatChange(e.target.value)}
         onBlur={handleFormatBlur}
--- a/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
+++ b/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
@@ -57,8 +57,8 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
       />
       <TextField
         fullWidth
-        label="Name"
-        placeholder="Tip: Use {{variable_name}}. Example: {{instance}} will be replaced with values such as 'webserver-123' and 'webserver-456'."
+        label="Legend Name"
+        placeholder="Tip: Use {{label_name}}. Example: {{instance}} will be replaced with values such as 'webserver-123' and 'webserver-456'."
         helperText="Set the name for each series in the legend and the tooltip."
         value={format ?? ''}
         onChange={(e) => handleFormatChange(e.target.value)}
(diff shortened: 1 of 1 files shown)


ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -58,10 +58,11 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
       <TextField
         fullWidth
         label="Series Name Format"
```
cndonovan, line 60: I changed this to "Name".

That might be controversial, but my thinking was:

- people don't know what we mean when we say "Series Name Format"
- `Name` is simple, and pretty close to what's happening (giving a name to each series)
- The example and helper text provide details to make it clear
- It covers the fact that we don't always use the formatting with `{{ }}`, sometimes we just enter text

cndonovan (commented):
Here was my best go at it: 

Screenshot 2023-04-14 at 9 56 04 AM

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
sjcobb: I'm good with this! It matches how it's defined in the JSON close enough and really like your suggestion for `Tip`

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -57,11 +57,12 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
...
       <TextField
         fullWidth
-        label="Series Name Format"
+        label="Name"
+        placeholder="Tip: Use {{variable_name}}. Example: {{instance}} will be replaced with values such as 'webserver-123' and 'webserver-456'."
+        helperText="Set the name for each series in the legend and the tooltip."
         value={format ?? ''}
         onChange={(e) => handleFormatChange(e.target.value)}
         onBlur={handleFormatBlur}
-        margin="dense"
```
cndonovan, line 64: Screenshot: 

Screenshot 2023-04-26 at 1 55 24 PM

@sjcobb Am I right to say `variable_name`? Or is `label_name`? (i.e. should they use the name of the label on the metric, or should they use the name of one of the variables that they created)?

sjcobb: label_name is correct! There is overlap since a lot of times the variables can match but what’s inside the handlebars {{ }} syntax is just regex that has nothing to do with variable context. (We do most likely also need to also start supporting template variable `$` syntax too, there is a bug ticket about this)

sjcobb: Maybe "Legend Name" for the TextField label? I think that's the only variation we haven't discussed yet? 😂

sjcobb: Pushed up 'Name' -> 'Legend Name' change, we can always adjust again later and I might propose in Matrix chat that we change the JSON from `series_name_format` to `legend_format` since it's best to do the breaking change sooner rather than later while we don't have many users

sjcobb: Also made the 'variable_name' to 'label_name' change in my commit [here](https://github.com/perses/perses/pull/1104/commits/74dcafe0e14e50f08ce8914e990a8627b70e04b1)

cndonovan: I like it - I also just pushed up to this branch because I had not reloaded this page yet to see your changes, but should be fine since I think it's already deleted

cndonovan: I like Legend Name a lot actually

PR #1107, opened 2023-04-14
[ENHANCEMENT] Highlight behavior slight improvement using opacity blur

sjcobb:
Very slight improvement to TimeSeriesChart hover / focused series emphasis behavior. Tried adjusting non-focused series line opacity on blur. Removed the line width increase on emphasis and added: 
- https://echarts.apache.org/en/option.html#series-line.emphasis.focus
- https://echarts.apache.org/en/option.html#series-line.emphasis.blurScope

## Before

https://user-images.githubusercontent.com/9369625/232081609-d278b38e-8137-49d1-9ed0-62b32498507a.mp4

### Default
https://user-images.githubusercontent.com/9369625/232081659-e19e83aa-eab4-4c99-b80c-1102e959ece5.mp4

### `visual.show_points: 'Always'`

https://user-images.githubusercontent.com/9369625/232084265-ab546f15-82a0-4561-bc8e-948f513cde11.mp4

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).

PR #1112, opened 2023-04-17
[BUGFIX] Line highlight on hover refactor using ECharts dispatch 

sjcobb:
~~TBD - need to fix flicker issue due to limitation described in~~: https://github.com/apache/echarts/issues/18495
**UPDATE**: Flicker issue is fixed now using Julie's `axisPointer.triggerEmphasis` approach, this PR is ready for review!

## Overview
Early on in the Perses project, we decided to build a custom tooltip with React so that we could have full control. This allows us to only show nearby series in the tooltip, better performance with 1000+ series, and ability to easily format the tooltip differently depending on the number of focused series. Since we went down this path, that also means we need to build custom highlighting behavior since the number of focused series can vary depending on what is set in our `yBuffer` param in the `getNearbySeries` function. 

This PR improves the implementation by using EChart's dispatch API so we can trigger custom [highlight](https://echarts.apache.org/en/api.html#action.highlight) and [downplay](https://echarts.apache.org/en/api.html#action.downplay) actions depending on seriesIndex conditions and which series are currently in the `currentFocusedData` array. 

**_Now hover styles are applied correctly so that only the focused series showing in the tooltip are emphasized!_**

The blur workaround in #1107 is also no longer needed.

## Screen Recordings
[Loom demo](https://www.loom.com/share/d79d2e29213c4c50990904c4a3514235) showing changes and ECharts dispatch API docs

Local example 1
https://user-images.githubusercontent.com/9369625/232504585-8ee1b0fd-41da-4f00-8238-4d4435668c71.mp4

Local example 2
https://user-images.githubusercontent.com/9369625/232507723-7a4c1088-2283-49ae-8cac-7b6fe33c3755.mp4

# Checklist

## UI Changes


juliepagano:
It's sort of separate from your change here, but I find our current approach to be kind of confusing in some cases like this one where you get this tooltip flicker of some items that occasionally have peaks near where you're moving the mouse left/right. I wonder if we need to have some chats with UX about the ideal behavior for tooltips/highlights/etc., so we can identify an ideal behavior for tooltip/emphasis/etc. we want to work towards.

https://user-images.githubusercontent.com/396962/232570754-5aba4ec5-2b4d-48e4-abb4-257d8b8ad2f5.mov

juliepagano:
Separate from my previous note, the lines seem extra flicker-y with this change. My guess is that the highlight & downplay events you're triggering are still fighting with the highlight/trigger events from the axis pointer.

https://user-images.githubusercontent.com/396962/232577280-c98be6ff-a7a4-4d00-ac19-a06ee27ccaad.mov

sjcobb:
> Separate from my previous note, the lines seem extra flicker-y with this change. My guess is that the highlight & downplay events you're triggering are still fighting with the highlight/trigger events from the axis pointer.

Using the nightly build, [your fix](https://github.com/apache/echarts/pull/18524) in ECharts is definitely helping with the flicker! 

One downside is when we set `axisPointer.triggerEmphasis` to false, the points that move with your cursor, stop showing. To adjust for this, I decided to [bring back symbols when the chart renders](https://github.com/perses/perses/pull/1112/files#diff-7717d573f17b47724d0d4db71dbd50982c25a7d7f059d9aa4f136e1bd2d0d69c), but only when there aren't a ton of datapoints (roughly correlates to when the dashboard time range is 15 minutes or less). 

This was something I wanted to do anyways and it makes sense to go out with the highlighting improvements, since having better highlighting makes up for cases when symbols don't render (note: HIDE_DATAPOINTS_LIMIT condition is also important since rendering too many symbols can hurt performance).

sjcobb:
> you get this tooltip flicker of some items that occasionally have peaks near where you're moving the mouse left/right

I've adjusted variables that control how many series show and added a new SHOW_MORE_NEARBY_SERIES_LIMIT condition that will help with this. Now when there are 5 or less series, it will almost always show all 5, but when there are more series than will fit in the tooltip (w/o pinning and scrolling), the yBuffer area is smaller. 

It doesn't completely address your concern, but between that and fixing the line color flicker using triggerEmphasis, it is a big improvement. I am planning to add panel options to override default tooltip behavior in the future. I need to sync with Mohit again, but these options could make sense:
- Single Series
- Nearby Series (Default)
- All Series

ui/components/package.json
```diff
@@ -40,7 +40,7 @@
     "@uiw/react-codemirror": "^4.19.1",
     "date-fns": "^2.28.0",
     "date-fns-tz": "^1.3.7",
-    "echarts": "^5.4.1",
+    "echarts": "npm:echarts-nightly@^5.5.0-dev.20230425",
```
sjcobb, line 43: This nightly build is needed to pull in Julie's ECharts [PR here](https://github.com/apache/echarts/pull/18524), which fixes cases where lines would flicker as they were de-emphasized. I'm hoping the next ECharts minor release comes out soon, but so far I haven't found any bugs and it seems like the `v5.5` branch has had a [decent amount of time](https://www.npmjs.com/package/echarts-nightly?activeTab=versions) to bake so maybe the risk is okay assuming we thoroughly test other use cases internally? 🤷‍♂️

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
```diff
@@ -64,25 +74,24 @@ export function getLineSeries(
...
+      opacity: 0.85,
     },
     areaStyle: {
       opacity: visual.area_opacity ?? DEFAULT_AREA_OPACITY,
     },
+    // https://echarts.apache.org/en/option.html#series-line.emphasis
     emphasis: {
-      focus: 'series',
-      blurScope: 'coordinateSystem',
+      focus: 'none', // 'none', 'self', 'series'
```
cndonovan, line 89: Is this not needed anymore because you're handling emphasis above?

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
```diff
@@ -55,6 +57,14 @@ export function getLineSeries(
 ): EChartsTimeSeries {
   const lineWidth = visual.line_width ?? DEFAULT_LINE_WIDTH;
   const pointRadius = visual.point_radius ?? DEFAULT_POINT_RADIUS;
+
+  // Shows datapoint symbols when selected time range is roughly 15 minutes or less
+  let showPoints = data.length <= HIDE_DATAPOINTS_LIMIT;
```
cndonovan, line 62: What are datapoint symbols?

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -170,10 +170,14 @@ export function LineChart({
...
-        },
+      },
+      // https://echarts.apache.org/en/option.html#axisPointer
+      axisPointer: {
+        type: isOptimizedMode ? 'none' : 'line',
+        z: 0, // ensure point symbol shows on top of dashed line
+        triggerEmphasis: false, // https://github.com/apache/echarts/issues/18495
+        triggerTooltip: false,
+        snap: true,
       },
```
cndonovan, line 181: I guess the `axisPointer` field moved to be less nested?

ui/components/src/TimeSeriesTooltip/focused-series.test.ts
```diff
@@ -62,7 +62,7 @@ describe('getNearbySeries', () => {
       kind: 'Decimal',
       decimal_places: 2,
     };
-    expect(getNearbySeries(chartData, pointInGrid, yBuffer, decimalUnit)).toEqual(focusedSeriesOutput);
+    expect(getNearbySeries(chartData, pointInGrid, yBuffer, undefined, decimalUnit)).toEqual(focusedSeriesOutput);
```
cndonovan, line 65: could be nice to name these someday, but not needed right now

i.e. `getNearbySeries({ chartData, pointInGrid, yBuffer, chart: undefined, unit: decimalUnit})`

ui/components/src/TimeSeriesTooltip/focused-series.ts
```diff
@@ -124,15 +146,30 @@ export function getFocusedSeriesData(
   const chartModel = chart['_model'];
   const yAxisInterval = chartModel.getComponent('yAxis').axis.scale._interval;
 
-  // tooltip trigger area gets smaller with more series, increase yAxisInterval multiplier to expand nearby series range
   const seriesNum = chartData.timeSeries.length;
-  const yBuffer = seriesNum > TOOLTIP_MAX_ITEMS ? yAxisInterval * 0.5 : yAxisInterval * 5;
+
+  // tooltip trigger area gets smaller with more series, increase yAxisInterval multiplier to expand nearby series range
+  const yBuffer =
+    seriesNum > SHOW_MORE_NEARBY_SERIES_LIMIT
```
cndonovan, line 153: i might call this `SHOW_FEWER_SERIES_LIMIT` because as the number of series grows from 1 --> more, when it reaches a certain point, we show fewer series. 

again up to you / not a problem as is!

cndonovan (approved):

ui/components/package.json
juliepagano: I think if we're pointing at a very specific nightly that we've tested against a bunch (as opposed to regularly pulling in the nightly) and keep an eye out for the next release and swap to it when it's ready, that it's probably ok.

juliepagano:
Still seeing a bit of flickering when you hover directly over lines. Any idea of that's something that can be worked around easily with some of the series emphasis settings? If not, I don't think it's too bad and can be revisited later. 


https://user-images.githubusercontent.com/396962/235540424-ff9e2670-48d3-4e08-ae74-d7da5bb32e58.mov

ui/components/src/LineChart/LineChart.tsx
juliepagano: IIRC, the nested one doesn't have as many options (e.g. the new `triggerEmphasis` I just added) and the top-level one is more comprehensive.

ui/components/src/TimeSeriesTooltip/focused-series.test.ts
juliepagano: +1 on this as a future thing. Once the number of args gets beyond a small handful, it's much easier to work with an object.

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
juliepagano: I think this is the little circle shown at points. It's called a "symbol" because you can use a variety of shapes. @sjcobb, correct me if I got that wrong.

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
juliepagano: Optional nit: It might be more helpful to have a comment clarifying why you set the value to "none" than the other possible values.

juliepagano (approved):
A couple small nits from me and Christine, but generally LGTM! Can you double-check with other folks that they're also ok with using the nightly before merging this?

ui/components/src/TimeSeriesTooltip/focused-series.ts
```diff
@@ -15,6 +15,11 @@ import { ECharts as EChartsInstance } from 'echarts/core';
 import { formatValue, UnitOptions, EChartsDataFormat } from '../model';
 import { CursorData, TOOLTIP_DATE_FORMAT, TOOLTIP_MAX_ITEMS } from './tooltip-model';
 
+// increase multipliers to show more series in tooltip
+export const DEFAULT_NEARBY_SERIES_RANGE_MULTIPLIER = 5.5; // adjusts how many focused series show in tooltip
```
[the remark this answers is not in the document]
sjcobb: I ended up renaming these constants so they are hopefully more clear, "narrow" was a little confusing and could make people think the range controls the width of the nearby series highlighted area, "shorten" is more accurate (since it adjusts yBuffer variable) but sounds weird so I landed on INCREASE_FOCUSED_SERIES_MULTIPLIER and REDUCE_FOCUSED_SERIES_MULTIPLIER for now

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
sjcobb: Yup that's right! [Example](https://echarts.apache.org/en/option.html#series-line.symbol) of supported symbols, I went with ['circle'](https://github.com/perses/perses/blob/v0.30.0/ui/components/src/utils/theme-gen.ts#L127) at the beginning instead of EChart's default `emptyCircle` since I ran benchmarks and `circle` is slightly faster

The big change here is now datapoints are visible when there aren't a ton of them (closer to how Grafana works now too)

image

ui/components/src/TimeSeriesTooltip/focused-series.test.ts
sjcobb: Yeah agreed! I'll probably wait on this though, Matt is going to be bringing in annotation rendering changes, which also adjusts this function (they render events as 'scatter' series so a condition is added to show different tooltip content). 

Avoiding further changing the params for now may make his life a bit easier since there are going to be a lot of merge conflicts already!

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
sjcobb: I ended up removing these since the defaults will work fine for now. 

I do want to try bringing back `blur` again later so we have more control over the opacity fading out for non-focused series (it's tricky to get this right with the custom 'downplay' actions and the current state is already a big improvement over what's on our main branch)

PR #1115, opened 2023-04-17
[ENHANCEMENT] Allow echarts wrapper to default to fluid width and height 

sjcobb:
Every usage of our echarts wrapper component `EChart.tsx` ends up setting an sx prop to control width and height resize behavior. This PR allows `width: '100%', height: '100%'` to be set by default. It is not a breaking change since combineSx is added and explicitly defining will not break anything, but those lines can be removed now if a user prefers to rely on defaults.

Building on PR #1112 to avoid merge conflicts and that PR will be merged first (legend highlight PR will be next in this PR stack).

# Checklist

## UI Changes

### DIFF
--- a/ui/components/src/EChart/EChart.tsx
+++ b/ui/components/src/EChart/EChart.tsx
@@ -188,6 +188,8 @@ export const EChart = React.memo(function EChart<T>({
       ref={containerRef}
       sx={combineSx(
         {
+          // Ensures chart fills container and updates size correctly when resizing,
+          // combineSx allows this default behavior to be overriden from parent component
           width: '100%',
           height: '100%',
         },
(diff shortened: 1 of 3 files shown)


ui/components/src/EChart/EChart.tsx
```diff
@@ -182,7 +183,20 @@ export const EChart = React.memo(function EChart<T>({
...
   }, [sx]);
 
-  return <Box ref={containerRef} sx={sx}></Box>;
+  return (
+    <Box
+      ref={containerRef}
+      sx={combineSx(
+        {
+          // Ensures chart fills container and updates size correctly when resizing,
+          // combineSx allows this default behavior to be overriden from parent component
```
cndonovan, line 192: I might say "this behavior can be overridden with combineSx on child components" (make it more direct)

cndonovan (approved):

ui/components/src/EChart/EChart.tsx
sjcobb: Hmm agreed this comment is a little confusing, but EChart.tsx actually is the child component and combineSx is allowing the parent component (LineChart, StatChart, GaugeChart) an easy way to customize it using sx as an escape hatch (in case the new default width and height aren't ideal for some edge case)

PR #1118, opened 2023-04-18
[IGNORE] add override flag that would reset the local DB

Nexucis:
you can now use it like that:

```bash
./scripts/api_backend_dev.sh --override
```

In case the folder `loca_db` doesn't exist when running this script whatever the flag is used, it will generate the DB.
But in case it already exists, then you need to use the flag `--override` to re-generate it


sjcobb (approved):
💯

PR #1123, opened 2023-04-19
[BREAKINGCHANGE] Make "TimeSeriesQuery" required in DefaultPluginKinds

eunicorn:
Because of the changes in #1032, if `TimeSeriesQuery` isn't defined in DefaultPluginKinds, it will have weird/unexpected behaviors in the query editor. 

**BEFORE** 
```
export interface PluginRegistryProps {
  children?: React.ReactNode;
  pluginLoader: PluginLoader;
  defaultPluginKinds?: DefaultPluginKinds; // optional before
}
```

```
export type DefaultPluginKinds = Partial<Record<PluginType, string>>; // all plugin types were optional
```

**AFTER**
```
export interface PluginRegistryProps {
  children?: React.ReactNode;
  pluginLoader: PluginLoader;
  defaultPluginKinds: DefaultPluginKinds; // this is now required
}
```

```
type PluginKinds = Partial<Record<PluginType, string>>;
export type DefaultPluginKinds = 
    Required<Pick<PluginKinds, 'TimeSeriesQuery'>> // "TimeSeriesQuery" is now required
    & Omit<PluginKinds, 'TimeSeriesQuery'>; 
```

### DIFF
--- a/ui/plugin-system/src/model/plugins.ts
+++ b/ui/plugin-system/src/model/plugins.ts
@@ -71,4 +71,5 @@ export type PluginImplementation<Type extends PluginType> = SupportedPlugins[Typ
 /**
  * Default plugin kinds by plugin type.
  */
-export type DefaultPluginKinds = Partial<Record<PluginType, string>>;
+type PluginKinds = Partial<Record<PluginType, string>>;
+export type DefaultPluginKinds = Required<Pick<PluginKinds, 'TimeSeriesQuery'>> & Omit<PluginKinds, 'TimeSeriesQuery'>;
--- a/ui/dashboards/src/stories/decorators/WithPluginRegistry.tsx
+++ b/ui/dashboards/src/stories/decorators/WithPluginRegistry.tsx
@@ -47,7 +47,10 @@ const bundledPluginLoader: PluginLoader = dynamicImportPluginLoader([
 
 export const WithPluginRegistry = (Story: StoryFn) => {
   return (
-    <PluginRegistry pluginLoader={bundledPluginLoader}>
+    <PluginRegistry
+      pluginLoader={bundledPluginLoader}
+      defaultPluginKinds={{ TimeSeriesQuery: 'PrometheusTimeSeriesQuery' }}
+    >
       <Story />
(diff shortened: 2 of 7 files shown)


ui/plugin-system/src/model/plugins.ts
```diff
@@ -71,4 +71,5 @@ export type PluginImplementation<Type extends PluginType> = SupportedPlugins[Typ
 /**
  * Default plugin kinds by plugin type.
  */
-export type DefaultPluginKinds = Partial<Record<PluginType, string>>;
+type PluginKinds = Partial<Record<PluginType, string>>;
+export type DefaultPluginKinds = Required<Pick<PluginKinds, 'TimeSeriesQuery'>> & Omit<PluginKinds, 'TimeSeriesQuery'>;
```
eunicorn, line 75: If anyone knows a more elegant way to write this, let me know 😄

juliepagano:
Should `defaultPluginKinds` also be required in the `PluginRegistryProps`? https://github.com/perses/perses/blob/main/ui/plugin-system/src/components/PluginRegistry/PluginRegistry.tsx#L30

ui/plugin-system/src/model/plugins.ts
juliepagano: I'm trying to remember why I even wrote it the way it was previously. Probably the result of some fighting with TypeScript. We only use the defaults for `TimeSeriesQuery` and `Panel`, right? If so, I wonder if this can be simplified to something like this. That'd be a lot simpler.

```
export type DefaultPluginKinds = {
  TimeSeriesQuery: string;
  Panel?: string;
};
```

eunicorn:
> Should `defaultPluginKinds` also be required in the `PluginRegistryProps`? 

Right! Also added it to `PluginRegistryContextType`

juliepagano (approved):
Could you update the description of the PR to clarify that `defaultPluginKinds` is also required now as part of the breaking change -- that will be helpful for any consumers trying to understand how to handle the change.

Other than that, LGTM! Thanks for fixing this up. 🎉

PR #1124, opened 2023-04-20
[UX Feedback] Categorical palette as default, switch to generative based on total series

sjcobb:
Mohit feedback:
> “Auto” palette is our end goal so users don’t have to meddle with color settings period.
> 1. For now, we can default to this for the color palette (https://colorbrewer2.org/#type=qualitative&scheme=Set3&n=12) with 12 colors for categorical. 
> 2. If we have more than 12 series we switch the palette to generative colors automatically, users should never have to deal with repetition of colors and don’t know what to do. This is my reasoning for leaving the colors to default to auto (ideally), one less thing to deal with. The option of categorical should no longer be available
> 3. Regarding tweaking the color palette for use cases like error - I think we can have multiple such generative palettes based on different use cases: Auto - General, Auto - Error, Auto - Warning, Custom Palette (user enters hex code?), etc.

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -150,6 +150,15 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     // Index is counted across multiple queries which ensures the categorical color palette does not reset for every query
     let seriesIndex = 0;
 
+    // Total series count across all queries is needed before mapping below to determine which color palette to use
+    // This calculation should not impact performance since total number of queries rarely exceeds ~5
+    let totalSeries = 0;
+    for (let i = 0; i < queryResults.length; i++) {
+      totalSeries += queryResults[i]?.data?.series?.length ?? 0;
+    }
+
+    // Mapping of each set of query results to be ECharts option compatible
+    // TODO: Look into performance optimizations and moving parts of mapping to the lower level chart
     for (const result of queryResults) {
       // Skip queries that are still loading or don't have data
       if (result.isLoading || result.isFetching || result.data === undefined) continue;
@@ -168,12 +177,14 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
           Array.isArray(echartsPalette) && echartsPalette[0]
             ? (echartsPalette[0] as string)
             : muiTheme.palette.primary.main;
-        // Check which color palette was chosen and get appropriate color.
+
+        // Decide which palette to use based on total series returned.
+        // When series number exceeds number of colors in palette, use generative colors instead.
+        const paletteLength = Array.isArray(echartsPalette) ? echartsPalette.length : 4;
         const seriesColor =
-          visual.palette?.kind === 'Categorical'
+          totalSeries < paletteLength
             ? getCategoricalPaletteColor(echartsPalette as string[], seriesIndex, fallbackColor)
             : getAutoPaletteColor(formattedSeriesName, fallbackColor);
-
         // Used for repeating colors in Categorical palette
         seriesIndex++;
 
--- a/ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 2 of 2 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -168,12 +177,14 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
           Array.isArray(echartsPalette) && echartsPalette[0]
             ? (echartsPalette[0] as string)
             : muiTheme.palette.primary.main;
-        // Check which color palette was chosen and get appropriate color.
+
+        // Decide which palette to use based on total series returned.
+        // When series number exceeds number of colors in palette, use generative colors instead.
+        const paletteLength = Array.isArray(echartsPalette) ? echartsPalette.length : 4;
```
sjcobb, line 183: One more fix coming soon! I'm planning to rename `echartsPalette` to `categoricalPalette` based on @cndonovan's [feedback here](https://github.com/perses/perses/pull/1088#discussion_r1175519897) on PR #1088. Otherwise, I think I've fixed the remaining comments on the previous PR and this one is officially ready for review now

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -165,15 +177,17 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
-          Array.isArray(echartsPalette) && echartsPalette[0]
-            ? (echartsPalette[0] as string)
+          Array.isArray(categoricalPalette) && categoricalPalette[0]
+            ? (categoricalPalette[0] as string) // needed since echarts color property isn't always an array
             : muiTheme.palette.primary.main;
-        // Check which color palette was chosen and get appropriate color.
+
+        // Decide which palette to use based on total series returned.
+        // When series number exceeds number of colors in palette, use generative colors instead.
+        const paletteLength = Array.isArray(categoricalPalette) ? categoricalPalette.length : 4;
```
eunicorn, line 186: is there a reason why we pick 4 here?

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -165,15 +177,17 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
         const seriesColor =
-          visual.palette?.kind === 'Categorical'
-            ? getCategoricalPaletteColor(echartsPalette as string[], seriesIndex, fallbackColor)
+          totalSeries <= paletteLength
```
eunicorn, line 188: What do you think about putting the color-related logic into a separate function/file (ex: `getSeriesColor(queryResults?.data?.series)`)? I think it'll make it easier to follow.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Changed to `0` instead so it will call getAutoPaletteColor in the event that the categorical [palette](https://github.com/perses/perses/blob/e3cb920002695309604d36d4ca62a233dd506f8e/ui/components/src/utils/theme-gen.ts#L43-L51) in the ECharts theme isn't set.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Good call, added `getSeriesColor` function and some additional tests [here](https://github.com/perses/perses/blob/5bff142bfe259cdc8679de4e1f1e61ab2ffcdcb4/ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.test.ts#L21)!

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -150,6 +153,15 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     // Index is counted across multiple queries which ensures the categorical color palette does not reset for every query
     let seriesIndex = 0;
 
+    // Total series count across all queries is needed before mapping below to determine which color palette to use
+    // This calculation should not impact performance since total number of queries rarely exceeds ~5
+    let totalSeries = 0;
+    for (let i = 0; i < queryResults.length; i++) {
+      totalSeries += queryResults[i]?.data?.series?.length ?? 0;
+    }
```
cndonovan, line 161: Awesome

PR #1125, opened 2023-04-21
[ENHANCEMENT] Use abbreviations for time units

cndonovan:
Right now, we display seconds as `100 seconds`, `200 seconds`, etc. along the y-axis.

We want to say `100s`, `200s`, etc. instead.


ui/components/src/model/units/time.ts
```diff
@@ -73,13 +73,16 @@ export enum TimeIntlDuration {
...
+  const decimalPlaces = decimal_places ?? DEFAULT_DECIMAL_PLACES;
+
+  const timeUnit: string = TimeIntlDuration[kind];
+  const isMonthOrYear = kind === 'Months' || kind === 'Years';
+
   const formatter = new Intl.NumberFormat('en-US', {
     style: 'unit',
     unit: timeUnit,
-    unitDisplay: 'long',
+    unitDisplay: isMonthOrYear ? 'long' : 'narrow',
```
cndonovan, line 85: See tests to see the difference that this makes!

I made the distinction with `months` and `years` because:

* If I only used `narrow`, 300 minutes would display as `300m` and 300 months would display as `300m`, so these two units would be ambiguous. 
* If I'm going to spell out `months`, might as well spell out the unit that's bigger than `months` too. 
* I think months and years will be pretty rare, so it's not such a big impact if we spell them out. I think milliseconds and seconds will be way more common.

ui/components/src/model/units/bytes.test.ts
```diff
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { formatValue } from './units';
-import { UnitTestCase } from './units.test';
+import { UnitTestCase } from './types';
```
cndonovan, line 15: `units.test.ts` had no more tests in it, so it was time to delete the whole file.

sjcobb: You had a good run units.test.ts, you’ll be missed 🫡

ui/components/src/model/units/time.ts
sjcobb: Amazing, cool to see you make full use of the Intl.NumberFormat api!

sjcobb (approved):

PR #1129, opened 2023-04-27
[FEATURE] Disable CRUD dialogs in readonly mode

Gladorme:
# Description
When you are readonly mode, CRUD dialogs shouldn't be usable. I was hesitating between this solution or disabling button opening dialogs. WDYT?

# Screenshots
image
image

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/CRUDButton/CRUDButton.tsx
+++ b/ui/app/src/components/CRUDButton/CRUDButton.tsx
@@ -31,20 +31,35 @@ export function CRUDButton(props: CRUDButtonProps) {
   const { text, variant, color, disabled, onClick } = props;
   const isReadonly = useIsReadonly();
 
+  if (isReadonly) {
+    return (
+      <Tooltip title="Resource managed via code only" placement="top">
+        <span>
+          <Button
+            variant={variant}
+            color={color}
+            size="small"
+            sx={{ textTransform: 'uppercase' }}
+            onClick={onClick}
+            disabled={disabled || isReadonly}
+          >
+            {text}
+          </Button>
+        </span>
+      </Tooltip>
+    );
+  }
+
   return (
-    <Tooltip title="Resource managed via code only" placement="top">
-      <span>
-        <Button
-          variant={variant}
-          color={color}
-          size="small"
-          sx={{ textTransform: 'uppercase' }}
-          onClick={onClick}
-          disabled={disabled || isReadonly}
-        >
-          {text}
-        </Button>
-      </span>
-    </Tooltip>
+    <Button
+      variant={variant}
+      color={color}
+      size="small"
+      sx={{ textTransform: 'uppercase' }}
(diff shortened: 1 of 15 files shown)


Nexucis:
perhaps that's a question for @foxbat07 

On my side I would prefer to avoid having to open a dialog to then understand I can't create / delete something.
So I would prefer disabling directly these buttons (`new project` for example). I will be faster and we can understand immediately the UI is in readonly mode.

sjcobb:
Yeah to Eunice’s point, having some type of sandbox flow where users can experiment and export out the JSON easily will be worth thinking about. 

If we definitely want to move forward with disabling for now, I do agree with Augustin that disabling the button so the dialog never opens makes sense, although I’m not sure how much extra work that ends up being

Nexucis:
mmm so we would have two different flows ? 

You won't be able to create a new project, but you would be able to access to the dashboard form. Isn't a bit weird from an UX point of view to have different flows for the same kind of button ?

Nexucis:
sounds good to me @foxbat07

Gladorme:
I still have a issue with tooltips not showing on some buttons... I need to check why

sjcobb (approved):
🙌

sjcobb:
Whoops missed the comment about the tooltips not showing, also didn't pull locally to test, but code LGTM otherwise!

Gladorme:
Fixed the PR, all tooltips are working fine 😄

ui/app/src/components/CRUDButton/CRUDButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Button, Tooltip } from '@mui/material';
+import { OverridableStringUnion } from '@mui/types';
+import { ButtonPropsColorOverrides, ButtonPropsVariantOverrides } from '@mui/material/Button/Button';
+import { useIsReadonly } from '../../model/config-client';
+
+interface CRUDButtonProps {
```
Nexucis, line 19: I'm wondering if this button should be moved to the package `component` instead. We probably want that also in the dashboard.
WDYT @eunicorn @sjcobb ?

Nexucis (approved):
LGTM otherwise. Thank you for doing it !

ui/app/src/components/CRUDButton/CRUDButton.tsx
Nexucis: Of course if we want to move the button to the package component, we should add more parameter like `isReadonly` instead of using directly the function `useIsReadonly`

and yes you are right it achieve the same purpose. This wrapper of the MUI button just add on top of that the tooltip that advertise why the button is disable.

PR #1131, opened 2023-04-28
[ENHANCEMENT] Make it possible to set decimal_places as undefined in UI

cndonovan:
This PR makes it possible to set `decimal_places` as `undefined` in the UI.

~To do this, I had to change the "Decimals" input from a dropdown to an input with `type="number"`.~

This allows the default behavior to kick in: When `decimal_places` is undefined, we can fall back to using significant figures to format numbers. 

## Before 
Screenshot 2023-04-28 at 9 21 27 AM

## After 
Screenshot 2023-05-02 at 12 09 30 PM

### DIFF
--- a/ui/components/src/UnitSelector/UnitSelector.tsx
+++ b/ui/components/src/UnitSelector/UnitSelector.tsx
@@ -11,15 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 import { Box, Switch, TextField, Autocomplete, SwitchProps } from '@mui/material';
-import {
-  UnitOptions,
-  UNIT_CONFIG,
-  UnitConfig,
-  isUnitWithDecimalPlaces,
-  isUnitWithAbbreviate,
-  DECIMAL_PLACES_MIN,
-  DECIMAL_PLACES_MAX,
-} from '../model';
+import { UnitOptions, UNIT_CONFIG, UnitConfig, isUnitWithDecimalPlaces, isUnitWithAbbreviate } from '../model';
 import { OptionsEditorControl } from '../OptionsEditorLayout';
 
 export interface UnitSelectorProps {
@@ -38,6 +30,19 @@ const KIND_OPTIONS: AutocompleteKindOption[] = Object.entries(UNIT_CONFIG)
   })
   .filter((config) => !config.disableSelectorOption);
 
+const DECIMAL_PLACES_OPTIONS = [
+  { label: 'Default', decimal_places: undefined },
+  { label: '0', decimal_places: 0 },
+  { label: '1', decimal_places: 1 },
+  { label: '2', decimal_places: 2 },
+  { label: '3', decimal_places: 3 },
+  { label: '4', decimal_places: 4 },
+];
+
+function getOptionByDecimalPlaces(decimal_places?: number) {
+  return DECIMAL_PLACES_OPTIONS.find((o) => o.decimal_places === decimal_places);
+}
+
 export function UnitSelector({ value, onChange }: UnitSelectorProps) {
   const hasDecimalPlaces = isUnitWithDecimalPlaces(value);
   const hasAbbreviate = isUnitWithAbbreviate(value);
@@ -48,11 +53,11 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
     });
   };
 
-  const handleDecimalChange = (newValue?: number) => {
+  const handleDecimalPlacesChange = (_: unknown, { decimal_places }: { decimal_places: number | undefined }) => {
     if (hasDecimalPlaces) {
       onChange({
         ...value,
-        decimal_places: newValue,
+        decimal_places: decimal_places,
       });
     }
   };
@@ -106,14 +111,15 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
       <OptionsEditorControl
         label="Decimals"
         control={
-          <TextField
-            type="number"
-            value={value.decimal_places ?? ''}
-            onChange={(e) => {
-              handleDecimalChange(e.target.value ? Number(e.target.value) : undefined);
-            }}
-            placeholder="Default"
-            inputProps={{ min: DECIMAL_PLACES_MIN, max: DECIMAL_PLACES_MAX }}
+          <Autocomplete
+            value={getOptionByDecimalPlaces(value.decimal_places)}
+            options={DECIMAL_PLACES_OPTIONS}
+            getOptionLabel={(o) => o.label}
+            isOptionEqualToValue={(option, value) => option.label === value.label}
+            renderInput={(params) => <TextField {...params} />}
+            onChange={handleDecimalPlacesChange}
+            disabled={!hasDecimalPlaces}
+            disableClearable
           />
         }
       />
--- a/ui/components/src/UnitSelector/UnitSelector.tsx
+++ b/ui/components/src/UnitSelector/UnitSelector.tsx
@@ -14,10 +14,11 @@ import { Box, Switch, TextField, Autocomplete, SwitchProps } from '@mui/material
 import {
   UnitOptions,
   UNIT_CONFIG,
-  DEFAULT_DECIMAL_PLACES,
   UnitConfig,
   isUnitWithDecimalPlaces,
   isUnitWithAbbreviate,
+  DECIMAL_PLACES_MIN,
+  DECIMAL_PLACES_MAX,
 } from '../model';
 import { OptionsEditorControl } from '../OptionsEditorLayout';
 
@@ -37,8 +38,6 @@ const KIND_OPTIONS: AutocompleteKindOption[] = Object.entries(UNIT_CONFIG)
   })
   .filter((config) => !config.disableSelectorOption);
 
-const DECIMAL_OPTIONS = [0, 1, 2, 3, 4];
-
 export function UnitSelector({ value, onChange }: UnitSelectorProps) {
   const hasDecimalPlaces = isUnitWithDecimalPlaces(value);
   const hasAbbreviate = isUnitWithAbbreviate(value);
@@ -49,7 +48,7 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
     });
   };
 
-  const handleDecimalChange = (_: unknown, newValue: number) => {
+  const handleDecimalChange = (newValue?: number) => {
     if (hasDecimalPlaces) {
       onChange({
         ...value,
@@ -82,7 +81,7 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
         }
       />
       <OptionsEditorControl
-        label="Units"
+        label="Unit"
         control={
           <Autocomplete
             value={{ id: value.kind, ...kindConfig }}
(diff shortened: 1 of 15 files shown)


ui/components/src/UnitSelector/UnitSelector.tsx
```diff
@@ -82,7 +81,7 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
         }
       />
       <OptionsEditorControl
-        label="Units"
+        label="Unit"
```
cndonovan, line 84: Changed label from `Units` --> `Unit`

ui/components/src/UnitSelector/UnitSelector.tsx
```diff
@@ -105,17 +104,17 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
         }
       />
       <OptionsEditorControl
-        label="Decimal"
+        label="Decimals"
```
cndonovan, line 107: Changed label for `decimal_places` from `Decimal` --> `Decimals`

ui/components/src/model/units/bytes.ts
```diff
@@ -37,37 +38,52 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
 
-export function formatBytes(bytes: number, { abbreviate, decimal_places }: BytesUnitOptions) {
-  if (bytes === 0) return '0 bytes';
+export function formatBytes(bytes: number, options: BytesUnitOptions) {
+  const { abbreviate, decimal_places } = options;
 
-  decimal_places = decimal_places ?? DEFAULT_DECIMAL_PLACES;
-  // Avoids maximumFractionDigits value is out of range error. Possible values are 0 to 20.
-  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#minimumfractiondigits
-  if (decimal_places < 0) {
```
cndonovan, line 46: Moved this code to a function called `limitDecimalPlaces()`, and used it in all the formatting files.

ui/components/src/model/units/constants.ts
```diff
@@ -13,4 +13,7 @@
 
 // Common constants needed across individual unit groups and the overall
 // combined units.
-export const DEFAULT_DECIMAL_PLACES = 2;
+export const MAX_SIGNIFICANT_DIGITS = 3;
+
+export const DECIMAL_PLACES_MIN = 0;
+export const DECIMAL_PLACES_MAX = 20;
```
cndonovan, line 19: We don't need `DEFAULT_DECIMAL_PLACES` anymore

We do need a constant to represent the default # of significant figures to use

The other two constants are for the `decimal_places` input with `type="number"`

ui/components/src/model/units/percent.ts
```diff
@@ -43,32 +44,24 @@ export const PERCENT_UNIT_CONFIG: Readonly<Record<PercentUnitKind, UnitConfig>>
...
-    });
+    formatterOptions.maximumFractionDigits = limitDecimalPlaces(decimal_places);
   } else {
-    // By default, use maximumSignificantDigits
-    formatter = new Intl.NumberFormat('en-US', {
-      style: 'percent',
-      maximumSignificantDigits: MAX_SIGNIFICANT_DIGITS,
-      useGrouping: true,
-    });
+    formatterOptions.maximumSignificantDigits = MAX_SIGNIFICANT_DIGITS;
```
cndonovan, line 62: I've played a lot of code golf trying to make the formatting code readable. I think this is pretty nice 🤷‍♀️ Maybe tomorrow I'll change my mind :) 

But anyway, on the last PR, people weren't sure what the difference was between the formatting when `hasDecimalPlaces` is `true` and `hasDecimalPlaces` is `false`, so I thought this would make it super clear.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -53,7 +53,6 @@ export type VisualOptions = {
 
 export const DEFAULT_UNIT: UnitOptions = {
   kind: 'Decimal',
-  decimal_places: 2,
```
cndonovan, line 56: Remove and default to `decimal_places` is `undefined`

sjcobb (approved):
LGTM!

sjcobb:
One weird thing I just noticed is the full 'Default' text gets cutoff in certain cases: 

image

ui/components/src/UnitSelector/UnitSelector.tsx
sjcobb: Should the label actually be 'Max Decimals'?

cndonovan:
> One weird thing I just noticed is the full 'Default' text gets cutoff in certain cases:
> 
> image

Whoa, no bueno, I had not seen that. I'll take a look, thank you

cndonovan:
@sjcobb I can't repro, what did you do to see that?

juliepagano:
> @sjcobb I can't repro, what did you do to see that?

I'm seeing the same thing with your latest changes. Didn't do anything special to trigger it. It looks that way when adding a new panel or when editing an existing one.

juliepagano:
I find the UX on this a little weird. It's not obvious that you should delete the number to get default behavior. I wonder if it'd make sense to keep it a dropdown with "default" as an option. Curious if @foxbat07 has thoughts.

juliepagano:
I think this is maybe a separate UX issue, but thought worth mentioning here since I think you proposed adjusting it. I find the UX of only showing decimals if they're non-zero a little weird when *some* of the items do have decimals. It makes it a little harder to visually skim the values because they're not aligned. 

image

ui/components/src/UnitSelector/UnitSelector.tsx
```diff
@@ -105,17 +104,17 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
...
-            disabled={!hasDecimalPlaces}
-            disableClearable
-          ></Autocomplete>
+          <TextField
+            type="number"
+            value={value.decimal_places ?? ''}
+            onChange={(e) => {
+              handleDecimalChange(e.target.value ? Number(e.target.value) : undefined);
+            }}
+            placeholder="Default"
```
juliepagano, line 115: I like @foxbat07's idea for calling this "auto" instead of default, especially because it's slightly different behavior (significant digits instead of decimals).

ui/components/src/model/units/bytes.ts
```diff
@@ -37,37 +38,52 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
-    decimal_places = 0;
-  } else if (decimal_places > 20) {
-    decimal_places = 20;
+  const showFullNumber = abbreviate === false;
```
juliepagano, line 44: Optional nit: Is `abbreviate` always a boolean value? If so, `!abbreviate` feels slightly cleaner here.

ui/components/src/model/units/bytes.ts
```diff
@@ -37,37 +38,52 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
+    const formatterOptions: Intl.NumberFormatOptions = {
       style: 'unit',
       unit: 'byte',
       unitDisplay: 'long',
-      maximumFractionDigits: decimal_places,
       useGrouping: true,
-    });
+    };
+
+    const hasDecimalPlaces = decimal_places !== undefined && decimal_places !== null;
```
juliepagano, line 57: Nit: would `typeof decimal_places === 'number'` be slightly safer here around weird edge cases?

ui/components/src/model/units/bytes.ts
```diff
@@ -37,37 +38,52 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
     return formatter.format(bytes);
   }
 
+  // numbro is able to add units like KB, MB, GB, etc. correctly
   return numbro(bytes).format({
     output: 'byte',
     base: 'decimal',
     spaceSeparated: true,
-    mantissa: decimal_places,
+    mantissa: decimal_places ?? 2,
```
juliepagano, line 73: Should this hardcoded fallback be pulled out into a constant like the other values?

ui/components/src/model/units/bytes.ts
```diff
@@ -37,37 +38,52 @@ export const BYTES_UNIT_CONFIG: Readonly<Record<BytesUnitKind, UnitConfig>> = {
...
     trimMantissa: true,
     optionalMantissa: true,
   });
 }
+
+function formatBytesAsFullNumber(bytes: number, { decimal_places }: BytesUnitOptions) {
+  const hasDecimalPlaces = decimal_places !== undefined && decimal_places !== null;
```
juliepagano, line 80: Same nit as above.

ui/components/src/model/units/constants.ts
juliepagano: Per our conversation with Mohit this morning, I think bringing this number down a bit would be a good idea. We can always increase it later if we get requests.

ui/components/src/model/units/time.ts
```diff
@@ -74,16 +75,21 @@ export enum TimeIntlDuration {
...
-    maximumFractionDigits: decimalPlaces,
-  });
+  };
+
+  const hasDecimalPlaces = decimal_places !== undefined && decimal_places !== null;
+  if (hasDecimalPlaces) {
+    formatterOptions.maximumFractionDigits = limitDecimalPlaces(decimal_places);
+  } else {
+    formatterOptions.maximumSignificantDigits = MAX_SIGNIFICANT_DIGITS;
+  }
```
juliepagano, line 91: It looks like you do this little bit of code (or something very similar) ~3 times in here. Is this a case where it's slightly different bits of code that I pattern matched too quickly or are these identical. If the latter, maybe worth pulling out into a little formatting util, so if we adjust these again it can happen in one place.

juliepagano (commented):
A bunch of small nits here. I also left some comments about the UI/UX in direct comments.

cndonovan:
> I find the UX on this a little weird. It's not obvious that you should delete the number to get default behavior. I wonder if it'd make sense to keep it a dropdown with "default" as an option. Curious if @foxbat07 has thoughts.

Oh yeah, that might be better. I think I went down that path and it hit some weirdness, but I'm down to try again. 

I'm sync-ing with Mohit tomorrow too to show him all of this.

cndonovan:
> I think this is maybe a separate UX issue, but thought worth mentioning here since I think you proposed adjusting it. I find the UX of only showing decimals if they're non-zero a little weird when _some_ of the items do have decimals. It makes it a little harder to visually skim the values because they're not aligned.
> 
> image
> 
> Similar example in tooltip, but they all have decimals, but they're uneven.
> 
> image

Ah yeah. I think that's something we could revisit after this change lands. 

This is also related to another bug on my plate where we will pick a better value for the y-axis minimum when we're zooming -- it doesn't need to be weird and have a bunch of decimal places. It can look more like the other y-axis values.

ui/components/src/UnitSelector/UnitSelector.tsx
cndonovan: Yeah I like "Auto" too! (Also what Grafana uses). 

I'll save that for a different PR since it would require updating all the inputs (because they all say "Default" right now).

ui/components/src/model/units/bytes.ts
cndonovan: Ah, I wanted to capture that it's explicitly set to `false` -- in other words, `undefined`  or `null` does not mean we should show the full number. 

It's because "abbreviating" is actually the common behavior, and not abbreviating is the uncommon behavior, so it's making sure the user explicitly asks for it. 

In an ideal world, I think it would have been cool if the parameter was `showFullNumber` (i.e. do not abbreviate) or something like that because then we could assume false/undefined to mean the same thing.

ui/components/src/model/units/bytes.ts
cndonovan: Oo maybe. It's simpler too!

ui/components/src/model/units/bytes.ts
cndonovan: Ah yeah, why not! It only applies to byte formatting because we have to use `numbro` and `numbro` has a different API, but I can still make a constant with a nice name.

ui/components/src/model/units/time.ts
cndonovan: Oo yes I will give that try

ui/components/src/UnitSelector/UnitSelector.tsx
cndonovan: I'll check with Mohit

cndonovan: Sounds like no -- 

After this change goes in, we can go back to: if the user sets decimal_places, then we always show that exact number of decimal places. 

(Before, that was causing problems because we didn't have the default option)

juliepagano (approved):
LGTM! 🎉

sjcobb (approved):
🚢 🚢

PR #1132, opened 2023-04-28
[BUGFIX] Series name formatter return empty string when no resolved label value

sjcobb:
## Overview

When a label_name token inside the double curly braces (aka handlebars syntax) for our `series_name_format` field in the query editor tab has no corresponding label values or is unintentionally set to value that isn't reflected in the query results, our formatter function used to show the token itself. Now instead, an empty string is returned, which makes more sense and is a better fallback for new users who might not 100% understand how this somewhat confusing field actually works under the hood (hopefully in the future, this will only be exposed to power users and we come up with a higher level, more intuitive way to format series names automatically and tie it to the legend more visibly).

## Before
image

## After

image

image

# Checklist

## UI Changes

### DIFF
--- a/ui/prometheus-plugin/src/utils/utils.ts
+++ b/ui/prometheus-plugin/src/utils/utils.ts
@@ -68,12 +68,18 @@ export const parseTemplateVariables = (text: string) => {
 export type SeriesLabels = Record<string, string>;
 
 /*
- * Formatter used for series name display in legends and tooltips
- * Regex replaces label {{ name }} with resolved label value
+ * Formatter used for series name display in legends and tooltips.
+ * Regex replaces label {{ name }} with resolved label value.
+ * If no resolved value, return empty string instead of the token inside double curly braces.
  */
 export function formatSeriesName(inputFormat: string, seriesLabels: SeriesLabels): string {
   const resolveLabelsRegex = /\{\{\s*(.+?)\s*\}\}/g;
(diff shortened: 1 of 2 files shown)


ui/prometheus-plugin/src/utils/utils.ts
```diff
@@ -68,12 +68,16 @@ export const parseTemplateVariables = (text: string) => {
...
+ * Formatter used for series name display in legends and tooltips.
+ * Regex replaces label {{ name }} with resolved label value.
+ * If no resolved value, return empty string instead of the token inside double curly braces.
  */
 export function formatSeriesName(inputFormat: string, seriesLabels: SeriesLabels): string {
   const resolveLabelsRegex = /\{\{\s*(.+?)\s*\}\}/g;
-  return inputFormat.replace(resolveLabelsRegex, (_, g) => (seriesLabels[g] ? seriesLabels[g] : g));
+  return inputFormat.replace(resolveLabelsRegex, (_match, token) => {
+    const resolvedValue = seriesLabels[token] ?? '';
+    return resolvedValue.toString();
```
eunicorn, line 79: i think you can remove `toString` here since seriesLabels[token] returns a string, right?

eunicorn (approved):

ui/prometheus-plugin/src/utils/utils.ts
sjcobb: I added it since I made a unit test where what was passed in wasn’t  valid but yeah I guess it wouldn’t ever actually happen, especially since this is only used in the Prometheus plugin currently

PR #1142, opened 2023-05-02
[BUGFIX] Prevent calendar icon from flipping over

cndonovan:
We noticed that the `Calendar` icon will flip over when the time range menu is opened. This happens because the icon is typically an arrow/caret that is supposed to flip over. This PR prevents that behavior.

### DIFF
--- a/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
+++ b/ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
@@ -42,7 +42,10 @@ export function TimeRangeSelector(props: TimeRangeSelectorProps) {
     <Select
       value={formattedValue}
       onChange={onSelectChange}
-      IconComponent={Calendar}
+      IconComponent={(props) => {
+        // "transform: none" prevents calendar icon from flipping over when menu is open
+        return <Calendar {...props} sx={{ transform: 'none !important' }} />;
+      }}
       inputProps={{
         'aria-label': `Select time range. Currently set to ${formattedValue}`,
       }}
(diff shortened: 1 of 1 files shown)


ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
```diff
@@ -42,15 +42,21 @@ export function TimeRangeSelector(props: TimeRangeSelectorProps) {
...
         'aria-label': `Select time range. Currently set to ${formattedValue}`,
       }}
       sx={{
         '.MuiSelect-icon': {
           marginTop: '1px',
         },
-        '.MuiSelect-select': height ? { lineHeight: height, paddingY: 0 } : {},
+        // "paddingRight" creates more space for the calendar icon (it's a bigger icon)
+        '.MuiSelect-select': height
+          ? { lineHeight: height, paddingY: 0, paddingRight: '36px !important' }
```
sjcobb, line 58: Any way to get more specific with the `.MuiSelect-select` class so !important isn't needed or I'm assuming MUI just doesn't give a helpful class here?

sjcobb (approved):

ui/components/src/DateTimeRangePicker/TimeRangeSelector.tsx
cndonovan: I'll give it a shot, let's see!

cndonovan: `!important` is not the best things to use if it can be avoided :)

cndonovan: @sjcobb That works much better, thank you

PR #1144, opened 2023-05-03
[FEATURE] Support ARM arch for binaries and docker

### LINKED ISSUE #957: Support ARM Chip arcitecture in docker image
Hi, looks like a great project! 👏 

I tried to run the docker on my m1 apple chip.
```
Unable to find image 'persesdev/perses:latest' locally
latest: Pulling from persesdev/perses
fc251a6e7981: Pull complete
99acb004d957: Pull complete
3c747f87a046: Pull complete
da755c413810: Pull complete
828cb559575c: Pull complete
b2ec7374fe6b: Pull complete
b7d2660d815c: Pull complete
fdbbd84be376: Pull complete
Digest: sha256:f160ee15b11f4855259bc7b230be117b113e251fb427d3d24c8ac0b41e34dadb
Status: Downloaded newer image for persesdev/perses:latest
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
1ac6454dbcb3ee04fe3f5636210b580aa8bc8055712b984a0a1d30009b2d9a0d
```

supporting multi-platform docker is possible via
https://docs.docker.com/build/building/multi-platform/
🙏

Nexucis:
It appears it would be great to be able to support the ARM architecture for binaries and for Docker too.

Managing different platform for docker with goreleaser is too verbose as we need to duplicate every flag across the different config to be consistent. It was ok to do it for two config, but adding the multiplatform support will make it impossible to maintain.

My proposition is to generate the file `.goreleaser.yaml` instead using a go script. Like that we are sure the flags are consistent across the different config.

This should fix #957


scripts/generate-goreleaser/generate-goreleaser.go
```diff
@@ -0,0 +1,143 @@
...
+				Goarch:     arch,
+				IDs:        binaryIDs,
+				Dockerfile: dockerfileName,
+				Use:        "buildx",
+				ImageTemplates: []string{
+					fmt.Sprintf("%s:{{ .Tag }}-%s-%s", dockerBaseImageTemplateName, templateNames[1], arch),
+				},
+				BuildFlagTemplates: []string{
+					"--pull",
+					"--label=org.opencontainers.image.authors=The Perses Authors <perses-team@googlegroups.com>",
```
juliepagano, line 84: I completely forgot we had a google group (it is in the governance docs, but I haven't looked at those in aaaages). Is that something I should get added to as a maintainer?

juliepagano (approved):
Code changes generally look ok to me, but my golang is pretty weak. If you want a closer look, it might be worth tagging one of the more backend-focused devs.

scripts/generate-goreleaser/generate-goreleaser.go
Nexucis: yup you should be added ;)

PR #1138, opened 2023-05-01
[BUGFIX] Tooltip locale errors can cause chart to break on hover

sjcobb:
This PR fixes a Firefox locale timezone issue where our graph can break completely on hover. 

Settings example:
image

## Before
image

On hover over 'Queue Length' line chart:
image

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [x] Visual tests are stable and unlikely to be flaky. See [Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests) and [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

PR #1149, opened 2023-05-04
[BREAKINGCHANGE] Allow saving `duration` from JSON editor, useDefaultTimeRange renamed to useDashboardDuration

sjcobb:
Now that a JSON editor was added in #1012, this PR allows saving a new `duration` from the editor. 

This is just the first PR of multiple that will allow users to pick if they want to save the selected time range as default and/or the selected variables as default.

## Screen recording

https://user-images.githubusercontent.com/9369625/236255218-cfdd270f-4a0a-4ff2-beac-55b055abd169.mp4

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
+++ b/ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
@@ -17,7 +17,7 @@ import { devtools } from 'zustand/middleware';
 import { immer } from 'zustand/middleware/immer';
 import { shallow } from 'zustand/shallow';
 import { createContext, ReactNode, useCallback, useContext, useEffect, useState } from 'react';
-import { DashboardResource, Display, ProjectMetadata, RelativeTimeRange, DurationString } from '@perses-dev/core';
+import { DashboardResource, Display, ProjectMetadata, RelativeTimeRange } from '@perses-dev/core';
 import { usePlugin, usePluginRegistry } from '@perses-dev/plugin-system';
 import { createPanelGroupEditorSlice, PanelGroupEditorSlice } from './panel-group-editor-slice';
 import { convertLayoutsToPanelGroups, createPanelGroupSlice, PanelGroupSlice } from './panel-group-slice';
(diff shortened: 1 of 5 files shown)


Nexucis (approved):
🥳

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
```diff
@@ -133,17 +133,18 @@ function initStore(props: DashboardProviderProps) {
           ...createEditJsonDialogSlice(...args),
           metadata,
           display,
-          defaultTimeRange: { pastDuration: duration },
+          timeRange: { pastDuration: duration },
```
juliepagano, line 136: Any reason for the renaming here? Guessing it's for clarity, but wanted to double-check I wasn't missing something else, especially since you left some of the hooks using "default time range" naming.

juliepagano (commented):
One small question inline. 

Other question: do you think it's worth adding a test for this? Maybe a small e2e test to validate the workflow of editing the JSON, seeing the time picker changed to match, and then reloading the url (not refreshing with the query params still in there) to validate the new default was saved? It sounds like this is the first of several PRs related to this behavior, so let me know if you'd rather do that in a follow-up.

cndonovan (approved):
🎉

sjcobb:
> Other question: do you think it's worth adding a test for this? 

Yeah I'm definitely planning to add an e2e, I've created a few tests with Playwright in previous PRs but I'm not super fast with it yet. Since this is going to be followed quickly by a few related features (saving defaults when currently selected change), I'd like to create the tests all at once.

Also, since the JSON edit flow isn't tested yet, I have some questions about sharing constants. For instance, I'll need to update the code icon button so it can be clicked and am planning to import [this 'Edit JSON' constant](https://github.com/perses/perses/blob/v0.29.1/ui/dashboards/src/constants/user-interface-text.ts#L19) but since it lives in @perses-dev/dashboards, how can I use it in @perses-dev/e2e? Should I move it to @perses-dev/core, but I noticed that e2e doesn't import from core yet, is this okay to add as a dependency?

ui/dashboards/src/context/DashboardProvider/DashboardProvider.tsx
sjcobb: Looking at it again, I actually think we should just rename this `timeRange` (prev defaultTimeRange) to `duration` since that is how it's defined in our dashboard schema and is what is stored. It also makes some of the logic in places like TimeRangeControls simpler. So now `useDefaultTimeRange` becomes `useDashboardDuration` and setDashboard calls `state.duration = duration;`

The full active timeRange from plugin-system (relative and absolute) is more relevant for runtime uses cases, whereas saving a full absolute time range doesn't make sense when storing a dashboard (confirmed this with Mohit too)

sjcobb:
Going ahead and merging since I want to include this in a snapshot release I'm about to do that will test out a few of my open PRs like #1138 and tooltip highlighting / echarts-nightly branch.

PR #1152, opened 2023-05-05
[ENHANCEMENT] Update formatting to use exact decimal places

cndonovan:
This is a follow-up to a bunch of number formatting work.

We noticed we had a lot of numbers with unnecessary decimal places like 100.00, 400.00, etc.

We changed "Decimals" and made it so that when "Decimals" is set to X, use _at maximum_ X decimal places. That solved the problem but wasn't quite complete.

Now, we've added the ability to set "Decimals" to undefined. Using this, we can avoid having numbers like 100.00. This means we can go back to being explicit: When "Decimals" is set to X, use exactly X decimal places.

### DIFF
--- a/ui/components/src/UnitSelector/UnitSelector.tsx
+++ b/ui/components/src/UnitSelector/UnitSelector.tsx
@@ -12,6 +12,7 @@
 // limitations under the License.
 import { Box, Switch, TextField, Autocomplete, SwitchProps } from '@mui/material';
 import { UnitOptions, UNIT_CONFIG, UnitConfig, isUnitWithDecimalPlaces, isUnitWithAbbreviate } from '../model';
+import { shouldAbbreviate } from '../model/units/utils';
 import { OptionsEditorControl } from '../OptionsEditorLayout';
 
 export interface UnitSelectorProps {
@@ -79,7 +80,7 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
         label="Abbreviate"
         control={
           <Switch
-            checked={hasAbbreviate ? !!value.abbreviate : false}
+            checked={hasAbbreviate ? shouldAbbreviate(value.abbreviate) : false}
             onChange={handleAbbreviateChange}
             disabled={!hasAbbreviate}
           />
--- a/ui/components/src/model/units/bytes.test.ts
+++ b/ui/components/src/model/units/bytes.test.ts
@@ -28,7 +28,7 @@ const BYTES_TESTS: UnitTestCase[] = [
   {
     value: -1234,
     unit: { kind: 'Bytes', abbreviate: false, decimal_places: 4 },
-    expected: '-1,234 bytes',
+    expected: '-1,234.0000 bytes',
   },
   {
     value: -1234,
@@ -38,7 +38,7 @@ const BYTES_TESTS: UnitTestCase[] = [
   {
     value: -1234,
     unit: { kind: 'Bytes', abbreviate: true, decimal_places: 4 },
-    expected: '-1.234 KB',
+    expected: '-1.2340 KB',
   },
   { value: 0, unit: { kind: 'Bytes' }, expected: '0 bytes' },
   { value: 1, unit: { kind: 'Bytes' }, expected: '1 byte' },
@@ -55,7 +55,7 @@ const BYTES_TESTS: UnitTestCase[] = [
   {
     value: 10,
     unit: { kind: 'Bytes', abbreviate: false, decimal_places: 4 },
-    expected: '10 bytes',
+    expected: '10.0000 bytes',
   },
   {
     value: 10,
@@ -65,7 +65,7 @@ const BYTES_TESTS: UnitTestCase[] = [
   {
     value: 10,
     unit: { kind: 'Bytes', abbreviate: true, decimal_places: 4 },
-    expected: '10 bytes',
+    expected: '10.0000 bytes',
   },
   {
     value: 10.1234,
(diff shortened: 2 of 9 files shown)


ui/components/src/model/units/bytes.test.ts
```diff
@@ -28,7 +28,7 @@ const BYTES_TESTS: UnitTestCase[] = [
   {
     value: -1234,
     unit: { kind: 'Bytes', abbreviate: false, decimal_places: 4 },
-    expected: '-1,234 bytes',
+    expected: '-1,234.0000 bytes',
```
cndonovan, line 31: All these changes show "decimal_places" becoming a setting that means "use exactly X decimal places".

ui/components/src/model/units/bytes.ts
```diff
@@ -52,6 +52,7 @@ export function formatBytes(bytes: number, options: BytesUnitOptions) {
     };
 
     if (hasDecimalPlaces(decimal_places)) {
+      formatterOptions.minimumFractionDigits = limitDecimalPlaces(decimal_places);
```
cndonovan, line 55: Creates formatting that uses exact number of decimal places

ui/components/src/model/units/bytes.ts
```diff
@@ -60,17 +61,16 @@ export function formatBytes(bytes: number, options: BytesUnitOptions) {
       }
     }
 
-    const formatter = Intl.NumberFormat('en-US', formatterOptions);
-    return formatter.format(bytes);
+    return Intl.NumberFormat('en-US', formatterOptions).format(bytes);
```
cndonovan, line 64: Figured why not write this in one line instead of two lines

ui/components/src/model/units/bytes.ts
```diff
@@ -60,17 +61,16 @@ export function formatBytes(bytes: number, options: BytesUnitOptions) {
...
   return numbro(bytes).format({
     output: 'byte',
     base: 'decimal',
     spaceSeparated: true,
-    mantissa: decimal_places ?? DEFAULT_NUMBRO_MANTISSA,
-    trimMantissa: true,
-    optionalMantissa: true,
+    mantissa: hasDecimalPlaces(decimal_places) ? decimal_places : DEFAULT_NUMBRO_MANTISSA,
+    trimMantissa: !hasDecimalPlaces(decimal_places),
+    optionalMantissa: !hasDecimalPlaces(decimal_places),
```
cndonovan, line 74: `trimMantissa` trims trailing 0s
`optionalMantissa` excludes the decimals places if they're all 0s

We don't want this behavior if `decimal_places` is set

ui/components/src/model/units/decimal.test.ts
```diff
@@ -16,18 +16,57 @@ import { UnitTestCase } from './types';
...
+  },
+  {
+    value: -4444,
+    unit: { kind: 'Decimal', abbreviate: true },
+    expected: '-4.44K',
+  },
+  {
+    value: -4444,
+    unit: { kind: 'Decimal', abbreviate: true, decimal_places: 4 },
+    expected: '-4.4440K',
```
cndonovan, line 41: Added more tests for negative values

ui/components/src/model/units/decimal.test.ts
```diff
@@ -60,27 +99,22 @@ const DECIMAL_TESTS: UnitTestCase[] = [
...
     value: 10.123456,
-    unit: { kind: 'Decimal', decimal_places: 3 },
-    expected: '10.123',
+    unit: { kind: 'Decimal', abbreviate: true },
+    expected: '10.1',
   },
   {
     value: 10.123456,
-    unit: { kind: 'Decimal', decimal_places: 4 },
+    unit: { kind: 'Decimal', abbreviate: true, decimal_places: 4 },
```
cndonovan, line 117: Changed the structure of this set of tests to match the other tests

ui/components/src/UnitSelector/UnitSelector.tsx
```diff
@@ -79,7 +80,7 @@ export function UnitSelector({ value, onChange }: UnitSelectorProps) {
         label="Abbreviate"
         control={
           <Switch
-            checked={hasAbbreviate ? !!value.abbreviate : false}
+            checked={hasAbbreviate ? shouldAbbreviate(value.abbreviate) : false}
```
cndonovan, line 83: Bug fix while I was in the code

ui/components/src/model/units/bytes.ts
juliepagano: Might be worth putting this in a comment above these for future us.

juliepagano (approved):
One optional comment. Otherwise, LGTM. I love all the tests you've added for these utils! It makes it so much easier to follow the impact of a change and avoid introducing regressions. 🎉

sjcobb (approved):

eunicorn (approved):

ui/components/src/model/units/bytes.ts
cndonovan: Yess

cndonovan:
> Does anything need to be updated in `unit.cue`? Probably not since decimal_places should already be optional everywhere

I shall double-check! Keep those files up-to-date

cndonovan:
> > Does anything need to be updated in `unit.cue`? Probably not since decimal_places should already be optional everywhere
> 
> I shall double-check! Keep those files up-to-date

Yeah it looks good, `unit.cue` marks all the decimal_places fields as optional

PR #1153, opened 2023-05-05
[IGNORE] update ui dependencies

Nexucis:


Nexucis:
looks like upgrading the react-router is big pain :(.

sjcobb:
And yeah updating react-router seems like it's always a pain! 😞

PR #1154, opened 2023-05-05
[ENHANCEMENT] Improve centering for stat chart

cndonovan:
While playing around with the Stat Chart, I noticed there was extra padding around the text when the chart doesn't have a sparkline. 

I figure this helps center the text a bit more.


sjcobb (approved):

eunicorn (approved):
thanks!

PR #1158, opened 2023-05-10
[BREAKINGCHANGE] Add interfaces for each panel's definition

cndonovan:
This PR adds an `interface` that represents the schema for each type of panel (`TimeSeriesChart`, `GaugeChart`, `StatChart`, and `MarkdownPanel`).

The goal is to use these in the converter code.

## Breaking Changes
Removes `spec.unit` from the TimeSeriesDefinition, because we actually use `spec.y_axis.unit`.  

To fix a broken dashboard, either move the data in `spec.unit` to `spec.y_axis.unit` OR remove `spec.unit` if `spec.y_axis.unit` is already defined.

### DIFF
--- a/ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
+++ b/ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
@@ -19,6 +19,9 @@ export const DEFAULT_UNIT: UnitOptions = { kind: 'PercentDecimal' };
 export const DEFAULT_MAX_PERCENT = 100;
 export const DEFAULT_MAX_PERCENT_DECIMAL = 1;
 
+/**
+ * The schema for a GaugeChart panel.
+ */
 export interface GaugeChartDefinition extends Definition<GaugeChartOptions> {
   kind: 'GaugeChart';
 }
--- a/ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
+++ b/ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
@@ -16,12 +16,13 @@ import { UnitOptions } from '@perses-dev/components';
 import { CalculationType, OptionsEditorProps } from '@perses-dev/plugin-system';
 
 export const DEFAULT_UNIT: UnitOptions = { kind: 'PercentDecimal' };
-
 export const DEFAULT_MAX_PERCENT = 100;
-
 export const DEFAULT_MAX_PERCENT_DECIMAL = 1;
 
-export type GaugeChartOptionsEditorProps = OptionsEditorProps<GaugeChartOptions>;
+export interface GaugeChart {
+  kind: 'GaugeChart';
+  spec: GaugeChartOptions;
+}
 
 /**
  * The Options object type supported by the GaugeChart panel plugin.
@@ -32,6 +33,8 @@ export interface GaugeChartOptions {
   thresholds?: ThresholdOptions;
   max?: number;
 }
+
+export type GaugeChartOptionsEditorProps = OptionsEditorProps<GaugeChartOptions>;
 
 /**
  * Creates the initial/empty options for a GaugeChart panel.
--- a/ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
+++ b/ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { ThresholdOptions } from '@perses-dev/core';
+import { Definition, ThresholdOptions } from '@perses-dev/core';
 import { UnitOptions } from '@perses-dev/components';
 import { CalculationType, OptionsEditorProps } from '@perses-dev/plugin-system';
 
@@ -19,9 +19,8 @@ export const DEFAULT_UNIT: UnitOptions = { kind: 'PercentDecimal' };
 export const DEFAULT_MAX_PERCENT = 100;
 export const DEFAULT_MAX_PERCENT_DECIMAL = 1;
 
-export interface GaugeChart {
+export interface GaugeChartDefinition extends Definition<GaugeChartOptions> {
   kind: 'GaugeChart';
-  spec: GaugeChartOptions;
 }
 
 /**
--- a/ui/panels-plugin/src/plugins/gauge-chart/index.ts
+++ b/ui/panels-plugin/src/plugins/gauge-chart/index.ts
@@ -12,3 +12,4 @@
 // limitations under the License.
 
 export * from './GaugeChart';
+export type { GaugeChartDefinition } from './gauge-chart-model';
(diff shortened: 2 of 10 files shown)


ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
```diff
@@ -33,6 +34,8 @@ export interface GaugeChartOptions {
   max?: number;
 }
 
+export type GaugeChartOptionsEditorProps = OptionsEditorProps<GaugeChartOptions>;
```
cndonovan, line 37: Moved `GaugeChartOptionsEditorProps` beneath `GaugeChartOptions`

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -15,13 +15,17 @@ import { UnitOptions, LegendOptions } from '@perses-dev/components';
...
+  spec: TimeSeriesChartOptions;
+}
+
 /**
  * The Options object supported by the TimeSeriesChartPanel plugin.
  */
 export interface TimeSeriesChartOptions {
   legend?: LegendOptions;
   y_axis?: YAxisOptions;
-  unit?: UnitOptions;
```
cndonovan, line 24: I think this is legacy, since we configure `unit` inside of `y_axis` at the moment

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -56,12 +60,25 @@ export const DEFAULT_UNIT: UnitOptions = {
...
 };
 
-export const DEFAULT_LINE_WIDTH = 1.5;
+export const DEFAULT_Y_AXIS: YAxisOptions = {
+  show: true,
+  label: '',
+  unit: DEFAULT_UNIT,
+  min: undefined,
+  max: undefined,
+};
```
cndonovan, line 69: Moved all `y_axis` constants above all `visual` constants

juliepagano:
I think your failed jobs might be related to the ongoing github incident. https://www.githubstatus.com/

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
```diff
@@ -16,12 +16,13 @@ import { UnitOptions } from '@perses-dev/components';
 import { CalculationType, OptionsEditorProps } from '@perses-dev/plugin-system';
 
 export const DEFAULT_UNIT: UnitOptions = { kind: 'PercentDecimal' };
-
 export const DEFAULT_MAX_PERCENT = 100;
-
 export const DEFAULT_MAX_PERCENT_DECIMAL = 1;
 
-export type GaugeChartOptionsEditorProps = OptionsEditorProps<GaugeChartOptions>;
+export interface GaugeChart {
```
juliepagano, line 22: I think you also need to export these from the `index.ts` file if you want them to be accessible for use outside perses internal dev, which I think was the goal with wanting to use these in converter code. Same comment goes for the other models.

juliepagano (commented):
A few recommendations. It also looks like there might be some slight type errors that need to get addressed.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
cndonovan: Gooooood call

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
cndonovan: Thank you, I was wondering about that but didn't go all the way into figuring it out

juliepagano: A good quick test if something is exported is to try out the quick and dirty `typedoc` command I added. If the value you want shows up in the docs it outputs, you know it is exported for public use.

juliepagano:
I just hit a case where I'd appreciate these in the work I'm doing on table legend. Thanks for working on this! 😄

ui/panels-plugin/src/plugins/gauge-chart/gauge-chart-model.ts
juliepagano: Looking at the code more closely because I hit a case I'd need this, should these types maybe have `Spec` at the end (e.g. `GaugeChartSpec`) and extend from the `Definition` interface here?  https://github.com/perses/perses/blob/main/ui/core/src/model/definitions.ts#L17-L20

eunicorn: I was going to make the same comment that you can probably just use `Definition<GaugeChartOptions>`

cndonovan:
> I just hit a case where I'd appreciate these in the work I'm doing on table legend. Thanks for working on this! 😄

Oh good to know, I got pulled away from this PR by the bugs I'm working on. If you want to, feel fully free to take this PR and change it around if you see a good direction for it 

But also no pressure to do so. But just for your info, I probably won't get back to this until next week sometime. @juliepagano

juliepagano:
> > I just hit a case where I'd appreciate these in the work I'm doing on table legend. Thanks for working on this! 😄
> 
> Oh good to know, I got pulled away from this PR by the bugs I'm working on. If you want to, feel fully free to take this PR and change it around if you see a good direction for it
> 
> But also no pressure to do so. But just for your info, I probably won't get back to this until next week sometime. @juliepagano

I have a few other side quests of my own around the storybook decorators and this isn't blocking me, so I'll probably wait until you get back to it next week unless it's urgent for anyone else.

cndonovan:
@juliepagano @eunicorn Ready for re-review! 

Thank you for taking a look, you helped me improve this change a lot because I realized that (1) we already have types called `TimeSeriesChart`, etc. (these types correspond to the plugins), and that (2) we want to extend the `Definition` interface in this case.  

Given that, these new interfaces are more properly considered `TimeSeriesChartDefinition`, etc.

ui/panels-plugin/src/plugins/gauge-chart/index.ts
```diff
@@ -12,3 +12,4 @@
 // limitations under the License.
 
 export * from './GaugeChart';
+export type { GaugeChartDefinition } from './gauge-chart-model';
```
cndonovan, line 15: I decided to be specific / explicit with the export here, but I'm cool with doing `*` if anyone thinks that's better

juliepagano: I think it's fine to start more focused. It's much easier to add things in the future if we want to expose them than to take them away if we realize the * export exposed a feature we didn't want to support externally.

juliepagano (approved):
Changes LGTM. Thanks for taking this on! 🚀 

Can you add a description of the BREAKING CHANGES and how to handle them in the description of your PR, so consumers can use it as a reference? I'd also recommend including that information in the commit message when you squash & merge the PR.

eunicorn (approved):

PR #1160, opened 2023-05-10
[BREAKINGCHANGE] use react-virtuoso for virtualization

juliepagano:
Previously, the list legend used react-window for virtualization. We want to switch to using react-virtuoso soon for the table legend because it has significantly better handling for tables. We do not want to use two different virtualization libraries, so migrating the list legend to react-virtuoso in prep for the table legend work.

BREAKING CHANGE: Legend no longer takes `listProps` because the associated `initialRowHeight` is no longer needed with the adjusted virtualization implementation.

# Notes for reviewers

- There is a known issue where charts aren't always resizing properly when you resize a panel. I validated this isn't caused by my changes here. @sjcobb is looking into fixing it.

# Checklist

## UI Changes

# Screenshots 

Note: there are some very tiny diffs in the happo screenshots, but they look like teeny tiny font differences related to a slight pixel difference in the two approaches. I'd still consider this functionally a no-op.

<table>
	<tr>
		<th>before</th>
		<th>after</th>
	</tr>
	<tr>
		<td>
image
</td>
		<td>
image
</td>
	</tr>
</table>


ui/components/src/Legend/ListLegend.tsx
```diff
@@ -30,140 +29,40 @@ export interface ListLegendProps {
...
+  // simulate top/bottom padding based on recommendation in this
+  // issue.
+  // https://github.com/petyosi/react-virtuoso/issues/238
   const LIST_PADDING = parseInt(theme.spacing(1), 10);
+  const mockPadding = <Box sx={{ width: '100%', height: `${LIST_PADDING}px` }}></Box>;
 
   // show full labels on hover when there are many total series
   const truncateLabels = items.length > 5;
 
-  // Gets the row height for a given item to enable the virtualized list to
```
juliepagano, line 60: This code is waaaay simpler because react-virtuoso has dynamic height handling baked inside the library, so you don't have to do it yourself. I probably should have done a teeny bit more looking ahead to this table work and considered this lib instead when I did the virtualization for the list, but hindsight is always 20/20. 🤷‍♀️

eunicorn (commented):
love how much simpler it is!

eunicorn (approved):

PR #1161, opened 2023-05-11
[BUGFIX] fix time series vertical positioning

jgbernalp:
This PR fixes the vertical positioning of the time series tooltip on a dashboard with several panels.

Issue:

https://github.com/perses/perses/assets/5461414/ef274d99-6696-4331-8ba1-9ef082a5b025

Fix:

https://github.com/perses/perses/assets/5461414/ad9da70f-0641-487a-86b6-e79f6d8f185a

### DIFF
--- a/ui/components/src/TimeSeriesTooltip/utils.ts
+++ b/ui/components/src/TimeSeriesTooltip/utils.ts
@@ -11,13 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {
-  CursorCoordinates,
-  CursorData,
-  TOOLTIP_MAX_HEIGHT,
-  TOOLTIP_MAX_ITEMS,
-  TOOLTIP_MAX_WIDTH,
-} from './tooltip-model';
+import { CursorCoordinates, CursorData, TOOLTIP_MAX_WIDTH } from './tooltip-model';
 
 /**
  * Determine position of tooltip depending on chart dimensions and the number of focused series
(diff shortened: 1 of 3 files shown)


ui/components/src/TimeSeriesTooltip/utils.ts
```diff
@@ -40,17 +45,12 @@ export function assembleTransform(
...
-  if (isCloseToBottom === true) {
-    if (seriesNum > 6) {
-      y = mousePos.page.y * 0.75;
-    } else {
-      y = mousePos.page.y * 0.9;
-    }
-  } else if (mousePos.plotCanvas.y > yPosAdjustThreshold) {
-    y = mousePos.page.y * 0.95;
+  // approximate tootlip height based on number of series
+  const tooltipHeight = Math.min(Math.min(seriesNum, TOOLTIP_MAX_ITEMS) * 24, TOOLTIP_MAX_HEIGHT);
```
sjcobb, line 49: Awesome, much better than the static `if (seriesNum > 6)` hack I did before, plus this actually fixes the core problem!

sjcobb: nit: typo in comment that says `tootlip` instead of `tooltip`

sjcobb (approved):
🙏 thanks for fixing this!

juliepagano:
I'm seeing some cases where the tooltip is cut off more often with this approach than the previous one when the mouse cursor is near the bottom of the viewport. @sjcobb -- you're a lot more familiar with this bit of code than me. Do you think there's some additional tuning that should be done before merging (maybe some adjustments to the height calc?) or is this still a significant improvement over what we had before?

## With this PR 
image

## With code from `main`
image

PR #1162, opened 2023-05-11
[FEATURE] Add HTTP endpoint to manage global variable

Nexucis:
This PR is adding two new resources in the backend : `GlobalVariable` and `ProjectVariable`. These variables can be then shared and used across different dashboard.

This is something coming up in the discussion https://github.com/perses/perses/discussions/226

For example:

```bash
curl -XGET https://perses.dev/api/v1/globalvariables
```
```json
[
{
  "kind": "GlobalVariable",
  "metadata": {
    "name": "MyList"
  },
  "spec": {
    "kind": "ListVariable",
    "spec": {
      "display": {
        "name": "my awesome variable",
        "hidden": false
      },
      "default_value": [
        "default1",
        "default2"
      ],
      "allow_all_value": false,
      "allow_multiple": true,
      "plugin": {
        "kind": "PrometheusLabelValuesVariable",
        "spec": {
          "label_name": "instance",
          "matchers": [
            "up"
          ]
        }
      }
    }
  }
}
]
```

### DIFF
--- a/pkg/model/api/v1/variable.go
+++ b/pkg/model/api/v1/variable.go
@@ -66,6 +66,7 @@ func (v *VariableSpec) unmarshal(unmarshal func(interface{}) error, staticMarsha
 	return nil
 }
 
+// GlobalVariable is a global variable that be used everywhere regardless the project.
 type GlobalVariable struct {
 	Kind     Kind         `json:"kind" yaml:"kind"`
 	Metadata Metadata     `json:"metadata" yaml:"metadata"`
@@ -84,6 +85,8 @@ func (v *GlobalVariable) GetSpec() interface{} {
 	return v.Spec
 }
 
+// Variable is a global variable at project level.
+// If you are looking for the variable contained in a dashboard, take a look at dashboard.Variable
 type Variable struct {
 	Kind     Kind            `json:"kind" yaml:"kind"`
 	Metadata ProjectMetadata `json:"metadata" yaml:"metadata"`
--- a/pkg/model/api/v1/variable.go
+++ b/pkg/model/api/v1/variable.go
@@ -14,14 +14,62 @@
 package v1
 
 import (
+	"encoding/json"
+	"fmt"
+
 	modelAPI "github.com/perses/perses/pkg/model/api"
-	"github.com/perses/perses/pkg/model/api/v1/dashboard"
+	"github.com/perses/perses/pkg/model/api/v1/variable"
+	"gopkg.in/yaml.v2"
 )
 
+type VariableSpec struct {
+	// Kind is the type of the variable. Depending on the value of Kind, it will change the content of Spec.
+	Kind variable.Kind `json:"kind" yaml:"kind"`
+	Spec interface{}   `json:"spec" yaml:"spec"`
+}
+
+func (v *VariableSpec) UnmarshalJSON(data []byte) error {
+	jsonUnmarshalFunc := func(variable interface{}) error {
+		return json.Unmarshal(data, variable)
+	}
+	return v.unmarshal(jsonUnmarshalFunc, json.Marshal, json.Unmarshal)
+}
+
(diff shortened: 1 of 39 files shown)


sjcobb:
It'd be cool someday to have an OpenAPI spec that is always update to with the various REST endpoints in Perses (not sure best lib to use, this one seems popular? https://github.com/go-swagger/go-swagger)

Or even something like a Postman collection that we can commit and share (useful, but hard to keep up to date)

Nexucis:
> It'd be cool someday to have an OpenAPI spec that is always update to with the various REST endpoints in Perses (not sure best lib to use, this one seems popular? https://github.com/go-swagger/go-swagger)
> 
> Or even something like a Postman collection that we can commit and share (useful, but hard to keep up to date)

On the paper it can be a good idea. But in reality, the issue it will result is that the swagger generated from the go code won't contain the full definition of the datamodel (since some piece of it is in cuelang)

So the swagger can provide you the full list of endpoint but with incomplete datamodel which can result of a wrong usage of the API.
So I would be quite cautious to add that.

But if have somewhere the complete data model for the dashboard maybe we could like it in the swagger and that could solve this issue.
So I think the priority would be more to display the complete datamodel in a nice way (maybe interactive way)

pkg/model/api/v1/variable.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (v *GlobalVariable) GetKind() string {
+	return string(v.Kind)
+}
+
+func (v *GlobalVariable) GetSpec() interface{} {
+	return v.Spec
+}
+
+type Variable struct {
```
Gladorme, line 87: IMO, I would prefer naming `Variable` => `ProjectVariable` even if it's longer, in order to not think it's a dashboard variable 😄

Nexucis: in a REST API terms, having a path starting by `/projects/perses/<resources>` already mean the resources is part of a project. So naming a resource `project<resource>` is redundant :/

Nexucis:
> It'd be cool someday to have an OpenAPI spec that is always update to with the various REST endpoints in Perses (not sure best lib to use, this one seems popular? https://github.com/go-swagger/go-swagger)
> 
> Or even something like a Postman collection that we can commit and share (useful, but hard to keep up to date)

apparently @celian-garcia is trying something internally to generate the swagger in a nice way and it should give some nice result. So probably we will be able to have it here as well :)

Nexucis:
but the datamodel will still be incomplete until we are able to generate it in a proper documentation

pkg/model/api/v1/variable.go
```diff
@@ -84,6 +85,8 @@ func (v *GlobalVariable) GetSpec() interface{} {
 	return v.Spec
 }
 
+// Variable is a global variable at project level.
+// If you are looking for the variable contained in a dashboard, take a look at dashboard.Variable
```
AntoineThebaud, line 89: Variable relates to variables defined at project level.
If you are looking for variable defined at dashboard level, see dashboard.Variable

celian-garcia:
> It'd be cool someday to have an OpenAPI spec that is always update to with the various REST endpoints in Perses (not sure best lib to use, this one seems popular? https://github.com/go-swagger/go-swagger)
> 
> Or even something like a Postman collection that we can commit and share (useful, but hard to keep up to date)

@sjcobb @Nexucis What I'm working on in our internal tool is to move from go-swagger to swaggo. 
It looks very less verbose because go-swagger forced us to define some structures for the requests/responses data model, and swaggo parse the code a bit better.

Here you have in addition a cuelang datamodel ? That would be complex, maybe we would have to do a cuelang > golang script to port the datamodel in something readable by swaggo or go-swagger

I can try with perses when I'm done with our internal project

Nexucis:
I think you shouldn't take care about the partial definition you can have in cuelang.
For the moment having the swagger generated with the partial definition will already be good. Probably we will have to put some warnings in the comment on the plugin.

PR #1164, opened 2023-05-11
[BUGFIX] Revert allow echarts wrapper to default to fluid width and height

sjcobb:
Reverts perses/perses#1115

Previous PR doesn't work since the sx prop is needed in the dependency array [here](https://github.com/perses/perses/blob/v0.29.1/ui/components/src/EChart/EChart.tsx#L178-L184)


juliepagano (approved):
LGTM! Thanks for the quick revert. I'd recommend doing a patch release after this is merged, so any consumers can get a quick fix for this.

PR #1165, opened 2023-05-11
[FEATURE] Option to save default variable values and time range

sjcobb:
Logic to check whether time range or variables have changed from what is stored (both list and text variables are supported). Function to check whether Zustand local state is out of sync with stored definitions happens onSave.

Note: For saving a new time range, we only store a `duration` so saving default absolute time ranges is not supported now.

Currently, we are using a dialog and checkboxes to let the user pick which to persist, in the future, we may experiment with other options.

## Screenshots and recording
image

image

https://github.com/perses/perses/assets/9369625/77b8e71a-23ec-4a4d-95a6-c3ad215315ff

https://github.com/perses/perses/assets/9369625/f6a0d214-8ee3-4765-9c5b-05df949a9869

# Checklist

## UI Changes


ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -61,11 +76,39 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
   );
 
   const onSaveButtonClick = () => {
+    const { newVariables, isSelectedVariablesUpdated } = updateVariableDefaultValues(savedVariables, variableValues);
```
eunicorn, line 79: Do you think it makes sense to make a `SaveButton` component, similar to how we export `EditJsonButton` and `AddPanelButton`? There is a lot of logic related to this save button now, so I am thinking it might be better if we encapsulate this in its own file

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
```diff
@@ -61,11 +76,39 @@ export const DashboardToolbar = (props: DashboardToolbarProps) => {
...
+      openSaveChangesConfirmationDialog({
+        onSaveChanges: (saveDefaultTimeRange, saveDefaultVariables) => {
+          if (isRelativeTimeRange(timeRange) && saveDefaultTimeRange === true) {
+            dashboard.spec.duration = timeRange.pastDuration;
+          }
+          if (saveDefaultVariables === true) {
+            dashboard.spec.variables = newVariables;
+          }
+          saveDashboard();
+          refresh();
```
eunicorn, line 94: is there a reason why we need to call `refresh` here?

ui/dashboards/src/components/SaveChangesConfirmationDialog/SaveChangesConfirmationDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    ? `(Last ${timeRange.pastDuration})`
+    : '(Absolute time ranges can not be saved)';
+
+  const timeRangeInfoText = `Save current time range as new default ${currentTimeRangeText}`;
+
+  return (
+    <Dialog open={isOpen}>
+      {dialog !== undefined && (
+        <>
+          <DialogTitle>Save Dashboard</DialogTitle>
```
eunicorn, line 50: We actually have a [reusable `Dialog` component](https://github.com/perses/perses/tree/main/ui/components/src/Dialog) in perses so you don't need to add these custom styles anymore!

ui/dashboards/src/components/Variables/TemplateVariable.tsx
```diff
@@ -107,6 +108,7 @@ function ListVariable({ name }: TemplateVariableProps) {
           label={title}
           value={selectValue}
           onChange={(e) => {
+            setVariableDefaultValue(name, e.target.value);
```
eunicorn, line 111: I feel like it makes more sense if we only call this `onSave` and not on every change

ui/dashboards/src/components/Variables/variable-model.ts
```diff
@@ -101,3 +102,42 @@ export const VARIABLE_TYPES = [
   { label: 'List', kind: 'ListVariable' },
   { label: 'Text', kind: 'TextVariable' },
 ] as const;
+
+/*
+ * Check whether saved variable definitions are out of date with current default list values in Zustand store
+ */
+export function updateVariableDefaultValues(
```
eunicorn, line 109: So i think you can actually move this logic to the TemplateVariableStore so you don't need to pass in savedVariables and can also directly call `setDefinitions` there, which will automatically update the dashboard json

eunicorn: happy to jump on a call to discuss if this isn't clear :)

cndonovan (commented):
This is me having thoughts about copy:

Since we know the dialog is going to appear if and only if the user DOES edit the variables or time range, I was thinking it could say:

`You have made changes to the time range or the variables values. Would you like to save these as defaults?`

Ultimately up to whatever Mohit thinks is best, I wrote it that way to be more obvious about why the dialog is appearing

ui/dashboards/src/components/SaveChangesConfirmationDialog/SaveChangesConfirmationDialog.tsx
sjcobb: Can't believe I forgot about that! I was being lazy and copied DiscardChangesConfirmationDialog.tsx into SaveChangesConfirmationDialog.tsx (and just changed the 'Discard' to 'Save'), might do a follow-up PR to backfill all dialogs to use the new one from @perses-dev./components

ui/dashboards/src/components/Variables/TemplateVariable.tsx
sjcobb: After refactoring and moving the updateVariableDefaultValues logic to the variable store, this isn't needed and is removed now!

ui/dashboards/src/components/DashboardToolbar/DashboardToolbar.tsx
sjcobb: Removed this, not needed anymore after updates in TemplateVariableProvider.tsx

ui/dashboards/src/components/SaveDashboardButton/SaveDashboardButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    // Save dashboard if active timeRange from plugin-system is relative and different than currently saved
+    if (timeRangeUpdated || variableDefaultValuesUpdated) {
+      openSaveChangesConfirmationDialog({
+        onSaveChanges: (variableDefinitions, saveDefaultTimeRange, saveDefaultVariables) => {
+          if (isRelativeTimeRange(timeRange) && saveDefaultTimeRange === true) {
+            dashboard.spec.duration = timeRange.pastDuration;
+          }
+          if (saveDefaultVariables === true) {
+            dashboard.spec.variables = variableDefinitions;
```
eunicorn, line 53: so i think we should be calling `setVariableDefaultValues` here because otherwise the json and store will be out of sync

```suggestion
            const variables = setVariableDefaultValues();
            dashboard.spec.variables = variables;
```

eunicorn: To return the new variable definitions, you can do this in the store: 

```suggestion
        setVariableDefaultValues: () => {
            const variableDefinitions = get().variableDefinitions;
            const variableState = get().variableDefinitions;
            const updatedVariables = produce( ... ); 
            set((state) => { state.variableDefinitions = updatedVariables })
            return updatedVariables;
        }
```

That way you don't need to call `useTemplateVariableDefinitions` in the save changes dialog too

sjcobb: Makes a lot of sense and much better than the hacky way I did it before! This is updated now, I'm still adjusting the checkSavedVariablesStatus util, and clarifying with Mohit about behavior after adding a brand new ListVariable (see failing e2e), but otherwise this is ready for another review

ui/e2e/src/tests/saveDefaults.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    await toolbarSaveButton.click();
+
+    // Save defaults confirmation dialog should open, click btn to save
+    const dialogText = await dashboardPage.page.getByText(SAVE_DEFAULTS_DIALOG_TEXT);
+    await expect(dialogText).toBeVisible();
+    const dialogSaveButton = await dashboardPage.page.getByRole('button', { name: 'Save Changes' });
+    await dialogSaveButton.click();
+
+    // Confirm correct default list and text variables are persisted
+    await page.reload();
```
eunicorn, line 62: i think we need to clear the query params in the url first, otherwise this test will always pass because the url has the correct values

ui/dashboards/src/components/SaveDashboardButton/SaveDashboardButton.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import { useState } from 'react';
+import { Button, ButtonProps } from '@mui/material';
+import { DashboardResource, isRelativeTimeRange } from '@perses-dev/core';
+import { useTimeRange } from '@perses-dev/plugin-system';
+import { useDashboard, useEditMode, useSaveChangesConfirmationDialog, useTemplateVariableActions } from '../../context';
+
+export interface SaveDashboardButtonProps extends Pick<ButtonProps, 'fullWidth'> {
+  onSave?: (entity: DashboardResource) => Promise<DashboardResource>;
+  isReadonly: boolean;
```
eunicorn, line 22: nit: maybe rename `isDisabled`?

ui/dashboards/src/components/Variables/variable-model.ts
```diff
@@ -101,3 +101,25 @@ export const VARIABLE_TYPES = [
...
+/*
+ * Check whether saved variable definitions are out of date with current default list values in Zustand store
+ */
+export function checkSavedVariablesStatus(definitions: VariableDefinition[], varState: VariableStateMap) {
+  let isSavedVariablesOutdated = false;
+  definitions.forEach((saveVariable) => {
+    if (saveVariable.kind === 'ListVariable') {
+      const currentVariable = varState[saveVariable.spec.name];
+      if (saveVariable.spec.default_value !== currentVariable?.value) {
+        isSavedVariablesOutdated = true;
```
eunicorn, line 114: i think we can just return true here once we know it's outdated

ui/dashboards/src/components/Variables/variable-model.ts
```diff
@@ -101,3 +101,25 @@ export const VARIABLE_TYPES = [
   { label: 'List', kind: 'ListVariable' },
   { label: 'Text', kind: 'TextVariable' },
 ] as const;
+
+/*
+ * Check whether saved variable definitions are out of date with current default list values in Zustand store
+ */
+export function checkSavedVariablesStatus(definitions: VariableDefinition[], varState: VariableStateMap) {
```
eunicorn, line 108: oh wait i noticed this is duplicated, does this need to be removed?

ui/e2e/src/tests/saveDefaults.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    await textVariableInput.clear();
+    await textVariableInput.type(TEXT_VARIABLE_UPDATED_VALUE, { delay: 100 });
+
+    // Switch to edit mode and click save
+    await dashboardPage.startEditing();
+    const toolbarSaveButton = await dashboardPage.page.getByRole('button', { name: 'Save' });
+    await toolbarSaveButton.click();
+
+    // Save defaults confirmation dialog should open, click btn to save
+    const dialogText = await dashboardPage.page.getByText(SAVE_DEFAULTS_DIALOG_TEXT);
```
eunicorn, line 56: FYI  there is also a `getDialog` helper function if you want to use it

eunicorn (approved):
YAY LGTM 🚀

ui/dashboards/src/components/Variables/variable-model.ts
sjcobb: Yeah lol whoops, I’ll update this, before I clarified with Mohit, I was keeping track of two booleans in this function but he said it was fine to show the dialog after a new variable is created

ui/dashboards/src/components/Variables/variable-model.ts
sjcobb: Yeah I’ll remove this!

ui/dashboards/src/components/SaveChangesConfirmationDialog/SaveChangesConfirmationDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+                label={timeRangeInfoText}
+              />
+              <FormControlLabel
+                control={
+                  <Checkbox
+                    checked={saveDefaultVariables}
+                    onChange={(e: React.ChangeEvent<HTMLInputElement>) => setSaveDefaultVariables(e.target.checked)}
+                  />
+                }
+                label="Save current variables values as new default"
```
juliepagano, line 61: Grammar nit: I think this might read better as "Save current variable values as new default."

ui/e2e/src/tests/jsonEditor.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import updatedDefaultsDashboard from '../data/updatedDefaultsDashboard.json';
+import { test, expect } from '../fixtures/dashboardTest';
+
+test.use({
+  dashboardName: 'EditJson',
+  modifiesDashboard: true,
+});
+
+test.describe('Dashboard: EditJson', () => {
```
juliepagano, line 22: Thanks for adding a test for this! 🥇

ui/e2e/src/tests/saveDefaults.spec.ts
juliepagano: +1 on this.

.github/workflows/e2e.yml
```diff
@@ -28,6 +28,9 @@ jobs:
       - name: Install UI deps
         working-directory: ./ui
         run: npm ci
+      - name: Build the app and all packages
+        working-directory: ./ui
+        run: npm run build
```
juliepagano, line 33: In theory we might also want to make this a dependency in the `turbo` config for running the e2e tests, but that gets kinda funky during local development because in that case you usually have already started in dev mode and have the code built and it'll get weird if you have competing builds. I'm inclined to say to leave it alone for now, BUT to add a note about it somewhere in the `e2e/README.md` getting started notes, so people know what to do if they hit issues with e2e in local dev.

juliepagano (commented):
A handful of small comments from a skim and playing with the running code. I didn't give this a super-deep look since multiple other reviewers already did so, and they looked like they already caught most of the things that were coming up for me.

sjcobb:
LOL I couldn't figure out why the variables test kept failing, turns out when I added the EditJson dashboard, it caused the temporary `Variables_can_add_simple_text_variable_0` dashboard to move to the second page in dashboards table, so the test would time out since it never navigated to that dashboard 🙃 

image

ui/app/src/views/home/ProjectsAndDashboards.tsx
```diff
@@ -89,6 +89,9 @@ function RenderDashboardList(props: RenderDashboardListProps) {
               dashboardList={row.dashboards}
               hideToolbar={true}
               initialState={{
+                pagination: {
+                  paginationModel: { pageSize: 25, page: 0 },
```
sjcobb, line 93: Added this as a workaround for e2e failures caused by temp dashboards not showing on the first page. Increasing pageSize from 10 to 25 ensures those dashboards can be easily accessed without having to click the next arrow (more info [here](https://github.com/perses/perses/pull/1165#issuecomment-1550557092))

eunicorn: oh can you also search for the dashboard?

PR #1175, opened 2023-05-16
[ENHANCEMENT] Refactor time series chart panel selection

juliepagano:
This commit refactors the time series chart panel selection mechanisms in the following ways:
- Switch from tracking selected items using `name` to using a unique `id` to avoid unintentionally setting multiple items if they have the same label.
- Switch from storing selected items using an array of names to a record of ids and booleans. This should make some operations more performant with large data sets and is more aligned with how we'll want to manage checkbox-based selection with the in progress table legend work.

# Notes for code reviewers

- This is another refactor in service to the in flight work I'm doing on the table legend to avoid that initial PR being super large and nasty to review. 
- I played around with some larger refactoring while doing this that I think would benefit the time series chart panel, but it made the diffs in git really nasty. I'll try to come back to it in a follow-up, so it's a cleaner change.

# Checklist

## UI Changes

# Screenshots

This should ideally be a noop from a user experience perspective. Below is a video showing a combo of single and modified clicks.

https://github.com/perses/perses/assets/396962/13e99a70-c6d2-4c0f-b79a-931548f617c6

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -13,7 +13,7 @@
 
 import { useState } from 'react';
 import { merge } from 'lodash-es';
-import { useDeepMemo, StepOptions, getXValues, getYValues } from '@perses-dev/core';
+import { useDeepMemo, StepOptions, getXValues, getYValues, TimeSeries } from '@perses-dev/core';
 import { PanelProps, useDataQueries, useTimeRange } from '@perses-dev/plugin-system';
 import type { GridComponentOption } from 'echarts';
 import { Box, Skeleton, useTheme } from '@mui/material';
@@ -27,6 +27,7 @@ import {
   ZoomEventData,
   useChartsTheme,
 } from '@perses-dev/components';
+import produce from 'immer';
 import {
   TimeSeriesChartOptions,
   DEFAULT_UNIT,
@@ -46,6 +47,16 @@ import {
 import { getSeriesColor } from './utils/palette-gen';
 
 export type TimeSeriesChartProps = PanelProps<TimeSeriesChartOptions>;
+
+// Using an "ALL" value to handle the case on first loading the chart where we
+// want to select all, but do not want all of the legend items to be visually highlighted.
+// This helps us differentiate those cases more clearly instead of inferring it
+// based on the state of the data. This also helps us avoid some coding
+// complexity around initializing a full record for the initial load that would
+// currently require significantly more refactoring of this component.
+// TODO: simplify this if we switch the list-based legend UI to use checkboxes,
+// where we *would* want to visually select all items in this case.
+type SelectedSeriesState = Record<string, boolean> | 'ALL';
 
 export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   const {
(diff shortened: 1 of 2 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -47,6 +48,16 @@ import { getSeriesColor } from './utils/palette-gen';
...
 export type TimeSeriesChartProps = PanelProps<TimeSeriesChartOptions>;
 
+// Using an "ALL" value to handle the case on first loading the chart where we
+// want to select all, but do not want all of the legend items to be visually highlighted.
+// This helps us differentiate those cases more clearly instead of inferring it
+// based on the state of the data. This also helps us avoid some coding
+// complexity around initializing a full record for the initial load that would
+// currently require significantly more refactoring of this component.
+// TODO: simplify this if we switch the list-based legend UI to use checkboxes,
+// where we *would* want to visually select all items in this case.
```
juliepagano, line 58: Some additional context that I thought was probably a bit much to put in here, but useful for in flight work:
- The prior implementation inferred "select all" from an empty selected series. This works great for the list legend where there is no "select none," but will become a problem for the table legend where the user can have a "select none" state because it uses checkboxes.
- Initializing a full record of "true" values for the "select all" state is tricky with how all of the data is managed in this component right now. We'd really benefit from breaking it out into 2-3 components, so that we can separate data fetching, transforming the data for rendering, and interaction state initialized from the prior two (e.g. selected items). I really didn't want to take that refactor on while other people are touching this component. Even if I did, it wouldn't address the case of not wanting to select all the list legend items in the initial "select all" state.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -47,6 +48,16 @@ import { getSeriesColor } from './utils/palette-gen';
...
+type SelectedSeriesState = Record<string, boolean> | 'ALL';
```
juliepagano, line 59: Swapping from an array to an object because:
- It'll line up better with the state needed for table legend 
- It should be more performant for lookup operations.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -94,34 +105,44 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   // convert Perses dashboard format to be ECharts compatible
   const echartsYAxis = convertPanelYAxis(y_axis);
 
-  const [selectedSeriesNames, setSelectedSeriesNames] = useState<string[]>([]);
+  const [selectedSeriesNames, setSelectedSeriesNames] = useState<SelectedSeriesState>('ALL');
 
   const { setTimeRange } = useTimeRange();
 
-  const onLegendItemClick = (e: React.MouseEvent<HTMLElement, MouseEvent>, seriesName: string) => {
+  const onLegendItemClick = (e: React.MouseEvent<HTMLElement, MouseEvent>, seriesId: string) => {
```
juliepagano, line 112: Swapping to seriesId to avoid accidentally selecting multiple series with the same label.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -167,7 +188,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
       if (result.isLoading || result.isFetching || result.data === undefined) continue;
 
       for (let i = 0; i < result.data.series.length; i++) {
-        const timeSeries = result.data.series[i];
+        const timeSeries: TimeSeries | undefined = result.data.series[i];
```
juliepagano, line 191: For some reason, TS had trouble inferring this type after one of the changes below. Explicitly setting it to fix that.

ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts
```diff
@@ -29,7 +34,7 @@ import {
 
 export type RunningQueriesState = ReturnType<typeof useTimeSeriesQueries>;
 
-export const EMPTY_GRAPH_DATA = {
+export const EMPTY_GRAPH_DATA: EChartsDataFormat = {
```
juliepagano, line 37: I ran into a type error with this at some point while working in here because this wasn't inferred to be the same as the other graph data. Explicitly typing it to match.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Love the idea about splitting up into multiple components the way you described

sjcobb (approved):

PR #1176, opened 2023-05-16
[ENHANCEMENT] Add `noDataVariant` to `LineChart`

juliepagano:
This new optional prop enables the user to choose one of two displays when the line chart does not have data:
- `message` (default): renders a simple "no data" message to the user. This was the prior behavior before the addition of this prop.
- `chart`: renders a very simple chart with the available information (e.g. x axis).

# Notes for reviewers

- This is yet another small PR in prep for the table legend. We want to show an empty chart (instead of a "no data" message) when the user unselects all the items in a table legend using the checkboxes.

# Checklist

## UI Changes

# Screenshots

This is only in storybook right now. Screenshot of the story that exercises this prop.

image

### DIFF
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -147,6 +147,9 @@ export function LineChart({
 
   const option: EChartsCoreOption = useMemo(() => {
     if (data.timeSeries === undefined) return {};
+
+    // The "chart" `noDataVariant` is only used when the `timeSeries` is an
+    // empty array because a `null` value will throw an error.
     if (data.timeSeries === null || (data.timeSeries.length === 0 && noDataVariant === 'message')) return noDataOption;
 
     // show symbols and axisPointer dashed line on hover
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -73,6 +73,7 @@ export interface LineChartProps {
   grid?: GridComponentOption;
   legend?: LegendComponentOption;
   tooltipConfig?: TooltipConfig;
+  noDataVariant?: 'chart' | 'message';
   onDataZoom?: (e: ZoomEventData) => void;
   onDoubleClick?: (e: MouseEvent) => void;
   __experimentalEChartsOptionsOverride?: (options: EChartsCoreOption) => EChartsCoreOption;
@@ -86,6 +87,7 @@ export function LineChart({
   grid,
   legend,
   tooltipConfig = { wrapLabels: true },
+  noDataVariant = 'message',
   onDataZoom,
   onDoubleClick,
   __experimentalEChartsOptionsOverride,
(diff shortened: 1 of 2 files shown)


ui/components/src/LineChart/LineChart.tsx
```diff
@@ -86,6 +87,7 @@ export function LineChart({
   grid,
   legend,
   tooltipConfig = { wrapLabels: true },
+  noDataVariant = 'message',
```
juliepagano, line 90: I went with this prop name because it's managing different variations of display when no data is available, but I'm option to other suggestions.

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -145,7 +147,10 @@ export function LineChart({
 
   const option: EChartsCoreOption = useMemo(() => {
     if (data.timeSeries === undefined) return {};
-    if (data.timeSeries === null || data.timeSeries.length === 0) return noDataOption;
+
+    // The "chart" `noDataVariant` is only used when the `timeSeries` is an
+    // empty array because a `null` value will throw an error.
```
juliepagano, line 152: The `null` case should also be pretty rare if people are using TS because I'm pretty sure the types say it should be an array.

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -191,7 +196,7 @@ export function LineChart({
       return __experimentalEChartsOptionsOverride(option);
     }
     return option;
-  }, [data, yAxis, unit, grid, legend, noDataOption, timeZone, __experimentalEChartsOptionsOverride]);
+  }, [data, yAxis, unit, grid, legend, noDataOption, timeZone, __experimentalEChartsOptionsOverride, noDataVariant]);
```
sjcobb, line 199: This dependency array has almost as many items in it as the one TimeSeriesChartPanel now 🙃 
I hope we get more time to refactor both components soon!

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -73,6 +73,7 @@ export interface LineChartProps {
   grid?: GridComponentOption;
   legend?: LegendComponentOption;
   tooltipConfig?: TooltipConfig;
+  noDataVariant?: 'chart' | 'message';
```
sjcobb, line 76: The other option I can think of is a more nested `variant` object, which would allow for customization of other properties in the future (might be overkill / pre-optimization though):
```ts
variant: {
  noData: 'chart' | 'message';
  <other future variants>
}
```

Which makes me second-guess the way I named that `tooltipConfig` prop too 🤔

sjcobb: If we did have a variant object like that, it could potentially help with my other comment about the dependency array growing since we could potentially move other props inside it (although not sure which ones currently, we can probably just get rid of ‘legend’ now since no one should be using the ECharts legend anymore)

ui/components/src/LineChart/LineChart.tsx
juliepagano: Saaaame. I have some thoughts for the `TimeSeriesPanel` and almost did a little breaking up in https://github.com/perses/perses/pull/1175, but realized it'd make the git diffs incomprehensible. I'm hoping I'll be able to come back and take a pass at it once I get the basic table legend PR out.

ui/components/src/LineChart/LineChart.tsx
eunicorn: what about `noDataDisplay`?

eunicorn: ok with `noDataVariant` too

eunicorn (approved):

eunicorn (approved):

sjcobb (approved):

sjcobb:
Going ahead and merging so I can pull these changes into my highlighting #1112 branch!

PR #1177, opened 2023-05-17
[FEATURE] Revamp dashboard variables form

Gladorme:
# Description
Adding datasource selection when creating/editing a list variable + revamping a little bit the variable form UI to be coherent with panel editor form:
- Similar metadata fields placement
- Moving preview at the top
- Preview available for all type of variable
- Replacing copy alert() by a snackbar
- Replacing PromQL Expression basic input by codemirror input
- Fixing a issue with preview values when a custom regex was typed, then removed
- And maybe more, but I forgot 😄 

# Images

## Custom list
image

## PromQL Result Value
image

##Label Names
image

## Text variable
image

# Checklist

## UI Changes


saminzadeh:
Nice! Looking from a UX / product lens - these changes all make sense.

> Moving preview at the top

Not too many strong opinions but I would prefer this to be at the bottom. The main reason why is that the number of values in the box will change and dictate the height of the container. That will shift the configuration around quite a bit depending on the results.

Gladorme:
> Not too many strong opinions but I would prefer this to be at the bottom. The main reason why is that the number of values in the box will change and dictate the height of the container. That will shift the configuration around quite a bit depending on the results.

I think it should be fine with the max preview value feature limiting the amount of values displayed

Gladorme:
About Happo issue, I move Prometheus Datasource selection at the top. It's something you should define first, before doing your query

ui/dashboards/src/context/index.ts
```diff
@@ -14,4 +14,5 @@
 export * from './DashboardProvider';
 export * from './DatasourceStoreProvider';
 export * from './TemplateVariableProvider';
+export * from './SnackbarProvider';
```
saminzadeh, line 17: This probably makes sense to go into `@components` package rather than Dashboarding

ui/dashboards/src/components/Variables/VariableEditorForm/VariablePreview.tsx
```diff
@@ -0,0 +1,108 @@
...
+const DEFAULT_MAX_PREVIEW_VALUES = 50;
+
+interface VariablePreviewProps {
+  values?: string[];
+  onRefresh?: () => void;
+  isLoading?: boolean;
+  error?: string;
+}
+
+export function VariablePreview(props: VariablePreviewProps) {
```
saminzadeh, line 33: Nice - makes sense to have this be a "dumb" view component

saminzadeh (approved):
LGTM - thank you! One suggestion of moving the snackbar to `@components` instead of `@dashboard`

saminzadeh:
> About Happo issue, I move Prometheus Datasource selection at the top. It's something you should define first, before doing your query

Accepted it

PR #1179, opened 2023-05-18
[BREAKINGCHANGE]: move legend interactions to legend

juliepagano:
Prior to this change, much of the legend selection interactions were managed inside `TimeSeriesChartPanel`, requiring consumers to duplicate this logic when using the legend in other use cases.

After this change, the legend selection interactions are managed inside `Legend` and its child components, so that it can be used wherever the legend is. The selection state remains "controlled."

BREAKINGCHANGE: Refactoring `Legend` to manage legend interactions includes the following breaking changes:
- The `Legend` component now requires the following props to control the selected state of the items: `selectedItems`, `onSelectedItemsChange`.
- The `LegendItem` type no longer allows the `isSelected` property. The selected state of items are now managed by the `selectedItems` prop on the `Legend` component.
- The `onClick` property on the `LegendItem` type is now optional because it is no longer used to manage the selected state of items. Instead use the `onSelectedItemsChange` prop on the `Legend` component if you need to listen for selection changes.

# Notes for reviewers

- This is yet another improvement I'm making as part of the lead-up to table legends landing, but it's also a good refactor to move the logic closer to the component that should own it.
- This is a no-op from a UX perspective for the time series panel, but does include some breaking changes in the props for Legend as noted above.

# Checklist

## UI Changes

# Screenshots

This is a no-op for the time series panel. Below is a video showing the interactions are still the same.

https://github.com/perses/perses/assets/396962/9d8e8326-b878-4dc5-9b6c-5df74e66f2f0

The visuals of the legend are the same now, but the interaction patterns for selection are now more tightly baked in. You can see that in the stories.

https://github.com/perses/perses/assets/396962/2f508e3f-d64c-41b3-b6e4-640ff08e11cf

Added a story to the legend to show the selection options.
image

### DIFF
--- a/ui/components/src/Legend/Legend.tsx
+++ b/ui/components/src/Legend/Legend.tsx
@@ -12,15 +12,29 @@
 // limitations under the License.
 
 import { Box } from '@mui/material';
-import { LegendOptions, LegendItem } from '../model';
-import { ListLegend } from './ListLegend';
-import { CompactLegend } from './CompactLegend';
+import { produce } from 'immer';
+import { LegendOptions, LegendItem, SelectedLegendItemState } from '../model';
+import { ListLegend, ListLegendProps } from './ListLegend';
+import { CompactLegend, CompactLegendProps } from './CompactLegend';
 
 export interface LegendProps {
   width: number;
   height: number;
   data: LegendItem[];
   options: LegendOptions;
+
+  /**
+   * State of selected items in the legend.
+   *
+   * Selected legend item state is a controlled value that should be managed using a
+   * combination of this prop and `onSelectedChange`.
+   */
+  selectedItems: SelectedLegendItemState;
+
+  /**
+   * Callback fired when the selected items in the legend changes.
+   */
+  onSelectedItemsChange: (newSelected: SelectedLegendItemState) => void;
 }
 
 // When the number of items to display is above this number, it is likely to
@@ -30,7 +44,50 @@ export interface LegendProps {
 // future as people test this out on different machines.
 const NEED_VIRTUALIZATION_LIMIT = 500;
 
-export function Legend({ width, height, options, data }: LegendProps) {
+export function Legend({ width, height, options, data, selectedItems, onSelectedItemsChange }: LegendProps) {
+  const onLegendItemClick = (e: React.MouseEvent<HTMLElement, MouseEvent>, seriesId: string) => {
+    const isModifiedClick = e.metaKey || e.shiftKey;
+
+    const newSelected = produce(selectedItems, (draft) => {
+      if (draft === 'ALL') {
+        return {
+          [seriesId]: true,
+        };
+      }
+
+      const isSelected = !!draft[seriesId];
+
+      // Clicks with modifier key can select multiple items.
+      if (isModifiedClick) {
+        if (isSelected) {
+          // Modified click on already selected item. Remove that item.
+          delete draft[seriesId];
+        } else {
+          // Modified click on not-selected item. Add it.
+          draft[seriesId] = true;
+        }
+        return draft;
+      }
+
+      if (isSelected) {
+        // Clicked item was already selected. Unselect it and return to
+        // ALL state.
+        return 'ALL' as const;
+      }
+
+      // Select clicked item.
+      return { [seriesId]: true };
+    });
+    onSelectedItemsChange(newSelected);
+  };
+
+  const commonLegendProps: ListLegendProps | CompactLegendProps = {
+    height,
+    items: data,
+    selectedItems,
+    onLegendItemClick,
+  };
+
   if (options.position === 'Right') {
     return (
       <Box
@@ -42,7 +99,7 @@ export function Legend({ width, height, options, data }: LegendProps) {
           right: 0,
         }}
       >
-        <ListLegend items={data} width={width} height={height} />
+        <ListLegend {...commonLegendProps} width={width} />
       </Box>
     );
   }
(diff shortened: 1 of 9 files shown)


ui/components/src/Legend/ListLegendItem.tsx
```diff
@@ -17,9 +17,16 @@ import { LegendItem } from '../model';
...
 import { LegendColorBadge } from './LegendColorBadge';
 
-interface ListLegendItemProps extends ListItemProps<'div'> {
+export interface ListLegendItemProps extends Omit<ListItemProps<'div'>, 'onClick'> {
   item: LegendItem;
 
+  /**
+   * When true, the item is visually highlighted to communicate it is selected.
+   */
+  isHighlighted?: boolean;
```
juliepagano, line 26: Calling this `isHighlighted` instead of `isSelected` because in the "ALL" case, everything is technically selected but is NOT visually highlighted as such.

ui/components/src/model/legend.ts
```diff
@@ -24,15 +24,23 @@ export interface LegendOptions {
 export interface LegendItem {
   id: string;
   label: string;
-  isSelected: boolean;
   color: string;
-  onClick: MouseEventHandler<HTMLElement>;
+  onClick?: MouseEventHandler<HTMLElement>;
```
juliepagano, line 28: Making this optional because it is no longer required to manage the legend selection. Now it's an optional handler in case you want to do something extra when a legend item is clicked.

ui/components/src/Legend/ListLegendItem.tsx
sjcobb: The only thing that might be confusing is a few people have asked about hovering on legend items, highlighting the corresponding line in the chart.

Once the legend is moved inside LineChart.tsx, that could be possible (since the legend would have access to the chart ref) by dispatching a [highlight](https://echarts.apache.org/en/api.html#action.highlight) action similar to the tooltip.

In that case highlighting === emphasized, so all the terms get a little confusing (there is also "focused" in the tooltip for the series that are nearby the cursor 😂 ) .

ui/components/src/model/legend.ts
sjcobb: Nice, in the future I could definitely see an onHover being useful here too for dispatching that chart highlight action:
```typescript
chart.dispatchAction({
  type: 'highlight',
  seriesIndex: seriesIndex, // legend item being hovered over
});

**Review by sjcobb — APPROVED** [review]
LGTM! Left one question about the term 'highlight' having multiple meanings, not sure the best solution there if selected isn't right anymore though

File: ui/components/src/model/legend.ts (hunk 2 above)
Comment (juliepagano) [reply ui/components/src/model/legend.ts:28]: 💯 Implementing that is on my list of things to do for the table legend, so I'll likely add an `onHover` or something like that in the relatively near future.

File: ui/components/src/Legend/ListLegendItem.tsx (hunk 1 above)
Comment (juliepagano) [reply ui/components/src/Legend/ListLegendItem.tsx:26]: That's a good callout. I could rename this to `isVisuallySelected` which is a little more verbose, but more explicit about what it is and reduces overlap with the terminology used in echarts.

Comment (juliepagano) [reply ui/components/src/Legend/ListLegendItem.tsx]: done!

Comment (sjcobb) [reply ui/components/src/Legend/ListLegendItem.tsx]: That sounds great!

File: ui/components/src/Legend/Legend.tsx [hunk 3]
```diff
@@ -30,7 +44,50 @@ export interface LegendProps {
...
+          // Modified click on not-selected item. Add it.
+          draft[seriesId] = true;
+        }
+        return draft;
+      }
+
+      if (isSelected) {
+        // Clicked item was already selected. Unselect it and return to
+        // ALL state.
+        return 'ALL' as const;
```
eunicorn, line 75: instead of setting it to 'ALL' state, do you think it should only deselect the one that you click?

ui/components/src/model/legend.ts
```diff
@@ -24,15 +24,23 @@ export interface LegendOptions {
...
 };
 
+/**
+ * State of selected items in the legend.
+ * - When "ALL", all legend items are selected, but not visually highlighted.
+ * - Otherwise, it is a Record that associates legend item ids with a boolean
+ *   value. When the associated entry for a legend item is `true`, that item
+ *   will be treated as selected and visually highlighted.
+ */
+export type SelectedLegendItemState = Record<LegendItem['id'], boolean> | 'ALL';
```
eunicorn, line 42: I wonder if we need `ALL` here because i think it should be up to the consumers to decide what it means when none is selected

ui/components/src/Legend/Legend.tsx
juliepagano: I think that would make more sense with a checkbox interface, but that's not what the list legend has right now. You can unselect a single item with a modified click (the conditional above). I think we're going to consider moving this to a checkbox interface after we get the table legend work (which does use checkboxes) working, so this logic will probably change in the near future. 

In the near term, the goal with this PR is to keep the logic as it was, but to move it down more closely to the legend component (instead of having it live in time series panel).

ui/components/src/model/legend.ts
juliepagano: I think it's reasonable for the Legend to be a little opinionated about the user experience/interaction for consistency across use and enabling some reasonable defaults for the behavior. The "ALL" is basically a special handling case for the behavior of "everything is selected, but not visually" that we want by default in the list behavior. 

The `selected` and `onSelectedItemsChange` props make this a controlled component, so the consumer can adjust the values if they want some slightly more custom behavior. If the consumer wants a true "none selected," they can set `selected` to an empty record or a record with all the entries set to false. If they want a "select all and visually select all," they can set `selected` to a record with all the entries set to true. 

Similar to the comment above, I'm hoping this special "ALL" case can eventually go away if we swap the list legend to checkboxes, but I don't want to do that in this particular PR.

eunicorn (approved):

ui/components/src/model/legend.ts
cndonovan, line 42: I wouldn't be opposed to calling this `SelectedLegendItems` type, but I think this is fine too!

cndonovan (approved):
Started to review, but saw your merge. Just leaving the one comment I had, but no big deal really.

PR #1134, opened 2023-05-01
[WIP] Prototype of ECharts time axis rebuild to fix x axis bucketing

sjcobb:
Allows x axis to be rounded to more intuitive values than before by using ECharts 'time' axis type instead of 'category'

## TODO

- [ ] Sync branch and fix merge conflicts from table legend PRs and tooltip / highlighting PRs
- [ ] Plan for a gradual roll out (feature flag in our app's plugin registry or a duplicate ExperimentalTimeSeriesChart?)
  - Could even be a URL param to switch between the legacy 'category' implementation and new 'time'
- [ ] Fix timezones
  - hopefully this approach will work: https://github.com/apache/echarts/issues/14453#issuecomment-800163920
- [ ] Fix thresholds convertPercentThreshold
- [ ] Fix types in new TimeChartTooltip getNearbySeries function
- [ ] Confirm no regressions in null values logic
- [ ] Cleanup components/src/model/graph.ts types
- [ ] Additional stories and docs for legacy LineChart and new TimeChart

### Last 1 hour
image

### Last 7 days
image

## Performance Benchmarks
| Total series | Category axis (ms)| Time axis (ms)|
| ------------| -------------------| -------|
| 1000            | 1707, 2083, 2239   | 1378, 1679, 1707, 1364 |
| 2000            | 3734, 3860, 4082 | 2949, 2758  |
| 3000            | 5830, 5872, 5744  | 5516, 5019, 4989 |
| 4000            | 8828, 8856, 8707  | 8202, 7621, 7616  |

PR #1181, opened 2023-05-22
[FEATURE] Tooltip bold multi series when close to cursor, fix pinning and add show all toggle

sjcobb:
## Summary

- Bold multiple series in tooltip when they are close to cursor
- Fix pinned tooltip by removing onMouseLeave unpin
- Add pin icon and help text inside tooltip so click to pin is more discoverable
- Add 'Show All Series' switch after entering pinned tooltip
- Refactor get focused and emphasized series logic
- Improve highlighting flicker
- Improve onBlur opacity fadeout

This PR is a follow up to #1112 and #1180 as I work to improve line highlighting and correlating which line corresponds to which series in the tooltip (especially when many series are returned in query results)

### Demo:
https://www.loom.com/share/9aa6fbd6edaa429089b860420fe1cc7c

## Recording / Screenshots

https://github.com/perses/perses/assets/9369625/63b2240a-1cfd-4010-a304-6a2413e750d6

Pinned tooltip with new 'Show All Series' toggle
image

After clicking 'Show All Series':
image

Highlighting and multi-emphasized series
image

Show all series - dark mode
image

# Checklist

## UI Changes


juliepagano:
The interactions for the pinned tooltip are a little weird to me. Whenever I click on the chart, the tooltip disappears and then reappears pinned if I move my cursor. I'm not sure if I'm doing something wrong or if there's something here that needs a little adjusting.

https://github.com/perses/perses/assets/396962/f4d8415a-6141-4ca0-a57b-6943f801fa5f

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -95,7 +95,7 @@ export function LineChart({
   const chartsTheme = useChartsTheme();
   const chartRef = useRef<EChartsInstance>();
   const [showTooltip, setShowTooltip] = useState<boolean>(true);
-  const [pinTooltip, setPinTooltip] = useState<boolean>(false);
+  const [tooltipPinned, setTooltipPinned] = useState<boolean>(false);
```
cndonovan, line 98: Could do `isTooltipPinned` if you like having the `is` / some verb in front of booleans (I forget it we have style guide around that, I think I had that at past company)

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -104,7 +104,7 @@ export function LineChart({
         if (onDataZoom === undefined) {
           setTimeout(() => {
             // workaround so unpin happens after click event
-            setPinTooltip(false);
+            setTooltipPinned(false);
```
cndonovan, line 107: Does this need the check for the `e` being an instance of a canvas?

ui/components/src/TimeSeriesTooltip/SeriesInfo.tsx
```diff
@@ -77,21 +78,23 @@ export function SeriesInfo(props: SeriesInfoProps) {
             maxWidth: TOOLTIP_LABELS_MAX_WIDTH,
             overflow: 'hidden',
             color: theme.palette.common.white,
+            fontWeight: emphasizeText ? theme.typography.fontWeightBold : theme.typography.fontWeightRegular,
             textOverflow: 'ellipsis',
             whiteSpace: wrapLabels ? 'normal' : 'nowrap',
           })}
+          aria-label={emphasizeText ? 'focused series' : 'emphasized series'}
```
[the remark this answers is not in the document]
cndonovan: Might have reversed the ternary (guessing you might have meant `emphasizeText ? 'emphasized series' : ... `)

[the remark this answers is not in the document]
cndonovan: But also similar Q to Julie, what does `focused series` mean?

ui/components/src/TimeSeriesTooltip/focused-series.ts
```diff
@@ -94,28 +103,65 @@ export function getNearbySeries(
...
+  unit,
+  showAllSeries = false,
+}: {
+  mousePos: CursorData['coords'];
+  chartData: EChartsDataFormat;
+  pinnedPos: CursorData['coords'];
+  chart?: EChartsInstance;
+  unit?: UnitOptions;
+  showAllSeries?: boolean;
+}) {
```
cndonovan, line 164: Such a minor thing, could put `chart` and `chartData` near each other since they are related ECharts things

ui/components/src/TimeSeriesTooltip/TimeSeriesTooltip.tsx
```diff
@@ -92,7 +113,31 @@ export const TimeSeriesTooltip = React.memo(function TimeSeriesTooltip({
           transform: cursorTransform,
         }}
       >
-        <TooltipContent focusedSeries={focusedSeries} wrapLabels={wrapLabels} />
+        <TooltipContent
+          focusedSeries={focusedSeries}
```
cndonovan, line 117: I might be wrong about this, but I'm thinking we could just call this prop `series` and pass in the `series={focusedSeries}`

If I'm getting it right, from `TooltipContent`'s perspective, it's receiving the list of series to render, and it will render whatever it gets.

Here in the parent, we are choosing to only pass in the `focusedSeries`.

ui/components/src/TimeSeriesTooltip/focused-series.ts
```diff
@@ -144,15 +190,8 @@ export function getFocusedSeriesData(
...
-  const seriesNum = chartData.timeSeries.length;
-
-  // tooltip trigger area gets smaller with more series, increase yAxisInterval multiplier to expand nearby series range
-  const yBuffer =
-    seriesNum > SHOW_FEWER_SERIES_LIMIT
-      ? yAxisInterval * REDUCE_FOCUSED_SERIES_MULTIPLIER
-      : yAxisInterval * INCREASE_FOCUSED_SERIES_MULTIPLIER;
-
+  const totalSeries = chartData.timeSeries.length;
+  const yBuffer = getYBuffer({ interval: yAxisInterval, totalSeries, showAllSeries });
```
cndonovan, line 194: I'm not opposed to calling the argument `yAxisInterval` or `yInterval` (just to make the connection to the y-axis obvious, I can see myself wondering what `interval` means by itself)

up to you!

cndonovan (commented):
Left some comments as overall thoughts, it's bit big for me to feel ready to stamp but if you need a stamp lemme know (I can take another closer look)

Gladorme:
Wouldn't be better to have "Show all" switch at the top of the tooltip (at the left of "Pin it") and be able to enable & disable it? 😄

ui/components/src/TimeSeriesTooltip/utils.ts
```diff
@@ -43,7 +42,7 @@ export function assembleTransform(
 
   // adjust so tooltip does not get cut off at bottom of chart
   if (mousePos.client.y + tooltipHeight + cursorPaddingY > window.innerHeight) {
-    y = mousePos.page.y - tooltipHeight;
+    y = mousePos.page.y - tooltipHeight * 0.75;
```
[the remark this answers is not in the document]
sjcobb: It just felt like the tooltip was getting overly adjusted when you cursor is close to the bottom of the viewport. So this slightly moves it down so it doesn't get close to the top of the viewport. I added a named constant and a comment with more details!

### Before:
image

### After:
image

ui/components/src/LineChart/LineChart.tsx
sjcobb: yeah good question, so this one is attached to ECharts' `datazoom` [event](https://echarts.apache.org/en/api.html#events.datazoom) so we don't have to explicitly check for HTMLCanvasElement like the others that are regular React mouse events

sjcobb: For a bit more context this `setTooltipPinned(false);` is only relevant when an onDataZoom handler isn't passed in and ECharts' zoom event is called (without updating the runtime time range from the panel). I'm not sure if anyone still really uses the chart that way, it's leftover from before we implemented TimeRangeProvider, but doesn't hurt to leave in just in case

ui/components/src/TimeSeriesTooltip/focused-series.ts
```diff
@@ -94,28 +103,65 @@ export function getNearbySeries(
...
               }
             }
           }
         }
       }
     }
   }
+  if (chart?.dispatchAction !== undefined) {
+    // Clears emphasis state of all lines that are not focused or emphasized.
+    // Focused is term for all nearby series showing in the tooltip.
```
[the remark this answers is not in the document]
sjcobb: Did a find & replace so everywhere 'focused' was used it's 'nearby' now!

sjcobb:
> The interactions for the pinned tooltip are a little weird to me. Whenever I click on the chart, the tooltip disappears and then reappears pinned if I move my cursor.

Yeah I [fixed the main bug](https://github.com/perses/perses/pull/1181/commits/2ad56d66d07312a2a12999ff9e6b20bfe1ab0467) where onMouseLeave was setting the pinned tooltip boolean to false way too frequently. Definitely other areas we can improve on, but hopefully that can be in follow-up PRs?

sjcobb:
> Wouldn't be better to have "Show all" switch at the top of the tooltip (at the left of "Pin it") and be able to enable & disable it? 😄

Yeah I like this idea, it'll take a bit of refactoring to TimeSeriesTooltip.tsx and TooltipContent.tsx to do properly, I'll try to come back to it soon though in a follow up PR?

ui/components/src/TimeSeriesTooltip/TimeSeriesTooltip.tsx
```diff
@@ -11,67 +11,88 @@
...
 
-  // ensure user is hovering over a chart before checking for nearby series
+  // Ensure user is hovering over a chart before checking for nearby series.
   if (pinnedPos === null && (mousePos.target as HTMLElement).tagName !== 'CANVAS') return null;
 
   const chart = chartRef.current;
-  const focusedSeries = getFocusedSeriesData(mousePos, chartData, pinnedPos, chart, unit);
-  const chartWidth = chart?.getWidth() ?? 750;
-  const cursorTransform = assembleTransform(
+  const chartWidth = chart?.getWidth() ?? FALLBACK_CHART_WIDTH; // Fallback width not likely to every be needed.
```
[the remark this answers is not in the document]
sjcobb: Yeah this is just to make TS happy, the chart should always have a width by this point or there are bigger problems and the whole parent component is probably broken

ui/components/src/TimeSeriesTooltip/focused-series.ts
sjcobb, line 1: I just did a find & replace everywhere for renaming **focused** -> **nearby**, ended up losing the diff in the process 😞 

It looks like I completely rewrote getNearbySeriesData and checkforNearbySeries, when really it was just moving around the dispatch action calls and adjusting yBuffer params / splitting out the logic into separate utils

ui/components/src/TimeSeriesTooltip/nearby-series.ts
```diff
@@ -0,0 +1,261 @@
...
+  isClosestToCursor: boolean;
+}
+
+export type NearbySeriesArray = NearbySeriesInfo[];
+
+/**
+ * Returns formatted series data for the points that are close to the user's cursor
+ * Adjust yBuffer to increase or decrease number of series shown
+ */
+export function checkforNearbySeries(
```
sjcobb, line 41: See [other comment](https://github.com/perses/perses/pull/1181/files#r1204737458), this file was renamed, I didn't actually rewrite both of these functions.

Definitely areas to improve here, adding more early returns is one area I'm planning to look into in the future so there isn't so much nested if conditions

ui/components/src/TimeSeriesTooltip/SeriesInfo.tsx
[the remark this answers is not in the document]
sjcobb: Fixed this ternary, thanks for catching! I agree "nearby" is a better name than focused so made that change everywhere and added more details in this aria-label

ui/components/src/TimeSeriesTooltip/focused-series.ts
```diff
@@ -52,17 +53,13 @@ export function getNearbySeries(
...
     for (let seriesIdx = 0; seriesIdx < data.timeSeries.length; seriesIdx++) {
       const currentSeries = data.timeSeries[seriesIdx];
-      // TODO: look into using batch or excludeSeriesId within downplay action to fix flicker
-      if (chart?.dispatchAction !== undefined) {
-        // clears emphasis state of lines that are not focused
-        chart.dispatchAction({
-          type: 'downplay',
-          seriesIndex: seriesIdx,
-        });
-      }
```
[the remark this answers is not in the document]
sjcobb: There is still flicker but changing from dispatching separate actions inside of the for loop and instead doing [a single chart.dispatchAction](https://github.com/perses/perses/blob/a4afe734a0e548f6525cbfa8c53da53618c485ae/ui/components/src/TimeSeriesTooltip/nearby-series.ts#L128-L132) for 'highlight' with the entire `emphasizedSeriesIndexes` array at once really helps. 

This was a big problem in the last PR #1181 since I didn't realize you could pass an array that way. Now `blur` in [getLineSeries](https://github.com/perses/perses/blob/a4afe734a0e548f6525cbfa8c53da53618c485ae/ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts#L96-L108) also works properly.

Long term I'm planning to dig more into this as part of the TimeChart.tsx refactor: https://github.com/perses/perses/pull/1134

In that branch using ECharts' 'time' axis, you can also define an "xBuffer" param which will potentially make the experience when moving your cursor horizontally a bit more smooth

ui/components/src/TimeSeriesTooltip/focused-series.ts
juliepagano: I hate when git does that. :(

juliepagano:
> > The interactions for the pinned tooltip are a little weird to me. Whenever I click on the chart, the tooltip disappears and then reappears pinned if I move my cursor.
> 
> Yeah I [fixed the main bug](https://github.com/perses/perses/pull/1181/commits/2ad56d66d07312a2a12999ff9e6b20bfe1ab0467) where onMouseLeave was setting the pinned tooltip boolean to false way too frequently. Definitely other areas we can improve on, but hopefully that can be in follow-up PRs?

It's not a regression, so fixing it in a follow-up seems reasonable to me!

juliepagano:
> > Wouldn't be better to have "Show all" switch at the top of the tooltip (at the left of "Pin it") and be able to enable & disable it? 😄
> 
> Yeah I like this idea, it'll take a bit of refactoring to TimeSeriesTooltip.tsx and TooltipContent.tsx to do properly, I'll try to come back to it soon though in a follow up PR?

Related design feedback for a follow-up: I could also see it being helpful to keep the header fixed when you scroll down large tooltips, so you don't lose the context and interactive bits of the header.

image

It might also be worth checking in with @foxbat07 about the general design of the tooltip to see if he has some additional thoughts there.

ui/components/src/TimeSeriesTooltip/nearby-series.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  if (cursorX === null || cursorY === null) {
+    return currentNearbySeriesData;
+  }
+
+  const nearbySeriesIndexes: number[] = [];
+  const emphasizedSeriesIndexes: number[] = [];
+  const nonEmphasizedSeriesIndexes: number[] = [];
+  const totalSeries = data.timeSeries.length;
+  if (Array.isArray(data.xAxis) && Array.isArray(data.timeSeries)) {
```
juliepagano, line 60: Not for you to fix in this PR because I know I'm only seeing this because git did a rename, but this deeply nested set of loops and conditionals is kinda hard to read. It sounds like you might touch this again for some follow-ups. If so, it'd be awesome if this could potentially get broken out into some smaller methods that are a little easier to read & maintain.

juliepagano (approved):
LGTM! Left one optional comment for potential future you improving this code more.

ui/components/src/TimeSeriesTooltip/nearby-series.ts
sjcobb: Yeah for sure, planning to add more early returns for the nested if's, separating out utils, and adding more unit tests!

I'd love to write an e2e for tooltip behavior too instead of just the RTL TooltipContent tests, I briefly tried before but I need to level up my Playwright knowledge before trying again

PR #1189, opened 2023-05-23
[ENHANCEMENT] Export types that are useful for working with core panels

sjcobb:
Just realized that we don't export out these basic panel option types, these are useful for consumers.

# Checklist

## UI Changes


ui/panels-plugin/src/plugins/time-series-chart/index.ts
```diff
@@ -12,4 +12,10 @@
 // limitations under the License.
 
 export * from './TimeSeriesChart';
-export type { TimeSeriesChartDefinition } from './time-series-chart-model';
+export type {
+  TimeSeriesChartDefinition,
+  TimeSeriesChartOptions,
+  YAxisOptions,
+  PaletteOptions,
+  VisualOptions,
```
juliepagano, line 20: Similar comment as above about giving these more verbose/prefixed names for clarity and to avoid conflicts with other panels.

ui/panels-plugin/src/plugins/stat-chart/index.ts
```diff
@@ -12,4 +12,4 @@
 // limitations under the License.
 
 export * from './StatChart';
-export type { StatChartDefinition } from './stat-chart-model';
+export type { StatChartDefinition, StatChartOptions, SparklineOptions } from './stat-chart-model';
```
juliepagano, line 15: Would it be worth renaming `SparklineOptions` to something like `StatChartSparklineOptions` to avoid conflict if we build a more specific sparkline-focused visualization in the future or do you think they'd have sufficient overlap that it's not worth worrying about it?

juliepagano:
Good catch! We have a bunch of cases like this. Thanks for tackling cleaning it up.

ui/panels-plugin/src/plugins/stat-chart/index.ts
cndonovan: Hm, that does seem more organized in a way, I wouldn't be opposed to adding prefixes like that. 

The cost is the verbosity... but it does kind of future-proof the types / make our types more organized/robust, which is nice. 

I'm good with either way, but I see the reason to think about it.

cndonovan (approved):
When I started exporting from these model files, I figured it would be nice to be explicit about what we're exporting, so I started by explicitly exporting the `XChartDefinition`. We *could* export `*` from these files and get all these types in a single export, but if we did that, we would also be exporting all the `const` variables too. Just calling that out in case we want to switch. I'm still good with the explicit approach. I think I like that each time we need to do an export, it forces us to think if we really want to export it or not. 

Also, I like Julie's idea about the prefixes, but up to you what seems best.

saminzadeh (approved):
Slight preference to add prefixes to the names, but either way I'm good. I'll let you make the call

ui/panels-plugin/src/plugins/time-series-chart/index.ts
sjcobb: I imagine we will be able to share some of these with other panel types in the future, in particular `PaletteOptions` could be common with the future potential BarChart. Adding the prefixes though we can revisit, at that point we'd probably want to move them to @perses-dev/core anyways

PR #1190, opened 2023-05-23
[BUGFIX] Fix Perses build for downstream consumers

eworstell:
Downstream consumers aren't able to depend on Perses because asset files referenced in the `/ui` package aren't supplied unless a `make build` has been run to generate them:

Screen Shot 2023-05-23 at 3 49 58 PM


Nexucis:
I don't know if you are trying to import some perses lib in a go application or if you are trying to build perses on your own side, but:

- if you want to import some libs, you should only use what is contained in `pkg` that's where we are putting our "public" lib.
- if you are trying to build Perses on your side, then you should considerate to generate the go files before trying to build it, because it's not the only one file generated. And you can be sure we won't remove the generation of these files, specially since the files with `embed` doesn't have the same name each time you are building it.

sjcobb:
> * if you want to import some libs, you should only use what is contained in `pkg` that's where we are putting our "public" lib.

I'll let Lizzy go into the specifics of what else we'd like to be public, but is there potential to expose more in the "public" API @Nexucis? 

Or even potential to publish a go module for consumers interested in the embed use-case? We're thinking about trying to hit the `/validate` endpoint from our UI since we have no other way to utilize the CUE schema right now. 

If that doesn't work out, we probably will need to look into adding another solution like [zod](https://github.com/colinhacks/zod) for client-side validation

eworstell:
> > I'll let Lizzy go into the specifics of what else we'd like to be public, but is there potential to expose more in the "public" API @Nexucis? We're thinking about trying to hit the /validate endpoint from our UI since we have no other way to utilize the CUE schema right now.
> 
> That's another question than what it is tackling here.
> 
> Here I still don't get what part of the build of Perses doesn't work. Still the files that need to be generated are generated on purpose and you shouldn't need to depend on it.
> 
> Regarding the fact you would like to use what is behind the `/validate` endpoint you can still import it even if it's not in the `pkg` path. It is more a convention rather than something that would block you from importing it.
> 
> For example, if you are looking at the `CLI`, you will see it is able to use the CUE file locally or it can use the `validate` endpoint if you are more into a online mode.
> 
> https://github.com/perses/perses/blob/main/internal/cli/cmd/lint/lint.go#L93-L143
> 
> > We're thinking about trying to hit the /validate endpoint from our UI since we have no other way to utilize the CUE schema right now.
> 
> You could simply run an instance of Perses that would be a cheap solution. Like that the endpoint will be available. Or you are doing what the CLI is doing: reading the cue files and use them to validate the dashboards locally

The goal of this PR is to unblock our ability to add the "github.com/perses/perses" module as a dependency in another Go module. 

We're deploying a Perses image to a container and plan to query this instance from a separate service (specifically, as @sjcobb mentioned, the `/validate` endpoint). In this other service, we want to add a Perses Go dependency to provide the contents of "github.com/perses/perses/pkg/client" and "github.com/perses/perses/pkg/model". This will allow this other service to import the Perses-defined API schemas for its client layer with Perses. In its current state, adding this dependency breaks the Go build because the UI package cannot build without assets. 

The original approach in this PR was an attempt to track everything the Go build needs — as you mentioned this didn't work because the compressed asset files are not stable. I've updated this draft with a different approach. Rather than leaving the Go module as-is and including additional files needed by the compiler, this approach updates the module structure so that the offending package (`/ui`) is entirely excluded; the `/ui` directory is separated into its own Go module. 

This prevents the contents of `/ui` from being included in the "github.com/perses/perses" module. Those interested in deploying the Perses UI can do so using the complete source code, either via the Perses Docker release or by cloning locally (the process remains unchanged). The "github.com/perses/perses" Go module is now isolated to contents that can be built by a Go compiler (thus unblocking the module from being used as a dependency).

Unfortunately there's still an issue — tests in `pkg/client` import the `e2eframework` package, which import the `ui` package via `core`. While this doesn't break the Go build (because all these test files are annotated with the `//go:build integration` constraint), it does break vendoring when importing "github.com/perses/perses/pkg/client" because the dependency chain can't be resolved when `/ui` is excluded from the root Go module.

This PR fixes the import/vendoring issue for `pkg/client` by moving the e2e tests from `pkg/client` into `internal/api/e2e`. But the Go build for the root module will now break when passing the integration compiler flag (which will attempt to build the files otherwise ignored by `go:build integration` constraint). This means the Go build will be broken when running in the `e2e` action pipeline.

I didn't want to go further here without some opinions from Perses people, but one fix to this could be moving `e2e` into a module of its own that adds the `github.com/perses/perses/ui` module as a dependency. The the root module would not need to import either `e2e` or `ui` modules.

Nexucis:
I also tried with a new go project and I didn't face the issue you are mentioning.

`go.mod`

```
module github.com/nexucis/pocperses

go 1.20

require github.com/perses/perses v0.31.0

require (
	cuelang.org/go v0.5.0 // indirect
	github.com/ajg/form v1.5.1 // indirect
	github.com/andybalholm/brotli v1.0.4 // indirect
	github.com/beorn7/perks v1.0.1 // indirect
	github.com/cespare/xxhash/v2 v2.2.0 // indirect
	github.com/cockroachdb/apd/v2 v2.0.2 // indirect
	github.com/davecgh/go-spew v1.1.1 // indirect
	github.com/emicklei/proto v1.10.0 // indirect
	github.com/fatih/color v1.13.0 // indirect
	github.com/fatih/structs v1.1.0 // indirect
	github.com/fsnotify/fsnotify v1.6.0 // indirect
	github.com/gavv/httpexpect/v2 v2.15.0 // indirect
	github.com/go-logr/logr v1.2.3 // indirect
	github.com/go-logr/stdr v1.2.2 // indirect
	github.com/go-sql-driver/mysql v1.7.1 // indirect
	github.com/gobwas/glob v0.2.3 // indirect
	github.com/golang-jwt/jwt v3.2.2+incompatible // indirect
	github.com/golang/glog v1.0.0 // indirect
	github.com/golang/protobuf v1.5.3 // indirect
	github.com/google/go-querystring v1.1.0 // indirect
	github.com/google/uuid v1.3.0 // indirect
	github.com/gorilla/websocket v1.5.0 // indirect
	github.com/huandu/go-sqlbuilder v1.21.0 // indirect
	github.com/huandu/xstrings v1.3.2 // indirect
	github.com/imkira/go-interpol v1.1.0 // indirect
	github.com/jpillora/backoff v1.0.0 // indirect
	github.com/json-iterator/go v1.1.12 // indirect
	github.com/klauspost/compress v1.16.3 // indirect
	github.com/labstack/echo/v4 v4.10.2 // indirect
	github.com/labstack/gommon v0.4.0 // indirect
	github.com/mattn/go-colorable v0.1.13 // indirect
	github.com/mattn/go-isatty v0.0.17 // indirect
	github.com/matttproud/golang_protobuf_extensions v1.0.4 // indirect
	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
	github.com/modern-go/reflect2 v1.0.2 // indirect
	github.com/mpvl/unique v0.0.0-20150818121801-cbe035fff7de // indirect
	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
	github.com/nexucis/lamenv v0.5.0 // indirect
	github.com/nxadm/tail v1.4.8 // indirect
	github.com/perses/common v0.20.0 // indirect
	github.com/pkg/errors v0.9.1 // indirect
	github.com/pmezard/go-difflib v1.0.0 // indirect
	github.com/prometheus/client_golang v1.15.1 // indirect
	github.com/prometheus/client_model v0.4.0 // indirect
	github.com/prometheus/common v0.44.0 // indirect
	github.com/prometheus/common/assets v0.2.0 // indirect
	github.com/prometheus/procfs v0.9.0 // indirect
	github.com/protocolbuffers/txtpbfmt v0.0.0-20220428173112-74888fd59c2b // indirect
	github.com/sanity-io/litter v1.5.5 // indirect
	github.com/sergi/go-diff v1.2.0 // indirect
	github.com/sirupsen/logrus v1.9.2 // indirect
	github.com/stretchr/testify v1.8.3 // indirect
	github.com/valyala/bytebufferpool v1.0.0 // indirect
	github.com/valyala/fasthttp v1.34.0 // indirect
	github.com/valyala/fasttemplate v1.2.2 // indirect
	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
	github.com/xeipuuv/gojsonschema v1.2.0 // indirect
	github.com/yalp/jsonpath v0.0.0-20180802001716-5cc68e5049a0 // indirect
	github.com/yudai/gojsondiff v1.0.0 // indirect
	github.com/yudai/golcs v0.0.0-20170316035057-ecda9a501e82 // indirect
	go.opentelemetry.io/otel v1.14.0 // indirect
	go.opentelemetry.io/otel/sdk v1.14.0 // indirect
	go.opentelemetry.io/otel/trace v1.14.0 // indirect
	golang.org/x/crypto v0.8.0 // indirect
	golang.org/x/net v0.10.0 // indirect
	golang.org/x/oauth2 v0.8.0 // indirect
	golang.org/x/sys v0.8.0 // indirect
	golang.org/x/text v0.9.0 // indirect
	golang.org/x/time v0.3.0 // indirect
	google.golang.org/appengine v1.6.7 // indirect
	google.golang.org/protobuf v1.30.0 // indirect
	gopkg.in/yaml.v2 v2.4.0 // indirect
	gopkg.in/yaml.v3 v3.0.1 // indirect
	moul.io/http2curl/v2 v2.3.0 // indirect
)
```

`main.go`

```go
package main

import (
	"fmt"

	clientV1 "github.com/perses/perses/pkg/client/api/v1"
	"github.com/perses/perses/pkg/client/perseshttp"
	v1 "github.com/perses/perses/pkg/model/api/v1"
)

func main() {
	entity := &v1.GlobalDatasource{}
	restClient, err := perseshttp.NewFromConfig(perseshttp.RestConfigClient{})
	if err != nil {
		panic(err)
	}
	client := clientV1.NewWithClient(restClient)
	fmt.Println("Hello, 世界", entity, client)
}
```

Nexucis:
while I was looking at my `go.mod` (used for the poc), I was wondering if moving the e2e tests outside of `pkg` would reduce drastically all these `indirect` dependencies that are mostly coming and used for the server and not for the `pkg`.

And it seems yes it reduced it and it is definitely cleaner: 

```
module github.com/nexucis/pocperses

go 1.20

require github.com/perses/perses v0.31.0

replace github.com/perses/perses => github.com/nexucis/perses v0.0.0-20230531095007-718efebef4e9

require (
        github.com/json-iterator/go v1.1.12 // indirect
        github.com/kr/pretty v0.3.1 // indirect
        github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
        github.com/modern-go/reflect2 v1.0.2 // indirect
        github.com/prometheus/common v0.44.0 // indirect
        gopkg.in/yaml.v2 v2.4.0 // indirect
)
```

Nexucis:
I'm closing this PR. The e2e tests have been moved outside of the `pkg` package. Like that the list of import will be tinier and more accurate.

If you feel there is still something broken, please let me know

PR #1195, opened 2023-05-25
[WIP] Virtualize tooltip, move show all and make date header sticky, improve pin logic

sjcobb:
## Summary

- [ENHANCEMENT] Tooltip add react-virtuoso to improve scrolling performance
- [ENHANCEMENT] Move show all toggle to tooltip header beside pin
- [BUGFIX] Tooltip does not flicker after clicking to pin
- [ENHANCEMENT] Add fixed tooltip header while scrolling many series

## Screenshots

image

image

# Checklist

## UI Changes


ui/components/src/LineChart/LineChart.tsx
```diff
@@ -205,20 +209,56 @@ export function LineChart({
...
-        if (e.target instanceof HTMLCanvasElement) {
-          setShowTooltip(false);
+        e.preventDefault();
+        const { clientX } = e;
+        setDragging(true);
+        setStartX(clientX);
+      }}
+      onMouseMove={(e) => {
+        const { clientX } = e;
+        if (isDragging) {
```
sjcobb, line 243: These lines fix the "tooltip disappears and reappears" bug @juliepagano mentioned [here](https://github.com/perses/perses/pull/1181#issuecomment-1558087880) in #1181. Before `setShowTooltip(false);` was being called while the tooltip was being pinned, I've added a useState to confirm whether the user is clicking and dragging their cursor to zoom before calling  `setShowTooltip(false);` (since you don't want the tooltip to get in the way while trying to select the range). 

Also note, the ECharts [datazoom](https://echarts.apache.org/en/api.html#events.datazoom) event above fires after the data range is selected, there may be other built-in events/actions we can leverage but doing it purely in React seems to work pretty well 🤷‍♂️

ui/components/src/TimeSeriesTooltip/TooltipContent.tsx
```diff
@@ -12,107 +12,89 @@
...
-        }}
-      >
-        {sortedFocusedSeries.map(
-          ({ datumIdx, seriesIdx, seriesName, y, formattedY, markerColor, isClosestToCursor }) => {
-            if (datumIdx === null || seriesIdx === null) return null;
-            const key = seriesIdx.toString() + datumIdx.toString();
+    );
+  }
+
+  // TODO: is there a better way to approximate height or a dynamic height and width workaround for Virtuoso?
```
sjcobb, line 60: @juliepagano this is my first time using react-virtuoso and I'm definitely not doing everything correctly. I tried following some of their examples and the code you wrote for the legend is really helpful, but curious if you can give me some pointers? 🙏 

Currently, I'm not sure the best way to set the `style={{ height: contentHeight, width: TOOLTIP_MULTI_SERIES_MIN_WIDTH }}` lines below. I'm approximating based on the number of series, which kind of works for longer series names, but results in weirdness like here:
image

I thought about improving the approximation by checking the number of characters in each series name to find the longest (and use that to infer how many lines it wraps to), but that seems hacky and bad for performance. I did try some workarounds after Googling "react-virtuoso dynamic height and width" but they didn't seem to work, it does seem like the outer container needs a static height defined and since this changes as you move your cursor when number of nearby series changes, I'm not how to address this 😕

sjcobb: Figured out an approach that works with useResizeObserver, closing this PR since I’m going to split it up into multiple PRs so it’s easier to review

PR #1196, opened 2023-05-25
BREAKING CHANGE: extract legend layout to reusable component

juliepagano:
This commit:
- Extracts the layout and spacing of content alongside a legend from the time series panel plugin into a ContentWithLegend component that can be used by both the time series plugin panel and other use cases.
- Adjustments to the styling of the layout and spacing of content alongside a legend and the legend itself to better match guidance from design. Primarily this impacts some margins and padding.

BREAKING CHANGE: this change makes some styling adjustments to the Legend component that existed to support the previous layout in the time series panel plugin. Consumers who were using the Legend outside of the time series panel, may need to adjust their styling to handle the change.

# Checklist

## UI Changes

### DIFF
--- a/ui/components/src/ContentWithLegend/model/content-with-legend-model.ts
+++ b/ui/components/src/ContentWithLegend/model/content-with-legend-model.ts
@@ -1,0 +1,153 @@
+// Copyright 2023 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { LegendProps } from '../../Legend';
+
+type Dimensions = {
+  width: number;
+  height: number;
+};
+
+export interface ContentWithLegendProps {
+  /**
+   * Width of the overall component layout in pixels.
+   */
+  width: number;
+  /**
+   * Height of overall component layout in pixels.
+   */
+  height: number;
+  /**
+   * Child content to render next to the legend. May be a react node or a
+   * function that returns a react node. The function provides the expected
+   * height and width for the content, which can be useful for passing down
+   * to chart components.
+   */
+  children: React.ReactNode | (({ width, height }: Dimensions) => React.ReactNode);
+  /**
+   * Props to configure the legend. If not set, the content is rendered without
+   * a legend.
+   */
+  legendProps?: Omit<LegendProps, 'width' | 'height'>;
+  /**
+   * Space to put between the children and the legend in pixels.
+   */
+  spacing?: number;
+
+  /**
+   * Minimum width required for the content specified by the `children` prop.
+   * If this width cannot be maintained with a right positioned legend, the
+   * legend will not be shown.
+   */
+  minChildrenWidth?: number;
+
+  /**
+   * Minimum height required for the content specified by the `children` prop.
+   * If this width cannot be maintained with a bottom positioned legend, the
+   * legend will not be shown.
+   */
+  minChildrenHeight?: number;
+}
+
+export interface ContentWithLegendLayoutOpts
+  extends Required<Omit<ContentWithLegendProps, 'children' | 'legendProps'>> {
+  legendOptions?: LegendProps['options'];
+}
+
+export interface ContentWithLegendLayout {
+  legend: Dimensions & {
+    show: boolean;
+  };
+  content: Dimensions;
+  margin: {
+    right: number;
+    bottom: number;
+  };
+}
+
+const PANEL_HEIGHT_LG_BREAKPOINT = 300;
+const LEGEND_HEIGHT_SM = 40;
+const LEGEND_HEIGHT_LG = 100;
+
+/**
+ * Returns information for laying out content alongside a legend.
+ */
+export function getContentWithLegendLayout({
+  width,
+  height,
+  legendOptions,
+  minChildrenHeight,
+  minChildrenWidth,
+  spacing,
+}: ContentWithLegendLayoutOpts): ContentWithLegendLayout {
+  const hasLegend = !!legendOptions;
+
+  const noLegendLayout: ContentWithLegendLayout = {
+    legend: {
+      show: false,
+      width: 0,
+      height: 0,
+    },
+    content: {
+      width,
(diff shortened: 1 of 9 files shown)


ui/components/src/ContentWithLegend/model/content-with-legend-model.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  if (!hasLegend) {
+    return noLegendLayout;
+  }
+
+  const { position } = legendOptions;
+
+  const legendWidth = position === 'Right' ? 200 : width;
+
+  // TODO: account for number of legend items returned when adjusting legend spacing
```
juliepagano, line 119: This is a carry-over TODO from the original logic in the time series panel. 

The majority of the logic in here is also the same as what's in there, but refactored for readability.

ui/components/src/ContentWithLegend/model/content-with-legend-model.ts
```diff
@@ -0,0 +1,153 @@
...
+  const contentWidth = position === 'Right' ? width - legendWidth - spacing : width;
+  const contentHeight = position === 'Bottom' ? height - legendHeight - spacing : height;
+
+  if (
+    (position === 'Right' && contentWidth < minChildrenWidth) ||
+    (position === 'Bottom' && contentHeight < minChildrenHeight)
+  ) {
+    // Legend does not fit. Just show the content.
+    return noLegendLayout;
+  }
```
juliepagano, line 136: This logic is new to be a little more responsive.

ui/components/src/Legend/ListLegend.tsx
```diff
@@ -31,14 +30,6 @@ export interface ListLegendProps {
  * large numbers of items when there is a single item per row.
  */
 export function ListLegend({ items, height, width, selectedItems, onLegendItemClick }: ListLegendProps) {
-  const theme = useTheme();
```
juliepagano, line 34: We don't need the mock padding with the new layout that more closely matches designs. However, this change could theoretically impact consumers using this directly, so I marked this PR as a breaking change.

ui/panels-plugin/src/plugins/time-series-chart/time-series-chart-model.ts
```diff
@@ -136,10 +136,6 @@ export const STACK_OPTIONS = Object.entries(STACK_CONFIG).map(([id, config]) =>
   };
 });
 
-export const PANEL_HEIGHT_LG_BREAKPOINT = 300;
```
juliepagano, line 139: These constants were moved to `content-with-legend-model.ts`

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -258,22 +251,12 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
-    legendHeight = LEGEND_HEIGHT_LG;
-  }
-
   // override default spacing, see: https://echarts.apache.org/en/option.html#grid
   const gridLeft = y_axis && y_axis.label ? 30 : 20;
   const gridOverrides: GridComponentOption = {
     left: !echartsYAxis.show ? 0 : gridLeft,
-    right: legend && legend.position === 'Right' ? legendWidth : 20,
-    bottom: legend && legend.position === 'Bottom' ? legendHeight : 0,
+    right: 20,
```
sjcobb, line 258: So ECharts grid doesn’t need to be adjusted like it was previously? I think I did it that way so that in panel edit drawer when you click to change between Bottom and Right it doesn’t cause a lag / weird jump while the padding was adjusting (might not be a problem anymore with all the other recent changes)

sjcobb (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
juliepagano: Lemme give that use case a double-check to make sure it looks ok with the new layout.

juliepagano: Here's a video of what it looks like in the next WIP PR where I actually add the options.

 Is the little period where there's weird overlap with the list legend what you're talking about? I think you don't see it with the table legend because it doesn't have a transparent background, which is something I could fix about the list legend.

https://github.com/perses/perses/assets/396962/ce57a9a5-032c-4a2b-9839-cc71057a92a4

juliepagano: Putting `overflow: hidden` on the content wrapper in `ContentWithLegend` seems to help this a bunch. I pushed up a change with that and story for the component that exercises it. How do you feel about this?


https://github.com/perses/perses/assets/396962/ea9bd6ce-ed4d-4988-a7d7-371b7edeb502

juliepagano: Digging a little mroe, I think the debouncing in the resize in `EChart` is what causes that delay. I'll take a closer look in the morning to see if there's a clever way to disable the debounce timeout for these one-time changes like a position swap while keeping it for cases like dragging to resize where you definitely want to debounce.

sjcobb: ooh interesting, we should talk about that EChart wrapper more sometime! I think long term we're going to need a hybrid of the approach tracing used internally in their wrapper and what we've built so far

And that video looks great 🙌

juliepagano: I was nerd sniped and poked at this some more after dinner. There's a bunch of issues with the debouncing logic in `EChart.tsx` that I don't think is actually doing what we want it to. I want to circle back to fixing it, but NOT in this PR, since it's a complex change that will benefit from more thorough, separate review. 

In the short-term, what I did change is making the `useEffect` debounce run on `leading` (i.e. it doesn't wait to run the first time) instead of the default `trailing` (it would wait 200ms for the first run, causing that lag you saw). I also left some comments in the related code annotating what I think is wrong with it. In the morning, I'll make a more detailed ticket around this that I'll try to get to after I get this initial round of MVP table legend PRs up. 

Does that sound ok to you?

juliepagano: Also, here's a video showing what it looks like after that switch to `leading. Note that the toggle is nice and snappy now. I also show a resize of a panel, which is theoretically where we want the debouncing to work and it seems the same as before, so I'm not harming anything with the change.


https://github.com/perses/perses/assets/396962/0cce61aa-d8fc-41dd-889b-5e7a80621337

Issue #1173, opened 2023-05-15
Switching between dark and light theme should not reload dashboard


sjcobb:
Theme toggle currently causes all queries to fire again

sjcobb:
> Theme toggle cause a complete page reload. Because of this [line](https://github.com/perses/perses/blob/28aa09f49312df68750926c40ac68cf3de2736ae/ui/app/src/context/DarkMode.tsx#L46)

Yeah I remember [trying to remove that line](https://github.com/perses/perses/pull/415#issuecomment-1108909667) and it required some further refactoring. I just never had a chance to go back and look more into it (the entire dark mode implementation could really be simplified)

nicolastakashi:
I would like to work on this @sjcobb

PR #1198, opened 2023-05-26
[BUGFIX] avoiding screen reload on theme toggle

nicolastakashi:
Changing the Theme Provider to not reload the entire screen when switch between themes.

This PR will fix https://github.com/perses/perses/issues/1173

# Checklist

- [X] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.

## UI Changes

- [X] Visual tests are stable and unlikely to be flaky. See [Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests) and [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues include:
  - Does it have loading states? You need to wait for the loading to complete.

## Screenshots

https://github.com/perses/perses/assets/8205546/d7eb91b6-c717-43cc-9d7a-0318b5bd2f12

### DIFF
--- a/ui/app/src/context/DarkMode.tsx
+++ b/ui/app/src/context/DarkMode.tsx
@@ -24,7 +24,7 @@ const DARK_MODE_PREFERENCE_KEY = 'PERSES_ENABLE_DARK_MODE';
 
 interface DarkModeContext {
   isDarkModeEnabled: boolean;
-  setDarkMode: (pref: boolean) => Promise<void>;
+  setDarkMode: (pref: boolean) => void;
 }
 
 export const DarkModeContext = createContext<DarkModeContext | undefined>(undefined);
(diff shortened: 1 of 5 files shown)


Nexucis:
oh nice, thanks @nicolastakashi.
Looks like the e2e tests are broken, but probably it won't be too hard to fix it :)

sjcobb:
> Looks like the e2e tests are broken, but probably it won't be too hard to fix it :)

Odd that those broke, rerunning to see if it was flaky, might need to rebase

ui/app/src/context/DarkMode.tsx
```diff
@@ -24,7 +24,7 @@ const DARK_MODE_PREFERENCE_KEY = 'PERSES_ENABLE_DARK_MODE';
 
 interface DarkModeContext {
   isDarkModeEnabled: boolean;
-  setDarkMode: (pref: boolean) => Promise<void>;
+  setDarkMode: (pref: boolean) => void;
```
sjcobb, line 27: Oh man, I did this in a rush so long ago 🙃
I was bringing over a bunch of code from an existing implementation that stored the theme preference in the db, thanks for fixing

sjcobb (approved):

sjcobb:
The typedoc CI errors will definitely go away after rebasing, Julie disabled in #1199

juliepagano:
I _think_ I figured out why the tests are failing. It looks like the tooltip for the toggle button is blocking most of the edit button, so attempting to click the edit probably fails. This didn't happen previously because the page would reload. You should be able to fix this by telling playwright to mouse away from the toggle, so the tooltip goes away. 

image

Can you try adding a line like after this line https://github.com/perses/perses/blob/main/ui/e2e/src/pages/DashboardPage.ts#L150

```ts
await this.page.mouse.move(-200, -200);
```

eunicorn:
Looks like the other tests are failing because they are trying to look for the edit button again after the theme change since the page used to reload before your changes.

To fix this, you just need to move the actions to outside the `dashboardPage.forEachTheme` function. You can cherry-pick this commit https://github.com/perses/perses/commit/916e49042df5590706a05165f68baf6789ddb502 and it should fix all the e2e tests.

Nexucis:
meh @eunicorn the e2e tests are still timeout :(.

eunicorn:
> meh @eunicorn the e2e tests are still timeout :(.

I don't see the fix in this branch yet. @nicolastakashi can you add the fix to your branch?

nicolastakashi:
@eunicorn can you check if I did it properly?

eunicorn:
yep looks correct, will check again once the tests pass :)

sjcobb (approved):
Going ahead and merging this one, I'll look into the recent e2e failure in a follow up, really appreciate the contribution @nicolastakashi!

PR #1204, opened 2023-05-30
[ENHANCEMENT] Set y-axis min relative to data in stat chart

eunicorn:
Right now y-axis min is calculated by echart, which causes the sparkline in stat chart to be rather flat in some cases. 

To fix this, we are now setting y-axis min to the minimum value of the data.

**BEFORE**
image
image

**AFTER**
image

image

### DIFF
--- a/ui/components/src/StatChart/StatChart.tsx
+++ b/ui/components/src/StatChart/StatChart.tsx
@@ -87,6 +87,13 @@ export function StatChart(props: StatChartProps) {
       yAxis: {
         type: 'value',
         show: false,
+        min: (value: { min: number; max: number }) => {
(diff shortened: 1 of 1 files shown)


ui/components/src/StatChart/StatChart.tsx
```diff
@@ -87,6 +87,13 @@ export function StatChart(props: StatChartProps) {
       yAxis: {
         type: 'value',
         show: false,
+        min: (value: { min: number; max: number }) => {
```
cndonovan, line 90: Could leave off `max` if you want

cndonovan (approved):
😍

sjcobb (approved):

PR #1209, opened 2023-05-31
[IGNORE] Upgrade react-router from 6.3.0 to 6.11.2 

sjcobb:
Temp branch to hopefully upgrade react router without breaking a bunch of stuff 🤞 

Follow up to closed PR #1153 - I think this is safe to upgrade now, finally got RTL tests to pass again

### DIFF
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -34,7 +34,7 @@
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
     "react-error-boundary": "^3.1.4",
-    "react-router-dom": "^6.3.0",
+    "react-router-dom": "^6.11.0",
     "use-immer": "^0.7.0",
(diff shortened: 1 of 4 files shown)


Nexucis:
looks good for me. I'm just wondering why we need the router history for the dashboard tests :o

sjcobb:
> looks good for me. I'm just wondering why we need the router history for the dashboard tests :o

We test query params for TimeRangeControls to make sure `start` is applied correctly [here](https://github.com/perses/perses/blob/v0.31.0/ui/dashboards/src/components/TimeRangeControls/TimeRangeControls.test.tsx#L60-L78), might be better to do that as an e2e in the future, but I did finally get it to work again 😂

Nexucis (approved):

eunicorn (approved):

PR #1210, opened 2023-05-31
[BREAKINGCHANGE] Save dashboard type change for embed gql users

sjcobb:
Not all embed users implement saving dashboards using REST, this type change will allow `SaveDashboardButton` to be used in cases where our consumers use GraphQL

### DIFF
--- a/ui/core/src/model/dashboard.ts
+++ b/ui/core/src/model/dashboard.ts
@@ -38,5 +38,3 @@ export interface DashboardSelector {
   project: string;
   dashboard: string;
 }
-
-export type OnSaveDashboard = (dashboard: DashboardResource) => Promise<unknown>;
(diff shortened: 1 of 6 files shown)


Nexucis:
that sounds like a breaking change for me ;)

sjcobb:
> that sounds like a breaking change for me ;)

Yeah good call! Just updated

ui/core/src/model/dashboard.ts
```diff
@@ -38,3 +38,5 @@ export interface DashboardSelector {
   project: string;
   dashboard: string;
 }
+
+export type OnSaveDashboard = (dashboard: DashboardResource) => Promise<unknown>;
```
Nexucis, line 42: Probably this should be define directly in the dashboard package. That doesn't sound it's part of the model for me.

sjcobb: Yeahh went back and forth on that, moved to dashboards package now, [this file](https://github.com/perses/perses/pull/1210/commits/347f18154cdb8f5878578eb83d59bc0f26d5e235) maybe makes sense and has some utils that don't hurt to export out

eunicorn (approved):

PR #1214, opened 2023-06-01
[BUGFIX] Saving defaults error adjustments and account for Empty StaticListVariable

sjcobb:
## Summary

- Fixes issue where empty StaticListVariable was triggering save default dialog unnecessarily
- Refactor error handling in SaveDashboardButton and change to async await
- RTL test setup adjustments to include SnackbarProvider
- Reduce Jest warnings in dashboards package by adding TimeRangeProvider in PanelDrawer.test.tsx (still a decent number of warnings though)

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
@@ -14,6 +14,7 @@
 import { screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { act } from 'react-dom/test-utils';
+import { TimeRangeProvider } from '@perses-dev/plugin-system';
 import { createDashboardProviderSpy, getTestDashboard, renderWithContext } from '../../test';
 import { DashboardProvider } from '../../context/DashboardProvider';
 import { PanelDrawer } from './PanelDrawer';
@@ -24,8 +25,10 @@ describe('Panel Drawer', () => {
 
     renderWithContext(
       <DashboardProvider initialState={{ dashboardResource: getTestDashboard(), isEditMode: true }}>
-        <DashboardProviderSpy />
-        <PanelDrawer />
+        <TimeRangeProvider initialTimeRange={{ pastDuration: '30m' }}>
+          <DashboardProviderSpy />
+          <PanelDrawer />
+        </TimeRangeProvider>
       </DashboardProvider>
     );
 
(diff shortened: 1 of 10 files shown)


ui/dashboards/src/components/SaveDashboardButton/SaveDashboardButton.tsx
```diff
@@ -64,18 +72,20 @@ export const SaveDashboardButton = ({ onSave, isDisabled, variant = 'contained'
...
+    if (onSave) {
+      try {
+        setSavingDashboard(true);
+        await onSave(dashboard);
+        setSavingDashboard(false);
+        closeSaveChangesConfirmationDialog();
+        setEditMode(false);
+      } catch (error) {
+        setSavingDashboard(false);
+        if (showErrorSnackbar) {
```
sjcobb, line 85: This is off by default since exceptionSnackbar logic is handled [here](https://github.com/perses/perses/blob/70bcb3856bc4b53bc3a37abb31d23fcb64ec5aad/ui/app/src/views/ViewDashboard.tsx#L103-L132) in Perses app but other consumers that use gql can opt-in to showing the snackbar here using new optional showErrorSnackbar prop

ui/dashboards/src/components/PanelDrawer/PanelDrawer.test.tsx
```diff
@@ -24,8 +25,10 @@ describe('Panel Drawer', () => {
 
     renderWithContext(
       <DashboardProvider initialState={{ dashboardResource: getTestDashboard(), isEditMode: true }}>
-        <DashboardProviderSpy />
-        <PanelDrawer />
+        <TimeRangeProvider initialTimeRange={{ pastDuration: '30m' }}>
```
sjcobb, line 28: Reduces a small bit of the noise we have in Jest logs currently

ui/dashboards/src/components/SaveDashboardButton/SaveDashboardButton.tsx
eunicorn: Can we just throw an error here and then let the consumer catch it and handle it however they want? That way the save dashboard button doesn't have a dependency on the snackbar provider

eunicorn (approved):

ui/dashboards/src/components/SaveDashboardButton/SaveDashboardButton.tsx
sjcobb: Yeah I’ll change this, will just take more refactoring internally, ideally we’d be able to use the whole toolbar and more logic could live there too

eunicorn: Can you rethrow the error in the catch block after calling setSavingDashboard?

Issue #1200, opened 2023-05-26
Fix typedoc issues


juliepagano:
See https://github.com/perses/perses/pull/1199 for additional context (for now -- I'll add more context here later today).

PR #1221, opened 2023-06-05
[IGNORE] upgrade and fix typedoc generation

juliepagano:
This commit:
- Upgrades typedoc to the latest release. I think they've fixed some of the issues that may have led to the OOM errors we saw before, but we'll need to keep an eye on it.
- Handles some breaking changes in the latest typedoc release related to generating docs for monorepos.
- Re-enables typedoc generation in CI.

Fixes https://github.com/perses/perses/issues/1200.

# Checklist

## UI Changes

# Screenshots

Typedoc assets again available after the CI run, which did not throw any memory errors.
image

Examples of what the docs look like now. They've changed the layout a little, but it's still very similar.
image
image

### DIFF
--- a/ui/typedoc.json
+++ b/ui/typedoc.json
@@ -1,7 +1,12 @@
 {
   "entryPointStrategy": "packages",
   "entryPoints": [
-    "."
+    "components",
+    "core",
+    "dashboards",
+    "panels-plugin",
+    "plugin-system",
+    "prometheus-plugin"
   ],
   "out": "typedoc"
 }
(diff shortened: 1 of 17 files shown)


ui/typedoc.json
```diff
@@ -1,7 +1,12 @@
 {
   "entryPointStrategy": "packages",
   "entryPoints": [
-    "."
```
juliepagano, line 4: Explicitly including the non-private packages here because otherwise you get weird empty entries in typedoc for `app` and some of the internal packages. 

At some point, we may want to reorganize the packages a little bit to make it easier to handle private vs. public packages in code like this and some of our release utils, but I don't want to tackle it in this particular PR.

saminzadeh (approved):

PR #1223, opened 2023-06-06
[BREAKINGCHANGE] reorg legend code to separate plugin spec & component concerns

juliepagano:
The purpose of this PR is to propose a change to break apart the model code for legends to better separate "props for the Legend component" (living in the `components` package) from "spec for legend usage inside a panel plugin" (living in the `plugin-system` package). 

Historically, these have had significant overlap, but that will likely change as we add more customization and functionality (e.g. I have some changes like this in mind for the table legend work). Commonalities between the two use cases have been lifted up to the `core` package for reuse across both use cases. 

TLDR:
- legend code related to the `Legend` component continues to live in the `components` package
- legend code related to panel specs moves to the `plugin-system` package, so it's available to the `panel-plugins` package  and any custom panel plugins that might want to use it.
- common legend code lives in `core`, so it can be used across both of the above.

# Description for squash commit

The purpose of this change is to improve the separation of concerns between the `Legend` component and legend-related behavior in panel plugin specifications. 

BREAKINGCHANGE: this code reorganizes some types, plugin system components, and utilities related to legends to better organize them. You may need to change the package you import from for some of these values. 
- Legend code related to the `Legend` component continues to live in the `components` package.
- Legend code related to panel specs moves to the `plugin-system` package.
- Common legend code lives in `core`.

# Checklist

## UI Changes


ui/components/src/Legend/legend-model.test.ts
```diff
@@ -11,16 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { LegendOptions, validateLegendSpec, isLegendItemVisuallySelected } from './legend';
-
-describe('validateLegendSpec', () => {
```
juliepagano, line 16: This method moved, so the test did too.

ui/components/src/Legend/legend-model.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright 2023 The Perses Authors
```
juliepagano, line 1: Eunice recommended just naming this `model`, which I like, but isn't aligned with the `component-name-model` I've seen in a few other places. I think it might be less confusing to fix the naming of those in a separate refactor PR.

saminzadeh (approved):
LGTM

PR #1228, opened 2023-06-06
[FEATURE] Shared crosshair for all time series panels on a dashboard

sjcobb:
## Overview
Support shared crosshair for all charts on a dashboard. This is only on for panel use-cases where `syncGroup` is set explicitly. Embed users of LineChart.tsx or EChart.tsx will have to explicitly set a `syncGroup` to opt-in to this feature.

See ECharts docs:
- https://echarts.apache.org/en/api.html#echarts.connect

## Recording / Screenshots
https://github.com/perses/perses/assets/9369625/b9c46df3-2306-4521-990c-e3c31cdf31da

image

### DIFF
--- a/ui/components/src/EChart/EChart.tsx
+++ b/ui/components/src/EChart/EChart.tsx
@@ -128,6 +128,7 @@ export const EChart = React.memo(function EChart<T>({
   useLayoutEffect(() => {
     if (containerRef.current === null || chartElement.current !== null) return;
     chartElement.current = init(containerRef.current, theme, { renderer: renderer ?? 'canvas' });
+    if (chartElement.current === undefined) return;
     chartElement.current.setOption(initialOption.current, true);
     onChartInitialized?.(chartElement.current);
     if (_instance !== undefined) {
@@ -140,8 +141,8 @@ export const EChart = React.memo(function EChart<T>({
     };
   }, [_instance, onChartInitialized, theme, renderer]);
 
+  // When syncGroup is explicitly set, charts within same group share interactions such as crosshair
   useEffect(() => {
-    // When syncGroup is explicitly set, charts within same group share interactions such as crosshair
     if (!chartElement.current || !syncGroup) return;
     chartElement.current.group = syncGroup;
     connect([chartElement.current]); // more info: https://echarts.apache.org/en/api.html#echarts.connect
--- a/ui/components/src/EChart/EChart.tsx
+++ b/ui/components/src/EChart/EChart.tsx
@@ -105,6 +105,7 @@ export interface EChartsProps<T> {
   sx?: SxProps<Theme>;
   onEvents?: OnEventsType<T>;
   _instance?: React.MutableRefObject<ECharts | undefined>;
+  syncGroup?: string;
   onChartInitialized?: (instance: ECharts) => void;
 }
 
@@ -115,6 +116,7 @@ export const EChart = React.memo(function EChart<T>({
   sx,
   onEvents,
   _instance,
+  syncGroup,
   onChartInitialized,
 }: EChartsProps<T>) {
   const initialOption = useRef<EChartsCoreOption>(option);
@@ -126,10 +128,6 @@ export const EChart = React.memo(function EChart<T>({
   useLayoutEffect(() => {
     if (containerRef.current === null || chartElement.current !== null) return;
     chartElement.current = init(containerRef.current, theme, { renderer: renderer ?? 'canvas' });
-    if (chartElement.current === undefined) return;
-    // https://echarts.apache.org/en/api.html#echarts.connect
-    chartElement.current.group = 'dashboard';
-    connect([chartElement.current]);
     chartElement.current.setOption(initialOption.current, true);
     onChartInitialized?.(chartElement.current);
     if (_instance !== undefined) {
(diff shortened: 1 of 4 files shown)


ui/components/src/EChart/EChart.tsx
```diff
@@ -127,6 +127,9 @@ export const EChart = React.memo(function EChart<T>({
     if (containerRef.current === null || chartElement.current !== null) return;
     chartElement.current = init(containerRef.current, theme, { renderer: renderer ?? 'canvas' });
     if (chartElement.current === undefined) return;
+    // https://echarts.apache.org/en/api.html#echarts.connect
+    chartElement.current.group = 'dashboard';
```
saminzadeh, line 131: It might be a good idea to add a new prop, `syncGroup` that allows you to this programmatically.

```tsx
// These are in sync
<EChart {...props} syncGroup="panel-group-1" />
<EChart {...props} syncGroup="panel-group-1" />

// then these are in sync
<EChart {...props} syncGroup="panel-group-2" />
<EChart {...props} syncGroup="panel-group-2" />
```

saminzadeh (approved):

ui/components/src/EChart/EChart.tsx
sjcobb: yeah for sure, I definitely was planning to add an easy way to opt out, I like the syncGroup idea though to let the parent have control over the group name too

ui/components/src/EChart/EChart.tsx
```diff
@@ -127,6 +127,9 @@ export const EChart = React.memo(function EChart<T>({
...
+    connect([chartElement.current]);
```
sjcobb, line 132: Probably a more performant way to do this, still just a prototyping

ui/components/src/EChart/EChart.tsx
```diff
@@ -105,6 +105,7 @@ export interface EChartsProps<T> {
   sx?: SxProps<Theme>;
   onEvents?: OnEventsType<T>;
   _instance?: React.MutableRefObject<ECharts | undefined>;
+  syncGroup?: string;
```
saminzadeh, line 108: If you're still unsure of if it's a good idea to expose this prop, you can throw on the `__experimental` prefix. However, I think it should be a good abstraction / we can change it if needed anyway.

saminzadeh (approved):

ui/components/src/EChart/EChart.tsx
sjcobb: I'm good with exposing it without `__experimental`, this is a pretty popular request

sjcobb: To be clear, I'm good with exposing it since it is off by default, embed users will have to opt-in to this by explicitly setting a `syncGroup`--it is only "on by default" for dashboards because of this line in [TimeSeriesChartPanel](https://github.com/perses/perses/blob/e30500c773feaa10cf0229f8a20e35df9c897e90/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx#L295)

Eventually, I like your idea of having a dashboard-level setting to control this if users want to opt out for all panels

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -292,6 +292,7 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
                 unit={unit}
                 grid={gridOverrides}
                 tooltipConfig={{ wrapLabels: true }}
+                syncGroup="default-panel-group"
```
saminzadeh, line 295: Eventually, this will be set only if the dashboard spec has `dashboard.spec.settings.graphtoolTip === 'shared'`.

ui/components/src/TimeSeriesTooltip/nearby-series.ts
```diff
@@ -129,13 +129,15 @@ export function checkforNearbySeries(
         type: 'highlight',
         seriesIndex: emphasizedSeriesIndexes,
         notBlur: false,
+        escapeConnect: true,
```
juliepagano, line 132: Might be worth putting a little comment in explaining why this needs `escapeConnect`.

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
juliepagano, line 295: Are there potential use cases where we want to have multiple groupings on a given page? If so, I think it would be good to think about how `syncGroup` could be configurable in the panel settings instead of hardcoded to a single value. I'm cool with throwing a TODO in there for now and fixing that up in a follow-up though.

juliepagano (approved):
A few small nits/questions, but overall this looks really cool. I'm glad it was so easy to set up! :rocket:

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Yeah definitely! Added the TODO, we'll want to think through dashboard settings as well as per panel overrides, especially important for embed use cases too where consumers are using multiple TimeSeriesChart.PanelComponent instances

PR #1237, opened 2023-06-07
[FEATURE] TimeSeriesChartPanel can show values in legend

juliepagano:
# Notes for reviewers

- This PR follows on from https://github.com/perses/perses/pull/1230
- There are a few additional tasks that I don't want to do in this PR in the name of small PRs and making incremental progress on delivering the table legend. I've noted these in TODOs in the code. Let me know if you feel strongly about any of them getting done now instead of in follow-ups.
- I'm holding off on updating the migpart migration scripts until this feature is a little farther along, but they are on my radar to do in a follow-up PR. 

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [x] Visual tests are stable and unlikely to be flaky. See [Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests) and [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

# Screenshots

image

image

PR #1238, opened 2023-06-08
[BREAKINGCHANGE]: move calculation model to `core` package AND s/lodash-es/lodash

juliepagano:
This change also replaces lodash-es with lodash, which should be a noop for most consumers, but will address some ESM-only issues in certain use cases (e.g. next.js).

BREAKINGCHANGE: calculation-related utilities and types have moved from the `plugin-system` package to the `core` package and associated imports will need to be adjusted.

# Notes for reviewers

- Doing this as part of responding to https://github.com/perses/perses/pull/1237#discussion_r1222288084. Doing the move of the utility to core in a separate PR, so any changes I make to combine the logic are a little easier to see in a git diff.
- This PR surfaced the esm-only issue from lodash-es again, so I'm fixing it while I'm in here.

# Checklist

## UI Changes


sjcobb (approved):
🙏

ui/components/src/EChart/EChart.tsx
```diff
@@ -14,7 +14,8 @@
 import React, { useEffect, useLayoutEffect, useRef } from 'react';
 import { ECharts, EChartsCoreOption, init, connect } from 'echarts/core';
 import { Box, SxProps, Theme } from '@mui/material';
-import { isEqual, debounce } from 'lodash-es';
+import isEqual from 'lodash/isEqual';
+import debounce from 'lodash/debounce';
```
juliepagano, line 18: I _think_ you don't need the nested imports in all use cases (e.g. I think webpack may be smart enough these days to tree shake with normal imports), but using them to be extra safe.

sjcobb: Yeah I'm still don't understand what the "correct" way to use lodash is, this approach is what I landed on [in core too](https://github.com/perses/perses/blob/v0.33.0/ui/core/src/utils/memo.ts#L15) when I needed the fix for e2e's in #1165

ui/jest.shared.ts
```diff
@@ -23,8 +23,6 @@ const config: Config.InitialOptions = {
   testEnvironment: 'jsdom',
   roots: ['<rootDir>/src'],
   moduleNameMapper: {
-    // Jest currently doesn't natively support ES Modules, so use the non ES Modules versions instead
-    '^lodash-es$': 'lodash',
```
sjcobb, line 27: 🎉

sjcobb (approved):
Lodash changes LGTM to me!

PR #1241, opened 2023-06-09
[FEATURE] Support Multi-Series Stat Chart

eunicorn:
This PR not only supports multi-series stat chart but also improves font size. Font size will now automatically grow to fit the available space.

https://github.com/perses/perses/assets/9817826/b04e2ca6-2c84-43ed-9d6b-0f67909b29c7

TO DO:
We want to add a manual option for adding your own font size as well.

### DIFF
--- a/dev/data/dashboard.json
+++ b/dev/data/dashboard.json
@@ -2281,19 +2281,17 @@
     "metadata": {
       "name": "StatChartPanel",
       "created_at": "2022-12-21T00:00:00Z",
-      "updated_at": "2023-01-27T00:35:26.333638Z",
-      "version": 8,
+      "updated_at": "2023-06-09T17:57:49.291171Z",
+      "version": 13,
       "project": "testing"
     },
     "spec": {
-      "duration": "6h",
       "panels": {
         "SimpleStat": {
           "kind": "Panel",
           "spec": {
             "display": {
-              "name": "Simple Stat",
-              "description": "Stat chart with a single value"
+              "name": "Multi-Series Stat with Sparkline"
             },
             "plugin": {
               "kind": "StatChart",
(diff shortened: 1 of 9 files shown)


dev/data/dashboard.json
```diff
@@ -2281,19 +2281,17 @@
     "metadata": {
       "name": "StatChartPanel",
       "created_at": "2022-12-21T00:00:00Z",
-      "updated_at": "2023-01-27T00:35:26.333638Z",
-      "version": 8,
+      "updated_at": "2023-06-09T17:57:49.291171Z",
```
eunicorn, line 2284: Updated testing/StatChartPanel dashboard with better mock data

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -40,14 +43,43 @@ export interface StatChartProps {
...
   const formattedValue = data.calculatedValue === undefined ? '' : formatValue(data.calculatedValue, unit);
 
+  const containerPadding = chartsTheme.container.padding.default;
+
+  // calculate series name font size and height
+  let seriesNameFontSize = useOptimalFontSize({
+    text: data?.seriesData?.name ?? '',
+    fontWeight: SERIES_NAME_FONT_WEIGHT,
+    width,
+    height: height * 0.125, // assume series name will take 12.5% of available height
```
juliepagano, line 61: Out of curiosity, where did 12.5% come from? Is it an approximation or is it based on some specific math?

ui/components/src/StatChart/calculateFontSize.ts
```diff
@@ -0,0 +1,63 @@
...
+// limitations under the License.
+
+import { useChartsTheme } from '../context/ChartsThemeProvider';
+
+interface CalculateFontSize {
+  text: string;
+  fontWeight: number;
+  width: number;
+  height: number;
+  lineHeight?: number; // default is 1.2
```
juliepagano, line 21: You can use `@default 1.2` to specify this if you put some typedoc above the `lineHeight`. I could see that being nice for folks using vscode and other tools that will show the typedoc on hover.

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -22,11 +22,14 @@ import { useChartsTheme } from '../context/ChartsThemeProvider';
 import { formatValue, UnitOptions } from '../model/units';
 import { EChart } from '../EChart';
 import { GraphSeries } from '../model/graph';
+import { useOptimalFontSize } from './calculateFontSize';
 
 use([EChartsLineChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
 
-const MIN_VALUE_SIZE = 12;
-const MAX_VALUE_SIZE = 36;
+const LINE_HEIGHT = 1.2;
```
juliepagano, line 29: Is there any association between this constant and `DEFAULT_LINE_HEIGHT` in `calculateFontSize.ts`? If so, should one be set using the other?

ui/components/src/StatChart/calculateFontSize.ts
```diff
@@ -0,0 +1,63 @@
...
+  maxSize?: number;
+}
+
+const DEFAULT_LINE_HEIGHT = 1.2;
+
+let canvasContext: CanvasRenderingContext2D | null;
+
+function useCanvasContext() {
+  if (!canvasContext) {
+    canvasContext = document.createElement('canvas').getContext('2d');
```
juliepagano, line 31: Do you happen to know if there's any perf issues with creating these over and over and not cleaning them up? I _think_ JS is smart about garbage collecting them if they're never inserted into the dom, but it'd be worth double-checking.

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
```diff
@@ -47,30 +52,68 @@ export function StatChartPanel(props: StatChartPanelProps) {
...
 
+  // Calculates chart width
+  const spacing = SPACING * (statChartData.length - 1);
+  let chartWidth = (contentDimensions.width - spacing) / statChartData.length;
+  if (isMultiSeries && chartWidth < MIN_WIDTH) {
+    chartWidth = MIN_WIDTH;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const noDataTextStyle = (chartsTheme.noDataOption.title as any).textStyle;
```
juliepagano, line 63: Out of curiosity, why is the `any` needed? Could you add a clarifying comment above the eslint-disable-line?

juliepagano (commented):
UI/UX looks great! I've got a handful of hopefully small-ish code questions.

ui/components/src/StatChart/StatChart.tsx
eunicorn: from @foxbat07, all these percentages are just approximation for how we want the stat chart to look

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
eunicorn: chartsTheme.noDataOption is type `EChartsCoreOption` but `EChartsCoreOption` doesn't have a `title` field so getting `Object is of type 'unknown'` error :(

any idea if there is a better type for noDataOption?

ui/components/src/StatChart/StatChart.tsx
eunicorn: Actually yeah calculateFontSize is more accurate if using the line height from stat chart so updating this to removing DEFAULT_LINE_HEIGHT, thanks!

ui/components/src/StatChart/calculateFontSize.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const textStyle = chartsTheme.echartsTheme.textStyle;
+  const fontSize = Number(textStyle?.fontSize) ?? 12;
+  const fontFamily = textStyle?.fontFamily ?? 'Lato';
+
+  // set the font on the canvas context
+  const fontStyle = `${fontWeight} ${fontSize}px ${fontFamily}`;
+  ctx.font = fontStyle;
+  // measure the width of the text with the given font style
+  const textMetrics: TextMetrics = ctx.measureText(text);
```
sjcobb, line 52: Nice, this is cool!

sjcobb (approved):
🚀

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
sjcobb: We should use ECharts ComposeOption actually, that would help with tooltip type weirdness too in LineChart, there is an example internally but something we can look into in separate PR later

sjcobb: ComposeOption / minimal types [docs](https://apache.github.io/echarts-handbook/en/basics/import/#creating-an-option-type-in-typescript:~:text=in%20your%20project.-,creating%20an%20option%20type%20in%20typescript,-For%20developers%20who), I had a branch where I started this I can try to find, just was going to take a bit of refactoring in LineChart but maybe easier to change in noDataOption / StatChart / GaugeChart

eunicorn: oh awesome, added a new type `NoDataOption` that uses `ComposeOption` :)

ui/panels-plugin/src/plugins/stat-chart/StatChartPanel.tsx
```diff
@@ -27,15 +31,17 @@ export function StatChartPanel(props: StatChartPanelProps) {
...
+  const isMultiSeries = statChartData.length > 1;
+
   const chartsTheme = useChartsTheme();
 
   if (queryResults[0]?.error) throw queryResults[0]?.error;
 
   if (contentDimensions === undefined) return null;
 
-  if (isLoading === true) {
+  if (isLoading || isFetching) {
```
saminzadeh, line 44: Nit: you should only need `isFetching` since it encapsulates `isLoading`. `isLoading` is only the first time

ui/components/src/StatChart/calculateFontSize.ts
saminzadeh: If we are doing something global, I would make it pretty clear with the name + a comment. The naming convention seems like a hook with the `use` prefix, so maybe `getGlobalCanvasContext()` instead since no react hooks are used.

saminzadeh (approved):

PR #1246, opened 2023-06-12
[BREAKINGCHANGE] Improve calculations for gauge and stat charts

juliepagano:
This change includes a handful of improvements for the calculation setting currently used in the gauge and stat charts and the associated utility functions for generating those calculations:
- Gauge and stat charts now communicate when the calculation is `null`. Previously, they would show a NaN in these cases, which is confusing.
- Calculation for `Mean` is now labeled as `Average` for slightly better usability/readablity. The underlying constant remains the same.
- Calculation utilities now can return `null` in cases where the data is `null`. Previously, they would return `NaN`.
- Adjusted several utility functions to handle the fact that values can now be `null`.
- Migration code now translates `firstNotNull` to `FirstNumber`, which is more equivalent than the previous translation to `First`.
- Added `FirstNumber` calculation option to provide users the ability to calculate the first non-null value.

BREAKINGCHANGE: the return values for calculation functions now return `null` instead of `NaN` when values are `null`. Several type signatures for working with values have been modified to include `null` values and may need to be accounted for.

# Checklist

## UI Changes

# Screenshots

## Gauge chart
image

image

## Stat chart

image
image


internal/api/shared/migrate/mapping.cuepart
```diff
@@ -20,7 +20,7 @@ import (
     // mapping table for the calculation attribute (key = grafana unit, value = perses equivalent)
     calc: {
         first: "First"
-        firstNotNull: "First"
+        firstNotNull: "FirstNumber"
```
juliepagano, line 23: Now that `FirstNumber` exists, migrate to that instead because it's more correct.

ui/core/src/model/calculations.ts
```diff
@@ -40,13 +40,17 @@ export const CALCULATIONS_CONFIG: Readonly<Record<CalculationType, CalculationCo
...
+  },
   LastNumber: {
     label: 'Last number',
     description: 'Last numeric value',
   },
   Mean: {
-    label: 'Mean',
-    description: 'Average value',
+    label: 'Average',
+    description: 'Average value excluding nulls',
```
juliepagano, line 53: Renaming per discussion with designers. 

Clarifying the exclusion of nulls to be explicit for the end user. Previously, this would return a `NaN` when there were `null` values, which probably isn't what we wanted.

ui/core/src/model/calculations.ts
```diff
@@ -56,38 +60,52 @@ export const CALCULATIONS_CONFIG: Readonly<Record<CalculationType, CalculationCo
 
 export const DEFAULT_CALCULATION: CalculationType = 'Sum';
 
-function first(values: TimeSeriesValueTuple[]): number | undefined {
+type CalculationValue = number | null | undefined;
+
+function first(values: TimeSeriesValueTuple[]): CalculationValue {
   const tuple = values[0];
   return tuple === undefined ? undefined : getValue(tuple);
 }
```
juliepagano, line 68: Changed a bunch of these to handle `null` instead of using `NaN`.

ui/core/src/model/calculations.ts
```diff
@@ -56,38 +60,52 @@ export const CALCULATIONS_CONFIG: Readonly<Record<CalculationType, CalculationCo
...
+function mean(values: TimeSeriesValueTuple[]): CalculationValue {
   if (values.length <= 0) return undefined;
-  return meanBy(values, getValue);
+  return lodashMean(getNonNullValues(values));
 }
 
-function sum(values: TimeSeriesValueTuple[]): number | undefined {
+function sum(values: TimeSeriesValueTuple[]): CalculationValue {
   if (values.length <= 0) return undefined;
-  return sumBy(values, getValue);
```
juliepagano, line 83: Honestly, we could probably get rid of all the lodash calc functions in here. I'm going to wait until my next PR where I look at calculating multiple values at once for table legend to take a pass at doing that.

ui/core/src/model/calculations.ts
```diff
@@ -12,14 +12,14 @@
 // limitations under the License.
 
 import findLast from 'lodash/findLast';
-import meanBy from 'lodash/meanBy';
-import sumBy from 'lodash/sumBy';
+import { default as lodashMean } from 'lodash/mean';
```
juliepagano, line 15: Renamed this to avoid conflict with the `mean` value below when I switched to the regular `mean` and not `meanBy`.

ui/core/src/model/calculations.ts
```diff
@@ -56,38 +60,52 @@ export const CALCULATIONS_CONFIG: Readonly<Record<CalculationType, CalculationCo
...
   if (values.length <= 0) return undefined;
-  return meanBy(values, getValue);
+  return lodashMean(getNonNullValues(values));
 }
 
-function sum(values: TimeSeriesValueTuple[]): number | undefined {
+function sum(values: TimeSeriesValueTuple[]): CalculationValue {
   if (values.length <= 0) return undefined;
-  return sumBy(values, getValue);
+
```
juliepagano, line 94: Previous, this would return `NaN` if there were any null values. Now it will sum the non-null values.

ui/core/src/model/calculations.ts
```diff
@@ -56,38 +60,52 @@ export const CALCULATIONS_CONFIG: Readonly<Record<CalculationType, CalculationCo
...
+}
+
+function lastNumber(values: TimeSeriesValueTuple[]): CalculationValue {
+  const tuple = findLast(values, (tuple) => typeof getValue(tuple) === 'number');
   return tuple === undefined ? undefined : getValue(tuple);
 }
 
-function mean(values: TimeSeriesValueTuple[]): number | undefined {
+function mean(values: TimeSeriesValueTuple[]): CalculationValue {
   if (values.length <= 0) return undefined;
```
juliepagano, line 88: Previously, this would return a `NaN` if there were any null values. Now it averages the non-null values.

ui/panels-plugin/src/plugins/stat-chart/StatChartOptionsEditorSettings.test.tsx
```diff
@@ -66,12 +66,12 @@ describe('StatChartOptionsEditorSettings', () => {
...
     userEvent.click(calcSelector);
     const firstOption = screen.getByRole('option', {
-      name: /First/,
+      name: /First number/,
     });
     userEvent.click(firstOption);
     expect(onChange).toHaveBeenCalledWith(
       expect.objectContaining({
-        calculation: 'First',
+        calculation: 'FirstNumber',
```
juliepagano, line 74: Switching a few tests to look for `FirstNumber` instead of `First` because it's more specific and doesn't run into RTL errors where it finds two matches.

ui/panels-plugin/src/plugins/gauge-chart/GaugeChartOptionsEditorSettings.test.tsx
```diff
@@ -66,7 +66,7 @@ describe('GaugeChartOptionsEditorSettings', () => {
     const calcSelector = screen.getByRole('combobox', { name: 'Calculation' });
     userEvent.click(calcSelector);
     const meanOption = screen.getByRole('option', {
-      name: /Mean/,
+      name: /Average/,
```
juliepagano, line 69: The label changed.

juliepagano:
For some reason github actions is being weird and auto-cancelling some of the jobs. No idea what's going on there. I'll try kicking it in a little while when the other jobs finish.

ui/core/src/model/calculations.ts
sjcobb: Due to bundle size concerns or just that these are so simple they would be clearer w/o lodash? @saminzadeh and I have talked about this in other cases too and I know he prefers using battle tested libraries for this type of thing but there are definitely tradeoffs

juliepagano: From a perf perspective, it'd be a bummer to iterate a long array n times where n is the number of calculations we need to do (so we can use each of those lodash functions) vs. having a single loop where we can do all the calculations at once. These are all pretty simple math, so I don't know that we really need a special lib for it and we can take a look at what lodash does under the hood, as needed. If we were doing much more complex calculations (e.g. the greatest common denominator stuff we get from the math lib), I'd definitely be more worried about doing it by hand.

juliepagano: The PR isn't 100% ready yet (hoping I'll give it one more pass in the morning and then un-draft it), but you can see an example of what I'm thinking hereabouts in the updated table legend values PR. https://github.com/perses/perses/pull/1237/files#r1227440821

PR #1253, opened 2023-06-16
[ENHANCEMENT] Table columns can be sorted

juliepagano:
This change adds sorting capabilities to the following components:
- `Legend` when `mode` is `table`.
- `Table` component.
- `TimeSeriesChartPanel` panel plugin.

Columns are not sorted by default and sorting must be enabled using the column definitions.

# Checklist

## UI Changes

# Screenshots

## Table

https://github.com/perses/perses/assets/396962/8c98231b-9484-41e6-ac01-94421339e527

## Table legend

https://github.com/perses/perses/assets/396962/602928dd-9cb3-4d66-a5d1-21f21f3297c6

## Time series chart panel

https://github.com/perses/perses/assets/396962/dc7db3a3-28b3-4ecf-b13b-c9c5bd4ac7ea


ui/core/src/model/calculations.ts
```diff
@@ -49,7 +49,7 @@ export const CALCULATIONS_CONFIG: Readonly<Record<CalculationType, CalculationCo
     description: 'Last numeric value',
   },
   Mean: {
-    label: 'Average',
+    label: 'Avg',
```
juliepagano, line 52: Making this smaller to fit better in the headers now that we have the sort icon taking space. I also recall someone recommending this previously as a shortening of this value, but I forgot to set it this way when doing the calculation update PRs.

ui/components/src/Table/Table.tsx
```diff
@@ -22,6 +30,12 @@ const DEFAULT_GET_ROW_ID = (data: unknown, index: number) => {
   return `${index}`;
 };
 
+// Setting these defaults one enables them to be consistent across renders instead
+// of being recreated every time, which can be important for perf because react
+// does not do deep equality checking for objects and arrays.
+const DEFAULT_ROW_SELECTION: NonNullable<TableProps<unknown>['rowSelection']> = {};
+const DEFAULT_SORTING: NonNullable<TableProps<unknown>['sorting']> = [];
```
eunicorn, line 37: Any idea why this is an array? Can you sort more than one column at a time?

ui/components/src/Table/TableHeaderCell.tsx
```diff
@@ -0,0 +1,76 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TableSortLabel, Typography, tableSortLabelClasses } from '@mui/material';
+import { SortDirection } from './model/table-model';
+import { TableCell, TableCellProps } from './TableCell';
+
+export interface TableHeaderCellProps extends TableCellProps {
+  canSort?: boolean;
```
eunicorn, line 19: Can we just check if `onSort` is defined or not to see if it should be sortable?

ui/components/src/Table/model/table-model.ts
```diff
@@ -19,10 +19,12 @@ import {
   CoreOptions,
   RowData,
   RowSelectionState,
+  SortingState,
 } from '@tanstack/react-table';
 import { CSSProperties } from 'react';
 
 export type TableDensity = 'compact' | 'standard';
+export type SortDirection = 'asc' | 'desc' | undefined;
```
eunicorn, line 27: i see SortDirection defined in tanstack as well, should we use that instead?

eunicorn (approved):

ui/components/src/Table/model/table-model.ts
juliepagano: If I recall correctly, there's  a sort direction defined in both the MUI Table and the TanStack table that are slightly different, so having an explicit definition here helps us have a single source of truth we control.

ui/components/src/Table/Table.tsx
juliepagano: We don't support multisort right now, but we could in the future, so starting with it as an array would allow us to add that without changing the type.

ui/components/src/Table/TableHeaderCell.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onSort?: ((event: unknown) => void) | undefined;
+  sortDirection?: SortDirection;
+  nextSortDirection?: SortDirection;
```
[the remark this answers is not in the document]
juliepagano: It's the next sort that will be used if you click again. It's used for some visual behavior (e.g. what direction the arrow should show up on hover if it's not sorted at the moment). I'll add a comment clarifying that!

ui/components/src/Table/TableHeaderCell.tsx
juliepagano: Good call! I'll refactor that bit.

PR #1254, opened 2023-06-16
[ENHANCEMENT] Add size option for content with legend

juliepagano:
This change impacts the following:
- `ContentWithLegend` component.
- `TimeSeriesChartPanel` panel plugin.

Sizes currently include: small and medium.

This change also includes an adjustment to the sized height of bottom positioned table legends to be smaller if there are not enough legend items to fill the specified space.

# Notes for reviewers

I'm intentionally not tackling the grafana migration piece in this PR because this one is a trickier case because there's not a 1:1 mapping between the two, and I don't want to block the basic behavior on the complex migration logic.

# Checklist

## UI Changes

# Screenshots

## Content with legend

image
image

## Time Series chart panel

image

image

https://github.com/perses/perses/assets/396962/1d8f60b4-1328-4081-862e-12c53fec6b8e

### DIFF
--- a/schemas/panels/time-series/time-series.cue
+++ b/schemas/panels/time-series/time-series.cue
@@ -22,6 +22,7 @@ import (
 #legend: {
 	position: "Bottom" | "Right"
 	mode?:    "List" | "Table"
+	size?:    "Small" | "Medium"
 	values?: [...#legend_value]
 }
 
--- a/ui/components/src/ContentWithLegend/model/content-with-legend-model.test.ts
+++ b/ui/components/src/ContentWithLegend/model/content-with-legend-model.test.ts
@@ -12,12 +12,20 @@
 // limitations under the License.
 
 import { createTheme } from '@mui/material';
-import { legendModes } from '@perses-dev/core';
+import { legendModes, legendSizes } from '@perses-dev/core';
+import * as table from '../../Table';
 import {
   ContentWithLegendLayoutOpts,
   TABLE_LEGEND_SIZE,
   getContentWithLegendLayout,
 } from './content-with-legend-model';
+
+// Workaround to get spyOn to work without a cannot redefine property error.
+// https://github.com/microsoft/TypeScript/issues/43081#issuecomment-1352352654
+jest.mock('../../Table', () => ({
+  __esModule: true,
+  ...jest.requireActual('../../Table'),
+}));
 
 const mockMuiTheme = createTheme({});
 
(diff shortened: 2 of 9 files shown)


ui/components/src/ContentWithLegend/model/content-with-legend-model.test.ts
```diff
@@ -46,40 +55,152 @@ describe('getContentWithLegendLayout', () => {
   });
```
juliepagano, line 56: The diff on these tests are a little funny because I added a wrapper describe around a bunch of them to get coverage on the size variations too.

ui/plugin-system/src/components/LegendOptionsEditor/LegendOptionsEditor.tsx
```diff
@@ -140,9 +161,25 @@ export function LegendOptionsEditor({ value, onChange }: LegendOptionsEditorProp
...
+            // TODO: enable sizes for list mode when we normalize the layout of
+            // lists to more closely match tables.
+            disabled={value === undefined || currentMode !== 'Table'}
+            disableClearable
+          ></SettingsAutocomplete>
+        }
+      />
       <OptionsEditorControl
         label="Values"
-        description="Computed values ignore nulls."
```
juliepagano, line 145: I forgot to take this note out when I updated the calculations. It's no longer valid, so ripping it out now.

schemas/panels/time-series/time-series.cue
```diff
@@ -22,6 +22,7 @@ import (
 #legend: {
 	position: "Bottom" | "Right"
 	mode?:    "List" | "Table"
+	size?: "Small" | "Medium"
```
juliepagano, line 25: FYI: some of the mocks include a "Large" size, but @foxbat07 suggested leaving it out based on some discussion of screen real estate, responsive behavior, etc.

saminzadeh (approved):
Thought: Part of me wonders if `size` should be an internal prop and not exposed to the Dashboard settings / spec. Seems like we should pick one (based on panel width) to keep it consistent and reduce decision fatigue.

juliepagano:
> Thought: Part of me wonders if `size` should be an internal prop and not exposed to the Dashboard settings / spec. Seems like we should pick one (based on panel width) to keep it consistent and reduce decision fatigue.

There's a follow-up "auto" option that @foxbat07 wants to add as a default that would behave like that. We decided to start with the hardcoded options to keep things simple.

PR #1256, opened 2023-06-19
[FEATURE] Injection of external variables

celian-garcia:
This PR introduce the logic to take into account a new type of variables into the dashboards. I call them here "external" and they contains all the variables that are not defined inside the Dashboard itself. 
Spoiler alert, in the Perses UI, it would describe the new "project scope" and "global scope" variables. 

The idea here is to include them as specific variables in the store, so they would be taken in account into the view, but not in the backend requests. 

> **Disclaimer from a react newbie**
> Despite the fact that it works well, I am not sure I'm using the store as I should. In particular, I put the ``overridden``/``overriding`` booleans inside the state object(s) of the store, but I need to recompute them in the drawer (see VariableEditor.tsx) as the store is not used there while we don´t save changes. 
> **I'm open to suggestions**.

Visually it would render the following type of screenshot, where we can see
- the external variables disabled as we cannot edit them;
- the overridden/overriding information alongside each variable in the drawer. They are dynamically updated if we add/remove a variable in the list;
- the overridden variables does not appear in the variable list behind.

> NB: the override mechanism means that a variable definition is used in place of another of same name
> local variable overrides all external variables of any source
> external variable from project source overrides external variable from global sources.

MicrosoftTeams-image
MicrosoftTeams-image (1)

### Proposal for the display of overridden/overriding variables
image

### Proposal for a shortcut to override easily an external variable
override_external_var_shortcut

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/views/ViewDashboard.tsx
+++ b/ui/app/src/views/ViewDashboard.tsx
@@ -13,7 +13,7 @@
 
 import { useNavigate, useParams } from 'react-router-dom';
 import { Box } from '@mui/material';
-import { ViewDashboard as DashboardView } from '@perses-dev/dashboards';
+import { ExternalVariableDefinition, ViewDashboard as DashboardView } from '@perses-dev/dashboards';
 import { ErrorAlert, ErrorBoundary, useSnackbar } from '@perses-dev/components';
 import { PluginRegistry } from '@perses-dev/plugin-system';
 import {
@@ -23,7 +23,7 @@ import {
   DEFAULT_DASHBOARD_DURATION,
   DEFAULT_REFRESH_INTERVAL,
 } from '@perses-dev/core';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { bundledPluginLoader } from '../model/bundled-plugins';
 import { useCreateDashboardMutation, useDashboard, useUpdateDashboardMutation } from '../model/dashboard-client';
 import AppBreadcrumbs from '../components/AppBreadcrumbs';
@@ -31,6 +31,8 @@ import { useIsReadonly } from '../model/config-client';
 import { CreateAction } from '../model/action';
 import { CachedDatasourceAPI, HTTPDatasourceAPI } from '../model/datasource-api';
 import { useNavHistoryDispatch } from '../context/DashboardNavHistory';
+import { useVariableList } from '../model/project-client';
+import { buildExternalVariableDefinition } from '../utils/variables';
 
 /**
  * Generated a resource name valid for the API.
@@ -66,6 +68,15 @@ function ViewDashboard() {
   const { isLoading } = useDashboard(projectName, dashboardName);
   let { data } = useDashboard(projectName, dashboardName);
   const isReadonly = useIsReadonly();
+
+  // Collect the Project variables and setup external variables from it
+  // TODO: Once we'll implement global variables CRUD, we'll do the same with 'global' as source
+  const { data: projectVars, isLoading: isLoadingProjectVars } = useVariableList(projectName);
+  const externalVariableDefinitions: ExternalVariableDefinition[] | undefined = useMemo(() => {
+    const result: ExternalVariableDefinition[] = [];
+    if (projectVars && projectVars.length > 0) result.push(buildExternalVariableDefinition('project', projectVars));
+    return result;
+  }, [projectVars]);
 
   const createDashboardMutation = useCreateDashboardMutation();
   const updateDashboardMutation = useUpdateDashboardMutation();
@@ -147,6 +158,7 @@ function ViewDashboard() {
   }, [actionRef, navigate, projectName]);
 
   if (isLoading) return null;
+  if (isLoadingProjectVars) return null;
 
   if (!data || data.spec === undefined || isReadonly === undefined) return null;
 
@@ -170,6 +182,7 @@ function ViewDashboard() {
             <DashboardView
               dashboardResource={data}
               datasourceApi={datasourceApi}
+              externalVariableDefinitions={externalVariableDefinitions}
               dashboardTitleComponent={
                 <AppBreadcrumbs dashboardName={getDashboardDisplayName(data)} projectName={data.metadata.project} />
               }
--- a/ui/dashboards/src/components/Variables/TemplateVariable.tsx
+++ b/ui/dashboards/src/components/Variables/TemplateVariable.tsx
@@ -19,23 +19,24 @@ import { useTemplateVariable, useTemplateVariableActions } from '../../context';
 
 type TemplateVariableProps = {
   name: VariableName;
+  source?: string;
 };
 
-export function TemplateVariable({ name }: TemplateVariableProps) {
-  const ctx = useTemplateVariable(name);
+export function TemplateVariable({ name, source }: TemplateVariableProps) {
+  const ctx = useTemplateVariable(name, source);
   const kind = ctx.definition?.kind;
   switch (kind) {
     case 'TextVariable':
-      return <TextVariable name={name} />;
+      return <TextVariable name={name} source={source} />;
     case 'ListVariable':
-      return <ListVariable name={name} />;
+      return <ListVariable name={name} source={source} />;
   }
 
   return <div>Unsupported Variable Kind: ${kind}</div>;
 }
 
-function ListVariable({ name }: TemplateVariableProps) {
-  const ctx = useTemplateVariable(name);
+function ListVariable({ name, source }: TemplateVariableProps) {
+  const ctx = useTemplateVariable(name, source);
   const definition = ctx.definition as ListVariableDefinition;
   const variablesOptionsQuery = useListVariablePluginValues(definition);
   const { setVariableValue, setVariableLoading, setVariableOptions } = useTemplateVariableActions();
@@ -45,11 +46,11 @@ function ListVariable({ name }: TemplateVariableProps) {
   const title = definition?.spec.display?.name ?? name;
 
   useEffect(() => {
-    setVariableLoading(name, variablesOptionsQuery.isFetching);
+    setVariableLoading(name, variablesOptionsQuery.isFetching, source);
     if (variablesOptionsQuery.data) {
-      setVariableOptions(name, variablesOptionsQuery.data);
+      setVariableOptions(name, variablesOptionsQuery.data, source);
     }
-  }, [variablesOptionsQuery, name, setVariableLoading, setVariableOptions]);
+  }, [variablesOptionsQuery, name, setVariableLoading, setVariableOptions, source]);
 
   let value = ctx.state?.value;
   const options = ctx.state?.options;
@@ -93,9 +94,9 @@ function ListVariable({ name }: TemplateVariableProps) {
 
     // If there is no value but there are options, set the value to the first option.
     if (!value && firstOption) {
-      setVariableValue(name, firstOption.value);
+      setVariableValue(name, firstOption.value, source);
     }
-  }, [finalOptions, setVariableValue, value, name, allowMultiple]);
+  }, [finalOptions, setVariableValue, value, name, allowMultiple, source]);
 
   return (
     <Box display={'flex'}>
@@ -110,11 +111,11 @@ function ListVariable({ name }: TemplateVariableProps) {
             // Must be selected
             if (e.target.value === null || e.target.value.length === 0) {
               if (allowAllValue) {
-                setVariableValue(name, DEFAULT_ALL_VALUE);
+                setVariableValue(name, DEFAULT_ALL_VALUE, source);
               }
               return;
             }
-            setVariableValue(name, e.target.value as VariableValue);
+            setVariableValue(name, e.target.value as VariableValue, source);
           }}
           multiple={allowMultiple}
         >
@@ -141,8 +142,8 @@ function ListVariable({ name }: TemplateVariableProps) {
   );
 }
 
-function TextVariable({ name }: TemplateVariableProps) {
-  const { state, definition } = useTemplateVariable(name);
+function TextVariable({ name, source }: TemplateVariableProps) {
+  const { state, definition } = useTemplateVariable(name, source);
   const [tempValue, setTempValue] = useState(state?.value ?? '');
   const { setVariableValue } = useTemplateVariableActions();
 
@@ -154,7 +155,7 @@ function TextVariable({ name }: TemplateVariableProps) {
     <TextField
       value={tempValue}
       onChange={(e) => setTempValue(e.target.value)}
-      onBlur={() => setVariableValue(name, tempValue)}
+      onBlur={() => setVariableValue(name, tempValue, source)}
       placeholder={name}
       label={definition?.spec.display?.name ?? name}
     />
--- a/ui/dashboards/src/components/Variables/VariableEditor.tsx
+++ b/ui/dashboards/src/components/Variables/VariableEditor.tsx
@@ -28,6 +28,7 @@ import {
   Alert,
   styled,
   capitalize,
+  Tooltip,
 } from '@mui/material';
 import AddIcon from 'mdi-material-ui/Plus';
 import { VariableDefinition } from '@perses-dev/core';
@@ -36,6 +37,7 @@ import PencilIcon from 'mdi-material-ui/Pencil';
 import TrashIcon from 'mdi-material-ui/TrashCan';
 import ArrowUp from 'mdi-material-ui/ArrowUp';
 import ArrowDown from 'mdi-material-ui/ArrowDown';
+import ContentDuplicate from 'mdi-material-ui/ContentDuplicate';
 import { Action, VariableEditForm, VariableState, VARIABLE_TYPES } from '@perses-dev/plugin-system';
 import { ExternalVariableDefinition, useDiscardChangesConfirmationDialog } from '../../context';
 import { hydrateTemplateVariableStates } from '../../context/TemplateVariableProvider/hydrationUtils';
@@ -156,6 +158,12 @@ export function VariableEditor(props: {
         draft[index + 1] = currentElement;
         draft[index] = nextElement;
       }
+    });
+  };
+
+  const overrideVariable = (v: VariableDefinition) => {
+    setVariableDefinitions((draft) => {
+      draft.push(v);
     });
   };
 
@@ -295,6 +303,14 @@ export function VariableEditor(props: {
                           </TableCell>
                           <TableCell>{getVariableLabelByKind(v.kind) ?? v.kind}</TableCell>
                           <TableCell align="right">
+                            <Tooltip title="Override">
+                              <IconButton
+                                onClick={() => overrideVariable(v)}
+                                disabled={!!variableState.get({ name: v.spec.name })}
+                              >
+                                <ContentDuplicate />
+                              </IconButton>
+                            </Tooltip>
                             <IconButton disabled>
                               <ArrowUp />
                             </IconButton>
--- a/ui/dashboards/src/components/Variables/VariableEditor.tsx
+++ b/ui/dashboards/src/components/Variables/VariableEditor.tsx
@@ -27,6 +27,7 @@ import {
   IconButton,
   Alert,
   styled,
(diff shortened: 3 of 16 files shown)


ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
```diff
@@ -58,7 +107,10 @@ export function ViewDashboard(props: ViewDashboardProps) {
           initialRefreshInterval={initialRefreshInterval}
           enabledURLParams={true}
         >
-          <TemplateVariableProvider initialVariableDefinitions={spec.variables}>
+          <TemplateVariableProvider
+            initialVariableDefinitions={spec.variables}
+            externalVariableDefinitions={externalVariableDefinitions}
```
sjcobb, line 112: Can this prop be optional? It looks like if it’s empty the project and global vars form controls won’t show so that’s good but just want to call out that embed users may not have implemented the project and global var spec so this and how the requests are made should be opt-in (For example, our onSave logic is passed into ViewDashboard from the app which could be an approach for the GET request TODO above)

sjcobb: (Really cool PR by the way! 🚀 Sorry if I jumped in to review too soon)

celian-garcia: Thanks! No problem, you're right. Don't hesitate to give feedback. There's still things that I will change though, like if I can simplify the store's statemap to prevent from things like that (state[source] || {})[name] which is completely ugly haha. I'm thinking of either a record with tuple [string,string|undefined] as key or a dedicated object. Not sure yet.

sjcobb: Just kind of skimming this so I’d need to dig in a bit more but it might be possible to have it a bit more integrated with the variables Zustand store, which leverages immer produce

sjcobb: Side note, @Gladorme, we could potentially start leveraging Zustand more in the rest of the app too as it gets more complex

celian-garcia: Note to myself: Don´t use tuple as Record key. As Macron said "This is huge mistake" :smile: 
https://www.typescriptlang.org/play?#code/MYewdgzgLgBA1gUwJ4EYBcMDa0BOBLMAcwBoZcDCYAfGAVzABMEAzAhBgXRgF4sAiMAEMAtghR9SfAAo4QAKwTAofDgG4AUKEixESAEwZsUfEVLki1OoxZtOPfkNHjJM+YuVr1m8BBAAbBAA6PxBCAApdFB5uXl09AEogA

@sjcobb If you don´t mind I would plan to create a dedicated MultiMap object. I see everywhere people using concatenation to create a single string key from multiple ones, but I'm not confident on which character will never be authorized in names.
Then I prefer to create something similar to this MultiMap object here https://stackoverflow.com/a/74594400/5690223 

Please shout if you disagree on that.

Gladorme: @sjcobb I didn't touch too much at Zustand stuff, but if it can simplify contexts / store: it would be awesome 😄

celian-garcia: I just updated the PR, to I think simplify the state map. But it doesn´t work yet perfectly, 
I have the list variables stuck in loading state, I don´t know why. I'll need anyway some more time to check and fix.

image

celian-garcia: ## I found out my issue !
After 2 days of investigation, it ended up that I didn't read enough zustand documentation..... 
I have forgotten the ``[immerable] = true`` in my state class .....
Now it works like a charm with it :rocket: 

Now, to go back to the original comment. I ensured the fact that the ``externalVariableDefinitions`` can be indeed not be set. It still works the same without it. So @sjcobb I will put the conv as resolved.

Still I have some `TODO(Célian)` comments to resolve before moving out of DRAFT state.

celian-garcia:
PR ready for review, but depends on #1267 to be finalized.

ui/dashboards/src/components/Variables/TemplateVariable.tsx
```diff
@@ -16,25 +16,27 @@ import { Select, FormControl, InputLabel, MenuItem, Box, LinearProgress, TextFie
 import { DEFAULT_ALL_VALUE, ListVariableDefinition, VariableName, VariableValue } from '@perses-dev/core';
 import { useTemplateVariable, useTemplateVariableActions } from '../../context';
 import { useListVariablePluginValues } from './variable-model';
+
 type TemplateVariableProps = {
   name: VariableName;
+  source?: string;
```
sjcobb, line 22: I like that this is optional but could it be more descriptive than `string`?

From the tests and discussion #226, it sounds like `'project' | 'global'` are the two expected values. Possible we could add 'dashboard' too and make that the default but I guess leaving undefined is better so `source` doesn't get set in the JSON unnecessarily.

sjcobb: So ignore my 'dashboard' idea, maybe this could work and TemplateVariableSource could be exported out of core package since looks like it's used in a few other files

```typescript
export type TemplateVariableSource = 'project' | 'global';

type TemplateVariableProps = {
  name: VariableName;
  source?: TemplateVariableSource;
};
```

sjcobb: Yeah true, that’s a good call!

if/when we implement the spec at Chronosphere, we’d likely want to change ‘project’ to ‘collection’ since that’s our app’s info architecture

ui/package-lock.json
celian-garcia, line 1: Strange that I have a diff on this one :thinking: I will check

ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
```diff
@@ -45,11 +48,29 @@ export function ViewDashboard(props: ViewDashboardProps) {
...
   const dashboardDuration = spec.duration ?? DEFAULT_DASHBOARD_DURATION;
-  const dashhboardRefreshInterval = spec.refreshInterval ?? DEFAULT_REFRESH_INTERVAL;
+  const dashboardRefreshInterval = spec.refreshInterval ?? DEFAULT_REFRESH_INTERVAL;
   const initialTimeRange = useInitialTimeRange(dashboardDuration);
-  const initialRefreshInterval = useInitialRefreshInterval(dashhboardRefreshInterval);
+  const initialRefreshInterval = useInitialRefreshInterval(dashboardRefreshInterval);
+
+  // Collect the Project variables and setup external variables from it
+  // TODO: Once we'll implement global variables CRUD, we'll do the same with 'global' as source
+  const { data: projectVars, isLoading: isLoadingProjectVars } = useVariableList(metadata.project);
```
celian-garcia, line 59: Checking twice how the components are architectured, it seems that the external variable retrieval should be done a bit upper and passed as a property

ui/package-lock.json
celian-garcia: reverted

AntoineThebaud (approved):
Looks good to the react-newbie that I am :D

ui/dashboards/src/components/Variables/VariableEditor.tsx
```diff
@@ -182,7 +189,7 @@ export function VariableEditor(props: {
               borderBottom: (theme) => `1px solid ${theme.palette.divider}`,
             }}
           >
-            <Typography variant="h2">Variables</Typography>
+            <Typography variant="h2">Variables (Dashboard scope)</Typography>
```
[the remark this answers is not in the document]
celian-garcia: ah yes I didn't change that. I did in the first place for me but I would agree on any proposal here. No opinion.

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -22,15 +22,45 @@ import {
...
+ * - 'global': variables defined for every dashboard of every project.
+ */
+type ExternalVariableSource = 'project' | 'global';
+
+/**
+ * Build the definition of the external variables from the variable resources collected from the API.
+ * @param source the source of external variables. Used in view only, to display and discriminate from other sources.
+ * @param variables variable resources collected from the API
+ */
+function buildExternalVariableDefinition(
```
sjcobb, line 49: I like that this lives in the app since it's specific to the Perses app data model, but it'd be good to move out of ViewDashboard, ExternalVariableSource could go in `ui/app/src/model` and buildExternalVariableDefinition to `ui/app/src/utils`

ui/dashboards/src/components/Variables/VariableEditor.tsx
```diff
@@ -260,6 +267,54 @@ export function VariableEditor(props: {
...
+                        <TableCell>Name</TableCell>
+                        <TableCell>Type</TableCell>
+                        <TableCell align="right" />
+                      </TableRow>
+                    </TableHead>
+                    <TableBody>
+                      {extVar.definitions.map((v) => (
+                        <TableRow key={v.spec.name}>
+                          <TableCell component="th" scope="row">
+                            <Switch checked={v.spec.display?.hidden !== true} disabled />
```
sjcobb, line 288: I'm confused on the expected behavior here, it makes senses that you can't edit the project variables or remove them when on a dashboard so they are disabled. But how do I hide them on a dashboard if the 'Visibility' toggle is disabled? I created a bunch of test project variables and want to use some of them on a dashboard but not necessarily show them all or else there are way too many variables in the TemplateVariableList in the dashboard toolbar. 

Am I missing a way to configure `spec.display.hidden` for project variables in the UI somewhere? Screen recording with example:

https://github.com/perses/perses/assets/9369625/83cd21eb-becb-4da2-8b7a-d21e11024980

ui/plugin-system/src/runtime/template-variables.ts
```diff
@@ -21,9 +22,106 @@ export type VariableState = {
   options?: VariableOption[];
   loading: boolean;
   error?: Error;
+  /**
+   * If a local variable is shadowing an external variable, local var will have the flag ``shadowing=true``.
+   */
+  shadowing?: boolean;
```
sjcobb, line 28: I'm not sure that I 100% understand the `shadowing` concept, is it similar to what @saminzadeh created `dependsOn` for?

sjcobb: Trying to use a project variable in a dashboard variable, it seems to work in the variable list drawer, but on a dashboard, the panels show No Data (assuming related to the loading order of the variables), also when I go to save I get:

image

Full recording [mp4 link here](https://drive.google.com/file/d/1-Sy1WbvSMZe7nAhcSb_G3Kf5OOGclY1V/view?usp=drive_link) (was too large to upload)

ui/dashboards/src/components/Variables/VariableEditor.tsx
sjcobb: Oh I missed the other PR #1290, might need to test with that one too locally? Since these changes are all related and global variables CRUD is coming too, would it make sense to merge these PRs into a feature branch first instead of going straight into `main`, I just created this [feature/global-and-project-variables](https://github.com/perses/perses/tree/feature/global-and-project-variables) branch where we could potentially consolidate and test these changes together before they go into main (cc @AntoineThebaud too if any thoughts?).

celian-garcia: Hello, the way to hide an external variable for now is to create a local variable with same name and set it to hidden (even with other PR merged) but I confess it may sound to be a bit tricky. 

We could imagine a more straightforward way to hide it, I'm personally fine with the current way but if you have a better idea I can implement it. 

Regarding the project var crud PR, it's been already merged actually ^^ Except if you cherry picked the commits to remove them from main, it's already on main branch. Then I'm not sure the feature branch is relevant here WDYT ?

ui/plugin-system/src/runtime/template-variables.ts
celian-garcia: Shadowing means that the variable definition is used in place of another of the same name but with lower priority. (i.e. a local variable with name interval will shadow/override any external variable with that name)

We could replace 'shadowing' by 'overriding' if it's clearer. The word is coming from the datasources design ping @AntoineThebaud

celian-garcia: I will check the behavior that you described in detail. There's maybe a bug in my implementation

ui/dashboards/src/components/Variables/VariableEditor.tsx
AntoineThebaud: @sjcobb Yes this the expected behavior for now, knowing that you always have the solution @celian-garcia described to hide the variable. Still, I agree that this is something that we should expect our future users to ask and I personally already had it in my personal todo (_"Study/submit proposal: editable visiblity for external variables on dashboards"_). 

Actually I would like to go even further and let people chose to "import" or not a given external variable (since with the visibility thing you already have the need to store a reference to the external variables in each dashboard individually, I don't expect that this would be more costly to implement). This way the external variables would be some kind of libraries that you'd decide to use or not.

It's probably worth creating a new Discussion for this topic. Anyway I think for now we should go with this first implem.

NB: just found the request for similar feature on Grafana side https://github.com/grafana/grafana/issues/1959, we may convert quite some people to Perses with our external vars ;)

ui/plugin-system/src/runtime/template-variables.ts
AntoineThebaud: Yes regarding the shadowing/shadowed behavior you have info on the Datasources design discussion https://github.com/perses/perses/discussions/404

Just in case, other ideas to name this:
- masking/masked
- overriding/overridden

ui/dashboards/src/components/Variables/VariableEditor.tsx
celian-garcia: Regarding the bug you spotted @sjcobb, let me rephrase from what I understood after our discussion and some quick investigation 

There are two bugs coming from the same situation : when you make a local variable (say instance) use an external variable (say job or "jobFromProj")  
- Bug one : you can´t save the dashboard as the backend make a validation to check the existence of the variable used
=> I will improve the perses backend to take in account global and project vars
- Bug two : a no data message appears on the panels using $instance.
=> I will check the depends on mechanism which is proper to the front to see if I didn't miss to include the external vars.

Thanks!

ui/dashboards/src/components/Variables/VariableEditor.tsx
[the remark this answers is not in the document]
celian-garcia: done

celian-garcia (commented):
Filing for myself one bug that I just noticed.
- Go to a dashboard
- Refresh the page
- Go to the project
- Go back to the dashboard
- Check console.logs

ui/dashboards/src/components/Variables/VariableEditor.tsx
sjcobb: Yeah that's a good summary and thanks for fixing the BE! 
For Bug two, I'm not 100% sure why the charts were showing 'No Data' in that case, but we'll definitely need to test charts where dashboard variables depend on project variables but the project variables take longer to load--same for when globals are added.

Put a summary of the idea we discussed on Mon in notes [here](https://github.com/perses/perses/discussions/947?sort=new#discussioncomment-6469469), in the future it'd be nice to look into that, basically a quicker way to define the lower level variable with the correct `spec.display` where the user just clicks the toggle and the UI takes care of the mapping under the hood

celian-garcia: I am currently looking into the second bug yeah, I'll find out (hopefully :face_with_spiral_eyes: )

For the display hidden, this is a good idea to have a shortcut, but putting this on the toggle would be in my opinion a bit too confusing for the user. Maybe a button at the very right "override locally" ? that would just create the local variable, and then people would just have to toggle the local var. I'm sure I can quickly do a proto to make you play with it and see if you find it better than what we have here.

celian-garcia: Here is the issue I created https://github.com/perses/perses/issues/1312

celian-garcia: Regarding the override shortcut, I pushed a proposal with a gif in the description.

I close this conv as resolved as everything is answered.

celian-garcia:
AAARGH I found a bug checking again my own gif... what is supposed to be the behavior if a variable depend on a hidden variable? Should it return an error ? Or should it just do the query for the hidden one silently and return the result for the visible one.

sjcobb:
> AAARGH I found a bug checking again my own gif... what is supposed to be the behavior if a variable depend on a hidden variable? Should it return an error ? Or should it just do the query for the hidden one silently and return the result for the visible one.

🙃 so my assumption has been it would be the second, that it would query for the hidden one and return result

celian-garcia:
> > AAARGH I found a bug checking again my own gif... what is supposed to be the behavior if a variable depend on a hidden variable? Should it return an error ? Or should it just do the query for the hidden one silently and return the result for the visible one.
> 
> upside_down_face so my assumption has been it would be the second, that it would query for the hidden one and return result

Thanks yes I would have assumed the same. 
Indeed I just checked on main and this is exactly the behavior. So this is a regression from my PR. I'm checking...

AntoineThebaud:
> Thanks yes I would have assumed the same. Indeed I just checked on main and this is exactly the behavior. So this is a regression from my PR. I'm checking...

Yes the goal of being able to hide a variable is "just" to remove the dropdown on the dashboard, it shouldn't change anything on the technical side (i.e variables dependencies, panel queries etc.)

celian-garcia:
> > > AAARGH I found a bug checking again my own gif... what is supposed to be the behavior if a variable depend on a hidden variable? Should it return an error ? Or should it just do the query for the hidden one silently and return the result for the visible one.
> > 
> > 
> > upside_down_face so my assumption has been it would be the second, that it would query for the hidden one and return result
> 
> Thanks yes I would have assumed the same. Indeed I just checked on main and this is exactly the behavior. So this is a regression from my PR. I'm checking...

@sjcobb 
Okay we spent some time with @AntoineThebaud investigating and brainstorming, we found several problems/facts that were here before the PR and kind of leading to this situation.

- When hidden, the variable value used is its default value.
- There's no default value mechanism for lists except the ``default_value`` field. (We could expect the first element as default value)
- We cannot set the ``default_value`` field of a global/project var from the UI. 
- We cannot set the ``display.hidden`` field of a global/project var from the UI. 
- Sometimes, values in URL / dropdown / panels are not synced properly. These are cases not easy to reproduce but present on main branch already. 

All of it together leads to the strange behavior we can see on the gif. And they explain also the bad feeling you experienced playing with the external variables as lists.

In our opinion, we can merge this PR as is and tackle the points one by one. I'll follow up on the global variable CRUD feature, where I will be able to provide a full CRUD for the ``display.hidden`` and ``default_value``.

sjcobb (approved):
I need to think about the default_value situation some more, but sounds good to address in follow ups

ui/dashboards/src/context/TemplateVariableProvider/hydrationUtils.test.ts
```diff
@@ -38,6 +39,113 @@ describe('hydrateTemplateVariableStates', () => {
...
+    expect(localStateResult.get({ source: 'project', name: 'project_var' })).toEqual({
+      value: 'something',
+      loading: false,
+      overriding: false,
+      overridden: true,
+    });
+    expect(localStateResult.get({ source: 'global', name: 'greetings' })).toEqual({
+      value: 'hello',
+      loading: false,
+      overriding: false,
```
sjcobb, line 141: Love the change to overriding and overridden, a lot more clear now!

sjcobb:
The new override shortcut is really cool! 🚀

celian-garcia:
> The new override shortcut is really cool! rocket

Thanks @sjcobb for all the reviews. I don't have the rights to merge btw so you can if you want, or I'll ask @Gladorme who's back this Monday :+1:

AntoineThebaud:
I can do it ;)

PR #1259, opened 2023-06-20
[BUGFIX] Highlighted series not always cleared when cursor moved quickly

sjcobb:
Move clear highlighted series function to onMouseLeave so it works even when cursor is moved quickly.

Before it could stay like this when leaving panel:
image


juliepagano (approved):
LGTM! 🚀

PR #1262, opened 2023-06-21
[BREAKINGCHANGE] Convert all panel spec values to kebab-case

sjcobb:
See discussion #1243, we agreed on standardizing that core panel spec values should be kebab-case instead of using PascalCase everywhere. 

We still use PascalCase for `kind` but all other properties like `calculation`, `position`, `stack`, `legend.mode`, etc. all use kebab-case. Comment [here](https://github.com/perses/perses/discussions/1243) describes the rationale.

Kubernetes docs: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Follow up PR for fixing `unit` to match casing guidelines with be #1278


AntoineThebaud (approved):

cndonovan (commented):
Is the idea that kebab-case is just the more common / expected way?

But there's a specific reason we want to keep `kind` as PascalCase?

(Tried to read through https://github.com/perses/perses/discussions/1243, but wondering about the TL;DR)

sjcobb:
> Is the idea that kebab-case is just the more common / expected way?
> 
> But there's a specific reason we want to keep `kind` as PascalCase?
> 
> (Tried to read through #1243, but wondering about the TL;DR)

Yeah that's right about kebab-case being more common! @juliepagano also mentioned PascalCase being unexpected when I asked for feedback on the schema before. 

`kind` is left PascalCase since that is a standard in Kubernetes resource definitions ([example](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)), although I definitely think we are overusing it especially in the more nested parts of the schema (see comments [here](https://github.com/perses/perses/discussions/1243#discussioncomment-6461437)), which is one reason I'm proposing we change the `unit` schema in #1278

PR #1268, opened 2023-06-26
[FEATURE] Time series chart rewritten for intuitive x axis bucketing, improved render perf

sjcobb:
## Overview
- Sync branch from previous time axis prototype #1134 with all changes from main. 
- Add `show_legacy_chart` panel option to allow testing TimeChart and LineChart side-by-side
- nearby-series functions adjusted to work in both scenarios
- prepend 'legacy' to types that will be deprecated

## Screenshot
image


ui/core/src/utils/time-series-data.ts
```diff
@@ -65,28 +65,68 @@ export function getYValues(series: TimeSeries, timeScale: TimeScale): Array<numb
...
+
+      // Otherwise, see if this query has a start or end outside of the current
+      // time range
+      if (data.timeRange.start < timeRange.start) {
+        timeRange.start = data.timeRange.start;
+      }
+      if (data.timeRange.end > timeRange.end) {
+        timeRange.end = data.timeRange.end;
+      }
+    } else if (data && 'values' in data) {
```
eunicorn, line 95: added this so we could also use this function in TimeChart component to get a default time scale

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -125,30 +128,41 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   const { setTimeRange } = useTimeRange();
 
   // Populate series data based on query results
-  const { graphData } = useDeepMemo(() => {
+  const { graphData, timeScale, timeChartData, timeSeriesMapping, legendItems } = useDeepMemo(() => {
```
eunicorn, line 131: moved `legendItems` out of `graphData` so it could be reused for TimeChart as well

ui/components/src/TimeSeriesTooltip/nearby-series.ts
```diff
@@ -36,10 +37,151 @@ export interface NearbySeriesInfo {
...
+  for (let seriesIdx = 0; seriesIdx < totalSeries; seriesIdx++) {
+    const currentSeries = seriesMapping[seriesIdx];
+    if (currentSeries === undefined) break;
+
+    const currentDataset = totalSeries > 0 ? data[seriesIdx] : null;
+    if (currentDataset == null) break;
+
+    const currentDatasetValues: TimeSeriesValueTuple[] = currentDataset.values;
+
+    for (const [timestamp] of currentDatasetValues) {
```
sjcobb, line 80: It might be possible to move this logic inside the for loop below so that we're only looping through currentDatasetValues once instead of twice

ui/core/src/utils/time-series-data.ts
sjcobb: Makes sense! So in a future release once we confirm the new chart works as expected, we can remove the logic above and only keep this new if condition (and remove the `show_legacy_chart` panel option / other conditions for LineChart).

ui/components/src/TimeSeriesTooltip/nearby-series.ts
sjcobb: nvm I see now, we do need to determine closestTimestamp _before_ the second loop that compares with xValue

eunicorn (approved):

eunicorn:
Yay 🚀

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -89,26 +80,7 @@ export interface LineChartProps {
...
 }
 
-// Intentionally making this an object to start because it is plausible we will
-// want to support focusing by other attributes (e.g. index, name) in the future,
-// and starting with an object will make adding them a non-breaking change.
-type LineChartHandleFocusOpts = {
-  id: string;
-};
-
-export type LineChartHandle = {
```
cndonovan, line 99: What is "handle" meant to mean in this situation? It looks like this object is related to highlighting series, but I can't quite figure out what that has to do with a "handle".

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -0,0 +1,364 @@
...
+  ToolboxComponent,
+  TooltipComponent,
+  LegendComponent,
+  CanvasRenderer,
+]);
+
+export interface TimeChartProps {
+  height: number;
+  data: TimeSeries[];
+  seriesMapping: TimeChartSeriesMapping;
```
cndonovan, line 73: What is a seriesMapping?

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -0,0 +1,364 @@
...
+  timeScale?: TimeScale;
+  yAxis?: YAXisComponentOption;
+  unit?: UnitOptions;
+  grid?: GridComponentOption;
+  tooltipConfig?: TooltipConfig;
+  noDataVariant?: 'chart' | 'message';
+  syncGroup?: string;
```
cndonovan, line 80: What is a syncGroup?

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -0,0 +1,364 @@
...
+    timeScale = timeScaleProp;
+  }
+
+  useImperativeHandle(
+    ref,
+    () => {
+      return {
+        highlightSeries({ name }: ChartHandleFocusOpts) {
+          if (!chartRef.current) {
+            // No chart. Do nothing.
```
cndonovan, line 137: If you like it, you could copy the updated comments I saw in another file: "When chart is undefined, do not highlight series when hovering over legend"

ui/components/src/TimeSeriesTooltip/nearby-series.ts
```diff
@@ -151,6 +286,70 @@ export function checkforNearbySeries(
  * If yes, convert from pixel values to logical cartesian coordinates and return all nearby series
  */
```
cndonovan, line 287: Looks like this was the comment for legacyGetNearbySeries - do we need to change anything about how we describe this newer function?

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -0,0 +1,364 @@
+// Copyright 2023 The Perses Authors
```
saminzadeh, line 1: Nit: `TimeChart` -> `TimeSeriesChart`

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -359,24 +406,41 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
+                  onDataZoom={handleDataZoom}
+                  //  Show an empty chart when there is no data because the user unselected all items in
+                  // the legend. Otherwise, show a "no data" message.
+                  noDataVariant={
+                    !graphData.timeSeries.length && legendItems && legendItems.length > 0 ? 'chart' : 'message'
+                  }
+                />
+              ) : (
+                <TimeChart
+                  ref={lineChartRef}
```
cndonovan, line 428: Should we call this something like `chartRef` since it can be used by both LineChart and TimeChart?

cndonovan (approved):
Doing my best to follow :) Looks good, looks like a lot of stuff that we had for LineChart and we're recreating it for TimeChart.

ui/components/src/TimeChart/TimeChart.tsx
sjcobb: As part of this PR, I decided to switch to use ECharts `dataset` so that raw time series data could be passed into the chart instead of having loops for graphData.xAxis and individual series storing their own values. So all data is stored inside dataset.source and the "styles" are separate and defined in `seriesMapping`. 

`seriesMapping` then uses datasetId instead of storing the data directly in the series object. I guess the downside here is embed users have to know a bit about how ECharts `dataset` works, but overall, I do like the approach and this tutorial explains the tradeoffs: https://apache.github.io/echarts-handbook/en/concepts/dataset/#define-data-in-dataset

ui/components/src/TimeChart/TimeChart.tsx
sjcobb: So it’s used for the shared crosshair we added recently, it tells ECharts which charts are connected

ui/components/src/LineChart/LineChart.tsx
sjcobb: Yeah that came in with recent legend changes, just renamed to ChartInstance

ui/components/src/TimeChart/TimeChart.tsx
sjcobb: I kind of like TimeChart for now, since there are various exports of TimeSeriesChart already from panels-plugin and internally the forked LineChart is called TimeSeriesChart, maybe after we consolidate or figure out the plan for a context-aware TimeChart, we can rename it (possibly TimeSeriesViz right?)

ui/components/src/TimeSeriesTooltip/nearby-series.ts
sjcobb: @eunicorn we might not even need the two functions right since xBuffer isn't necessary anymore? Might just leave it for now though just in case and cleanup in a follow up PR when we remove all the LineChart logic

eunicorn: i think this function is still called in the line chart tooltip?

but yeah agreed we should just leave it for now until we completely clean up LineChart just to be safe

sjcobb: Gotcha, I was thinking that it’s possible the same function can be called in LineChart and TimeChart now, we probably will need to keep LineChart around for a while since it’s used so much internally but yeah, the panel can be simplified a ton once the show_legacy_chart option is removed

PR #1269, opened 2023-06-26
[ENHANCEMENT] Add ability to select font size for stat chart

ellenwyllie:
- Adds the ability to manually select the font size in stat charts

https://github.com/perses/perses/assets/22085207/53eb7671-99aa-4b6a-ad50-743ae85d347f

# Checklist

## UI Changes


ui/panels-plugin/src/plugins/stat-chart/stat-chart-model.ts
```diff
@@ -26,6 +27,7 @@ export interface StatChartOptions {
   unit: UnitOptions;
   thresholds?: ThresholdOptions;
   sparkline?: StatChartSparklineOptions;
+  fontSize?: FontSizeOptions;
```
sjcobb, line 30: Would it be worth distinguishing that this is `value_font_size`? We currently don't display any other labels but I could see that being added in the future similar to Grafana ([ex](https://play.grafana.org/d/Zb3f4veGk/2-stats?orgId=1)). 

Also can you update the CUE schema for StatChart [here](https://github.com/perses/perses/blob/main/schemas/panels/stat/stat.cue), could be something like: 
```
calculation: common.#calculation
unit?:  common.#unit
thresholds?:  common.#thresholds
sparkline?:  #sparkline
value_font_size?: number
```

Then some of the make commands can be useful like: 
- make fmt
- make cue-eval
- make cue-test

ui/components/src/FontSizeSelector/FontSizeSelector.tsx
```diff
@@ -0,0 +1,63 @@
...
+  '48',
+  '56',
+  '64',
+  '72',
+  '96',
+  '128',
+  '160',
+  '192',
+] as const;
+export type FontSizeOptions = (typeof FONT_SIZE_OPTIONS)[number];
```
juliepagano, line 37: Nit: could you make this singular `FontSizeOption`? The plural makes sense for the array because it's a list of things, but this represents a single value (a string that can be one of values from a union).

ui/components/src/StatChart/calculateFontSize.ts
```diff
@@ -37,10 +39,23 @@ function getGlobalCanvasContext() {
...
+  width,
+  height,
+  lineHeight,
+  maxSize,
+  fontSizeOverride,
+}: CalculateFontSize) {
   const ctx = getGlobalCanvasContext();
   const chartsTheme = useChartsTheme();
 
+  // if user has selected a font size in the settings, use it instead of calculating the optimal size
```
juliepagano, line 54: Is it worth adding a test for the use case where the font is overridden?

ui/panels-plugin/src/plugins/stat-chart/stat-chart-model.ts
```diff
@@ -42,5 +44,6 @@ export function createInitialStatChartOptions(): StatChartOptions {
       kind: 'Decimal',
     },
     sparkline: {},
+    fontSize: 'Default',
```
eunicorn, line 47: i think it's ok to remove this since fontSize is optional

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -75,6 +77,8 @@ export function StatChart(props: StatChartProps) {
   const availableHeight = height - seriesNameHeight;
   const valueFontSize = useOptimalFontSize({
     text: formattedValue,
+    // override the font size if user selects it in the settings
+    fontSizeOverride: fontSize,
```
eunicorn, line 81: I'm thinking it might be simpler if we just return font size 

```
    const valueFontSize = fontSize !== undefined ? fontSize : useOptimalFontSize()
```

eunicorn: Wait never mind, you cannot conditionally call a hook. Ignore this 😅

juliepagano:
Is this ready for another pass or are you still working through all the feedback? Either is ok -- just wanted to make sure I knew if I should re-review yet or not.

ui/panels-plugin/src/plugins/stat-chart/stat-chart-model.ts
ellenwyllie: Yeah it can't hurt to be more specific. I changed it to `value_font_size` and updated the cue file!

ui/components/src/FontSizeSelector/FontSizeSelector.tsx
ellenwyllie: Good point, updated!

ellenwyllie:
> Is this ready for another pass or are you still working through all the feedback? Either is ok -- just wanted to make sure I knew if I should re-review yet or not.

@juliepagano It's ready for another review now!

ui/panels-plugin/src/plugins/stat-chart/stat-chart-model.ts
ellenwyllie: Removed!

eunicorn (approved):
looks great! 🚀

juliepagano (approved):
LGTM! Thanks for working through all the trickiness with the cuelang and naming feedback. 🚀

PR #1271, opened 2023-06-26
[BREAKINGCHANGE] Clean up refresh_interval in dashboard spec

sjcobb:
I usually prefer camelCase but for consistency, I don't want some dashboard spec properties snake_case and others camelCase. Discussion up here if anyone has strong feelings: https://github.com/perses/perses/discussions/1243

### DIFF
--- a/ui/core/src/model/dashboard.ts
+++ b/ui/core/src/model/dashboard.ts
@@ -29,7 +29,7 @@ export interface DashboardSpec {
   display?: Display;
   datasources?: Record<string, DatasourceSpec>;
   duration: DurationString;
-  refreshInterval: DurationString;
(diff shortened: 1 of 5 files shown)


sjcobb:
Should we just call it `refresh`? Just realized how it's stored in URL params is different and it's refresh:
- ex) `steven-dashboard?start=5m&refresh=0s`

Gladorme:
`refresh` alone seems to be a little vague, no? It could define a boolean for enabling auto-refresh or something else. With `_interval` there is no hesitation.
The `refresh` in the url is probably something took from Grafana. IMO, I prefer to have everything aligned too, maybe just `refresh` is the right choice 😄

PR #1278, opened 2023-07-06
[BREAKINGCHANGE] Move unit inside new `format` obj + change to kebab case + `abbreviate` renamed as `shortValues`

sjcobb:
# Description

- Our `unit` property currently breaks two rules proposed in discussion #1243 since it uses `kind` but has no `spec` to group child properties (and to fix this, changing `kind` to `unit` means the values should be kebab-case). At this inner level of a panel’s schema, I prefer moving away from using `kind` to avoid this extra unnecessary nesting (similar argument for why I’d like to change palette.kind as well)
- It also doesn't really make sense for `decimal_places` and `abbreviate` to be grouped inside `unit`, instead I'm proposing we change to group these properties inside a new `format` object. 
- proposing also to rename `abbreviate` into `shortValues`.

All in all, example of old vs new proposed schema:

before:
```typescript
{
  "kind": "TimeSeriesChart",
  "spec": {
    "yAxis": {
      "unit": { 
        "kind": "PercentDecimal",
        "decimals": 2,
        "abbreviate": true
      }
    }
  }
}
```
after:
```typescript
{
  "kind": "TimeSeriesChart",
  "spec": {
    "yAxis": {
      "format": { 
        "unit": "percent-decimal",
        "decimals": 2,
        "shortValues": true
      }
    }
  }
}
```

~~Another option is to not group with ‘format’ at all and move the properties to be at top level inside y_axis. This would be a larger refactor since many function params would need to change, while renaming unit to format as proposed is much simpler to implement.~~

# Screenshots

image

# Checklist

## UI Changes


Gladorme (approved):
LGTM

ui/components/src/FormatControls/FormatControls.tsx
```diff
@@ -52,13 +52,13 @@ function getOptionByDecimalPlaces(decimalPlaces?: number) {
   return DECIMAL_PLACES_OPTIONS.find((o) => o.decimalPlaces === decimalPlaces);
 }
 
-export function UnitSelector({ value, onChange, disabled = false }: UnitSelectorProps) {
```
celian-garcia, line 55: I would have kept the word Selector as I feel it's the same meaning than DatasourceSelector

+ Controls will come in force with the validation use case

celian-garcia (changes requested):

ui/components/src/FormatControls/FormatControls.tsx
celian-garcia: after discussion, nevermind it has nothing to do with DatasourceSelector

celian-garcia (approved):

PR #1279, opened 2023-07-06
[FEATURE] Implement BarChart component

ellenwyllie:
Implement a `BarChart` component based on echarts BarChart. The corresponding panel plugin will be implemented in a follow up PR.

Screen Shot 2023-07-06 at 3 24 25 PM
Screen Shot 2023-07-06 at 3 09 18 PM

# Checklist

## UI Changes

### DIFF
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -80,13 +80,7 @@ export function BarChart(props: BarChartProps) {
       },
       tooltip: {
         formatter: (params: { name: string; data: number[] }) =>
-          params.data[1] && `<b>${params.name}</b> - ${formatValue(params.data[1], unit)}`,
-        backgroundColor: theme.palette.background.tooltip,
-        borderColor: theme.palette.background.tooltip,
-        textStyle: {
-          color: theme.palette.text.primary,
-          fontSize: 12,
-        },
+          params.data[1] && `<b>${params.name}</b> &emsp; ${formatValue(params.data[1], unit)}`,
       },
     };
   }, [data, chartsTheme, theme, unit]);
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -31,7 +31,7 @@ export interface BarChartData {
 export interface BarChartProps {
   width: number;
   height: number;
-  data: BarChartData[] | null | undefined;
+  data: BarChartData[] | null;
   unit: UnitOptions;
 }
 
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -38,7 +38,6 @@ export interface BarChartProps {
 export function BarChart(props: BarChartProps) {
   const { width, height, data, unit } = props;
   const chartsTheme = useChartsTheme();
-  const theme = useTheme();
 
   const option: EChartsCoreOption = useMemo(() => {
     if (data == null) return chartsTheme.noDataOption;
@@ -68,14 +67,7 @@ export function BarChart(props: BarChartProps) {
       series: {
         type: 'bar',
         label: {
-          position: 'right',
-          show: true,
           formatter: (params: { data: number[] }) => params.data[1] && formatValue(params.data[1], unit),
-          color: theme.palette.text.primary,
-        },
-        itemStyle: {
-          borderRadius: 4,
-          color: theme.palette.designSystem.blue[400],
         },
       },
       tooltip: {
@@ -83,7 +75,7 @@ export function BarChart(props: BarChartProps) {
           params.data[1] && `<b>${params.name}</b> &emsp; ${formatValue(params.data[1], unit)}`,
       },
     };
-  }, [data, chartsTheme, theme, unit]);
+  }, [data, chartsTheme, unit]);
 
   return (
     <EChart
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -23,19 +23,19 @@ import { EChart } from '../EChart';
 
 use([EChartsBarChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
 
-export interface BarGaugeChartData {
+export interface BarChartData {
   label: string;
   value: number | null | undefined;
 }
 
-export interface BarGaugeChartProps {
+export interface BarChartProps {
   width: number;
   height: number;
-  data: BarGaugeChartData[] | null | undefined;
+  data: BarChartData[] | null | undefined;
   unit: UnitOptions;
 }
 
-export function BarGaugeChart(props: BarGaugeChartProps) {
+export function BarChart(props: BarChartProps) {
   const { width, height, data, unit } = props;
   const chartsTheme = useChartsTheme();
   const theme = useTheme();
@@ -43,7 +43,7 @@ export function BarGaugeChart(props: BarGaugeChartProps) {
   const option: EChartsCoreOption = useMemo(() => {
     if (data == null) return chartsTheme.noDataOption;
 
-    const source: Array<Array<BarGaugeChartData['label'] | BarGaugeChartData['value']>> = [];
+    const source: Array<Array<BarChartData['label'] | BarChartData['value']>> = [];
     data.map((d) => {
       source.push([d.label, d.value]);
     });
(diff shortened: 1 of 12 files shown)


ui/components/src/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+          show: true,
+          formatter: (params: { data: number[] }) => params.data[1] && formatValue(params.data[1], unit),
+          color: theme.palette.text.primary,
+        },
+        itemStyle: {
+          borderRadius: 4,
+          color: theme.palette.designSystem.blue[400],
+        },
+      },
+      tooltip: {
```
eunicorn, line 81: i think we might have to implement a custom tooltip here similar to what we have for time series chart so we can pin the tooltip 

@sjcobb what do you think?

ui/components/src/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { CanvasRenderer } from 'echarts/renderers';
+import { useTheme } from '@mui/material';
+import { useChartsTheme } from '../context/ChartsThemeProvider';
+import { EChart } from '../EChart';
+
+use([EChartsBarChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
+
+export interface BarChartData {
+  label: string;
+  value: number | null | undefined;
```
eunicorn, line 28: i think it's ok if we don't accept `undefined` here since it doesn't make sense to have an undefined value

ui/components/src/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+          show: false,
+        },
+      },
+      series: {
+        type: 'bar',
+        label: {
+          position: 'right',
+          show: true,
+          formatter: (params: { data: number[] }) => params.data[1] && formatValue(params.data[1], unit),
+          color: theme.palette.text.primary,
```
eunicorn, line 74: maybe we can add bar label and itemStyle to the charts theme

ui/components/src/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export interface BarChartData {
+  label: string;
+  value: number | null | undefined;
+}
+
+export interface BarChartProps {
+  width: number;
+  height: number;
+  data: BarChartData[] | null | undefined;
```
eunicorn, line 34: not sure if we need to do it now, but i think we need to account for multiple series in the future

eunicorn:
This looks great!!

ui/components/src/BarChart/BarChart.tsx
sjcobb: I’m not sure if tooltip pinning is really as useful on BarChart, it’s more relevant on TimeSeriesChart since there can be so many series (pin to scroll, click ‘Show All’, potentially filter in future, etc.).

I was hoping we could use ECharts tooltip and style it to look as close to the custom React TimeSeriesTooltip as possible (and make those styles inherited in chartsTheme so they’re applied by default anytime ECharts tooltip is used)

Really the only reason I built the custom React tooltip in the first place was for performance when there are hundreds or thousands of series, the ECharts tooltip is pretty great in all other cases and gives embed users a lot more freedom to customize options, adjust the formatter, decide if they want trigger axis or trigger item, etc.

sjcobb: (Or change the TimeSeriesTooltip styles to match if we end up liking how the ECharts tooltip looks better)

But yeah, we should confirm with @foxbat07, how we integrate with tracing in the future (do we need a way to display a context menu?) will also play a factor

ui/components/src/BarChart/BarChart.tsx
sjcobb: +1 to building out the charts theme as much as possible (especially since we have so many bar charts internally that shouldn’t be reimplementing so many options)

ui/components/src/BarChart/BarChart.tsx
sjcobb: I’ve been out so I need to get caught up on all the use cases we are trying to cover because if this was supposed to cover BarGauge too, I was expecting it to be a stacked bar so thresholds could be supported. Do we have a plan for adding thresholds later, is that a separate component and then there will be logic in the panel that shows either component depending if thresholds are populated?

sjcobb: Really I think BarGauge could be covered by adding panel options (visual.display: ‘radial’ as default and visual.display: ‘bar’) to GaugeChart and adding logic to *that* panel but mostly trying to understand the schema for BarChart since sounded like Karan wanted it to cover both use cases

ui/components/src/BarChart/BarChart.tsx
saminzadeh: +1 I don't think we need a custom tooltip since it's not focused on timeseries data. However, I would be in favor of getting styling close as possible. I would say do a fast follow it we can't do it quickly

ui/components/src/BarChart/BarChart.tsx
saminzadeh: > not sure if we need to do it now, but i think we need to account for multiple series in the future


When you say multiple series, do you mean sub grouping? Example below
image

I think we could extend (not needed now) this to look like this:

```ts 

type BarChartData = {
   label: string,
   grouping: [{
     label: string,
     value: number
   }]
}
```

Where "Variety" is the top level group and "Genotype" is the sub group.

eunicorn: Oh yeah that's what i meant! Since this isn't time series data, does that also mean we need to update the DataQueryProvider?

ui/components/src/BarChart/BarChart.tsx
ellenwyllie: I'll look into moving the tooltip styling into the echarts theme and work with Mohit to make sure all tooltip styling is the same!

ui/components/src/BarChart/BarChart.tsx
ellenwyllie: Good point, updated it!

ui/components/src/BarChart/BarChart.tsx
saminzadeh: > Oh yeah that's what i meant! Since this isn't time series data, does that also mean we need to update the DataQueryProvider?

Not necessarily, you can transform a TimeSeries data interface by using 2 labels on each series. 1 label would be the top level group, and one would be the sub group. Example:


```ts
// first compute the last calculation of a timeseries
[
   { labels: { app: "app-1", device: "ios" }, lastValue: 10 },
   { labels: { app: "app-1", device: "android" }, lastValue: 20 },
   { labels: { app: "app-2", device: "ios" }, lastValue: 2 },
   { labels: { app: "app-2", device: "android" }, lastValue: 4 },
]
```

Then group them for the bar chart
```tsx
[
   { label: "app-1", grouping: [{ label: "ios", value: 10 }, { label: "android", value: 20 }] },
   { label: "app-2", grouping: [{ label: "ios", value: 2 }, { label: "android", value: 4 }] },
]
```

eunicorn: ohhhh got it thanks!

ui/components/src/BarChart/BarChart.tsx
ellenwyllie: I updated the tooltip styling to match the `TimeSeriesTooltip` and moved it into the echarts theme 6d10815c5ef38cc9db53a5a2ad7724792faa85e2
Screen Shot 2023-07-10 at 1 38 40 PM

ui/components/src/BarChart/BarChart.tsx
ellenwyllie: Done! f21e62d7e4d0f4164688c20fcb2150dfad91a46c

ui/components/src/BarChart/BarChart.tsx
ellenwyllie: Multi series and thresholds aren't part of the MVP but I think we will likely add them later. We don't have designs for either yet so I'm not sure exactly what it would like

ui/components/src/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,89 @@
...
+  height: number;
+  data: BarChartData[] | null;
+  unit: UnitOptions;
+}
+
+export function BarChart(props: BarChartProps) {
+  const { width, height, data, unit } = props;
+  const chartsTheme = useChartsTheme();
+
+  const option: EChartsCoreOption = useMemo(() => {
```
sjcobb, line 41: I had a branch at one point where I started to do this for LineChart but someday I want to start using ComposeOption instead of EChartsCoreOption, not a blocker or anything 

Here is more info about it and tracing started using it internally too

https://apache.github.io/echarts-handbook/en/basics/import/#:~:text=in%20your%20project.-,creating%20an%20option%20type%20in%20typescript,-For%20developers%20who

sjcobb (approved):
🎉

Issue #1033, opened 2023-03-20
UI: CRUD for Datasources


AntoineThebaud:
Add the necessary screens & forms to be able to create, read, update and delete datasources from the UI.

Notable difference with e.g dashboards, a datasource can be created a 3 different levels (global, project and dashboard).

AntoineThebaud:
Current status:
- Project datasources CRUD to be tackled with https://github.com/perses/perses/pull/1282.
- Global datasources CRUD to be tackled with https://github.com/perses/perses/pull/1319.
- Local datasources CRUD still to be done.

celian-garcia:
@AntoineThebaud I took the liberty to update your status comment with the last PR. 
This PR, when done, should close this issue. If you think of something that I missed tell me!

AntoineThebaud:
Hmm I think your PR tackles something different, by Local Datasources CRUD I meant having access to the create/edit datasource form for the local/dashboard datasources, just like we have for dashboard variables.

celian-garcia:
oh yes okay! I forgot that. Let me remove my edit and change the closes to relates

AntoineThebaud:
No worries, this is something we also forgot upstream when we planned the activities for this summer 😅

PR #1282, opened 2023-07-07
[FEATURE] Add project datasources CRUD

AntoineThebaud:
Issue ref: https://github.com/perses/perses/issues/1033

This PR adds project datasources CRUD, leveraging on the tab layout from #1267. It also comes with:
- A bugfix on backend side: updating the already-defined-as-default datasource was causing unwanted request rejection (_"cannot set A as default since there's already a default datasource set (A)"_).
- New reusable component `OptionsEditorRadio` for RadioGroup-based options.

2023-07-10_09h37_59

_NB: contrary to what was done for variables, here you have no action button in the list view: clicking on a row opens the drawer form for this datasource and it's there that you can act on it. We agreed with @Gladorme to have this divergence between variables & datasources UX temporarily: since we were not able to identify which one is better, the goal is to have people being able to play with the 2 designs for a while (should end before v1 release :D), then ultimately decide which one we prefer to keep & implement consistency between datasources & variables UX in a later PR._

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
+++ b/ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
@@ -1,0 +1,97 @@
+// Copyright 2023 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Datasource } from '@perses-dev/core';
+import { DispatchWithoutAction, useState } from 'react';
+import { DiscardChangesConfirmationDialog, Drawer, ErrorAlert, ErrorBoundary } from '@perses-dev/components';
+import { PluginRegistry } from '@perses-dev/plugin-system';
+import { bundledPluginLoader } from '../../model/bundled-plugins';
+import { DeleteDatasourceDialog } from '../dialogs/DeleteDatasourceDialog';
+import { DatasourceEditorForm } from './DatasourceEditorForm';
+
+interface DatasourceDrawerProps {
+  datasource: Datasource;
+  isOpen: boolean;
+  saveAction: string;
+  onSave: (datasource: Datasource) => void;
+  onClose: DispatchWithoutAction;
+}
+
+export function DatasourceDrawer(props: DatasourceDrawerProps) {
+  const { datasource, isOpen, saveAction, onSave, onClose } = props;
+  const [isDeleteDatasourceDialogStateOpened, setDeleteDatasourceDialogStateOpened] = useState<boolean>(false);
+  const [isDiscardDialogStateOpened, setDiscardDialogStateOpened] = useState<boolean>(false);
+  const [hasDraftChanges, setHasDraftChanges] = useState<boolean>(false);
+
+  const handleSave = (datasource: Datasource) => {
+    setHasDraftChanges(false);
+    onSave(datasource);
+  };
+
+  const handleClose = () => {
+    if (hasDraftChanges) {
+      setDiscardDialogStateOpened(true);
+    } else {
+      onClose();
+    }
+  };
+
+  const handleDelete = () => {
+    setHasDraftChanges(false);
+    onClose();
+  };
+
+  const handleDiscard = () => {
+    setHasDraftChanges(false);
+    setDiscardDialogStateOpened(false);
+    onClose();
+  };
+
+  return (
+    <Drawer isOpen={isOpen} onClose={handleClose} data-testid="datasource-editor">
+      <ErrorBoundary FallbackComponent={ErrorAlert}>
+        <PluginRegistry
+          pluginLoader={bundledPluginLoader}
+          // TODO this required field is useless here
+          defaultPluginKinds={{
+            Panel: 'TimeSeriesChart',
+            TimeSeriesQuery: 'PrometheusTimeSeriesQuery',
+          }}
+        >
+          {isOpen && (
+            <DatasourceEditorForm
+              initialDatasource={datasource}
+              saveAction={saveAction}
+              flagAsDraft={() => setHasDraftChanges(true)}
+              onSave={handleSave}
+              onCancel={handleClose}
+              onDelete={saveAction == 'Update' ? () => setDeleteDatasourceDialogStateOpened(true) : undefined}
+            />
+          )}
+        </PluginRegistry>
+        <DeleteDatasourceDialog
+          open={isDeleteDatasourceDialogStateOpened}
+          onClose={() => setDeleteDatasourceDialogStateOpened(false)}
+          onDelete={handleDelete}
+          datasource={datasource}
+        />
+        <DiscardChangesConfirmationDialog
+          description="Are you sure you want to discard your changes? Changes cannot be recovered."
+          isOpen={isDiscardDialogStateOpened}
+          onCancel={() => setDiscardDialogStateOpened(false)}
+          onDiscardChanges={handleDiscard}
+        />
+      </ErrorBoundary>
+    </Drawer>
+  );
+}
--- a/ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
+++ b/ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
@@ -12,53 +12,36 @@
 // limitations under the License.
 
 import { Datasource } from '@perses-dev/core';
-import { DispatchWithoutAction, useState } from 'react';
-import { DiscardChangesConfirmationDialog, Drawer, ErrorAlert, ErrorBoundary } from '@perses-dev/components';
+import { DispatchWithoutAction, useCallback, useState } from 'react';
+import { Drawer, ErrorAlert, ErrorBoundary } from '@perses-dev/components';
 import { PluginRegistry } from '@perses-dev/plugin-system';
 import { bundledPluginLoader } from '../../model/bundled-plugins';
 import { DeleteDatasourceDialog } from '../dialogs/DeleteDatasourceDialog';
 import { DatasourceEditorForm } from './DatasourceEditorForm';
 
+export type ActionStr = 'Create' | 'Update';
+
 interface DatasourceDrawerProps {
   datasource: Datasource;
   isOpen: boolean;
-  saveAction: string;
+  saveActionStr: ActionStr;
   onSave: (datasource: Datasource) => void;
   onClose: DispatchWithoutAction;
 }
 
 export function DatasourceDrawer(props: DatasourceDrawerProps) {
-  const { datasource, isOpen, saveAction, onSave, onClose } = props;
+  const { datasource, isOpen, saveActionStr, onSave, onClose } = props;
   const [isDeleteDatasourceDialogStateOpened, setDeleteDatasourceDialogStateOpened] = useState<boolean>(false);
-  const [isDiscardDialogStateOpened, setDiscardDialogStateOpened] = useState<boolean>(false);
-  const [hasDraftChanges, setHasDraftChanges] = useState<boolean>(false);
 
-  const handleSave = (datasource: Datasource) => {
-    setHasDraftChanges(false);
-    onSave(datasource);
-  };
-
-  const handleClose = () => {
-    if (hasDraftChanges) {
-      setDiscardDialogStateOpened(true);
-    } else {
-      onClose();
-    }
-  };
-
-  const handleDelete = () => {
-    setHasDraftChanges(false);
-    onClose();
-  };
-
-  const handleDiscard = () => {
-    setHasDraftChanges(false);
-    setDiscardDialogStateOpened(false);
-    onClose();
-  };
+  // When user clicks out of the drawer, do not close it, just do nothing
+  // This is a quick-win solution to avoid losing draft changes
+  // -> TODO allow closing by clicking-out without being too sensitive to missclicks
+  const handleClickOut = useCallback(() => {
+    return;
+  }, []);
 
   return (
-    <Drawer isOpen={isOpen} onClose={handleClose} data-testid="datasource-editor">
+    <Drawer isOpen={isOpen} onClose={handleClickOut} data-testid="datasource-editor">
       <ErrorBoundary FallbackComponent={ErrorAlert}>
         <PluginRegistry
           pluginLoader={bundledPluginLoader}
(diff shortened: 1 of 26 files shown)


ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
```diff
@@ -0,0 +1,97 @@
...
+            />
+          )}
+        </PluginRegistry>
+        <DeleteDatasourceDialog
+          open={isDeleteDatasourceDialogStateOpened}
+          onClose={() => setDeleteDatasourceDialogStateOpened(false)}
+          onDelete={handleDelete}
+          datasource={datasource}
+        />
+        <DiscardChangesConfirmationDialog
```
celian-garcia, line 88: If you check on that PR https://github.com/perses/perses/pull/1290, I moved the discard change behavior into the VariableEditorForm. This can remove some duplication if you use it somewhere else and prevent you from using this flagAsDraft over there no ?

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/types.ts
```diff
@@ -0,0 +1,43 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { RequestHeaders } from '@perses-dev/core';
+
+// TODO unify this one with PrometheusDatasourceSpec from prometheus-datasource.tsx ? or rename one or the other for disambiguition?
+export interface PrometheusDatasourceSpec {
```
celian-garcia, line 17: Can you tell me again the difference between the two? You told me already but I forgot. This SpecFull thingy is a terrible name indeed :smile: I'm wondering if we can't find a more meaningful name.

celian-garcia (changes requested):
This is a technical review here. 

Generally I think you can do a kind of "visual rebase" from the last version of the components created by Guillaume as I think you took a lot of inspiration :stuck_out_tongue:, specially after the changes I made here https://github.com/perses/perses/pull/1290

ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
AntoineThebaud: I see that you went for a comparison based on JSON.stringify to detect if there was changes.. Yes I think it's better than the solution I did, I'll reuse this thanks

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,231 @@
...
+interface DatasourceEditorFormProps {
+  initialDatasource: Datasource;
+  saveAction: string;
+  onSave: (datasource: Datasource) => void;
+  onCancel: DispatchWithoutAction;
+  onDelete: DispatchWithoutAction | undefined;
+  flagAsDraft: DispatchWithoutAction;
+}
+
+export function DatasourceEditorForm(props: DatasourceEditorFormProps) {
```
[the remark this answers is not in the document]
AntoineThebaud: Yes this discard modal is pretty much the same as for variables, I just added this flagAsDraft based stuff to avoid triggering it when nothing was changed (since on my side there is no read-specific mode, the case could occur quite often)

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/PrometheusDatasourceEditor.tsx
```diff
@@ -0,0 +1,326 @@
...
+}
+
+export function PrometheusDatasourceEditor(props: PrometheusDatasourceEditorProps) {
+  const { value, onChange, isReadonly } = props;
+  const strDirect = 'Direct access';
+  const strProxy = 'Proxy';
+
+  // utilitary function used for headers when renaming a property
+  // TODO maybe there's an cleaner way to manage this case?
+  const buildNewHeaders = (oldHeaders: RequestHeaders | undefined, oldName: string, newName: string) => {
```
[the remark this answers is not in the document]
AntoineThebaud: Yes this is something I'd like to rework later, and in the meantime we could discuss with @Nexucis when he comes back about this datamodel change proposal. Anyway yes manipulating a list on the frontend side seems to be a better approach no matter what we decide for the model.

ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
celian-garcia: Not perfect maybe but that's the trick I found for deep compares. Maybe there would still be an issue if you have inner map ? I'm not sure if JSON.stringify orders maps by key by default or if it's kinda random here

ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
```diff
@@ -12,53 +12,36 @@
...
-
-  const handleDiscard = () => {
-    setHasDraftChanges(false);
-    setDiscardDialogStateOpened(false);
-    onClose();
-  };
+  // When user clicks out of the drawer, do not close it, just do nothing
+  // This is a quick-win solution to avoid losing draft changes
+  // -> TODO allow closing by clicking-out without being too sensitive to missclicks
+  const handleClickOut = useCallback(() => {
```
AntoineThebaud, line 39: As discussed in another place with @celian-garcia, moving the discard change logic into the DatasourceEditorForm component made it more complex to tackle the "click out of drawer" case, on which we would like the same behavior as the cancel button. In the current state it would mean moving around a datasource object between the drawer & the form component to achieve json comparison at the drawer level, which we would like to avoid.. Letting @celian-garcia think about a better solution in a next PR, for now going with this quick-win of just disabling it.

ui/e2e/src/pages/AppProjectPage.ts
```diff
@@ -127,4 +134,14 @@ export class AppProjectPage {
   getVariableEditor() {
     return new VariableEditor(this.variableEditor);
   }
+
+  async startCreatingDatasources() {
+    await this.addDatasourceButton.click();
```
sjcobb, line 139: Thanks for adding the e2e!

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,245 @@
...
+                    draft.spec.default = v.target.value === 'yes';
+                  });
+                }}
+                sx={{ width: '100%' }}
+              >
+                <MenuItem value="yes">Yes</MenuItem>
+                <MenuItem value="no">No</MenuItem>
+              </Select>
+            </FormControl>
+            <FormHelperText>
```
sjcobb, line 212: nit: Help text here gets tight with control above:
image

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,245 @@
...
+  Stack,
+  TextField,
+  Typography,
+} from '@mui/material';
+import { DispatchWithoutAction, useCallback, useMemo, useState } from 'react';
+import { PluginEditor } from '@perses-dev/plugin-system';
+import { DiscardChangesConfirmationDialog } from '@perses-dev/components';
+import { useIsReadonly } from '../../model/config-client';
+
+// TODO: Replace with proper validation library
```
sjcobb, line 35: cc @Gladorme @celian-garcia  I kind of alluded to this in the sync, but if you guys want to try out a form state / validation library in the app, I think that'd be cool. We were trying to avoid it last time this came up until we could do some further research, but these forms are getting more and more complex

If we find one we like, could also potentially use on dashboards later. As far as validation, this [zod](https://github.com/colinhacks/zod) lib has been on my radar and it integrates with a bunch of form packages: https://github.com/colinhacks/zod#form-integrations

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,245 @@
...
+    spec: {
+      ...datasource.spec,
+      display: patchedDisplay,
+    },
+  };
+}
+
+interface DatasourceEditorFormProps {
+  initialDatasource: Datasource;
+  saveAction: string;
```
sjcobb, line 78: Something like `saveActionText` would be a bit more descriptive

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,245 @@
...
+    if (JSON.stringify(initialState) !== JSON.stringify(state)) {
+      setDiscardDialogStateOpened(true);
+    } else {
+      onClose();
+    }
+  }, [state, initialState, setDiscardDialogStateOpened, onClose]);
+
+  return (
+    <>
+      <Box
```
sjcobb, line 114: I could see some nice-to-have refactoring to break up this component in the future, also some of the spacing between form controls looks a little tight in places and slightly further apart in others:
image

sjcobb:
Created a new datasource and brand new dashboard, noticed a bunch of errors in the console, but totally possible some were already there before:
image
image
Full console errors:
[debug_log.txt](https://github.com/perses/perses/files/12099913/debug_log.txt)

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/PrometheusDatasourceEditor.tsx
```diff
@@ -0,0 +1,327 @@
...
+                  <Grid item xs={4}>
+                    <TextField
+                      fullWidth
+                      label="Header value"
+                      value={value.proxy?.spec.headers?.[headerName]}
+                      InputProps={{
+                        readOnly: isReadonly,
+                      }}
+                      onChange={(e) =>
+                        onChange({
```
sjcobb, line 170: Could try immer in these forms too to avoid all the spread operator usage, see `useImmer` in VariableEditForm as example

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/types.ts
sjcobb: I'm not crazy about the `PrometheusDatasourceSpecFull` alias name either but I don't have a great solution on how these types could be refactored right now

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/PrometheusDatasourceEditor.tsx
AntoineThebaud: at first I used it but the performance where quite bad, maybe not because of useImmer specifically though since I changed other things since. I will try again to use it & see.

ui/app/src/components/DatasourceList/DatasourceDrawer.tsx
celian-garcia: ack :+1: I will propose a refactor of that in the following days

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
celian-garcia: Yeah I have no knowledge/opinion on React ecosystem. I would trust you.
This zod seems powerful. Reading the doc an idea of having the same validation on Backend and Frontend came to mind. But maybe I'm overestimating the tool :smile:

celian-garcia: Because developing in the UI, I've been sometimes facing 400 issues that were not discovered by the UI so if any mechanism could ensure the validations in the backend are the same as the validation in the frontend, that would be :fire:

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/PrometheusDatasourceEditor.tsx
sjcobb: Yeah worth a try, but depending what comes out of [this conversation](https://github.com/perses/perses/pull/1282#discussion_r1268626997) about form / validation libraries we could wait and refactor later I guess

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
sjcobb: +1 definitely would be great, I've [brought this up](https://matrix.to/#/!XNmgYIAndZOkEvQAbb:matrix.org/$WA_zjNYIr4v0lGOryD9NbF6QFcyWq1b4gEoB6F8ebGM?via=matrix.org&via=herail.net) in #perses-dev too, there would be some friction in keeping the two schemas in sync, but it might be worth it regardless. Orr you / @AntoineThebaud should write a CUE -> Zod converter! 😄  Surprised there isn't one yet, it already has integrations to convert from all of these: https://github.com/colinhacks/zod#x-to-zod

AntoineThebaud:
> Created a new datasource and brand new dashboard, noticed a bunch of errors in the console, but totally possible some were already there before: image image Full console errors: [debug_log.txt](https://github.com/perses/perses/files/12099913/debug_log.txt)

Yes I checked and this is not related to this PR. About the "document not found" error, it seems to trigger everytime you create a new dashboard, it seems we are right away trying to get the dashboard from the API while it wasnt saved/registered yet.

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
AntoineThebaud: This form control was changed to a switch, the spacing is better now :)

image

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
AntoineThebaud: Overall spacing has been improved with latest changes: 
image

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/PrometheusDatasourceEditor.tsx
AntoineThebaud: Yes if you don't mind I'd like to tackle this one in a follow-up PR.

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,245 @@
...
+                  }
+                });
+              }}
+            />
+          </Grid>
+          <Grid item xs={6}>
+            <FormControl fullWidth>
+              {/* TODO: How to ensure ids are unique? */}
+              <InputLabel id="datasource-default-label">Default</InputLabel>
+              <Select
```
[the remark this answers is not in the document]
AntoineThebaud: image

AntoineThebaud:
@sjcobb I have to adjust the e2e test due to form control change, otherwise it's ready for re-review ;) https://github.com/perses/perses/pull/1282/commits/1360c770d8d6c9db2010b9abc3de515578abeaa4

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
```diff
@@ -0,0 +1,228 @@
...
+              onChange={(v) => {
+                setState((draft) => {
+                  if (draft.spec.display) {
+                    draft.spec.display.description = v.target.value;
+                  }
+                });
+              }}
+            />
+          </Grid>
+          <Grid item xs={6} sx={{ paddingTop: '5px !important' }}>
```
AntoineThebaud, line 178: 'would have like avoiding using !important here but didn't find an easy way... it seems quite some people already complained about this https://github.com/mui/material-ui/issues/8509

ui/app/src/components/DatasourceList/DatasourceEditorForm.tsx
celian-garcia: Yeah maintaining two schemas or maintaining two piece of validation code, I prefer the first :smile: 

Also, we could imagine that the validation is all part of the backend. I think it would be worth it to consider an approach where the frontend validation consist in making a debounced query to the backend (making the query only if you edit and don´t touch for one sec for example)

=> The first solution with zod would be ok if we only speak about static validation, but for dynamic validation it may be more complex.

The Variable injection validation that I'm actually not doing here in this frontend PR is a very good example of dynamic validation that won't be possible with schemas. In the backend we are doing a dynamic validation checking that the variables mentioned in the dashboard are well existing (as external or local vars). But we're not doing that check in the frontend.

celian-garcia (approved):
LGTM technically. I let others do the styles/functional review.

sjcobb (approved):
Will be good to iterate in future PRs, excited to start using it!

PR #1284, opened 2023-07-09
[FEATURE] Time series bar support using `visual.display` in TimeSeriesChart spec

sjcobb:
## Overview
Adds ability to have a time series bar chart type using existing TimeSeriesChart panel.
The new optional property added to the spec is `visual.display: 'line' | 'bar'` (with the default being 'line' if undefined)

## Screenshot
Screenshot 2023-07-08 at 10 11 53 PM

### DIFF
--- a/schemas/panels/time-series/time-series.cue
+++ b/schemas/panels/time-series/time-series.cue
@@ -31,9 +31,10 @@ import (
 }
 
 #visual: {
+	display?:       "line" | "bar"
 	line_width?:    number & >=0.25 & <=3
 	area_opacity?:  number & >=0 & <=1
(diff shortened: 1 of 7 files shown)


saminzadeh (approved):
Awesome - I assume this also works with stacked bar?

eunicorn (approved):

sjcobb:
> Awesome - I assume this also works with stacked bar?

Yeah it does for the most part, our custom tooltip sometimes isn’t exactly right though, it’s possible to fix but I didn’t look more into it—worst case I could show the ECharts `trigger: ‘item’` tooltip when ‘bar’ is selected instead

PR #1286, opened 2023-07-09
[ENHANCEMENT] Clean up legacy LineChart conditions in TimeSeriesChartPanel

sjcobb:
Now that v0.37.0 has been published, this is a follow up to #1268 to remove temporary `show_legacy_chart` and all dead code in TimeSeriesChart panel that is no longer needed.

All component-level types, functions, components necessary for LineChart are kept since there are many instance where embed users rely on LineChart. Deprecated comment warnings have been added to nudge new users to try TimeChart instead.


ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -253,6 +250,7 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
     noDataVariant,
     timeZone,
   ]);
+  console.log({ option });
```
eunicorn, line 253: remove?

eunicorn (approved):
thanks!

PR #1289, opened 2023-07-10
[FEATURE] Add option to specify the portal container for the chart tooltip

eunicorn:
By default, the chart tooltip is attached to `document.body` using mui portal, which can lead to the pinned tooltip moving when scrolling is applied to a container element instead of the entire document. To address this issue, this PR adds the option to specify the container element to which the chart tooltip's MUI portal should be attached.


sjcobb (approved):
🙏

sjcobb:
Thanks for fixing! Not a big deal since can cherry-pick, but it’ll be slightly easier to do the v0.37.1 patch release if you make this PR into release/v0.37 branch instead of main

PR #1294, opened 2023-07-12
[BUGFIX] Fix tooltip pinning when apps set a `tooltipPortalContainerId`

eunicorn:
This PR fixes 2 bugs:

1. The position of tooltip is recalculated every time it renders, causing the tooltip to move even after it is pinned. To fix this, we need to store the position of tooltip in a ref. 
3. Tooltip is cut off if it is attached to a different element other than document.body. To fix this, we have to make sure the tooltip stays within the element boundaries. 

More info about using `tooltipPortalContainerId` in PR #1289

### DIFF
--- a/ui/components/src/TimeSeriesTooltip/TimeChartTooltip.tsx
+++ b/ui/components/src/TimeSeriesTooltip/TimeChartTooltip.tsx
@@ -17,7 +17,7 @@ import { ECharts as EChartsInstance } from 'echarts/core';
 import { UnitOptions, TimeSeries } from '@perses-dev/core';
 import useResizeObserver from 'use-resize-observer';
 import { TimeChartSeriesMapping } from '../model';
-import { CursorCoordinates, FALLBACK_CHART_WIDTH, TOOLTIP_PADDING, useMousePosition } from './tooltip-model';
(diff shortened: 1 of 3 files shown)


ui/components/src/TimeSeriesTooltip/TimeChartTooltip.tsx
```diff
@@ -63,7 +65,12 @@ export const TimeChartTooltip = memo(function TimeChartTooltip({
   const chartWidth = chart?.getWidth() ?? FALLBACK_CHART_WIDTH; // Fallback width not likely to ever be needed.
 
   const containerElement = containerId ? document.querySelector(containerId) : undefined;
-  const cursorTransform = assembleTransform(mousePos, chartWidth, pinnedPos, height ?? 0, width ?? 0, containerElement);
+  // if tooltip is attached to a container, set max height to the height of the container so tooltip does not get cut off
+  const maxHeight = containerElement ? containerElement.getBoundingClientRect().height : undefined;
```
sjcobb, line 69: Nice! This is a really cool solution and I like that it's opt-in so if containerId isn't set

sjcobb (approved):
🎉

PR #1296, opened 2023-07-13
[BUGFIX] Time chart fix for incorrect connected values 

sjcobb:
## Before issue where data was connected
http://localhost:3000/projects/perses/dashboards/TimeChart_vs_LineChart?start=1689236880000&refresh=0s&end=1689240690000
image

## After fix with nulls filled in so empty space is shown correctly
http://localhost:3000/projects/perses/dashboards/TimeChart_vs_LineChart?start=1689236880000&refresh=0s&end=1689240690000
image

### DIFF
--- a/ui/core/src/utils/time-series-data.ts
+++ b/ui/core/src/utils/time-series-data.ts
@@ -30,34 +30,39 @@ export function getXValues(timeScale: TimeScale): number[] {
   return xValues;
 }
 
-export type PromValue = [number, string | null];
+/**
+ * Given a TimeSeries from a query and a common time scale (see `getCommonTimeScale`),
+ * process values and checks for 'NaN', while filling in any timestamps that are
+ * missing from the time series data with `null` values.
+ */
+export function getTimeSeriesValues(series: TimeSeries, timeScale: TimeScale) {
+  let timestamp = timeScale.startMs;
 
-export function getTimeSeriesValues(values: TimeSeriesValueTuple[], timeScale: TimeScale) {
-  const inputValues = values as PromValue[];
-  const completeResponse: PromValue[] = [];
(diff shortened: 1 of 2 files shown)


ui/core/src/utils/time-series-data.ts
```diff
@@ -30,6 +30,38 @@ export function getXValues(timeScale: TimeScale): number[] {
   return xValues;
 }
 
+/**
+ * Given a TimeSeries from a query and a common time scale (see `getCommonTimeScale`),
+ * processes the time series values, filling in any timestamps that are missing
+ * from the time series data with `null` values.
+ */
+export function getTimeSeriesValues(series: TimeSeries, timeScale: TimeScale) {
```
sjcobb, line 38: Definitely ways we can improve this, but this approach ports how [getYValues](https://github.com/perses/perses/blob/0fcb632ff4e7e4ab3ae60f53dd98886302f9f83f/ui/core/src/utils/time-series-data.ts#L70C17-L84) was implemented before, so now `show_legacy_chart` and the new TimeChart render the same

eunicorn (approved):

ui/core/src/utils/time-series-data.ts
sjcobb: Planning to write better tests for this in a follow up, mocking a similar query to the one where we caught this bug with msw and a visual regression test should do it. Merging for now to get this patch release out quickly

sjcobb: Moving this and the getCommonTimeScaleForQueries logic into DataQueriesProvider and merging / processing the data before it gets to TimeSeriesChartPanel is also worth looking into

PR #1300, opened 2023-07-16
[WIP] Closest datapoint styles and pinned crosshair remains after cursor leaves chart 

sjcobb:
## Summary
- Crosshair remains in place when tooltip is pinned instead of disappearing
- Closest datapoint to cursor 'select' styles
- Do not snap to nearest datapoint in xAxis.axisPointer options

image

### DIFF
--- a/ui/components/src/TimeChart/TimeChart.tsx
+++ b/ui/components/src/TimeChart/TimeChart.tsx
@@ -38,7 +38,7 @@ import {
 } from 'echarts/components';
 import { CanvasRenderer } from 'echarts/renderers';
 import { EChart, OnEventsType } from '../EChart';
-import { ChartInstanceFocusOpts, ChartInstance, TimeChartSeriesMapping } from '../model/graph';
+import { ChartInstanceFocusOpts, ChartInstance, TimeChartSeriesMapping, TimeSeriesOption } from '../model/graph';
 import { useChartsTheme } from '../context/ChartsThemeProvider';
 import {
   clearHighlightedSeries,
@@ -238,6 +238,7 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
     if (__experimentalEChartsOptionsOverride) {
       return __experimentalEChartsOptionsOverride(option);
     }
+
     return option;
   }, [
     data,
@@ -256,8 +257,46 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
     <Box
       sx={{ height }}
       onClick={(e) => {
+        // https://echarts.apache.org/en/api.html#echartsInstance.convertFromPixel
+        const pointInPixel = [e.nativeEvent.offsetX ?? 0, e.nativeEvent.offsetY ?? 0];
+        if (chartRef.current !== undefined && chartRef.current.containPixel('grid', pointInPixel)) {
+          const pointInGrid: number[] = chartRef.current.convertFromPixel('grid', pointInPixel);
+          console.log('pointInGrid -> ', pointInGrid);
+        }
+
         // Pin and unpin when clicking on chart canvas but not tooltip text.
         if (e.target instanceof HTMLCanvasElement) {
+          // https://github.com/perses/perses/compare/main...sjcobb/tooltip-pin-attach-mark-line-init
+          const pinnedCrosshair: LineSeriesOption = {
+            name: 'Pinned Crosshair',
+            type: 'line',
+            // data: [[startTime, null] as unknown],
+            // data: [1689444705000, 0.02],
+            // data: [timeScale.startMs, 0.02],
+            markLine: {
+              symbol: 'none',
+              symbolSize: 0,
+              itemStyle: {
+                color: '#eee',
+              },
+              data: [
+                {
+                  // xAxis: 1689444705000,
+                  // yAxis: 0.02,
+                  xAxis: timeScale.startMs,
+                },
+              ],
+              // data: [timeScale.startMs, 0.02],
+              lineStyle: {
+                width: 1,
+                type: 'dashed',
+              },
+              label: {
+                // distance: [20, 8],
+              },
+            },
+          };
+          seriesMapping.push(pinnedCrosshair as TimeSeriesOption);
           setTooltipPinnedCoords((current) => {
             if (current === null) {
               return {
--- a/ui/components/src/TimeChart/TimeChart.tsx
+++ b/ui/components/src/TimeChart/TimeChart.tsx
@@ -44,6 +44,7 @@ import {
   clearHighlightedSeries,
   enableDataZoom,
   getFormattedAxisLabel,
+  getPointInGrid,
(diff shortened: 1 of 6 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -264,12 +265,13 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
 
-        // generates array of [time, step.value] where time ranges from timescale.startMs to timescale.endMs with an interval of 30s
+        // Generates array of [time, step.value] where time ranges from timescale.startMs to timescale.endMs with an interval of 15s
         const thresholdTimeValueTuple: TimeSeriesValueTuple[] = [];
         let currentTimestamp = timeScale.startMs;
         while (currentTimestamp <= timeScale.endMs) {
           thresholdTimeValueTuple.push([currentTimestamp, stepOption.value]);
-          currentTimestamp += 1000 * 30;
+          // Used to plot fake thresholds datapoints so correct nearby threshold series shows in tooltip without flicker
+          currentTimestamp += 1000 * THRESHOLD_PLOT_INTERVAL;
```
sjcobb, line 274: @eunicorn I reduced this interval from 30s to 15s since moving your cursor horizontally close to a threshold line had this flicker before (probably should have made a separate PR but noticed it when testing datapoint hover state): 

https://github.com/perses/perses/assets/9369625/1ac02a71-3d34-4da5-9723-b617b23b337a

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -254,13 +267,38 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
...
       sx={{ height }}
       onClick={(e) => {
+        // Determine where on chart canvas to plot pinned crosshair as markLine.
+        const pointInGrid = getPointInGrid(e.nativeEvent.offsetX, e.nativeEvent.offsetY, chartRef.current);
+        if (pointInGrid === null) {
+          return;
+        }
+
+        // Clear previously set pinned crosshair
+        const isCrosshairPinned = seriesMapping[seriesMapping.length - 1]?.name === 'Pinned Crosshair';
```
sjcobb, line 285: Sorry for the large PR, these ~15 lines is the majority of the pinned crosshair implementation. When I added this and disabled xAxis.axisPointer.snap, I realized we _really_ needed the datapoint hover state too to complete this feature (or else when the crosshair was pinned, you could not necessarily tell which datapoint on the left or right was actually showing in the tooltip).

sjcobb: @foxbat07 also asked for datapoint symbols to have a hover state when close to your cursor

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -254,13 +267,38 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
...
     noDataVariant,
     timeZone,
+    tooltipPinnedCoords,
     isStackedBar,
   ]);
 
   return (
     <Box
       sx={{ height }}
       onClick={(e) => {
```
sjcobb, line 277: Another option could be to use approach here: https://apache.github.io/echarts-handbook/en/concepts/event/#listen-to-events-on-the-blank-area

Using regular `onClick` the way we've been doing it seems like it's more straightforward for React devs than using `chart.getZr().on('click', (e) => {});` but open to feedback or other ideas. 

There are also these out-of-the-box [events](https://echarts.apache.org/en/api.html#events), but they mostly trigger when interacting with lines / datapoint symbols instead of clicking anywhere on the canvas like we need for pinned tooltip behavior

saminzadeh (approved):
Just played around with the branch, seems sweet. I didn't do a thorough review of the logic but the general approach seems sound for an MVP we can iterate on.

ui/components/src/TimeChart/TimeChart.tsx
eunicorn: > it became more clear that we really need the datapoint hover state to complete this feature

+1, can we keep xaxis snaping until we implement the data point hover state? otherwise it's hard to tell the relationship between the tooltip and the data point it correlates

cc @saminzadeh

ui/components/src/TimeChart/TimeChart.tsx
eunicorn: wonder if we can reuse the logic in the tooltip to find the nearest data point and snap the pinned crosshair there?

ui/components/src/model/graph.ts
```diff
@@ -68,3 +68,35 @@ export type ChartInstance = {
    */
   clearHighlightedSeries: () => void;
 };
+
+export const DEFAULT_PINNED_CROSSHAIR: LineSeriesOption = {
+  name: 'Pinned Crosshair',
```
eunicorn, line 73: maybe good to use a variable here since it's used in TimeChart.tsx

ui/components/src/TimeChart/TimeChart.tsx
sjcobb: Well datapoint hover state is part of this PR! But maybe I need to tweak that so it’s more clear (symbols get larger) 🤔

For longer time ranges in seemed like hiding them was okay but could also show them more often, ideally they would show only on hover (that is end goal but was hoping this PR was adequate for now)

ui/components/src/TimeChart/TimeChart.tsx
sjcobb: Would adding back snapping help with that though?

ui/components/src/TimeChart/TimeChart.tsx
eunicorn: ohh i was looking at last 6 hrs, so maybe that was why i didn't notice the hover state

ui/components/src/TimeChart/TimeChart.tsx
eunicorn: i don't think so because i think this is what we want ideally:

1. while hovering the graph, the crosshair doesn't snap but the nearest data points are highlighted
2. the crosshair only snaps to the nearest data point when user pins the tooltip 

does this sound right?

ui/components/src/utils/theme-gen.ts
```diff
@@ -172,6 +172,22 @@ export function generateChartsTheme(
...
+    axisPointer: {
+      lineStyle: {
+        color: muiTheme.palette.grey[500],
+      },
+    },
+    markLine: {
+      symbol: 'none',
+      symbolSize: 0,
+      itemStyle: {
+        color: muiTheme.palette.grey[700],
```
eunicorn, line 184: i think we should use a different color, something more transparent?

right now it seems too bold in light mode

image

eunicorn (approved):

PR #1305, opened 2023-07-18
[BUGFIX] Time series stacked bar correct palette color and tooltip

sjcobb:
## Summary

- Fix palette colors for TimeSeriesChart `visual.display: 'bar'`
- Adjust charts theme to account for vertical bar charts
- Fix stacked bar tooltip by using ECharts tooltip when visual.stack is set

### Before palette and charts theme fixes
image

### After change to ECharts `trigger: 'item'` tooltip

https://github.com/perses/perses/assets/9369625/b6490658-eb57-43d3-a467-faecd26f9db6

### After palette fix by removing itemStyle.color from charts theme-gen
image

### DIFF
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -66,7 +66,13 @@ export function BarChart(props: BarChartProps) {
       series: {
         type: 'bar',
         label: {
+          show: true,
+          position: 'right',
           formatter: (params: { data: number[] }) => params.data[1] && formatValue(params.data[1], unit),
+        },
+        itemStyle: {
+          borderRadius: 4,
+          color: chartsTheme.echartsTheme[0],
         },
       },
       tooltip: {
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -208,8 +208,13 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
         const showTimeSeries = isSelected || isSelectAll;
 
         if (showTimeSeries) {
+          // Use timeChartData.length to ensure the data that is passed into the tooltip accounts for
+          // which legend items are selected.
+          const datasetIndex = timeChartData.length;
+          // Each series is stored as a separate dataset source.
+          // https://apache.github.io/echarts-handbook/en/concepts/dataset/#how-to-reference-several-datasets
           timeSeriesMapping.push(
-            getTimeSeries(seriesId, timeChartData.length, formattedSeriesName, visual, timeScale, seriesColor)
+            getTimeSeries(seriesId, datasetIndex, formattedSeriesName, visual, timeScale, seriesColor)
           );
 
           timeChartData.push({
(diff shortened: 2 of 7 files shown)


ui/components/src/BarChart/BarChart.tsx
```diff
@@ -66,8 +66,14 @@ export function BarChart(props: BarChartProps) {
       series: {
         type: 'bar',
         label: {
+          show: true,
+          position: 'right',
```
sjcobb, line 70: Moving this back to BarChart since wasn't thinking about how it would impact vertical bar charts when I suggested adding it to chartsTheme in #1279 (sorry @ellenwyllie!). And moving color back as well so that TimeSeriesChart stacked bar loops through the palette again

ellenwyllie (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -358,6 +363,10 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
     setTimeRange({ start: new Date(event.start), end: new Date(event.end) });
   };
 
+  const isStackedBar = visual.display === 'bar' && visual.stack === 'All';
+
+  const tooltipConfig: TooltipConfig = { wrapLabels: true };
```
juliepagano, line 368: If this object is constant, it might be beneficial to pull the variable definition outside the render function. That way react always sees it as the same value (as opposed to it being recreated on every render cycle).

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -215,10 +209,21 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
         const showTimeSeries = isSelected || isSelectAll;
 
         if (showTimeSeries) {
+          // Use timeChartData.length to ensure the data that is passed into the tooltip accounts for
+          // which legend items are selected.
+          const datasetIndex = timeChartData.length;
```
juliepagano, line 214: Optional suggestion: It took me a minute to figure out what was going on here. It might be worth clarifying in the comment that this _must_ happen before the `timeChartData.push` below to avoid an off-by-one error. I could see someone accidentally breaking this while trying to refactor the component. It's also possible this is just "still getting back into work after a 2 week vacation" brain, so no big deal either way.

juliepagano (approved):
A couple optional suggestions, but generally LGTM. 🎉

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Thanks this is fixed, added DEFAULT_TOOLTIP_CONFIG and can merge in future spec.tooltip properties, going to add one to disable tooltip pinning in a follow up

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Added the comment! By the way the diff here was out of date, this PR was originally going into branch for #1302 (which was merged into main this morning)

That PR has more info about why this change was needed, after the LineChart -> TimeChart refactor, clicking a legend item would cause tooltip to not render since it was looping through entire dataset instead of just the filtered data based on selected legend items

PR #1306, opened 2023-07-18
[FEATURE] Consider external vars in dashboard validation

celian-garcia:
Backend should take in account the external variables (project/global variables) when it is verifying that the variables used by a dashboard are defined.

This is the case for all validation by endpoints (on dashboard create/update + on each call of the /validate endpoint).

The static validation done by percli lint (without --online) doesn't assume on any external variable. It may fail if somebody try to use it to validate a dashboard which uses external variables.

# Checklist

### DIFF
--- a/pkg/model/api/v1/utils/variable_build_order.go
+++ b/pkg/model/api/v1/utils/variable_build_order.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package dashboard
+package utils
 
 import (
 	"fmt"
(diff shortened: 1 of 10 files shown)


pkg/model/api/v1/utils/variable_build_order.go
```diff
@@ -54,37 +58,107 @@ func buildGraph(variables []Variable) (*graph, error) {
...
-		var matches [][]string
-		switch variableSpec := variable.Spec.(type) {
-		case *TextVariableSpec:
-			matches = parseVariableUsed(variableSpec.Value)
-		case *ListVariableSpec:
-			matches = findAllVariableUsedInPlugin(variableSpec.Plugin)
-		}
-		deps := make(map[string]bool)
+
+	// Declare some
```
AntoineThebaud, line 78: unfinished comment

AntoineThebaud (approved):

PR #1309, opened 2023-07-19
[BUGFIX] Fix tooltip pinning in Line Chart

eunicorn:
Same fix as https://github.com/perses/perses/pull/1294 but for Line Chart


sjcobb (approved):
🙏

PR #1313, opened 2023-07-20
[ENHANCEMENT] Ability to disable tooltip pinning in panel spec 

sjcobb:
## Summary

- Adds ability to opt out tooltip pinning in TimeSeriesChart panel using `spec.tooltip.enable_pinning`
- Can also opt out in lower level TimeChart using prop: `tooltipConfig.enablePinning`
- `tooltipConfig.enableEChartsTooltip` also added to replace `isStackedBar` prop since it is more descriptive and useful generally

## TODO for follow up
- [ ] rename ChartsThemeProvider -> ChartsProvider
- [ ] add ability to disable tooltip pinning and shared crosshair app wide 
- [ ] add ability to only pin one tooltip at a time

## Screenshot 
image


ui/components/src/LineChart/LineChart.tsx
```diff
@@ -88,7 +88,7 @@ export const LineChart = forwardRef<ChartInstance, LineChartProps>(function Line
     unit,
     grid,
     legend,
-    tooltipConfig = { wrapLabels: true },
+    tooltipConfig = { wrapLabels: true, enablePinning: true },
```
juliepagano, line 91: Should this one also use `DEFAULT_TOOLTIP_CONFIG`?

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -85,7 +85,6 @@ export interface TimeChartProps {
   tooltipConfig?: TooltipConfig;
   noDataVariant?: 'chart' | 'message';
   syncGroup?: string;
-  isStackedBar?: boolean;
```
juliepagano, line 88: Does the removal of this prop justify making this PR a breaking change or is this such a new feature that it's fine not to worry about it?

ui/components/src/TimeSeriesTooltip/tooltip-model.ts
```diff
@@ -135,4 +135,16 @@ export const useMousePosition = (): CursorData['coords'] => {
 export type TooltipConfig = {
   wrapLabels: boolean;
   hidden?: boolean;
+  enablePinning?: boolean;
+  enableEChartsTooltip?: boolean;
```
juliepagano, line 139: Does this prop differentiate whether to use the tooltip baked into echarts vs. the perses-based react tooltip or is it something a little different from that? If that is what it is, I wonder if this would be a little clearer (and enable supporting other tooltip kinds in the future) if it was an enum instead. Maybe something like `tooltipVariant?: 'echarts' | 'perses'` or something similar?  I'd also be cool with putting a typedoc comment above here clarifying the behavior, which is a lot less work.

juliepagano (commented):
Quick first pass before I need to run for the day, but I didn't get to everything. I'll try to give this another pass first thing tomorrow morning. If you get a 👍 from Eunice before I get to it, feel free to merge without that second pass since y'all are both on east coast time, and I don't want to be a blocker.

juliepagano:
I noticed that when I'm inside the panel editor, the pinning doesn't seem to work right. The line stays there, but not the tooltip. Is this a bug in your PR or potentially a separate issue?

image

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -372,6 +373,11 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
   };
 
   const isStackedBar = visual.display === 'bar' && visual.stack === 'All';
+  const tooltipConfig: TooltipConfig = {
+    ...DEFAULT_TOOLTIP_CONFIG,
+    enablePinning: tooltip?.enable_pinning ?? true,
+    enableEChartsTooltip: isStackedBar,
```
juliepagano, line 379: It might be worth adding a comment here explaining why we use the native echarts tooltip for stacked bars. I'm personally curious why since I missed those code changes while I was out on vacation, and I could see other folks being curious in the future when looking at the code.

juliepagano (commented):
One more small nit and I also left a top-level comment about a potential bug. The code changes generally look ok to me, so I'll give a 👍  when the bug is fixed (or clarified that it's a preexisting issue unrelated to your PR).

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -227,9 +225,9 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
       animation: false,
       tooltip: {
         show: true,
-        trigger: isStackedBar ? 'item' : 'axis',
-        // ECharts tooltip content hidden since we use custom tooltip instead
-        showContent: isStackedBar,
+        trigger: tooltipConfig.enableEChartsTooltip ? 'item' : 'axis',
```
eunicorn, line 228: i think maybe we should still keep `isStackedBar`, otherwise it seems confusing why trigger is set to item when enableEChartsTooltip is true

eunicorn (approved):

sjcobb:
> I noticed that when I'm inside the panel editor, the pinning doesn't seem to work right. The line stays there, but not the tooltip. Is this a bug in your PR or potentially a separate issue?

Yeah that's one reason I'm making this PR as well as the ChartsProvider changes in my other WIP branch #1317 so that we can disable tooltip pinning when inside modals/drawers. We have one case internally where we'll use `spec.enable_pinning: false` and then for panel preview we can wrap it with another `<ChartsProvider enabledPinning={false}>`. Also, might change so tooltip pinning is opt-in, it works well in the Perses app but other apps can have odd CSS that makes finding the correct container to use with Portal tricky

ui/components/src/TimeChart/TimeChart.tsx
sjcobb: I can change back, we had another request internally before where I suggested using the ECharts tooltip with `__experimentalEChartsOptionsOverride` so thought having a more explicit way to do that could be useful (they only ever had at most 2 series and needed multi y axis support which we don't support yet)

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: So the custom tooltip didn’t always reflect the subgroup you were hovering over accurately, there’s a way to fix it but just using the ECharts tooltip in this case was less work and since you shouldn’t really use that chart type with a ton of series—the nearby-series / highlighting logic isn’t as important

PR #1315, opened 2023-07-20
[FEATURE] Ability to customize panel extra content

saminzadeh:
This adds a "slot" to the Dashboard Panel so consumers can customize how they want to render extra information in the Panel header. This enables embedded use-cases so they can provide their own links, actions, etc to their external systems.

For example:
```tsx
function CustomPanelOptions() {
  const ctx = useDataQueries(); // access to data context too!
  const resolvedQueryString = ctx.queryResults?.[0]?.data?.metadata.resolvedQuery;
  return (
    <div>
      <a href={`http://my-special-logs.com/query?q=${resolvedQueryString}`>View Logs</a>
    </div>
  );
}

function MyApp() {
  return (
    <Dashboard
      panelOptions={{
        extra: CustomPanelOptions,
      }}
      emptyDashboardProps={{
        onEditButtonClick,
        ...emptyDashboardProps,
      }}
    />
  );
}

```

image

Note: This will only render when the dashboard is not in edit state.

# Checklist

## UI Changes


ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
```diff
@@ -105,6 +105,7 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
     }),
     metadata: {
       notices,
+      resolvedQuery: query,
```
saminzadeh, line 108: Some extra metadata that plugins can set that has the resolvedQuery string with interpolated variables. Similar to Grafana's `exectedQueryString`
https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/data.ts#L85
Probably should rename to that?

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -23,13 +23,18 @@ import { PanelContent } from './PanelContent';
 export interface PanelProps extends CardProps<'section'> {
   definition: PanelDefinition;
   editHandlers?: PanelHeaderProps['editHandlers'];
+  panelOptions?: PanelOptions;
 }
 
+export type PanelOptions = {
+  extra?: () => React.ReactNode;
```
saminzadeh, line 30: Named this prop `extra` which was inspired by Ant Design's extra prop [[1]](https://ant.design/components/card#api) [[2]](https://ant.design/components/drawer#api) they add to their components
image

However, I'm open to another name but figured I copy a popular component library paradigm.

juliepagano: I could see us wanting to create some more granular configuration props in the future, but for now, the generic `extra` seems ok to me. I do think it's worth adding a jsdoc above it clarifying what `extra` is and where the result of the function ends up rendering -- just something short like what Ant has in the description you screenshotted.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
juliepagano: No strong feelings about the name, but this is another one that could probably benefit from a little jsdoc above the type def to provide a little more context for consumers about what it contains and why they might want to use it.

ui/dashboards/src/components/Dashboard/Dashboard.tsx
```diff
@@ -16,20 +16,21 @@ import { ErrorBoundary, ErrorAlert } from '@perses-dev/components';
...
-
+import { PanelOptions } from '../Panel';
 export type DashboardProps = BoxProps & {
   /**
    * Props for `EmptyDashboard` component that will be rendered when the dashboard
    * is empty (i.e. has no panel groups). If not specified, the defaults will
    * be used.
    */
   emptyDashboardProps?: EmptyDashboardProps;
+  panelOptions?: PanelOptions;
```
juliepagano, line 27: Future-looking question: do you think there's ever a case where a dashboard might want more granular panel options (e.g. by panel type or even at the individual panel level) or do you anticipate it always being global? I don't think it's worth spending time on a ton of complex premature optimization here, so mostly a curiosity question.

juliepagano (commented):
Code generally looks ok to me besides some requests to add clarifying jsdoc, but I wasn't sure how to test it in action without modifying the code. Is there a min viable storybook story you could add to show consumers how to use this and what the result looks like?

ui/dashboards/src/components/Panel/Panel.tsx
saminzadeh: Great call - I added the JSdoc!

ui/dashboards/src/components/Dashboard/Dashboard.tsx
saminzadeh: Definitely, happy you thought about that as well. 

I left it out in this PR but I was thinking of injecting some context like the `panelDefinition` that would be all optional. Fortunately, we have some context hooks to get some info (eg. the TimeSeriesQuery through `useDataQueries()`) but I think we may need to inject some via args.

```tsx
type PanelOptions = {
 extra: (args: { panelDefinition?: PanelDefinition }) => React.Node
}
```

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
saminzadeh: Decided to use the same thing as grafana `executedQueryString` (no good reason to be different IMO). I also added a jsdoc comment

juliepagano (approved):
LGTM! Thanks for adding the extra story. 🚀

Issue #1303, opened 2023-07-17
Add `supportedQueryInterfaces` support to PanelPlugin interface


sjcobb:
See 'Decoupling data querying from panel plugin' discussion #956, this feature is follow up now that we have `DataQueriesProvider`. 

Currently, we only have TimeSeriesQuery plugins that are used in our core panels (TimeSeriesChart, GaugeChart, StatChart). We also want the ability to add more query plugins such as TraceQuery and LogQuery. To support these we need a way for a panel plugin to specify which interfaces it supports (see PR #922 for more info in [comment here](https://github.com/perses/perses/pull/922#discussion_r1080546854))

```ts
export interface PanelPlugin<Spec = UnknownSpec> extends Plugin<Spec> {
  PanelComponent: React.ComponentType<PanelProps<Spec>>;
  supportedQueryInterfaces?: Array<'TimeSeriesQuery' | 'LogQuery' | 'TraceQuery'>; // NEW
  panelOptionsEditorComponents?: Array<PanelOptionsEditorComponent<Spec>>;
  hideQueryEditor?: boolean; // can be removed 
}
```

In the example above, a future Table panel plugin would support all query interfaces, while TimeSeriesChart would not. hideQueryEditor can be removed and inferred when supportedQueryInterfaces is empty Array

Also, relates to #1151 for adding instant query support since a new query plugin is the ideal way to implement that functionality.

eunicorn:
After #1318 is merged, we can introduce “supportedQueryInterface” in panel plugins. The idea here is to give the framework the ability to render the appropriate query editor depending on which panel plugin is selected, enhancing dashboard user experience. For example, if user selects a time series chart, our panel editor will know to render a time series query editor. 

Additionally, the panel plugin can determine the active query interface to dynamically adjust its options. To do this, we can have `PanelEditor` becomes the children of `DataQueriesProvider`. This way, the options editor components can access `useDataQueries` or other relevant hooks to identify the data type returned by the query.

celian-garcia:
Hello! if you don´t mind I'm gonna take this one this week ;) I will ask you directly if I face any issue. Assigning to myself.

celian-garcia:
Regarding the PR upper 👆, I will close it and split it into three parts in order to be more readable and more feasible for me to implement 
- :one: the `PluginEditor` should support several query types`` (this is almost done in that PR) 
[...]

PR #1318, opened 2023-07-21
[BREAKINGCHANGE] Update DataQueriesProvider to support other data types

eunicorn:
Currently our DataQueriesProvider serves as the core component for data querying. It decouples the visualization layer from the data source layer. This separation allows users to update queries easily without affecting the visualization. 

However, our current implementation is limited to supporting time series queries only. We want to expand the capabilities of the DataQueriesProvider to support other data types, such as traces and logs. This improvement will  provide developers more versatility and flexibility when running queries.

**Changes**
- Modify the `DataQueriesProvider` to filter the definitions into different query types, returning a mix of data types
- Update `useDataQueries` hook to accept a `queryType` prop and return results specifically related to that query type.

**Usage**
Here’s an example of usage of the modified interface:
```
export function MyCustomPanel() {
  const definitions = [
    // TimeSeriesQueryPluginDefinition
    { kind: 'PrometheusTimeSeriesQuery', spec: prometheusTimeSeriesSpec },
    // LogsQueryPluginDefinition
    { kind: 'ElasticLogsQuery', spec: elasticLogsSpec },
  ];

  return (
    <DataQueriesProvider definitions={definitions}>
      <InstantQueryComponent />
      <LogsComponent />
      <TableComponent />
    </DataQueriesProvider>
  );
}

export function TableComponent() {
  // table can render any data type
  const { queryResults: { data: timeSeriesData } = useDataQueries('TimeSeresQuery');
  const { queryResults: { data: logs } = useDataQueries('LogsQuery');

  if (timeSeriesData) {
    return <TimeSeriesTable />
  }

  if (logs) {
    return <LogsTable />
  }
}
```

### DIFF
--- a/ui/core/src/model/query.ts
+++ b/ui/core/src/model/query.ts
@@ -20,6 +20,7 @@ interface QuerySpec<PluginSpec> {
 /**
  * A generic query definition interface that can be extended to support more than just TimeSeriesQuery
  */
+// Kind needs to be `any` because otherwise typescript will complain 'unknown' is not assignable to type '"TimeSeriesQuery"' in a few places
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export interface QueryDefinition<Kind = any, PluginSpec = UnknownSpec> {
   kind: Kind;
@@ -32,12 +33,8 @@ export interface QueryDefinition<Kind = any, PluginSpec = UnknownSpec> {
 export interface QueryType {
   TimeSeriesQuery: TimeSeriesData;
   // in the future we can add other query plugin and data types
-  // LogsQuery: LogsData;
+  // for example: we can add something like `LogsQuery: LogsData;`
 }
-
-// export interface LogsData {
-//   message: string;
-// }
 
 /**
  * Values of QueryType
--- a/ui/core/src/model/query.ts
+++ b/ui/core/src/model/query.ts
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 import { Definition, UnknownSpec } from './definitions';
+import { TimeSeriesData } from './time-series-data';
 
 interface QuerySpec<PluginSpec> {
   plugin: Definition<PluginSpec>;
@@ -19,9 +20,27 @@ interface QuerySpec<PluginSpec> {
 /**
  * A generic query definition interface that can be extended to support more than just TimeSeriesQuery
  */
-export interface QueryDefinition<Kind = QueryType, PluginSpec = UnknownSpec> {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface QueryDefinition<Kind = any, PluginSpec = UnknownSpec> {
   kind: Kind;
   spec: QuerySpec<PluginSpec>;
 }
 
-export type QueryType = 'TimeSeriesQuery';
+/**
+ * Mapping the query plugin to the data type it returns
+ */
+export interface QueryType {
+  TimeSeriesQuery: TimeSeriesData;
+  // in the future we can add other query plugin and data types
+  // LogsQuery: LogsData;
+}
+
+// export interface LogsData {
+//   message: string;
+// }
+
+/**
+ * Values of QueryType
+ * ex: 'TimeSeriesData'
+ */
+export type QueryDataType = QueryType[keyof QueryType];
--- a/ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
+++ b/ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
@@ -12,16 +12,17 @@
 // limitations under the License.
 
 import { createContext, useCallback, useContext, useMemo } from 'react';
-import { Definition, TimeSeriesQueryDefinition, UnknownSpec } from '@perses-dev/core';
+import { QueryType, TimeSeriesQueryDefinition } from '@perses-dev/core';
 import { useTimeSeriesQueries } from '../time-series-queries';
-import { DataQueriesProviderProps, QueryData, UseDataQueryResults } from './model';
+import {
+  DataQueriesProviderProps,
+  UseDataQueryResults,
+  transformQueryResults,
+  DataQueriesContextType,
+  QueryData,
+} from './model';
 
-export function useDataQueries(): UseDataQueryResults {
-  const ctx = useDataQueriesContext();
-  return ctx;
-}
-
-export const DataQueriesContext = createContext<UseDataQueryResults | undefined>(undefined);
+export const DataQueriesContext = createContext<DataQueriesContextType | undefined>(undefined);
 
 export function useDataQueriesContext() {
   const ctx = useContext(DataQueriesContext);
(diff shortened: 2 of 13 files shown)


ui/core/src/model/query.ts
```diff
@@ -12,16 +12,35 @@
...
+export interface QueryDefinition<Kind = any, PluginSpec = UnknownSpec> {
   kind: Kind;
   spec: QuerySpec<PluginSpec>;
 }
 
-export type QueryType = 'TimeSeriesQuery';
+/**
+ * Mapping the query plugin to the data type it returns
+ */
+export interface QueryType {
```
eunicorn, line 32: This allows typescript to be able to infer the return type of 'useDataQueries` given the parameter type.

ui/plugin-system/src/runtime/DataQueriesProvider/model.ts
```diff
@@ -11,32 +11,55 @@
...
 
 type QueryOptions = Record<string, unknown>;
 
-interface DataQueriesDefinitions<QueryPluginDefinition> {
-  definitions: QueryPluginDefinition[];
+export interface DataQueriesDefinition extends Definition<UnknownSpec> {
+  /**
+   * @default 'TimeSeriesQuery'
+   */
+  type?: keyof QueryType;
```
eunicorn, line 24: We need a new prop `type` here since it is now capable of supporting other query types

sjcobb:
🎉 can you update #1303 with some more info about how it relates to these changes?

ui/core/src/model/query.ts
```diff
@@ -12,16 +12,35 @@
...
 
 interface QuerySpec<PluginSpec> {
   plugin: Definition<PluginSpec>;
 }
 /**
  * A generic query definition interface that can be extended to support more than just TimeSeriesQuery
  */
-export interface QueryDefinition<Kind = QueryType, PluginSpec = UnknownSpec> {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface QueryDefinition<Kind = any, PluginSpec = UnknownSpec> {
```
juliepagano, line 24: Would an `unknown` kind work better here? If it's important to use `any` could you put a comment above the `eslint-disable` explaining why for future people working in the code.

ui/core/src/model/query.ts
```diff
@@ -12,16 +12,35 @@
...
+/**
+ * Mapping the query plugin to the data type it returns
+ */
+export interface QueryType {
+  TimeSeriesQuery: TimeSeriesData;
+  // in the future we can add other query plugin and data types
+  // LogsQuery: LogsData;
+}
+
+// export interface LogsData {
```
juliepagano, line 38: Could you clarify in these comments that these are here as examples for future dev (as opposed to real implementations that were commented out for some other reason). I'm hoping that'll help reduce confusion in the future.

ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
```diff
@@ -31,47 +32,66 @@ export function useDataQueriesContext() {
...
   const { definitions, options, children } = props;
 
-  // For now we will map each query plugin definition to TimeSeriesQueryDefinition
-  // Later on when we add support for other query types,
-  // we will have to map each query maps to the correct QueryDefinition
-  const timeSeriesQueries = definitions.map(
-    (definition) =>
-      ({
+  const queryDefinitions = definitions.map((definition) => {
+    if (definition.type === undefined || definition.type === 'TimeSeriesQuery') {
```
juliepagano, line 62: If these are explicitly `TimeSeriesQueries`, would it makes sense to adjust the variable name or is the idea that this conditional will be adjusted to include other types when they are added?

ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
```diff
@@ -12,16 +12,17 @@
 // limitations under the License.
```
juliepagano, line 13: Is it worth adding a couple simple unit tests for some of the logic changes in this file?

ui/core/src/model/query.ts
eunicorn: I had to use `any` here because otherwise I was getting typescript errors complaining that unknown cannot be converted to TimeSeriesQueryDefinition

ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
eunicorn: > is the idea that this conditional will be adjusted to include other types when they are added? 

yep this!

ui/plugin-system/src/runtime/DataQueriesProvider/model.ts
saminzadeh: If we made the data query type unique across all types, it would be cool to infer this.
eg. You could look up that `PrometheusTimeSeriesQuery` was a `TimeSeriesQuery` without explicitly saying so.

ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
eunicorn: yep done!

juliepagano:
Is the added screenshot because of some diffs with what's on `main` or is that an example of something going funny with storybook/happo? I'm trying to keep a close eye out for weirdness there.

juliepagano (approved):

ui/plugin-system/src/runtime/DataQueriesProvider/DataQueriesProvider.tsx
eunicorn: > could we look up the type via the kind of query?

@saminzadeh yep added getQueryType and removed `type` from the DataQueriesProvider definitions prop

eunicorn:
> Is the added screenshot because of some diffs with what's on `main` or is that an example of something going funny with storybook/happo? I'm trying to keep a close eye out for weirdness there.

just rebased the branch with the latest main branch so we'll see if the happo test is still failing in the next run

ui/plugin-system/src/runtime/DataQueriesProvider/model.ts
sjcobb: Would that be a problem for our BE validation though? One of the goals is complete static validation of the entire schema in CI for GitOps use cases

saminzadeh: > Would that be a problem for our BE validation though? One of the goals is complete static validation of the entire schema in CI for GitOps use cases

I don't believe so. Can you give an example of where the issue would be?

sjcobb: > You could look up that PrometheusTimeSeriesQuery was a TimeSeriesQuery without explicitly saying so.

I misunderstood, I thought you were referring to how we've talked about someday trying to remove the [nesting](https://github.com/perses/perses/blob/v0.38.0/dev/data/dashboard.json#L163-L174) of having PrometheusTimeSeriesQuery inside of TimeSeriesQuery in our dashboard schema

For that topic, I wasn't sure if the BE relies on the nesting, but I'd have to dig into the implementation more to understand: [internal/api/shared/schemas/schemas.go#L76-L98](https://github.com/perses/perses/blob/a999795fdd1c2ce9ee76db7e429465012d7d6623/internal/api/shared/schemas/schemas.go#L76-L98)

ui/plugin-system/src/runtime/DataQueriesProvider/model.ts
```diff
@@ -11,32 +11,85 @@
...
+import { useListPluginMetadata } from '../plugin-registry';
 
 type QueryOptions = Record<string, unknown>;
-
-interface DataQueriesDefinitions<QueryPluginDefinition> {
-  definitions: QueryPluginDefinition[];
-}
-export interface DataQueriesProviderProps<QueryPluginDefinition = Definition<UnknownSpec>>
-  extends DataQueriesDefinitions<QueryPluginDefinition> {
+export interface DataQueriesProviderProps {
```
saminzadeh, line 20: `interface DataQueriesProviderProps<T = UnknownSpec>` use a generic here. That way consumers can get type safety as well.

```tsx
<DataQueriesProvider<PrometheusTimeSeriesQuery | ElasticLogsQuery> definitions={defs}>
  {children}
</DataQueriesProvider>
```

ui/plugin-system/src/runtime/DataQueriesProvider/model.ts
saminzadeh: I think it does right now, but I don't think there is a strong reason why it needs to. 

For this change however, we are focusing on the framework code rather than the dashboard spec translation which can be slightly different.

saminzadeh (approved):
This is going to be great! Thanks!

PR #1319, opened 2023-07-21
[FEATURE] Add global datasources CRUD

AntoineThebaud:
Issue ref: https://github.com/perses/perses/issues/1033

This PR adds global datasources CRUD, located in the Admin section.

2023-07-21_22h34_06

⚠️There are lots of code duplicated from the project datasources #1282. I spent quite some time trying to factorize as much as possible but in the end I got stucked, especially by the fact union types combined with useState/useImmer seem like a bottleneck here (e.g it doesnt seem possible to infer the concrete type of the object being spread/destructured in order to reuse it for the updated state?). Maybe using an intermediary type to gather Datasource and GlobalDatasource (with e.g project info as optional) would be the way to go? Anyway I'm pretty new to React so I may have missed something here. To whoever will review this, please let me know if you know a way to refactor this and I'll be happy to do it. ⚠️

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/DatasourceList/DatasourceDataGrid.tsx
+++ b/ui/app/src/components/DatasourceList/DatasourceDataGrid.tsx
@@ -20,6 +20,7 @@ import {
   GridToolbarFilterButton,
   GridToolbarQuickFilter,
   GridRow,
+  GridRowParams,
   GridColumnHeaders,
 } from '@mui/x-data-grid';
 import { memo, useMemo } from 'react';
@@ -42,7 +43,7 @@ const MemoizedRow = memo(GridRow);
 const MemoizedColumnHeaders = memo(GridColumnHeaders);
 
 export interface Row {
-  project: string;
+  project?: string;
   name: string;
   displayName: string;
   version: number;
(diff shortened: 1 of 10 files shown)


sjcobb:
This would be a major change but starting to wonder if this could benefit from being a slice in Zustand (how we handle dashboard / panel / variable state), there isn’t an app-wide store setup yet though and need to think some more about how the overriding would work at the dashboard-level but this would let us use Redux devtools which can make debugging so much easier

celian-garcia:
FYI @AntoineThebaud, since you're off, I am currently stealing this PR to make a DRY version for the CRUD, thanks to the same mechanism that I did for Global/Project variables. I hope you don't mind :heart: 
I will keep your commit and add a "refactor" one :wink: 

@sjcobb Regarding the zustand store, I am not fluent enough unfortunately to integrate it in the Global CRUD, I think we have a lot to do with eventual bugfixing and the builtin variables before the launch (started by @Gladorme :heart:). Then this will come in a second place, except if somebody have enough bandwith to propose a refactoring ofc!

PR #1325, opened 2023-07-24
[ENHANCEMENT] Bar chart updates

ellenwyllie:
- Updates formatting of x axis values
- Adds a "Reset to Defaults" button to the bar chart settings
- Fixes empty state so that "No data" message is shown when appropriate
- Prevents y axis labels from getting cut off

X axis before:
Screen Shot 2023-07-24 at 1 25 20 PM
X axis after:
Screen Shot 2023-07-24 at 1 24 45 PM

Reset to Defaults button:
Screen Shot 2023-07-24 at 1 26 36 PM

Empty state:
Screen Shot 2023-07-24 at 1 27 56 PM

# Checklist

## UI Changes

### DIFF
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -21,7 +21,7 @@ import { Box } from '@mui/material';
 import { useChartsTheme } from '../context/ChartsThemeProvider';
 import { EChart } from '../EChart';
 import { ModeOption } from '../ModeSelector';
-import { getYAxes } from '../utils';
+import { formatAxesValues } from '../utils';
 
 use([EChartsBarChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
 
@@ -63,7 +63,7 @@ export function BarChart(props: BarChartProps) {
           source: source,
         },
       ],
-      xAxis: getYAxes({}, unit),
+      xAxis: formatAxesValues({}, unit),
       yAxis: {
         type: 'category',
         splitLine: {
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -21,7 +21,7 @@ import { Box } from '@mui/material';
 import { useChartsTheme } from '../context/ChartsThemeProvider';
 import { EChart } from '../EChart';
 import { ModeOption } from '../ModeSelector';
-import { formatAxesValues } from '../utils';
+import { getFormattedAxis } from '../utils';
 
 use([EChartsBarChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
 
@@ -63,7 +63,7 @@ export function BarChart(props: BarChartProps) {
           source: source,
         },
       ],
-      xAxis: formatAxesValues({}, unit),
+      xAxis: getFormattedAxis({}, unit),
       yAxis: {
         type: 'category',
         splitLine: {
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -105,6 +105,7 @@ export function BarChart(props: BarChartProps) {
         formatter: (params: { name: string; data: number[] }) =>
           params.data[1] && `<b>${params.name}</b> &emsp; ${formatValue(params.data[1], unit)}`,
       },
+      // increase distance between grid and container to prevent y axis labels from getting cut off
       grid: {
         left: '5%',
         right: '5%',
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -105,6 +105,10 @@ export function BarChart(props: BarChartProps) {
         formatter: (params: { name: string; data: number[] }) =>
           params.data[1] && `<b>${params.name}</b> &emsp; ${formatValue(params.data[1], unit)}`,
       },
+      grid: {
+        left: '5%',
+        right: '5%',
+      },
     };
   }, [data, chartsTheme, width, mode, unit]);
 
--- a/ui/components/src/BarChart/BarChart.tsx
+++ b/ui/components/src/BarChart/BarChart.tsx
@@ -21,6 +21,7 @@ import { Box } from '@mui/material';
 import { useChartsTheme } from '../context/ChartsThemeProvider';
 import { EChart } from '../EChart';
 import { ModeOption } from '../ModeSelector';
+import { getYAxes } from '../utils';
 
 use([EChartsBarChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
 
@@ -45,7 +46,7 @@ export function BarChart(props: BarChartProps) {
   const chartsTheme = useChartsTheme();
 
   const option: EChartsCoreOption = useMemo(() => {
-    if (data == null) return chartsTheme.noDataOption;
+    if (data == null || !data.length) return chartsTheme.noDataOption;
 
     const source: Array<Array<BarChartData['label'] | BarChartData['value']>> = [];
     data.map((d) => {
@@ -62,7 +63,7 @@ export function BarChart(props: BarChartProps) {
           source: source,
         },
       ],
-      xAxis: {},
+      xAxis: getYAxes({}, unit),
       yAxis: {
         type: 'category',
         splitLine: {
--- a/ui/components/src/utils/axis.ts
+++ b/ui/components/src/utils/axis.ts
@@ -12,15 +12,15 @@
 // limitations under the License.
 
 import merge from 'lodash/merge';
-import type { YAXisComponentOption } from 'echarts';
+import type { XAXisComponentOption, YAXisComponentOption } from 'echarts';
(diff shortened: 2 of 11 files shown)


ui/components/src/BarChart/BarChart.tsx
```diff
@@ -62,7 +63,7 @@ export function BarChart(props: BarChartProps) {
           source: source,
         },
       ],
-      xAxis: {},
+      xAxis: getYAxes({}, unit),
```
ellenwyllie, line 66: @sjcobb Any thoughts on updating `getYAxes` to work for either x or y axis? Could be something like this: 

```
/*
 * Populate yAxis or xAxis properties, returns an Array since multiple axes will be supported in the future
 */
export function formatAxesValues(axis?: YAXisComponentOption | XAXisComponentOption, unit?: UnitOptions) {
  // TODO: support alternate yAxis that shows on right side
  const AXIS_DEFAULT = {
    type: 'value',
    boundaryGap: [0, '10%'],
    axisLabel: {
      formatter: (value: number) => {
        return formatValue(value, unit);
      },
    },
  };
  return [merge(AXIS_DEFAULT, axis)];
}
```

sjcobb: Nice yeah I like that!

ellenwyllie: Sweet, updated it!

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
```diff
@@ -65,6 +66,17 @@ export function BarChartOptionsEditorSettings(props: BarChartOptionsEditorProps)
     );
   };
 
+  const handleResetSettings: React.MouseEventHandler<HTMLButtonElement> = () => {
+    onChange(
+      produce(value, (draft: BarChartOptions) => {
+        draft.calculation = 'LastNumber';
```
cndonovan, line 72: Looks like the DEFAULT_CALCULATION variable is set to 'Sum'

Which do you think is better default / reset value?

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
```diff
@@ -65,6 +66,17 @@ export function BarChartOptionsEditorSettings(props: BarChartOptionsEditorProps)
...
+        draft.unit = DEFAULT_UNIT;
+        draft.sort = 'desc';
+        draft.mode = 'value';
```
cndonovan, line 75: Definitely up to you, but we could potentially make these all constants and use them as the fallback value in the inputs below too (i.e. use them both in this reset function and as the defaults for each input) 

It's more verbose, so up to you if you think it's more helpful

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.test.tsx
```diff
@@ -133,4 +133,31 @@ describe('BarChartOptionsEditorSettings', () => {
     const unitSelector = screen.getByRole('combobox', { name: 'Unit' });
     expect(unitSelector).toBeDisabled;
   });
+
+  it('should reset settings to defaults', () => {
```
cndonovan, line 137: ⭐

ui/components/src/utils/axis.ts
```diff
@@ -12,15 +12,15 @@
...
-import type { YAXisComponentOption } from 'echarts';
+import type { XAXisComponentOption, YAXisComponentOption } from 'echarts';
 import { formatValue, UnitOptions } from '@perses-dev/core';
 
 /*
- * Populate yAxis properties, returns an Array since multiple y axes will be supported in the future
+ * Populate yAxis or xAxis properties, returns an Array since multiple axes will be supported in the future
  */
-export function getYAxes(yAxis?: YAXisComponentOption, unit?: UnitOptions) {
+export function formatAxesValues(axis?: YAXisComponentOption | XAXisComponentOption, unit?: UnitOptions) {
```
cndonovan, line 21: Could make sense to call this `getAxisWithFormatting` / `getFormattedAxis` or something since I think it still returns the entire axis object

cndonovan (approved):

ui/components/src/utils/axis.ts
cndonovan: (Especially since many of the call sites are naming the output of this function `axis`, so it could be like `yAxis: getFormattedAxis(...)`)

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
ellenwyllie: Hmm looks like `DEFAULT_CALCULATION` won't actually ever be used. Stat chart and gauge chart use `LastNumber` as the default so I'm inclined to stick with that.

@eunicorn @sjcobb @juliepagano Any thoughts on what the default calculation should be?

ui/components/src/utils/axis.ts
ellenwyllie: I like that, updated it to `getFormattedAxis`!

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
ellenwyllie: Good call, the defaults were being defined in a few places so I think it's good to set them as constants. Updated! [6d798df](https://github.com/perses/perses/pull/1325/commits/6d798dfe9e04be2956d98d45cbc7dba464cb14e8)

ui/panels-plugin/src/plugins/bar-chart/bar-chart-model.ts
```diff
@@ -16,6 +16,8 @@ import { CalculationType, Definition, UnitOptions } from '@perses-dev/core';
 import { OptionsEditorProps } from '@perses-dev/plugin-system';
 
 export const DEFAULT_UNIT: UnitOptions = { kind: 'Decimal' };
```
eunicorn, line 18: maybe we should also set `abbreviate` to true similar to TimeSeriesChart

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
eunicorn: looks like Grafana also uses `LastNumber` as default so `LastNumber` sounds good to me

ui/components/src/BarChart/BarChart.tsx
```diff
@@ -104,6 +105,10 @@ export function BarChart(props: BarChartProps) {
         formatter: (params: { name: string; data: number[] }) =>
           params.data[1] && `<b>${params.name}</b> &emsp; ${formatValue(params.data[1], unit)}`,
       },
+      grid: {
```
eunicorn, line 108: can you add a comment here why this is needed?

eunicorn (approved):

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
cndonovan: +1 to that plan

ui/panels-plugin/src/plugins/bar-chart/bar-chart-model.ts
cndonovan: Oh that's a good call, `abbreviate` is probably what users will want more of time (it means we'll use shortened, more readable values like `10K` instead of `10,000`)

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
juliepagano: +1

ui/components/src/BarChart/BarChart.tsx
ellenwyllie: Sure! [caae104](https://github.com/perses/perses/pull/1325/commits/caae1049f423d9b6bde6d35ac829883109239b7d)

ui/panels-plugin/src/plugins/bar-chart/bar-chart-model.ts
ellenwyllie: Good idea, added `abbreviate` [caae104](https://github.com/perses/perses/pull/1325/commits/caae1049f423d9b6bde6d35ac829883109239b7d)

ui/panels-plugin/src/plugins/bar-chart/BarChartOptionsEditorSettings.tsx
ellenwyllie: Thanks y'all, went with `LastNumber`! 

I also removed `export const DEFAULT_CALCULATION: CalculationType = 'Sum';` from ui/core/src/model/calculations.ts. It was referenced in the gauge and stat chart options editors but wasn't actually being used. Let me know if I'm missing something here and we need to keep it 496d16f563de0b3197a8839cffff9c2095801f32

PR #1327, opened 2023-07-25
[ENHANCEMENT] Choose largest time unit that produces natural numbers

cndonovan:
The goal of this work is to produce y-axis values that are easier to understand. In the code, we pick the largest time unit that still produces natural numbers (numbers > 1). 

## Goal 
| We currently say 3000ms | We want to say 3s |
| -- | -- |
| Screenshot 2023-07-25 at 11 51 12 AM | Screenshot 2023-07-25 at 11 51 22 AM |

## Perses Examples
| Before | After |
| -- | -- |
| Screenshot 2023-07-25 at 11 58 18 AM | Screenshot 2023-07-25 at 12 02 09 PM |
| Screenshot 2023-07-25 at 11 56 55 AM | Screenshot 2023-07-25 at 11 56 02 AM |

### DIFF
--- a/ui/core/src/model/units/time.ts
+++ b/ui/core/src/model/units/time.ts
@@ -74,10 +74,12 @@ export enum PersesTimeToIntlTime {
   Years = 'year',
 }
 
-// Note: This conversion will not be exactly accurate for months and years,
-// due variations in the lengths of months (i.e. 28 - 31 days) and years (i.e. leap years).
-// For precision with months and years, we would need more complex algorithms and/or external libraries.
-// However, we expect that measurements in months and years will be rare.
+/**
+ * Note: This conversion will not be exactly accurate for months and years,
+ * due variations in the lengths of months (i.e. 28 - 31 days) and years (i.e. leap years).
+ * For precision with months and years, we would need more complex algorithms and/or external libraries.
+ * However, we expect that measurements in months and years will be rare.
+ */
 const TIME_UNITS_IN_SECONDS: Record<TimeUnitKind, number> = {
   Years: 31536000, // 365 days
   Months: 2592000, // 30 days
@@ -100,7 +102,9 @@ const LARGEST_TO_SMALLEST_TIME_UNITS: TimeUnitKind[] = [
   'Milliseconds',
 ];
 
-// Choose the first time unit that produces a number greater than 1, starting from the biggest time unit.
+/**
+ * Choose the first time unit that produces a number greater than 1, starting from the biggest time unit.
+ */
 function getValueAndKindForNaturalNumbers(value: number, kind: TimeUnitKind): { value: number; kind: TimeUnitKind } {
   const valueInSeconds = value * TIME_UNITS_IN_SECONDS[kind];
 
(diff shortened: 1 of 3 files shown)


ui/core/src/model/units/units.ts
```diff
@@ -19,7 +19,7 @@ import { UnitGroup, UnitGroupConfig, UnitConfig } from './types';
 
 /**
  * Most of the number formatting is based on Intl.NumberFormat, which is built into JavaScript.
- * Prefer Intl.NumbeFormat because it covers most use cases and will continue to be well-supported with time.
+ * Prefer Intl.NumbeFormat because it covers most use cases and will continue to be supported with time.
```
cndonovan, line 22: I can't vouch for how supported it will be :)

eunicorn (approved):
🚀

ui/core/src/model/units/time.ts
```diff
@@ -74,13 +74,67 @@ export enum PersesTimeToIntlTime {
...
+  'Months',
+  'Weeks',
+  'Days',
+  'Hours',
+  'Minutes',
+  'Seconds',
+  'Milliseconds',
+];
+
+// Choose the first time unit that produces a number greater than 1, starting from the biggest time unit.
```
juliepagano, line 103: Optional suggestion: if you use the other style of comments, this description will show up in some text editors (e.g. vscode) when someone hovers over usage of `getValueAndKindForNaturalNumbers`, which I could see being helpful.

```ts
/**
 * Choose the first time unit that produces a number greater than 1, starting from the biggest time unit.
 */
```

ui/core/src/model/units/time.ts
```diff
@@ -74,13 +74,67 @@ export enum PersesTimeToIntlTime {
...
+  return kind === 'Months' || kind === 'Years';
+}
+
 export function formatTime(value: number, { kind, decimal_places }: TimeUnitOptions): string {
-  const isMonthOrYear = kind === 'Months' || kind === 'Years';
+  if (value === 0) return '0s';
+
+  const valueAndKindForNaturalNumbers = getValueAndKindForNaturalNumbers(value, kind);
+  value = valueAndKindForNaturalNumbers.value;
+  kind = valueAndKindForNaturalNumbers.kind;
```
juliepagano, line 132: nit: do you want to re-assign the variables from the args here? I could see it being slightly safer to use different variable names for clarity, but no strong feelings.

juliepagano (approved):
A few very optional suggestions. Looks good to me. 🚀

ui/core/src/model/units/time.ts
cndonovan: Oo yeah I'll play around with that, this was a part where I just keep trying to find a nicer way to do it / I wished I could reassign the variables with one line of code.

ui/core/src/model/units/time.ts
```diff
@@ -74,13 +74,73 @@ export enum PersesTimeToIntlTime {
...
+}
+
 export function formatTime(value: number, { kind, decimal_places }: TimeUnitOptions): string {
-  const isMonthOrYear = kind === 'Months' || kind === 'Years';
+  const { value: valueForNaturalNumbers, kind: kindForNaturalNumbers } = getValueAndKindForNaturalNumbers(value, kind);
+
+  return _formatTime(valueForNaturalNumbers, { kind: kindForNaturalNumbers, decimal_places });
+}
+
+function _formatTime(value: number, { kind, decimal_places }: TimeUnitOptions): string {
```
cndonovan, line 137: Do you find this code ^ more readable @juliepagano 

I'm playing around, so open to yays or nays

cndonovan: (I basically created a private internal function so we don't have to reassign `value` and `kind` inside of that function, `value` and `kind` are already determined by the time we get to this function)

juliepagano: Seems ok to me, although I could see being a little confused about the extraction of some of the code into `_formatTime` in the future when I've forgotten this discussion. I was trying to think of a name for `_formatTime` that would communicate the remaining work it was doing, but couldn't think of anything helpful. If you can think of anything more descriptive, that might be helpful, but otherwise I think it's fine as is.

PR #1328, opened 2023-07-26
[FEATURE] Global Variables CRUD

celian-garcia:
After the CRUD of variables by project, let's give the ability to make variables in a more global scope!

# Checklist

## UI Changes


ui/app/src/components/VariableList/VariablesList.tsx
```diff
@@ -23,44 +23,45 @@ import { useSnackbar } from '@perses-dev/components';
...
 import { DeleteVariableDialog } from '../dialogs';
-import { useUpdateVariableMutation } from '../../model/project-client';
 import { VariableDataGrid, Row } from './VariableDataGrid';
 import { VariableFormDrawer } from './VariableFormDrawer';
 
-export interface VariableListProperties {
-  projectName: string;
-  variableList: VariableResource[];
+export interface VariableListProperties<T extends Variable> {
+  project?: string;
```
celian-garcia, line 30: ``project`` is unused

ui/app/src/model/global-variable-client.ts
```diff
@@ -0,0 +1,143 @@
...
+  const name = entity.metadata.name;
+  const url = buildURL({ resource, name });
+  return fetch(url, {
+    method: HTTPMethodDELETE,
+    headers: HTTPHeader,
+  });
+}
+
+/**
+ * Used to get a variable from the API.
```
celian-garcia, line 67: global

ui/app/src/model/global-variable-client.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Will automatically be refreshed when cache is invalidated
+ */
+export function useGlobalVariable(name: string) {
```
celian-garcia, line 70: global

ui/core/src/utils/text.ts
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Datasource, DashboardResource, VariableResource } from '../model';
+import { Datasource, DashboardResource, GlobalVariableResource, Variable } from '../model';
```
celian-garcia, line 14: GlobalVariableResource unused ?

celian-garcia:
Compare behaviors when an update/delete action on the API does not succeed

celian-garcia:
Reuse buildQueryKey in other clients + unit test it

sjcobb:
I haven't dug too far into a technical review yet, I do like how you went with generics though! Noticed a couple small functional issues (could be existing bugs too), after a variable is created, when switching between variables, the list variable input can become disabled and I can't add new values unless I refresh:

https://github.com/perses/perses/assets/9369625/053e380b-3564-462d-b734-8f4c3c8dab24

Another is after creating a new project variable, it doesn't show immediately in the table, success banner shows but you have to refresh to see it (it works correctly when creating a new global variable)

https://github.com/perses/perses/assets/9369625/3dbbb6a9-8aa2-4d8e-b233-b5e46c951251

And a small nit when editing variables on a dashboard, the spacing is inconsistent, 'Global Variables' heading is too close to the border above. Design-wise it could be useful to have an InfoTooltip beside 'Global Variables' heading with info about how they work to help users and there could be a link to `/admin/variables` inside it that opens in a new browser tab

Global_Var_heading_spacing

ui/app/src/components/VariableList/VariablesList.tsx
```diff
@@ -23,44 +23,47 @@ import { useSnackbar } from '@perses-dev/components';
...
   initialState?: GridInitialStateCommunity;
   isLoading?: boolean;
 }
 
 /**
  * Display variables in a table style.
- * @param props.variableList Contains all variables to display
+ * @param props.data Contains all variables to display
  * @param props.hideToolbar Hide toolbar if enabled
+ * @param props.onUpdate Event received when an update action has been requested
```
sjcobb, line 44: I didn’t try it but curious, can a GlobalVariable be used inside a ProjectVariable?

celian-garcia: I think yes, but nothing is checked when you create a project variable. The same question for the dependency between project variables. 

That's an interesting point that we need to clarify for us, and in the view as well.

celian-garcia:
@sjcobb 
> after a variable is created, when switching between variables, the list variable input can become disabled and I can't add new values unless I refresh:
For this issue I actually this is because you've been tricked by the fact that clicking on the row is different from clicking on the edit button of the row. 
- on the row -> Open the drawer in view mode
- on the edit button => Open the drawer in edit mode.
This is something that we probably need to improve, maybe having a way to switch between view and edit mode in the drawer itself? The view mode is important to keep IMHO when perses will for the Perses readonly mode or for when we'll implement some users and roles mechanisms.

> Another is after creating a new project variable, it doesn't show immediately in the table, success banner shows but you have to refresh to see it (it works correctly when creating a new global variable)

Definitely a bug, I will fix it in that PR.

> And a small nit when editing variables on a dashboard, the spacing is inconsistent, 'Global Variables' heading is too close to the border above. Design-wise it could be useful to have an InfoTooltip beside 'Global Variables' heading with info about how they work to help users and there could be a link to /admin/variables inside it that opens in a new browser tab

Ah yes indeed I didn't notice that space lack. I like the idea of having more information + a way to go to the definition. This is definitely a must indeed. I will integrate something in that PR.

ui/app/src/model/datasource-api.ts
```diff
@@ -14,8 +14,10 @@
 import { Datasource, DatasourceSelector, GlobalDatasource } from '@perses-dev/core';
 import { DatasourceApi } from '@perses-dev/dashboards';
 import LRUCache from 'lru-cache';
-import { fetchDatasourceList, fetchGlobalDatasourceList } from './datasource-client';
+import { fetchDatasourceList } from './datasource-client';
+import { fetchGlobalDatasourceList } from './global-datasource-client';
 
+//TODO: Move to tanstack query for caching
```
sjcobb, line 20: Definitely would love to see this using react-query, I briefly looked into it at one point and it wasn’t as easy as I hoped (my first attempt violated rules of hooks but there has to be a way)

ui/app/src/model/querykey-builder.ts
```diff
@@ -0,0 +1,53 @@
...
+ *   - [variables, myProject, myVar]
+ *   - [variables, myProject]
+ *   - [variables]
+ *
+ * NB: We can do this because, in the backend, all resources that can be in a parent always have a parent
+ *   => you cannot update a variable resource with PUT /api/v1/variables/myVar,
+ *      so we cannot build a query key like this [variables, myVar]
+ * @param params
+ */
+export function buildQueryKey(params: QueryKeyParams): QueryKey {
```
sjcobb, line 42: Awesome, I see you’re a react-query expert now! 😄

ui/app/src/utils/variables.ts
```diff
@@ -12,17 +12,38 @@
 // limitations under the License.
 
 import { ExternalVariableDefinition } from '@perses-dev/dashboards';
-import { VariableResource } from '@perses-dev/core';
+import { Variable } from '@perses-dev/core';
 import { ExternalVariableSource } from '../model/variables';
 
+export function buildProjectVariableDefinition(projectName: string, variables: Variable[]): ExternalVariableDefinition {
+  return {
+    editLink: `/projects/${projectName}/variables`,
```
sjcobb, line 20: This is great, love how it links directly to the correct project’s variable tab and describes how overriding works

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -70,13 +71,15 @@ function ViewDashboard() {
   const isReadonly = useIsReadonly();
 
   // Collect the Project variables and setup external variables from it
-  // TODO: Once we'll implement global variables CRUD, we'll do the same with 'global' as source
+  const { data: globalVars, isLoading: isLoadingGlobalVars } = useGlobalVariableList();
   const { data: projectVars, isLoading: isLoadingProjectVars } = useVariableList(projectName);
   const externalVariableDefinitions: ExternalVariableDefinition[] | undefined = useMemo(() => {
     const result: ExternalVariableDefinition[] = [];
-    if (projectVars && projectVars.length > 0) result.push(buildExternalVariableDefinition('project', projectVars));
+    if (projectVars && projectVars.length > 0) result.push(buildProjectVariableDefinition(projectName, projectVars));
```
sjcobb, line 78: For my own understanding, I’m not 100% following what all we’re doing with the response in these two functions and `buildExternalVariableDefinition`, are there any changes on the BE that could simplify this mapping in the future? I also noticed this comment which confused me on what is in the spec vs metadata:

*“The name of the variable resource can be missing from the spec but present in the metadata.”*

I’m also starting to wonder if the way we store separate kinds for GlobalVariable and ProjectVariable could be simplified somehow since we are mapping them to Variable anyways and using `source` but I guess we have to be explicit so that we know the hierarchy for which overrides the other?

sjcobb (approved):
I left a few questions but overall this LGTM!

@Gladorme @AntoineThebaud could be good to have another set of 👀 though before merging

ui/app/src/model/datasource-api.ts
celian-garcia: Same for me yeah I red the code and have been directly afraid of breaking something :fearful:  I think I will ask a bit of the time of @Nexucis at some point to do it together and be sure it follows what he wanted. I'm afraid he will have forgot as it's been a long time :laughing:

Gladorme (approved):
LGTM 😄

ui/app/src/views/ViewDashboard.tsx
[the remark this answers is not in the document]
celian-garcia: I can do that if I add a condition in the VariableEditor to not display the title "Global Variables" when the extVar.definitions is empty. 
Okay it's probably safer in any case. I'll do it thanks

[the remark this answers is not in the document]
celian-garcia: Updated to this 
```ts
() => [
      buildProjectVariableDefinition(projectName, projectVars || []),
      buildGlobalVariableDefinition(globalVars || []),
    ],
```

Note that the builder doesn't return a list of var definition directly but a list for a dedicated source, so the externalVariableDefinitions list here have only 2 elements with following format : 

```ts
export type ExternalVariableDefinition = {
  source: string; // e.g "project" or "global"
  tooltip?: {
    title?: string;
    description?: string;
  };
  editLink?: string;
  definitions: VariableDefinition[]; // ALL the defs of the project or global variable
};
```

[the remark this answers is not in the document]
celian-garcia: Plus the conditions added in the VariableEditor
https://github.com/perses/perses/pull/1328/files#diff-a9f19d289024de68e7f85a86767258199b0d0bc3122c8122894f4105edef1040R282-R287

ui/app/src/model/querykey-builder.ts
celian-garcia: I wouldn't be that sure hehe. I still have some oddities that I don´t understand. For example, I found out that when I create a global variable, to make the list refresh, this is not working to invalidate the ``[variables, myVariable]`` key, I had to invalidate ``[variables]``. 
Maybe it's because I'm declaring the mutation with wrong key. I'm not sure yet.

celian-garcia:
@sjcobb Looks like a silent failure of the tests occurred. 
MicrosoftTeams-image (2)

I will fix the regression but shouldn't it be red in case tests fail ??

Gladorme:
> @sjcobb Looks like a silent failure of the tests occurred. MicrosoftTeams-image (2)
> 
> I will fix the regression but shouldn't it be red in case tests fail ??

The first try failed but not the second one, it's normal

PR #1317, opened 2023-07-21
[BREAKINGCHANGE] Only pin one tooltip at a time by default, `ChartsThemeProvider` renamed to `ChartsProvider`

sjcobb:
## Summary

- [BREAKINGCHANGE] `ChartsThemeProvider` renamed to `ChartsProvider`, app-level enablePinning prop added
- [ENHANCEMENT] Only pin one tooltip at a time by default, `Cmd + Click` to pin multiple
- [ENHANCEMENT] Snap crosshair to nearest datapoint when clicking to pin 
- #1330
- #1329

We got feedback that unpinning multiple tooltip individually is tedious, this PR will address this by only allow 1 tooltip to be pinned at a time by default. If the user clicks to pin a second tooltip, the first will be unpinned. To pin more than one, the Ctrl key can be held down to opt in to the multi pinned tooltip feature.

This PR will be a major BREAKINGCHANGE since all embed users will need to rename ChartsThemeProvider -> ChartsProvider (more flexible name and we can expand the scope to control other app-wide functionality like shared crosshair)

### Snap crosshair to nearest datapoint

https://github.com/perses/perses/assets/9369625/0da53a31-6ea2-465d-ac5b-c9fc7412947e

### Unpin subsequent tooltips:

https://github.com/perses/perses/assets/9369625/cb79eaeb-4ece-4323-947e-813f578b3108

https://github.com/perses/perses/assets/9369625/dedefc54-57e8-47fe-90c4-4b981b1bdf2f

### ~~Unpin change WITH pinned crosshair on synced charts:~~

~~https://github.com/perses/perses/assets/9369625/552b214b-ff2a-4827-8ea2-d69fe600fb42~~

## TODO
- [x] Ability to disable tooltip pinning app-wide 
- [x] Rename ChartsThemeProvider -> ChartsProvider
- [x] Pin subsequent tooltips and unpin prevous
- [x] Opt-in to multi tooltip pinning when CTRL key held down
- [ ] Pin crosshair on synced charts aka temporary annotation
  - Removed synched chart pinned crosshair to simplify this PR, that can be added separately later
- [ ] ~~Double check synced chart crosshairs, seems adding is inconsistent~~ Fixed by removing this feature for now, only one crosshair is pinned at a time again

PR #1258, opened 2023-06-20
[BUGFIX] Ensure pinned tooltip shows behind modals / sticky header

sjcobb:
Pinned tooltip fixed to not show above edit panel drawer and sticky header.

## Before
edit_panel_overlap_bug

sticky_header_overlap_bug

## After

image

PR #1329, opened 2023-07-26
[BUGFIX] Disable tooltip pinning in panel preview

sjcobb:
## Overview
In order to fix zIndex issues when clicking to pin on a dashboard, changes were made in #1258 that break tooltip pinning once inside a Drawer. Since pinning isn't as useful in PanelPreview.tsx, this PR disables tooltip pinning by adding a nested `<ChartsProvider chartsTheme={chartsTheme} enablePinning={false}>`

This is follow up to #1317, that branch will be merged first

## Screenshot
image

### DIFF
--- a/ui/components/src/TimeChart/TimeChart.tsx
+++ b/ui/components/src/TimeChart/TimeChart.tsx
@@ -343,7 +343,7 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
             if (current === null) {
               const cursorX = pointInGrid[0];
               const closestTimestamp = getClosestTimestampInFullDataset(data, cursorX);
-              const pinnedCrosshair = merge(DEFAULT_PINNED_CROSSHAIR, {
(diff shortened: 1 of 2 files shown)


juliepagano (approved):
I have some idle curiosity around the complexity and perf impact of having nested context as we grow the library, but I think that's something for longer-term thinking. This small case looks reasonable to me!

ui/components/src/TimeChart/TimeChart.tsx
```diff
@@ -343,7 +343,7 @@ export const TimeChart = forwardRef<ChartInstance, TimeChartProps>(function Time
             if (current === null) {
               const cursorX = pointInGrid[0];
               const closestTimestamp = getClosestTimestampInFullDataset(data, cursorX);
-              const pinnedCrosshair = merge(DEFAULT_PINNED_CROSSHAIR, {
```
sjcobb, line 346: This diff was fixed in other PR #1317

PR #1330, opened 2023-07-26
[ENHANCEMENT] Improve performance in `checkforNearbyTimeSeries`

sjcobb:
Rely on `getCommonTimeScale` and `getTimeSeriesValues` to ensure consistent xAxis intervals. We don't need to manually check every series.

### DIFF
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
@@ -245,7 +245,6 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
         seriesIndex++;
       }
     }
-    graphData.xAxis = xAxisData;
 
     if (thresholds && thresholds.steps) {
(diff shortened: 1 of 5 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -245,7 +245,6 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
         seriesIndex++;
       }
     }
-    graphData.xAxis = xAxisData;
```
eunicorn, line 248: can we just remove graphData completely?

eunicorn (approved):

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
sjcobb: Dang I thought I did remove most of this in #1286, but clearly missed some places 😂

Issue #1280, opened 2023-07-07
Bug: refresh interval does not update x axis


sjcobb:
Query refreshes but x axis min and max don’t reflect the newly fetched data

sjcobb:
Discussed [potential fixes](https://github.com/perses/perses/discussions/947?sort=new#discussioncomment-6406568) in weekly sync, should be able to increment refreshKey to bust cache, hopefully no changes in panels are needed, only to query plugin 🤞

PR #1344, opened 2023-08-03
[FEATURE] Add builtin variables ($__dashboard, $__project, $__from, $__to)

### LINKED ISSUE #1346: Whitelist builtin variables in backend validation
The backend validation on dashboard creation should accept all variables starting with $__.

### LINKED ISSUE #1347: Forbid creation of variable with builtin prefix `$__`

Gladorme:
# Description

Added support for basic builtin variables, wdyt about the implementation? I still need to figure out how to implement plugin builtin variables (i.e.: $__rate_interval for prometheus-plugin)

Builtin variables are dynamic variables computed when the dashboard is rendered / refreshed.
Builtin variables implemented:
- `$__dashboard`: the dashboard display name name, if not set dashboard metadata name
- `$__project`: the dashboard project name
- `$__from`: the start time of the dashboard
- `$__to`: the end time of the dashboard

Know issues: 
- $__from and $_to are not updated with auto-refresh, related to: #1280 
- $__from and $__to variables are not consistent with our current naming, in the code / url, it's `start` and `end`. Should we rename variables?

# Screenshots

## Variable Form Editor
Builtin variables

# Related issues
Closes #1347 
Closes #1346 

# Checklist

## UI Changes

### DIFF
--- a/internal/api/shared/validate/validate.go
+++ b/internal/api/shared/validate/validate.go
@@ -25,6 +25,12 @@ import (
 	"github.com/perses/perses/pkg/model/api/v1/utils"
 )
 
+// We want to keep only variables that are not only a number.
+// A number that represents a variable is not meaningful, and so we don't want to consider it.
+// It's also a way to avoid a collision in terms of variable template syntax.
+// For example in PromQL, in the function `label_replace`, it used the syntax $1, $2, for the placeholder.
+var variableTemplateNameRegexp = regexp.MustCompile(`^\w*?[^0-9]\w*$`)
+
 func Dashboard(entity *modelV1.Dashboard, sch schemas.Schemas) error {
 	if _, err := utils.BuildVariableOrder(entity.Spec.Variables, nil, nil); err != nil {
 		return err
@@ -85,15 +91,13 @@ func validateUnicityOfDefaultDTS[T modelV1.DatasourceInterface](entity T, list [
 }
 
 func validateVariableName(variable string) error {
-	var variableRegex = regexp.MustCompile(`^\w+$`)
-	valid := variableRegex.MatchString(variable)
+	valid := variableTemplateNameRegexp.MatchString(variable)
 	if !valid {
 		return fmt.Errorf("variable name '%s' is not valid", variable)
 	}
 
 	// Checking if variable do not have builting variable prefix: __
-	var builtinVarPrefixRegex = regexp.MustCompile(`^__`)
-	isBuiltinVar := builtinVarPrefixRegex.MatchString(variable)
+	isBuiltinVar := utils.BuiltinVariablePrefixRegexp.MatchString(variable)
 	if isBuiltinVar {
 		return fmt.Errorf("variable name '%s' can not have builtin variable prefix: __", variable)
 	}
--- a/internal/api/shared/validate/validate.go
+++ b/internal/api/shared/validate/validate.go
@@ -15,6 +15,8 @@ package validate
 
 import (
 	"fmt"
+	"github.com/perses/perses/pkg/model/api/v1/dashboard"
+	"regexp"
 
 	"github.com/perses/perses/internal/api/shared/schemas"
 	modelV1 "github.com/perses/perses/pkg/model/api/v1"
@@ -51,6 +53,16 @@ func Datasource[T modelV1.DatasourceInterface](entity T, list []T, sch schemas.S
 	return nil
 }
 
+func Variable[T modelV1.VariableInterface](entity T, sch schemas.Schemas) error {
+	if err := validateVariableName(entity.GetMetadata().GetName()); err != nil {
+		return err
+	}
+	if err := sch.ValidateGlobalVariable(entity.GetVarSpec()); err != nil {
+		return err
+	}
+	return nil
+}
+
 func validateUnicityOfDefaultDTS[T modelV1.DatasourceInterface](entity T, list []T) error {
 	name := entity.GetMetadata().GetName()
 	spec := entity.GetDTSSpec()
@@ -72,6 +84,31 @@ func validateUnicityOfDefaultDTS[T modelV1.DatasourceInterface](entity T, list [
 	return nil
 }
 
+func validateVariableName(variable string) error {
+	var variableRegex = regexp.MustCompile(`^\w+$`)
+	valid := variableRegex.MatchString(variable)
+	if !valid {
+		return fmt.Errorf("variable name '%s' is not valid", variable)
+	}
+
+	// Checking if variable do not have builting variable prefix: __
+	var builtinVarPrefixRegex = regexp.MustCompile(`^__`)
+	isBuiltinVar := builtinVarPrefixRegex.MatchString(variable)
+	if isBuiltinVar {
+		return fmt.Errorf("variable name '%s' can not have builtin variable prefix: __", variable)
+	}
+	return nil
+}
+
+func validateVariableNames(variables []dashboard.Variable) error {
+	for _, variable := range variables {
+		if err := validateVariableName(variable.Spec.GetName()); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func validateDTSPlugin(plugin common.Plugin, sch schemas.Schemas) error {
 	if _, err := http.ValidateAndExtract(plugin.Spec); err != nil {
 		return err
@@ -80,6 +117,10 @@ func validateDTSPlugin(plugin common.Plugin, sch schemas.Schemas) error {
 }
 
 func validateDashboard(entity *modelV1.Dashboard, sch schemas.Schemas) error {
+	if err := validateVariableNames(entity.Spec.Variables); err != nil {
+		return err
+	}
+
 	if sch != nil {
 		if err := sch.ValidateDashboardVariables(entity.Spec.Variables); err != nil {
 			return err
--- a/internal/api/shared/validate/validate.go
+++ b/internal/api/shared/validate/validate.go
@@ -59,7 +59,7 @@ func Datasource[T modelV1.DatasourceInterface](entity T, list []T, sch schemas.S
 	return nil
 }
 
-func Variable[T modelV1.VariableInterface](entity T, sch schemas.Schemas) error {
+func Variable(entity modelV1.VariableInterface, sch schemas.Schemas) error {
 	if err := validateVariableName(entity.GetMetadata().GetName()); err != nil {
 		return err
 	}
@@ -97,7 +97,7 @@ func validateVariableName(variable string) error {
 	}
 
 	// Checking if variable do not have builting variable prefix: __
-	isBuiltinVar := utils.BuiltinVariablePrefixRegexp.MatchString(variable)
+	isBuiltinVar := modelV1.IsBuiltinVariable(variable)
 	if isBuiltinVar {
 		return fmt.Errorf("variable name '%s' can not have builtin variable prefix: __", variable)
 	}
--- a/pkg/model/api/v1/utils/variable_build_order.go
+++ b/pkg/model/api/v1/utils/variable_build_order.go
@@ -25,11 +25,9 @@ import (
 	"strconv"
 )
 
-// We want to keep only variables that are not only a number.
-// A number that represents a variable is not meaningful, and so we don't want to consider it.
-// It's also a way to avoid a collision in terms of variable template syntax.
+// Not similar to variable validation regex (\w*?[^0-9]\w*), because some PromQL expression may need variable name with only number
 // For example in PromQL, in the function `label_replace`, it used the syntax $1, $2, for the placeholder.
-var variableTemplateSyntaxRegexp = regexp.MustCompile(`\$(\w*?[^0-9]\w*)`)
+var variableTemplateSyntaxRegexp = regexp.MustCompile(`\$(\w+)`)
 
 type VariableGroup struct {
 	Variables []string
--- a/pkg/model/api/v1/utils/variable_build_order.go
+++ b/pkg/model/api/v1/utils/variable_build_order.go
@@ -22,6 +22,7 @@ import (
 	"golang.org/x/exp/slices"
 	"reflect"
 	"regexp"
+	"strconv"
 )
 
 // We want to keep only variables that are not only a number.
@@ -29,9 +30,6 @@ import (
 // It's also a way to avoid a collision in terms of variable template syntax.
 // For example in PromQL, in the function `label_replace`, it used the syntax $1, $2, for the placeholder.
 var variableTemplateSyntaxRegexp = regexp.MustCompile(`\$(\w*?[^0-9]\w*)`)
-
-// BuiltinVariablePrefixRegexp is the regex for detecting builtin variables (i.e.: $__dashboard, $__from, $__to, ...)
-var BuiltinVariablePrefixRegexp = regexp.MustCompile(`^__`)
 
 type VariableGroup struct {
 	Variables []string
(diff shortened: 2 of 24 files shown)


ui/app/src/utils/variables.ts
```diff
@@ -36,6 +107,18 @@ export function buildGlobalVariableDefinition(variables: Variable[]): ExternalVa
     ...buildExternalVariableDefinition('global', variables),
   };
 }
+
+export function buildBuiltinVariableDefinition(variables: VariableDefinition[]): ExternalVariableDefinition {
```
celian-garcia, line 111: I'm not sure we want them to appear in the variables list everytime. When we edit variables.

ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
```diff
@@ -53,6 +56,12 @@ export function ViewDashboard(props: ViewDashboardProps) {
   const dashboardRefreshInterval = spec.refreshInterval ?? DEFAULT_REFRESH_INTERVAL;
   const initialTimeRange = useInitialTimeRange(dashboardDuration);
   const initialRefreshInterval = useInitialRefreshInterval(dashboardRefreshInterval);
+  const builtinVariableDefinitions = getBuiltinVariableDefinitions(dashboardResource);
```
celian-garcia, line 59: If you do that, it means that builtin variables from/to won´t be available in a panel embedded use case. 

@sjcobb, is this a use case of Perses to embed only a panel without dashboard? What would be then the builtin variables expected in that case? 

I'm not sure also about the embedded dashboard use case. I guess the builtin variable project will not be valuable in that case?

celian-garcia (changes requested):
Great that it works in the app view! but IMHO, we need to think deeper of the builtin variables in embedded use case. What would be the builtin variables available?

Reusing the external variables is practical but sounds a hack to me. Either add stuff in the `ExternalVariableDefinition` datamodel to tell they are builtin variables and tell they are valid in an embedded use case or not. Or use a separate list.

celian-garcia:
Please close the following issues in this PR as well
- close #1347 
- close #1346

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
```diff
@@ -136,11 +146,28 @@ export function useTemplateVariableStore() {
...
   const originalValues = useTemplateVariableValues();
   const definitions = useTemplateVariableDefinitions();
   const externalDefinitions = useTemplateExternalVariableDefinitions();
 
+  const initialTimeRange = useInitialTimeRange(dashboardDuration ?? DEFAULT_DASHBOARD_DURATION);
+  const convertedTimeRange = useMemo(() => {
+    console.log(isRelativeTimeRange(initialTimeRange) ? toAbsoluteTimeRange(initialTimeRange) : initialTimeRange);
+    return isRelativeTimeRange(initialTimeRange) ? toAbsoluteTimeRange(initialTimeRange) : initialTimeRange;
+  }, [initialTimeRange]);
+  const from = useMemo(() => {
```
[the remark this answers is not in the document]
Gladorme: It's useful memoize expensive operations, not the case here. I will remove it 😄

ui/app/src/utils/variables.ts
Gladorme: Same question for global var / project var. IMO, I prefer to always have them, it easily show users they exists and they are located at the end.

ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
Gladorme: Builtin variables definitions should be injected the same way dashboard variables definitions are injected. And I think it the case with current implementation, no?

sjcobb: Yeah you’re right, embedded dashboards and panels would want these, it should be more ingrained into the plugin-system instead of using the similar approach to externalVariableDefinitions

ui/app/src/utils/variables.ts
celian-garcia: There's still some differences 
1/ External variables are dynamic so user cannot guess them by experience or knowledge. While builtin variables are static and known from (future?) documentation. To me they could appear somewhere but with a different format: - Without visibility toggle since it cannot be switched on
- With a description of what is inside, and some examples 

2/ Another big difference is that we cannot override a builtin variable.

ui/dashboards/src/views/ViewDashboard/ViewDashboard.tsx
sjcobb: I also wonder if plugins should be able to opt-out somehow if project doesn’t apply to them (here is datasource specific [built-in ex](https://github.com/perses/perses/blob/3a637aae1eafcd3d24e4f8ff7a3c5e8b6c971d9e/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts#L51), no totally relevant here but in case you haven’t seen how that works)

Gladorme: I didn't know about this `$__rate_interval`, I will take a look

celian-garcia: some will have to be ingrained (from, to, dashboard?), some not (project).

celian-garcia: This is the complexity of this feature :smile: we have similar problems as with var def / external var def, but now all at once.

The difference is that now, there's no action coming from user, so action to change definitions. But still probably actions to change some state. 

We have to decide if we want to create other variables / state in the store, or create a second store/provider etc... To me put them in the same store makes sense as it is still variables and it will help you to inject them in the different panels. But definitely I'm not fan neither of reusing externalVariableDefinitions.


Also you have several layers now to manage: 
- $__project ............. will be available in Perses App
- $__dashboard ...... will be available in Perses App + embedded dashboard
- $__from & $__to ... will be available in Perses App + embedded dashboard + embedded panels

sjcobb: Yeah I don't have a great solution for app-specific variables, but `$__from` and `$__to` should be straightforward, some quick pseudo-code: 

```typescript
const { absoluteTimeRange } = useTimeRange();
replace('$__from', absoluteTimeRange.start); 
replace('$__to', absoluteTimeRange.end); 
```

For app-specific, I wonder if we could use a similar approach the [our chartsTheme](https://github.com/perses/perses/blob/3a637aae1eafcd3d24e4f8ff7a3c5e8b6c971d9e/ui/app/src/context/DarkMode.tsx#L50-L59). In that case, we wanted ability to let apps use a different ECharts theme or define different default threshold colors, opt out of tooltip pinning, etc.

How `defaultPluginKinds` is implemented could also be helpful to reference

celian-garcia: ah okay so you wouldn't leverage at all the templateVariableProvider but just the time range hook implementation.

Note that we have also in a second phase, the datasource-specific builtin variable :partying_face:

celian-garcia: Aaaah I missed some messages okay nevermind you already talekd about that

sjcobb: So I think that `replace` could live in plugin-system (possibly replace spec.query before it ever gets passed to prometheus-plugin?) 

A couple old PRs from Shan too that could be good to look at: [1](https://github.com/perses/perses/pull/1225/files#diff-52838fcf6ba629b1e359aae1ed07659b1c8165e64172d803937dcc96937bc0f4), [2](https://github.com/perses/perses/pull/1233/files#diff-fd8b76d3a0b28ddc83e9fc2f372e859de53cbce7e47b6626ab46b4f08f428fd0)
And his [comment here](https://github.com/perses/perses/discussions/226#discussioncomment-6092683) is important, we could have regex checking for `__` and would need to add validation in user-defined that does not allow users to name a variable with the `__` prefix. If we had that standardized could even build it directly into [useReplaceVariablesInString](https://github.com/perses/perses/blob/3a637aae1eafcd3d24e4f8ff7a3c5e8b6c971d9e/ui/plugin-system/src/runtime/template-variables.ts#L161C10-L167) 🤔

sjcobb: This repo is out of date now but to understand embed-use case it can be helpful to look at, [here are where providers are defined](https://github.com/sjcobb/perses-embed-example/blob/eeefcb48451faab301f88f6341031e645d29dc52/components/dashboards/PersesDashboardProvidersAlt.tsx#L46-L49) and that app can decide to change 'TimeSeriesChart' as the default panel kind if they wanted to.

And here is example of how an app can [override the default ECharts theme](https://github.com/sjcobb/perses-embed-example/blob/eeefcb48451faab301f88f6341031e645d29dc52/components/dashboards/PersesDashboardProviders.tsx#L19-L42C14). I'm thinking that the new provider @celian-garcia mentioned will be needed and it can inject app-specific context about built-in variables. Then TemplateVariableProvider can use the injected dependencies without caring about where they come from (so if that demo Next.js app has no concept of "projects" it could inject info about "workspaces" or "collections" or if RedHat / OpenShift has some other grouping concept in their info model)

sjcobb: Or hopefully the Prometheus UI someday will use Perses 🤞 and could leverage this new provider. In that case, they'll be using a panel to replace [the UI here](https://prometheus.demo.do.prometheus.io/graph?g0.expr=up&g0.tab=0&g0.stacked=0&g0.range_input=1h)

celian-garcia: Thanks for the examples, this will be very useful as we're currently thinking to integrate a perses panel in our home made alerting configuration tool in Amadeus. :wink:

ui/app/src/utils/variables.ts
Gladorme: I changed the listing design for builtin variables 😄

celian-garcia:
image

celian-garcia:
MicrosoftTeams-image (3)

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
```diff
@@ -136,11 +146,27 @@ export function useTemplateVariableStore() {
...
+  children: ReactNode;
+  dashboardDuration?: DurationString;
+}
+
+function PluginProvider({ children, dashboardDuration }: PluginProviderProps) {
   const originalValues = useTemplateVariableValues();
   const definitions = useTemplateVariableDefinitions();
   const externalDefinitions = useTemplateExternalVariableDefinitions();
 
+  const initialTimeRange = useInitialTimeRange(dashboardDuration ?? DEFAULT_DASHBOARD_DURATION);
```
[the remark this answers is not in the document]
celian-garcia: Now I guess you're using ``useTimeRange`` everywhere. Less confusing for me :+1: . You can resolve this

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
```diff
@@ -136,11 +146,27 @@ export function useTemplateVariableStore() {
...
+  const convertedTimeRange = useMemo(() => {
+    return isRelativeTimeRange(initialTimeRange) ? toAbsoluteTimeRange(initialTimeRange) : initialTimeRange;
```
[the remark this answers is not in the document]
celian-garcia: good to know ! resolve discussion ?

internal/api/shared/validate/validate.go
```diff
@@ -51,6 +59,16 @@ func Datasource[T modelV1.DatasourceInterface](entity T, list []T, sch schemas.S
 	return nil
 }
 
+func Variable[T modelV1.VariableInterface](entity T, sch schemas.Schemas) error {
+	if err := validateVariableName(entity.GetMetadata().GetName()); err != nil {
+		return err
+	}
+	if err := sch.ValidateGlobalVariable(entity.GetVarSpec()); err != nil {
```
celian-garcia, line 66: validateGlobalVariable ?

pkg/model/api/v1/utils/variable_build_order.go
```diff
@@ -229,20 +238,7 @@ func extractVariableInStringOrInSomethingElse(v reflect.Value, matches *[][]stri
...
-	matches := variableTemplateSyntaxRegexp.FindAllStringSubmatch(str, -1)
-	var result [][]string
-	for _, match := range matches {
-		if _, err := strconv.Atoi(match[1]); err != nil {
-			// We want to keep only variables that are not only a number.
-			// A number that represents a variable is not meaningful, and so we don't want to consider it.
-			// It's also a way to avoid a collision in terms of variable template syntax.
-			// For example in PromQL, in the function `label_replace`, it used the syntax $1, $2, for the placeholder.
-			//
-			// If the string cannot be parsed as an integer, then we can keep it because that means it contains other characters than just numbers.
```
celian-garcia, line 241: From the description I read, this sounds important to keep

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -42,19 +43,21 @@ describe('Panel', () => {
...
 
     renderWithContext(
-      <TemplateVariableProvider
-        initialVariableDefinitions={[
-          {
-            kind: 'TextVariable',
-            spec: {
-              name: 'foo',
-              value: 'bar ',
+      <TimeRangeProvider initialTimeRange={{ pastDuration: '1h' }}>
```
celian-garcia, line 46: It means the TemplateVariableProvider cannot be used without being under a time range provider? What is the error message in this case ?

internal/api/shared/validate/validate.go
Gladorme: https://github.com/perses/perses/blob/main/internal/api/shared/schemas/schemas.go#L65C1-L73C2

ui/dashboards/src/components/Panel/Panel.test.tsx
Gladorme: Yep, you need the Time Range provider because of the `$__from`/`$__to` variables. It will print something like this: No provider found... Did you forget to import a Provider...

pkg/model/api/v1/variable.go
```diff
@@ -16,12 +16,18 @@ package v1
...
-
 	modelAPI "github.com/perses/perses/pkg/model/api"
 	"github.com/perses/perses/pkg/model/api/v1/variable"
 	"gopkg.in/yaml.v2"
 )
 
+var BuiltinVariableNames = []string{"$__project", "$__dashboard", "$__from", "$__to"}
+
+type VariableInterface interface {
+	GetMetadata() modelAPI.Metadata
```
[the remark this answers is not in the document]
celian-garcia: haha ok

ui/plugin-system/src/runtime/builtin-variables.ts
```diff
@@ -0,0 +1,68 @@
...
+  }, [names, states]);
+
+  if (names === undefined) {
+    return states;
+  }
+
+  return values;
+}
+
+export function useBuiltinVariableDefinitions(): BuiltinVariableDefinition[] {
```
celian-garcia, line 65: You are cheating here :smile: 

The only drawback of that is that you won't be able to control the order of variables. If the default way to order entries of a map in typescript is ok for you I guess it's fine.

celian-garcia (approved):
LGTM! nice ! We finally have builtin variables

Nexucis (approved):
changes looks great 😊. I didn't review in details the frontend sorry about that. I'm trusting @sjcobb and @celian-garcia it was ok for them ;)

PR #1351, opened 2023-08-03
[BUGFIX] Default selection for variables of type list

### LINKED ISSUE #1312: Default selection for variables of type list
If the default_value field of a variables is not set, the default selection is equal to "", even if the list contains some values. 
I would naturally expect to see the first value as a default selection.  

If the All selection is allowed, we could also set All as default selection, but that's a bit more opinionated.

celian-garcia:
Encapsulates the complex part which manages the default selected value of a list variable into a function. 
It allowed me to make some interesting tests and fixes.
 
fixes #1312 

It's implementing the flow described in the issue + an additional behavior : when the default value or value from query param is not present in the list, it falls back to the first value of the list. 

# Checklist

## UI Changes

- [ ] ~~Changes that impact the UI include screenshots and/or screencasts of the relevant changes.~~
> No screenshots as it is quite hard to capture in a screen. I let people reviewing the PR play with it locally.

### DIFF
--- a/ui/dashboards/src/components/Variables/TemplateVariable.tsx
+++ b/ui/dashboards/src/components/Variables/TemplateVariable.tsx
@@ -13,8 +13,16 @@
 
 import { useEffect, useMemo, useState } from 'react';
 import { Select, FormControl, InputLabel, MenuItem, Box, LinearProgress, TextField } from '@mui/material';
-import { DEFAULT_ALL_VALUE, ListVariableDefinition, VariableName, VariableValue } from '@perses-dev/core';
-import { useListVariablePluginValues } from '@perses-dev/plugin-system';
+import {
+  DEFAULT_ALL_VALUE,
+  ListVariableDefinition,
+  ListVariableSpec,
+  UnknownSpec,
+  VariableName,
+  VariableValue,
+} from '@perses-dev/core';
+import { useListVariablePluginValues, VariableOption, VariableState } from '@perses-dev/plugin-system';
+import { UseQueryResult } from '@tanstack/react-query';
 import { useTemplateVariable, useTemplateVariableActions } from '../../context';
 
 type TemplateVariableProps = {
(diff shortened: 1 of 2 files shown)


ui/dashboards/src/components/Variables/TemplateVariable.tsx
```diff
@@ -74,29 +78,76 @@ function ListVariable({ name, source }: TemplateVariableProps) {
...
+
+  // Once we computed value, we set it as the selected one, if it is available in the options
+  let selectedValue = value;
   if (!valueIsInOptions) {
-    selectValue = allowMultiple ? [] : '';
+    selectedValue = allowMultiple ? [] : '';
   }
 
+  if (ctx.definition.spec.name === 'instance') {
+    console.log('=> value', value);
```
Gladorme, line 117: 👀

sjcobb (approved):

ui/dashboards/src/components/Variables/TemplateVariable.tsx
```diff
@@ -74,29 +82,68 @@ function ListVariable({ name, source }: TemplateVariableProps) {
...
+
+  // Update value when changed
   useEffect(() => {
-    const firstOption = finalOptions?.[0];
+    if (value) {
+      setVariableValue(name, value, source);
+    }
+  }, [setVariableValue, name, value, source]);
+
+  // Update loading when changed
```
sjcobb, line 136: Approved but I have a hunch that some of these hooks could be refactored and removed, didn’t get a chance to look into more specifics but newer React docs are emphasizing that useEffect can be unnecessary in many cases https://react.dev/learn/you-might-not-need-an-effect

celian-garcia: I fully agree to refactor it! Here I just encapsulate and test but for sure it doesn't make the code more readable, just a bit safer. Also in this particular case, it's tricky as we don´t want to update value too often or we have an infinite loop with the subsequent setVariableValue :smile:

celian-garcia: I can try to remove the useEffect for loading and options, but I'm sure that removing it for value would do some infinite loop...

Gladorme (approved):

PR #1359, opened 2023-08-10
move pkg integration test outside

Nexucis:
This PR is moving the integration in the internal package.
This is following the conversation from #1190. Moving them outside of `pkg` will reduce the number of indirect dependencies you have in the go.mod when you are importing perses.

### DIFF
--- a/internal/api/e2e/api/dashboard_test.go
+++ b/internal/api/e2e/api/dashboard_test.go
@@ -13,7 +13,7 @@
 
 //go:build integration
 
-package e2e
+package api
 
 import (
 	"fmt"
--- a/internal/api/e2e/api/datasource_test.go
+++ b/internal/api/e2e/api/datasource_test.go
@@ -13,7 +13,7 @@
 
 //go:build integration
 
-package e2e
+package api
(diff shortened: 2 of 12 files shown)


internal/api/e2e/api/dashboard_test.go
```diff
@@ -13,7 +13,7 @@
 
 //go:build integration
 
-package e2e
```
celian-garcia, line 16: I´d be in favor of keeping e2e as package, this is easier to read in my opinion but maybe this is only me 

Like having ``internal/api/e2e`` and ``internal/client/e2e``

Nexucis: we will create a client package just for the e2e tests, that's a bit weird don't you think ?

celian-garcia: aaah nevermind I thought client was standing for the cli but it is the api client. Nevermind fine

celian-garcia (approved):

Gladorme (approved):
LGTM

PR #1360, opened 2023-08-10
[BREAKINGCHANGE] Datasource selection by group

celian-garcia:
# Description

Currently the datasource selection when we use it in a panel or variable is quite limited. It is not displaying the eventually overridden datasource (e.g a global datasource when a project datasource exist with the same name)

I'm proposing to keep this well split by group (default VS local VS project VS global datasource ) so the user understand better from where the datasource selected comes from. 

As a bonus, it continues on the implementation of the local datasources. Specially the part that computes the proxy url on the fly + related proxy middleware in the backend.

# Breaking changes
- ``DatasourceApi``: Moved build of proxy url to dedidated optional method
- ``DatasourceStore``: Modified the datasource "metadata" to a list of selection items, grouped by group

# Screenshots

image

# References
- internal
  - relates #1033 
- react
  - impossible to make the chip clickable if the item is disabled. I guess this is ok to keep it like this for now.
https://github.com/mui/material-ui/issues/36572

# Checklist

## UI Changes

### DIFF
--- a/internal/api/core/middleware/proxy.go
+++ b/internal/api/core/middleware/proxy.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/labstack/echo/v4"
+	"github.com/perses/perses/internal/api/interface/v1/dashboard"
 	"github.com/perses/perses/internal/api/interface/v1/datasource"
 	"github.com/perses/perses/internal/api/interface/v1/globaldatasource"
 	"github.com/perses/perses/internal/api/interface/v1/globalsecret"
@@ -35,15 +36,16 @@ import (
 )
 
 var (
-	globalProxyMatcher  = regexp.MustCompile(`/proxy/globaldatasources/([a-zA-Z-0-9_-]+)(/.*)?`)
-	projectProxyMatcher = regexp.MustCompile(`/proxy/projects/([a-zA-Z-0-9_-]+)/datasources/([a-zA-Z-0-9_-]+)(/.*)?`)
+	globalProxyMatcher    = regexp.MustCompile(`/proxy/globaldatasources/([a-zA-Z-0-9_-]+)(/.*)?`)
+	projectProxyMatcher   = regexp.MustCompile(`/proxy/projects/([a-zA-Z-0-9_-]+)/datasources/([a-zA-Z-0-9_-]+)(/.*)?`)
+	dashboardProxyMatcher = regexp.MustCompile(`/proxy/projects/([a-zA-Z-0-9_-]+)/dashboards/([a-zA-Z-0-9_-]+)/datasources/([a-zA-Z-0-9_-]+)(/.*)?`)
 )
 
 // TODO cache the request to the database
 
 func extractGlobalDatasourceAndPath(requestPath string) (dtsName string, path string, err error) {
 	matchingGroups := globalProxyMatcher.FindAllStringSubmatch(requestPath, -1)
-	if len(matchingGroups) > 1 || len(matchingGroups[0]) <= 1 {
+	if len(matchingGroups) > 1 || len(matchingGroups) == 0 || len(matchingGroups[0]) <= 1 {
 		return "", "", echo.NewHTTPError(http.StatusBadGateway, "unable to forward the request to the datasource, request not properly formatted")
 	}
 	dtsName = matchingGroups[0][1]
@@ -58,7 +60,7 @@ func extractGlobalDatasourceAndPath(requestPath string) (dtsName string, path st
 
 func extractProjectDatasourceAndPath(requestPath string) (projectName string, dtsName string, path string, err error) {
 	matchingGroups := projectProxyMatcher.FindAllStringSubmatch(requestPath, -1)
-	if len(matchingGroups) > 1 || len(matchingGroups[0]) <= 2 {
+	if len(matchingGroups) > 1 || len(matchingGroups) == 0 || len(matchingGroups[0]) <= 2 {
 		return "", "", "", echo.NewHTTPError(http.StatusBadGateway, "unable to forward the request to the datasource, request not properly formatted")
 	}
 	projectName = matchingGroups[0][1]
@@ -72,7 +74,25 @@ func extractProjectDatasourceAndPath(requestPath string) (projectName string, dt
 	return
 }
 
+func extractProjectDashboardDatasourceAndPath(requestPath string) (projectName string, dashboardName string, dtsName string, path string, err error) {
+	matchingGroups := dashboardProxyMatcher.FindAllStringSubmatch(requestPath, -1)
+	if len(matchingGroups) > 1 || len(matchingGroups) == 0 || len(matchingGroups[0]) <= 3 {
+		return "", "", "", "", echo.NewHTTPError(http.StatusBadGateway, "unable to forward the request to the datasource, request not properly formatted")
+	}
+	projectName = matchingGroups[0][1]
+	dashboardName = matchingGroups[0][2]
+	dtsName = matchingGroups[0][3]
+	// Based on the HTTP 1.1 RFC, a `/` should be the minimum path.
+	// https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2
+	path = "/"
+	if len(matchingGroups[0]) > 4 && len(matchingGroups[0][4]) > 0 {
+		path = matchingGroups[0][4]
+	}
+	return
+}
+
 type Proxy struct {
+	Dashboard    dashboard.DAO
 	Secret       secret.DAO
 	GlobalSecret globalsecret.DAO
 	DTS          datasource.DAO
@@ -84,15 +104,19 @@ func (e *Proxy) Proxy() echo.MiddlewareFunc {
 		return func(c echo.Context) error {
 			requestPath := c.Request().URL.Path
 			globalDatasourceMatch := globalProxyMatcher.MatchString(requestPath)
-			localDatasourceMatch := projectProxyMatcher.MatchString(requestPath)
-			if !globalDatasourceMatch && !localDatasourceMatch {
+			projectDatasourceMatch := projectProxyMatcher.MatchString(requestPath)
+			dashboardDatasourceMatch := dashboardProxyMatcher.MatchString(requestPath)
+			if !globalDatasourceMatch && !projectDatasourceMatch && !dashboardDatasourceMatch {
 				// this is likely a request for the API itself
 				return next(c)
 			}
 			if globalDatasourceMatch {
 				return e.proxyGlobalDatasource(c)
 			}
-			return e.proxyProjectDatasource(c)
+			if projectDatasourceMatch {
+				return e.proxyProjectDatasource(c)
+			}
+			return e.proxyDashboardDatasource(c)
 		}
 	}
 }
@@ -133,6 +157,24 @@ func (e *Proxy) proxyProjectDatasource(ctx echo.Context) error {
 	return pr.serve(ctx)
 }
 
+func (e *Proxy) proxyDashboardDatasource(ctx echo.Context) error {
+	projectName, dashboardName, dtsName, path, err := extractProjectDashboardDatasourceAndPath(ctx.Request().URL.Path)
+	if err != nil {
+		return err
+	}
+	dts, err := e.getDashboardDatasource(projectName, dashboardName, dtsName)
+	if err != nil {
+		return err
+	}
+	pr, err := newProxy(dts, path, func(name string) (*v1.SecretSpec, error) {
+		return e.getProjectSecret(projectName, dtsName, name)
+	})
+	if err != nil {
+		return err
+	}
+	return pr.serve(ctx)
+}
+
 func (e *Proxy) getGlobalDatasource(name string) (v1.DatasourceSpec, error) {
 	dts, err := e.GlobalDTS.Get(name)
 	if err != nil {
@@ -145,6 +187,7 @@ func (e *Proxy) getGlobalDatasource(name string) (v1.DatasourceSpec, error) {
 	}
 	return dts.Spec, nil
 }
+
 func (e *Proxy) getProjectDatasource(projectName string, name string) (v1.DatasourceSpec, error) {
 	dts, err := e.DTS.Get(projectName, name)
 	if err != nil {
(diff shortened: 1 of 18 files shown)


ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -181,35 +182,72 @@ function findDashboardDatasource(dashboardDatasources: DashboardSpec['datasource
...
+    const index = groupIndices[group] ?? -1;
+    let results = resultsByGroup[index];
+    if (!results) {
+      groupIndices[group] = currentGroupIndex;
+      results = { data: [], group, editLink };
+      resultsByGroup[currentGroupIndex] = results;
+      currentGroupIndex++;
+    }
+
     // If we already have a datasource with this selector name, ignore it, otherwise add to end of list
```
celian-garcia, line 238: with this selector name and group

ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -181,35 +182,72 @@ function findDashboardDatasource(dashboardDatasources: DashboardSpec['datasource
...
         name: selectorName,
+        group,
       },
     });
-    usedNames.add(selectorName);
+    usedNames.add({ name: selectorName, group });
   };
 
-  return { results, addResult };
+  return { results: resultsByGroup, addResult };
```
celian-garcia, line 252: keep it named results for clarity and rename the one in the function

ui/plugin-system/src/components/DatasourceSelect.tsx
```diff
@@ -31,32 +32,73 @@ export interface DatasourceSelectProps extends Omit<SelectProps<string>, Omitted
...
+        group: ds.selector.group,
+        value: selectorToOptionValue(ds.selector),
+      })),
+    }));
   }, [data]);
 
   // While loading available values, just use an empty string so MUI select doesn't warn about values out of range
   const optionValue = isLoading ? '' : selectorToOptionValue(value);
 
+  console.log('TEST ====');
```
celian-garcia, line 56: console.log

ui/plugin-system/src/components/DatasourceSelect.tsx
```diff
@@ -11,9 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Select, SelectProps, MenuItem } from '@mui/material';
+import OpenInNewIcon from 'mdi-material-ui/OpenInNew';
+import { Select, SelectProps, MenuItem, Stack, Divider, ListItemText, Chip, IconButton } from '@mui/material';
 import { DatasourceSelector } from '@perses-dev/core';
-import { useMemo } from 'react';
+import React, { useMemo } from 'react';
```
celian-garcia, line 17: unused

ui/plugin-system/src/runtime/datasources.ts
```diff
@@ -27,10 +27,16 @@ export interface DatasourceStore {
   /**
    * Gets a list of datasource metadata for a plugin kind.
    */
-  listDatasourceMetadata(datasourcePluginKind: string): Promise<DatasourceMetadata[]>;
+  listDatasourceMetadata(datasourcePluginKind: string): Promise<DatasourceMetadata>;
 }
 
-export interface DatasourceMetadata {
+export type DatasourceMetadata = Array<{
```
celian-garcia, line 33: Rename ``DatasourceMetadata`` and ``DatasourceMetadataEntry`` to something more meaningful like 

``DatasourceSelectItemGroup[]`` and ``DatasourceSelectItem[]``

ui/prometheus-plugin/src/model/prometheus-selectors.ts
```diff
@@ -31,7 +31,7 @@ export const DEFAULT_PROM: PrometheusDatasourceSelector = { kind: PROM_DATASOURC
  * Returns true if the provided PrometheusDatasourceSelector is the default one.
  */
 export function isDefaultPromSelector(selector: PrometheusDatasourceSelector) {
-  return selector.name === undefined;
+  return isDefaultSelector(selector);
```
celian-garcia, line 34: Not relevant anymore as finally not used

AntoineThebaud:
FYI we brainstormed about this display with mohit some time ago and we came up with some mockups, I dont know if you had a look at them? Cant find the link right now, but e.g if I remember well we were thinking to put the display of the overriden datasources at the end, to not impact the navigation of a user already aware of what he can use and what he can't.

celian-garcia:
> FYI we brainstormed about this display with mohit some time ago and we came up with some mockups, I dont know if you had a look at them? Cant find the link right now, but e.g if I remember well we were thinking to put the display of the overriden datasources at the end, to not impact the navigation of a user already aware of what he can use and what he can't.

@AntoineThebaud We talked together and we are not sure it won't bring more confusion so I won't do it in that PR. I let you create a discussion if you really think it's worth it.

Gladorme (approved):
LGTM 😄

Gladorme (changes requested):
When clicking on panel, datasource is not selected

internal/api/core/middleware/proxy.go
```diff
@@ -55,24 +57,28 @@ func Proxy(dts datasource.DAO, globalDTS globaldatasource.DAO) echo.MiddlewareFu
...
+	if projectDatasourceMatch {
+		return getProjectDatasourceAndPath(dts, requestPath)
+	}
+	return getLocalDatasourceAndPath(dashboardDAO, requestPath)
 }
 
 func getGlobalDatasourceAndPath(dao globaldatasource.DAO, requestPath string) (v1.DatasourceSpec, string, error) {
 	matchingGroups := globalProxyMatcher.FindAllStringSubmatch(requestPath, -1)
-	if len(matchingGroups) > 1 || len(matchingGroups[0]) <= 1 {
+	if len(matchingGroups) == 0 || len(matchingGroups[0]) <= 1 {
```
Nexucis, line 81: I think you should keep the check `len(matchingGroups) > 1 `

ui/app/src/model/datasource-api.ts
```diff
@@ -11,40 +11,49 @@
...
+   * - /proxy/projects/{project}/datasources/{name}
+   * - /proxy/projects/{project}/dashboards/{dashboard}/{name}
+   *
+   * NB: despite the fact it's possible, it is useless to give a dashboard without a project as the url will for sure
+   * correspond to nothing.
+   * @param props.name
+   * @param props.dashboard
+   * @param props.project
+   */
+  buildProxyUrl(props: { project?: string; dashboard?: string; name: string }): string {
```
Nexucis, line 34: this method is still a utils method and as such doesn't need to be attached to a class.

celian-garcia: I thought so but I thought of embeded use case, if the build of proxy url was part of the implementation of DatasourceApi before, then it should remain the same for local dashboards as well. Otherwise somebody integrating a embedded dashboard will be forced to use the proxy url format of the backend

Nexucis: well if you are linked this method to the interface, then you are inducing the usage of proxy when it's not necessary the case depending of your usage.

You shouldn't provide or induce how the proxy url should be built. So for me it is still an utils method.

celian-garcia: I understand, but it was linked with the interface, I didn't change that. And actually I understand why it was linked with the interface.(Cf my previous answer)
I just moved it into a dedicated method instead of the return type. I can maybe set the method optional so that somebody implementing the DatasourseApi interface doesn't have to implement the buildProxyUrl. I will try.

celian-garcia: Let me give you an example with something you know well. If I develop a panel for the Argos UI, I will want the proxy url to be "/proxy/prometheus". I don´t want it to be computed according to an assomption of my backend. This was already possible before my implementation because the get methods of DatasourceApi was returning the proxyurl of each found datasource. But now, as we need also the datasource from the dashboard itself, we need something. 

I confess it's not ideal to expose all of it to somebody who want just an embedded panel, but well it's the way it's done as far as I understand :/

Nexucis: yeah ok.

I re-read the code and ok you just change the way the `proxyURL` was exposed.

@jgbernalp This PR will break the datasource selector. I don't know exactly where are you in the integration of Perses but perhaps you would like to look at this PR. Perhaps you will have a different opinion about how the proxyURL should be exposed (like before, like it is proposed in this PR through a method in the interface, or in a different way)

celian-garcia:
> When clicking on a panel, datasource is not selected

this is fixed now, the backend was sending back a datasource selector with only kind and name, so I added a code that would find back the group in the `DatasourceSelect` component.

ui/core/src/model/datasource.ts
```diff
@@ -42,10 +42,21 @@ export interface DatasourceSpec<PluginSpec = UnknownSpec> {
...
 export interface DatasourceSelector {
+  /**
+   * Kind of the datasource.
+   */
   kind: string;
+  /**
+   * Group of the datasource.
+   * Omit it if you don't store datasource by group.
+   */
+  group?: string;
```
Nexucis, line 56: what do you mean `store datasource by group` ? Can you explain a bit furthermore ?

celian-garcia: I mean that for embedded use case, if it is useless for the developer to have different level of datasources, it can just omit it. Then there won't be any grouping of the datasources and the chip in the selector won't be displayed at all. I can elaborate a bit more in the comment.

Nexucis: I don't get what you are doing here. The datasource selector represents the way the datasource will be selected based on the info stored in the dashboard.
A group is not reflected in the dashboard data-model and it shouldn't be since we defined a long time ago that only `name` and `kind` will be the parameters used to select the datasource you want to use.

Then the system behind will find the appropriate datasource. We talked also about selecting a special datasource coming from a dedicated group but we decided to not consider it to have something simpler and an override system with no exception.

This shouldn't be altered when changing the display of the datasource dropdown from my opinion.

celian-garcia: This group field is used in UI only. Ignored by backend. 

In a perfect world, I would do a dedicated data model for the view but it requires a mapper that I didn't manage to put in place because of the architecture of the UI, tightly coupled with the API datamodel. :/ 

Any help from you or @AntoineThebaud  would be appreciated, if we can do so alright!

ui/app/src/model/datasource-api.ts
jgbernalp: @Nexucis we are using the direct_url field on the datasource to fetch data from a service present inside a k8s cluster. So I have no strong opinion about the proxyURL.

ui/core/src/model/datasource.ts
celian-garcia: Exactly what we talked about yes.
Good idea to check variables, but the VariableDefinition is actually the type used at the same time for the view and backend request. The difference in variables is that the provider is done differently, with a zustand store. It allows us to decouple definition coming from the backend, and state coming from the view. 

The spirit is completely different in the datasource provider. The solution is probably to integrate zustand in the datasource provider the variable one but it would require a big refactoring, I'm not sure I am skilled enough. 

It makes me think of Steven who was regularly saying "hey guys what about having zustand store for the datasources ?" This is typically the situation where we can say he's right haha

ui/app/src/model/datasource-api.ts
celian-garcia: @jgbernalp I'm thinking of having the method that build the proxy url as optional so you could even not have to return a fake "" (as you might do currently)

jgbernalp: It seems more clear the new approach, to return what is being requested, rather than a wrapper object like before `{ resource: GlobalDataSource ...`.  I'm not against having the buildProxyUrl into a new function, but since this is not a react component maybe we can remove the `props` and just pass a plain parameter list.

celian-garcia: Yeah indeed I prefer the returns like that. 

Regarding the props, actually I like the way the props are passed in component and I wanted on purpose to do the same here as I have some optional/not optional parameters. 

First, we don´t have to guess the order of parameters, have project and/or dashboard undefined can be cumbersome sometimes. 
E.g if you want to have order name, project, dashboard and you don't want to define project, you have to put 
"mydatasource", undefined, "mydashboard"  

Second, we have named parameters which is quite easier to read. 

I can rename props to something else if you prefer like params

ui/app/src/model/datasource-api.ts
```diff
@@ -11,40 +11,49 @@
...
+    const { project, dashboard, name } = props;
+    let url = `${!project && !dashboard ? 'globaldatasources' : 'datasources'}/${name}`;
+    if (dashboard) {
+      url = `dashboards/${dashboard}/${url}`;
+    }
+    if (project) {
+      url = `projects/${project}/${url}`;
+    }
+    return `/proxy/${url}`;
+  }
```
jgbernalp, line 44: @celian-garcia continuing the conversation here as I needed to add more lines to the suggestion. In that case it can be changed as follows: ( I added the encodeURIComponent as this names might contain special characters )
```suggestion
  buildProxyUrl({ project, dashboard, name } : { project?: string; dashboard?: string; name: string }): string {
    let url = `${!project && !dashboard ? 'globaldatasources' : 'datasources'}/${encodeURIComponent(name)}`;
    if (dashboard) {
      url = `dashboards/${encodeURIComponent(dashboard)}/${url}`;
    }
    if (project) {
      url = `projects/${encodeURIComponent(project)}/${url}`;
    }
    return `/proxy/${url}`;
  }
```

celian-garcia: LGTM thanks, I will do the change

celian-garcia: Updated the PR with your suggestion + the buildProxyUrl method is now optional, so if you don't use the proxy feature, you don't need to implement it.

ui/app/src/model/datasource-api.ts
celian-garcia: So far I think the original thread can be resolved, moving the build of proxy url to a dedicated method seems to be ok. 

I mark this issue as resolved

internal/api/shared/schemas/base_def_query.cue
```diff
@@ -15,9 +15,9 @@ package base
 
 kind: string
 spec: {
-	datasource?: close({
```
AntoineThebaud, line 18: Just in case, context for future reference: closing the struct here was preventing to add the level/group info on frontend side, that should be ignored by the backend. 
In the long run we may want to decouple frontend & backend definitions. The day this is done, we can put back the `close()` call here.

internal/api/core/middleware/proxy.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 The Perses Authors
```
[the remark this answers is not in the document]
celian-garcia: Nom de Zeus !

ui/plugin-system/src/components/DatasourceSelect.tsx
```diff
@@ -32,51 +33,124 @@ export interface DatasourceSelectProps extends Omit<SelectProps<string>, Omitted
  */
 export function DatasourceSelect(props: DatasourceSelectProps) {
   const { datasourcePluginKind, value, onChange, ...others } = props;
-  const { data, isLoading } = useListDatasources(datasourcePluginKind);
+  const { data, isLoading } = useListDatasourceSelectItems(datasourcePluginKind);
 
-  // Convert the datasource list into menu items with name/value strings that the Select input can work with
+  // Rebuild the group of the value if not provided
+  const defaultValue = useMemo(() => {
```
celian-garcia, line 39: Note to myself : this is still not fully working. In some cases, nothing is selected. I think it is when there is no datasource selected and no default datasource exist in the items list. 

This specific case could be encountered and should be more explicit to the user. 

Then if it occurs
- should we select the first of the list as default?
- should we put a warning (error?) saying that if you don't explicitly select a datasource, none of them are declared as default?

celian-garcia: For the record, we discussed together and here is the output: 
If no default datasource exist, take the first one, but it should still be considered as default (no ds selected in the backend). This is possible using the same way the store provider setup the classic default. It means that there will always be a default value in the selection list.

ui/core/src/model/datasource.ts
celian-garcia: @Nexucis @AntoineThebaud Is it ok for you if we leave it like that or no ? Decouple the backend model and view model for the ``DatasourceSelector`` will be a challenge.

I agree that it's not ideal to put a group there knowing that it is considered differently in the backend and frontend, but that would be too much work for this PR. 
I just added a comment 
image

Nexucis: I would be ok with that if we are ok and are going to solve the TODO

celian-garcia: Yes, I can work on that later. As you know, I'm already thinking of adopting a perses panel in our UI, and in that context, I would like to implement a simpler datasource provider for this kind of purpose. Something more agnostic to the backend.

Just to say that I'm interested in all that datasource provider topic and I'm curious to work on that and discover the other benefits to have a zustand store here.

Nexucis (approved):
thank you for working on this and sorry for the time it took to merge it. It was quite complicated to evaluate the results of the breaking change

AntoineThebaud (approved):

PR #1366, opened 2023-08-17
[BUGFIX] Fix direct link with unknown project redirect now to the home page

### LINKED ISSUE #1349: Direct link to a not existing dashboard/project should error and redirect
Current behavior:

=> Dashboard: empty
image

=> Project: This one is fun, it's not empty but display a fake project
image

Nexucis:
This PR is introducing a sort of guard on the route.

It will check if the project exist and in case it doesn't exist, the user is redirected to the home page.

fix #1349

### DIFF
--- a/ui/app/src/model/project-client.ts
+++ b/ui/app/src/model/project-client.ts
@@ -112,3 +112,11 @@ export function useDeleteProjectMutation() {
     },
   });
 }
+
+export function fetchProject(name: string) {
+  const url = buildURL({
+    resource: resource,
+    name: name,
+  });
+  return fetchJson<ProjectResource>(url);
+}
--- a/ui/app/src/views/ViewDashboard.tsx
+++ b/ui/app/src/views/ViewDashboard.tsx
@@ -66,7 +66,7 @@ function ViewDashboard() {
     datasourceApi.listDatasources(projectName);
     datasourceApi.listGlobalDatasources();
   }, [datasourceApi, projectName]);
-  const { isLoading } = useDashboard(projectName, dashboardName);
+  const { isLoading, error } = useDashboard(projectName, dashboardName);
   let { data } = useDashboard(projectName, dashboardName);
   const isReadonly = useIsReadonly();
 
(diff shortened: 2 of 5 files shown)


ui/app/src/model/project-client.ts
```diff
@@ -112,3 +112,11 @@ export function useDeleteProjectMutation() {
     },
   });
 }
+
+export function fetchProject(name: string) {
```
celian-garcia, line 116: It is preferrable to make a useProject function using tanstack as it is the direction taken by all the clients. 

If you want to receive a promise, you should use mutation and mutateAsync method

celian-garcia: https://tanstack.com/query/v3/docs/react/guides/mutations
https://tanstack.com/query/v3/docs/react/guides/mutations#promises

celian-garcia (changes requested):

ui/app/src/model/project-client.ts
Nexucis: mutate or mutateAsync is not the correct way when you want to fetch data :/. It's more used when you update or create data

ui/app/src/views/ViewDashboard.tsx
```diff
@@ -161,6 +161,10 @@ function ViewDashboard() {
   }, [actionRef, navigate, projectName]);
 
   if (isLoading) return null;
+  if (error !== null) {
```
celian-garcia, line 164: ``if (error && actionRef.current !== CreateAction) `` ? this is it @Gladorme ?

Gladorme: Yes

ui/app/src/model/project-client.ts
Nexucis: I don't really see the value of doing that. Using fetch directly is doing exactly what is expected and it is simpler. Yes it doesn't have a cache or whatever react-query has, but it's not necessary here. The simpler the better

celian-garcia: Yes we need to discuss about that custom cache in datasources. I think it should be migrated to react-query. 
I don´t understand what justified the creation of this cache wrapper. Is this historical ? because react-query was not used before? Or is there a specific feature that you make and which is not supported by react-query? Maybe you can explain me. Steven told me he tried also to migrate it to react-query some day, without success.

celian-garcia: I guess we can keep it like that as it seems you don't want to have cache for project guard...

PR #1373, opened 2023-08-21
[FEATURE] Support secret CRUD on the API side

Nexucis:
This PR is adding two new kind of object named `Secret` and `GlobalSecret`.

These two objects should be used to configure the HTTP Proxy for the datasource. It can be used for other purpose if we need secret for something else.

Note:
1. I didn't update yet the proxy so it can use these new configurations
2. I didn't take care about how to store properly the secret in the database (aka hash + salt of the sensitive informations)


internal/api/impl/v1/globalsecret/service.go
```diff
@@ -0,0 +1,89 @@
...
+	// Update the time contains in the entity
+	entity.Metadata.CreateNow()
+	if err := s.dao.Create(entity); err != nil {
+		return nil, err
+	}
+	return entity, nil
+}
+
+func (s *service) Update(entity api.Entity, parameters shared.Parameters) (interface{}, error) {
+	if projectObject, ok := entity.(*v1.GlobalSecret); ok {
```
celian-garcia, line 55: object

internal/api/impl/v1/globalsecret/service.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func NewService(dao globalsecret.DAO) globalsecret.Service {
+	return &service{
+		dao: dao,
+	}
+}
+
+func (s *service) Create(entity api.Entity) (interface{}, error) {
+	if projectObject, ok := entity.(*v1.GlobalSecret); ok {
```
celian-garcia, line 39: object

celian-garcia (approved):
I didn't find how you obfuscate the result in GET requests. I guess you will do it following PRs. Apart from that and some very minor typo, LGTM

Nexucis:
>  I guess you will do it following PRs

Yep that's what I have in mind

PR #1377, opened 2023-08-23
[ENHANCEMENT] Use secret in the proxy

Nexucis:
This PR is adding the usage of the secret in the proxy.

In the meantime, the proxy is now a list of endpoint instead of being a middleware. It simplify a couple of things (like it  removes the usage of regexp to catch the path, the name of the datasource)


internal/api/core/middleware/proxy.go
```diff
@@ -25,123 +25,197 @@ import (
...
-	// Based on the HTTP 1.1 RFC, a `/` should be the minimum path.
-	// https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2
-	path := "/"
-	if len(matchingGroups[0]) > 3 {
-		path = matchingGroups[0][3]
+	return &scrt.Spec, nil
+}
+
+func (e *Proxy) getProjectSecret(projectName string, dtsName string, name string) (*v1.SecretSpec, error) {
+	scrt, err := e.Secret.Get(projectName, name)
```
celian-garcia, line 177: sometimes I hate Golang ...

celian-garcia (approved):
I can't wait to try the SkipTLS :x. I guess I will be the one having to deal with the conflcts hehe they seem to be easy though

PR #1378, opened 2023-08-24
[ENHANCEMENT] Hide and encrypt sensitive data coming from the datasource spec

Nexucis:
This PR is introducing the notion of public secret where the data is hidden when returned by the API to the user.
It also contains a way to encrypt and decrypt the sensitive data when it's necessary


AntoineThebaud (approved):
LGTM

PR #1379, opened 2023-08-24
[FEATURE] Add builtin variables ($__interval, $interval_ms and $__rate_interval)

### LINKED ISSUE #1370: 1:244: parse error: missing unit character in duration
### What did you do?

I imported the dashboard Node Exporter Full coming from Grafana 

No error

image

### System information

_No response_

### Perses version

_No response_

### Perses configuration file

_No response_

### Logs

_No response_

Gladorme:
# Description

- Add builtin variables:
  - `$__interval` & `$__interval_ms`: Interval that can be used to group by time in queries. When there are more data points than can be shown on a graph then queries can be made more efficient by grouping by a larger interval. RESERVED TO PROMETHEUS QUERIES ONLY!
  - `$__rate_interval`: Interval at least four times the value of the scrape interval, it avoid problems specific to Prometheus when using 'rate' and 'increase' functions. RESERVED TO PROMETHEUS QUERIES ONLY!
- Add millisecond support to some time utils
- Add scrape interval support to Prometheus Datasources
- Add min step on the UI for Prometheus queries

fixes #1370 

# Screenshots

- New `Scrape Interval` parameter in the datasource form when type=PrometheusDatasource (+ some reorg of the fields)
image
- New `Min Step` parameter in panel's query editor 
image
- New section in the variables drawer to display the Prometheus built-in variables
image

# Checklist


ui/plugin-system/src/runtime/plugin-registry.ts
```diff
@@ -82,3 +83,18 @@ export function useListPluginMetadata(pluginType: PluginType, options?: UseListP
...
+export async function usePluginBuiltinVariableDefinitions() {
+  const { getPlugin, listPluginMetadata } = usePluginRegistry();
+
+  const datasources = await listPluginMetadata('Datasource');
+  const datasourceKinds = new Set(datasources.map((datasource) => datasource.kind));
+  let result: BuiltinVariableDefinition[] = [];
+  for (const kind of datasourceKinds) {
+    const plugin = await getPlugin('Datasource', kind);
+    if (plugin.getBuiltinVariableDefinitions) {
+      result = result.concat(plugin.getBuiltinVariableDefinitions());
```
celian-garcia, line 96: Maybe we'll want to keep somewhere the information of from which datasource plugin the builtin variable was coming from. 

To do that you will certainly have to transform ``result`` to a list of list instead of just a list

Gladorme: I plan to add it in the BuiltinVar spec

ui/plugin-system/src/runtime/plugin-registry.ts
```diff
@@ -82,3 +83,18 @@ export function useListPluginMetadata(pluginType: PluginType, options?: UseListP
   const { listPluginMetadata } = usePluginRegistry();
   return useQuery(['listPluginMetadata', pluginType], () => listPluginMetadata(pluginType), options);
 }
+
+export async function usePluginBuiltinVariableDefinitions() {
```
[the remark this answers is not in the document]
Gladorme: For me, it's related to the plugin system, so it should be here

ui/dashboards/src/components/Variables/BuiltinVariableAccordions.tsx
```diff
@@ -0,0 +1,109 @@
...
+type BuiltinVariableAccordionsProps = {
+  builtinVariableDefinitions: BuiltinVariableDefinition[];
+};
+
+export function BuiltinVariableAccordions({ builtinVariableDefinitions }: BuiltinVariableAccordionsProps) {
+  const builtinVariablesBySource = useMemo(() => {
+    const result: Record<string, BuiltinVariableDefinition[]> = {};
+    for (const definition of builtinVariableDefinitions) {
+      const value = result[definition.spec.source];
+      if (value) {
```
nicolastakashi, line 40: Nothing special but I guess it might reduce the cyclomatic complexity a little.

```suggestion
      if (value) {
          value.push(definition);
          continue;
      }
      result[definition.spec.source] = [definition];
```

ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor/PrometheusDatasourceEditor.tsx
```diff
@@ -64,6 +64,19 @@ export function PrometheusDatasourceEditor(props: PrometheusDatasourceEditorProp
...
             onChange={(e) => onChange({ ...value, direct_url: e.target.value })}
           />
+          <Typography py={2} variant="h4">
+            Prometheus Settings
+          </Typography>
+          <TextField
+            fullWidth
+            label="Scrape Interval"
+            value={value.scrape_interval || ''}
+            placeholder="Default: 15s"
```
nicolastakashi, line 74: I'm wondering if we should use Prometheus default scrape interval which is 1m according to their documentation.
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config:~:text=%5B%20scrape_interval%3A%20%3Cduration%3E%20%7C%20default%20%3D%201m%20%5D

Gladorme: I used the same default scrape interval of Grafana. I didn't check Prometheus, it makes sense to use the same as Prometheus

AntoineThebaud (changes requested):
Some `$__rate_interval` computing issues when the datasource scrape interval is set (here = 60s) but no value set for the query's min step, e.g:

2023-09-01_11h58_22

Gladorme:
> Some `$__rate_interval` computing issues when the datasource scrape interval is set (here = 60s) but no value set for the query's min step, e.g:
> 
> 2023-09-01_11h58_22

Fixed, thank you for testing! 😄

AntoineThebaud (approved):

PR #1408, opened 2023-09-01
[IGNORE] bump golangci-lint version

Nexucis:


celian-garcia (approved):

internal/cli/cmd/describe/describe.go
```diff
@@ -17,7 +17,7 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/perses/perses/internal/cli/cmd"
+	persesCMD "github.com/perses/perses/internal/cli/cmd"
```
nicolastakashi, line 20: just keep consistency, should it be `persescmd`?

Nexucis: actually I don't have a strong opinion around that. I change the name of the package so the linter `revive` is happy with it. But I found easier to read when it is following the camelCase convention don't you think ?

PR #1414, opened 2023-09-04
[FEATURE] Display name is optional for variables

### LINKED ISSUE #1384: Cannot set description if display name is not set
### What did you do?

- Create a variable
- Set a description but not display name
- Save + save dashboard

No error

An error

### System information

_No response_

### Perses version

_No response_

### Perses configuration file

_No response_

### Logs

_No response_

Gladorme:
# Description
You can't set a description without having a display name for variables & datasources. This requirement is removed with this PR. 
Closes #1384

# Screenshots

# Checklist

### DIFF
--- a/pkg/model/api/v1/common/display.go
+++ b/pkg/model/api/v1/common/display.go
@@ -13,45 +13,7 @@
 
 package common
 
-import (
-	"encoding/json"
-	"fmt"
-)
-
 type Display struct {
-	Name        string `json:"name" yaml:"name"`
+	Name        string `json:"name,omitempty" yaml:"name,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/model/api/v1/common/display.go
```diff
@@ -29,9 +28,6 @@ func (d *Display) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, (*plain)(&tmp)); err != nil {
```
Gladorme, line 28: Should we keep this function if there no more validation? 🤔

AntoineThebaud (approved):
LGTM!

PR #1150, opened 2023-05-04
[BUGFIX] Remove palette edit buttons after merge issue

sjcobb:
This was meant to be removed in #1124 (I messed up a rebase at some point 😞 since it was deleted at one point)

Mohit had asked for visual editing of palette to be removed, it also doesn't 100% work now that we automatically change the palette based on total series.

PR #1421, opened 2023-09-08
[BREAKINGCHANGE] rename palette's `kind` + change its value to kebab-case

AntoineThebaud:
# Description

As suggested by @sjcobb in https://github.com/perses/perses/pull/1278, this change is to get rid of an inconsistency in our datamodel, here with palette having a `kind` without a `spec`.
At first I changed `palette` to be a simple string with the value of the current `kind` attribute, but in the end I kept it as a struct and changed `kind` to `mode` as there are plans to add more attributes there in the future - for custom colors etc, see current todos.
Also since it's no longer named `kind`, the value is changed to kebab-case.

_Afaiu this palette attribute has currently no use since https://github.com/perses/perses/pull/1150, but kept in the model for later use._

# Checklist

### DIFF
--- a/schemas/panels/time-series/time-series.cue
+++ b/schemas/panels/time-series/time-series.cue
@@ -31,7 +31,8 @@ import (
 }
 
 #palette: {
-	kind: "Auto" | "Categorical"
+	mode: "auto" | "categorical"
+	// colors: [...string]; // TODO: add colors to override ECharts theme
 }
 
 #visual: {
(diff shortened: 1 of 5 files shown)


celian-garcia (commented):
I like `type`. I found also in Grafana they use `mode` which is not bad either. 
In material, the word mode is used to differentiate dark theme from light theme

celian-garcia (approved):
LGTM :)

PR #1424, opened 2023-09-11
[BUGFIX] Fix the way to get an unique document. It is now case sensitive

### LINKED ISSUE #1420: Project API requests are not case sensitive
### What did you do?

Create projects: "perses" and add dashboard in the project. Migrate a dashboard and set "Perses" as project name.

Only see "perses" dashboards when requesting API `/api/v1/projects/perses/dashboards`

"perses" and "Perses" dashboards when requesting API `/api/v1/projects/perses/dashboards`

### System information

_No response_

### Perses version

_No response_

### Perses configuration file

_No response_

### Logs

_No response_

Nexucis:
This is PR fixed #1420.

Like that you will need the accurate case to get your document. Otherwise it will return nothing.


Gladorme (approved):
❤️

Issue #1364, opened 2023-08-16
Improve migration page with auto complete inputs


Gladorme:
### Proposal

Add autocomplete inputs on the migration page, in order to help users to fill inputs (project name & datasource name inputs)

AntoineThebaud:
#1425 brought project name completion & #3373 brought an option to remove any previously-defined datasource selector, so that it falls back to the default datasource. I guess with these two this is enough to close this issue. Further improvements could still be proposed later on.

PR #1425, opened 2023-09-12
[ENHANCEMENT] Autocomplete project name in the migration page

Nexucis:
I'm adding an autocompletion for the project name in the migration page. It's one of the two things needed for #1364

### DIFF
--- a/ui/app/src/views/MigrateView.tsx
+++ b/ui/app/src/views/MigrateView.tsx
@@ -13,6 +13,7 @@
 
 import {
   Alert,
+  Autocomplete,
   Box,
   Button,
   CircularProgress,
@@ -27,11 +28,12 @@ import AutoFix from 'mdi-material-ui/AutoFix';
 import Upload from 'mdi-material-ui/Upload';
 import Import from 'mdi-material-ui/Import';
 import { ChangeEvent, useState } from 'react';
-import { JSONEditor } from '@perses-dev/components';
+import { JSONEditor, useSnackbar } from '@perses-dev/components';
 import { useNavigate } from 'react-router-dom';
 import { useMigrate } from '../model/migrate-client';
 import { useCreateDashboardMutation } from '../model/dashboard-client';
 import { useIsReadonly } from '../model/config-client';
+import { useProjectList } from '../model/project-client';
 
 interface GrafanaLightDashboard {
   // The only part that is interesting us is the list of the input that can exists in the Grafana dashboard definition.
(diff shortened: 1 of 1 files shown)


ui/app/src/views/MigrateView.tsx
```diff
@@ -148,18 +152,27 @@ function MigrateView() {
             {migrateMutation.error.message}
           </Alert>
         )}
-        {migrateMutation.isSuccess && (
-          <Stack direction={'row'} spacing={1}>
+        {!isLoading && data !== undefined && data !== null && migrateMutation.isSuccess && (
```
Gladorme, line 155: `{!isLoading && data && migrateMutation.isSuccess && (`

ui/app/src/views/MigrateView.tsx
```diff
@@ -53,6 +55,8 @@ function MigrateView() {
   const dashboardMutation = useCreateDashboardMutation((data) => {
     navigate(`/projects/${data.metadata.project}/dashboards/${data.metadata.name}`);
   });
+  const { exceptionSnackbar } = useSnackbar();
+  const { data, isLoading } = useProjectList({ onError: exceptionSnackbar });
```
AntoineThebaud, line 59: wouldn't "projects" be a better name than "data" here?

AntoineThebaud (approved):
LGTM, just dropped a suggestion for var naming

ui/app/src/views/MigrateView.tsx
```diff
@@ -148,18 +152,27 @@ function MigrateView() {
             {migrateMutation.error.message}
           </Alert>
         )}
-        {migrateMutation.isSuccess && (
-          <Stack direction={'row'} spacing={1}>
+        {!isLoading && data !== undefined && data !== null && migrateMutation.isSuccess && (
+          <Stack direction={'row'}>
```
Gladorme, line 156: The best practice is to remove curly brackets when it's only containing a hardocded string 😄 Easier to read too

ui/app/src/views/MigrateView.tsx
Gladorme: data is the default var returned by the useQuery function

ui/app/src/views/MigrateView.tsx
Nexucis: I am following the [Perses UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md#additional-best-practices) here: 

> Use explicit `null` or `undefined` checks instead of relying on JS falsiness

ui/app/src/views/MigrateView.tsx
Nexucis: yeah definitely, I'm just wondering if it's possible because I think there is a sort of binding so the name should be the same than the one returned by `useProjectList`

AntoineThebaud: ok my react-noobie fault here, 'thought it was possible to give any name you want here like for useState

Gladorme: You can create an alias if you want `const { data: projects, isLoading }` 😄

PR #1426, opened 2023-09-12
[BUGFIX] dedicated route for dashboard creation to fix wrong redirections

### LINKED ISSUE #1393: Redirect + Document not found error after succesful dashboard creation when the name contains a space
### What did you do?

- go to project view
- click on Add Dashboard
- Provide a name that contains a space
- Once on the dashboard view, add 1 panel
- Click on save

Stay on the dashboard view (as when no space is provided in the name)

You get redirected to the project page, both document not found error snackbar & dashboard successfully created info snackbar are displayed. The dashboard is available & got created with an underscore instead of the space.

### Perses version

0.39

AntoineThebaud:
# Description

This PR changes:

- the dashboard creation route from `/projects/<project>/dashboards/create` to `/projects/<project>/dashboard/new`. 
  - This fixes a redirection issue we had with the dashboard creation page, i.e the `navigate()` call to standard dashboard view was ignored and the page was instead reloading on `/create`. This was the root cause behind #1393 & also a minor issue not detected previously: in case of successful dashboard creation, reloading the page manually was causing the display of a warning message "document already exists".
  - _NB: side effect of this change: currently when you try to create a dashboard and provide a name that already exists, you get redirected to the existing dashboard instead with a warning "document already exists". Now with this PR, you land on the new dashboard view & will get "document already exists" error if you try to save. To solve the conflict, the only solution currently is to edit the dashboard JSON directly (hence the added condition around warning display on the JSON dialog, since metadata update is possible when creating a new dashboard). This is something we can probably improve in the future with a proper field for name edition. EDIT a simpler solution would be to have some backend validation/check upfront, when providing the initial name in the dashboard creation dialog._

- routes to follow same naming convention & lazy-loading for all (still except the Home page).

# Checklist

## UI Changes


ui/app/src/views/ViewDashboard.tsx
```diff
@@ -128,7 +128,6 @@ function ViewDashboard() {
           onSuccess: (createdDashboard: DashboardResource) => {
             actionRef.current = undefined;
             successSnackbar(`Dashboard ${getDashboardDisplayName(createdDashboard)} has been successfully created`);
-            navigate(`/projects/${createdDashboard.metadata.project}/dashboards/${createdDashboard.metadata.name}`);
```
AntoineThebaud, line 131: fyi this was basically never called; when `navigate()` call is placed here in this promise, the auto redirection caused by the form submit is taking precedence.

ui/dashboards/src/components/SaveDashboardButton/SaveDashboardButton.tsx
```diff
@@ -67,6 +69,8 @@ export const SaveDashboardButton = ({ onSave, isDisabled, variant = 'contained'
     } else {
       saveDashboard();
     }
+
+    navigate(`/projects/${dashboard.metadata.project}/dashboards/${dashboard.metadata.name}`);
```
Nexucis, line 73: actually we shouldn't do that directly in the dashboards package since you are taking assumption about how the routing is done.

This should come from the parent, that's why initially it was done in the callback of the method `saveDashboard`

ui/app/src/views/projects/dashboards/GenericDashboardView.tsx
```diff
@@ -0,0 +1,104 @@
...
+  onDiscard?: (entity: DashboardResource) => void;
+  isReadonly: boolean;
+  isEditing: boolean;
+  isCreating?: boolean;
+}
+
+/**
+ * The View for displaying a Dashboard.
+ */
+export function GenericDashboardView(props: GenericDashboardViewProps) {
```
AntoineThebaud, line 39: Ugly name I know, if you have a better idea feel free to propose :D

ui/app/src/views/projects/dashboards/DashboardView.tsx
```diff
@@ -0,0 +1,86 @@
...
+        onError: (err) => {
+          exceptionSnackbar(err);
+          throw err;
+        },
+      });
+    },
+    [exceptionSnackbar, successSnackbar, updateDashboardMutation]
+  );
+
+  const handleDashboardDiscard = useCallback(() => {
```
Gladorme, line 63: callback can be remove `() => undefined`

ui/app/src/views/projects/dashboards/CreateDashboardView.tsx
```diff
@@ -0,0 +1,110 @@
...
+  if (!data || data.spec === undefined) return null;
+
+  return (
+    <GenericDashboardView
+      dashboardResource={data}
+      onSave={handleDashboardSave}
+      onDiscard={handleDashboardDiscard}
+      isReadonly={false}
+      isEditing={true}
+      isCreating={true}
```
AntoineThebaud, line 105: In order to display conditionnally the warning message on the Edit JSON dialog (I want it in case of an existing dashboard but not for a new one), I'm obliged to add this new prop `isCreating` at this stage and thus to pass it down repetitively to GenerericDashboardView > ViewDashboard > DashboardApp > EditJsonDialog > EditJsonDialogForm. That's quite a long road but I dont think putting in place a context/provider here would be worth it

ui/app/src/views/projects/dashboards/DashboardView.tsx
Gladorme: Maybe put onDiscard optional if it's not used here

ui/app/src/views/projects/dashboards/GenericDashboardView.tsx
```diff
@@ -0,0 +1,104 @@
...
+    datasourceApi.listDatasources(dashboardResource.metadata.project);
+    datasourceApi.listGlobalDatasources();
+  }, [datasourceApi, dashboardResource]);
+
+  // Collect the Project variables and setup external variables from it
+  const { data: globalVars, isLoading: isLoadingGlobalVars } = useGlobalVariableList();
+  const { data: projectVars, isLoading: isLoadingProjectVars } = useVariableList(dashboardResource.metadata.project);
+  const externalVariableDefinitions: ExternalVariableDefinition[] | undefined = useMemo(
+    () => [
+      buildProjectVariableDefinition(dashboardResource.metadata.project, projectVars || []),
```
Gladorme, line 54: It's better to use `??` instead of `||`

AntoineThebaud: ah ok, this comes from the existing code but I can change it 👍

ui/dashboards/src/components/EditJsonDialog/EditJsonDialog.tsx
```diff
@@ -18,18 +18,24 @@ import { useTimeRange } from '@perses-dev/plugin-system';
 import { useEditJsonDialog } from '../../context/DashboardProvider';
 import { useDashboard } from '../../context/useDashboard';
 
-export const EditJsonDialog = () => {
+export interface EditJsonDialogProps {
+  isCreating?: boolean;
```
Gladorme, line 22: I think we could use a more generic term for the component: enableMetadataEdition or disableMetadataEdition

Gladorme: Because isCreating is related to the user flow, not really the component

ui/dashboards/src/components/EditJsonDialog/EditJsonDialog.tsx
```diff
@@ -18,18 +18,24 @@ import { useTimeRange } from '@perses-dev/plugin-system';
...
+export const EditJsonDialog = (props: EditJsonDialogProps) => {
+  const { isCreating } = props;
   const { editJsonDialog, closeEditJsonDialog } = useEditJsonDialog();
 
   return (
     <Dialog open={!!editJsonDialog?.isOpen} scroll="paper" fullWidth maxWidth="lg">
-      <Dialog.Header onClose={() => closeEditJsonDialog()}>Edit Dashboard</Dialog.Header>
-      {editJsonDialog?.isOpen && <EditJsonDialogForm />}
+      <Dialog.Header onClose={() => closeEditJsonDialog()}>Edit Dashboard JSON</Dialog.Header>
+      {editJsonDialog?.isOpen && <EditJsonDialogForm isCreating={isCreating} />}
```
Gladorme, line 32: Same for the form "isCreating"

ui/app/src/views/projects/dashboards/DashboardView.tsx
```diff
@@ -0,0 +1,86 @@
...
+    // do nothing
+  }, []);
+
+  if (isLoading) return null;
```
Gladorme, line 67: We should but a loader, no? Something like this
``` tsx
{isLoading && (
              <CircularProgress />
            </Stack>
          )}
```

ui/app/src/views/projects/dashboards/GenericDashboardView.tsx
Nexucis: what about WrapperDashboardView or HelperDashboardView ?

Gladorme (approved):
LGTM 😄

ui/app/src/views/projects/dashboards/GenericDashboardView.tsx
AntoineThebaud: Yep that's slightly better I think, I'll go for Helper

Nexucis (approved):
I added a suggestion to replace the `GenericDashboardView` but otherall looks pretty good.
Well done !

PR #1427, opened 2023-09-12
[ENHANCEMENT] Autocomplete variable list

### LINKED ISSUE #1381: Filter variable list selection with a search field
### Proposal

Would be good to be able to filter a bit the selection that we are doing in a variable typing some prefix or something. It would help to reduce the big lists.

Suggestion? 
https://mui.com/material-ui/react-autocomplete/#combo-box

Nexucis:
This PR is modifying the rendering of the variable. Like that we can search across the list.

There is still place for improvement in this area: 
1. When the label of variable is too big, we cannot see the complete value.
2. On multiple select option, we could have a checkbox

Note: this PR fix #1381

https://github.com/perses/perses/assets/4548045/6a5cd04d-53a4-41b9-8622-9a9fa20ff808

### DIFF
--- a/ui/dashboards/src/components/Variables/TemplateVariable.tsx
+++ b/ui/dashboards/src/components/Variables/TemplateVariable.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


AntoineThebaud:
Can you please add a screenshot or recording 🤓

AntoineThebaud:
New PR opened https://github.com/perses/perses/pull/1576

PR #1441, opened 2023-09-21
[BUGFIX] Fix memory leak caused by a cue context kept during the load of the schemas

### LINKED ISSUE #1440: Constantly increasing memory usage
### What did you do?

I have setup perses via the docker image (per described in the readme), and connected it to my prometheus instance.  I have a few dashboards and widgets (including one of the attached image).

Memory should remain relatively flat or exhibit GC-collection saw-tooth form.

Perses memory consistently increases over time, even when not being used.  Eventually the docker image crashes (when it hits its memory limit, which I have set at 128 MB).

image

The green line reset sooner because I switched it from 0.39 to 0.40 to see if that would help.

### System information

linux amd64 (docker)

### Perses version

0.39 and 0.40

### Perses configuration file

```yaml
Default for the docker file
```

### Logs

_No response_

Nexucis:
This PR fixed #1440 

We are loading every 5min the cuelang schema. And we never reset the cue schema during this reload. The cue context is tracking every loaded instances of the schema, which made the memory increasing during every loop.

Even if we don't want to keep a cue context forever, we still need to keep one that would have in memory the different schema loaded. The cue context is then used during the validation phase.

It also removed the usage of the sync map. 
Explanations are described here: https://teivah.medium.com/maps-and-memory-leaks-in-go-a85ebe6e7e69

### DIFF
--- a/internal/api/shared/migrate/loader.go
+++ b/internal/api/shared/migrate/loader.go
@@ -18,7 +18,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
+	"sync/atomic"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/perses/common/async"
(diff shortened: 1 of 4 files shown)


celian-garcia (approved):
Nice to see this atomic.Pointer in action! Good work for investigation also. It seems it was not simple to figure it out.

PR #1459, opened 2023-10-03
[BUGFIX] Prom DS: change to XOR between `directUrl` and `proxy`

AntoineThebaud:
# Description

This PR brings changes to the PrometheusDatasource cuelang schema:
- XOR condition to ensure either `directUrl` or `proxy` is present. It's not supposed to be able to provide either both or none of them.
- format validation for the url fields.

_NB: this is the error you'd now see if you try to create a datasource without direct URL or proxy config set:_
image
_So the cue validation is working as expected but the error message is not very intelligible. Imho this is not a problem since we also want field validation on the frontend side, which would prevent submitting the form with no or wrongly-formatted URL provided. The latter is not tackled by this PR but instead should come with https://github.com/perses/perses/pull/1438._

ℹ️ This change is technically a breaking change, however I don't expect it to have a real impact at this stage, + on the frontend side it was already a XOR (the radio buttons force you to chose between the 2 configs).
-> Because of this FE/BE misalignment, in the end we decided to categorize this PR as bugfix.

# Checklist

### DIFF
--- a/schemas/common/proxy/http.cue
+++ b/schemas/common/proxy/http.cue
@@ -13,6 +13,10 @@
 
 package proxy
 
+import (
+	"github.com/perses/perses/schemas/common"
+)
+
 #HTTPAllowedEndpoint: {
 	endpointPattern: string
 	method:          "POST" | "PUT" | "PATCH" | "GET" | "DELETE"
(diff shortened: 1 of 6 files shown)


AntoineThebaud:
@Nexucis this PR is at a very early stage but just want to confirm we're align on its purpose here: Do we agree that prom datasource should either come with a directUrl or proxy config, and that you're not supposed to provide either both or none of them?

celian-garcia (approved):
LGTM.
Although the direct url vs proxy url difference is not straightforward in my opinion, it makes sense to not use both. The UI itself will anyway use one or the other and not both so yes.

nicolastakashi (approved):

Nexucis (approved):
I would probably call it a bugfix rather than a breaking change

PR #1467, opened 2023-10-12
[IGNORE] Complete doc about the config

Nexucis:
I'm just completing the documentation about the config. Even if the explanations are not exhaustive, it will be nicer than having to look at the code to know exactly what to put in the yaml config file

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -63,29 +63,30 @@ Generic placeholders are defined as follows:
 
 ```yaml
   # A flag that will disable any HTTP POST, PUT and DELETE endpoint in the API.
-  # It will also change the UI to reflect th
+  # It will also change the UI to reflect this config, by removing any action button and will prevent the access to a form.
   [ readonly: <boolean> | default = false ]
 
   # The secret key used to encrypt and decrypt sensitive data stored in the database such as any data in the Secret and GlobalSecret object.
-  # Note that if it is not provided it will be generated. When perses is used in a multi instance mode, you should provide the key.
-  # Otherwise, each instance will have a different key and therefore won't be able to decrypt what the other is encrypting.
-  # Also note the key must be at least 32 bytes long.
+  # Note that if it is not provided it will be generated. 
+  # When perses is used in a multi instance mode, you should provide the key, otherwise, each instance will have a different key 
+  # and therefore won't be able to decrypt what the other is encrypting.
+  # Also note that the key must be at least 32 bytes long.
   [ encryption_key: <secret> ]
-    
+
   # The path to the file containing the secret key.
   [ encryption_key_file: <filename> ]
 
   # Database configuration 
   database: <database_config>
-  
-  # The configuration to get access to the CUE schemas
+
+  # The configuration to access the CUE schemas
   [ schemas: <schemas_config> ]
-  
-  # A list of dashboard you would like to display in the UI home page
-  important_dashboards: 
-  - [ <dashboard_selector_config> ]
-  
-  # The markdown content to be display on the UI home page
+
+  # A list of dashboards you would like to display in the UI home page
+  important_dashboards:
+    - [ <dashboard_selector_config> ]
+
+  # The markdown content to be displayed on the UI home page
   [ information: <string> ]
 ```
 
@@ -93,7 +94,7 @@ Generic placeholders are defined as follows:
 
 ```yaml
   # Config in case you want to use a file DB.
-  # Prefer the SQL config in case you are running multiple Perses instance.
+  # Prefer the SQL config in case you are running multiple Perses instances.
   [ file: <database_file_config> ]
   
   # The SQL config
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+  # Also note the key must be at least 32 bytes long.
+  [ encryption_key: <secret> ]
+    
+  # The path to the file containing the secret key.
+  [ encryption_key_file: <filename> ]
+
+  # Database configuration 
+  database: <database_config>
+  
+  # The configuration to get access to the CUE schemas
```
AntoineThebaud, line 81: The configuration to access the CUE schemas

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+* `<secret>`: a regular string that is a secret, such as a password
+* `<string>`: a regular string
+
+```yaml
+  # A flag that will disable any HTTP POST, PUT and DELETE endpoint in the API.
+  # It will also change the UI to reflect th
+  [ readonly: <boolean> | default = false ]
+
+  # The secret key used to encrypt and decrypt sensitive data stored in the database such as any data in the Secret and GlobalSecret object.
+  # Note that if it is not provided it will be generated. When perses is used in a multi instance mode, you should provide the key.
```
AntoineThebaud, line 70: [...] the key, otherwise [...]

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+* `<duration>`: a duration matching the regular expression `((([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?|0)`, e.g. `1d`, `1h30m`, `5m`, `10s`
+* `<filename>`: a valid path in the current working directory
+* `<path>`: a valid URL path
+* `<int>`: an integer value
+* `<secret>`: a regular string that is a secret, such as a password
+* `<string>`: a regular string
+
+```yaml
+  # A flag that will disable any HTTP POST, PUT and DELETE endpoint in the API.
+  # It will also change the UI to reflect th
```
AntoineThebaud, line 66: missing end of sentence

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+  # The path to the file containing the secret key.
+  [ encryption_key_file: <filename> ]
+
+  # Database configuration 
+  database: <database_config>
+  
+  # The configuration to get access to the CUE schemas
+  [ schemas: <schemas_config> ]
+  
+  # A list of dashboard you would like to display in the UI home page
```
AntoineThebaud, line 84: dashboard**s**

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+  important_dashboards: 
+  - [ <dashboard_selector_config> ]
+  
+  # The markdown content to be display on the UI home page
```
AntoineThebaud, line 88: display**ed**

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+
+```yaml
+  # A flag that will disable any HTTP POST, PUT and DELETE endpoint in the API.
+  # It will also change the UI to reflect th
+  [ readonly: <boolean> | default = false ]
+
+  # The secret key used to encrypt and decrypt sensitive data stored in the database such as any data in the Secret and GlobalSecret object.
+  # Note that if it is not provided it will be generated. When perses is used in a multi instance mode, you should provide the key.
+  # Otherwise, each instance will have a different key and therefore won't be able to decrypt what the other is encrypting.
+  # Also note the key must be at least 32 bytes long.
```
AntoineThebaud, line 72: note that

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+  # Path to the Cue schemas of the queries
+  [ queries_path: <path> | default = "schemas/queries" ]
+
+  # Path to the Cue schemas of the datasources
+  [ datasources_path: <path> | default = "schemas/datasources" ]
+
+  # Path to the Cue schemas of the variables
+  [ variables_path: <path> | default = "schemas/variables" ]
+
+  # The refresh interval of the cue schema regardless their localisation
```
Gladorme, line 205: localisation => path. I think we should keep the same wording 😄

AntoineThebaud, line 205: schema**s**

docs/configuration.md
```diff
@@ -41,6 +43,211 @@ uppercase. Every environment variable for this config are prefixed by `PERSES`
...
+  
+  # The markdown content to be display on the UI home page
+  [ information: <string> ]
+```
+
+### `<database_config>`
+
+```yaml
+  # Config in case you want to use a file DB.
+  # Prefer the SQL config in case you are running multiple Perses instance.
```
Gladorme, line 96: instances

PR #1470, opened 2023-10-12
[BUGFIX] Fix too-restrictive regex used for URL validation

AntoineThebaud:
# Description

- Previous regex was not matching for valid urls like e.g `https://thanos-query-global.mycompany.net`.

- I also took the opportunity with this PR to (finally) improve the error feedback returned to the user when CUE validation fails. E.g when initially facing this URL issue, the error displayed was:
`bad request: invalid datasource : spec: 2 errors in empty disjunction: (and 2 more errors)`
Now for the same you'd get:
`bad request: invalid datasource NewDatasource: spec: 2 errors in empty disjunction: spec.proxy: field not allowed: 1:1 1:40 ..\schemas\datasources\prometheus\prometheus.cue:22:7 ..\schemas\datasources\prometheus\prometheus.cue:23:2 ..\schemas\datasources\prometheus\prometheus.cue:27:13 spec.proxy.spec.url: invalid value "www.google.com" (out of bound =~"^https?://[^\\s/$.?#].[^\\s]*$"): ..\schemas\common\url.cue:16:7 1:261 ..\schemas\common\proxy\http.cue:30:8`
..which we could still improve by doing some cleanup for better readability. Still it leads you to the real root cause contrary to previously.

# Checklist


schemas/common/url.cue
```diff
@@ -13,4 +13,4 @@
 
 package common
 
-#url: =~"^https?:\\/\\/[a-z]+(?:\\.[a-z]+)*(:\\d+)?(?:\\/[a-z\\d]+)*$"
+#url: =~"^https?:\/\/[^\\s\/$.?#].[^\\s]*$"
```
celian-garcia, line 16: it looks like the previous regex was escaping twice the / with \\/ but now it's \/ is that change wanted?

pkg/model/api/v1/datasource.go
```diff
@@ -36,7 +36,7 @@ func FilterDatasource[T DatasourceInterface](kind string, defaultDTS *bool, list
 
 type DatasourceInterface interface {
 	GetMetadata() modelAPI.Metadata
-	GetDTSSpec() DatasourceSpec
+	GetDatasourceSpec() DatasourceSpec
```
celian-garcia, line 39: or GetSpec ? :smile:

AntoineThebaud: ah indeed, better :D

AntoineThebaud: Oh actually not possible as-is, theres's already a GetSpec

Nexucis (approved):

PR #1471, opened 2023-10-12
[IGNORE] Add doc about secret

Nexucis:
I'm starting a documentation about the `Secret` and the `GlobalSecret`.

### DIFF
--- a/docs/secret.md
+++ b/docs/secret.md
@@ -4,12 +4,10 @@ When defining a datasource, you will probably need to provide a basic authentica
 used when the Perses backend will contact your Datasource.
 
 We have two different objects to store sensitive data: `GlobalSecret` and `Secret`.
-To know what which one you need, it will depend on what object representing a datasource you are using.
+You should use one or the other depending on which object your datasource corresponds to.
 
-If you want to store sensitive data for a `GlobalDatasource`, then you need to create a `GlobalSecret`.
-For a `Datasource`, you need to create a `Secret`, `GlobalSecret` cannot be used in a `Datasource`.
-Finally, for the datasource defined directly in a dashboard, you can use a `Secret` as it is defined in the same project
-as the dashboard.
+- To store sensitive data for a `GlobalDatasource`, you need to create a `GlobalSecret`.
+- For a `Datasource` object or a datasource defined directly in a dashboard, you need to create a `Secret`. `GlobalSecret` cannot be used here.
 
 A `Secret` is defined like that:
 
--- a/docs/secret.md
+++ b/docs/secret.md
@@ -1,0 +1,171 @@
+# Secret
+
+When you are defining a datasource, likely you will need to provide a basic authentication, a certificate, or a token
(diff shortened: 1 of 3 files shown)


docs/secret.md
```diff
@@ -0,0 +1,171 @@
+# Secret
+
+When you are defining a datasource, likely you will need to provide a basic authentication, a certificate, or a token
```
AntoineThebaud, line 3:
```suggestion
When defining a datasource, you will probably need to provide a basic authentication, a certificate, or a token
```

docs/secret.md
```diff
@@ -0,0 +1,171 @@
...
+to be used when the Perses backend will contact your Datasource.
+
+`Secret` or `GlobalSecret` are the two objects that you need. The choice between a `Secret` or
+a `GlobalSecret` will mainly depend on if you are defining a `Datasource` or a `GlobalDatasource`.
```
AntoineThebaud, line 7: Maybe we should highlight more the difference in terms of "scope behavior" with the rest of the resources here?
I mean e.g a project variable can rely on a global datasource for instance, while here we have:
- GlobalDatasource -> GlobalSecret
- (Project) Datasource -> Secret only
- Dashboard datasource -> didnt check but Secret only too I guess ?

Maybe it's an issue to have such discrepancy btw, not sure but anyway that's out of the topic of this PR.

WDYT?

Nexucis: ah yeah indeed it's certainly unclear like that. I will change that

docs/secret.md
```diff
@@ -1,10 +1,15 @@
...
 
-When you are defining a datasource, likely you will need to provide a basic authentication, a certificate, or a token
-to be used when the Perses backend will contact your Datasource.
+When defining a datasource, you will probably need to provide a basic authentication, a certificate, or a token to be
+used when the Perses backend will contact your Datasource.
 
-`Secret` or `GlobalSecret` are the two objects that you need. The choice between a `Secret` or
-a `GlobalSecret` will mainly depend on if you are defining a `Datasource` or a `GlobalDatasource`.
+We have two different objects to store sensitive data: `GlobalSecret` and `Secret`.
+To know what which one you need, it will depend on what object representing a datasource you are using.
```
AntoineThebaud, line 7:
```suggestion
You should use one or the other depending on which object your datasource corresponds to.
```

docs/secret.md
```diff
@@ -1,10 +1,15 @@
...
+
+If you want to store sensitive data for a `GlobalDatasource`, then you need to create a `GlobalSecret`.
```
AntoineThebaud, line 9:
```suggestion
- To store sensitive data for a `GlobalDatasource`, you need to create a `GlobalSecret`.
- For a `Datasource` object or a datasource defined directly in a dashboard , you need to create a `Secret`. `GlobalSecret` cannot be used here.
```

AntoineThebaud (approved):

PR #1472, opened 2023-10-13
[FEATURE] Add support of user

Nexucis:
This PR is adding the support of a new resource "User".
This should help for the authentication.

### DIFF
--- a/internal/api/impl/v1/user/service.go
+++ b/internal/api/impl/v1/user/service.go
@@ -1,0 +1,123 @@
+// Copyright 2023 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package user
+
+import (
+	"fmt"
+
+	"github.com/perses/perses/internal/api/interface/v1/user"
+	"github.com/perses/perses/internal/api/shared"
+	databaseModel "github.com/perses/perses/internal/api/shared/database/model"
+	"github.com/perses/perses/pkg/model/api"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/bcrypt"
+)
+
+func hashAndSalt(pwd []byte) ([]byte, error) {
+	return bcrypt.GenerateFromPassword(pwd, bcrypt.DefaultCost)
+}
+
+type service struct {
+	user.Service
+	dao user.DAO
+}
+
+func NewService(dao user.DAO) user.Service {
+	return &service{
+		dao: dao,
+	}
+}
+
+func (s *service) Create(entity api.Entity) (interface{}, error) {
+	if userObject, ok := entity.(*v1.User); ok {
+		return s.create(userObject)
+	}
+	return nil, fmt.Errorf("%w: wrong entity format, attempting user format, received '%T'", shared.BadRequestError, entity)
+}
+
+func (s *service) create(entity *v1.User) (*v1.PublicUser, error) {
+	// Update the time contains in the entity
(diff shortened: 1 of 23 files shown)


internal/api/impl/v1/user/service.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, updateErr
+	}
+	return v1.NewPublicUser(entity), nil
+}
+
+func (s *service) Delete(parameters shared.Parameters) error {
+	return s.dao.Delete(parameters.Name)
+}
+
+func (s *service) Get(parameters shared.Parameters) (interface{}, error) {
```
celian-garcia, line 117: The dao return a User. Does it mean that the get will return the password ?

internal/api/impl/v1/user/service.go
```diff
@@ -0,0 +1,123 @@
...
+	return s.dao.Get(parameters.Name)
+}
+
+func (s *service) List(q databaseModel.Query, _ shared.Parameters) (interface{}, error) {
```
celian-garcia, line 121: Same for List will it return the password ?

pkg/client/api/v1/user.go
```diff
@@ -0,0 +1,98 @@
...
+	client *perseshttp.RESTClient
+}
+
+func newUser(client *perseshttp.RESTClient) UserInterface {
+	return &user{
+		client: client,
+	}
+}
+
+func (c *user) Create(entity *v1.User) (*v1.User, error) {
```
celian-garcia, line 49: AFAIU the returns should all be PublicUser no ?

celian-garcia (changes requested):

pkg/model/api/v1/public_user.go
```diff
@@ -0,0 +1,62 @@
...
+
+package v1
+
+import (
+	modelAPI "github.com/perses/perses/pkg/model/api"
+	"github.com/perses/perses/pkg/model/api/v1/secret"
+)
+
+type PublicUserSpec struct {
+	FirstName string        `json:"firstName,omitempty" yaml:"firstName"`
```
Gladorme, line 22: `omitempty` missing for yaml, is it normal or is it a trick for something? 👀 
If it's intended maybe explain a bit in a comment 😄

pkg/model/api/v1/public_user.go
```diff
@@ -0,0 +1,62 @@
...
+
+func NewPublicUserSpec(u UserSpec) PublicUserSpec {
+	return PublicUserSpec{
+		FirstName: u.FirstName,
+		LastName:  u.LastName,
+		Password:  secret.Hidden(u.Password),
+	}
+}
+
+type PublicUser struct {
```
Gladorme, line 35: No yaml?

pkg/model/api/v1/user.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	modelAPI "github.com/perses/perses/pkg/model/api"
+)
+
+type UserSpec struct {
+	FirstName string `json:"firstName,omitempty" yaml:"firstName"`
```
Gladorme, line 24: `omitempty` missing for yaml, is it normal or is it a trick for something? 👀 
If it's intended maybe explain a bit in a comment 😄

pkg/model/api/v1/user.go
```diff
@@ -0,0 +1,83 @@
...
+	LastName  string `json:"lastName,omitempty" yaml:"lastName"`
+	Password  []byte `json:"password,omitempty" yaml:"password"`
+}
+
+func (p *UserSpec) UnmarshalJSON(data []byte) error {
+	type plain = struct {
+		FirstName string `json:"firstName,omitempty"`
```
Gladorme, line 31: No yaml?

internal/api/impl/v1/user/service.go
Nexucis: ah yeah I miss this part. That wasn't intended. I will fix that

pkg/model/api/v1/public_user.go
Nexucis: oups miss that too, will fix it

pkg/client/api/v1/user.go
Nexucis: it does't really matter for the client. `PublicUser` on the API side is just here to change the behaviour of the `Marshalling` when the document is going outside of Perses (database included)

celian-garcia (approved):

PR #1473, opened 2023-10-13
[FEATURE] Add authentication endpoint

Nexucis:
This PR is adding a new endpoint `/api/auth` that should be used to authenticate a user with a login and a password.
When the login is successful it doesn't return anything yet. Later it will return a JWT token


nicolastakashi:
TIL: Perses already have user methods 😄 
LGTM

nicolastakashi (approved):

PR #1475, opened 2023-10-13
[BREAKINGCHANGE] Allow TimeRangeProvider and TemplateVariableProvider to live without QueryParamsProvider

celian-garcia:
# Description

This PR comes from the experience of creating an embedded panel in a separate UI. 

This requires for sure a `TimeRangeProvider`, but even if we set `enabledURLParams` property to `false`, it still requires to define a `QueryParamsProvider` upper, which leads to this kind of non-sense:
```tsx
<BrowserRouter>
  <QueryParamProvider adapter={ReactRouter6Adapter} options={{ params: {} }}>
    <TimeRangeProvider
      initialRefreshInterval="0s"
      initialTimeRange={{ pastDuration: '30m' }}
      enabledURLParams={false}
    >
      <TemplateVariableProvider>
      ...
```

With this PR this PR you can fully decide to use or not query params

- If you don´t want to use them
```tsx
<TimeRangeProvider
  refreshInterval="0s"
  timeRange={{ pastDuration: '30m' }}
>
  <TemplateVariableProvider>
  ...
```
- If you want to use them
```tsx
<BrowserRouter>
  <QueryParamProvider adapter={ReactRouter6Adapter} options={{ params: {} }}>
    <TimeRangeProviderWithQueryParams
      initialRefreshInterval="0s"
      initialTimeRange={{ pastDuration: '30m' }}
    >
      <TemplateVariableProviderWithQueryParams>
  ...
```

# Breaking changes
- The `enabledURLParams` in the `TimeRangeProvider` disappear to let you choose between `TimeRangeProvider` and `TimeRangeProviderWithQueryParams`
- The default behaviour of using query params inside the `TemplateVariableProvider` disappear to let you choose between `TemplateVariableProvider` and `TemplateVariableProviderWithQueryParams`

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
+++ b/ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
@@ -424,6 +424,21 @@ export function TemplateVariableProvider({
   externalVariableDefinitions = [],
   builtinVariables = [],
 }: TemplateVariableProviderProps) {
+  const [store] = useState(createTemplateVariableSrvStore({ initialVariableDefinitions, externalVariableDefinitions }));
+
+  return (
+    <TemplateVariableStoreContext.Provider value={store}>
+      <PluginProvider builtinVariables={builtinVariables}>{children}</PluginProvider>
+    </TemplateVariableStoreContext.Provider>
+  );
+}
+
+export function TemplateVariableProviderWithQueryParams({
+  children,
+  initialVariableDefinitions = [],
+  externalVariableDefinitions = [],
+  builtinVariables = [],
+}: TemplateVariableProviderProps) {
   const allVariableDefs = mergeVariableDefinitions(initialVariableDefinitions, externalVariableDefinitions);
   const queryParams = useVariableQueryParams(allVariableDefs);
   const [store] = useState(
--- a/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
+++ b/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
@@ -11,20 +11,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, { useMemo, useState, useCallback, createContext, useContext } from 'react';
+import React, { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react';
 import {
   AbsoluteTimeRange,
   DurationString,
-  isRelativeTimeRange,
   TimeRangeValue,
+  isRelativeTimeRange,
   toAbsoluteTimeRange,
 } from '@perses-dev/core';
-import { useSetRefreshIntervalParams, useSetTimeRangeParams } from './query-params';
+import { getRefreshIntervalInMs } from './refresh-interval';
 
 export interface TimeRangeProviderProps {
-  initialTimeRange: TimeRangeValue;
-  initialRefreshInterval?: DurationString;
-  enabledURLParams?: boolean;
+  timeRange: TimeRangeValue;
+  refreshInterval?: DurationString;
+  setTimeRange?: (value: TimeRangeValue) => void;
+  setRefreshInterval?: (value: DurationString) => void;
   children?: React.ReactNode;
 }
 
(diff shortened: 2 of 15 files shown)


ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
```diff
@@ -19,7 +19,12 @@ import {
   TimeRangeValue,
   toAbsoluteTimeRange,
 } from '@perses-dev/core';
-import { useSetRefreshIntervalParams, useSetTimeRangeParams } from './query-params';
+import {
+  useSetRefreshIntervalParamsWithQueryParams,
+  useSetRefreshIntervalParams,
+  useSetTimeRangeParamsWithQueryParam,
+  useSetTimeRangeParams,
+} from './query-params';
```
celian-garcia, line 27: Maybe the ones that don´t use query params should not live in a file named query-params

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
```diff
@@ -416,13 +416,43 @@ export interface TemplateVariableProviderProps {
...
   builtinVariables?: BuiltinVariableDefinition[];
+  enabledURLParams?: boolean;
 }
 
 export function TemplateVariableProvider({
   children,
   initialVariableDefinitions = [],
   externalVariableDefinitions = [],
   builtinVariables = [],
+  enabledURLParams = true,
```
Gladorme, line 427: Enabled by default? 🤔 
IMO, we should provide minimum features and user enable whatever he wants

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
```diff
@@ -56,37 +57,104 @@ export function useTimeRange(): TimeRange {
...
+ * gave some initial time range / refresh interval.
+ * @param ctx
+ */
+function useContextMemo(
+  ctx: Omit<TimeRange, 'refreshKey' | 'absoluteTimeRange' | 'refresh' | 'refreshIntervalInMs'>
+): TimeRange {
+  const [refreshCounter, setRefreshCounter] = useState(0);
+
+  const refresh = useCallback(() => {
+    setRefreshCounter(refreshCounter + 1);
```
Gladorme, line 72:
```suggestion
    setRefreshCounter(prevCounter => prevCounter + 1);
```
An you can remove the useCallback wrapping the function

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
```diff
@@ -56,37 +57,104 @@ export function useTimeRange(): TimeRange {
...
+/**
+ * Utils function to transform a refresh interval in {@link DurationString} format into a number of ms.
+ * @param refreshInterval
+ */
+function getRefreshIntervalInMs(refreshInterval?: DurationString): number {
+  let refreshIntervalInMs = 0;
+  if (refreshInterval) {
+    const refreshIntervalDuration = parseDurationString(refreshInterval);
+    if (refreshIntervalDuration && refreshIntervalDuration.seconds) {
+      refreshIntervalInMs = refreshIntervalDuration?.seconds * 1000;
```
Gladorme, line 96:
```suggestion
      refreshIntervalInMs = milliseconds(refreshIntervalDuration);
```
And if condition can be removed

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
celian-garcia: You mean something like this ? 
image


My IDE is not happy with that.
image



TBH I don't want to modify too much the existing because I'm afraid of regression

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
celian-garcia: Where this milliseconds is supposed to come from ??
image

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
Gladorme: Ah yes, never mind about the useCallback removal

ui/dashboards/src/context/TemplateVariableProvider/TemplateVariableProvider.tsx
celian-garcia: Hi @jgbernalp What would you think of these breaking change from redhat? Are you using enabledURLParams ? Do you mind if we make the modifications that I listed in this previous message?

jgbernalp: Hi @celian-garcia, I think the correct way to go will be to decouple these providers. IMO the TimeRangeProvider should not rely on the URL state. That should be a choice when using the component, otherwise it comes to this issues in which the component is not reusable if you happen to have other means to store the range

celian-garcia: Yep absolutely I think we agree on that and that's exactly why I made this PR to make sure that this existing ``enabledURLParams`` boolean will actually ensure that the QueryParamsProvider is not necessary if flase.
Except if you'd have better way to decouple in head?

The breaking change would come with the changes proposed by @Gladorme.
- if we change the existing ``enabledURLParams`` to ``isURLParamsEnabled`` in TimeRangeProvider.. People using ``enabledURLParams`` will have to rename.
- if we make the new ``isURLParamsEnabled`` by default to false in TemplateVariableProvider (the current behaviour is that QueryParamProvider is needful by default). People using currently the TemplateVariableProvider  will have to explicitely set enabeldURLParams to true if they want the same behaviour

jgbernalp: A better way to decouple would be invert control and pass `timeRange` `setTimeRange` to the TimeRangeProvider, then you can choose to use them from the QueryParamsProvider or any other you choose. I think this boolean condition does not decouple the providers as consumers must know that using `false` will allow to use without the QueryProvider otherwise there will be an error.

celian-garcia: Ah yeah but then it becomes complex for the user who actually use the ``QueryParamsProvider`` they would have to run themselves the ``useQueryParams`` or use a wrapper that we would expose like ``useTimeRangeContextFromQueryParam`` and build from that the time range context (because there's not only time range but also refresh interval)

It would be fine to me but what I don't like is that it constraint to create a component that would be place under the QueryParamsProvider. Exit the straightforward way to integrate like in the doc https://github.com/perses/perses/blob/main/docs/embedded-panels.md#minimal-code
You may answer that nobody will anyway put all the providers like that in one only component, but this is exactly what I'm doing currently as I'm using a Perses panel in Angular environment :laughing: 

I'd be in favour of keeping what I did (maybe renaming the boolean to usingQueryParams) but with a default value to false. So that by default no error nothing specific apart from the given TimeRange. 
Note that if somebody want to modify the time range, they still can use the ``useTimeRange`` and modify from there. I even placed directly the ``<TimeRangeControls />`` in my code under the TimeRangeProvider and everything works well!

celian-garcia: Thank you for your participation I updated the PR with the code you gave to me.
I also adapted the `TemplateVariableProvider` to follow more or less the same principles except that I didn´t put the setter as it was a bit more complex than the time range provider (here there is zustand store with immer to partially modify variable definitions)

I will resolve this discussion as the breaking changes are properly listed in the PR description

jgbernalp (approved):
LGTM 🎉

PR #1476, opened 2023-10-13
[ENHANCEMENT] Auth return a JWT token

Nexucis:
With this PR, the auth endpoint is returning a signed JWT token


nicolastakashi:
LGTM!
Just one question.
What's `body.login`, is the user content?

nicolastakashi (approved):

Nexucis:
> What's `body.login`, is the user content?

Yeah exactly. In the body, the user is providing its login and its password.

PR #1477, opened 2023-10-13
[FEATURE] adding explore page

nicolastakashi:
# Description

Resolves: #1503

# Screenshots

image

# Checklist

## UI Changes


Nexucis:
I'm wondering if it would make sense to have a dedicated `npm` package for this feature. What do you think @nicolastakashi, @jgbernalp ?

jgbernalp:
That would be desirable. As I was commenting to @nicolastakashi, this might be a separate component that can be included inside a dashboard panel in the future, hopefully the `ViewExplore` component. So having it as its own package will make it easy to embed in other UIs.

nicolastakashi:
Yeah!
It makes sense to me.
I started under the dashboard package, just to be simpler for now.

But I'll take this in consideration to move to another package.

ui/app/src/views/explore/HelperExploreView.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { data: projectVars, isLoading: isLoadingProjectVars } = useVariableList(projectName);
+  const externalVariableDefinitions: ExternalVariableDefinition[] | undefined = useMemo(
+    () => [
+      buildProjectVariableDefinition(projectName || '', projectVars ?? []),
+      buildGlobalVariableDefinition(globalVars ?? []),
+    ],
+    [projectName, projectVars, globalVars]
+  );
+
+  if (isLoadingProjectVars || isLoadingGlobalVars) return null;
```
Gladorme, line 58: I think we could add a loading indicator here 😄 
```suggestion
  if (isLoadingProjectVars || isLoadingGlobalVars) {
    return (
      <Stack width="100%" sx={{ alignItems: 'center', justifyContent: 'center' }}>
        <CircularProgress />
      </Stack>
    );
  }
```

ui/app/src/views/explore/ExploreView.tsx
```diff
@@ -0,0 +1,59 @@
+// Copyright 2023 The Perses Authors
```
Gladorme, line 1: UI file archi is a little bit messy, but I think this folder should be moved to `ui/app/src/views/projects`, next to `dashboards` folder :)

ui/app/src/views/projects/ProjectTabs.tsx
```diff
@@ -99,7 +103,16 @@ function TabButton(props: TabButtonProps) {
...
         <>
-          <CRUDButton text="Add Dashboard" variant="contained" onClick={() => setCreateDashboardDialogOpened(true)} />
+          <Box>
+            <Stack direction="row" spacing={1}>
+              <CRUDButton
+                text="Add Dashboard"
+                variant="contained"
+                onClick={() => setCreateDashboardDialogOpened(true)}
+              />
+              <CRUDButton color="inherit" text="Explore" variant="contained" onClick={handleGoToExplore} />
```
Gladorme, line 113: The color in light theme is strange, no?

ui/explore/src/components/ExploreToolbar/ExploreToolbar.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <Box px={2} py={1} display="flex">
+        {exploreTitleComponent}
+      </Box>
+
+      <Box
+        sx={{
+          display: 'flex',
+          width: '100%',
+          alignItems: 'start',
+          padding: (theme) => theme.spacing(1, 2, 0, 2),
```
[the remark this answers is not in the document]
Gladorme: There is still a margin top that can remove, but even with marginTop: 0, it doesn't override it 😠

ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+              handleQueryDelete={handleQueryDelete}
+              queriesCollapsed={queriesCollapsed}
+              queries={queries}
+              defaultTimeSeriesQueryKind={defaultTimeSeriesQueryKind}
+            />
+          </Grid>
+          <Grid item xs={12}>
+            <Box
+              sx={{
+                padding: (theme) => theme.spacing(1),
```
Gladorme, line 97:
```suggestion
                paddingX: 1,
```
Let's reduce the blank space a little 😄

ui/app/src/views/projects/ProjectTabs.tsx
nicolastakashi: A bit!
Any suggestion?

Gladorme:
Awesome!
New feature: add a toggle to show/hide metric list below panel, I think it can be useful 😄

ui/app/src/views/projects/ProjectTabs.tsx
nicolastakashi: WDYT about secondary?
image

ui/app/src/views/projects/explore/ExploreView.tsx
```diff
@@ -0,0 +1,59 @@
...
+import ProjectBreadcrumbs from '../../../components/ProjectBreadcrumbs';
+import HelperExploreView from './HelperExploreView';
+
+const DEFAULT_DASHBOARD_NAME = 'Explore Page';
+
+/**
+ * TBD
+ */
+function ExploreView() {
+  const { projectName } = useParams();
```
AntoineThebaud, line 26: Probably for a future PR: What about having this explore view also available on the home page / not linked to a project?

ui/app/src/Router.tsx
```diff
@@ -39,6 +41,7 @@ function Router() {
           <Route path="/projects/:projectName" element={<GuardedProjectRoute />}>
             <Route path="" element={<ProjectView />} />
             <Route path=":tab" element={<ProjectView />} />
+            <Route path="explore" element={<ExploreView />} />
```
AntoineThebaud, line 44: Probably for a future PR: As it's working on Prometheus & Thanos UI (+ Grafana explore view), we may want to have the queries as part of the query params, so that you can share preconfigured explore views.

Gladorme (approved):
LGTM 👍

ui/app/src/views/projects/explore/ExploreView.tsx
nicolastakashi: this was something me and @Nexucis we discussed.
The idea of having it on the project level is to use to leverage global and project datasources.

I think might be confuse to the user Explore on two levels.
I mean the user cognitive load is a bit higher, because you need to think if you want to explore metrics from on project or from global source.

ui/app/src/Router.tsx
nicolastakashi: Yeah!
That makes sense.

ui/app/src/views/projects/explore/ExploreView.tsx
AntoineThebaud: Yeah valid point, still I think that if we make it explicit & with the same UX in both cases, this could work.

E.g we could imagine having a dropdown a the top to optionally select a project. Then, accessing the Explore view from your project space would prefill (+ lock) this dropdown, while accessing it from home page would leave it empty.

It could look like this:

perses_proposal_explore_project_setting

WDYT?

nicolastakashi: This is a good idea btw!
Maybe we can simplify even more and have only one global explore page with a project drop down, wdyt?

jgbernalp (approved):
LGTM 🎉

ui/app/src/views/projects/explore/ExploreView.tsx
Gladorme: I agree with a single explore page with dropdown 😄

Nexucis (approved):
Incredible work !! Thank you so much for doing it !

ui/core/package.json
```diff
@@ -33,7 +33,8 @@
     "lodash": "^4.17.21",
     "mathjs": "^10.6.4",
     "numbro": "^2.3.6",
-    "zod": "^3.21.4"
+    "zod": "^3.21.4",
+    "use-query-params": "^2.1.2"
```
sjcobb, line 37: 👋 Quickly skimming, but the ProjectStoreProvider move to core makes me nervous, wouldn’t it bring other dependencies too like react-query? Really core is meant to be the lowest level, most minimal package, mostly just for shared types and utils, not really providers, data fetching, or query param logic (really I had always wanted to refactor and make use-query-params completely optional even in plugin-system) 

What we probably need is some type of AppMetadataProvider that could even be decoupled from the concept of “projects”
That will take some trial and error to get around circular dependencies but even if that had to live in plugin-system, I think it’d be better than coupling core to the app

nicolastakashi: Yeah!
I won't lie I feel the same.
But I didn't get a better idea.
I'm not sure if we will be able build such a generic metadata provider but we can give a try

sjcobb: Yeah maybe it’s something we aspire to in the future, if this PR works for Red Hat’s use case it’s probably fine!

Would moving ProjectStoreProvider somewhere besides core be possible though (maybe plugin-system?), the other reason is I don’t want to indirectly make the components package depend on react-query, use-query-params, etc (this [old dependency diagram](https://github.com/perses/perses/blob/v0.22.0/ui/README.md#library-architecture-diagram) might be helpful to reference too)

nicolastakashi: Let's move it to plugin-system for now and we can feel if we are going to have more use cases like that and see what we should do.

Wdyt?

ui/explore/src/views/ViewExplore/ViewExplore.tsx
```diff
@@ -0,0 +1,74 @@
...
+export function ViewExplore(props: ViewExploreProps) {
+  const { dashboardResource, datasourceApi, externalVariableDefinitions, sx, exploreTitleComponent, ...others } = props;
+
+  const initialTimeRange = useInitialTimeRange(DEFAULT_DASHBOARD_DURATION);
+  const initialRefreshInterval = useInitialRefreshInterval(DEFAULT_REFRESH_INTERVAL);
+
+  return (
+    <DatasourceStoreProvider dashboardResource={dashboardResource} datasourceApi={datasourceApi}>
+      <DashboardProvider initialState={{ dashboardResource, isEditMode: true }}>
+        <TimeRangeProviderWithQueryParams
```
sjcobb, line 47: We should probably make a wrapper provider component at some point that encapsulates the common providers like TimeRangeProvider, TemplateVariableProvider, etc.

sjcobb: So something like this, but a first class version exported out from dashboards package:
https://github.com/sjcobb/perses-embed-example/blob/main/components/dashboards/PersesDashboardProvidersAlt.tsx#L38

ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+            <Box
+              sx={{
+                paddingX: 1,
+              }}
+            >
+              <Typography variant="h4" marginBottom={1}>
+                Preview
+              </Typography>
+              <ErrorBoundary FallbackComponent={ErrorAlert}>
+                <PanelPreview panelDefinition={panelDefinition} />
```
sjcobb, line 104: I think using the TimeSeriesChart panel directly would be more composable that using PanelPreview, it’d also be neat to always show the table legend in this case since we know there is room when in explore? Here are some examples:

https://github.com/sjcobb/prometheus/blob/sjcobb/mui-refactor-prototype/web/ui/react-app/src/components/perses/TimeSeriesPanel.tsx


https://github.com/sjcobb/prometheus/blob/sjcobb/mui-refactor-prototype/web/ui/react-app/src/components/perses/PersesQueryPanel.tsx#L57

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -29,6 +29,7 @@ export interface PluginEditorProps extends Omit<BoxProps, OmittedMuiProps> {
   pluginKindLabel: string;
   value: Definition<UnknownSpec>;
   isReadonly?: boolean;
+  isExplore?: boolean;
```
sjcobb, line 32: I’m confused why isExplore prop is added in so many places, can explore instead use the lower level building blocks directly? We can export them out if any are missing but it would give more control than using the higher level components that are more for dashboards and when the spec needs to be saved (explore is view-only right?)

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -66,46 +59,41 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
...
-          labelId="prom-datasource-label"
-          label="Prometheus Datasource"
-        />
-      </FormControl>
-      <PromQLEditor
-        completeConfig={{ remote: { url: promURL } }}
-        value={query}
-        onChange={handleQueryChange}
-        onBlur={handleQueryBlur}
+  if (isExplore) {
```
sjcobb, line 62: Does this isExplore condition need to be in prometheus-plugin this way? Can ui/explore import relevant lower components and rebuild the parts it needs, I’m thinking useState for managing active queries and interact with the query string directly

ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
sjcobb: Well I guess I’m too focused on time series queries in my comments actually, maybe the isExplore approach is okay after all 😅

sjcobb: Sorry for the stream of consciousness but my next question is are there changes we could make to the query editing on dashboards that would make this easier? It seems what we want is a more responsive experience that looks good in the panel edit drawer and in full screen explorer (side note: we could also make the panel edit drawer larger) plus some way to hide or show that ‘Projects’ dropdown. 

I just don’t totally understand why the UX would be different enough to warrant the isExplore prop 🤔

ui/explore/src/views/ViewExplore/ViewExplore.tsx
nicolastakashi: I almost sure @AntoineThebaud or @Gladorme is working in something like that

AntoineThebaud: I think you meant @celian-garcia here, he was working on refactorings around those providers to facilitate the embed use cases, e.g https://github.com/perses/perses/pull/1475

ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
nicolastakashi: Hey @sjcobb nothing to apologize mate ❤️ 
We'd like to reuse the entire plugin system we have but with a different flavor in terms of UI for the explore page.
I thougth about have a dedicated TimeSeries Plugin for explore page but @Nexucis @jgbernalp and I we think it might be a bit complex maintain both plugins.

In theory I'm expecing the TimeSeries Plugin be less couple to the UI, but currently it isn't (that's totally Ok), I'm open for a better suggestion on how to implement this tbh, I didn't have any better idea.

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
nicolastakashi: We thought about that, but this will require duplicate many components we're already have just because we want a different render style. Don't you think?

ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
sjcobb: I have some ideas but I don’t want to hold this up, I really need to start by pulling the changes down and

sjcobb (approved):

ui/explore/src/views/ViewExplore/ViewExplore.tsx
celian-garcia: Yeah I'm not sure what would be the best between a wrapper or more fine tuned provider asking less mandatory parameters sometimes not useful. 

e.g here the DatasourceStoreProvider ask for a dashboard which should be optional for sure. Then a dashboardProvider is necessary ? maybe this shouldn't

AntoineThebaud:
Spotted a minor issue when playing with it locally, when you select a project-specific datasource and then switch to another project:

https://github.com/perses/perses/assets/7058693/c383df41-d49e-48dc-ac15-1cbaad6f8828

jgbernalp:
LGTM.

Just some nits that should not block this PR and can be addressed in a follow up:

The dropdowns seem all to have different heights, not sure if its caused by its contents.
Screenshot 2023-11-14 at 11 49 22

Is necessary to have the legend name and min step?, it seems this is needed when editing an existing panel, but in this case the explore view purpose might be just to see what's there. A nice feature could be to create a panel from the current explore query.

nicolastakashi:
> LGTM.
> 
> Just some nits that should not block this PR and can be addressed in a follow up:
> 
> The dropdowns seem all to have different heights, not sure if its caused by its contents. Screenshot 2023-11-14 at 11 49 22
> 
> Is necessary to have the legend name and min step?, it seems this is needed when editing an existing panel, but in this case the explore view purpose might be just to see what's there. A nice feature could be to create a panel from the current explore query.

Good catch mate!

Just fix it.
image

Nexucis (approved):
That's an incredible work you made here @nicolastakashi. Thank you very much !

I agree with @sjcobb about plugin shouldn't know how they are used (explorer, dashboard, single usage outside of Perses etc.)

And as everyone else already said, yup provider should be simplified so it's easier to embed what we are providing.

But for sure that's a long term work and will require different followup PR. Until that, let's move forward on this PR.

I don't know if you were waiting my approval on this, but if it's the case, I'm sorry about the time it took me to look at it !

PR #1478, opened 2023-10-15
[FEATURE] Activate JWT middleware on any non anonymous endpoints

Nexucis:
This PR is activating the JWT middleware on the various non anonymous endpoints we have.

The PR is quite big unfortunately, mainly because I had to review the way we are registering the different route.
Most of them must be protected excepting the one required to create a new user. As I find quite nice that the `Endpoint` is providing itself its route and I didn't want to change that, I needed a way at the `Endpoint` level a way to say if the route is anonymous or not.
Depending of this parameter then, the JWT middleware will be activated or not.

So I ended up by creating our own `Route` and `Group` struct to be able to save the fact a route is anonymous or not.
Once the different `Route` are collected, the Server is going over all of them and register each of them in `echo` with the appropriate middleware depending of the route configuration.

And since I needed to modify the endpoint `user` to say that the route `/api/v1/users` required to create a user is an anonymous route, I had to stop ignoring the different file `endpoint`.


internal/api/config/config.go
```diff
@@ -88,6 +90,10 @@ func (c *Config) Verify() error {
 		return fmt.Errorf("encryption_key must be longer than 32 bytes")
 	}
 	c.EncryptionKey = promConfig.Secret(hex.EncodeToString([]byte(c.EncryptionKey)))
+	if c.ActivatePermission == nil {
+		var activatePermission = true
+		c.ActivatePermission = &activatePermission
```
nicolastakashi, line 95: Not required but we could leverage this package "k8s.io/utils/ptr" and use the `ptr.To` 😄

internal/api/core/server.go
```diff
@@ -65,25 +66,68 @@ func NewPersesAPI(serviceManager dependency.ServiceManager, persistenceManager d
...
-	a.registerAPIV1Route(e)
+	// First, let's collect every route.
+	// The expecting result is a tree we will need to loop over.
+	groups := a.collectRoutes()
+	// Now let's create a simple struct that will help us to loop over the route tree.
+	type queue struct {
+		parent *echo.Group
+		group  *shared.Group
+	}
+	var queueList []queue
```
nicolastakashi, line 89:
```suggestion
	queueList := make([]queue, 0, len(groups))
```

internal/api/core/server.go
```diff
@@ -65,25 +66,68 @@ func NewPersesAPI(serviceManager dependency.ServiceManager, persistenceManager d
...
 	}
 }
 
 func (a *api) RegisterRoute(e *echo.Echo) {
-	a.registerAPIV1Route(e)
+	// First, let's collect every route.
+	// The expecting result is a tree we will need to loop over.
+	groups := a.collectRoutes()
+	// Now let's create a simple struct that will help us to loop over the route tree.
+	type queue struct {
```
nicolastakashi, line 85: When I saw this I thought about queue data struct.
But I saw on the comment route tree I think it might be a bit confuse.
Just sharing a felling, I don't have a better name suggestion yet.

Nexucis: yeah I agree the name is not accurate since it's not the queue but a element of the queue actually.
Maybe `queueNode` ?

internal/api/core/server.go
Nexucis: The maximum size of `queueList` can be more than the length of `groups`. Not sure if it really worth to initialise the size of `queueList` in that case

internal/api/core/server.go
nicolastakashi: yeah, `queueNode` seems better, another suggestion would be `queueElement` feel free to chose anyone, I don't have preference over those.

internal/api/core/server.go
nicolastakashi: Oh nvm!
I see we might append to the queueList on the second loop.

internal/api/core/server.go
```diff
@@ -65,25 +66,68 @@ func NewPersesAPI(serviceManager dependency.ServiceManager, persistenceManager d
...
+		queueList = append(queueList, queue{group: g})
+	}
+	// It is our current element on each iteration.
+	var q queue
+	for len(queueList) > 0 {
+		// Let's grab the first element of the queue and remove it so the size of the queue is decreasing.
+		q, queueList = queueList[0], queueList[1:]
+		// Now we need to initialize the echo group that will be used to finally register in the router the different route.
+		var group *echo.Group
+		if q.parent != nil {
```
nicolastakashi, line 100: I think if we invert the if we can remove the else statement, wdyt?

internal/api/core/server.go
```diff
@@ -65,25 +66,68 @@ func NewPersesAPI(serviceManager dependency.ServiceManager, persistenceManager d
...
+		} else {
+			group = e.Group(q.group.Path)
+		}
+		// Then let's collect every child group, so we can loop over them during a future iteration.
+		for _, g := range q.group.Groups {
+			queueList = append(queueList, queue{group: g, parent: group})
+		}
+		// Finally, register the route with the echo.Group previously created.
+		// We will consider also if the route needs to remain anonymous or not and then inject the JWT middleware accordingly.
+		for _, route := range q.group.Routes {
```
nicolastakashi, line 113: This might reduce cyclomatic complexity
```suggestion
		for _, route := range q.group.Routes {
			if route.IsAnonymous {
				route.Register(group)
				continue
			}
			route.Register(group, a.jwtMiddleware)
		}
```

internal/api/core/server.go
Nexucis: `parent` is equal to nil at the first iteration, so unfortunately `group = q.parent.Group(q.group.Path)` without checking first `parent` will make the code panic with a null pointer exception

nicolastakashi: Hmm got it.
So we can keep the variable as is and move the `group = q.parent.Group(q.group.Path)` after the expression.
This should work or I'm missing something?

Nexucis: I'm not sure I understand what you mean by moving the assignment with the parent group after the expression :s

internal/api/core/server.go
Nexucis: I did a different thing here, since the `isAnonymous` is just changing the list of middleware, it makes more sense (for me) to reflect that than having two different way to register a route.

internal/api/core/server.go
nicolastakashi: Nvm, it's wrong what I thought about.

PR #1481, opened 2023-10-16
[IGNORE] Add authorization doc

Gladorme:
# Description
Add doc for future authorization

# Checklist

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -1,0 +1,171 @@
+# Authorization / Permission
+
+*WIP: The authorization is still not implemented in Perses*
+
+Perses will use a Role-based access control (RBAC) for regulating access to resources based on the role of the user.
+The RBAC API is based on four kinds of resource: GlobalRole, Role, GlobalRoleBinding and RoleBinding.
+Perses RBAC implementation is highly inspired by K8S RBAC implementation.
+
+
+## Role and GlobalRole
+
+An RBAC `Role` or `GlobalRole` contains a set of permissions (rules). Permissions are purely additive (there are no "deny" permissions).
+
+A `Role` defines a set of permissions within a particular project. When you create a `Role` you need to specify the project it belongs in.
+`GlobalRole`, by contrast, is not limited to a project scope.
+
+
+### Role example
+
+Here is an example of Role in "MySuperProject" project that can be used to grant edit access to dashboards:
+```yaml
+kind: Role
+metadata:
+  name: dashboard-editor
+  project: MySuperProject
+spec:
+  permissions:
+    - action: edit
+      scopes: ["Dashboard"]
+```
+
+
+### Global Role example
+
+A `GlobalRole` can be used to grant the same permissions as a Role. However, because `GlobalRole` are global, you can also use them to grant access to:
+- global resources (like Global Datasources, Global Variables, Users, ...)
+- project resources (like Dashboards) across all projects
+
+Here is an example of a `GlobalRole` that can be used to grant edit access to variables in all projects:
+```yaml
+kind: GlobalRole
+metadata:
+  name: variable-editor
+spec:
+  permissions:
+    - action: edit
+      scopes: ["Variable"]
+```
+
+
+## RoleBinding and GlobalRoleBinding
+
+A role binding grants the permissions defined in a role to a user or set of users. 
+It holds a list of subjects (users or teams) and a reference to the role being granted. A `RoleBinding` grants permissions within a specific project whereas a `GlobalRoleBinding` grants that access global-wide.
+
+A `RoleBinding` may reference any `Role` in the same project. Similarly, a `GlobalRoleBinding` can reference any `GlobalRole`.
+
+
+### RoleBinding example
+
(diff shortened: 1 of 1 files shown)


docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+
+## RBAC Synchro
+
+Roles and RoleBindings of an user are stored in the user's JWT.
+If Perses is deployed with multiple instances, Perses RBAC roles and role bindings cache need to be synchronized/replicated between all instances.
+To do that there are multiple mechanisms:
+- cache is refreshed every X minutes
+- cache is refreshed if roles and role bindings retrieve from the user's JWT are different from the cache
```
Nexucis, line 120: I would say it is refreshed only if the token is saying the user has more right than what is contained in the cache.

docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+*WIP: The authorization is still not implemented in Perses*
+
+Perses will use a Role-based access control (RBAC) for regulating access to resources based on the role of the user.
+The RBAC API is based on four kinds of resource: GlobalRole, Role, GlobalRoleBinding and RoleBinding.
+Perses RBAC implementation is highly inspired by K8S RBAC implementation.
+
+
+## Role and GlobalRole
+
+An RBAC *Role* or *GlobalRole* contains a set of permissions (rules). Permissions are purely additive (there are no "deny" permissions).
```
Nexucis, line 12:
```suggestion
An RBAC `Role` or `GlobalRole` contains a set of permissions (rules). Permissions are purely additive (there are no "deny" permissions).

File: docs/authorization.md [hunk 3]
```diff
@@ -0,0 +1,171 @@
...
+
+A *Role* set permission within a particular project, when you create a *Role* you need to specify the project it belongs in.
```
Nexucis, line 14:
```suggestion
A `Role` set permission within a particular project, when you create a *Role* you need to specify the project it belongs in.
```

docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+*GlobalRole*, by contrast, is not limited to a project scope.
```
Nexucis, line 15:
```suggestion
`GlobalRole`, by contrast, is not limited to a project scope.
```

docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+      scopes: ["Variable"]
+```
+
+
+## RoleBinding and GlobalRoleBinding
+
+A role binding grants the permissions defined in a role to a user or set of users. 
+It holds a list of subjects (users or teams) and a reference to the role being granted. A *RoleBinding* grants permissions within a specific namespace whereas a *GlobalRoleBinding* grants that access global-wide.
+
+A *RoleBinding* may reference any *Role* in the same namespace. Similarly, a *ClusterRoleBinding* can reference any *ClusterRole*.
```
AntoineThebaud, line 56:
```suggestion
A *RoleBinding* may reference any *Role* in the same project. Similarly, a *GlobalRoleBinding* can reference any *GlobalRole*.
```

docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+spec:
+  permissions:
+    - action: edit
+      scopes: ["Dashboard"]
+```
+
+
+### Global Role example
+
+A *GlobalRole* can be used to grant the same permissions as a Role. However, because *GlobalRole* are global, you can also use them to grant access to:
```
Nexucis, line 35:
```suggestion
A `GlobalRole` can be used to grant the same permissions as a Role. However, because `GlobalRole` are global, you can also use them to grant access to:
```

docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+
+
+## RBAC Synchro
+
+Roles and RoleBindings of an user are stored in the user's JWT.
+If Perses is deployed with multiple instances, Perses RBAC roles and role bindings cache need to be synchronized/replicated between all instances.
+To do that there are multiple mechanisms:
+- cache is refreshed every X minutes
+- cache is refreshed if roles and role bindings retrieve from the user's JWT are different from the cache
+- cache is refresh when a new role is created, edited or deleted
```
AntoineThebaud, line 121:
```suggestion
- cache is refreshed when a new role is created, edited or deleted
```

docs/authorization.md
```diff
@@ -0,0 +1,171 @@
...
+- cache is refresh when a new rolebinding is created, edited or deleted
```
AntoineThebaud, line 122:
```suggestion
- cache is refreshed when a new rolebinding is created, edited or deleted
```

Nexucis (approved):

docs/authorization.md
AntoineThebaud: I think the correct one is to put refresh**ed** on all rows

docs/authorization.md
AntoineThebaud: Are you sure it's namespace and not project here?

AntoineThebaud (approved):

nicolastakashi (approved):

PR #1483, opened 2023-10-16
[IGNORE] review the dashboard documentation

Nexucis:
Must be merged after #1500 as the dashboard is referencing the doc of the variable.


schemas/queries/prometheus/query.cue
```diff
@@ -21,6 +21,7 @@ kind: "PrometheusTimeSeriesQuery"
 spec: close({
 	datasource?: {
 		kind: prometheus.kind
+		name?: string
```
Nexucis, line 24: it looks like the datasource selector missed the `name` of the datasource for the query. Maybe we should create a common `datasource` selector ?

docs/api/plugin/prometheus.md
```diff
@@ -102,3 +102,39 @@ spec:
...
+A simple one: 
+
+```yaml
+kind: "TimeSeriesQuery"
+spec:
+  plugin:
+    kind: "PrometheusTimeSeriesQuery"
+    spec:
+      query: "rate(caddy_http_response_duration_seconds_sum[$interval])"
+```
```
Nexucis, line 140: Could be cool to have a more complexe example using every field, specially the `resolution`

docs/api/plugin/prometheus.md
```diff
@@ -102,3 +102,39 @@ spec:
...
+```yaml
+  # The promql query
+  query: <string>
+
+  # If empty, then the default PrometheusDatasource is chosen
+  # See the documentation about the datasources to understand how it is selected.
+  [ datasource: <datasource_selector> ]
+  [ seriesNameFormat: <string> ]
+  [ minStep: <duration> ]
+  [ resolution: number ]
```
Nexucis, line 126: if someone has a good explanation for each field, that would be nice to share !

docs/api/plugin/panel.md
```diff
@@ -0,0 +1,120 @@
+# Panel
```
Nexucis, line 1: Here also some help would be needed to add an explanation on the different attributes.

schemas/queries/prometheus/query.cue
AntoineThebaud: Lol this keeps coming back :D the name attribute is already defined in the base definition for queries https://github.com/perses/perses/blob/dbf9cb4c09aea90ab5970bcc7800e1ac786d93e4/internal/api/shared/schemas/base_def_query.cue#L20

This is true that it's kinda misleading / not clear here, but I don't know if we can improve this other than with proper documentation; One might think about using imports here to clearly show that this prometheus query definition is built over a generic one, but doing this would allow the future plugin designers to not do this import thus not comply with the expected datamodel..

docs/api/plugin/prometheus.md
```diff
@@ -103,6 +103,42 @@ spec:
...
+
+```yaml
+  # The promql query
+  query: <string>
+
+  # If empty, then the default PrometheusDatasource is chosen
+  # See the documentation about the datasources to understand how it is selected.
+  [ datasource: <datasource_selector> ]
+  [ seriesNameFormat: <string> ]
+  [ minStep: <duration> ]
```
AntoineThebaud, line 125:
```yaml
# `minStep` is the minimum time interval you want between each data points.
```

Nexucis:
Amazing review @AntoineThebaud thank you so much !

AntoineThebaud (approved):
LGTM!

PR #1484, opened 2023-10-16
[ENHANCEMENT] CLI cmd login can use a token

Nexucis:
This PR is modifying the CLI command `login` to be able to store a token or to use some credentials to get a token from the API


docs/cli.md
```diff
@@ -51,10 +51,16 @@ command.
...
-$ percli login https://perses.dev
+$ percli login https://demo.perses.dev
 ```
 
-The URL will be stored in JSON file that is by default `<UserHome>/.perses/config.json`.
+Depending on the server configuration, you will certainly have to provide a token or a user/password.
+
+- `--token` can be used to set a Bearer JWT token
+- `--username` and `--password` can be used to set a username & password. The command will contact the Perses server
+  with the credential. It will return a Bearer JWT token which expires in 1h
```
AntoineThebaud, line 61:
```suggestion
  with the credential(s). It will return a Bearer JWT token which expires after 1h.
```

internal/cli/cmd/login/login_test.go
```diff
@@ -25,13 +25,13 @@ func TestLoginCMD(t *testing.T) {
 			Title:           "empty args",
 			Args:            []string{},
 			IsErrorExpected: true,
-			ExpectedMessage: "no URL has been provided or has been found in the previous configuration",
+			ExpectedMessage: "no URL has been provided or has not been found in the previous configuration",
```
AntoineThebaud, line 28:
```suggestion
			ExpectedMessage: "no URL has been provided neither found in the previous configuration",
```

AntoineThebaud (approved):

PR #1486, opened 2023-10-17
[FEATURE] Enabling data provisioning

Nexucis:
This PR is proposing a way to have some data provisioning in Perses.

The provisioning service can read a list of folder and extract the known data from them. Each of them are then injected in the database regardless what exist. So any data that already exist are overridden by the provisioning service.

As a side (but nice) effect, we don't need to use the script `populate.sh` to generate the local database. The provisioning service is doing it for us.


Nexucis:
@nicolastakashi let me know what you think about that :)

internal/api/config/provisioning.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"encoding/json"
+	"time"
+)
+
+// jsonSchemas is only used to marshal the config in a proper json format
+// (mainly because of the duration that is not yet supported by json).
+type jsonProvisioningConfig struct {
+	Folders  []string `json:"folders,omitempty"`
```
Gladorme, line 24: yaml is missing 😄

Nexucis: nop it is not, look after ;)

Gladorme: 😇 Oops

Gladorme (approved):
LGTM 😄

AntoineThebaud (approved):

docs/configuration.md
```diff
@@ -92,6 +92,10 @@ Generic placeholders are defined as follows:
 
   # The markdown content to be displayed on the UI home page
   [ information: <string> ]
+
+  # If provided, Perses server will look to the different folders configured and populate the database based on what it is found
+  # The data coming from the provisioning folder will totally override what exists in the database.
```
nicolastakashi, line 97: When you say `The data coming from the provisioning folder will totally override what exists in the database.`, you mean it will delete the existing data and than insert the new one?

Nexucis: actually it updates the existing data. But it will replace whatever you change in the `spec`

internal/api/config/provisioning.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (p *ProvisioningConfig) Verify() error {
+	if p.Interval <= 0 {
+		p.Interval = defaultInterval
+	}
+	return nil
+}
+
+func (p ProvisioningConfig) MarshalJSON() ([]byte, error) {
+	j := &jsonProvisioningConfig{
```
nicolastakashi, line 42: I didn't get why we can't use this: ProvisioningConfig

internal/api/shared/dependency/provisioning.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		// retrieve if exists the entity from the Perses API
+		_, apiError := svc.Get(param)
+		if apiError != nil && !databaseModel.IsKeyNotFound(apiError) {
+			logrus.WithError(apiError).Errorf("unable to retrieve the %q from the database", kind)
+			continue
+		}
+
+		if databaseModel.IsKeyNotFound(apiError) {
```
nicolastakashi, line 83: Maybe.
```suggestion
		if databaseModel.IsKeyNotFound(apiError) {
			// the document doesn't exist, so we have to create it.
			if _, createError := svc.Create(entity); createError != nil {
				logrus.WithError(createError).Errorf("unable to create the %q %q", kind, name)
			}
			continue
		}
		// the document doesn't exist, so we have to create it.
		if _, updateError := svc.Update(entity, param); updateError != nil {
			logrus.WithError(updateError).Errorf("unable to update the %q %q", kind, name)
		}
```

internal/api/config/provisioning.go
Nexucis: that's because `time.Duration` is not marshalled properly in JSON unfortunately. For a duration the desired format is something like `1h` not just un big number

internal/api/shared/dependency/provisioning.go
Nexucis: you don't like the `if else` closure do you ? 😄

PR #1487, opened 2023-10-17
[BUGFIX] Fix grafana migration failing in some corner cases

AntoineThebaud:
# Description

2 different issues were making the grafana migration fail:
- Library panels: looks like "collapse them to make it work" is no longer valid, so we were trying to access a `type` attribute that doesn't exist in their case. It's now properly tackled (placeholder panel gets applied). Warning on the Migrate UI removed accordingly.
- Old query datamodel: nowadays the `query` attribute from the prometheus template variables is a struct and no longer a string. Grafana is not automatically doing the conversion so we have to support both here.

This also comes with some improvements for PromQL variables migration: stronger checks + no more replacement of builtin vars like `$__range` since we properly support them now.

# Checklist


Nexucis (approved):
🥳

PR #1489, opened 2023-10-18
[BREAKINGCHANGE] Change plugin structure to rely on regular .cue files only

AntoineThebaud:
# Description

Since https://github.com/perses/perses/pull/1032 we actually don't need anymore to rely on the homemade `.cuepart` extension for the plugin migration files (no more nested placeholder replacement).

Moving those to `.cue` extension implied to change further the plugin structure, because by default the schema loading was considering all .cue files it could find in the considered plugin folder. Thus to avoid considering the migration file as part of the regular plugin datamodel, the solution was to restrict the schema load to a specific package (that I named `model`, but open to proposals here if you have better ideas). I don't think this is an issue (it's just another rule to follow to develop plugins) but let me know if you disagree on that.

So this is technically a breaking change in regards to the plugin structure on backend side, but I assume that in reality this should not affect anyone at this stage.

Also this change involves excluding the migration files from `cue fmt` (because it's removing the trailing commas we need when assembling the conditions with the Go logic) as well as `make checklicense`.

# Checklist


schemas/panels/stat/stat.cue
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package stat
+package model
```
Nexucis, line 14: is it really a good thing to change the package name and have the same for everything ?

scripts/cue.sh
```diff
@@ -5,7 +5,8 @@
 set -e
 
 function fmt() {
-  find ./internal ./pkg ./schemas -name "*.cue" -exec cue fmt {} \;
+  # exclude migration files from the formatting since it removes the trailing commas we need
+  find ./internal ./pkg ./schemas -name "*.cue" -and -not -name "migrate.cue" -exec cue fmt {} \;
```
Nexucis, line 9: can't we add it programmatically and keep the formatter on every cue file ?

AntoineThebaud: the problem is that it's not as simple as adding a comma at the end of the payload: you can have several conditional blocks in a same file (like https://github.com/perses/perses/pull/1489/files#diff-15277e993e33a4adf4c22a6d46eed9f9582f2b0906824f999cf6a45d2b8cdb03)

schemas/panels/stat/stat.cue
AntoineThebaud: well imho as I put in the description _I don't think this is an issue (it's just another rule to follow to develop plugins)_. It shouldn't cause any kind of conflict since each plugin is evaluated independently.

Nexucis (approved):

PR #1490, opened 2023-10-18
[IGNORE] review datasource documentation

Nexucis:


docs/api/datasource.md
```diff
@@ -28,20 +26,107 @@ Once we will have user feedbacks, if they don’t like it, it won’t be that ha
...
+  metadata:
+    name: <string>
+    project: <string>
+  spec: <datasource_spec>
+```
+
+### Global level
+
+When we talk about scope and user permission in a REST API, the easiest way is to associate one permission per endpoint.
+If we want to provide a datasource shared by all projects, then it makes sense to create a different object that is
```
AntoineThebaud, line 42:
```suggestion
If we want to provide a datasource shared by all projects, then it makes sense to have a different object that is
```

docs/api/datasource.md
```diff
@@ -28,20 +26,107 @@ Once we will have user feedbacks, if they don’t like it, it won’t be that ha
...
+living outside a project.
+
+That’s why we will have a new resource called `GlobalDatasource`
```
AntoineThebaud, line 45:
```suggestion
That’s why we have another resource called `GlobalDatasource`
```

docs/api/datasource.md
```diff
@@ -28,20 +26,107 @@ Once we will have user feedbacks, if they don’t like it, it won’t be that ha
...
+So if by any chance you have a local datasource that is named exactly like a Project datasource, or a Global datasource,
+we will consider that the user intentionally wanted to override the upper datasource. We will use the one with the
+smallest scope.
+
+### How to use the Perses' proxy
+
+As described before, you can provide a proxy configuration that will be used by the Perses server to redirect any
+queries to the datasource.
+
+It means in case of the Prometheus datasource, if the field `directUrl` is not set, then the FE needs to use the Perses
```
AntoineThebaud, line 104:
```suggestion
It means in the case of the Prometheus datasource, if the field `directUrl` is not set then the FE needs to use the Perses
```

docs/api/plugin/prometheus.md
```diff
@@ -0,0 +1,102 @@
...
+  # Leave it empty if you don't want to restrict the access to the datasource.
+  allowedEndpoints:
+    - [ <allowedEndpoints_spec> ]
+
+  # It can be used to provide additional headers that need to be forwarded when requesting the datasource
+  headers:
+    [ <string>: <string> ]
+  # This is the name of the secret that should be used for the proxy or discovery configuration
+  # It will contain any sensitive information such as password, token, certificate.
+  # Please the documentation about the secret to understand how to create it
```
AntoineThebaud, line 46:
```suggestion
# Please the documentation about secrets to understand how to create one
```

docs/api/datasource.md
```diff
@@ -1,11 +1,8 @@
-# Datasource Specification
+# Datasource
 
-This documentation is describing how a datasource can be defined and can be used in a dashboard. It is the conclusion of
-the discussion #404.
+## Different Datasource scope
```
AntoineThebaud, line 3:
```suggestion
## Choose a scope
```

docs/api/datasource.md
```diff
@@ -1,11 +1,8 @@
...
 
-## How datasource can be defined
-
-We will have three different ways to define a datasource, depending on the scope we want to give
+We have three different ways to define a datasource, depending on the scope we want to give
```
AntoineThebaud, line 5:
```suggestion
There are 3 different scopes in which you can define a datasource, depending at which point you want it to be shared.
- for common use cases, use higher scopes in order to reuse the same datasource on multiple dashboards
- for more specific needs, use lower scopes to restrict the datasource availability to a specific set of (or even a single) dashboard(s).
```

docs/api/datasource.md
```diff
@@ -28,20 +26,108 @@ Once we will have user feedbacks, if they don’t like it, it won’t be that ha
...
+interface DatasourceSelector {
+  kind: string;
+  // name is the unique name of the datasource to use
+  // If name is omitted, that effectively means "use the default datasource for this kind".
+  name?: string;
 }
 ```
 
-#### API definition
+Priority is `"Local datasource in the dashboard" > "Project datasource" > "Global datasource"`.
```
AntoineThebaud, line 93:
```suggestion
The priority order is `"Local datasource in the dashboard" > "Project datasource" > "Global datasource"`.
```

docs/api/datasource.md
```diff
@@ -1,47 +1,131 @@
-# Datasource Specification
+# Datasource
 
-This documentation is describing how a datasource can be defined and can be used in a dashboard. It is the conclusion of
-the discussion #404.
+## Choose a scope
 
-## How datasource can be defined
+There are three different scopes in which you can define a datasource, depending on which point you want it to be shared.
```
AntoineThebaud, line 5:
```suggestion
There are three different scopes in which you can define a datasource, depending on how much you want it to be shared.
```

docs/api/plugin/prometheus.md
```diff
@@ -0,0 +1,104 @@
...
+## Datasource
+
+```yaml
+kind: "PrometheusDatasource"
+spec: <prometheus_plugin_spec>
+```
+
+### Plugin Specification
+
+Prometheus as a datasource is basically an HTTP server. So we need to define am HTTP config.
```
AntoineThebaud, line 14:
```suggestion
Prometheus as a datasource is basically an HTTP server. So we need to define an HTTP config.
```

AntoineThebaud (approved):

Issue #1353, opened 2023-08-05
Upgrade to React 18


sjcobb:
Will be a decent amount of work but it will need to happen eventually. Can also leverage many newer features, ex: enhanced dashboard loading states and there are TODOs in our panel key generation to leverage the new built-in hook

Part of this will be telling our known embed users such as Red Hat to understand what that will mean for them and to give notice

AntoineThebaud:
FYI this move is now [part of the roadmap](https://github.com/perses/perses/blob/main/ROADMAP.md#dropping-support-for-react-17) with the deadline of mid-july next year for the upgrade.

PR #1494, opened 2023-10-19
[BREAKINGCHANGE] Upgrade to React 18.2.0

Gladorme:
# Description

Upgrade all package to React 18

# Checklist

## UI Changes


ui/prometheus-plugin/package.json
```diff
@@ -40,7 +40,7 @@
     "immer": "^9.0.15"
   },
   "peerDependencies": {
-    "react": "^17.0.2 || ^18.0.0"
```
Nexucis, line 43: do we have a reason to remove the support of react v17 ?
I don't see any changes in this PR that required this drop, am I missing something ?

Gladorme: If we want to support newer React features #1353, React 17 support will need to drop

Nexucis: yeah I agree but that's not the case for the moment excepting if you are willing to do it :)

Gladorme: Updated ❤️

Nexucis (commented):
sounds good to me. @jgbernalp is ok on RedHat side to drop the support of React 17 ?

jgbernalp:
@Gladorme Unfortunately I think it might break some components for us. Is there a strong reason why we should change the `useId` hook already provided by perses components package?

Gladorme:
@jgbernalp No strong reasons, I saw we had some warnings about deps vulnerability and I wanted to update of our deps. I know we had an topic about React 17 -> React 18, so I started by this 😄 
Let's put in standby for now

Nexucis:
@jgbernalp is it still not ok to drop the support of react v17 ?

jgbernalp:
@Nexucis unfortunately yes. We still depend on react 17 for dynamic plugins on the OpenShift console, this is to add support for Perses dashboards to web consoles for older OpenShift versions. But we can make an assessment of the impact once we have the console plugin ready for distributed tracing and revisit this PR.

Nexucis:
let's close this PR.
When we will be able to make the upgrade there will be more changes anyway. So this PR cannot be merged just like that.

PR #1496, opened 2023-10-20
[FEATURE] display the server config

Nexucis:
This PR is adding a new page to display the backend config.

Maybe we should put every button in a menu on the right corner.

I took the opportunity to split the breadcrumbs wrapper we have in two different components. I changed also the icon for the admin panel

image

### DIFF
--- a/ui/app/src/views/admin/AdminView.tsx
+++ b/ui/app/src/views/admin/AdminView.tsx
@@ -21,7 +21,7 @@ function AdminView() {
   const { tab } = useParams();
   return (
     <Stack sx={{ width: '100%' }} m={2} gap={2}>
-      <AppBreadcrumbs rootPageName="Admin" />
+      <AppBreadcrumbs rootPageName="Administration" />
       <Box>
         <Stack direction="row" alignItems="center" justifyContent="space-between">
           <Stack direction="row" alignItems="center" gap={1}>
--- a/ui/app/src/views/config/ConfigView.tsx
+++ b/ui/app/src/views/config/ConfigView.tsx
@@ -12,27 +12,27 @@
 // limitations under the License.
 
 import { Box, Stack, Typography } from '@mui/material';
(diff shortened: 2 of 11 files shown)


AntoineThebaud (approved):
Great to have this! I think it would be better to display it in YAML rather than JSON, but we dont have a ready-to-use component for this.

ui/app/src/views/config/ConfigView.tsx
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+
+import { Box, Stack, Typography } from '@mui/material';
+import Cog from 'mdi-material-ui/Cog';
+import { JSONEditor } from '@perses-dev/components';
+import AppBreadcrumbs from '../../components/AppBreadcrumbs';
+import { useConfig } from '../../model/config-client';
+
+function ConfigView() {
+  const cfg = useConfig();
```
Gladorme, line 21: I think we should add loading support:
```suggestion
  const { data, isLoading } = useConfig();
```

Gladorme: And add loading element: 
```
if (isLoading) {
    return (
        <CircularProgress />
      </Stack>
    );
  }
```

ui/app/src/views/admin/AdminView.tsx
```diff
@@ -21,11 +21,11 @@ function AdminView() {
   const { tab } = useParams();
   return (
     <Stack sx={{ width: '100%' }} m={2} gap={2}>
-      <AppBreadcrumbs admin={true} />
+      <AppBreadcrumbs rootPageName="Admin" />
```
Gladorme, line 24:
```suggestion
      <AppBreadcrumbs rootPageName="Administration" />
```
Let's use the same name as the title, no?

Gladorme:
Not a huge fan of add "Config" to the navbar like this, as you said it will be probably better on the right.

Gladorme (approved):

PR #1500, opened 2023-10-20
[IGNORE] add doc about variable

Nexucis:


docs/api/variable.md
```diff
@@ -0,0 +1,230 @@
...
+  # If you are creating a GlobalVariable or a Variable, you don't have to use this attribute as it is replaced by metadata.name.
+  # This is the unique name of the variable that can be used in another variable or in the different dashboard to use
+  [ name: <string> ]
+
+  [ display: <display_spec> ]
+
+  # It can be a single value or a list.
+  [ defaultValue: <string> | <array of string> ]
+  [ allowAllValue: <boolean> | default = false ]
+  [ allMultiple: <boolean> | default = false ]
```
Nexucis, line 117: it would be cool if someone have a good explanation for each field here. It would help to complete the doc :)

docs/api/plugin/prometheus.md
```diff
@@ -102,3 +102,148 @@ spec:
...
+  # If empty, then the default PrometheusDatasource is chosen
+  # See the documentation about the datasources to understand how it is selected.
+  [ datasource: <datasource_selector> ]
+  matchers:
+   [ - <string> ]
+```
+
+#### Example
+
+A simple Prometheus LabelValues variable would be
```
AntoineThebaud, line 127:
```suggestion
A simple Prometheus LabelNames variable would be
```

docs/api/variable.md
```diff
@@ -0,0 +1,230 @@
...
+```yaml
+  kind: "GlobalVariable"
+  metadata:
+    name: <string>
+  spec: <variable_spec>
+```
+
+## Variable specification
+
+We are supporting two different types of variable. `TextVariable` and `ListVariable`
```
AntoineThebaud, line 52:
```suggestion
We are supporting two different types of variables: `TextVariable` and `ListVariable`.
```

docs/api/variable.md
```diff
@@ -0,0 +1,230 @@
...
+
+  [ display: <display_spec> ]
+  value: <string>
+  [ constant: <boolean> | default = false ]
+```
+
+#### Example
+
+```yaml
+kind: "Variable"
```
AntoineThebaud, line 77:
```suggestion
kind: "Variable" # alternatively "GlobalVariable"
```

docs/api/variable.md
```diff
@@ -0,0 +1,230 @@
...
+
+#### Get a list of `Variable`
+
+```bash
+GET /api/v1/projects/<project_name>/variables
+```
+
+URL query parameters:
+
+- name = `<string>` : should be used to filter the list of variables based on the prefix name.
```
AntoineThebaud, line 170:
```suggestion
name = <string> : filters the list of variables based on their name (prefix match).
```

docs/api/variable.md
```diff
@@ -0,0 +1,230 @@
...
+
+#### Get a list of `GlobalVariable`
+
+```bash
+GET /api/v1/globalvariables
+```
+
+URL query parameters:
+
+- name = `<string>` : should be used to filter the list of datasource based on the prefix name.
```
AntoineThebaud, line 206:
```suggestion
name = <string> : filters the list of variables based on their name (prefix match).
```

docs/api/variable.md
AntoineThebaud: I'd say

defaultValue: value from the list to be selected by default.
allowAllValue: whether or not to append the "All" value that allows to select all available values at once.
allowMultiple: whether or not to allow multi-selection of values.

AntoineThebaud (approved):

PR #1511, opened 2023-10-24
[FEATURE] Add Role, RoleBinding, GlobalRole and GlobalRoleBinding kinds

Gladorme:
# Description

Add Role, RoleBinding, GlobalRole and GlobalRoleBinding kinds for futur authorization as describe in the [doc](https://github.com/perses/perses/blob/main/docs/authorization.md) to the backend


docs/api/role.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Permission definition
+
+```yaml
+  # The type of the action. For example: `create`, `read`, `update` or `delete`
+  action: <string>
+
+  # The list of kind targeted by the permission. For example: `Datasource`, `Dashboard`, ...
+  # With Role, you can't target global kinds
+  scopes: [<string>]
```
Nexucis, line 48:
```suggestion
  scopes:
  - <string>
```

Like that it's clearer it's an array of string

docs/api/role.md
```diff
@@ -0,0 +1,144 @@
...
+  metadata:
+    name: <string>
+  spec: <role_spec>
+```
+
+## Role specification
+
+```yaml
+  # List of permissions owned by the role
+  permissions: [<permission>]
```
Nexucis, line 37:
```suggestion
  permissions:
  - <permission_spec>
```

docs/api/role.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+### Permission definition
```
Nexucis, line 40:
```suggestion
### `<permission_spec>`
```

docs/api/rolebinding.md
```diff
@@ -0,0 +1,145 @@
...
+  spec: <rolebinding_spec>
+```
+
+## RoleBinding specification
+
+```yaml
+  # Name of the Role or GlobalRole concerned by the role binding (metadata.name)
+  role: <string>
+  # Subjects that will inherit permissions from the role
+  subjects: [<subject>]
```
Nexucis, line 39:
```suggestion
  subjects:
  - <subject_spec>
```

docs/api/rolebinding.md
```diff
@@ -0,0 +1,145 @@
+# RoleBinding
```
Nexucis, line 1: a short explanation about what is the roleBinding won't hurt ;)

docs/api/role.md
```diff
@@ -0,0 +1,144 @@
+# Role
```
Nexucis, line 1: a quick explanation about what is a role won't hurt ;)

.gitignore
```diff
@@ -21,7 +21,7 @@ Thumbs.db
 
 coverage.txt
 
-/internal/api/e2e/test/
+/internal/api/e2e/api/test/
```
Nexucis, line 24: sounds weird we need that here

internal/api/impl/v1/rolebinding/service.go
```diff
@@ -0,0 +1,116 @@
...
+	// find the previous version of the roleBinding
+	oldEntity, err := s.dao.Get(parameters.Project, parameters.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	// If you do want to change the role for a binding, you need to remove the binding object and create a replacement.
+	// More info at: https://github.com/perses/perses/blob/main/docs/authorization.md#rolebinding-and-globalrolebinding-update-restriction
+	if entity.Spec.Role != oldEntity.Spec.Role {
+		return nil, shared.HandleBadRequestError("spec.role can't be updated")
```
Nexucis, line 95: I'm wondering if it's a big issue to be able to change the role

internal/api/interface/v1/globalrolebinding/interface.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/perses/perses/internal/api/shared"
+	databaseModel "github.com/perses/perses/internal/api/shared/database/model"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+type Query struct {
+	databaseModel.Query
+	// NamePrefix is a prefix of the GlobalRoleBinding.metadata.name that is used to filter the list of the GlobalRoleBinding.
+	// NamePrefix can be empty in case you want to return the full list of GlobalRoleBinding available.
+	NamePrefix string `query:"name"`
```
Nexucis, line 26: do we need a search by user ? (same question for the `RoleBinding` ofc)

pkg/model/api/v1/kind.go
```diff
@@ -124,3 +144,24 @@ func GetStruct(kind Kind) (modelAPI.Entity, error) {
 		return nil, fmt.Errorf("%q has no associated struct", kind)
 	}
 }
+
+func IsGlobal(kind Kind) bool {
+	switch kind {
+	case KindGlobalDatasource:
```
Nexucis, line 150:
```suggestion
case KindGlobalDatasource, KindGlobalRole, KindGlobalRoleBinding, KindGlobalSecret, KindGlobalVariable, KindProject, KindUser:
  return true
```
```

File: pkg/model/api/v1/role.go [hunk 11]
```diff
@@ -0,0 +1,178 @@
...
+	KindWildcard ActionKind = "*"
+	KindRead     ActionKind = "read"
+	KindCreate   ActionKind = "create"
+	KindUpdate   ActionKind = "update"
+	KindDelete   ActionKind = "delete"
+)
+
+type Permission struct {
+	// Action of the permission (read, create, update, delete, ...)
+	Action ActionKind `json:"action" yaml:"action"`
```
Nexucis, line 41: probably we want a list here so we can have a role that cover multiple actions, like `create`, `read`

pkg/model/api/v1/role.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	*g = tmp
+	return nil
+}
+
+func (g *GlobalRole) validate() error {
+	if g.Kind != KindGlobalRole {
+		return fmt.Errorf("invalid kind: %q for a GlobalRole type", g.Kind)
+	}
+	if reflect.DeepEqual(g.Spec, RoleSpec{}) {
```
Nexucis, line 89:
```suggestion
if g.Spec == RoleSpec{} {
}
```
that should be simple as that.

pkg/model/api/v1/role.go
```diff
@@ -0,0 +1,178 @@
...
+
+const (
+	KindWildcard ActionKind = "*"
+	KindRead     ActionKind = "read"
+	KindCreate   ActionKind = "create"
+	KindUpdate   ActionKind = "update"
+	KindDelete   ActionKind = "delete"
+)
+
+type Permission struct {
```
Nexucis, line 39: we certainly want to check that `scopes` and `actions` are not empty (both of them)

pkg/model/api/v1/rolebinding.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	*r = tmp
+	return nil
+}
+
+func (r *RoleBinding) validate() error {
+	if r.Kind != KindRoleBinding {
+		return fmt.Errorf("invalid kind: %q for a RoleBinding type", r.Kind)
+	}
+	if reflect.DeepEqual(r.Spec, RoleBindingSpec{}) {
```
Nexucis, line 207:
```suggestion
if r.Spec == RoleBindingSpec{} {
}
```

pkg/model/api/v1/rolebinding.go
```diff
@@ -0,0 +1,227 @@
...
+
+func (g *GlobalRoleBinding) GetMetadata() modelAPI.Metadata {
+	return &g.Metadata
+}
+
+func (g *GlobalRoleBinding) GetKind() string {
+	return string(g.Kind)
+}
+
+func (g *GlobalRoleBinding) GetRoleBindingSpec() RoleBindingSpec {
```
Nexucis, line 161: do you plan to add some generic method ? (I didn't see them for the moment)

internal/api/impl/v1/globalrolebinding/service.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *service) Create(entity api.Entity) (interface{}, error) {
+	if object, ok := entity.(*v1.GlobalRoleBinding); ok {
+		return s.create(object)
+	}
+	return nil, shared.HandleBadRequestError(fmt.Sprintf("wrong entity format, attempting GlobalroleBinding format, received '%T'", entity))
+}
+
+func (s *service) create(entity *v1.GlobalRoleBinding) (*v1.GlobalRoleBinding, error) {
+	// TODO: validate user + role exists
+	if len(entity.Spec.Subjects) == 0 {
```
Nexucis, line 50: this check should be at the Unmarshal level I believe. That's because, It's not a functional verification, it's more something that is checking the integrity of the entity itself (and also it is a static check)

Nexucis (commented):
Thank you for this (huge ;)) PR. I think I hope I covered everything !
Let me know if you have further questions.

.gitignore
Gladorme: Well it's where json test files from e2e api integration tests are currently created 😄

internal/api/impl/v1/rolebinding/service.go
Gladorme: I found this restriction by looking K8S perm doc and I'm ok with their [reasons](https://github.com/perses/perses/blob/main/docs/authorization.md#rolebinding-and-globalrolebinding-update-restriction)

internal/api/interface/v1/globalrolebinding/interface.go
Gladorme: Maybe, I will implement it in the next PR if necessary 🤔

pkg/model/api/v1/role.go
Gladorme: I copy/paste from datasource/dashboard validation. If I recall correctly it was for checking that spec was correctly set

pkg/model/api/v1/rolebinding.go
Gladorme: I don't known, I will remove it for now

pkg/model/api/v1/role.go
Gladorme: Roles have a list of permissions. One permission per action kind, then any scope you want. 
It will be easier to manage when you want to add a permission to a scope.
ie: add read perm to var

You check if the role has already a perm with read action and add variable to scope list, else create new read perm with var in scope list

If permissions have list of actions, I don't know how you will select which permission to add scope.

Really easier with only one action

internal/api/interface/v1/globalrolebinding/interface.go
Nexucis: yeah probably we will see if we need it later

pkg/model/api/v1/role.go
Nexucis: > If permissions have list of actions, I don't know how you will select which permission to add scope.

if you have 

```yaml
permissions:
- actions:
  - read
  - create
  scopes:
  - "Dashboards"
  - "Variables"
```

it just means you can `create` and `read` `Dashboards` and `Variables` I don't see why it's more complexe. And it gives more flexibility from the user perspective.

Besides, since you have an array of `Permission`, that's already possible, it's just longer to do it.

For the moment to be able to do what I did above, I need to do that:

```yaml
permissions:
- actions:
  - read
  scopes:
  - "Dashboards"
  - "Variables"
- actions:
  - create
  scopes:
  - "Dashboards"
  - "Variables"
```

At the end that's the same thing, it's just longer to write. So from my perspective, you are able to manage what is currently proposed in the PR, you should be able to manage what I am suggesting (aka an array of action)

Gladorme: Totally agree it's easier for users, but I think it will be more tricky to implem. Let's see

Nexucis: tricky doesn't mean impossible, it's just mean you need to think a bit more ;)

Nexucis (approved):
well done :)

PR #1557, opened 2023-10-30
[ENHANCEMENT] Store JWT token in cookies and redirect if the token is expired

Nexucis:
In order to easily authenticate the UI, I have created two different cookies:
- `jwtPayload`: it contains the readonly payload of the token that can be decoded by the UI.
- `jwtSignature`: it contains the signature of the token. This shouldn't be read or manipulate by the UI.

In case both cookies are present, the backend will take both value to recreate the complete token and then inject it in the header `Authorization`

Note: the back still accept a JWT token that would be in the header `Authorization`. We are still able to use the CLI with the authentication activated.

On the UI side, I have created a guard that is checking if the cookie `jwtPayload` is present and then if the token expired. In case the token expired (or doesn't exist) then it will redirect to the sign-in page.

I think on a future PR, we should manage automatically a refresh of the token. Currently it is set to one hour and after one hour regardless you are still using Perses, you will have to re sign in.


internal/api/shared/crypto/jwt.go
```diff
@@ -28,6 +38,11 @@ type JWTCustomClaims struct {
 
 type JWT interface {
 	SignedToken(login string) (string, error)
+	// CreateJWTCookies will create two different cookies that contain a piece of the token.
+	// As a reminder, a JWT token has the following structure: header.payload.signature
+	// The first cookie will contain the struct header.payload that can then be manipulated by Javascript
```
jgbernalp, line 43: In which use case do we need the client to manipulate the token payload? If the payload is changed then the token will be tampered and the signature won't match.

Nexucis: maybe the word `manipulate` is not the correct one. It's more on the client side, we want to decode the token to know if it is expired (at least) and then later to know which permission have the user so we can modify the UI to reflect these permission.

jgbernalp: In that case the complete token can be used, as decoding does not require signature validation.

Nexucis: yeah I know :). That's what I'm doing. I just store the token in two different cookies based on what I read here: https://stackoverflow.com/a/61106253. And I'm quite convinced to follow.

AntoineThebaud (approved):
LGTM

Gladorme (approved):
LGTM

PR #1559, opened 2023-10-31
[ENHANCEMENT] File DB is the default choice if no db configuration is provided

Nexucis:
By adding this default choice, we are able to run `Perses` with no configuration now.

### DIFF
--- a/internal/api/config/database.go
+++ b/internal/api/config/database.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/prometheus/common/config"
+	"github.com/sirupsen/logrus"
 )
 
 type FileExtension string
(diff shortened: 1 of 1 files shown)


internal/api/config/database.go
```diff
@@ -126,7 +127,10 @@ type Database struct {
 
 func (d *Database) Verify() error {
 	if d.File == nil && d.SQL == nil {
-		return fmt.Errorf("you must specify if Perses has to use SQL or filesystem as a database")
+		logrus.Debug("no database has been specified, therefore a file system database is used")
+		d.File = &File{
+			Folder: "./local_db",
```
AntoineThebaud, line 132: could you add a const for this as we have for other configs?

```golang
const (
	//...
        DefaultFileDBFolder = "./local_db"
)

//...
			Folder: DefaultFileDBFolder,

```

PR #1560, opened 2023-10-31
[BREAKINGCHANGE] Move any securiy config to a dedicated struct

Nexucis:
I'm adding a new config called `security` that will contain any configuration regarding the permission, roles ..etc.


Gladorme (approved):
Nice idea :)

Issue #1551, opened 2023-10-27
Grafana migration: user inputs ignored in the Perses part


AntoineThebaud:
### What did you do?

- Go to migrate screen
- Provide a correct payload in the "Provide a Grafana dashboard" part
- Click on migrate
- In the "Migration output", modify a field e.g metadata.name because you expect to encounter a conflict with an existing dashboard when saving.
- provide project an click Import
- It fails with "document already exists" although you did provide a different name. Actually any user input in the Perses json part is ignored: if I want to change the dashboard name, I should do it in the Grafana part & click migrate again

### What did you expect to see?

I expect to have either 
- post-migration modifications done by user taken into account,
- or "Migration output" field to be readonly, to not give the wrong impression you can modify things afterwards.

### What did you see instead? Under which circumstances?

User inputs are allowed but get ignored behind the scenes.

### Perses version

main-2023-10-26-d44ceabd

Issue #1518, opened 2023-10-25
CTRL+F not working in all JSON views


AntoineThebaud:
~~### Proposal~~

~~Support searching via CTRL+F in the JSON display. Currently it would be the default browser one & the input text can only match the currently-displayed content on the window.~~

Typing CTRL+F in JSON Editor should always display the find&replace interface, but in some cases it doesn't.

AntoineThebaud:
NB currently the search within the full JSON is already possible on the Migration screen, but only this one: we dont have the same search fields appearing on the dashboard view.

AntoineThebaud:
It's actually a bug since this is a default behavior, the find & replace interface actually shows up for a brief moment but then closes immediately.

PR #1562, opened 2023-11-01
[BUGFIX] Fix several issues with JSON views

AntoineThebaud:
# Description

All JSON views:
- find & replace interface is now showing up correctly. Fixes https://github.com/perses/perses/issues/1518.

Migrate screen:
- Grafana input: changed from **TextField** to **JSONEditor**. No more weird behavior when trying to modify things after initial copy-paste, & better UX overall.
- Migrate button: stronger checks for enabling (JSON parse should succeed).
- Migration output: it's now readonly. Fixes https://github.com/perses/perses/issues/1551.

# Screenshots

image

image

image

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/model/migrate-client.ts
+++ b/ui/app/src/model/migrate-client.ts
@@ -20,7 +20,7 @@ const resource = 'migrate';
 
 export interface MigrateBodyRequest {
   input?: Record<string, string>;
-  grafanaDashboard: string;
+  grafanaDashboard: Record<string, unknown>;
 }
 
 export function useMigrate() {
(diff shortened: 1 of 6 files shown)


ui/app/src/model/migrate-client.ts
```diff
@@ -20,7 +20,7 @@ const resource = 'migrate';
 
 export interface MigrateBodyRequest {
   input?: Record<string, string>;
-  grafanaDashboard: string;
+  grafanaDashboard: Record<string, unknown>;
```
Nexucis, line 23: why do you need a record here ?

Nexucis (approved):
overall lgtm

ui/app/src/model/migrate-client.ts
AntoineThebaud: since I changed the type of `grafanaDashboard` to Record<string, unknown> upstream, I though it would be better to pass it like this to the migrate-client and stringify it just before sending the request, so that we have both args `input` and `grafanaDashboard` behaving the same here. But I dont mind changing it back if you prefer.

PR #1566, opened 2023-11-02
[IGNORE] Fix redirection when token expired

Nexucis:
When the token expired, the cookies are removed since we set `maxAge` in the previous PR.

The redirection is currently broken because the cookies are removed when expire.

I think it's cleaner to remove the cookies when they expire and I think it is simpler to check if the cookies exist rather than trying to know if the token has been decoded and if it is expired (for the redirection I mean)

### DIFF
--- a/ui/app/src/guard/GuardedAuthRoute.tsx
+++ b/ui/app/src/guard/GuardedAuthRoute.tsx
@@ -16,14 +16,14 @@ import { useSnackbar } from '@perses-dev/components';
 import { Suspense, useEffect, useState } from 'react';
 import { LinearProgress } from '@mui/material';
 import { useGetConfigMutation } from '../model/config-client';
-import { useAuthToken } from '../model/auth-client';
+import { useIsTokenExist } from '../model/auth-client';
 import { SignInRoute } from '../model/route';
 
 function GuardedAuthRoute() {
   const { mutateAsync } = useGetConfigMutation();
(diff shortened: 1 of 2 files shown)


ui/app/src/model/auth-client.ts
```diff
@@ -29,9 +30,17 @@ export interface AuthBody {
   password: string;
 }
 
+export function useIsTokenExist() {
```
Gladorme, line 33: I don't really know what is returning useJwt, could we use the decoded payload for the check? At least we know the payload format is fine

Nexucis: well if there is no token then there is decoded payload ;)

ui/app/src/guard/GuardedAuthRoute.tsx
```diff
@@ -40,7 +40,8 @@ function GuardedAuthRoute() {
           if (!conf.security.activate_permission) {
             return true;
           }
-          if (isExpired) {
+          // In case the token is null, it means we weren't able to find the cookie.
+          if (!isTokenExist) {
```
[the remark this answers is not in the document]
Nexucis: when the token doesn't exist, isExpired is equal to false. And when the token expire => the cookies are removed =>  so the token doesn't exist anymore => so `isExpired` is equal to `false`

PR #1589, opened 2023-11-15
[BUGFIX] Fix YAML files not working for provisioning

karsov:
# Description

When using YAML files for initial provisioning (located in a folder defined in the `config.yaml` file's `provisioning.folders` field),
Perses fails to load the `Dashboard` and `GlobalDatasource` definitions throwing the following error:

```text
...
time="2023-11-14T09:43:46Z" level=debug msg="unable to marshal the plugin \"TimeSeriesChart\"" error="json: unsupported type: map[interface {}]interface {}"
time="2023-11-14T09:43:46Z" level=error msg="unable to create the \"Dashboard\" \"Demo_Dashboard\"" error="bad request: json: unsupported type: map[interface {}]interface {}"
time="2023-11-14T09:43:46Z" level=debug msg="unable to marshal the plugin \"PrometheusDatasource\"" error="json: unsupported type: map[interface {}]interface {}"
time="2023-11-14T09:43:46Z" level=error msg="unable to create the \"GlobalDatasource\" \"DemoPrometheus\"" error="bad request: json: unsupported type: map[interface {}]interface {}"
...
```

I turns out this is caused by the `gopkg.in/yaml.v2` library unmarshaling some YAML parts (specifically the `Plugin` spec) into `map[interface {}]interface {}`, which is not supported for marashaling by Go standard library's `enconding/json`.

This PR:

- updates Perses to use the newer `gopkg.in/yaml.v3` package which uses `map[string]interface{}` (supported by `encoding/json`), thus fixing the errors;
- removes the usage of `github.com/json-iterator/go`, because previously it was only added to address the exact issue described above (apparently that did not fix it completely), and now that it is addressed by the YAML package update there is no further need for it;
- changes slightly the output of `percli version` to have bigger indentation, because `gopkg.in/yaml.v3` uses 4 spaces for indentation by default (compared to 2 spaces in `gopkg.in/yaml.v2`).

# Checklist


Nexucis (approved):
nice thank you for taking time to fix this bug !!

PR #1602, opened 2023-11-23
[ENHANCEMENT] Implement Refresh token logic

Nexucis:
Implementation of the refresh token logic based on what is described in the [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749): as advised in https://stackoverflow.com/a/69631673

### What's new

- There is a new endpoint called `/auth/refresh`. This endpoint will get the refresh token from the cookies or the body and if the refresh token is still valid, then it will store and return a new access token.
- The endpoint `/auth` is now returning a refresh token and an access token.

As a side note, the access and the refresh token doesn't have the same signature to avoid using the refresh token as an access token.

- On the UI side side, I wrapped `fetch` to integrate an automatic refresh of the token based on the HTTP error code.

As an another side note, doing that removed the usage of the `AuthGuard`


internal/api/impl/auth/auth.go
```diff
@@ -56,15 +58,69 @@ func (e *Endpoint) auth(ctx echo.Context) error {
...
 	}
-	jwtHeaderPayloadCookie, signatureCookie := e.jwt.CreateJWTCookies(token)
-	ctx.SetCookie(jwtHeaderPayloadCookie)
-	ctx.SetCookie(signatureCookie)
+	refreshToken, err := e.refreshToken(ctx, body.Login)
+	if err != nil {
+		return err
+	}
+
 	return ctx.JSON(http.StatusOK, api.AuthResponse{
```
celian-garcia, line 70: Don´t you need to change the UI part because of this change ? To at least have a working dev env?

Nexucis: yup I need to work on that to complete this PR !

ui/app/src/model/fetch.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SignInRoute } from './route';
+
+export async function fetch(...args: Parameters<typeof global.fetch>) {
+  return initialFetch(...args).catch((error: UserFriendlyError | FetchError) => {
+    if (error.status !== 401) {
+      throw error;
+    }
+    refreshToken().catch((refreshTokenError: UserFriendlyError | FetchError) => {
+      if (refreshTokenError.status === 400) {
+        window.location.href = SignInRoute;
```
Gladorme, line 25: RIP error toast 🍞 

I think we should add the location of the current page in param, so if the user relogin he will be redirect to the wanted page. Maybe in another PR?

Gladorme (approved):
LGTM, small comments. With logout button auth UI side should be good 😄

ui/app/src/model/fetch.ts
Nexucis: aaah yeah you are right !
Let me check if it's a quick win

internal/api/shared/crypto/jwt.go
```diff
@@ -30,12 +30,31 @@ import (
 const (
 	CookieKeyJWTPayload   = "jwtPayload"
 	CookieKeyJWTSignature = "jwtSignature"
+	CookieKeyRefreshToken = "jwtRefreshToken"
+)
+
+var (
+	accessTokenExpiration  = time.Minute * 15
+	refreshTokenExpiration = time.Hour * 24
```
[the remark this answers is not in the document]
Nexucis: yeah you are probably right. I will add that then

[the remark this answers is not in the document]
Nexucis: config added

Gladorme (approved):

PR #1603, opened 2023-11-25
[ENHANCEMENT] Prevent using dashboard datasources before they're saved

sinkingpoint:
# Description

Previously, when a dashboard level datasource is added to a dashboard, it is immediately available for use in the dashboard. This causes an error because the server side (e.g. the proxy endpoints) isn't created until the dashboard is saved.

Here we introduce the idea of "saved" datasources, which are the datasources that are available for use in the dashboard because they have been created through a dashboard save.

We then use this list of saved datasources to disable entries in the datasource select dropdown, with an indication that the dashboard needs to be saved before the datasource can be used.

# Screenshots

image

https://github.com/perses/perses/assets/4386405/63a71f53-1636-41fb-8e46-e036bca62643

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/components/Datasources/EditDatasourcesButton.tsx
+++ b/ui/dashboards/src/components/Datasources/EditDatasourcesButton.tsx
@@ -23,8 +23,9 @@ import { DatasourceEditor } from './DatasourceEditor';
 
 export function EditDatasourcesButton() {
   const [isDatasourceEditorOpen, setIsDatasourceEditorOpen] = useState(false);
-  const { getLocalDatasources, setLocalDatasources } = useDatasourceStore();
+  const { getLocalDatasources, setLocalDatasources, getSavedDatasources, setSavedDatasources } = useDatasourceStore();
   const localDatasources: Record<string, DatasourceSpec> = getLocalDatasources();
+  const savedDatasources: Record<string, DatasourceSpec> = getSavedDatasources();
   const { dashboard, setDashboard } = useDashboard();
 
   const openDatasourceEditor = () => {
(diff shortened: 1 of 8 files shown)


Gladorme (approved):
LGTM, really cool!

AntoineThebaud (changes requested):
Two issues in the current state:
1. You can still encounter the error by changing a saved datasource with mode=direct to mode=proxy (and not saving yet)
2. You cannot use a not-yet-saved datasource that has mode=direct, while the querying would work.

IMHO it's ok to let point 1 aside for now, what you did is already a good improvement. However I would like to be able to use new datasources right away when they have mode=direct.

sinkingpoint:
Updated this with a change to the `EditDatasourcesButton` - we now allow datasources that are direct to be used immediately, and we also _disallow_ datasources that are proxied, but weren't before (i.e. they've been updated to be a proxy). This does introduce a rather gnarly conditional though

ui/dashboards/src/components/Datasources/EditDatasourcesButton.tsx
```diff
@@ -35,6 +36,40 @@ export function EditDatasourcesButton() {
...
+      .filter((key) => {
+        // Datasources are allowed to be used if a) they are direct, or b) they are proxied, and their
+        // proxy is the same as what we have saved.
+        const isDirect = 'directUrl' in (datasources[key]?.plugin?.spec ?? {});
+        const isSavedProxy =
+          !isDirect &&
+          !('directUrl' in (savedDatasources[key]?.plugin?.spec ?? {})) &&
+          datasources[key]?.plugin?.spec?.proxy === savedDatasources[key]?.plugin?.spec?.proxy;
+
+        console.log('name', key, 'isDirect', isDirect, 'isSavedProxy', isSavedProxy);
```
AntoineThebaud, line 51: leftover log

AntoineThebaud (approved):
Awesome, thanks for the fix! LGTM (there's just a remaining console.log to cleanup ;)

sinkingpoint:
Knew I missed something. Fixed!

PR #1604, opened 2023-11-27
[ENHANCEMENT] Use a context for the config to share it across the whole app

Nexucis:
This PR is just refactoring the way to access the config. I'm introducing a context so the different component can have access to the config without having to deal with async and network issue.

Today the config is already heavily use and I think it makes sense to simplify the usage and its access of it.

I find the code cleaner with this way (but perhaps it's just matter of test ;))

### DIFF
--- a/ui/app/src/context/Config.tsx
+++ b/ui/app/src/context/Config.tsx
@@ -1,0 +1,63 @@
+// Copyright 2023 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


ui/app/src/context/Config.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface ConfigContextType {
+  config: ConfigModel;
+}
+
+const ConfigContext = createContext<ConfigContextType | undefined>(undefined);
+
+export function ConfigContextProvider(props: { children: React.ReactNode }) {
+  const { data, isLoading } = useConfig();
+  if (isLoading || data === undefined) {
+    return <></>;
```
Gladorme, line 28:
```suggestion
    return <Stack width="100%" sx={{ alignItems: 'center', justifyContent: 'center' }}>
        <CircularProgress />
      </Stack>;
```

Nexucis: okey. I just wished we would have the Perses logo animated instead of the MUI spinner 😢

Gladorme (approved):
LGTM 😄

PR #1609, opened 2023-11-28
[IGNORE] Allow the use of JS/TS falsiness

Gladorme:
# Description

I want to propose a change to the UI guideline: allow to use falsiness of TS/JS.

It's way easier and shorter:
- for checking if a string is `undefined`, `null` or empty ("")
- for checking if an object is `undefined` or `null`

Code can still use explicit `null` or `undefined` checks if needed.

For info, these are the following falsy values in TypeScript and JavaScript:
- `0`
- `false`
- `NaN`
-  `""` (empty string)
- `0n` (bigint version of zero)
- `null`
- `undefined`

# Checklist

### DIFF
--- a/ui/ui-guidelines.md
+++ b/ui/ui-guidelines.md
@@ -280,7 +280,6 @@ to ensure you are writing modular UI code.
 - Start thinking about breaking a module into smaller modules or sub components when
   it exceeds 300 lines, unless there is a good reason to couple all of the code in a
   single module)
-- Use explicit `null` or `undefined` checks instead of relying on JS falsiness
 - Bias towards using optional properties (i.e. `foo?: bar`) and `undefined` over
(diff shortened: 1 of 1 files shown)


AntoineThebaud (approved):
Personally I don't have a strong opinion on this topic, and it doesn't seem like there's a consensus on whether relying on javascript truthiness/falsiness is a bad practice.

It provides a convenient shorthand syntax so I think it's ok to rely on it when it makes sense (= you want the same behavior no matter which truthy/falsy value you encountered). After all you can always change to explicit undefined vs null vs empty etc.. checks when you expect different behaviors for each case.

sjcobb (approved):
SGTM! I don’t think we were consistent with following that guideline anyways

PR #1617, opened 2023-12-04
[ENHANCEMENT] Healthcheck datasources when they're saved

sinkingpoint:
# Description

This updates the datasource editor to test the datasource when it's
saved. This pretty much mirrors how Grafana does this, with a "Save and Test"
button when editing datasources that saves the datasource and then runs
a test.

Most of this PR is making the saving process properly async so that we can call the test 
_after_ the datasource is saved.

There's a few things of note:

- Users of the <DatasourceEditorForm /> component no longer need to call
their `onClose` when the datasource is saved - the editor will call
it when the health check is sucessful.

- Dashboard local datasources _are not healthchecked_, for mostly the
same reasons as why they can't be immediately used - they don't exist
serverside until the dashboard is saved.

Signed-off-by: sinkingpoint <colin@quirl.co.nz>

# Screenshots

https://github.com/perses/perses/assets/4386405/eb9fc27a-bf80-4b80-a351-8b27ef8f24e5

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/datasource/DatasourceDrawer.tsx
+++ b/ui/app/src/components/datasource/DatasourceDrawer.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { Action, Datasource, DatasourceSpec, DispatchWithPromise } from '@perses-dev/core';
-import { Dispatch, DispatchWithoutAction, useState } from 'react';
+import { DispatchWithoutAction, useState } from 'react';
 import { Drawer, ErrorAlert, ErrorBoundary } from '@perses-dev/components';
 import { DatasourceEditorForm, PluginRegistry } from '@perses-dev/plugin-system';
 import { bundledPluginLoader } from '../../model/bundled-plugins';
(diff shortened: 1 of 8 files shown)


Nexucis:
nice !

I know that's not possible for the moment, but I think it would be nice as a user to be able to make the healthcheck before saving the datasource.

Perhaps we should provide an endpoint where you can have in the body the datasource definition and the backend/proxy is performing the check

sinkingpoint:
I was thinking about that when writing this - between this and my previous PR about local datasources, I think an endpoint like that would actually be really valuable. I'm happy to abandon this PR and do that instead - it _would_ make this a lot nicer. I've never really liked Grafanas pattern here

Nexucis:
> I was thinking about that when writing this - between this and my previous PR about local datasources, I think an endpoint like that would actually be really valuable. I'm happy to abandon this PR and do that instead - it _would_ make this a lot nicer. I've never really liked Grafanas pattern here

let's do that then !!

I'm just wondering where this endpoint should go:
* under the /api/v1
* or under the /proxy. Probably makes more sense but the `/proxy` is a middleware for the moment, not the best place to add a new endpoint.

sinkingpoint:
I'll do a bit of investigation and see what's feasible - we can discuss in chat

PR #1621, opened 2023-12-05
[BUGFIX] Wrap PrometheusDatasourceEditor with react-hook-form Controllers

sinkingpoint:
# Description

Previously, the fields created with the PrometheusDatasourceEditor weren't wired into the form infrastructure. This means that, among other things, if you edit the URL of the datasource, the form doesn't know that the datasource has changed and doesn't allow you to save.

This change wraps the PrometheusDatasourceEditor inputs with react-hook-form Controllers, which means that the form is now aware of changes.

# Screenshots

Before: (Editing the URL doesn't allow you to save)

https://github.com/perses/perses/assets/4386405/1d4602f6-b913-40e3-8e8f-c9513623c01c

After:

https://github.com/perses/perses/assets/4386405/fd146205-53b4-4a6c-b492-5fe6818953b7

# Checklist

## UI Changes

### DIFF
--- a/ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor.tsx
+++ b/ui/prometheus-plugin/src/plugins/PrometheusDatasourceEditor.tsx
@@ -16,6 +16,7 @@ import { OptionsEditorRadios } from '@perses-dev/plugin-system';
 import { Grid, IconButton, MenuItem, TextField, Typography } from '@mui/material';
 import React, { Fragment, useState } from 'react';
 import { produce } from 'immer';
+import { Controller } from 'react-hook-form';
 import MinusIcon from 'mdi-material-ui/Minus';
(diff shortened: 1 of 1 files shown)


Gladorme (approved):
I really need to finish my PR on the validation #1438 and have all form editors fixed and aligned with what has been done with Role & RoleBinding Form Editor...
We shouldn't have to pass `value` in Controller as it should be initialized by `useForm`, it will use the default value of the controller name.
When there is an array of controllers like allowed endpoints, we should use useFieldArray ([example here](https://github.com/perses/perses/blob/79e68259c5997e9fa519515653ab908935105383/ui/app/src/components/roles/RoleEditorForm.tsx#L54)), it will simplify the current logic to add or remove a new endpoint.
Approving because it still better than before, but I will need to improve it again when I will continue on my validation PR 😄

PR #1626, opened 2023-12-05
[ENCHANCEMENT] Refresh cache after Role or RoleBinding update + impro…

Gladorme:
…ve validation

# Description

Adding refresh when a GlobalRole, Role, GlobalRoleBinding and RoleBinding is created, updated or deleted.
Adding validation on RoleBinding payload, checking if role exists and if user subjects exist

# Checklist


internal/api/impl/v1/globalrole/service.go
```diff
@@ -52,6 +55,9 @@ func (s *service) create(entity *v1.GlobalRole) (*v1.GlobalRole, error) {
 	if err := s.dao.Create(entity); err != nil {
 		return nil, err
 	}
+	if err := s.rbac.Refresh(); err != nil {
```
Nexucis, line 58: Can you add a comment explaining why we are refreshing the `rbac` cache ?

internal/api/impl/v1/globalrole/service.go
```diff
@@ -78,11 +84,17 @@ func (s *service) update(entity *v1.GlobalRole, parameters apiInterface.Paramete
...
+	}
 	return entity, nil
 }
 
 func (s *service) Delete(_ apiInterface.PersesContext, parameters apiInterface.Parameters) error {
-	return s.dao.Delete(parameters.Name)
+	if err := s.dao.Delete(parameters.Name); err != nil {
+		return err
+	}
+	return s.rbac.Refresh()
```
Nexucis, line 97: it would be weird to return an error because the refresh cache failed right ? Specially because the resources has been successfully deleted

Nexucis: actually everytime we are refreshing the cache and it fails probably we should just log the error and not returning it to the user

internal/api/impl/v1/globalrolebinding/service.go
```diff
@@ -101,3 +124,22 @@ func (s *service) Get(_ apiInterface.PersesContext, parameters apiInterface.Para
...
+func (s *service) validateGlobalRoleBinding(globalRoleBinding *v1.GlobalRoleBinding) error {
+	_, err := s.globalRoleDAO.Get(globalRoleBinding.Spec.Role)
+	if err != nil {
+		return shared.HandleBadRequestError(fmt.Sprintf("global role '%s' doesn't exist", globalRoleBinding.Spec.Role))
+	}
+
+	for _, subject := range globalRoleBinding.Spec.Subjects {
+		if subject.Kind == v1.KindUser {
+			_, err := s.userDAO.Get(subject.Name)
+			if err != nil {
```
Nexucis, line 138:
```suggestion
if _, err := s.userDAO.Get(subject.Name); err != nil {
  if databaseModel.IsKeyNotFound(err) {
    return shared.HandleBadRequestError(fmt.Sprintf("user subject name '%s' doesn't exist", subject.Name))
  }
  logrus.WithError(err).Errorf("unable to find the user with the name %q", subject.Name)
  return err
}
```

internal/api/impl/v1/globalrolebinding/service.go
```diff
@@ -101,3 +124,22 @@ func (s *service) Get(_ apiInterface.PersesContext, parameters apiInterface.Para
 func (s *service) List(_ apiInterface.PersesContext, q databaseModel.Query, _ apiInterface.Parameters) (interface{}, error) {
 	return s.dao.List(q)
 }
+
+// Validating role and subjects are existing
+func (s *service) validateGlobalRoleBinding(globalRoleBinding *v1.GlobalRoleBinding) error {
+	_, err := s.globalRoleDAO.Get(globalRoleBinding.Spec.Role)
+	if err != nil {
+		return shared.HandleBadRequestError(fmt.Sprintf("global role '%s' doesn't exist", globalRoleBinding.Spec.Role))
```
Nexucis, line 132:
```suggestion
return shared.HandleBadRequestError(fmt.Sprintf("global role %q doesn't exist", globalRoleBinding.Spec.Role))
```

internal/api/impl/v1/globalrolebinding/service.go
Gladorme: oops

PR #1635, opened 2023-12-10
[ENHANCEMENT] Add in a proxy for unsaved datasources

sinkingpoint:
# Description

Add ability to proxy unsaved datasources

This adds the ability to proxy unsaved datasources. This is effectively
the same as the existing proxy functionality, but loading the datasource
from a spec in the request body rather than from the database. This is useful
for testing datasources before saving them, and for using local datasources
before the dashboard has been saved.

In order to not make this an open proxy, we authenticate calls to the
proxy endpoint, ensuring that users have create datasource permissions.
This is a bit awkward to do in the proxy, because it required another middleware
to extract the claims, but it seems to work - I'd appreciate any comments
on making that cleaner.

# Checklist

### DIFF
--- a/internal/api/impl/proxy/proxy.go
+++ b/internal/api/impl/proxy/proxy.go
@@ -15,6 +15,7 @@ package proxy
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
@@ -29,23 +30,52 @@ import (
 	"github.com/perses/perses/internal/api/interface/v1/globaldatasource"
 	"github.com/perses/perses/internal/api/interface/v1/globalsecret"
 	"github.com/perses/perses/internal/api/interface/v1/secret"
+	"github.com/perses/perses/internal/api/shared"
 	"github.com/perses/perses/internal/api/shared/crypto"
 	databaseModel "github.com/perses/perses/internal/api/shared/database/model"
+	"github.com/perses/perses/internal/api/shared/rbac"
 	"github.com/perses/perses/internal/api/shared/route"
 	"github.com/perses/perses/internal/api/shared/utils"
 	v1 "github.com/perses/perses/pkg/model/api/v1"
 	datasourceHTTP "github.com/perses/perses/pkg/model/api/v1/datasource/http"
+	"github.com/perses/perses/pkg/model/api/v1/role"
 	promConfig "github.com/prometheus/common/config"
 	"github.com/sirupsen/logrus"
 )
+
+var _ = json.Unmarshaler(&unsavedProxyBody{})
 
 // unsavedProxyBody is the body of the request when the datasource is not saved yet.
 // It contains the body of the request and the datasource spec, which is used to build the proxy rather than
 // retrieving the datasource from the database.
 type unsavedProxyBody struct {
-	Method string             `json:"method" yaml:"method"`
-	Body   []byte             `json:"body" yaml:"body"`
-	Spec   *v1.DatasourceSpec `json:"spec" yaml:"spec"`
+	Method string            `json:"method" yaml:"method"`
+	Body   []byte            `json:"body,omitempty" yaml:"body"`
+	Spec   v1.DatasourceSpec `json:"spec" yaml:"spec"`
+}
+
+func (u *unsavedProxyBody) UnmarshalJSON(data []byte) error {
+	type Alias unsavedProxyBody
+	aux := Alias{}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	if aux.Method == "" {
+		return fmt.Errorf("missing field 'method'")
+	}
+
+	if aux.Method != http.MethodGet && aux.Method != http.MethodPost && aux.Method != http.MethodPut && aux.Method != http.MethodDelete {
+		return fmt.Errorf("invalid method %q", aux.Method)
+	}
+
+	if _, err := datasourceHTTP.ValidateAndExtract(aux.Spec.Plugin.Spec); err != nil {
+		return fmt.Errorf("invalid http config: %w", err)
+	}
+
+	*u = unsavedProxyBody(aux)
+	return nil
 }
 
 func (u *unsavedProxyBody) setRequestParams(ctx echo.Context) {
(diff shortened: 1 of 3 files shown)


Nexucis (commented):
excepting the permission, it looks ok.

It makes sense to check the permission for this part like you said, but probably we need to change the way the proxy is implemented to make it work

internal/api/core/core.go
```diff
@@ -51,6 +51,10 @@ func New(conf config.Config, banner string) (*app.Runner, dependency.Persistence
 		DTS:          persistenceManager.GetDatasource(),
 		GlobalDTS:    persistenceManager.GetGlobalDatasource(),
 		Crypto:       serviceManager.GetCrypto(),
+		RBAC:         serviceManager.GetRBAC(),
+		JWTMiddleware: serviceManager.GetJWT().Middleware(func(echo.Context) bool {
```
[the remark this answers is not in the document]
sinkingpoint: :/ I was hoping by just generating the middleware and then calling it directly in the proxy that it'd keep those anonymous endpoints, but it seems you're right. It was definitely a cause of confusion for me that the proxy is implemented as a middleware rather than just a regular route so I'm happy to investigate moving it across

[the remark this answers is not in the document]
Nexucis: If you want to rewrite it using the route, I would propose to do it in another PR and then go back to this once the refactoring is done.
By the way if rewriting the proxy using only thé route doesn't work, perhaps a combination with route and middleware would make it.
Like fake route that doesn't do anything but is just there to register the path and then a middleware apply to this route that is actually doing the job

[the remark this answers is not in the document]
sinkingpoint: https://github.com/perses/perses/pull/1644

sinkingpoint:
Updated this to use the new proxy routes. The authentication is a lot cleaner.

internal/api/impl/proxy/proxy.go
```diff
@@ -28,67 +29,179 @@ import (
...
 		Crypto:       crypto,
+
+		rbac: rbac,
+	}
+}
+
+func (e *endpoint) checkPermission(ctx echo.Context, projectName string) error {
+	claims := crypto.ExtractJWTClaims(ctx)
+	if claims == nil {
+		// Claims can be nil with anonymous endpoint and unauthenticated users, no need to continue
```
Nexucis, line 94: Actually in that context, it can be nil if we have skipped the authentication (by configuration). Otherwise it won't be possible.

Can you update the comments please ?

internal/api/impl/proxy/proxy.go
```diff
@@ -28,67 +29,179 @@ import (
...
 		GlobalSecret: globalSecretDAO,
 		DTS:          dtsDAO,
 		GlobalDTS:    globalDtsDAO,
 		Crypto:       crypto,
+
+		rbac: rbac,
+	}
+}
+
+func (e *endpoint) checkPermission(ctx echo.Context, projectName string) error {
```
Nexucis, line 91:
```suggestion
func (e *endpoint) checkPermission(ctx echo.Context, projectName string, scope role.Scope) error {
        claims := crypto.ExtractJWTClaims(ctx)
	if claims == nil {
	  return nil
	}
	
	if role.IsGlobalScope(scope) {
	  [...]
	}
	[....]
}
```

I think you need to pass the scope in the parameter. In the current implementation the scope is hardcoded so even for the `GlobalDatasource` you won't check the accurate permission.

And then in the method `proxyUnsavedGlobalDatasource` you can use it like that 


```go
if err := e.checkPermission(ctx, rbac.GlobalProject, role.GlobalDatasourceScope); err != nil {
		return err
}
```

internal/api/impl/proxy/proxy.go
```diff
@@ -28,67 +29,179 @@ import (
...
+		return nil
+	}
+
+	scope, err := role.GetScope(string(v1.KindDatasource))
+	if err != nil {
+		return err
+	}
+
+	if role.IsGlobalScope(*scope) {
+		if ok := e.rbac.HasPermission(claims.Subject, role.CreateAction, rbac.GlobalProject, *scope); !ok {
```
Nexucis, line 104: if you pass the action in parameter, you will be able to check the permission for the saved datasource as well. For the saved one, you should need the `readAction`

Nexucis (commented):
nice that's better indeed !

internal/api/impl/proxy/proxy.go
sinkingpoint: Done!

internal/api/impl/proxy/proxy.go
sinkingpoint: Ah yep - missed that there was a  `KindGlobalDatasource` distinct from `KindDatasource`. Updated  this to take a kind and an action

internal/api/impl/proxy/proxy.go
sinkingpoint: Forgot we were going to do that - done!

internal/api/impl/proxy/proxy.go
```diff
@@ -28,67 +29,187 @@ import (
...
 		GlobalSecret: globalSecretDAO,
 		DTS:          dtsDAO,
 		GlobalDTS:    globalDtsDAO,
 		Crypto:       crypto,
+
+		rbac: rbac,
 	}
 }
 
+func (e *endpoint) checkPermission(ctx echo.Context, projectName string, scopeKind v1.Kind, action role.Action) error {
```
Nexucis, line 91: instead of using the `v1.Kind` for `scopeKind`, why not using directly the `role.Scope` that match the type of datasource we want to use ? https://github.com/perses/perses/blob/main/pkg/model/api/v1/role/scope.go#L28

Like that we would avoid the usage of the method `role.GetScope`.

internal/api/impl/proxy/proxy.go
```diff
@@ -28,67 +29,187 @@ import (
...
+}
+
 type endpoint struct {
 	Dashboard    dashboard.DAO
 	Secret       secret.DAO
 	GlobalSecret globalsecret.DAO
 	DTS          datasource.DAO
 	GlobalDTS    globaldatasource.DAO
 	Crypto       crypto.Crypto
+
```
Nexucis, line 74: no need for this new line.

and actually we could privatise every attribute of this struct.

```go
type endpoint struct {
	dashboard    dashboard.DAO
	secret       secret.DAO
	globalSecret globalsecret.DAO
	dts          datasource.DAO
	globalDTS    globaldatasource.DAO
	crypto       crypto.Crypto
	rbac rbac.RBAC
}
```

sinkingpoint: Done

internal/api/impl/proxy/proxy.go
sinkingpoint: I was sort of assuming that `role.GetScope` did something a bit more complicated 😅 Done

internal/api/impl/proxy/proxy.go
```diff
@@ -28,67 +29,181 @@ import (
...
 	"github.com/sirupsen/logrus"
 )
 
+// unsavedProxyBody is the body of the request when the datasource is not saved yet.
+// It contains the body of the request and the datasource spec, which is used to build the proxy rather than
+// retrieving the datasource from the database.
+type unsavedProxyBody struct {
+	Method string             `json:"method" yaml:"method"`
+	Body   []byte             `json:"body" yaml:"body"`
+	Spec   *v1.DatasourceSpec `json:"spec" yaml:"spec"`
```
Nexucis, line 51: excepting the `body` attribute, all others attributes are mandatory. I suggest to :
- add the key `omitempty` for the body
- Override the method `UnmarshalJSON` to include the verification that `method` is not empty
- I guess we also need to check that `method` is a correct HTTP method
- And why not removing the pointer for `Spec`. Like that we are ensuring the `Spec` cannot be empty

sinkingpoint: Done! I was tossing up doing validation in the `UnmarshalJSON`, or having an external `Validate` that was called after each `Bind`. Calling it in `Bind` would allow us to validate, regardless of Content-Type WDYT?

Nexucis: Nah it's ok to do it in the `UnmarshalJSON` as there is no plan to support anything else than JSON when contacting the Perses HTTP API

Nexucis (approved):
awesome ! Thank you for doing it and keeping up to date after so many reviews !

PR #1636, opened 2023-12-11
[ENHANCEMENT] move config to pkg so its available publicly

jgbernalp:
# Description

In order to use the same types for Perses configuration, it should be available publicly

# Checklist


Nexucis:
I think it would be better if the package is under the /api folder : `pkg/api/config`. (because it's the API's config)

Looks good excepting that !

Nexucis (approved):
👍

PR #1637, opened 2023-12-11
[FEATURE]: Authentication with external OAuth/OIDC providers (Authorization code flow)

celian-garcia:
# Description

Let's introduce the possibility to setup some external authentication provider. 
It regards two type of provider here : Oauth 2.0 and OIDC. 

With this new feature, we can for example sign in with Azure (OIDC) or Github (Oauth 2.0) and all is driven by configuration on backend side.
Check authentication documentation added for more details.

### Important notes 
- Only **Authorization Code** flow is supported here. **Device code** and **Client Credentials** flow will be implemented in a subsequent PR.
- this is not fully working, some thoughts and implementation is necessary to save the user in DB properly. That's why **I propose to merge it in a feature branch** instead of breaking the main branch.

### What is this PR taking in charge:

- ✅ Authorization code flow for interactive flow (human being) documentation
- ✅ Multiple auth providers configuration, with endpoints created accordingly
- ✅ Authorization code flow with a login endpoint in the backend
- ❌ ~Logout from OIDC/OAuth providers~ => abandonned for now
- ✅ UI part with "sign in with" section

### What will be done in future PR(s):
- Display the right logo alongside each provider in the signin page
- Oauth providers should collect more efficiently user information from token + user infos
- Sync user in DB when login
- Device code flow for interactive flow (👤) through the CLI (implementation + documentation)
- Client Credentials flow for non interactive flow ( :robot: ) through the CLI (implementation + documentation)

# Screenshots
image

| image | image |
|---|---|
| image | image|

image

# Checklist

## UI Changes

  include:


scripts/api_backend_dev.sh
```diff
@@ -12,7 +12,8 @@ if [[ $1 == "--e2e" ]]; then
   previous_file="./dev/config.previous.yaml"
   config_file="./dev/config.yaml"
   cp ${config_file} ${previous_file}
-  sed 's/enable_auth: true/enable_auth: false/g' ${previous_file} >${config_file}
+  #TODO(OIDC): we need to reproduce old "enable_auth: false" by replacing providers value by [] ?
+  #sed 's/enable_auth: true/enable_auth: false/g' ${previous_file} >${config_file}
```
Nexucis, line 16: I think it's simpler to keep the attribute `enable_auth` since it's widely used to activate or deactivate things across the API and the UI.

dev/config-mariadb.yaml
```diff
@@ -12,7 +11,10 @@ security:
           - create
         scopes:
           - Project
-
+  authentication:
+    providers:
+      - kind: "database"
```
Nexucis, line 16: I think `native` would require less explanation. Specially when you are reading the config like that: here my list of auth providers: `database`.

That doesn't sound a great name for an auth provider :)

internal/api/impl/auth/builtin.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/perses/perses/pkg/model/api"
+	"github.com/sirupsen/logrus"
+)
+
+type builtinEndpoint struct {
+	dao user.DAO
+	jwt crypto.JWT
+}
+
+func NewBuiltinEndpoint(dao user.DAO, jwt crypto.JWT) Endpoint {
```
Nexucis, line 20: Based on what is implemented for the moment, there isno need to expose this function outside of the package

internal/api/impl/auth/builtin.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+
+	return ctx.JSON(http.StatusOK, api.AuthResponse{
+		AccessToken:  accessToken,
+		RefreshToken: refreshToken,
+	})
+}
+
+func (e *builtinEndpoint) accessToken(ctx echo.Context, login string) (string, error) {
```
Nexucis, line 61: since `accessToken` and `refreshToken` are common in all implementation, it would worth to create a common struct that hold this implementation.

Like


```golang
type tokenManagement struct {
  jwt crypto.JWT
}

func (e *tokenManagement) accessToken(ctx echo.Context, login string) (string, error) {
	accessToken, err := e.jwt.SignedAccessToken(login)
	if err != nil {
		logrus.WithError(err).Errorf("unable to generate the access token")
		return "", shared.InternalError
	}
	jwtHeaderPayloadCookie, signatureCookie := e.jwt.CreateAccessTokenCookie(accessToken)
	ctx.SetCookie(jwtHeaderPayloadCookie)
	ctx.SetCookie(signatureCookie)
	return accessToken, nil
}

func (e *tokenManagement) refreshToken(ctx echo.Context, login string) (string, error) {
	refreshToken, err := e.jwt.SignedRefreshToken(login)
	if err != nil {
		logrus.WithError(err).Errorf("unable to generate the refresh token")
		return "", shared.InternalError
	}
	ctx.SetCookie(e.jwt.CreateRefreshTokenCookie(refreshToken))
	return refreshToken, nil
}
```

and then in the different endpoint

```golang
type builtinEndpoint struct {
	dao user.DAO
	jwt crypto.JWT
        tokenManagement *tokenManagement
}

func (e *builtinEndpoint) auth(ctx echo.Context) error {
	body := &api.Auth{}
	if err := ctx.Bind(body); err != nil {
		return shared.HandleBadRequestError(err.Error())
	}
	usr, err := e.dao.Get(body.Login)
	if err != nil {
		if databaseModel.IsKeyNotFound(err) {
			return shared.HandleBadRequestError("wrong login or password ")
		}
	}

	if !crypto.ComparePasswords(usr.Spec.Password, body.Password) {
		return shared.HandleBadRequestError("wrong login or password ")
	}
	accessToken, err := e.tokenManagement.accessToken(ctx, body.Login)
	if err != nil {
		return err
	}
	refreshToken, err := e.tokenManagement.refreshToken(ctx, body.Login)
	if err != nil {
		return err
	}

	return ctx.JSON(http.StatusOK, api.AuthResponse{
		AccessToken:  accessToken,
		RefreshToken: refreshToken,
	})
}
```

internal/api/impl/auth/auth.go
```diff
@@ -15,70 +15,55 @@ package auth
...
 	"github.com/perses/perses/pkg/model/api"
+	"github.com/perses/perses/pkg/model/api/config"
 	"github.com/sirupsen/logrus"
 )
 
-type Endpoint struct {
-	dao          user.DAO
-	jwt          crypto.JWT
-	isAuthEnable bool
+type Endpoint interface {
```
Nexucis, line 30: mmm a bit sad to redefine the interface just for that.

 I will propose something to make it simpler.

scripts/api_backend_dev.sh
celian-garcia: I agree

ui/app/src/model/config-client.ts
```diff
@@ -86,17 +86,21 @@ export interface AuthorizationConfig {
   guest_permissions: Permission[];
 }
 
+export interface AuthenticationProvider {
+  slug_id: string;
```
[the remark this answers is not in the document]
celian-garcia: Sure I will put more later :stuck_out_tongue:

dev/config-mariadb.yaml
celian-garcia: ok for native :)

internal/api/impl/auth/builtin.go
celian-garcia: Sure the thing is the accessToken/refreshToken is managed by the OIDC library with good old http writer. No problem the duplication has already been identified, I will work on that later

internal/api/impl/auth/auth.go
celian-garcia: I can't wait to see it merged thanks  :)

ui/app/src/model/config-client.ts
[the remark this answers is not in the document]
celian-garcia: Aligned with the exposed one, let's close it now and see if we expose more params and/or reshuffle the way params are deserialized and exposed (serialized) by the backend.

internal/api/impl/auth/builtin.go
celian-garcia: Splitted the token management in two steps : the creation of token/cookies (framework independant) and then the set of the cookies (framework dependent so duplicated for now)

pkg/model/api/config/authentication.go
```diff
@@ -0,0 +1,206 @@
...
+	Kind   AuthKind `json:"kind"`
+	Params any      `json:"params"`
+}
+
+type jsonAuthenticationConfig struct {
+	AccessTokenTTL  string             `json:"access_token_ttl,omitempty"`
+	RefreshTokenTTL string             `json:"refresh_token_ttl,omitempty"`
+	DisableSignUp   bool               `json:"disable_sign_up"`
+	Providers       []jsonAuthProvider `json:"providers,omitempty"`
+}
```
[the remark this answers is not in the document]
celian-garcia: Great thanks I will do that you save me some headaches :)

ui/app/src/views/auth/SignWrapper.tsx
```diff
@@ -11,33 +11,72 @@
...
+            <PersesLogoCropped />
+          ) : isDarkModeEnabled ? (
+            <DarkThemePersesLogo />
+          ) : (
+            <LightThemePersesLogo />
+          )}
+          <Divider orientation={isLaptopSize ? 'vertical' : 'horizontal'} variant="middle" flexItem />
+          <Stack gap={1}>{props.children}</Stack>
+        </Stack>
+        <Divider orientation={'horizontal'} variant="fullWidth" flexItem>
```
Nexucis, line 54: I think it would be nice if this part is kept on the right side of the logo like the native auth.

It would avoid a weird moment when the native auth is not activated and you have nothing on the right.

Nexucis: by the way you need to consider when the native auth is not enabled here.

Nexucis: on a different part, would be cool to display the logo of each provider. Maybe we could add an attribute that would give the type of the provider. And we could support a list of provider logo (like GitHub, GitLab, Microsoft, Google, AWS as a starting base). WDYT ?

internal/api/impl/auth/oidc.go
```diff
@@ -0,0 +1,156 @@
...
+	relyingParty    rp.RelyingParty
+	jwt             crypto.JWT
+	tokenManagement tokenManagement
+	slugID          string
+	urlParams       map[string]string
+}
+
+func newOIDCEndpoint(provider config.OIDCProvider, jwt crypto.JWT) (authEndpoint, error) {
+	// As the cookie is used only at login time, we don't need a persistent value here.
+	key := securecookie.GenerateRandomKey(16)
```
Nexucis, line 47: Could you elaborate on that please? Can you explain why we need to generate a cookie for the authentication ?

internal/api/impl/auth/token.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (tm *tokenManagement) accessToken(login string, provider string) (string, []*http.Cookie, error) {
+	accessToken, err := tm.jwt.SignedAccessToken(login, provider)
+	if err != nil {
+		logrus.WithError(err).Errorf("unable to generate the access token")
+		return "", nil, shared.InternalError
+	}
+	jwtHeaderPayloadCookie, signatureCookie := tm.jwt.CreateAccessTokenCookie(accessToken)
+
+	return accessToken, []*http.Cookie{jwtHeaderPayloadCookie, signatureCookie}, nil
```
Nexucis, line 36: based on the usage you are doing with these methods, probably you can keep the set of the Cookies in this method. It would avoid to do it everytime we are contacting these methods

internal/api/impl/auth/oauth.go
```diff
@@ -0,0 +1,128 @@
...
+	resp, err := e.conf.Client(ctx.Request().Context(), token).Get(e.userInfoURL)
+	if err != nil {
+		return err
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	var userInfos struct {
```
Nexucis, line 103: probably we can declare this struct outside of the method. It would avoid to redeclare it everytime this method is called

internal/api/impl/auth/oidc.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (e *oIDCEndpoint) CollectRoutes(g *route.Group) {
+	g.GET(fmt.Sprintf("/%s/%s", e.String(), loginPath), e.buildAuthHandler(), true)
+	g.GET(fmt.Sprintf("/%s/%s", e.String(), callbackPath), e.buildCodeExchangeHandler(), true)
+	g.GET(fmt.Sprintf("/%s/%s", e.String(), logoutPath), e.logout, true)
+}
+
+func (e *oIDCEndpoint) String() string {
+	return strings.Join([]string{"oidc", e.slugID}, "-")
```
Nexucis, line 82: I think I would prefer to have the route `/auth/providers/oauth/github` instead of `/auth/providers/oauth-github`.

Probably it can removed the usage of the method `String()` and the needs of having a new interface `authEndpoint` since you can define the route like that:

```go
func (e *oIDCEndpoint) CollectRoutes(g *route.Group) {
	oauthGroup = g.Group("/oauth")
	oauthGroup.GET(fmt.Sprintf("/%s", e.slugID, loginPath), e.buildAuthHandler(), false)
	oauthGroup.GET(fmt.Sprintf("/%s", e.slugID, callbackPath), e.buildCodeExchangeHandler(), true)
	oauthGroup.GET(fmt.Sprintf("/%s", e.slugID, logoutPath), e.logout, true)
```

internal/api/impl/auth/oauth.go
```diff
@@ -0,0 +1,128 @@
...
+		conf:            conf,
+		jwt:             jwt,
+		tokenManagement: tokenManagement{jwt: jwt},
+		slugID:          params.SlugID,
+		userInfoURL:     params.UserInfosURL,
+	}
+}
+
+func (e *oAuthEndpoint) CollectRoutes(g *route.Group) {
+	g.GET(fmt.Sprintf("/%s/%s", e.String(), loginPath), e.authHandler, true)
```
Nexucis, line 60: same suggestion than for OIDC

```suggestion
g.GET(fmt.Sprintf("/%s/%s", e.String(), loginPath), e.authHandler, false)
```

the login endpoint shouldn't require a JWT token to use it I think

ui/app/src/views/auth/SignWrapper.tsx
celian-garcia: Yes I didn't focus yet on the login page. I need to make sure it adapt to every situation (native+external, only native, only external) + also check the responsive which I guess is not working.

I agree that we should have logo indeed. We need to have all the known logos in the assets and somehow knows which one to use based on the config. I would wish to do that without adding yet another config line parsing urls for example but I'm not sure. Let me check what others do.

internal/api/impl/auth/token.go
celian-garcia: The cookie set is not always the same. The OIDC library force me to use the http writer, in the classic echo endpoints, we use the ctx.

pkg/model/api/config/authentication.go
```diff
@@ -0,0 +1,81 @@
...
+	ClientSecret config.Secret `json:"client_secret" yaml:"client_secret"`
+	RedirectURI  string        `json:"redirect_uri" yaml:"redirect_uri"`
+	Scopes       []string      `json:"scopes" yaml:"scopes"`
+	AuthURL      string        `json:"auth_url" yaml:"auth_url"`
+	TokenURL     string        `json:"token_url" yaml:"token_url"`
+	LogoutURL    string        `json:"logout_url" yaml:"logout_url"`
+	UserInfosURL string        `json:"user_infos_url" yaml:"user_infos_url"`
+}
+
+type AuthProviders struct {
```
Nexucis, line 52: based on what I saw in the code, it seems `slug_id` must be unique for the same kind of provider.

Would you mind to implement the method `Verify` for the `AuthProviders` to adresse that please ?

internal/api/impl/auth/oauth.go
celian-garcia: yes I will

Note also that this all "login discovery" needs to be enriched a lot, as it's not necessarily a standard to have login in the user infos payload. I'm pretty sure that we don't support a lot more than Github provider with that... 
I was not sure it's the right way but this is also what Grafana is doing, checking it in several places, the token, user infos answer.. I don't like it that much but I don't have better option. Except having an oauth provider type dedicated by provider, which is not better

internal/api/impl/auth/oidc.go
```diff
@@ -0,0 +1,156 @@
...
+		relyingParty:    relyingParty,
+		jwt:             jwt,
+		tokenManagement: tokenManagement{jwt: jwt},
+		slugID:          provider.SlugID,
+		urlParams:       provider.URLParams,
+	}, nil
+}
+
+func (e *oIDCEndpoint) CollectRoutes(g *route.Group) {
+	g.GET(fmt.Sprintf("/%s/%s", e.String(), loginPath), e.buildAuthHandler(), true)
```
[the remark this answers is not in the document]
celian-garcia: mmh I don't understand this one, the /login endpoint has to be anonymous otherwise we cannot generate any token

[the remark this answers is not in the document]
Nexucis: aaah sorry I thought it was the opposite. I thought `true` was saying the endpoint was authenticated.
My bad

internal/api/impl/auth/oauth.go
celian-garcia: nono the login endpoint will trigger the authorization code flow that will redirect to the callback (code exchange) endpoint that will generate the token. so tehse two ednpoints should be anonymous

pkg/model/api/config/authentication.go
celian-garcia: Ok I will

internal/api/impl/auth/oidc.go
celian-garcia: I don't have much more details, if any cookie is generated by the lib here, we are not using it directly and setting random key doesn't change anything... 

Oh maybe I'm configuring a cookie handler for nothing here, let me check

celian-garcia: :smile: I removed the line and it still work... nevermind we'll remove it

celian-garcia: okok I checked twice and no it is used during the authorization code flow, then removed. So this is what I said we don´t need it persistent.
Before redirect to authorize provider's endpoint, the OIDC lib save the state and a pkce code challlenge into cookies, this is then reused to verify the state and code received into our code exchange callback, and removed. 

I will elaborate into the comment itself.

internal/api/impl/auth/oidc.go
celian-garcia: I simplified a lot. Iremoved the provider claim from the token and I removed the logout by provider. Let's do this like that for now and we'll see in the future if want to put some mechanism to ensure people are also logged out from providers

internal/api/impl/auth/token.go
celian-garcia: I introduced a setCookie function parameter to save some lines of code

pkg/model/api/config/authentication.go
celian-garcia: Done and tested :)

ui/app/src/views/auth/SignWrapper.tsx
celian-garcia: I updated the sign in with a much simpler aproach. As you suggested all the sign in together at the right side. Code is simpler thanks to that. 

Regarding the logo. I propose you to add in a subsequent PR.

Nexucis (approved):
This is so cool. Thank you for taking the time to do that !!

I just have one question: do you know if it will be possible to add e2e test on oauth / oidc ?

celian-garcia:
> This is so cool ! Thank you for taking the time to do that !!
> 
> I just have one question: do you know if it will be possible to add e2e test on oauth / oidc ?

Thank you! That's not an easy one for something appearing to be so standard on paper but at the end which become a nightmare as nobody respect the standard lol.

The e2e tests the great question yeah! I didn't invest yet on that but I think it's possible, we could do something running a fake provider server on a different port but I have to check if such a test helper already exist, in the library I used (oidc and oauth). That would facilitate a lot the build of this test.

PR #1639, opened 2023-12-11
[IGNORE] move again the config into the pkg model

Nexucis:
made a mistake when commenting #1636 where the config should go from my opinion. This PR is fixing that.


jgbernalp (approved):
LGTM

PR #1641, opened 2023-12-12
[IGNORE] move custom route and endpoint interface to a dedicated package

Nexucis:
I'm just moving our custom route to a dedicated package. Should help for #1637


celian-garcia (approved):
Lgtm thanks

PR #1644, opened 2023-12-13
[ENHANCEMENT] Move the proxy endpoint from a middleware to a route

sinkingpoint:
Adding this in as a first pass. Manual testing seems to work, but I'm looking at how we can unit test it.

# Description

This moves the proxy endpoint from a midleware, to an actual route. This allows us to use echos routing system to remove a lot of the regex logic and allows us to apply the auth system to the proxy endpoint.

# Checklist


sinkingpoint:
Any idea what issues you were hitting with this @Nexucis ? From my manual testing this seems to be working

Nexucis:
> Any idea what issues you were hitting with this @Nexucis ? From my manual testing this seems to be working

I don't reminder and I can't find if I created a PR about it ... So probably I did it wrong. I just remembered the redirection to the actual datasource didn't work. I'm wondering if I was trying to make an actual redirection (using a 3xx HTTP code)

internal/api/impl/proxy/proxy.go
```diff
@@ -101,33 +47,26 @@ type Proxy struct {
...
 }
 
-func (e *Proxy) proxyGlobalDatasource(ctx echo.Context) error {
-	dtsName, path, err := extractGlobalDatasourceAndPath(ctx.Request().URL.Path)
-	if err != nil {
-		return err
-	}
+func (e *endpoint) CollectRoutes(g *route.Group) {
+	g.ANY(fmt.Sprintf("/%s/:%s/*", utils.PathGlobalDatasource, utils.ParamName), e.proxyGlobalDatasource, true)
+	g.ANY(fmt.Sprintf("/%s/:%s/%s/:%s/*", utils.PathProject, utils.ParamProject, utils.PathDatasource, utils.ParamName), e.proxyProjectDatasource, true)
```
Nexucis, line 63: probably at some point we would like to protect these routes with the permission `read` on the datasource.

But can be down later as it is not the case for the moment

Nexucis (approved):
Awesome that it is working!
It's definitely a really cool improvement and remove a lot of "home made" code to get the proper route !

PR #1658, opened 2023-12-18
[ENHANCEMENT] Use mode.Duration from Prometheus

Nexucis:
Today it's not possible to unmarshal the config in JSON there is an error with the duration.
Introducing the type `mode.Duration` from Prometheus is helping to fix this part. It also helps to simplify the code by removing every `MarshalJSON` method overriding.


celian-garcia (approved):
Great ! I'll have some conflicts but this is really worth it

``time.Duration`` should work actually for unmarshalling if you use <1d duration like 3h etc, if you want to use d notation (1d, 2d, ...) we need model.Duration indeed.

I had some troubles with this ``MarshalJSON`` that works with non-pointer receivers but not with pointer receivers. Do you know why? Is it just because the field itself is not a pointer in the parent struct? I'm asking because I'm afraid that I'll need this to expose the authentication provider.

Nexucis:
> I had some troubles with this `MarshalJSON` that works with non-pointer receivers but not with pointer receivers. Do you know why? Is it just because the field itself is not a pointer in the parent struct? I'm asking because I'm afraid that I'll need this to expose the authentication provider. ref: https://github.com/perses/perses/pull/1637/files#diff-4ffce29bd915be7e7a5de3ff7f3e0baf1407ae34ebd372e17574fe3dea86d4a2R193-R206

Nice thanks! To mitigate your issue, I'm proposing another way to manage the config: https://github.com/perses/perses/pull/1637/files#r1430462873

And regarding your question, yes I think it's because of that.

PR #1664, opened 2023-12-19
[WIP] 1rst step towards a CUE-based Dashboard-as-Code framework 

AntoineThebaud:
# Description

- [IN PROGRESS] CUE utilities for DaC.
  - current panel builder is actually specific to prometheus, 'should be "extracted" or renamed
- [TODO] new github action to ensure cue libs are always valid (should run cue get go etc..)
- [TODO] a bit of examples & other user docs 

# Checklist


scripts/dac_setup.sh
```diff
@@ -0,0 +1,59 @@
...
+Usage: $(basename "$0") [-h] [-v VERSION]
+This script takes care of adding the CUE sources from Perses as external dependencies to your DaC repo.
+/!\ It must be executed at the root of your repo.
+
+Options:
+    -h  show this help text
+    -v  version of Perses (in the form of \"X.Y.Z\") from which to retrieve the dependencies
+
+Example: 
+$ ./dac_setup.sh -v 0.42.1"
```
Nexucis, line 14: to ease the cross platform support, why not including that in the CLI ? 

like under the command `percli dac init --language=cue`

scripts/dac_setup.sh
```diff
@@ -0,0 +1,59 @@
...
+    tar -xzf sources.tar.gz
+
+    # Copy the dependencies to the destination folder
+    cd perses-$VERSION/
+    # - plugin schemas
+    cp -r --parents cue/schemas ../cue.mod/pkg/github.com/perses/perses/
+    # - DaC utilities
+    cp -r --parents cue/dac-utils ../cue.mod/pkg/github.com/perses/perses/
+    # - CUE definitions patches
+    find "pkg/model/api/v1/" -name "*.cue" -exec cp --parents {} "../cue.mod/pkg/github.com/perses/perses/" \;
```
Nexucis, line 33: I'm wondering if it wouldn't make sense to have all cue definitions in a single place instead of having them in the middle of the go package.

I'm just thinking in case one day we are providing a Java pkg, the classes won't be in the middle of the go package 

Would make more sense since you move the `schemas` folder under the `cue` folder

docs/user-guides/dashboard-as-code.md
```diff
@@ -0,0 +1,89 @@
...
+- automated deployments
+- and more..
+
+## Getting started
+
+### Prerequisites
+
+- `percli`, the [CLI of Perses](../tooling/cli.md).
+- `cue`, the [CLI of Cuelang](https://cuelang.org/).
+- [Go](https://go.dev/).
```
Nexucis, line 26: that's a weird things to require `Go` when you are providing a `cue` lib, don't you think ?
I don't even really get why do we need Go here

.gitignore
```diff
@@ -27,8 +27,8 @@ coverage.txt
 
 EXTRACTED_CHANGELOG.md
 
-cue.mod/pkg
-cue.mod/usr
+cue.mod/gen
+/pkg/model/api/v1/**/*_go_gen.cue
```
Nexucis, line 31: I think we are saving these files instead of ignoring them, then it will ease the DAC setup because you won't need to import the go definition to generate the cue definition.

cue/dac-utils/panel-group.cue
```diff
@@ -0,0 +1,57 @@
...
+#width: math.Trunc(#gridCols / #cols)
+#panelsX: [for i, _ in #panelsAsList {
+    #width * math.Round(math.Mod(i, #cols))
+}]
+#panelsY: [for i, _ in #panelsAsList {
+    #height * math.Trunc(i / #cols)
+}]
+
+// output: the final layout, in the format expected by the Perses dashboard.
+v1Dashboard.#Layout & {
```
Nexucis, line 44: I think a docs on each utilities method/function you are providing would help a lot to have more inside about what we are providing.

Probably a doc for each files and a sort of scenario to gives a context why we are providing such utilities method/function

Nexucis (commented):
Amazing work already !

I think we can simplify the setup by saving in the perses repo the cue files generated and by adding a `dac` subcommand in the CLI !

scripts/dac_setup.sh
AntoineThebaud: Yes since we already planned to add such `dac` subcommand in the CLI for other stuff, I was thinking of something like this but for a next PR, but yes we can already do this move if you want

scripts/dac_setup.sh
AntoineThebaud: This actually comes from the fact to use the `--local` flag on `cue get go` command to generate the base CUE def. I did this otherwise it's not possible to do DaC within the perses repo itself (I can explain live, would be hard by writing :D). Then to apply these patches seamlessly (= seamless merge with the generated files) they have to be located in the same place as the generated files.
I know this is not ideal, maybe we can think of intermediary files to invoke somehow.. need to think about it.

docs/user-guides/dashboard-as-code.md
AntoineThebaud: Needed for dep management (you need to `go get` before being able to `cue get go`) https://github.com/perses/perses/pull/1664/files#diff-6f12614d0527f3d0486999783e4b2656892df91b600ae4017eae65c58a7829d9R48

Hopefully this is something that will vanish when CUE will release its dep management?

.gitignore
AntoineThebaud: Ok yeah I suppose this is a valid workaround, but ideally we should rely on this CUE feature + not commit files generated in a deterministic way.

cue/dac-utils/panel-group.cue
AntoineThebaud: For such lib doc useful to developers, I prefer to provide "embedded" doc (ideally jsdoc style that could be rendered in a webpage) rather than external markdown files, that have a higher risk to later get desynchronized with the actual code imho, hence why I put comments in each file & a complete example rather than documented these in the user guide. Still yes for now it's only few lines of text and no fancy jsdoc-style content so I'll try to enrich this part.

AntoineThebaud: After discussing this live, 'agreed that we could have the best of both worlds here: dev doc in a jsdoc way that could be used to generate markdown files, to be integrated in the perses website. That's work for a future PR though :D

scripts/dac_setup.sh
AntoineThebaud: generated files & their patches are now available at from the `cue` folder like the rest of the CUE sources

.gitignore
AntoineThebaud: After discussion, agreed to commit these files & do some operations to have them in a dedicated folder

docs/user-guides/dashboard-as-code.md
AntoineThebaud: Go prereq removed since we now commit the generated file (DaC users can retrieve these ones directly)

pkg/model/api/v1/metadata.cue
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+// This definition provides placeholder values for the dashboard metadata,
+// required to pass the CUE evaluation, as those attributes are flagged as
+// mandatory in the (Go) datamodel but populated by the server in the end.
+#Metadata: {
```
Nexucis, line 19: I guess that's something that is not used anymore

Nexucis (approved):
so cool, that's a really nice start !!

As a side note I think you need to re-generate the cue files with the new URL type we added in the pkg `/model/api/v1/common`

pkg/model/api/v1/metadata.cue
AntoineThebaud: Oups indeed I forgot to move that file, nice catch

PR #1665, opened 2023-12-19
[IGNORE] Add authentication details to backend dev script

sinkingpoint:
# Description

Just submitting this as a quick proposal.

We've had a couple of cases in the chat where folks have been confused about how to log in to the backend dev server. This adds a echo to the script that tells them how to do it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,7 +126,7 @@ The secure way to configure your monitoring.               <\
 
 ```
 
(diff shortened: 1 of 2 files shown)


Gladorme (approved):

Nexucis:
Ah yeah good idea.
Would you mind to add this info also in https://github.com/perses/perses/blob/main/CONTRIBUTING.md#detailed-version ?

Nexucis (approved):

PR #1668, opened 2023-12-20
[FEATURE] CLI: add a new command  to refresh the access token

Nexucis:
when the authentication is enabled, it's quite annoying to login every 15min. 

Since we have a refresh token mecanism, I'm proposing a new command `refresh` that will contact the API to get a new access token

That's a beginning, perhaps for long term we want an automatic way to refresh the token like the UI has.


celian-garcia (approved):
As discussed in private, hell yes if it's reduce pain. But the target should be to have something a bit more automatic like on each command if we receive a 401 but have a refresh token let's try to refresh under the hood before redoing the command

PR #1672, opened 2023-12-21
[FEATURE] Initial ScatterChart Panel Plugin

zhuje:
# Description

## TLDR: Initial ScatterChart Panel Plugin

- Add Panel Plugin for Scatter Chart
    - Scatter Chart only handles Trace data for now
    - Scatter Chart currently doesn't offer any spec options to format chart (e.g., y-axis label, min value, max value)
- Add Tempo Query Editor for query editing in UI
    - There is no auto-complete extension in CodeMirror for Tempo

## Note

ScatterChart Panel Plugin is based on the work done by Steve Cobb from [[perses-embed-example](https://github.com/sjcobb/perses-embed-example)](https://github.com/sjcobb/perses-embed-example).

## Backend Changes

Add schemas for scatter plot panels in `schemas/panels/scatterplot` . In this initial iteration, no `spec` options are added to this panel.

## Frontend Changes

**1. Add scatter plot panel for data visualization**

- Files Changes in perses/ui/panels-plugin/src/scatterplot
    - **ScatterChart.ts** — define a PanelPlugin Chart. The PluginRegistry uses this file to fetch the component dynamically (see the file PanelContent.tsx and the usePlugin() function).
    - **ScatterChartPanel.tsx** — define a PanelComponent, transforms the Tempo API data into a ‘dataset’ that can be consumed by Apache ECharts. Pass dataset and additional visualization options as props to <Scatterplot> component.
        - *DataSet*
            - Documentation for data structures accepted by Apache ECharts: https://echarts.apache.org/handbook/en/concepts/dataset
        - *Formatting*
            - Datapoint size (bubble size). A single data point is a trace. The number of spans in a trace dictates the data point size.
            - Data point color. If a trace contains an error, it will color the datapoint in red.
            - Examples for chart formatting in Apache ECharts: https://echarts.apache.org/examples/en/index.html#chart-type-scatter
    - **ScatterChartPanel.test.tsx** — test if ScatterChartPanel can render correctly
    - **Scatterplot.tsx** - accept props from Scatterplot and format the Apache EChart with the props.
    - **scatter-chart-model.ts** - define models for components in ScatterChart.ts

**1B. Component TreeView**

- ScatterChart
    - ScatterChartPanel
        - ScatterPlot

**2. Changes to add a UI panel editing in a scatter plot with Tempo Data**
Note: These changes were based on Time Series Query Editor. In the future, if additional data types are added, we should consider how to create more reusable components (i.e., if we can make a general query editor that is less tightly coupled to a specific query type).

- File Changes in perses/ui/plugin-system/src/components
    - **PanelSpecEditor.tsx** —  modified to include a switch statement to render the correct query editor based on the dashboard definition spec > query kind (e.g., TimeSeriesQuery, TraceQuery). For example : `switch(queryType) >> case ‘TraceQuery’ >> render <TraceQueryEditor>`
    - **TraceQueryEditor.tsx** — holds the logic for handling add/removing queries in trace query editor and passes these as props to TraceQueryInput
    - **TraceQueryInput.tsx**— Renders collapsing and delete buttons. Contains the PluginEditor component, which uses the Panel’s query type (i.e., TraceQuery) to render the corresponding TraceQuery editing components
        - **PluginEditor.tsx**
            - **PluginKindSelect.tsx** — renders a drop-down containing kinds for our query type. For example, `<PluginKind queryType={TraceQuery}/>` drop-down will render  TempoTraceQuery.
            - **PluginSpecEditor.tsx** — dynamically renders the corresponding OptionsEditorComponent given a query type. For example, `<PluginSpecEditor queryType={TraceQuery}/>` will fetch the query plugin from the file TempoTraceQuery.tsx and get the `OptionsEditorComponent: TempoTraceQueryEditor`
- File Changes in perses/ui/tempo-plugin/src/tempo-trace-query
    - **TempoTraceQuery.tsx** — define OptionsEditorComponent: TempoTraceQueryEditor in the query plugin
    - **TempoTraceQueryEditor.tsx** — handles data source change, contains <DashboardTraceQueryEditor>
    - **DashboardTraceQueryEditor.tsx** — contains <DatasourceSelect> and <TraceQLEditor>
    - **TraceQLEditor.tsx**— embed <CodeMirror>, which is a text input for a query. Currently, there is no autocomplete extension for Tempo.

**3. Enable TimeRangeSelector on TraceQuery**

- File changes for /perses/ui/plugin-system/src/runtime
    - **trace-queries.tsx** - fetch selected time range (e.g. ‘last 1 hour, last 1 day’) and pass it as  a prop to the trace fetching function getTraceData() in the get-trace-data.tsx file
- File changes for perses/ui/tempo-plugin/src/tempo-trace-query
    - **get-trace-data.tsx** — contains getTraceData(). Modify the query to include a ‘start’ and ‘end’ time. For example `{fooQuery}&start={unixStartTime}&end{unixEndTime}`. Pass this query to `client.getEnrichedTraceQuery(getQuery(), datasourceUrl)`  to  fetch trace data from API.

# Screenshots

*Figure 1. Scatter Chart*
Datapoint size corresponds to the number of spans within a trace.
Red datapoints indicate the trace contains and error.
Perses_Scatterplot_PR_Screenshot_ScatterChart_Dec_20_2023
*Figure 2. Trace Query Editor*
Perses_Scatterplot_PR_Screenshot_TraceQueryEditor_Dec_20_2023

# Checklist

- [x]  Pull request has a descriptive title and context useful to a reviewer.
- [x]  Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
- [x]  All commits have [[DCO signoffs](https://github.com/probot/dco#how-it-works)](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x]  Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x]  Code follows the [[UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md)](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [x]  Visual tests are stable and unlikely to be flaky.
See [[Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests)](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests)
and [[e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests)](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues
include:


ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const containsTempoQuery = () => {
+    let isTempoQuery = false;
+    queries?.forEach((query) => {
+      if (query.spec.plugin.kind === 'TempoTraceQuery') {
+        isTempoQuery = true;
+      }
+    });
+    return isTempoQuery;
+  };
+  const isGreaterThanSevenDays = (): boolean => {
```
sjcobb, line 109: This should probably be handled in the query plugin similar to how there is Prometheus specific logic here: https://github.com/perses/perses/blob/main/ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts

docs/plugin/panel.md
```diff
@@ -51,6 +51,14 @@ spec:
   [ width: <int> ]
 ```
 
+## ScatterChart
+Note: Currently, ther are are no spec options avaiable for this panel
+
+```yaml
+kind: "ScatterChart"
```
sjcobb, line 58: Very cool PR! Do you see this evolving to support TimeSeries queries and possibly logs in the future? If not, we could potentially lean into this being a plugin that is focused on visualizing traces and rename it to something like TraceViewer (TraceChart sounds a little weird)

ui/plugin-system/src/components/TraceQueryEditor/TraceQueryEditor.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  // show one query input if queries is empty
+  const queryDefinitions: TraceQueryDefinition[] = queries.length
+    ? queries
+    : [
+        {
+          kind: 'TraceQuery',
+          spec: {
+            plugin: {
+              kind: 'TempoTraceQuery',
```
sjcobb, line 105: We might need a new package for the Tempo-specific parts of this that are in plugin-system, similar to how we have prometheus-plugin to encapsulate its query plugin, variables, datasource / client, editor, etc.

sjcobb: @Nexucis since we don’t have a finished plugin system yet to install / manage third party plugins, should we just make a more generic package like `misc-plugins` for now? Or we could do a tempo-plugin but that probably won’t scale to add new packages like that in ui folder

sjcobb: NVM sorry I missed that we have a tempo-plugin already now 😣

ui/tempo-plugin/src/plugins/tempo-trace-query/get-trace-data.ts
```diff
@@ -0,0 +1,95 @@
...
+      spanCount,
+      errorCount,
+      traceId,
+      name,
+    };
+  });
+
+  const traceData: TraceData = {
+    traces,
+    metadata: {
```
sjcobb, line 89: To my other comment about ‘greater than last 7 days’ not being supported by Tempo logic, we could potentially use `metadata.notices` to convey that to the panel—didn’t finish that implementation yet but the idea there is notices can display in the panel header (info, warning, and error tags with a tooltip to show the full message) 

Doesn’t need to be in this PR, I might write up some thoughts in a separate issue, it’d be useful for TimeSeriesQuery too, for instance if one query throws errors but the others succeed, you could use a notice to not block rendering but still show the error in panel header. Or for performance if we wanted to truncate queries when too many series are returned and use an info tag in panel header so user is aware

docs/plugin/panel.md
zhuje: TBH, I have yet to make plans for TimeSeries and logs support. I wanted to focus on tracing for now. However, it would be a great feature to support multiple data sources in the future.  So, let's try to keep it as 'ScatterChart.'

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
zhuje: Good catch. In the latest commit, I removed this function after discovering that the Tempo API allows time-range queries over seven days. Users can override this default configuration. I double-checked this with one of my colleagues at Red Hat, who works on the backend with Tempo.

ui/tempo-plugin/src/plugins/tempo-trace-query/get-trace-data.ts
zhuje: Interesting! I see there's a beginning iteration of this in `get-time-series-data.ts.` I'll leave the `metadata.notices` param out of this PR for now and wait for the separate issue.

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { ErrorAlert } from '@perses-dev/components';
+import { getUnixTime } from 'date-fns';
+import { Scatterplot } from './Scatterplot';
+import { ScatterChartOptions } from './scatter-chart-model';
+
+export interface EChartTraceValue extends Omit<TraceValue, 'startTimeUnixMs'> {
+  startTime: Date;
+}
+
+export function getUnixTimeRange(timeRange: AbsoluteTimeRange) {
```
sjcobb, line 28: Can't remember exactly, but we might have a similar util in @perses-dev/core

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
```diff
@@ -0,0 +1,165 @@
...
+  if (!traceIsLoading && traceData?.traces.length === 0) {
+    const query = traceData?.metadata?.executedQueryString;
+    return generateErrorAlert(`No traces found for the query : " ${query} " .`);
+  }
+
+  // Formatting for the dataset
+  // 1. Map x,y coordinates
+  // 2. Datapoint size corresponds to the number of spans in a trace
+  // 3. Color datapoint red if the trace contains an error
+  const seriesTemplate: SeriesOption = {
```
sjcobb, line 108: The dataset lines above are in a useMemo so that's good, but these other lines could benefit from being memozied too. It would break the rules of hooks though so having a separate component like ScatterChartPanelLoaded.tsx would be the way to go (other panel types would benefit from that too)

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
```diff
@@ -0,0 +1,165 @@
...
+    },
+    itemStyle: {
+      color: function (params) {
+        const traceData: EChartTraceValue = params.data as EChartTraceValue;
+        // If the trace contains an error, color the datapoint in red
+        if (traceData.errorCount !== undefined && traceData.errorCount > 0) {
+          return 'red';
+        }
+        // Else return default color
+        const defaultColor = '#56B4E9';
```
sjcobb, line 129: chartsTheme is nice to leverage for these types of colors since it allows embed users to customize their palette, fonts, etc. ([ex of defaultColor ](https://github.com/perses/perses/blob/072cb4f8aa8ad6a36a014c490cd357c31b87fdba/ui/panels-plugin/src/plugins/stat-chart/utils/data-transform.ts#L25)for thresholds)

ui/panels-plugin/src/plugins/scatterplot/Scatterplot.tsx
```diff
@@ -0,0 +1,103 @@
...
+export function Scatterplot(props: ScatterplotProps) {
+  const { width, height, options } = props;
+  const chartsTheme = useChartsTheme();
+
+  // Apache EChart Options Docs: https://echarts.apache.org/en/option.html
+  const eChartOptions: EChartsCoreOption = {
+    dataset: options.dataset,
+    series: options.series,
+    grid: {
+      bottom: 40,
```
sjcobb, line 38: Similar comment about possibly leveraging [chartsTheme](https://github.com/perses/perses/blob/072cb4f8aa8ad6a36a014c490cd357c31b87fdba/ui/components/src/utils/theme-gen.ts#L38) if possible

ui/tempo-plugin/src/plugins/tempo-trace-query/query-editor-model.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  // Update our local state's copy as the user types
+  const handleQueryChange = (e: string) => {
+    setQuery(e);
+  };
+
+  // Propagate changes to the query's value when the input is blurred to avoid constantly re-running queries in the
+  // PanelPreview
+  const handleQueryBlur = () => {
```
sjcobb, line 48: Side note: sometimes this onBlur logic annoys me in our Prometheus editor and I just want an explicit 'Run queries' button 😂

sjcobb (approved):
🚀

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
zhuje: I’ve removed this in the latest commit. This was a helper function used by greaterThanSevenDays().

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
zhuje: TBH, I’m a novice with hooks and don’t see how it would break the rules. Could you elaborate on the rules that are being broken and the implementation you’d like to see for the separate component `ScatterChartPanelLoaded`? In the latest push, I attempted to use the useMemo() hook directly in scatterChartPanel to memoize the series formatting, and it didn’t throw any errors…

I am not sure how the [rules](https://react.dev/warnings/invalid-hook-call-warning)](https://react.dev/warnings/invalid-hook-call-warning) are being broken because the memoized function is not inside a condition, loop, or event handler, and there are no other hooks within the function. And I moved the error checks that would return the function early, after the hooks.

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
zhuje: Absolutely, agree. Great catch; this is updated in the latest commit.

ui/tempo-plugin/src/plugins/tempo-trace-query/query-editor-model.ts
zhuje: I agree. This could be generated as a separate issue. Updating the panel to have an explicit 'Run queries' button might also be more intuitive for the user.

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
sjcobb: Oh great! I thought it was after the loading if conditions but I might have misread too, if no errors were thrown or anything then you should be good

ui/panels-plugin/src/plugins/scatterplot/ScatterChartPanel.tsx
sjcobb: Awesome and you can always add new colors too if what’s there isn’t enough

sjcobb (approved):
LGTM!

Nexucis (approved):
incredible work @zhuje thank you so much for taking the time to do it !

And thank you  @sjcobb for reviewing and @zhuje for your patience

Issue #1524, opened 2023-10-25
Custom serie color based on string match


AntoineThebaud:
### Proposal

E.g this is a quite standard use case to have a same graph showing an ok trend vs an error trend. In such case you may want to display the error trend in red.

sjcobb:
Our ‘auto’ palette has logic to automatically [generate series names with error in them to a red hue](https://github.com/perses/perses/blob/20a69c0d8e063bcb193f9f209d4d571c1bbadde2/ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts#L103), imo we should change this back to the default and make sure it still works. I’m also going to pick up #557 since I have branch with that logic started 

We could also allow further customization through ‘color aliases’ based on series name or regex, @AntoineThebaud curious, do you see value in allowing other properties like line width to be overridden too? I think Grafana goes a little overboard with series overrides but certain visual properties and more control over formatting seems important

AntoineThebaud:
I do have an example in mind yes: in our company it's quite common to have, for a given graph, the current trend(s) and a "baseline" computed from past data (we often have a seasonality at week level), and for such trend it's convenient if you can display it with a thiner line, and/or as a dotted line, to highlight the difference with the live data.

Still, I agree that this part of trends customization is quite complex in Grafana, so we should be careful with the UX when adding such options.

AntoineThebaud:
I'm closing this issue since things were implemented since & further capabilities could be needed but should be considered individually plugin by plugin (= eventual new plugin-specific issues to be created):
- As mentioned above, TimeSeriesChart now has [Query Settings](https://github.com/perses/perses/pull/2046) that mitigate a bit the need of the feature requested here.
- For multiple panels we got "Value Mappings" implemented based on a common spec, see https://github.com/perses/perses/issues/2337, that among other things allow to color series based on string match.

PR #1678, opened 2023-12-26
[FEATURE] Palette override when single series returned

### LINKED ISSUE #557: Allow set of a color when you have only one timeseries for query
In the `LineChart`, some of our users would like to be able to set a color only if the query you can configure returned a single timeseries.

In case there is more than one, then the color set in the query is ignored by the panel.

sjcobb:
# Description

When only one timeseries returned, allow setting an explicit color using color picker. When more than one series, fallback to what is set in visual.palette.mode ('auto' is now default).

Proposing a new optional panel option for TimeSeriesChart: `visual.palette.singleSeriesColor`

Closes #557 

### Additional updates 
- Change default color palette to 'auto' (see comment [here](https://github.com/perses/perses/issues/1524#issuecomment-1868190329))
- VisualOptionsEditor refactored to use immer 
- Reordered visual controls in panel editor
  - Open to changing this, I thought it looked better and palette options are probably more important than lineWidth and areaOpacity 
- Refactor useMemo logic in TimeSeriesChartPanel

# Screenshots

New visual controls in TimeSeriesChart panel editor 
image

Color picker with predefined "approved" colors from chartsTheme palette 
image

Yellow set in `visual.palette.singleSeriesColor ` and working fallback palette example
image

Proposed spec
image

# Checklist

## UI Changes

  include:


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -108,27 +107,33 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
       : undefined;
+  }, [props.spec.legend]);
 
   // TODO: add support for y_axis_alt.format
   const format = props.spec.yAxis?.format ?? DEFAULT_FORMAT;
 
   // ensures there are fallbacks for unset properties since most
   // users should not need to customize visual display
-  const visual = merge({}, DEFAULT_VISUAL, props.spec.visual);
+  const visual = useMemo(() => {
```
sjcobb, line 121: Refactored these merge lines so they don't run unnecessarily often. Also removed useDeepMemo since it isn't actually helping and fixed up resulting eslint errors

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -28,30 +28,23 @@ export interface SeriesColorProps {
...
   // Explicit way to always cycle through classical palette instead of changing when based on number of series.
   if (visual.palette?.mode === 'categorical') {
     return getCategoricalPaletteColor(categoricalPalette, seriesIndex, fallbackColor);
   }
 
-  // Decide which palette to use based on total series returned.
-  // When series number exceeds number of colors in palette, use generative colors instead.
-  const paletteLength = Array.isArray(categoricalPalette) ? categoricalPalette.length : 0;
-  const seriesColor =
-    totalSeries <= paletteLength
```
sjcobb, line 51: This changes the default palette behavior to use `palette.mode: 'auto'` which allows for consistent colors for same series names across panels (and the automatic red hue for errors kicks in, see comment [here](https://github.com/perses/perses/issues/1524#issuecomment-1868190329)). Works better with singleSeriesColor too since changing the entire color strategy when the palette ran out of colors led to inconsistent/confusing behavior

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -35,24 +40,38 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
...
 
   const currentStack: StackOptions = value.stack ?? 'none';
   const stackConfig = STACK_CONFIG[currentStack];
 
+  const handleSingleSeriesColorChange = (color: string) => {
+    onChange(
+      produce(value, (draft) => {
+        draft.palette = {
+          ...draft.palette,
+          mode: 'auto',
```
sjcobb, line 68: TODO: Need to remove ‘auto’ here and always fallback to whatever is set in palette.mode 

Planning to fix this and I’ll add a test case for ‘categorical’ as fallback (when singleSeriesColor is set).
Also would love to add visual regression / an e2e for this but not sure I’ll have enough time as part of this PR

schemas/panels/time-series/time-series.cue
```diff
@@ -31,8 +31,8 @@ import (
 }
 
 #palette: {
-	mode: "auto" | "categorical"
-	// colors: [...string]; // TODO: add colors to override ECharts theme
+	mode:               "auto" | "categorical"
+	singleSeriesColor?: string
```
sjcobb, line 35: @Nexucis I’m reading #557 again, sounds like this should be an option on the query instead of how I have it here?

If so, what level would it be, on TimeSeriesQuery or nested inside PrometheusTimeSeriesQuery?

ui/panels-plugin/src/plugins/time-series-chart/SingleSeriesColorPicker.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const closeColorPicker = () => {
+    setAnchorEl(null);
+  };
+
+  const { echartsTheme } = useChartsTheme();
+
+  // Show first four colors from palette as selectable in color picker
+  const predefinedSeriesColors = useMemo(() => {
+    if (Array.isArray(echartsTheme.color) && echartsTheme.color.length > 4) {
```
Gladorme, line 41: I think if there is less than 4 colors, we still want to show them

ui/panels-plugin/src/plugins/time-series-chart/SingleSeriesColorPicker.tsx
```diff
@@ -0,0 +1,86 @@
...
+      return echartsTheme.color.slice(0, 4).map((color: Color) => color.toString()); // account for ZRColor LinearGradientObject edge case
+    }
+  }, [echartsTheme.color]);
+
+  return (
+    <>
+      <ColorIconButton
+        size="small"
+        aria-label={`change single series color`}
```
Gladorme, line 50:
```suggestion
        aria-label="change single series color"
```

ui/panels-plugin/src/plugins/time-series-chart/SingleSeriesColorPicker.tsx
```diff
@@ -0,0 +1,86 @@
...
+  );
+}
+
+const ColorIconButton = styled(IconButton, {
+  shouldForwardProp: (props) => props !== 'isSelected' && props !== 'iconColor',
+})<{
+  iconColor?: string;
+  isSelected?: boolean;
+}>(({ iconColor, isSelected }) => ({
+  backgroundColor: isSelected && iconColor ? `${iconColor}3F` : 'undefined', // 3F represents 25% opacity
```
Gladorme, line 84: Cool trick! Great there is a comment explaining this 🥸

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -82,6 +101,59 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
...
+              );
+            }}
+          >
+            <ToggleButton value="auto">Auto</ToggleButton>
+            <ToggleButton value="categorical">Categorical</ToggleButton>
+          </ToggleButtonGroup>
+        }
+      />
+      <OptionsEditorControl
+        label={'Single Series Color'}
```
Gladorme, line 130:
```suggestion
        label='Single Series Color'
```

ui/panels-plugin/src/plugins/time-series-chart/VisualOptionsEditor.tsx
```diff
@@ -82,6 +101,59 @@ export function VisualOptionsEditor({ value, onChange }: VisualOptionsEditorProp
           </ToggleButtonGroup>
         }
       />
+      <OptionsEditorControl
+        label={'Palette'}
```
Gladorme, line 105:
```suggestion
        label='Palette'
```

Gladorme (commented):
Cool! On the screenshot I see the color picker is aligned on the left, I think it should aligned as others inputs/switchs on the right 😄

schemas/panels/time-series/time-series.cue
celian-garcia: Hello @sjcobb we were just discussing about that with Antoine and saw your comment, so let me jump on it :D

We were talking about a use case that seems to be exactly that: 
In our tool we embed perses panel displaying ALERTS query result. For now we display currently only firing result as we don't have a way to visually separate them. 
Having the possibility to make two queries with let's say red for firing and gray for pending for example would be very interesting for us.

sjcobb: Yeahh like right after I completed this approach, I read Augustin’s ticket again and it finally clicked why on the query would be so much better 😂 

We have a metadata option on TimeSeriesQuery already, I’d propose we add a `config` since styles like color aren’t really metadata. Then I can probably reuse some of this work and make it happen

There are some other updates in this branch though like the useMemo changes so I’ll break those out into a separate PR

sjcobb:
Closing in favor of better UX described [in comment thread here](https://github.com/perses/perses/pull/1678#discussion_r1442975526), should be straightforward to adjust the approach here but I’ll make a new PR when I find time to come back to it

Issue #1542, opened 2023-10-25
"Test connectivity" button in datasource form


AntoineThebaud:
### Proposal

Currently when creating/editing a datasource, you have no idea if the configuration you provided actually works. Thus it would be great to have a simple button to launch a dummy query using the provided config, to validate that the connectivity is OK.

adrianSepiol:
I can work on it.

jgbernalp:
@adrianSepiol thnx for your contribution. I was checking the PR sequence and I wonder if you will be open to take a more pragmatic approach. I think we can reuse the DatasourceClient interface, which already includes a healthCheck function:

```typescript
// @perses-dev/plugin-system
export interface DatasourceClient {
  kind?: string;
  healthCheck?: () => Promise<boolean>;
}
```

This will allow plugins to implement their health check in their datasource clients, some of them already have it. Then, in the HTTPSettingsEditor we can create a client with the current spec and call the function. This will allow us to make less assumptions on how to call a specific URL or which method to use by delegating that to plugins. Also we will avoid creating a new provider as all we need is the plugin client, which can be created from the spec. I think the client can be created lazily as soon as the button is pressed so it uses the current spec values. WDYT? Did I miss something that would avoid us to take this path?

adrianSepiol:
> [@adrianSepiol](https://github.com/adrianSepiol) thnx for your contribution. I was checking the PR sequence and I wonder if you will be open to take a more pragmatic approach. I think we can reuse the DatasourceClient interface, which already includes a healthCheck function:
> 
> // @perses-dev/plugin-system
> export interface DatasourceClient {
>   kind?: string;
>   healthCheck?: () => Promise<boolean>;
> }
> 
> This will allow plugins to implement their health check in their datasource clients, some of them already have it. Then, in the HTTPSettingsEditor we can create a client with the current spec and call the function. This will allow us to make less assumptions on how to call a specific URL or which method to use by delegating that to plugins. Also we will avoid creating a new provider as all we need is the plugin client, which can be created from the spec. I think the client can be created lazily as soon as the button is pressed so it uses the current spec values. WDYT? Did I miss something that would avoid us to take this path?

@jgbernalp 
[...]

PR #1680, opened 2023-12-30
[FEATURE] Add Test button to datasource creation

sinkingpoint:
# Description

This is the UI work for utilising the previously created Unsaved Datasources framework to be able to test datasources before they're saved.

There's a few components to this:

 - Work in the datasource provider to take two different types of selector. A "Saved" selector, which functions as previously - loading the datasource from the API and using that, and an "Unsaved" selector which generates datasource clients from the spec included in the selector.

- This is the one I'm least sure about. A change to the datasource client api to allow us to handle both saved and unsaved datasources. Because unsaved datasources have more complicated semantics than unsaved ones (they need to package the spec, method, and request body into the body of the request), we need to be able to process the requests sent by datasource clients. To do this, we add in another argument - `fetch` to the query options of datasources. For saved datasources, again, this is the same. For unsaved ones, it's a wrapper around `globals.fetch` that allows us to intercept and repackage requests.

- Adding the button and wiring it into the DatasourceStoreProvider

# Screenshots

## Global Datasources

https://github.com/perses/perses/assets/4386405/3ed021a9-3a08-4977-b7e2-7abcb5b17c99

## Project Datasources

https://github.com/perses/perses/assets/4386405/829b740f-d4dd-4dd4-bb13-e46ae16816b7

## Dashboard datasources ( + requests to saved datasources still working)

https://github.com/perses/perses/assets/4386405/83a7a293-6428-479d-b505-5b7b8a25b6dd

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/app/src/model/datasource-api.ts
+++ b/ui/app/src/model/datasource-api.ts
@@ -34,15 +34,7 @@ export class HTTPDatasourceAPI implements DatasourceApi {
    * @param dashboard
    * @param project
    */
-  buildProxyUrl({
-    project,
-    dashboard,
-    name,
-  }: {
-    project?: string;
-    dashboard?: string;
-    name?: string;
-  }): string {
+  buildProxyUrl({ project, dashboard, name }: { project?: string; dashboard?: string; name?: string }): string {
     const unsaved = name === undefined;
     const prefix = unsaved ? '/proxy/unsaved/' : '/proxy/';
     const type = !project && !dashboard ? 'globaldatasources' : 'datasources';
--- a/ui/app/src/model/datasource-api.ts
+++ b/ui/app/src/model/datasource-api.ts
@@ -23,7 +23,10 @@ export class HTTPDatasourceAPI implements DatasourceApi {
    * Give the following output according to the definition or not of the input.
    * - /proxy/globaldatasources/{name}
    * - /proxy/projects/{project}/datasources/{name}
-   * - /proxy/projects/{project}/dashboards/{dashboard}/{name}
+   * - /proxy/projects/{project}/dashboards/{dashboard}/datasources/{name}
+   * - /proxy/unsaved/globaldatasources
+   * - /proxy/unsaved/{project}/datasources
+   * - /proxy/unsaved/{project}/dashboards/{dashboard}/datasources
    *
    * NB: despite the fact it's possible, it is useless to give a dashboard without a project as the url will for sure
    * correspond to nothing.
@@ -31,15 +34,29 @@ export class HTTPDatasourceAPI implements DatasourceApi {
    * @param dashboard
    * @param project
    */
-  buildProxyUrl({ project, dashboard, name }: { project?: string; dashboard?: string; name: string }): string {
-    let url = `${!project && !dashboard ? 'globaldatasources' : 'datasources'}/${encodeURIComponent(name)}`;
+  buildProxyUrl({
+    project,
+    dashboard,
+    name,
+  }: {
+    project?: string;
+    dashboard?: string;
+    name?: string;
+  }): string {
+    const unsaved = name === undefined;
+    const prefix = unsaved ? '/proxy/unsaved/' : '/proxy/';
+    const type = !project && !dashboard ? 'globaldatasources' : 'datasources';
+    let url = unsaved ? `${type}` : `${type}/${encodeURIComponent(name || '')}`;
+
     if (dashboard) {
       url = `dashboards/${encodeURIComponent(dashboard)}/${url}`;
     }
+
     if (project) {
       url = `projects/${encodeURIComponent(project)}/${url}`;
     }
-    return `/proxy/${url}`;
+
+    return `${prefix}${url}`;
   }
 
   getDatasource(project: string, selector: DatasourceSelector): Promise<ProjectDatasource | undefined> {
--- a/ui/dashboards/src/context/DatasourceStoreProvider.tsx
+++ b/ui/dashboards/src/context/DatasourceStoreProvider.tsx
@@ -20,6 +20,7 @@ import {
   DatasourceSpec,
   GlobalDatasource,
   useEvent,
+  UnsavedDatasourceSelector,
 } from '@perses-dev/core';
 import {
   DatasourceStoreContext,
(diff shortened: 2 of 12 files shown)


ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -130,19 +131,75 @@ export function DatasourceStoreProvider(props: DatasourceStoreProviderProps) {
...
-  const getDatasourceClient = useCallback(
-    async function getClient<Client extends DatasourceClient>(selector: DatasourceSelector): Promise<Client> {
+  const getSavedDatasourceClient = useCallback(
+    async (selector: DatasourceSelector) => {
       const { kind } = selector;
       const [{ spec, proxyUrl }, plugin] = await Promise.all([findDatasource(selector), getPlugin('Datasource', kind)]);
 
-      // allows extending client
-      const client = plugin.createClient(spec.plugin.spec, { proxyUrl }) as Client;
+      return plugin.createClient(spec.plugin.spec, { proxyUrl, fetch });
```
sjcobb, line 139: Just skimming the PR, not crazy about passing fetch this way, but agreed this is tricky 🤔 

Is the custom version only needed for healthCheck? If so, I wonder if a similar approach to the onCreate prop could be used, so inside getDatasourceClient if the datasource is unsaved, create a wrapper that defines the custom fetch behavior and return something like: `{ …client, healthCheck: () => customFetchWrapperExample(client.healthCheck, spec) }`

sinkingpoint: I _was_ thinking that we could apply it to things like datasources in dashboard that haven't been saved yet, but honestly if limiting it to just healthchecks makes it cleaner then I'm all for it

sjcobb: Oh that is neat, sorry missed that recording before. I’ll wait for others to weigh in about expected behavior though (also curious what Grafana supports for comparison). If a particular endpoint is misconfigured, does the error message give info about which one?

Wrapper idea probably isn’t relevant if more than healthchecks though—and if it was just healthchecks, it might be possible to just tweak the logic directly in DatasourceEditorForm

Gladorme (commented):
Commented but I didn't understand everything. For example I forgot why we need a custom fetch 😅 So maybe my comments are not relevant

ui/app/src/model/datasource-api.ts
```diff
@@ -23,23 +23,32 @@ export class HTTPDatasourceAPI implements DatasourceApi {
...
+   * - /proxy/unsaved/{project}/dashboards/{dashboard}/datasources
    *
    * NB: despite the fact it's possible, it is useless to give a dashboard without a project as the url will for sure
    * correspond to nothing.
    * @param name
    * @param dashboard
    * @param project
    */
-  buildProxyUrl({ project, dashboard, name }: { project?: string; dashboard?: string; name: string }): string {
-    let url = `${!project && !dashboard ? 'globaldatasources' : 'datasources'}/${encodeURIComponent(name)}`;
```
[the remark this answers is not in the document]
sinkingpoint: Not sure I follow?

Saved datasources look like: `/proxy/globaldatasource/`

Unsaved ones look like: `/proxy/unsaved/globaldatasource`

We can tidy this up a bit though, sure.

sinkingpoint:
> For example I forgot why we need a custom fetch

To perhaps clarify a bit: Requests to saved datasources get proxied as is to the backing datasource, because we already have all the required metadata (datasource url etc) on the server side. This means that a request to a saved datasource is exactly as if you're fetching the datasource directly.

 For _unsaved_ datasources, we don't have that metadata already on the server side. So we need to package the spec up and send it along with the request. Because we _have_ to send a body, this means that we have to use a POST (or at least, not a get), so we have to also include the requested method, and actual request body in the request body. 

In order to make that transparent from the plugins point of view, I have a custom fetch wrapper that checks if we're calling an unsaved datasources and transparently wraps the request in the format that an unsaved datasource requires. That way, plugins can continue to make fetchs as usual and not worry about handling the case of saved vs unsaved. 

Does that sort of make sense?

sinkingpoint:
@sjcobb I note that we already have a fetch wrapper of sorts in https://github.com/perses/perses/blob/main/ui/core/src/utils/fetch.ts - I could piggy back off that? WDYT?

ui/dashboards/src/context/DatasourceStoreProvider.tsx
Nexucis: I think we should limit it to the health-checks for the moment, because creating a datasource directly in the dashboard shouldn't be the usual way. So we could restrict a bit the things you can do with local datasource

Nexucis:
> > For example I forgot why we need a custom fetch
> 
> To perhaps clarify a bit: Requests to saved datasources get proxied as is to the backing datasource, because we already have all the required metadata (datasource url etc) on the server side. This means that a request to a saved datasource is exactly as if you're fetching the datasource directly.
> 
> For _unsaved_ datasources, we don't have that metadata already on the server side. So we need to package the spec up and send it along with the request. Because we _have_ to send a body, this means that we have to use a POST (or at least, not a get), so we have to also include the requested method, and actual request body in the request body.
> 
> In order to make that transparent from the plugins point of view, I have a custom fetch wrapper that checks if we're calling an unsaved datasources and transparently wraps the request in the format that an unsaved datasource requires. That way, plugins can continue to make fetchs as usual and not worry about handling the case of saved vs unsaved.
> 
> Does that sort of make sense?

Thank you for the explanations ! I was wondering too why we need to have another fetch wrapper.

Now I'm wondering if it is worth it. 
The whole point initially was just to test the connection to the datasource, not really to use an unsaved datasource to make a timeseries query. 
That's definitely interesting, I'm just worried it creates more difficulties to create a datasource plugin. But maybe I'm worried for nothing.

sinkingpoint:
I think there's still value here. Working within limiting it to just health checks, and piggy-backing off of https://github.com/perses/perses/blob/main/ui/app/src/model/fetch.ts I think I could put something together

AntoineThebaud:
this is just to bind the issue ref to keep the history: https://github.com/perses/perses/issues/1542

PR #1685, opened 2024-01-02
[ENHANCEMENT]: Some improvements around OIDC/OAuth provider configuration

celian-garcia:
# Description
This PR comes after having tested several external OIDC provider.

So far I tested Azure, Google, Gitlab, Linkedin as OIDC providers. 
This last one seems to not support all the features like the others.

That's why this PR allow more fine grain in configuration:

### ``discovery_url`` config
Set a custom Discovery URL (the famous "/.well-known/openid-configuration" where all the config can be accessed)

With linkedin the problem is that the lib we use expected the discovery url to be {issuer}/.well-known/openid-configuration, which is not the case
```
// 20240102161055
// https://www.linkedin.com/oauth/.well-known/openid-configuration

{
  "issuer": "https://www.linkedin.com",
```

### PKCE check
By default my first implementatio assumed that PKCE verification is supported by all the OIDC providers, which is not the case. 
So we can now disable it through ``disable_pkce`` configuration.

### More efficient user info parsing
In anticipation, I also include in that PR an interface and a service that will have to be implemented to upsert user at login time (see userinfo.go file)

> Important Note: the left part of the email is now used to generate username, and I am planning in a first place to forbid the login with same username and different providers. (this is a best effort solution and will have to be rethink later)
> It is also still subject to change in the following PRs when we'll do the sync. 
> image

# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/internal/api/impl/auth/oauth.go
+++ b/internal/api/impl/auth/oauth.go
@@ -215,13 +215,13 @@ func (e *oAuthEndpoint) authHandler(ctx echo.Context) error {
 	// Save the state cookie, will be verified in the codeExchangeHandler
 	state := uuid.NewString()
 	if err := e.saveStateCookie(ctx, state); err != nil {
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	// Save the PKCE code verifier cookie, will be verified in the codeExchangeHandler
 	verifier := oauth2.GenerateVerifier()
 	if err := e.saveCodeVerifierCookie(ctx, verifier); err != nil {
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	// Redirect user to consent page to ask for permission
@@ -241,21 +241,21 @@ func (e *oAuthEndpoint) codeExchangeHandler(ctx echo.Context) error {
 	// Verify that the state in the query match the saved one
 	if _, err := e.readStateCookie(ctx); err != nil {
 		e.logWithError(err).Error("An error occurred while verifying the state")
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	// Verify that the PKCE code verifier is present
 	verifier, err := e.readCodeVerifierCookie(ctx)
 	if err != nil {
 		e.logWithError(err).Error("An error occurred while verifying the state")
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	// Exchange the authorization code with a token
 	token, err := e.conf.Exchange(ctx.Request().Context(), code, oauth2.VerifierOption(verifier))
 	if err != nil {
 		e.logWithError(err).Error("An error occurred while exchanging code with token")
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	uInfoBody, err := e.requestUserInfo(ctx, token)
@@ -267,24 +267,24 @@ func (e *oAuthEndpoint) codeExchangeHandler(ctx echo.Context) error {
 	uInfo, err := e.buildUserInfo(uInfoBody)
 	if err != nil {
 		e.logWithError(err).Error("Failed to collect user infos.")
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	// Save the user in database
 	user, err := e.svc.SyncUser(uInfo)
 	if err != nil {
 		e.logWithError(err).Error("Failed to sync user in database.")
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	username := user.GetMetadata().GetName()
 	_, err = e.tokenManagement.accessToken(username, ctx.SetCookie)
 	if err != nil {
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 	_, err = e.tokenManagement.refreshToken(username, ctx.SetCookie)
 	if err != nil {
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	return ctx.Redirect(302, "/")
--- a/internal/api/impl/auth/oauth.go
+++ b/internal/api/impl/auth/oauth.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 
 	"github.com/google/uuid"
 	"github.com/gorilla/securecookie"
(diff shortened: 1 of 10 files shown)


pkg/model/api/config/authentication.go
```diff
@@ -36,6 +36,7 @@ type OIDCProvider struct {
 	Scopes       []string          `json:"scopes" yaml:"scopes"`
 	Issuer       string            `json:"issuer" yaml:"issuer"`
 	DiscoveryURL string            `json:"discovery_url" yaml:"discovery_url"`
+	DisablePKCE  bool              `json:"disable_pkce" yaml:"disable_pkce"`
```
Nexucis, line 39: can you update the doc about the config ?

celian-garcia: I actually did it in a next commit

internal/api/impl/auth/userinfo.go
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+package auth
+
+import "github.com/zitadel/oidc/v3/pkg/oidc"
+
+// UserInfoProfile is a simplified version of the oidc.UserInfoProfile structure.
+// It's been created as we want only a limited amount of user information in Perses and some data like locale can be
+// provided wrongly by some OIDC providers.
+type UserInfoProfile struct {
```
Nexucis, line 21: don't you need to use this struct somewhere ?

Nexucis: maybe we can put this PR in hold until this refactoring is finished, wdyt ?

celian-garcia: say hello to the new interface :) I'll resolve this discussion and let you review the changes as it is quite different now.

internal/api/impl/auth/oauth.go
```diff
@@ -268,6 +297,40 @@ func (e *oAuthEndpoint) requestUserInfos(ctx echo.Context, token *oauth2.Token)
...
+//
+// In the future, this method could evolve taking some configuration parameters to allow more customization
+// on claim paths/JSON paths used.
+func (e *oAuthEndpoint) buildUserInfo(token *oauth2.Token, body []byte) (externalUserInfo, error) {
+	userInfos := oauthUserInfo{}
+	// Parse token
+	var errToken error
+	if token != nil {
+		// TODO(cegarcia): Is it really necessary?, claims that we are searching for are id_token claims,
+		//  and if there is an id token, then it is OIDC
```
celian-garcia, line 314: Putting back into draft mode in order to resolve this comment. 
I will try more oauth providers as for now only github has been tried and doesn't have jwt access_token neither id_token. 
I have a good hope that this token parsing is actually not necessary at all.

celian-garcia: I made the choice to not parse the token here and rely only on API /userinfos data of the provider to retrieve needful information. This will be acceptable as first shot and we'll see with usages if we need to improve.

Anyway, with generic oauth, access_token is not guaranteed to be a jwt token neither to contain the right claims at the right place. A true id_token would guarantee that but if we receive id_token, then it is OIDC.

From what I see in the web, the mainstream will be more and more OIDC and this is what we should push for as much as possible.

internal/api/impl/auth/oidc.go
```diff
@@ -93,24 +134,32 @@ func (e *oIDCEndpoint) buildAuthHandler() echo.HandlerFunc {
...
 		if redirectURI == "" {
 			redirectURI = "/"
 		}
-		login := info.Subject
+		// We don´t forget to set the issuer before making any sync in the database.
+		info.issuer = e.issuer
+
+		user, err := e.svc.SyncUser(info)
+		if err != nil {
+			w.WriteHeader(500)
```
Nexucis, line 147: what does this value (`500`) mean ?

internal/api/impl/auth/userinfo.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2023 The Perses Authors
```
Nexucis, line 1:
```suggestion
// Copyright 2024 The Perses Authors
```

internal/api/impl/auth/oidc.go
celian-garcia: I confess that putting 500 is a magic number yeah haha but common everybody knows what 500 means. I will put something more clear.

Btw I should maybe put a 401 instead. I will stick with the existing from native auth

pkg/model/api/config/authentication_test.go
```diff
@@ -16,13 +16,29 @@ package config
...
+	testYamlInput := `issuer: "http://localhost:4200"`
+
+	// Run the resolver
+	err := config.NewResolver[OIDCProvider]().
+		SetConfigData([]byte(testYamlInput)).
+		Resolve(&OIDCProvider{}).
+		Verify()
+
+	// Check the error is well an error coming from the parent "Provider" struct
+	assert.ErrorContains(t, err, "`slug_id` is mandatory")
```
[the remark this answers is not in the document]
celian-garcia: omg you're right it is calling the Verify of the parent but not the child. Is it a bug of config resolver ? 
```go
// TestProvider_Verify makes sure the Verify of child struct is well called by the config Resolver
func TestProvider_VerifyChild(t *testing.T) {
	// Make sure it is a valid Provider but not a valid OIDCProvider (Verify of the child is called before the parent's one)
	testYamlInput := `
slug_id: "github"
name: "Github"
client_id: "secretthing"
client_secret: "secretthing"
redirect_uri: "http://localhost:8080"
`

	// Run the resolver
	err := config.NewResolver[OIDCProvider]().
		SetConfigData([]byte(testYamlInput)).
		Resolve(&OIDCProvider{}).
		Verify()

	// Check the error is well an error coming from the parent "Provider" struct
	assert.ErrorContains(t, err, "`issuer` is mandatory")
}
```
-> 
```
Error Trace:	/home/cegarcia/src/perses/perses/pkg/model/api/config/authentication_test.go:64
Error:      	An error is expected but got nil.
Test:       	         TestProvider_VerifyChild
```

internal/api/impl/auth/oidc.go
celian-garcia: ah it is 400 bad request that is returned in the native. Should I put the same ? Do you return vague 400 intentionally for security reason ?

pkg/model/api/config/authentication_test.go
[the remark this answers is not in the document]
celian-garcia: I committed the failing test if you want to try.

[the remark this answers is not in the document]
celian-garcia: Saw with you in private. This is actually working well, just that the lamenv library was going to the inner struct of common.URL and was doing weird stuff. Adding tags `` `json:"-" yaml:"-"` `` to ``common.URL`` fixed it. 

There seems to be a remaining threat in the lamenv library though. You seem to understand it better so I let you fix. I don´t have much time anyway :/

internal/api/impl/auth/oidc.go
celian-garcia: Aligned /login endpoints. 400 if bad creds and 500 otherwise. Note that for external providers, bad creds is handled in their UI with red message in the html

internal/api/impl/auth/oauth.go
```diff
@@ -267,24 +267,24 @@ func (e *oAuthEndpoint) codeExchangeHandler(ctx echo.Context) error {
...
 		e.logWithError(err).Error("Failed to sync user in database.")
-		return shared.UnauthorizedError
+		return shared.InternalError
 	}
 
 	username := user.GetMetadata().GetName()
 	_, err = e.tokenManagement.accessToken(username, ctx.SetCookie)
 	if err != nil {
-		return shared.UnauthorizedError
+		return shared.InternalError
```
Nexucis, line 283: can you add a log everytime we are doing that ? Just so we have a trace of the actual error

Nexucis (approved):
excepting my last comment, lgtm !

internal/api/impl/auth/oidc.go
```diff
@@ -93,23 +134,31 @@ func (e *oIDCEndpoint) buildAuthHandler() echo.HandlerFunc {
...
 		if redirectURI == "" {
 			redirectURI = "/"
 		}
-		login := info.Subject
+		// We don´t forget to set the issuer before making any sync in the database.
+		info.issuer = e.issuer
+
+		user, err := e.svc.SyncUser(info)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
```
Nexucis, line 147: log

internal/api/impl/auth/oidc.go
```diff
@@ -93,23 +134,31 @@ func (e *oIDCEndpoint) buildAuthHandler() echo.HandlerFunc {
...
-			w.WriteHeader(500)
+		if _, err := e.tokenManagement.accessToken(username, setCookie); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
 			writeResponse(w, []byte(shared.InternalError.Error()))
 			return
 		}
-		if _, err := e.tokenManagement.refreshToken(login, setCookie); err != nil {
-			w.WriteHeader(500)
+		if _, err := e.tokenManagement.refreshToken(username, setCookie); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
```
Nexucis, line 161: log

Nexucis (approved):

PR #1688, opened 2024-01-05
[ENHANCEMENT] Use common.URL in the proxy

Nexucis:


internal/api/impl/proxy/proxy.go
```diff
@@ -406,14 +406,14 @@ func (h *httpProxy) serve(c echo.Context) error {
...
 	req.URL.Path = h.path
-	logrus.Debugf("request will be redirected to %q", h.config.URL.String())
+	logrus.Debugf("request will be redirected to %q", h.config.URL.URL.String())
 
 	// Set up the proxy
 	var proxyErr error
-	reverseProxy := httputil.NewSingleHostReverseProxy(h.config.URL)
+	reverseProxy := httputil.NewSingleHostReverseProxy(h.config.URL.URL)
 	reverseProxy.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, err error) {
-		desc := h.config.URL.String()
```
celian-garcia, line 415: Do you really need these .URL ? I think that .String() is accessible anyway from the parent.

PR #1692, opened 2024-01-09
[BREAKING CHANGE] Sync external user with database at login time

celian-garcia:
# Description
This PR add the last important missing brick which consist in saving the user once logged in. 
It allows then to add some permissions to that user. 

# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/internal/api/impl/auth/service.go
+++ b/internal/api/impl/auth/service.go
@@ -1,0 +1,117 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"errors"
+
+	"github.com/perses/perses/internal/api/interface/v1/user"
+	databaseModel "github.com/perses/perses/internal/api/shared/database/model"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+// useNewIfPresent decides if we take the old or new string.
+// Return a boolean saying if the result is different from old value.
+func useNewIfPresent(old, new string) (string, bool) {
+	if len(new) > 0 {
+		return new, old != new
+	}
+	return old, false
+}
+
+// saveProfileInfo build a user spec merging the old spec with a given user profile
+// Return a boolean saying if the result is different from old value.
+func saveProfileInfo(old v1.UserSpec, uInfoProfile externalUserInfoProfile) (v1.UserSpec, bool) {
+	firstChanged := false
+	lastChanged := false
+	old.FirstName, firstChanged = useNewIfPresent(old.FirstName, uInfoProfile.GivenName)
+	old.LastName, lastChanged = useNewIfPresent(old.LastName, uInfoProfile.FamilyName)
+	return old, firstChanged || lastChanged
+}
+
+// saveProviderInfo takes provider context of external user info and control that there is not already a different
+// provider registered for that user.
+// Return a boolean saying if the result is different from old value.
+// If it is a case then an error is returned.
+// NB: Currently done like this for sake of simplicity. This is the most straightforward way to ensure that we can't
+// impersonate another user.
+func saveProviderInfo(old v1.UserSpec, uInfoProvider v1.OAuthProvider) (v1.UserSpec, bool, error) {
+	if len(old.NativeProvider.Password) > 0 {
+		return old, false, errors.New("this user is already registered with the native provider")
+	}
+	foundPerfectMatch := false
+	for _, provider := range old.OauthProviders {
+		if provider != uInfoProvider {
+			return old, false, errors.New("this user is already registered with a different oauth provider context")
+		}
+		foundPerfectMatch = true
+	}
+	if !foundPerfectMatch {
+		// No Native provider, no different oauth provider, no matching oauth provider,
+		// it probably means that oauth providers is empty and the old spec is actually a new user that we want to create
+		old.OauthProviders = append(old.OauthProviders, uInfoProvider)
+	}
+	return old, !foundPerfectMatch, nil
+}
+
+func newSpecIfChanged(old v1.UserSpec, uInfo externalUserInfo) (v1.UserSpec, bool, error) {
+	specWithProfile, profileChanged := saveProfileInfo(old, uInfo.GetProfile())
+	newSpec, providerChanged, err := saveProviderInfo(specWithProfile, uInfo.GetProviderContext())
+	return newSpec, profileChanged || providerChanged, err
+}
+
+type service struct {
+	dao user.DAO
+}
+
+func (s *service) getOrPrepareUserEntity(login string) (*v1.User, bool, error) {
+	result, err := s.dao.Get(login)
+	if err != nil {
+		if databaseModel.IsKeyNotFound(err) {
+			result = &v1.User{Kind: v1.KindUser, Metadata: v1.Metadata{Name: login}}
+			return result, true, nil
+		}
+		return nil, false, err
+	}
+	return result, false, nil
+}
+
+func (s *service) syncUser(uInfo externalUserInfo) (*v1.User, error) {
+	entity, isNew, err := s.getOrPrepareUserEntity(uInfo.GetLogin())
+	if err != nil {
+		return nil, err
+	}
+
+	var specHasChanged bool
+	entity.Spec, specHasChanged, err = newSpecIfChanged(entity.Spec, uInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	if isNew {
+		entity.Metadata.CreateNow()
+		err = s.dao.Create(entity)
+		if err != nil {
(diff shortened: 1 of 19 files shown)


pkg/model/api/v1/public_user.go
```diff
@@ -18,17 +18,25 @@ import (
...
+	Password secret.Hidden `json:"password,omitempty" yaml:"password,omitempty"`
+}
+
 type PublicUserSpec struct {
-	FirstName string        `json:"firstName,omitempty" yaml:"firstName,omitempty"`
-	LastName  string        `json:"lastName,omitempty" yaml:"lastName,omitempty"`
-	Password  secret.Hidden `json:"password,omitempty" yaml:"password,omitempty"`
+	FirstName      string               `json:"firstName,omitempty" yaml:"firstName,omitempty"`
+	LastName       string               `json:"lastName,omitempty" yaml:"lastName,omitempty"`
+	NativeProvider PublicNativeProvider `json:"native_provider,omitempty" yaml:"native_provider,omitempty"`
```
Nexucis, line 28:
```suggestion
NativeProvider PublicNativeProvider `json:"nativeProvider,omitempty" yaml:"nativeProvider,omitempty"`
```

celian-garcia: updated thanks

pkg/model/api/v1/user.go
```diff
@@ -20,10 +20,21 @@ import (
 	modelAPI "github.com/perses/perses/pkg/model/api"
 )
 
+type NativeProvider struct {
+	Password string `json:"password,omitempty" yaml:"password,omitempty"`
+}
+
+type OAuthProvider struct {
```
Gladorme, line 27: No json, yaml?

celian-garcia: ah yes I will add them, even if the default works well, thanks

celian-garcia: updated

internal/api/impl/auth/service.go
```diff
@@ -0,0 +1,119 @@
...
+		if databaseModel.IsKeyNotFound(err) {
+			result = &v1.User{Kind: v1.KindUser, Metadata: v1.Metadata{Name: login}}
+			return result, true, nil
+		}
+		return nil, false, err
+	}
+	return result, false, nil
+}
+
+func (s *service) SyncUser(uInfo externalUserInfo) (*v1.User, error) {
```
Nexucis, line 86: I guess there is no need to expose this method :)

internal/api/impl/auth/service.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if specsHasChanged {
+		entity.Metadata.Update(entity.Metadata)
+		err = s.dao.Update(entity)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return entity, nil
```
Nexucis, line 117: my two cents suggestion to simplify and clarify this function:

```go
func (s *service) syncUser(uInfo externalUserInfo) (*v1.User, error) {
	entity, isNew, err := s.getOrPrepareUserEntity(uInfo.GetLogin())
	if err != nil {
		return nil, err
	}
	spec, specHasChanged, err := newSpecIfChange(entity.Spec, uInfo)
	if err != nil {
		return nil, err
	}
	if isNew {
		entity.Metadata.CreateNow()
		err = s.dao.Create(entity)
		if err != nil {
			return nil, err
		}
	} else if specHasChanged {
		entity.Metadata.Update(entity.Metadata)
		err = s.dao.Update(entity)
		if err != nil {
			return nil, err
		}
	}
	return entity, nil
}
```

internal/api/impl/auth/service.go
```diff
@@ -0,0 +1,119 @@
...
+// NB: Currently done like this for sake of simplicity. This is the most straightforward way to ensure that we can't
+// impersonate another user.
+func saveProviderInfo(old v1.UserSpec, uInfoProvider v1.OAuthProvider) (v1.UserSpec, bool, error) {
+	if len(old.NativeProvider.Password) > 0 {
+		return old, false, errors.New("this user is already registered with the native provider")
+	}
+	foundPerfectMatch := false
+	for _, provider := range old.OauthProviders {
+		if provider != uInfoProvider {
+			return old, false, errors.New("this user is already registered with a different oauth provider context")
```
Nexucis, line 58: I think it will be nice if this error is not lost because it will be hide with the `InternalServerError`

Nexucis: perhaps in that particular case we could have an info or debug log ?

internal/api/impl/auth/service.go
```diff
@@ -0,0 +1,119 @@
...
+	return old, hasChanged
+}
+
+// saveProviderInfo takes provider context of external user info and control that there is not already a different
+// provider registered for that user.
+// Return a boolean saying if the result is different from old value.
+// If it is a case then an error is returned.
+// NB: Currently done like this for sake of simplicity. This is the most straightforward way to ensure that we can't
+// impersonate another user.
+func saveProviderInfo(old v1.UserSpec, uInfoProvider v1.OAuthProvider) (v1.UserSpec, bool, error) {
```
Nexucis, line 51: can you provide some unit test on this function ?

internal/api/impl/auth/service.go
```diff
@@ -0,0 +1,119 @@
...
+	return old, false
+}
+
+// saveProfileInfo build a user spec merging the old spec with a given user profile
+// Return a boolean saying if the result is different from old value.
+func saveProfileInfo(old v1.UserSpec, uInfoProfile externalUserInfoProfile) (v1.UserSpec, bool) {
+	hasChanged := false
+	singleChange := false
+	old.FirstName, singleChange = arbitrate(old.FirstName, uInfoProfile.GivenName)
+	hasChanged = hasChanged || singleChange
```
Nexucis, line 39:
```suggestion
hasChanged = singleChange
```

because reading quite quickly you suppose that `hasChanged` can have a different value than the initial one. I had to double check it was the case or not.
Assigning directly `singleChange` remove this question when reading the code

internal/api/impl/auth/service.go
```diff
@@ -0,0 +1,119 @@
...
+	"errors"
+
+	"github.com/perses/perses/internal/api/interface/v1/user"
+	databaseModel "github.com/perses/perses/internal/api/shared/database/model"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+// arbitrate decides if we take the old or new string.
+// Return a boolean saying if the result is different from old value.
+func arbitrate(old, new string) (string, bool) {
```
Nexucis, line 26:
```suggestion
func isNew(old, new string) (string, bool) {
```

Nexucis (commented):
nice to see we are reaching a first version of OIDC/OAUTH support !!!

Nexucis:
from private tchat with @celian-garcia, because I think it will be nice to have a trace of what we decided :

> I still foresee two problems
the fact to return 500 error when something is wrong. The login with external oauth providers is not really a rest API flow as we're used to. This is more frontend oriented with GET and redirections.
=> Shall I redirect to "/"instead of returning a 500 error

well you can do both I think. And error is needed to know what happens.
For the redirection, would be interesting why you would like to redirect to the home page if the loggin failed. Because there is an automatic redirection in the frontend to the login page if there is no token stored in the cookies.

> When there's only one oauth provider and no native provider, I removed the step of the perses signin page. I think it's a bad idea as the logout is logging us in again automatically.
=> Shall I get back the perses signin page when there is only one oauth provider and no native provider?

Yeah I think it's better if the login page is there whatever is the provider enabled. Specially because like said above, there is a redirection to the login page when you are logged.

celian-garcia:
> from private tchat with @celian-garcia, because I think it will be nice to have a trace of what we decided :
> 
> > I still foresee two problems
> > the fact to return 500 error when something is wrong. The login with external oauth providers is not really a rest API flow as we're used to. This is more frontend oriented with GET and redirections.
> > => Shall I redirect to "/"instead of returning a 500 error
> 
> well you can do both I think. And error is needed to know what happens. For the redirection, would be interesting why you would like to redirect to the home page if the loggin failed. Because there is an automatic redirection in the frontend to the login page if there is no token stored in the cookies.
Yes it is to be then redirected by the frontend itself. The current issue is that it returns 500 and display 500 Internal Server Error to the browser. 
We could also return a web page with a button to go back to "/"or go back to "/signin"

> > When there's only one oauth provider and no native provider, I removed the step of the perses signin page. I think it's a bad idea as the logout is logging us in again automatically.
> > => Shall I get back the perses signin page when there is only one oauth provider and no native provider?
> 
> Yeah I think it's better if the login page is there whatever is the provider enabled. Specially because like said above, there is a redirection to the login page when you are logged.
:+1:  alright we agree on that, I will remove the line that does the auto redirect.

internal/api/impl/auth/service.go
celian-garcia: You mean for the first one only, yes I can change this one

internal/api/impl/auth/service.go
celian-garcia: we have already an error log on top when the sync does not work. 

image

In this case, I was already logged in with celian.garcia1@gmail.com on google so I was not able to use the same email for gitlab.

internal/api/impl/auth/service.go
celian-garcia: If it was only a bool returned I would agree. But I agree also that arbitrate is not good 

What if I rename it `chooseNewIfPresent` ?

Nexucis: yup that works for me !

celian-garcia: I renamed it to `useNewIfPresent`

Issue #1537, opened 2023-10-25
Wrong error feedback when querying datasource returns an SSL error


AntoineThebaud:
Today if you try to query a datasource via https and encounter an SSL error, instead of considering this as a datasource misconfiguration error and informing the end users accordingly, the backend returns "Internal server error" which is misleading & technically wrong.

PR #1693, opened 2024-01-09
[BUGFIX] Fix the error generated  when datasource is contacted

Nexucis:
when we move the proxy to a list of endpoint with https://github.com/perses/perses/pull/1644, it appears the final error created when contacting the datasource is hidden by the middleware [HandleError](https://github.com/perses/perses/blob/main/internal/api/core/middleware/error.go).

This PR intends to be transparent about what the datasource is returning. It might fix https://github.com/perses/perses/issues/1537.

As an example, I have a created a dummy instance, and it returns the accurate error: 

image

### DIFF
--- a/internal/api/impl/proxy/proxy.go
+++ b/internal/api/impl/proxy/proxy.go
@@ -114,6 +114,16 @@ func New(dashboardDAO dashboard.DAO, secretDAO secret.DAO, globalSecretDAO globa
 	}
 }
 
+func (e *endpoint) CollectRoutes(g *route.Group) {
+	g.ANY(fmt.Sprintf("/%s/:%s/*", utils.PathGlobalDatasource, utils.ParamName), e.proxySavedGlobalDatasource, true)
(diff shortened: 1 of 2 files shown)


internal/api/impl/proxy/proxy.go
```diff
@@ -424,15 +424,20 @@ func (h *httpProxy) serve(c echo.Context) error {
 	// Reverse proxy request.
 	reverseProxy.ServeHTTP(res, req)
 	// Return any error handled during proxying request.
-	return proxyErr
+	if proxyErr != nil {
+		// we need to wrap the error with an Echo Error,
+		// otherwise the error will be hidden by the middleware "middleware.HandleError".
+		return echo.NewHTTPError(res.Status, proxyErr.Error())
```
Nexucis, line 430: this is the actual fix, the rest of the changes is just an alignment about what kind of error an endpoint should use, and it should use the error `shared.PersesError` if it's possible. And as you can see, that's not really possible for this fix ;)

AntoineThebaud (approved):

PR #1695, opened 2024-01-10
[ENHANCEMENT] Replace DaC init script by `percli dac setup`

AntoineThebaud:
# Description

`dac_setup.sh` script replaced by a new CLI command `percli dac setup`.

# Checklist


docs/user-guides/dashboard-as-code.md
```diff
@@ -37,7 +37,13 @@ See the [CUE documentation](https://cuelang.org/docs/concepts/packages/) for mor
 
 #### 2. Retrieve the CUE sources from Perses
 
-Unfortunately for now there is no clean way to import the CUE sources from Perses (no proper dependency management available in CUE as already mentionned), so in the meantime we provide the script [dac_setup.sh](../../scripts/dac_setup.sh) that takes care of adding the CUE sources from Perses as external dependencies of your repo. Simply copy this script at the root of your repository and run it. You will have to indicate the version of Perses from which to retrieve the files - unless you have specific case, better to look for the latest one.
+Ideally we should rely on a native dependency management here, but since it's not yet available for CUE as already mentionned, we provide in the meantime a dedicated CLI command in order to add the CUE sources from Perses as external dependencies to your repo:
+
+```
+percli dac setup --version 0.42.1
```
AntoineThebaud, line 43: NB: doesn't work for this version 😅, due to the repo restructuration. I will update this doc & mention a minimal version once the feature branch `feat/dashboard-as-code` will be merged & a release available with its content.

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -0,0 +1,249 @@
...
+		health, err := apiClient.V1().Health().Check()
+		if err != nil {
+			logrus.WithError(err).Debug("can't reach Perses server")
+			return fmt.Errorf("can't retrieve version from Perses server")
+		}
+		o.version = health.Version
+	}
+
+	// Validate the format of the provided version
+	if !semver.IsValid(fmt.Sprintf("v%s", o.version)) {
```
Nexucis, line 66: I would put this step in the `Validate` function. It think it fits nicely there

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -0,0 +1,249 @@
...
+percli dac setup
+
+# DaC setup when you are not connected to a server, you need to provide the Perses version to consider for dependencies retrieval
+percli dac setup --version 0.42.1
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return persesCMD.Run(o, cmd, args)
+		},
+	}
+	cmd.Flags().StringVar(&o.version, "version", "", "Version of Perses from which to retrieve the CUE dependencies.")
```
Nexucis, line 246: I'm wondering if we should manage the setup for the go sdk as well here. That's kind of weird from my point of view the command `dac init` would support only one language

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -0,0 +1,249 @@
...
+	cmd := &cobra.Command{
+		Use:   "setup",
+		Short: "Sets up a local development environment to do Dashboard-as-Code",
+		Long: `
+takes care of adding the CUE sources from Perses as external dependencies to your DaC repo.
+/!\ It must be executed at the root of your repo.
+`,
+		Example: `
+# DaC setup when you are connected to a server
+percli dac setup
```
Nexucis, line 237: you didn't like `percli dac init` ?

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -0,0 +1,249 @@
...
+	// Cleanup
+	err = os.Remove(archiveName)
+	if err != nil {
+		return fmt.Errorf("error removing the temp archive: %v", err)
+	}
+
+	return output.HandleString(o.writer, "DaC setup finished")
+}
+
+func downloadSources(version string) error {
```
Nexucis, line 107:
```suggestion
func (o *option) downloadSources() error {
```

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	// Extract the CUE deps from the archive to the destination folder
+	err = extractCUEDepsToDst()
+	if err != nil {
+		return fmt.Errorf("error extracting the CUE dependencies: %v", err)
+	}
+
+	// Cleanup
+	err = os.Remove(archiveName)
```
Nexucis, line 99: I think this should be done in a defer function right after the downloading it.

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/perses/perses/internal/cli/output"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
+)
+
+const archiveName = "sources.tar.gz"
+const depsFolderName = "cue/"
+const depsRootDstPath = "cue.mod/pkg/github.com/perses/perses" // for more info see https://cuelang.org/docs/concepts/packages/
+const maxFileSize = 10240                                      // = 10 KiB. Estimated max size for CUE files. Limit required by gosec.
```
Nexucis, line 36:
```suggestion
const (
  archiveName = "sources.tar.gz"
  depsFolderName = "cue/"
  depsRootDstPath = "cue.mod/pkg/github.com/perses/perses" // for more info see https://cuelang.org/docs/concepts/packages/
  maxFileSize = 10240              
)

File: internal/cli/cmd/dac/setup/setup.go (hunk 2 above)
Comment (Nexucis) [reply internal/cli/cmd/dac/setup/setup.go:66]: Also perhaps we should check if the version start already with a `v`. You could have both syntax supported `v0.42.0` and `0.42.0`

File: docs/user-guides/dashboard-as-code.md (hunk 1 above)
Comment (Nexucis) [reply docs/user-guides/dashboard-as-code.md:43]: maybe we should have a minimum version also. To avoid an error because someone thinks it was clever to put the version 0.2.0

File: internal/cli/cmd/dac/setup/setup.go (hunk 3 above)
Comment (AntoineThebaud) [reply internal/cli/cmd/dac/setup/setup.go:246]: IMHO better to keep it simple thus CUE-opinionated for now, and to refactor with a next PR when we'd want to make it work with the Go SDK, but that's a matter of preference I guess.

File: internal/cli/cmd/dac/setup/setup.go (hunk 4 above)
Comment (AntoineThebaud) [reply internal/cli/cmd/dac/setup/setup.go:237]: actually that's because I couldnt name the package "init" since that's its reserved by the language 😅 myself I have no clear preference here, I can bring back `init` if you want and think about a good-enough package name..

Comment (Nexucis) [reply internal/cli/cmd/dac/setup/setup.go:237]: ah yeah I forgot that `init` is reserved.
Nethermind !

File: internal/cli/cmd/dac/setup/setup.go [hunk 8]
```diff
@@ -0,0 +1,267 @@
...
+
+	// Add "v" prefix to the version if not present
+	if !strings.HasPrefix(o.version, "v") {
+		o.version = fmt.Sprintf("v%s", o.version)
+	}
+
+	return nil
+}
+
+func (o *option) Validate() error {
```
Nexucis, line 163: what do you think about checking if the binary `cue` is accessible. I'm thinking also we could provide a flag that will give us the path to the binary `cue`. So you can point exactly what binary you want to use and also in case `cue` is not in the `PATH`

Nexucis: Let's do that in a different PR. It's actually not that simple.

Nexucis (approved):
well done ! 🥳

PR #1698, opened 2024-01-12
[ENHANCEMENT] OAuth/OIDC possibly get redirect_uri from request

celian-garcia:
# Description

In some companies you may be under different proxy when you run your perses instance. 
This PR is make it possible to replace the redirect_uri hardly defined in oauth config to something computed dynamically

# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/internal/api/impl/auth/auth.go
+++ b/internal/api/impl/auth/auth.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 
 	"github.com/labstack/echo/v4"
 	"github.com/perses/perses/internal/api/interface/v1/user"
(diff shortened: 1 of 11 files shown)


internal/api/impl/auth/auth.go
```diff
@@ -28,6 +30,19 @@ import (
 	"github.com/perses/perses/pkg/model/api/config"
 )
 
+func getRedirectURI(r *http.Request, authKind string, slugID string) string {
+	rd := url.URL{}
+	rd.Host = util.GetRequestHost(r)
+	rd.Scheme = util.GetRequestProto(r)
```
Nexucis, line 36: I think importing this package just to extract the request host and the request proto is not worth it. The complexity of these methods is really low to avoid a copy and past

internal/api/impl/auth/auth.go
```diff
@@ -28,6 +30,19 @@ import (
 	"github.com/perses/perses/pkg/model/api/config"
 )
 
+func getRedirectURI(r *http.Request, authKind string, slugID string) string {
+	rd := url.URL{}
+	rd.Host = util.GetRequestHost(r)
```
Nexucis, line 35: I don't think this will work the way you would like. This method (`GetRequestHost`) is calling another method from another package: `middlewareapi.GetRequestScope(req)`: https://github.com/oauth2-proxy/oauth2-proxy/blob/master/pkg/apis/middleware/scope.go

It basically extracts the struct `RequestScope` from the request context. But it's only possible if this value is set before (as stated by the struct comment, probably set by a middleware to share information from a middleware to another one) and it's not the case in our current implementation.

Nexucis (approved):

PR #1709, opened 2024-01-16
[IGNORE] Fix failing unit test on windows

AntoineThebaud:
# Checklist


internal/cli/cmd/apply/apply_test.go
```diff
@@ -64,7 +74,7 @@ object "Project" "perses" has been applied
 			Args:            []string{"-d", "../../test/sample_resources", "--project", "perses"},
 			APIClient:       fakeapi.New(),
 			IsErrorExpected: true,
-			ExpectedMessage: `resource "game" from file "../../test/sample_resources/unknown_resource.json" not supported by the command`,
+			ExpectedMessage: strings.Replace(`resource "game" from file "..%s..%stest%ssample_resources%sunknown_resource.json" not supported by the command`, "%s", separator, -1),
```
Nexucis, line 77: that's weird you don't need for test about the command `delete`

AntoineThebaud: yeah good point.. I will have a look

AntoineThebaud: Ok found it.. it's because in the case of the `apply` test we are passing a directory, then it retrieves the list of children files via `filepath.Walk()`. That's why we get OS-specific paths in output of this, while the `delete` test just takes the single file we provide in input, and the linux-style string is not altered during the processing.

celian-garcia (approved):

PR #1710, opened 2024-01-16
[ENHANCEMENT] `percli dac setup` adds output dir to .gitignore

AntoineThebaud:
# Description

Just a simple addition to avoid further manual steps to the DaC users.
This step is skipped if the .gitignore file is not already present.

# Checklist

### DIFF
--- a/internal/cli/cmd/dac/setup/setup.go
+++ b/internal/cli/cmd/dac/setup/setup.go
@@ -15,6 +15,7 @@ package setup
 
 import (
 	"archive/tar"
+	"bufio"
 	"compress/gzip"
 	"fmt"
 	"io"
@@ -126,6 +127,43 @@ func removeFirstFolder(filePath string) string {
 	return resultPath
 }
 
+func addOutputDirToGitignore() error {
+	gitignorePath := ".gitignore"
+	comment := "# folder used to store the results of the `percli dac build` command"
+
+	// Skip if .gitignore doesn't exist
+	if _, err := os.Stat(gitignorePath); os.IsNotExist(err) {
+		logrus.Debugf("%s is not present", gitignorePath)
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	// Open the .gitignore file
+	file, err := os.OpenFile(gitignorePath, os.O_RDWR|os.O_APPEND, os.ModeAppend)
+	if err != nil {
+		return fmt.Errorf("error opening %s: %v", gitignorePath, err)
+	}
+	defer file.Close()
+
+	// Check & skip if the output dir is already listed
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if strings.TrimSpace(scanner.Text()) == config.Global.Dac.OutputFolder {
+			logrus.Debugf("%s dir is already ignored", config.Global.Dac.OutputFolder)
+			return nil
+		}
+	}
+
+	// Append the output folder to the list
+	if _, err := file.WriteString(fmt.Sprintf("\n%s\n%s\n", comment, config.Global.Dac.OutputFolder)); err != nil {
+		return fmt.Errorf("error appending to %s: %v", gitignorePath, err)
+	}
+
+	logrus.Debugf("%s dir appended to %s", config.Global.Dac.OutputFolder, gitignorePath)
+	return nil
+}
+
 type option struct {
 	persesCMD.Option
 	writer  io.Writer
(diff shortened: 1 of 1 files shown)


AntoineThebaud:
'should make this idempotent by adding a check to avoid appending to gitignore if the `built` folder is already listed.

celian-garcia (commented):
If you want my opinion, I'm not fan of pre-assuming that the user of your tool want to create a git repo, or want to use git as a version manager or even want to ignore the built folder.

That's easier to have a doc saying.

```bash
git init
percli dac setup
echo "built" >> .gitignore
```

Less code to maintain and the three lines are self explanatory.

Nexucis:
I would be in the middle here:

- In case `.gitignore` exists, then we could append it with the folder `built` and I think we should add a comment in the .gitignore explaining what is this folder ignored. Like 

```.gitignore
# folder used to store the result of the command percli dac build
/built
```

- In other case, I think we shouldn't assume the user is using git or want to use it and so for `.gitignore`. A user could even have a global `.gitignore` and so it doesn't need to create one for each project. 

FInally, it would be great to be able to change the targeted folder too.

internal/cli/cmd/dac/build/build.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	outputFolderName = "built"
+	OutputFolderName = "built"
```
Nexucis, line 31: I think this parameter should come from the config and that can be set during the command `setup` if you want to have a different default folder for the build.

And then you could still have a flag in the command `build` in case you want to define/override the default folder

Regarding the config, probably having a dedicated struct would be nice like:

```go
type DacConfig struct {
  Language        string `json: "language"`
  BuildFolderPath string `json: "build_folder_path"`
}

type Config struct {
	RestClientConfig perseshttp.RestConfigClient `json:"rest_client_config"`
	Project          string                      `json:"project,omitempty"`
	RefreshToken     string                      `json:"refresh_token,omitempty"`
        Dac              DacConfig                   `json:"dac"`
	filePath         string
	apiClient        api.ClientInterface
}
```

Nexucis: I can help on this topic if you want

AntoineThebaud: As told offline imho this is kinda out of scope here, I'm planning to address this in a next PR.

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -181,8 +218,14 @@ func (o *option) Validate() error {
 func (o *option) Execute() error {
 	logrus.Debugf("Starting DaC setup with Perses %s", o.version)
 
-	// Create the destination folder
-	err := os.MkdirAll(depsRootDstPath, os.ModePerm)
+	// Add the DaC output folder to .gitignore, if applicable
+	err := addOutputDirToGitignore()
+	if err != nil {
+		return fmt.Errorf("error adding the 'built' folder to .gitignore: %v", err)
```
Nexucis, line 224: I think it's not a mandatory step, so we could just add a warning message instead.

```go
if err := addOutputDirToGitignore(); err != nil {
   logrus.WithError(error).Warningf("unable to add the folder %q to .gitignore", OutputFolderName)
}
```

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -181,8 +218,14 @@ func (o *option) Validate() error {
...
+	}
+
+	// Create the destination folder for the dependencies
+	err = os.MkdirAll(depsRootDstPath, os.ModePerm)
```
Nexucis, line 228:
```suggestion
	if err := os.MkdirAll(depsRootDstPath, os.ModePerm); err != nil {
		return fmt.Errorf("error creating the dependencies folder structure: %v", err)
	}
```

internal/cli/cmd/dac/build/build.go
Nexucis: you could also put this PR on hold and come back on it when the folder is correctly managed. That would be the better way to go imh

AntoineThebaud: See https://github.com/perses/perses/pull/1753 ;)

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -126,6 +127,41 @@ func removeFirstFolder(filePath string) string {
 	return resultPath
 }
 
+func addOutputDirToGitignore() error {
+	gitignorePath := ".gitignore"
+	comment := "# folder used to store the results of the `percli dac build` command"
+
+	// Skip if .gitignore doesn't exist
+	if _, err := os.Stat(gitignorePath); os.IsNotExist(err) {
```
Nexucis, line 135: what would happen if the error is something else ?

Nexucis (approved):

PR #1712, opened 2024-01-17
[ENHANCEMENT] Run ID validation on Metadata.Name in all cases to avoid discrepancies

AntoineThebaud:
# Description

Run `ValidateID()` as part of the v1.Metadata unmarshalling, so the constraints apply consistently for any kind of resource having metadata.
This solves some validation discrepancies (e.g DaC unit tests, unmarshalling was working but trying to apply the dashboard to a Perses server was failing).

# Checklist

### DIFF
--- a/cue/model/api/v1/metadata_patch.cue
+++ b/cue/model/api/v1/metadata_patch.cue
@@ -13,11 +13,30 @@
 
 package v1
 
-// This definition provides placeholder values for the dashboard metadata,
-// required to pass the CUE evaluation, as those attributes are flagged as
-// mandatory in the (Go) datamodel but populated by the server in the end.
+import (
+	"strings"
+	"time"
+)
+
 #Metadata: {
+	name: string @go(Name)
+	// extra constraints for the name attribute, that reproduces some validation we have on
+	// the Golang side placeholder values for the dashboard metadata,
+	name: strings.MinRunes(1) & strings.MaxRunes(75)
+	name: =~"^[a-zA-Z0-9_.-]+$"
+
+	createdAt: time.Time @go(CreatedAt)
+	updatedAt: time.Time @go(UpdatedAt)
+	version:   uint64    @go(Version)
+	// placeholder values required to pass the CUE evaluation, as those attributes are flagged
+	// as mandatory in the (Go) datamodel but populated by the server in the end.
 	createdAt: "1970-01-01T00:00:00.000000000Z"
 	updatedAt: "1970-01-01T00:00:00.000000000Z"
 	version:   1
+}
+
+// ProjectMetadata is the metadata struct for resources that belongs to a project.
+#ProjectMetadata: {
+	#Metadata
+	project: string @go(Project)
 }
--- a/cue/model/api/v1/metadata_patch.cue
+++ b/cue/model/api/v1/metadata_patch.cue
@@ -18,25 +18,24 @@ import (
 	"time"
 )
 
+// Extra constraints for the name attributes, that reproduce validation we have on the Golang side.
+#metadataName: strings.MinRunes(1) & strings.MaxRunes(75) & =~"^[a-zA-Z0-9_.-]+$"
+
 #Metadata: {
-	name: string @go(Name)
-	// extra constraints for the name attribute, that reproduces some validation we have on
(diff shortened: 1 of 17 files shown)


pkg/model/api/v1/metadata.go
```diff
@@ -54,14 +57,73 @@ func NewProjectMetadata(project string, name string) *ProjectMetadata {
...
+}
+
+func (m *Metadata) validate() error {
+	return common.ValidateID(m.Name)
+}
+
+// This wrapping struct is required to allow defining a custom unmarshall on Metadata
+// without breaking the Project attribute (the fact Metadata is injected line in
+// ProjectMetadata caused Project string to be ignored when unmarshalling)
+type ProjectAsStruct struct {
```
AntoineThebaud, line 99: If you have a better idea for a name..

Nexucis: maybe `ProjectMedataWrapper` ?

cue/model/api/v1/metadata_patch.cue
```diff
@@ -13,11 +13,30 @@
...
+	version:   uint64    @go(Version)
+	// placeholder values required to pass the CUE evaluation, as those attributes are flagged
+	// as mandatory in the (Go) datamodel but populated by the server in the end.
 	createdAt: "1970-01-01T00:00:00.000000000Z"
 	updatedAt: "1970-01-01T00:00:00.000000000Z"
 	version:   1
 }
+
+// ProjectMetadata is the metadata struct for resources that belongs to a project.
+#ProjectMetadata: {
```
Nexucis, line 39: isn't that strange we are defining twice the struct `ProjectMetadata` ?

cue/model/api/v1/metadata_patch.cue
```diff
@@ -13,11 +13,30 @@
...
+	createdAt: time.Time @go(CreatedAt)
+	updatedAt: time.Time @go(UpdatedAt)
+	version:   uint64    @go(Version)
+	// placeholder values required to pass the CUE evaluation, as those attributes are flagged
+	// as mandatory in the (Go) datamodel but populated by the server in the end.
 	createdAt: "1970-01-01T00:00:00.000000000Z"
 	updatedAt: "1970-01-01T00:00:00.000000000Z"
 	version:   1
 }
+
```
Nexucis, line 37: I'm wondering if `ProjectMetadata` shouldn't be redefined but instead we are overriding the definition of `ProjectAsStruct`

Should be more coherent no ? 

```cue
#ProjectAsStruct {
        project: string @go(Project)
	// extra constraints for the name attribute, that reproduces some validation we have on
	// the Golang side placeholder values for the dashboard metadata,
	project: strings.MinRunes(1) & strings.MaxRunes(75)
	project: =~"^[a-zA-Z0-9_.-]+$"
}
```

Nexucis: and maybe it will solve the test issue also

Nexucis:
With this new constraint in the Metadata struct, I think we can remove this line https://github.com/perses/perses/blob/main/internal/api/shared/utils/utils.go#L92

cue/model/api/v1/metadata_patch.cue
AntoineThebaud: Oups you're right, that was a miss..

Nexucis:
@AntoineThebaud it looks like what you are doing with the Unmarshalling for the metadata doesn't work well. I just wrote this little unit test and it fails : 

```go
func TestUnmarshalJSONProjectMetadata(t *testing.T) {
	jason := `
{
  "name": "f o o",
  "project": "bar"
}
`
	m := &ProjectMetadata{}
	assert.Error(t, json.Unmarshal([]byte(jason), m))
}
```

and it shouldn't fail since `name` is not a correct name

Nexucis:
but it works with yaml, interesting: 

```go
func TestUnmarshalYAMLProjectMetadata(t *testing.T) {
	yamele := `
name: "bu ll sh it",
project: "indeed"
`
	m := &ProjectMetadata{}
	assert.Error(t, yaml.Unmarshal([]byte(yamele), m))
}
```

Nexucis (approved):

Issue #1729, opened 2024-01-25
Ephemeral dashboards


AntoineThebaud:
### Proposal

Implement ephemeral dashboards as described in https://github.com/perses/perses/discussions/1727#discussioncomment-8243762

PR #1730, opened 2024-01-26
[FEATURE] backend support of a new resource `EphemeralDashboard`

AntoineThebaud:
# Description

This adds support of a new `EphemeralDashboard` resource as described in https://github.com/perses/perses/issues/1729. This PR tackles only the new resource on the backend part. The mechanism to remove the ephemeral dashboards once the TTL is passed will come in a next PR.

NB: The implemented datamodel differs a bit from what was designed in https://github.com/perses/perses/discussions/1727#discussioncomment-8243762. Basically instead of having a v1.DashboardSpec within `spec.dashboardSpec`, it is included "inline" directly in `spec`. IMHO it's cleaner this way (avoids an additional "floor" in the model) + will make it easier I think on the frontend side to reuse the same code between regular & ephemeral dashboards.

# Checklist


internal/cli/service/ephemeraldashboard.go
```diff
@@ -0,0 +1,68 @@
...
+package service
+
+import (
+	"github.com/perses/perses/internal/cli/output"
+	v1 "github.com/perses/perses/pkg/client/api/v1"
+	modelAPI "github.com/perses/perses/pkg/model/api"
+	modelV1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+type ephemeralDashboard struct {
```
Nexucis, line 23: you need to use it in the service in the same package

pkg/model/api/v1/ephemeraldashboard_test.go
```diff
@@ -0,0 +1,556 @@
...
+							Items: []dashboard.GridItem{
+								{
+									X:      0,
+									Y:      0,
+									Width:  3,
+									Height: 4,
+									Content: &common.JSONRef{
+										Ref:    "#/spec/panels/MyPanel",
+										Path:   []string{"spec", "panels", "MyPanel"},
+										Object: panel,
```
Nexucis, line 496: Object is actually equal to nil after the Unmarshal process of the DashboardSpec

AntoineThebaud: I don't get it, this part is basically copy-pasted from the same test on Dashboard side, and it does reference the panel object:
image

Nexucis: that's because the code checking and creating the reference is executed at the Dashboard struct level not at the spec level.

This is because if you execute at the spec level, then you don't know the exact json path.

internal/api/shared/validate/validate.go
```diff
@@ -32,20 +32,20 @@ import (
 // For example in PromQL, in the function `label_replace`, it used the syntax $1, $2, for the placeholder.
 var variableTemplateNameRegexp = regexp.MustCompile(`^\w*?[^0-9]\w*$`)
 
-func Dashboard(entity *modelV1.Dashboard, sch schemas.Schemas) error {
-	if _, err := utils.BuildVariableOrder(entity.Spec.Variables, nil, nil); err != nil {
+func DashboardSpec(spec modelV1.DashboardSpec, sch schemas.Schemas) error {
```
AntoineThebaud, line 35: NB: signature modified to allow reuse with ephemeral dashboards

internal/cli/service/ephemeraldashboard.go
```diff
@@ -0,0 +1,68 @@
...
+		data = append(data, line)
+	}
+	return data
+}
+
+func (e *ephemeralDashboard) GetColumHeader() []string {
+	return []string{
+		"NAME",
+		"PROJECT",
+		"AGE",
```
AntoineThebaud, line 66: Should I add the TTL here? it's not a metadata per se, so not sure here..

pkg/model/api/v1/dashboard.go
```diff
@@ -193,16 +195,16 @@ func (d *Dashboard) validate() error {
...
 		return fmt.Errorf("spec cannot be empty")
 	}
-	return d.verifyAndSetJSONReferences()
+	return verifyAndSetJSONReferences(d.Spec.Layouts, d.Spec.Panels)
 }
 
 // verifyAndSetJSONRef will check that each JSON Reference are pointing to an existing object and will set the related pointer in the JSONRef.Object
-func (d *Dashboard) verifyAndSetJSONReferences() error {
-	for _, layout := range d.Spec.Layouts {
+func verifyAndSetJSONReferences(layouts []dashboard.Layout, panels map[string]*Panel) error {
```
AntoineThebaud, line 202: NB function signatures modified to allow reuse by ephemeral dashboards

internal/cli/service/ephemeraldashboard.go
celian-garcia: In my opinion an initial TTL in itself is not that visually interesting, You could have the wish to display the remaining time it has to live? I don't think there is a strong rule saying that we should necessarily put metadata here. To me this is more "what we want to display" VS "is that too much"

It makes me think that TTL is maybe not a good name. I's ambiguous if we speak of the time remaining to live or the time to live after it's created.

pkg/client/api/v1/ephemeraldashboard.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2023 The Perses Authors
```
celian-garcia, line 1: 2024

pkg/model/api/v1/dashboard.go
celian-garcia: can´t you change the signature to something belonging to DashboardSpec ? As you are already reusing DashboardSpec into the ephemeralDashboard. It would imply less changes. 

```go
func (d *DashboardSpec) verifyAndSetJSONReferences() error
```

pkg/model/api/v1/dashboard.go
```diff
@@ -212,24 +214,21 @@ func (d *Dashboard) verifyAndSetJSONReferences() error {
 	return nil
 }
 
-func (d *Dashboard) checkAndSetRef(ref *common.JSONRef) error {
-	// ref.Path should like that [ "spec", "panels", <name> ].
-	// So if the array is not equal to three then the reference is wrong.
-	if len(ref.Path) != 3 {
+func checkAndSetRef(ref *common.JSONRef, panels map[string]*Panel) error {
+	// ref.Path should be like [ "spec", "panels", <name> ]
+	if len(ref.Path) != len(panelsRefPath)+1 {
```
celian-garcia, line 219: I bugged 2 sec on "what the hell is that ``panelsRefPath`` variable". Maybe it's not a big performance issue if you redeclare the list inside the function? WDYT?

pkg/model/api/v1/dashboard.go
```diff
@@ -212,24 +214,21 @@ func (d *Dashboard) verifyAndSetJSONReferences() error {
 	return nil
 }
 
-func (d *Dashboard) checkAndSetRef(ref *common.JSONRef) error {
-	// ref.Path should like that [ "spec", "panels", <name> ].
-	// So if the array is not equal to three then the reference is wrong.
-	if len(ref.Path) != 3 {
+func checkAndSetRef(ref *common.JSONRef, panels map[string]*Panel) error {
```
celian-garcia, line 217: This refactor seems interesting but is it tested somewhere? I'm afraid of any regression here as you don´t have all the same error messages here for example.

celian-garcia: Maybe this refactor could take place in another PR?

pkg/model/api/v1/ephemeraldashboard.go
```diff
@@ -0,0 +1,147 @@
...
+	type plain EphemeralDashboardSpecBase
+	if err := unmarshal((*plain)(&tmp)); err != nil {
+		return err
+	}
+	*edsb = tmp
+	return nil
+}
+
+type EphemeralDashboardSpec struct {
+	EphemeralDashboardSpecBase
```
celian-garcia, line 50: and what about the Marshal, it is creating the right structure ?? I'll never understand all these oddities...

celian-garcia: I see in the following tests that the answer of my question is yes....

celian-garcia (changes requested):
Cool to see these ephemeral Dashboard coming alive :)
From an idea out of a discussion to a real resource in less than a week! Great!

pkg/model/api/v1/dashboard.go
AntoineThebaud: Yes it's tested in both dashboard_test and ephemeraldashboard_test, it checks if the ref has been successfully resolved with `Object: panel` like here https://github.com/perses/perses/pull/1730/files#diff-0ad2f6727162626eb0f59d9f583d2520b8a5229a7a8a7b8cc1c390a447ce02b9R496

internal/cli/service/ephemeraldashboard.go
```diff
@@ -0,0 +1,72 @@
...
+		"REMAINING TTL",
```
AntoineThebaud, line 70: FYI final result looks like this:

```
$ percli get ephemeraldashboards
    NAME   | PROJECT | AGE | REMAINING TTL  
-----------+---------+-----+----------------
  TempDemo | perses  | 23s | 6d
```

celian-garcia (approved):
LGTM 👍

pkg/model/api/v1/ephemeraldashboard.go
Nexucis, line 50: can you add that: `json:",inline" yaml:",inline"`

Nexucis (approved):
nice

pkg/model/api/v1/dashboard.go
```diff
@@ -142,6 +142,43 @@ func (d *DashboardSpec) validate() error {
...
+			}
+
+		}
+	}
+	return nil
+}
+
+func (d *DashboardSpec) checkAndSetRef(ref *common.JSONRef) error {
+	// ref.Path should be like [ "spec", "panels", <name> ]
+	var panelsRefPath = []string{"spec", "panels"}
```
Nexucis, line 163: at this level, there is no way to be sure that the complete path is starting by `spec`. It could start by something else or have a different depth

PR #1745, opened 2024-02-06
[ENHANCEMENT] add kubebuilder annotations to runtime types

jgbernalp:
# Description

This PR adds kubebuilder annotations to interface types that are resolved at runtime, this allows that the CRD schema can store these dynamic fields

# Checklist

## UI Changes

no UI changes

### DIFF
--- a/pkg/model/api/v1/common/plugin.go
+++ b/pkg/model/api/v1/common/plugin.go
@@ -19,7 +19,9 @@ import (
 )
 
 type Plugin struct {
-	Kind string      `json:"kind" yaml:"kind"`
+	Kind string `json:"kind" yaml:"kind"`
+	// +kubebuilder:pruning:PreserveUnknownFields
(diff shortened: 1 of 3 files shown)


Nexucis (approved):
:+1:

Wouldn't be better if we add the kubeBuilder annotation on every field (with the appropriate associated schema perhaps)?

jgbernalp:
> 👍
> 
> Wouldn't be better if we add the kubeBuilder annotation on every field (with the appropriate associated schema perhaps)?

other types are resolved to its defined type. But if I understand correctly, these dynamic fields are based on the available plugins. Which is unknown when generating the CDR schema.

Nexucis:
> But if I understand correctly, these dynamic fields are based on the available plugins. Which is unknown when generating the CDR schema.

yeah that depends on what is available in the Perses instance

PR #1746, opened 2024-02-06
[FEATURE] Add Golang SDK support for `percli dac build`

Gladorme:
# Description

Add Golang SDK support for command: `percli dac build`

# Checklist

### DIFF
--- a/docs/user-guides/dashboard-as-code.md
+++ b/docs/user-guides/dashboard-as-code.md
@@ -9,13 +9,22 @@ DaC benefits can be summarized as follows:
 - **Implementation cost reduction**, by leveraging on existing components when creating new dashboards.
 - **Maintenance cost reduction**, by making it easier to: cascade the same update on multiple dashboards, keep multiple components aligned with each other, etc..
 
-Most of these benefits comes from not dealing with the Perses JSON format directly: instead, we work with [CUE](https://cuelang.org/), a powerful templating language with a strong emphasis on data validation, that also brings the factorization & dependency management capabilities (well, [not yet!](https://github.com/cue-lang/cue/discussions/2330)) we need here.
+Most of these benefits comes from not dealing with the Perses JSON format directly: instead, we provide SDKs in languages that enable factorization, code imports and more, namely:
+* [CUE](https://cuelang.org/), a powerful templating language with a strong emphasis on data validation.
+* [Go](https://go.dev/), an opensource programming language, that probably doesn't need to be introduced...
+
+Just pick your favorite to start with DaC. If you don't have one, give a try to both!
+
+> [!NOTE]
+> CUE is the language used for the data model of the plugins, which means you'll always be able to include any external plugin installed in your Perses server into your code when using the CUE SDK. 
+> However, the Golang SDK may not support all the plugins: it's basically up to each plugin development team to provide a Go package to enable the DaC use case. 
+> This statement applies also to any other language we might have a SDK for in the future.
 
 Also, as-code means it's GitOps-friendly, meaning that you can also benefit from:
 - versions history
 - peer-review of changes before rollout
 - automated deployments
-- and more..
+- and more...
 
 ## Getting Started With Cue
 
@@ -53,18 +62,8 @@ It's first strongly recommended to ramp up on CUE if you are not familiar with t
 - The [official website](https://cuelang.org/) of Cuelang.
 - [Cuetorials](https://cuetorials.com/), a 3rd party source of information that is a very good complement.
 
-Then, you can check an example of DaC usage [here](../../internal/test/dac/input.cue). This example is heavily relying on the DaC utilities we provide. To get a deeper understanding of the Go SDK and how to use it, the best thing to do for now is to check directly its source code.
+Then, you can check an example of DaC usage [here](../../internal/test/dac/input.cue). This example is heavily relying on the DaC utilities we provide. To get a deeper understanding of these libs and how to use them, the best thing to do for now is to check directly their source code.
 
-Anytime you want to build the final dashboard definition (i.e Perses dashboard in JSON or YAML format) corresponding to your as-code definition, you can use the `dac build` command, as the following:
-
-```
-percli dac build my_dashboard.cue -ojson
-```
-
-If the build is successful, the result can be found in the generated `built` folder.
-
-> [!NOTE]
-> the `-o` (alternatively '--output') flag is optional (the default output format is YAML).
 
 ## Getting started with Go SDK
 
@@ -104,7 +103,7 @@ It's first strongly recommended to ramp up on Go if you are not familiar with th
 - The [official website](https://go.dev/) of Go.
 
 Then, you can check an example of DaC usage [here](../../internal/cli/cmd/dac/build/testdata_go/main.go).
-To get a deeper understanding of these libs and how to use them, the best thing to do for now is to check directly their source code.
+To get a deeper understanding of the Go SDK and how to use it, the best thing to do for now is to check directly its source code.
 All the SDK utilities are located in the `github.com/perses/perses/go-sdk` package.
 
 > [!WARNING]  
@@ -159,10 +158,13 @@ func main() {
 ```
 
 
+## Build dashboards
+
 Anytime you want to build the final dashboard definition (i.e: Perses dashboard in JSON or YAML format) corresponding to your as-code definition, you can use the `dac build` command, as the following:
 
 ```
 percli dac build main.go -ojson
+percli dac build my_dashboard.cue -ojson
 ```
 
 If the build is successful, the result can be found in the generated `built` folder.
@@ -170,9 +172,11 @@ If the build is successful, the result can be found in the generated `built` fol
 > [!NOTE]
 > the `-o` (alternatively '--output') flag is optional (the default output format is YAML).
 
+
 ## Deploy dashboards
 
 Once you are satisfied with the result of your DaC definition for a given dashboard, you can finally deploy it to Perses with the `apply` command:
+
 ```
 percli apply -f built/my_dashboard.json
 ```
--- a/docs/user-guides/dashboard-as-code.md
+++ b/docs/user-guides/dashboard-as-code.md
@@ -17,7 +17,7 @@ Also, as-code means it's GitOps-friendly, meaning that you can also benefit from
 - automated deployments
 - and more..
 
-## Getting started with Cue
+## Getting Started With Cue
 
 ### Prerequisites
 
@@ -45,7 +45,7 @@ percli dac setup --version 0.42.1
 
 You can omit the version flag if you are connected to a Perses server (it will retrieve its version). Otherwise, unless you have a specific case, better to pass the latest version available.
 
-## Develop dashboards
+### Develop dashboards
 
 You are now fully ready to start developping dashboards as code!
 
@@ -53,7 +53,7 @@ It's first strongly recommended to ramp up on CUE if you are not familiar with t
 - The [official website](https://cuelang.org/) of Cuelang.
 - [Cuetorials](https://cuetorials.com/), a 3rd party source of information that is a very good complement.
 
-Then, you can check an example of DaC usage [here](../../internal/test/dac/input.cue). This example is heavily relying on the DaC utilities we provide. To get a deeper understanding of these libs and how to use them, the best thing to do for now is to check directly their source code.
+Then, you can check an example of DaC usage [here](../../internal/test/dac/input.cue). This example is heavily relying on the DaC utilities we provide. To get a deeper understanding of the Go SDK and how to use it, the best thing to do for now is to check directly its source code.
 
 Anytime you want to build the final dashboard definition (i.e Perses dashboard in JSON or YAML format) corresponding to your as-code definition, you can use the `dac build` command, as the following:
 
@@ -84,7 +84,7 @@ go mod init <module name>
 ```
 See the [Go documentation](https://go.dev/doc/tutorial/create-module) for more information about this step.
 
-#### 2.Install the Perses SDK
+#### 2. Install the Perses SDK
 
 ```
 go get github.com/perses/perses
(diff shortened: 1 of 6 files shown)


docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,108 @@ If the build is successful, the result can be found in the generated `built` fol
...
+Create a new folder that will become your DaC repository, then follow the steps below:
+
+#### 1. Initialize the Go module
+
+```
+go mod init <module name>
+```
+See the [Go documentation](https://go.dev/doc/tutorial/create-module) for more information about this step.
+
+#### 2.Install the Perses SDK
```
AntoineThebaud, line 87:
```suggestion
#### 2. Install the Perses SDK
```

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,108 @@ If the build is successful, the result can be found in the generated `built` fol
...
+
+You are now fully ready to start developing dashboards as code!
+
+It's first strongly recommended to ramp up on Go if you are not familiar with this technology. For this have a look at:
+- The [official website](https://go.dev/) of Go.
+
+Then, you can check an example of DaC usage [here](../../internal/cli/cmd/dac/build/testdata_go/main.go). This example is heavily relying on the SDK utilities we provide. 
+To get a deeper understanding of these libs and how to use them, the best thing to do for now is to check directly their source code.
+All the SDK utilities are located in the `github.com/perses/perses/go-sdk` package.
+Do not log / print on the standard stdout, it will broke the output of the `dac build` command.
```
AntoineThebaud, line 109:
```suggestion
> [!WARNING]  
> Do not log / print on the standard stdout! It would break the output of the `dac build` command.
```

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,108 @@ If the build is successful, the result can be found in the generated `built` fol
 > [!NOTE]
 > the `-o` (alternatively '--output') flag is optional (the default output format is YAML).
 
+## Getting started with Go SDK
```
AntoineThebaud, line 69: imho it's weird to have at same level
- get started with cue
- develop dashboards
- get started with go
- develop dashboards

We should have
- cue
  - get started
  - develop dashboards
- go
  - get started
  - develop dashboards

-> maybe worth having 2 different files then. As you prefer.

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,108 @@ If the build is successful, the result can be found in the generated `built` fol
...
+```
+
+## Develop dashboards
+
+You are now fully ready to start developing dashboards as code!
+
+It's first strongly recommended to ramp up on Go if you are not familiar with this technology. For this have a look at:
+- The [official website](https://go.dev/) of Go.
+
+Then, you can check an example of DaC usage [here](../../internal/cli/cmd/dac/build/testdata_go/main.go). This example is heavily relying on the SDK utilities we provide.
```
AntoineThebaud, line 106:
```suggestion
Then, you can check an example of DaC usage [here](../../internal/cli/cmd/dac/build/testdata_go/main.go).
```

-> the second sentence makes sense in the case of CUE because you are not obliged to use the DaC "builder" utilities provided, but it's not the case here.

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,108 @@ If the build is successful, the result can be found in the generated `built` fol
...
+
+## Develop dashboards
+
+You are now fully ready to start developing dashboards as code!
+
+It's first strongly recommended to ramp up on Go if you are not familiar with this technology. For this have a look at:
+- The [official website](https://go.dev/) of Go.
+
+Then, you can check an example of DaC usage [here](../../internal/cli/cmd/dac/build/testdata_go/main.go). This example is heavily relying on the SDK utilities we provide. 
+To get a deeper understanding of these libs and how to use them, the best thing to do for now is to check directly their source code.
```
AntoineThebaud, line 107:
```suggestion
To get a deeper understanding of the Go SDK and how to use it, the best thing to do for now is to check directly its source code.
```

internal/cli/cmd/dac/build/build.go
```diff
@@ -98,10 +100,13 @@ func (o *option) Execute() error {
 			return nil
 		}
 
-		err = o.processFile(path)
-		if err != nil {
-			// Tell the user about the error but don't stop the processing
-			fmt.Printf("error processing file %s: %v\n", path, err)
+		extension := filepath.Ext(path)
+		if extension == goExtension || extension == cueExtension {
```
AntoineThebaud, line 104: Why are you checking the extension at this level in the case of directory processing, but not in the case of a single file just above? Given the logic of processFile() I guess you can just remove this check

internal/cli/cmd/dac/build/build.go
```diff
@@ -114,13 +119,21 @@ func (o *option) Execute() error {
...
-	// NB2: Since cue is written in Go, we could consider relying on its code instead of going the exec way.
-	//      However the cue code is (for now at least) not well packaged for such external reuse.
-	//      See https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/eval.go#L87
-	// NB3: #nosec is needed here even if the user-fed parts of the command are sanitized upstream
-	cmd := exec.Command("cue", "eval", file, "--out", o.Output, "--concrete") // #nosec
+func (o *option) processFile(file string, extension string) error {
+	var cmd *exec.Cmd
+
+	if extension == goExtension {
+		cmd = exec.Command("go", "run", file) // #nosec
```
AntoineThebaud, line 126: no way to call the code behind `go run` directly?

internal/cli/cmd/dac/build/build.go
```diff
@@ -174,10 +187,12 @@ func NewCMD() *cobra.Command {
 		Short: "Build the given DaC file, or directory containing DaC files",
 		Long: `
 Generate the final output (YAML by default, or JSON) of the given DaC file - or directory containing DaC files.
-Only CUE files are supported for now.
+Only CUE and Go files are supported for now.
```
AntoineThebaud, line 190:
```suggestion
The supported languages for a DaC are:
- CUE
- Go
```

internal/cli/cmd/dac/build/build_test.go
```diff
@@ -85,6 +85,18 @@ func TestDacBuildCMD(t *testing.T) {
 			IsErrorExpected: true,
 			ExpectedMessage: "if any flags in the group [file directory] are set none of the others can be; [directory file] were all set",
 		},
+		{
+			Title:           "nominal case with a single go file",
```
AntoineThebaud, line 89: can you change the titles of the existing tests to precise they are about CUE ?

AntoineThebaud (changes requested):
Noice, only minor things to adjust

docs/user-guides/dashboard-as-code.md
Gladorme: You can still create dashboards / resources using only Go constructor
```
v1.Dashboard{...}
```

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,108 @@ If the build is successful, the result can be found in the generated `built` fol
...
+- `go`, the [programming language](https://go.dev/).
+
+### Repository setup
+
+Create a new folder that will become your DaC repository, then follow the steps below:
+
+#### 1. Initialize the Go module
+
+```
+go mod init <module name>
```
Nexucis, line 83: that's something that the command `percli dac setup` could handle.

internal/cli/cmd/dac/build/build.go
Gladorme: Early check for reducing error spam for unrelated files, example: README, git file, ...

Could be removed but it could produce more noise in log

internal/cli/cmd/dac/build/testdata_go/go.mod
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+module dac
+
+go 1.21.6
+
+replace github.com/perses/perses => ../../../../../../ // TODO: remove when released
```
Nexucis, line 18: I don't think we want to remove that after the release since we will always want to test the current implementation and not the previous release

docs/user-guides/dashboard-as-code.md
AntoineThebaud: Ah indeed, nvm then

go-sdk/exec.go
```diff
@@ -0,0 +1,40 @@
...
+func executeDashboardBuilder(builder dashboard.Builder, writer io.Writer, errWriter io.Writer) {
+	output, err := json.Marshal(builder.Dashboard)
+	if err != nil {
+		fmt.Fprint(errWriter, err)
+		os.Exit(-1)
+	}
+	fmt.Fprint(writer, string(output))
+}
+
+func ExecuteDashboard(builder dashboard.Builder, err error) {
```
Nexucis, line 34: Can you add a comment explaining how it is supposed to be used ?

internal/cli/cmd/dac/build/build.go
Gladorme: AFAIK, no. Except if we do a setup as Dagger, [where the CLI is opening a browser](https://github.com/dagger/dagger/blob/f9219591913f76323f83afb8f34d1b75871ee699/cmd/dagger/run.go#L96).

internal/cli/cmd/dac/build/build.go
Nexucis: perhaps we can ignore this check and just call directly `processFile` and rely on the function to ensure to execute only file that are known

internal/cli/cmd/dac/build/build.go
```diff
@@ -114,13 +119,21 @@ func (o *option) Execute() error {
...
+		cmd = exec.Command("go", "run", file) // #nosec
+	} else if extension == cueExtension {
+		// NB: most of the work of the `build` command is actually made by the `eval` command of the cue CLI.
+		// NB2: Since cue is written in Go, we could consider relying on its code instead of going the exec way.
+		//      However the cue code is (for now at least) not well packaged for such external reuse.
+		//      See https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/eval.go#L87
+		// NB3: #nosec is needed here even if the user-fed parts of the command are sanitized upstream
+		cmd = exec.Command("cue", "eval", file, "--out", o.Output, "--concrete") // #nosec
+	} else {
+		return fmt.Errorf("invalid extension provided: must be either `cue` or `go`")
```
Nexucis, line 135: we could just ignore a file that is not supported and instead of an error we can just log a warning

Gladorme: Fine to me to keep returning err, method calling it handle the error as they want, logging or returning error

docs/user-guides/dashboard-as-code.md
Gladorme: I moved to:
```
## Getting Started With Cue
### Prerequisites
### Develop dashboards

## Getting Started With Go
### Prerequisites
### Develop dashboards
```

internal/cli/cmd/dac/build/build.go
Nexucis: the method calling this one cannot make the difference between the error from the command and a file not supported.

By the way: `go run `my_project/` can work too. You don't need to provide a specific file, you can provide a complete package and it will find for yourself the main file.

Gladorme: Changed to warning.

For "go run folder/", I think it will hard to detect if the folder is containing go main exec without user passing the info via param. And this command it not walking in the repo as current implem. With current implem, you could target a folder containing multiple package having main file.

docs/user-guides/dashboard-as-code.md
AntoineThebaud: there's the same step on the CUE side (cue.mod), imho in that case we should include mod file generation either in none or both.

docs/user-guides/dashboard-as-code.md
```diff
@@ -37,23 +37,23 @@ See the [CUE documentation](https://cuelang.org/docs/concepts/packages/) for mor
...
+### Develop dashboards
 
 You are now fully ready to start developping dashboards as code!
 
 It's first strongly recommended to ramp up on CUE if you are not familiar with this technology. For this have a look at:
 - The [official website](https://cuelang.org/) of Cuelang.
 - [Cuetorials](https://cuetorials.com/), a 3rd party source of information that is a very good complement.
 
-Then, you can check an example of DaC usage [here](../../internal/test/dac/input.cue). This example is heavily relying on the DaC utilities we provide. To get a deeper understanding of these libs and how to use them, the best thing to do for now is to check directly their source code.
+Then, you can check an example of DaC usage [here](../../internal/test/dac/input.cue). This example is heavily relying on the DaC utilities we provide. To get a deeper understanding of the Go SDK and how to use it, the best thing to do for now is to check directly its source code.
```
AntoineThebaud, line 56: wrong paragraph replaced ;)

docs/user-guides/dashboard-as-code.md
```diff
@@ -37,23 +37,23 @@ See the [CUE documentation](https://cuelang.org/docs/concepts/packages/) for mor
...
 
 Anytime you want to build the final dashboard definition (i.e Perses dashboard in JSON or YAML format) corresponding to your as-code definition, you can use the `dac build` command, as the following:
```
AntoineThebaud, line 58: the `dac build` part here could be removed and factorize with the Go one below, as it's exactly the same. Only the example differ, and you could put an example of each in the same code block

AntoineThebaud, line 58: the `dac build` part here could be removed and factorize with the Go one below, as it's exactly the same. Only the example differ, and you could put an example of each in the same code block

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,110 @@ If the build is successful, the result can be found in the generated `built` fol
...
+			),
+		),
+
+		dashboard.AddDatasource("promDemo", promDs.Prometheus(promDs.HTTPProxy("https://demo.prometheus.com"))),
+	)
+	sdk.ExecuteDashboard(builder, buildErr)
+}
+```
+
+
```
AntoineThebaud, line 161:
```suggestion
## Build dashboards
```

docs/user-guides/dashboard-as-code.md
```diff
@@ -66,6 +66,110 @@ If the build is successful, the result can be found in the generated `built` fol
...
+Anytime you want to build the final dashboard definition (i.e: Perses dashboard in JSON or YAML format) corresponding to your as-code definition, you can use the `dac build` command, as the following:
+
+```
+percli dac build main.go -ojson
```
AntoineThebaud, line 165:
```suggestion
percli dac build main.go -ojson
percli dac build my_dashboard.cue -ojson
```

internal/cli/cmd/dac/build/build.go
AntoineThebaud: Ok, forget about it then

internal/cli/cmd/dac/build/build.go
AntoineThebaud: Ok, fair enough

Nexucis (approved):

Issue #1497, opened 2023-10-20
Tooltip display when stack series is "all"


celian-garcia:
### What did you do?

https://github.com/perses/perses/assets/10258608/526e3129-fc02-4857-9548-4e250138d5aa

### What did you expect to see?

I expect to see a different tooltip by line hovered

### What did you see instead? Under which circumstances?

See capture

### Perses version

0.40.1

### Perses configuration file

```yaml
https://sandbox.perses.dev/projects/Celian/dashboards/Test?start=1h&refresh=0s
```

AntoineThebaud:
The attached video is not working :(

celian-garcia:
argh. Let me share the two main screenshots in this video: 
Hover the line on top (should display the tooltip no?)
![image](https://github.com/perses/perses/assets/10258608/2af0c02b-a3de-40b4-a265-5e67980105a7)

Hover the line on bottom (display well the tooltip but for every line?)
![image](https://github.com/perses/perses/assets/10258608/dbd9aa8e-d6be-4580-9ea3-5a8c97647bc0)

AntoineThebaud:
I confirm this is still occuring in 0.53.0-beta.2

AntoineThebaud:
Duplicate of https://github.com/perses/perses/issues/3515

PR #1750, opened 2024-02-08
[BUGFIX] Tooltip positioning broken in Firefox

### LINKED ISSUE #1396: (Firefox only) Timeseries panel: series tooltip on hover is shaky when moving around
### What did you do?

Using Mozilla Firefox, hover on a timeseries panel to display the series tooltip, and move around.

https://github.com/perses/perses/assets/7058693/55e7b2f9-3db9-48e7-874e-f25a67c01642

Stable tooltip to be able to read it while moving.

Shaking tooltip that makes it impossible to read.

### Perses version

```text
0.39
```

sjcobb:
# Description
Fixes #1396

Tooltip is broken in Firefox, fix up the mousemove logic (offsetX and offsetY have subtle differences with Chrome)

zrender is the canvas/svg graphics lib that ECharts is built on, main change is to always use their mousemove coords that [handle browser inconsistencies](https://github.com/ecomfe/zrender/blob/5.5.0/src/core/event.ts#L46-L120) for us

# Screenshots

## Before
https://github.com/perses/perses/assets/9369625/f8d9433e-4147-47fc-bae7-87ccf86242b1

Also see recording in #1396

## After

https://github.com/perses/perses/assets/9369625/a346979d-e143-452f-a3fc-e4f3b065ca57

https://github.com/perses/perses/assets/9369625/8f7ec20a-1d7e-4f13-b99f-31f37f1b4ab3

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/components/src/LineChart/LineChart.tsx
+++ b/ui/components/src/LineChart/LineChart.tsx
@@ -123,11 +123,11 @@ export const LineChart = forwardRef<ChartInstance, LineChartProps>(function Line
             // when chart undef, do not clear highlight series
             return;
           }
-          clearHighlightedSeries(chartRef.current, data.timeSeries.length);
+          clearHighlightedSeries(chartRef.current);
         },
       };
     },
-    [data.timeSeries.length]
+    []
   );
 
   const handleEvents: OnEventsType<LineSeriesOption['data'] | unknown> = useMemo(() => {
@@ -276,7 +276,7 @@ export const LineChart = forwardRef<ChartInstance, LineChartProps>(function Line
           setShowTooltip(false);
         }
         if (chartRef.current !== undefined) {
-          clearHighlightedSeries(chartRef.current, data.timeSeries.length);
+          clearHighlightedSeries(chartRef.current);
         }
       }}
       onMouseEnter={() => {
--- a/ui/components/src/TimeSeriesTooltip/LineChartTooltip.tsx
+++ b/ui/components/src/TimeSeriesTooltip/LineChartTooltip.tsx
@@ -59,7 +59,7 @@ export const LineChartTooltip = memo(function LineChartTooltip({
   const [showAllSeries, setShowAllSeries] = useState(false);
   const mousePos = useMousePosition();
   const { height, width, ref: tooltipRef } = useResizeObserver();
-  const transform = useRef('');
+  const transform = useRef<string | undefined>();
 
   const isTooltipPinned = pinnedPos !== null && enablePinning;
 
@@ -121,6 +121,7 @@ export const LineChartTooltip = memo(function LineChartTooltip({
         })}
         style={{
           transform: transform.current,
+          display: transform.current ? 'block' : 'none',
         }}
       >
         <Stack spacing={0.5}>
--- a/ui/components/src/TimeSeriesTooltip/tooltip-model.ts
+++ b/ui/components/src/TimeSeriesTooltip/tooltip-model.ts
@@ -53,10 +53,15 @@ export const defaultCursorData = {
   chartWidth: 0,
 };
 
-export const emptyTooltipData = {
-  cursor: defaultCursorData,
-  focusedSeries: null,
-};
+export const EMPTY_TOOLTIP_DATA: NearbySeriesArray = [];
+
+/**
+ * ECharts is built with zrender, zrX and zrY are undefined when not hovering over a chart canvas
+ */
+export interface ZRCoordinate {
+  x?: number;
+  y?: number;
+}
 
 export interface Coordinate {
   x: number;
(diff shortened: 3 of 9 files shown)


ui/components/src/utils/chart-actions.ts
```diff
@@ -56,14 +56,17 @@ export function restoreChart(chart: EChartsInstance) {
  * Clear all highlighted series when cursor exits canvas
  * https://echarts.apache.org/en/api.html#action.downplay
  */
-export function clearHighlightedSeries(chart: EChartsInstance, totalSeries: number) {
+export function clearHighlightedSeries(chart: EChartsInstance) {
   if (chart.dispatchAction !== undefined) {
-    for (let i = 0; i < totalSeries; i++) {
```
sjcobb, line 61: Not directly related to the Firefox issue, but ECharts doesn't actually require you set an explicit seriesIndex this way in dispatchAction. It can automatically "unhighlight" the correct series (which is way better for perf when moving your cursor between quickly between panels)

ui/components/src/TimeSeriesTooltip/tooltip-model.ts
```diff
@@ -109,12 +110,9 @@ export const useMousePosition = (): CursorData['coords'] => {
           y: e.clientY,
         },
         plotCanvas: {
-          x: e.offsetX,
```
sjcobb, line 112: See inline comment below, this is the most important part of the fix to use [zrender](https://github.com/ecomfe/zrender) mousemove coords

ui/components/src/LineChart/LineChart.tsx
```diff
@@ -123,11 +123,11 @@ export const LineChart = forwardRef<ChartInstance, LineChartProps>(function Line
             // when chart undef, do not clear highlight series
             return;
           }
-          clearHighlightedSeries(chartRef.current, data.timeSeries.length);
+          clearHighlightedSeries(chartRef.current);
         },
       };
     },
-    [data.timeSeries.length]
+    []
```
sjcobb, line 130: This entire component LineChart.tsx that uses 'category' xAxis should be deprecated and removed, it was replaced by TimeChart a while back in #1268

ui/components/src/TimeSeriesTooltip/LineChartTooltip.tsx
```diff
@@ -59,7 +59,7 @@ export const LineChartTooltip = memo(function LineChartTooltip({
   const [showAllSeries, setShowAllSeries] = useState(false);
   const mousePos = useMousePosition();
   const { height, width, ref: tooltipRef } = useResizeObserver();
-  const transform = useRef('');
+  const transform = useRef<string | undefined>();
```
sjcobb, line 62: Fixed these since type errors were thrown, but LineChartTooltip can also be removed now, as well as legacyCheckforNearbySeries and [getLineSeries](https://github.com/perses/perses/blob/8cf6c9af4ce7d17523269f28fbecc41258dd14f3/ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts#L71) (see [other comment](https://github.com/perses/perses/pull/1750/files#r1482388936))

ui/panels-plugin/src/plugins/scatterplot/Scatterplot.tsx
```diff
@@ -60,7 +60,9 @@ export function Scatterplot(props: ScatterplotProps) {
       axisPointer: {
         type: 'cross',
       },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       formatter: function (params: any) {
```
sjcobb, line 64: This [threw lint errors](https://github.com/perses/perses/actions/runs/7824675331/job/21347614114) on my PR, but should have been caught in #1672, wonder if something is up with PR checks 😕 (added TODO for changing `any` to import directly from ECharts types)

sjcobb: Oh I guess no-explicit-any is just a warning so won't fail that check, but probably should be an error

celian-garcia:
Cool ! Any chance that it solves the #1497 as well ? This was occuring on chrome, the top lines was not triggering the tooltip and the lines were highlighted all at once everytime, if I remember well

AntoineThebaud (approved):
awesome! thanks for fixing this 🙏

Nexucis (approved):
niiiicceee ! Thanks 🙏

sjcobb:
> Cool ! Any chance that it solves the #1497 as well ? This was occuring on chrome, the top lines was not triggering the tooltip and the lines were highlighted all at once everytime, if I remember well

Ah, nah that’s a different issue, we loop through the raw dataset values in our custom tooltip so it doesn’t account for aggregated values in our stacked variant correctly, but we can actually just use the ECharts [mouseover event](https://echarts.apache.org/en/api.html#events.Mouse%20events.mouseover) to find the correct series when isStacked since they won’t overlap.

Just assigned myself that one!

PR #1753, opened 2024-02-11
[ENHANCEMENT] CLI: add config to tune the DaC output folder + internal improvements for CLI testing

AntoineThebaud:
# Description

- added new config to tune the DaC output folder:
  - it's wraped in a `Dac` struct in anticipation of other dac-specific configs to come
  - "built" is still the value by default
- CLI test framework modified to allow customizing any part of the config
- fixed some broken unit tests on windows

# Checklist


Gladorme (approved):
lgtm 😄

internal/cli/config/config.go
```diff
@@ -48,38 +49,48 @@ func Init(configPath string) {
...
+	OutputFolder string `json:"output_folder,omitempty" yaml:"output_folder,omitempty"`
+}
+
 type Config struct {
 	RestClientConfig perseshttp.RestConfigClient `json:"rest_client_config"`
 	Project          string                      `json:"project,omitempty"`
 	RefreshToken     string                      `json:"refresh_token,omitempty"`
 	filePath         string
-	apiClient        api.ClientInterface
+	APIClient        api.ClientInterface
```
Nexucis, line 62: I would avoid to expose this parameter. It's not supposed to be accessible directly, but through the Getter.
Exposing it that encourage to use it directly.

internal/cli/config/config.go
```diff
@@ -80,6 +88,13 @@ func (c *Config) SetFilePath(filePath string) {
 	c.filePath = filePath
 }
 
+func (c *Config) GetDacOutputFolder() string {
```
Nexucis, line 91: if you want to keep a getter, then you should not expose the attribute `OutputFolder`.

Or you keep it exposed, then you remove this getter and instead you init this value in the method `init` above

internal/cli/config/config.go
```diff
@@ -54,6 +63,8 @@ type Config struct {
 	RefreshToken     string                      `json:"refresh_token,omitempty"`
 	filePath         string
 	apiClient        api.ClientInterface
+	// Dac wraps the configuration related to Dashboard-as-Code
+	Dac Dac `json:"dac,omitempty" yaml:"dac,omitempty"`
```
Nexucis, line 67: I would keep the exposed attribute at the beginning and the others at the end of the struct.
Also no need to specify a tag for yaml, the struct is only marshalled in JSON.

Nexucis (approved):

PR #1754, opened 2024-02-12
[ENHANCEMENT] Add output support on Go SDK

Gladorme:
# Description

Go SDK is now supporting output flag: yaml and json

# Checklist


AntoineThebaud (approved):

internal/cli/cmd/dac/build/testdata_go/main.go
```diff
@@ -102,5 +104,5 @@ func main() {
 
 		dashboard.AddDatasource("promDemo", promDs.Prometheus(promDs.HTTPProxy("#####"))),
 	)
-	sdk.ExecuteDashboard(builder, buildErr)
+	exec.ExecuteDashboard(builder, buildErr)
```
Nexucis, line 107: I think `ExecuteDashboard` is not the correct term. Perhaps `BuildDashboard` would be more accurate.

I also think that if `dashboard.New` is returning a builder, then perhaps it would be more interesting that the builder has a method `Build` that returns the dashboard and the error instead of getting the dashboard directly in the attribute of the builder

PR #1760, opened 2024-02-13
[ENHANCEMENT] `cmd dac setup` support the go language

Nexucis:


internal/cli/cmd/dac/setup/golang.go
```diff
@@ -0,0 +1,33 @@
...
+package setup
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func (o *option) setupGo() error {
+	if _, err := os.Stat("go.mod"); os.IsNotExist(err) {
```
AntoineThebaud, line 23: could you add the same check on cue side for the `cue.mod` folder ?

Nexucis: done

AntoineThebaud (approved):

PR #1785, opened 2024-02-27
Fix cue files path in docker image and fix when cue context loading that could not exist

Nexucis:


internal/api/schemas/schemas.go
```diff
@@ -236,7 +236,12 @@ func (s *sch) validatePlugin(plugin common.Plugin, modelKind string, modelName s
 		return err
 	}
 	// compile the JSON plugin into a CUE Value
-	value := cueDefs.context.Load().CompileBytes(pluginData)
+	cueContext := cueDefs.context.Load()
+	if cueContext == nil {
+		logrus.Warningf("unable to validate the plugin %q %q, associated cue definition not loaded", modelKind, modelName)
```
Nexucis, line 241: I'm wondering if it's accurate to just raise a warning and return no error here. That would mean we accept to inject in the database data that we cannot validate.

AntoineThebaud (approved):

PR #1798, opened 2024-02-29
[FEATURE] Add dac preview command

Nexucis:
I'm adding a new command that would help for the preview of the dashboard in the context of the `dashboard as code`.

```bash
percli dac preview -d ./built
```

The command will give a give of URLs for the previews. For example:

```yaml
- project: perses
  dashboard: Demo
  preview: http://localhost:8080/api/v1/projects/perses/ephemeraldashboards/pr-1664-Demo
- project: perses
  dashboard: Benchmark
  preview: http://localhost:8080/api/v1/projects/perses/ephemeraldashboards/pr-1664-Benchmark
```

Today through flags provided by the command, you can add two prefixes: the repository name & the pull request name.

I'm wondering if we should add another flags for generic prefix or it's fine like that.

Note: result can be printed in JSON too

### DIFF
--- a/internal/cli/cmd/dac/preview/preview.go
+++ b/internal/cli/cmd/dac/preview/preview.go
@@ -1,0 +1,189 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package preview
+
+import (
+	"fmt"
+	"io"
+	"net/url"
+	"strings"
+
+	"github.com/perses/perses/internal/api/utils"
+	persesCMD "github.com/perses/perses/internal/cli/cmd"
+	"github.com/perses/perses/internal/cli/config"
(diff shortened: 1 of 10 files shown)


internal/cli/service/service.go
```diff
@@ -32,6 +34,27 @@ func convertToEntityIfNoError[T modelAPI.Entity](entities []T, err error) ([]mod
...
+		return fmt.Errorf("unable to retrieve the %q from the Perses API. %w", modelV1.Kind(entity.GetKind()), apiError)
+	}
+
+	if errors.Is(apiError, perseshttp.RequestNotFoundError) {
+		// the document doesn't exist, so we have to create it.
+		if _, createError := svc.CreateResource(entity); createError != nil {
+			return createError
+		}
+	} else {
+		// the document doesn't exist, so we have to create it.
```
AntoineThebaud, line 50:
```suggestion
// the document exists, so we have to update it.
```

internal/cli/cmd/dac/preview/preview.go
```diff
@@ -0,0 +1,194 @@
...
+	opt.ProjectOption
+	opt.FileOption
+	opt.DirectoryOption
+	opt.OutputOption
+	dashboards        []*modelV1.Dashboard
+	writer            io.Writer
+	apiClient         api.ClientInterface
+	ttl               model.Duration
+	ttlAsAString      string
+	pullRequestNumber string
```
AntoineThebaud, line 69: I'm wondering if we shouldnt go for a single and generic "prefix" flag instead, so people put whatever they want (PR number & repo name looks like standard use cases, but maybe for local work they want to put their username or whatever)

Nexucis: yeah I agree. I have replaced these flags with the prefix flag

AntoineThebaud (approved):

PR #1752, opened 2024-02-09
[ENHANCEMENT] Add grouped variable helper in Go SDK

Gladorme:
# Description

Align Go SDK with Cue feature, you can create grouped variables. Grouped variables will be filtered by their parent variables. For now, it only applied automatically to label values and label names variables matchers.

Prev:
```golang
	dashboard.AddVariable("stack",
		listVar.List(
			labelValuesVar.PrometheusLabelValues("stack",
				labelValuesVar.Matchers("thanos_build_info{}"),
				labelValuesVar.Datasource("promDemo"),
			),
			listVar.DisplayName("PaaS"),
		),
	),
	dashboard.AddVariable("namespace", listVar.List(
		labelValuesVar.PrometheusLabelValues("namespace",
			labelValuesVar.Matchers("kube_namespace_labels{\"stack\"=\"$stack\"}"),
			labelValuesVar.Datasource("promDemo"),
		),
	)),
	dashboard.AddVariable("namespaceLabels", listVar.List(
		labelNamesVar.PrometheusLabelNames(
			labelNamesVar.Matchers("kube_namespace_labels{\"stack\"=\"$stack\", \"namespace\"=\"$namespace\"}"),
			labelNamesVar.Datasource("promDemo"),
		),
	)),
	dashboard.AddVariable("pod", listVar.List(
		labelValuesVar.PrometheusLabelValues("pod",
			labelValuesVar.Matchers("kube_namespace_labels{\"stack\"=\"$stack\", \"namespace\"=\"$namespace\"}"),
			labelValuesVar.Datasource("promDemo"),
		),
	)),
```

New:
```golang
dashboard.AddVariableGroup(
	group.AddVariable("stack",
		listVar.List(
			labelValuesVar.PrometheusLabelValues("stack",
				labelValuesVar.Matchers("thanos_build_info"),
				labelValuesVar.Datasource("promDemo"),
			),
			listVar.DisplayName("PaaS"),
		),
	),
	group.AddVariable("namespace", listVar.List(
		labelValuesVar.PrometheusLabelValues("namespace",
			labelValuesVar.Matchers("kube_namespace_labels"),
			labelValuesVar.Datasource("promDemo"),
		),
	)),
	group.AddIgnoredVariable("namespaceLabels", listVar.List(
		labelNamesVar.PrometheusLabelNames(
			labelNamesVar.Matchers("kube_namespace_labels"),
			labelNamesVar.Datasource("promDemo"),
		),
	)),
[...]

PR #1807, opened 2024-03-01
[ENHANCEMENT] DaC CUE SDK: rework the variables part to support more use cases

AntoineThebaud:
# Description

- Add a generic builder for "variables group" like it was done for Go SDK (https://github.com/perses/perses/pull/1752).
- Prometheus variables: change from a single builder to one builder for each kind of var. Text variable removed (changed as a standalone builder outside of prometheus lib where it shouldn't have belong).
- Move the prometheus filter logic to a dedicated builder.
- Add a builder for static list variable (previously missing).

# Checklist


internal/test/dac/input.cue
```diff
@@ -15,87 +15,93 @@ package test
```
AntoineThebaud, line 15: FYI most relevant file to look at, as it shows the library change from DaC user PoV.

cue/dac-utils/variable/list/list.cue
```diff
@@ -0,0 +1,47 @@
+// Copyright 2023 The Perses Authors
```
Nexucis, line 1:
```suggestion
// Copyright 2024 The Perses Authors
```

PR #1813, opened 2024-03-06
[ENHANCEMENT] DaC SDKs: allow using different identifiers for name vs label

AntoineThebaud:
# Description

The goal is to allow with both SDKs to use a different identifier/alias between a (prometheus) variable name & label attributes.
Also factorizes some duplicated code on CUE SDK side.

# Checklist

### DIFF
--- a/cue/dac-utils/prometheus/filter/filter.cue
+++ b/cue/dac-utils/prometheus/filter/filter.cue
@@ -21,11 +21,11 @@ import (
 
 #input: [...varBuilder]
 
-// TODO support label arg if provided like ""\(d.label)=\"$\(d.name)\"""
 filter: strings.Join(
 	[for var in #input {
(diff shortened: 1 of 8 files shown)


AntoineThebaud:
To be checked with @Gladorme why it generates a wrong matcher when going through AddVariableGroup() (`paas=\"$paas\"` instead of the expected `stack=\"$paas\"`)

AntoineThebaud:
Cancelling this one, in the end I don't feel this is very useful for now. What would be useful instead would be to allow providing a list of aliases for a given variable, so that when using metrics from different sources on a same dashboard (e.g one metric has a label "namespace" while another one has the label "ns"), your variable would still work in both cases when generating the filter. To ponder.

I'll bring back the factorization part in a new PR.

Issue #1514, opened 2023-10-25
Variable for datasource selection


AntoineThebaud:
### Proposal

Instead of "hardcoding" a given datasource in each and every panel & variable query, having a variable for datasource selection & passing its reference to them would allow to make a same dashboard work with multiple datasources easily.

AntoineThebaud:
Assigning the issue to [OrReuben](https://github.com/OrReuben) who started to look into it.

In the end since the multi-selection can also make sense in the case of datasources, let's implement that as a plugin of ListVariable instead of as a new core type as it was started with github.com/perses/perses/pull/1819. No need to start back from feat/datasource-variable branch then

See thread on slack for more context https://cloud-native.slack.com/archives/C07KQR95WBE/p1744811008896369

PR #1819, opened 2024-03-07
[FEATURE] Support new DatasourceVariable on backend side

AntoineThebaud:
# Description

Ref issue https://github.com/perses/perses/issues/1514.

# Checklist

### DIFF
--- a/internal/api/schemas/schemas.go
+++ b/internal/api/schemas/schemas.go
@@ -67,9 +67,9 @@ type Schemas interface {
 	ValidateDatasource(plugin common.Plugin, dtsName string) error
 	ValidatePanels(panels map[string]*modelV1.Panel) error
 	ValidatePanel(plugin common.Plugin, panelName string) error
-	ValidateGlobalVariable(v modelV1.VariableSpec) error
-	ValidateDashboardVariables([]dashboard.Variable) error
+	ValidateVariableSpec(v modelV1.VariableSpec) error
 	ValidateVariable(plugin common.Plugin, varName string) error
+	ValidateDashboardVariables([]dashboard.Variable) error
 	GetLoaders() []Loader
 }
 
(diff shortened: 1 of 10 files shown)


pkg/model/api/v1/variable/datasource.go
```diff
@@ -0,0 +1,38 @@
...
+	Display *Display `json:"display,omitempty" yaml:"display,omitempty"`
+	// Value from the list to be selected by default.
+	DefaultValue *DefaultValue `json:"defaultValue,omitempty" yaml:"defaultValue,omitempty"`
+	// CapturingRegexp is the regexp used to catch and filter the list of datasources.
+	// If empty, then nothing is filtered. That's the equivalent of setting CapturingRegexp with (.*)
+	CapturingRegexp string `json:"capturingRegexp,omitempty" yaml:"capturingRegexp,omitempty"`
+	// Sort method to apply when rendering the list of datasources
+	Sort *Sort `json:"sort,omitempty" yaml:"sort,omitempty"`
+	// Type is the type (= kind) of datasources
+	Type string `json:"type" yaml:"type"`
```
Nexucis, line 30: why not using Kind in that case ? We never used the word `Type` in the datamodel. I don't think we should start here to use it

pkg/model/api/v1/dashboard/variable_test.go
```diff
@@ -53,6 +55,31 @@ func TestUnmarshalJSONVariable(t *testing.T) {
...
 		},
+		{
+			title: "simple DatasourceVariable",
+			jason: `
+{
+  "kind": "DatasourceVariable",
+  "spec": {
+    "name": "datasourceVar",
+    "defaultValue": "prom-global",
+    "type": "PrometheusDatasource",
```
Nexucis, line 66: Perhaps we could check at the service level that the datasource type selected is known in the different plugins. WDYT ?

pkg/model/api/v1/variable/datasource.go
AntoineThebaud: I had a feeling that this would look weird:

```
kind: ...
spec: {
  kind: ...
  ...
}
```

but I'm ok to change it.

pkg/model/api/v1/dashboard/variable_test.go
Nexucis: yup that's exactly what I meant.
Perhaps the most appropriate place would be here: https://github.com/perses/perses/blob/main/internal/api/impl/validate/validate.go#L73

internal/api/schemas/schemas.go
```diff
@@ -178,15 +180,45 @@ func (s *sch) ValidateQuery(plugin common.Plugin, queryName string) error {
...
+	}
+}
+
+func (s *sch) validateDatasourceVariable(v modelV1.VariableSpec) error {
+	datasourceVariableSpec, ok := v.Spec.(*variable.DatasourceSpec)
+	if !ok {
+		return errors.New("Error converting Variable to DatasourceVariable")
 	}
 
+	schemasPointer := atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&s.dts.schemas)))
```
Nexucis, line 202: why don't you use `s.dts.schemas.Load` instead ?

AntoineThebaud: Ah you're right, I forgot the schemas map is already an atomic.Pointer.. I'll change that.

internal/api/schemas/schemas.go
```diff
@@ -178,15 +178,40 @@ func (s *sch) ValidateQuery(plugin common.Plugin, queryName string) error {
...
+}
+
+func (s *sch) validateDatasourceVariable(v modelV1.VariableSpec) error {
+	datasourceVariableSpec, ok := v.Spec.(*variable.DatasourceSpec)
+	if !ok {
+		return errors.New("Error converting Variable to DatasourceVariable")
+	}
+
+	schemasMap := *s.dts.schemas.Load()
+	for kind := range schemasMap {
```
Nexucis, line 201: probably you will need to check if `schemasMap == nil` before using it

AntoineThebaud: Well in that case it wont iterate and go straight to error "no plugin found" after, I think that's enough?

Nexucis: Actually you need to verify before, because when using `*s.dts.schemas.Load()` you accessing to the value of the pointer immediately.

But if the pointer is nil, you will have a panic issue here.

You can try this code :

```go
package main

import (
	"fmt"
	"sync/atomic"
)

func main() {
	var schemas atomic.Pointer[map[string]string]
	testMap := *schemas.Load()
	for k := range testMap {
		fmt.Println(k)
	}
}
```

What you need to do is the following code here:

```go
schemasMap := s.dts.schemas.Load()
if schemasMap == nil {
 // do something probably return an error
}
	
for kind := range *schemasMap {
}
```

PR #1837, opened 2024-03-13
[REFACTOR] Add new possible returned error in the API, used in oauth context

celian-garcia:
# Description

Objective of this PR is to have a complete control of the different error returned by the provider (being OIDC or OAuth). 
We base the code on the two main OAuth 2.0 error response specifications
https://www.rfc-editor.org/rfc/rfc8628#section-3.5
https://www.rfc-editor.org/rfc/rfc6749#section-5.2

The added value from always returning an error 500 is 
- in the device code flow, we can stop directly the poll if we already know that the call will never work (wrong credentials, timeout, etc...)
- in any case, the user will receive the oauth error returned by the provider directly and this will be easier to understand the problem (see first screenshots)

# Screenshots
### Client credentials (using wrong creds)
image
image
Special case for github that returns a non standard 404. It is translated to a 500 by default
image

### Device code errors (using a fake grant type in the code)
image
image
Special case for github that returns a weird wrong device code error if grant_type is faked 
image

Classic behaviours still works well 
image
image

# Checklist


internal/api/impl/auth/auth.go
```diff
@@ -141,3 +143,52 @@ func (e *endpoint) logout(ctx echo.Context) error {
 
 	return ctx.Redirect(302, "/")
 }
+
+// withOAuthErrorMdw applies to the handler a middleware that the error is well managed in oauth2.0 context.
+// This will transform any oauth2.0 retrieve token error into a bad request error, as suggested in the specs:
+// https://www.rfc-editor.org/rfc/rfc6749#section-5.2
+// https://www.rfc-editor.org/rfc/rfc8628#section-3.5
+// Otherwise, and only if the error is not an explicit Perses or HTTP error, we send back an opaque InternalError.
+func withOAuthErrorMdw(handler echo.HandlerFunc) echo.HandlerFunc {
```
celian-garcia, line 152: We can maybe put it directly in HandleError, I was not sure so I have it here now. To be discussed

pkg/client/api/auth.go
```diff
@@ -104,13 +104,14 @@ func (c *auth) ClientCredentialsToken(authKind, slugID, clientID, clientSecret s
 	body := &api.TokenRequest{
 		GrantType:    api.GrantTypeClientCredentials,
 		ClientID:     clientID,
-		ClientSecret: clientSecret,
+		ClientSecret: clientSecret + " hihi",
```
celian-garcia, line 107: revert that ! :o

pkg/client/perseshttp/request.go
```diff
@@ -319,6 +321,24 @@ type errorResponse struct {
 	Message string `json:"message"`
 }
 
+// parseErrorFromBody is trying to parse the two possible error responses that the backend can return.
+func (r *Response) parseErrorFromBody() (string, error) {
+	var err error
+	response1 := api.OAuthError{}
+	err = json.Unmarshal(r.body, &response1)
```
Nexucis, line 328: here if the error message is well formatted json, then it won't return an error even if the json doesn't match the struct `OauthError`

pkg/client/api/auth.go
```diff
@@ -104,13 +104,14 @@ func (c *auth) ClientCredentialsToken(authKind, slugID, clientID, clientSecret s
...
 	}
 	result := &api.AuthResponse{}
-	return result, c.client.Post().
+	err := c.client.Post().
```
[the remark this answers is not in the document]
celian-garcia: Yeah I wanted to align with the other method but I think I will just align the other way around

pkg/client/perseshttp/request.go
celian-garcia: Ouch ok then it's a big deal.. I will verify and maybe think again of another solution. 
An error kind ? Arh

celian-garcia: The best would probably to use a common struct for all the errors. Oauth or Perses. 
The oauth error codes could be extended with some Perses error codes. I need to think

celian-garcia: Okay I find three solutions: 

### Solution1: 
Set ``message`` or ``error`` fields as "required" respectively in Perses or Oauth error response. 
To do so I would have to override the UnmarshalJSON of each and if the field is not there, return an error 

### Solution2:
Introduce error kind and add a kind field on each struct

### Solution3:
Have a single Error response object satisfying the OAuth requirements : 
- with ``error`` containing all oauth errors + a default perses one (or use the "server_error" of oauth)
- with ``error_description`` with possibly more details (optional and would correspond to the ``message`` field of the current struct


WDYT ? 
I personally prefer the 3rd one but I wanted your opinion on that

internal/api/impl/auth/auth.go
celian-garcia: @Nexucis can you give me opinion on that ?

pkg/client/perseshttp/request.go
Nexucis: what about having a single Error response like : 

```
message: the error message or the error_description in case we are using Oauth (mandatory field)
error: additional error (optional field).
```

I am either wondering if we should really need to put the oauth error (code?) in the error field.
Does the CLI / the client needs to know the exact error ?

internal/api/impl/auth/auth.go
```diff
@@ -141,3 +143,52 @@ func (e *endpoint) logout(ctx echo.Context) error {
...
+	return func(ctx echo.Context) error {
```
Nexucis, line 153:
```suggestion
	return func(ctx echo.Context) error {
		err := next(ctx)
		if err == nil {
			return nil
		}
		
		// This error is the token error that could come from the oauth2 library
		var retrieveError *oauth2.RetrieveError
		if errors.As(err, &retrieveError) {
			return ctx.JSON(http.StatusBadRequest, api.OAuthError{
				ErrorCode:        retrieveError.ErrorCode,
				ErrorDescription: retrieveError.ErrorDescription,
			})
		}
		// This error is the oidc error that could come from the oidc library
		var oidcError *oidc.Error
		if errors.As(err, &oidcError) {
			return ctx.JSON(http.StatusBadRequest, api.OAuthError{
				ErrorCode:        string(oidcError.ErrorType),
				ErrorDescription: oidcError.Description,
			})
		}
		
		// for all other error, we don't need to manage it as it will be handled by the next middleware `HandleError`.
		return err
}
```

internal/api/impl/auth/auth.go
Nexucis: yeah I think it's good to do it like that. I'm just proposing some changes to be more focus on oauth and oidc error management and let the generic `HandleError` taking care on all other errors

internal/api/impl/auth/oauth.go
```diff
@@ -253,12 +253,12 @@ func (e *oAuthEndpoint) readCodeVerifierCookie(ctx echo.Context) (state string,
 
 func (e *oAuthEndpoint) CollectRoutes(g *route.Group) {
 	// Add routes for the "Authorization Code" flow
-	g.GET(fmt.Sprintf("/%s/%s/%s", utils.AuthKindOAuth, e.slugID, utils.PathLogin), e.authHandler, true)
-	g.GET(fmt.Sprintf("/%s/%s/%s", utils.AuthKindOAuth, e.slugID, utils.PathCallback), e.codeExchangeHandler, true)
+	g.GET(fmt.Sprintf("/%s/%s/%s", utils.AuthKindOAuth, e.slugID, utils.PathLogin), withOAuthErrorMdw(e.authHandler), true)
```
Nexucis, line 256: with the PR #1838 we could do something like that:

```suggestion
oauthGroup := g.Group(fmt.Sprintf("/%s/%s", utils.AuthKindOAuth, e.slugID), withOauthErrorMDW)
oauthGroup.Get(fmt.Sprintf("/%s", utils.PathLogin, true)
oauthGroup.Get(fmt.Sprintf("/%s", utils.PathCallback, true)

oauthGroup.Post(fmt.Sprintf("/%s", utils.PathDeviceCode, true)
oauthGroup.Post(fmt.Sprintf("/%s", utils.PathToken, true)
```

internal/api/impl/auth/auth.go
```diff
@@ -141,3 +143,41 @@ func (e *endpoint) logout(ctx echo.Context) error {
...
+		err := handler(ctx)
+		if err == nil {
+			return nil
+		}
+
+		urlError := &url.Error{}
+		if errors.As(err, &urlError) {
+			return urlError
```
Nexucis, line 161: this error is not managed by Perses, so it will produce at the end an internalError.

What's the point of this check ?

celian-garcia: yes I forgot it, it was a check. I'm removing it

pkg/client/perseshttp/request.go
celian-garcia: Cool, then I implemented it, I let you review that.

internal/cli/cmd/login/login.go
```diff
@@ -177,13 +181,29 @@ func (o *option) authAndSetToken() error {
...
-				continue
+			if tokenErr == nil {
+				// Handle the access token
+				o.accessToken = tokenResponse.AccessToken
+				o.refreshToken = tokenResponse.RefreshToken
+				return nil
+			}
+
+			reqErr := &perseshttp.RequestError{}
+			if errors.As(tokenErr, &reqErr) && reqErr.OAuthError != nil {
```
Nexucis, line 192: I think you are missing the case when `tokenErr` is not a `perseshttp.RequestError`

Nexucis: and the case when `reqErr.OauthErro == nil`

pkg/client/perseshttp/request.go
```diff
@@ -280,6 +282,7 @@ type RequestError struct {
 	Message    string
 	StatusCode int
 	Err        error
+	OAuthError *api.OAuthError
```
Nexucis, line 285: I guess you want to include that in the method `Error` no ?

celian-garcia: I hesitated haha, I see that we're aligned I update to use Err

internal/cli/cmd/login/login.go
celian-garcia: updated

internal/cli/cmd/login/login.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/perses/perses/internal/cli/output"
 	"github.com/perses/perses/pkg/client/api"
 	"github.com/perses/perses/pkg/client/perseshttp"
+	api2 "github.com/perses/perses/pkg/model/api"
```
Nexucis, line 31:
```suggestion
modelAPI "github.com/perses/perses/pkg/model/api"
```

internal/cli/cmd/login/login.go
```diff
@@ -177,13 +182,36 @@ func (o *option) authAndSetToken() error {
...
+					case errSlowDown:
+						// https://datatracker.ietf.org/doc/html/rfc8628#section-3.5
+						// "the interval MUST be increased by 5 seconds for this and all subsequent requests"
+						interval += 5
+						ticker.Reset(time.Duration(interval) * time.Second)
+						continue
+					case errAuthorizationPending:
+						// Do nothing.
+						continue
+					case errAccessDenied, errExpiredToken:
```
Nexucis, line 206: maybe it doesn't worth to put this case if we ended to execute the default case, wdyt ?

celian-garcia: Yes, I wondered the same haha and I don't know why but I copied from golang/oauth2 and they do this. I think you're right I will modify

Nexucis (approved):
awesome work @celian-garcia !! Thank you !

PR #1838, opened 2024-03-14
[ignore] Support middleware in the custom perses routes

Nexucis:
This PR is adding the support of middleware on any custom Perses route or group.
That should help to factorise a bit the code from the pr #1837


celian-garcia (approved):
Perfect thanks, I checked if such a thing was existing and just gave up on trying to implement it :smile:

PR #1850, opened 2024-03-21
[ENHANCEMENT] Allow empty value for non-constant TextVariables (= textbox use case)

AntoineThebaud:
# Description

Allow empty value for non-constant TextVariables. This is especially useful for the Grafana migrate use case, for which the non-empty value constraint is currently preventing to migrate dashboards that define Textbox variables without default value. 
The constraint is kept for constants, as empty constants don't really make sense.

# Checklist


celian-garcia (approved):
I trust you on the functional part :)

PR #1852, opened 2024-03-22
[BUGFIX] Grafana migrate: fix issue with inputs containing double quotes

AntoineThebaud:
# Description

- [BUGFIX] Grafana migrate: fix issue with inputs containing double quotes
- [ENHANCEMENT] Grafana migrate: set current input value as default suggestion

# Screenshots

For the enhancement part:

Before, current values of constants were lost in the process:
image

Now, current values are provided as default values. You can always change them.
image

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/app/src/views/import/GrafanaFlow.tsx
+++ b/ui/app/src/views/import/GrafanaFlow.tsx
@@ -23,10 +23,15 @@ import { useProjectList } from '../../model/project-client';
 import { useCreateDashboardMutation } from '../../model/dashboard-client';
 import { useIsReadonly } from '../../context/Config';
 
+type Input = {
+  name: string;
+  value?: string;
+};
+
 // GrafanaLightDashboard is a Grafana dashboard that may have some variables that need to be replaced by the user.
 interface GrafanaLightDashboard {
   // The only part that is interesting us is the list of the input that can exists in the Grafana dashboard definition.
-  __inputs?: Array<{ name: string }>;
+  __inputs?: Input[];
   // In order to have an accurate type when matching this interface with the Grafana JSON,
   // we just say we have an unknown list of key that exists, but we don't really care about what they are.
   [key: string]: unknown;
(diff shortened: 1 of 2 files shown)


ui/app/src/views/import/GrafanaFlow.tsx
```diff
@@ -48,11 +53,24 @@ function GrafanaFlow({ dashboard }: GrafanaFlowProps) {
     navigate(`/projects/${data.metadata.project}/dashboards/${data.metadata.name}`);
   });
 
+  // initialize the map with the provided input values if exist
+  dashboard?.__inputs?.map((input) => {
+    grafanaInput[input.name] = input.value ?? '';
```
celian-garcia, line 58: Dumb question: why not call setInput here and do the escape in setInput ?

ui/app/src/views/import/GrafanaFlow.tsx
```diff
@@ -23,10 +23,15 @@ import { useProjectList } from '../../model/project-client';
...
+type Input = {
+  name: string;
+  value?: string;
+};
+
 // GrafanaLightDashboard is a Grafana dashboard that may have some variables that need to be replaced by the user.
 interface GrafanaLightDashboard {
   // The only part that is interesting us is the list of the input that can exists in the Grafana dashboard definition.
-  __inputs?: Array<{ name: string }>;
+  __inputs?: Input[];
```
celian-garcia, line 34: I'm not sure I get what this __inputs stands for anyway. Is this part of the grafana data model? and you say then that you make possible modification only on that part? 

You mentioned in private that maybe you should escape also the rest of the fields. Does it regards the fields below, materialized by ``[key: string]: unknown``

celian-garcia (commented):
Dumb questions as it's too long I didn´t use states in react :P

Also I don't understand why the http client is not doing the escape by itself. There is probably something I didn´t get in that fix

ui/app/src/views/import/GrafanaFlow.tsx
AntoineThebaud: Yes it's part of it when you tick the "export for sharing externally" checkbox when exporting a dashboard on Grafana. See the first PR that brought migration support for these inputs in Perses https://github.com/perses/perses/pull/816

ui/app/src/views/import/GrafanaFlow.tsx
AntoineThebaud: calling setInput triggers a re-render, thus here it causes an infinite loop :/

AntoineThebaud: also we dont want the backslashes to appear in the text field, hence why afaiu there's no other choice that appending them right before the final submit

celian-garcia (changes requested):
As seen in private with you. You have to change the backend function that replace the input variable by the value in all the dashboard. As the dashboard is a json at this step. We need to escape the value at this step, using ``strconv.Quote``

celian-garcia (approved):

Issue #1536, opened 2023-10-25
Allow migrating dashboards with no panels defined


AntoineThebaud:
### Proposal

I guess people may have grafana dashboards that don't have any panel but only variables & links generated from those variables.

We don't provide this links feature (yet), however anyway today migrating such dashboard would fail because Perses requires that a dashboard defines at least 1 panel. Thus we should either remove this constraint, or add some logic to append a dummy panel when migrating a Grafana dashboard that doesn't embed any.

AntoineThebaud:
Closed by https://github.com/perses/perses/pull/1854

PR #1854, opened 2024-03-24
[ENHANCEMENT] Allow migration of empty panel dashboard

ying-jeanne:
# Description

Hi folks, I am first time contributor of this repo, not sure we are aligned the fix on this ticket https://github.com/perses/perses/issues/1536. This PR intent to allow migration of empty panels dashboard from Grafana, and remove the panels check in the schema of dashboard v1. 
# Screenshots

# Checklist

## UI Changes

- na Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- na Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- na Visual tests are stable and unlikely to be flaky.
  include:

### DIFF
--- a/internal/api/migrate/mapping.cuepart
+++ b/internal/api/migrate/mapping.cuepart
@@ -128,55 +128,52 @@ spec: {
         }
     }]
     // go through the top-level panels 1st time, to fill the panels section of Perses
+    panels: { for grafanaPanelId, grafanaPanel in #grafanaDashboard.panels {
+        // function-like pattern to factorize the panel conversion logic
+        _panelConversion: {
+            #panel: _
 
-    if #grafanaDashboard.panels != null {
-        panels: {for grafanaPanelId, grafanaPanel in #grafanaDashboard.panels {
-            // function-like pattern to factorize the panel conversion logic
-            _panelConversion: {
-                #panel: _
-
-                kind: "Panel"
-                spec: {
-                    display: {
-                        name: [ // switch
-                            if #panel.title != _|_ if #panel.title != "" {
-                                #panel.title
-                            },
-                            "empty"
-                        ][0]
-                        description: *#panel.description | ""
-                    }
-                    plugin: [ // switch
-                        %(conditional_panels)
+            kind: "Panel"
+            spec: {
+                display: {
+                    name: [ // switch
+                        if #panel.title != _|_ if #panel.title != "" {
+                            #panel.title
+                        },
+                        "empty"
                     ][0]
-                    // in case of no targets, the resulting empty array will be removed via unmarshaling in Go (omitempty)
-                    queries: [ if #panel.targets != _|_ for _, target in #panel.targets {
-                        kind: "TimeSeriesQuery"
-                        spec: {
-                            #target: target
-                            plugin: [ // switch
-                                %(conditional_timeseries_queries)
-                            ][0]
-                        }
-                    }]
+                    description: *#panel.description | ""
                 }
-            }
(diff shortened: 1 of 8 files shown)


internal/api/migrate/mapping.cuepart
```diff
@@ -128,52 +128,55 @@ spec: {
...
-                    %(conditional_panels)
-                ][0]
-                // in case of no targets, the resulting empty array will be removed via unmarshaling in Go (omitempty)
-                queries: [ if #panel.targets != _|_ for _, target in #panel.targets {
-                    kind: "TimeSeriesQuery"
-                    spec: {
-                        #target: target
-                        plugin: [ // switch
-                            %(conditional_timeseries_queries)
+    if #grafanaDashboard.panels != null {
```
AntoineThebaud, line 132: Could you elaborate on the addition of this check? From what I can see when you manually remove the `panels` attribute from a grafana dashboard via the JSON view, after saving it it's appended back by the backend with an empty array as a value, thus I'm not sure it's worth supporting the case where it would be `null`.

pkg/model/api/v1/dashboard.go
```diff
@@ -129,7 +130,8 @@ func (d *DashboardSpec) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 func (d *DashboardSpec) validate() error {
 	if len(d.Panels) == 0 {
-		return fmt.Errorf("spec.panels cannot be empty")
+		logrus.Trace("the dashboard panels list is empty")
```
AntoineThebaud, line 133: Personally I'm fine to no longer raise a validation error when having an empty panels list, WDYT @Nexucis? Alternatively the solution for the grafana migration here would be (as described in https://github.com/perses/perses/issues/1536) "_add some logic to append a dummy panel when migrating a Grafana dashboard that doesn't embed any._"

internal/api/migrate/mapping.cuepart
AntoineThebaud: Ok I see that the case was happening because of our own preprocessing in `rearrangeGrafanaPanelsWithinExpandedRows()` but you already fixed it by initializing the map, thanks for that. So I guess we can remove this check then.

ying-jeanne: Hi @AntoineThebaud thanks for taking a look, the cue schema change is reverted.

pkg/model/api/v1/dashboard.go
Nexucis: well on the functional side, I find it weird to be able to create an empty dashboard. But at the end probably it might simplify other flow such as the creation of the dashboard. Maybe something to be done in another PR

Nexucis:
I think we can merge it right @AntoineThebaud  ?

pkg/model/api/v1/dashboard.go
Nexucis: but we shouldn't use a logger in this package. That's not the correct place for that.
In this package you mostly have the definition of the different resource type used in Perses. This package is supposed to be "free" to import. That means you won't have any external definition.
The only exception we did was about the type `duration` coming from Prometheus.

Can you remove the log @ying-jeanne please ? And then I believe we should be good to merge it.
Thank you for this nice contribution !!

ying-jeanne: hi folks, logline is removed.

Nexucis (approved):
awesome thank you @ying-jeanne !

PR #1860, opened 2024-03-26
[ignore] add roadmap

Nexucis:
Please share if you see something missing in this highlevel roadmap


celian-garcia (approved):
No objection ^^' Did you ask a chatbot to generate it? :D

Nexucis:
Na I write it by myself. I'm not really into IA text generation and it doesn't work for Perses (not enough content haha)

PR #1864, opened 2024-03-27
[REFACTOR] Remove isExplore + allow TimeSeriesQueryEditor to be used without defining explicitly default plugin kind (take the first)

celian-garcia:
# Description

First PR of a long list that would aim refactoring the explorer. 
Here I propose to remove ``isExplore`` toggle from the ``TimeSeriesQueryEditor`` plugin. 
It degrades slightly the design (see screenshot) but removes cumbersome code.
I think that we agree anyway on the fact that the explore should stop reuse it in favor of reuse some lower level components that needs to be created.

Waiting for me to redesign the explore, I can do this PR in a feature branch if we don´t want it to be in main like this. My opinion is it's not that uglier finally :sweat_smile: 

At the same time I realized that The ``TimeSeriesQueryEditor`` needed default plugin to be defined, where in my opinion, it could just take the first defined plugin as default if defaultPluginKinds is not set. @zhuje I can modify by the way the ``TraceQueryEditor`` if you want the same behavior. So you wouldn't have to define ``defaultPluginKinds`` theorically ^^'.

# Screenshots
image

# Checklist

## UI Changes

  include:


ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
```diff
@@ -12,28 +12,19 @@
...
   initialValues: PanelEditorValues;
   initialAction: Action;
-  exploreTitleComponent?: JSX.Element;
-}
-
-export interface PanelSpecEditorProps {
-  panelDefinition: PanelDefinition;
-  onQueriesChange: (queries: QueryDefinition[]) => void;
-  onPluginSpecChange: (spec: UnknownSpec) => void;
+  exploreTitleComponent?: React.JSX.Element;
```
jgbernalp, line 27:
```suggestion
  exploreTitleComponent?: React.ReactNode;
```

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -59,41 +66,46 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
...
-      minStepPlaceholder={minStepPlaceholder}
-      minStep={minStep}
-      handleMinStepChange={handleMinStepChange}
-      handleMinStepBlur={handleMinStepBlur}
-    />
+    <Stack spacing={2}>
+      <FormControl margin="dense" fullWidth={false}>
+        {/* TODO: How do we ensure unique ID values if there are multiple of these? Can we use React 18 useId and
+                maintain 17 compatibility somehow with a polyfill/shim? */}
+        <InputLabel id="prom-datasource-label">Prometheus Datasource</InputLabel>
```
jgbernalp, line 74: Couldn't we use the datasource name?
```suggestion
        <InputLabel id={`prom-datasource-label-${datasource.name}`">Prometheus Datasource</InputLabel>
```

jgbernalp (changes requested):

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
celian-garcia: Updated with a variable computed this way 
```ts
const datasourceSelectLabelID = `prom-datasource-label-${selectedDatasource.name || 'default'}`;
```

ui/explore/src/components/PanelEditor/PanelEditorForm.tsx
celian-garcia: updated

AntoineThebaud (approved):
LGTM! Nice cleanup 💯

PR #1872, opened 2024-03-29
[ENHANCEMENT] Explorer - don't reuse PanelPreview but directly the Timeseries chart

celian-garcia:
# Description

This PR aims to remove another dependency from explorer to dashboard : The ``PanelPreview``. Indeed it's been done like this to prevent from code duplication, but the benefit is not worth it, since we'll refactor the view anyway.

Doing so, I move the ``useSuggestedStepMs`` utils directly alongside with the ``useTimeRange`` as very close. 

# Screenshots

This screenshot simulates a after/before, but what will remain is only the first one. This is to see that this is not very different from the previous one.
image

# Checklist

## UI Changes

  include:


ui/explore/src/components/TimeSeriesExplorer/TimeSeriesExplorer.tsx
```diff
@@ -0,0 +1,91 @@
+// Copyright 2023 The Perses Authors
```
celian-garcia, line 1: This is actually a rename of the ``PanelEditorForm`` component right on top.

ui/explore/package.json
```diff
@@ -23,14 +23,13 @@
     "build:types": "tsc --project tsconfig.build.json",
     "type-check": "tsc --noEmit",
     "start": "concurrently -P \"npm:build:* -- {*}\" -- --watch",
-    "test": "cross-env TZ=UTC jest",
```
celian-garcia, line 26: Doing so many removes that it remains nothing tested :sunglasses:

ui/explore/src/components/TimeSeriesExplorer/TimeSeriesExplorer.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <ExploreToolbar exploreTitleComponent={exploreTitleComponent} />
+      <Grid container spacing={2}>
+        <Grid item xs={12}>
+          <Stack gap={1}>
+            <TimeSeriesQueryEditor onChange={setQueries} queries={queries} />
+          </Stack>
+        </Grid>
+        <Grid item xs={12}>
+          <Stack gap={1}>
+            <Typography variant="h4">Preview</Typography>
```
AntoineThebaud, line 82: Not directly linked to your change but imho this "Preview" title should be either change to "Result" or simply removed.

AntoineThebaud (approved):
LGTM! Just got a single display remark

ui/explore/src/components/TimeSeriesExplorer/TimeSeriesExplorer.tsx
celian-garcia: Sure! For now, I'm not taking care that much of changing the display as I'm working in parallel on a complete redesign. So I won´t modify that if you agree with me.

Spoiler alert (work in progress)
image

Gladorme (approved):
lgtm

PR #1888, opened 2024-04-05
[ignore] refactor the way different provider are managed in the login 

Nexucis:
This PR is refactoring the way we are managing the different provider in the CLI cmd `login`.

Like that each login flows are isolated from each other and we cannot execute a default flow if nothing is entered.


celian-garcia (approved):
LGTM, I like the way you splitted it! thanks!

Issue #1170, opened 2023-05-12
Dashboard usage metrics


nicolastakashi:
Maybe this is too advanced for this moment, but would be nice if we provide metrics about the dashboard usage in general, such as.

1. Number of views
2. Last view
3. Time to load

Those are just examples, but those metrics are important for the ones are managing a solution like Perses and they care about dashboards usability.

jgbernalp:
This sounds very useful. Are you talking about RUM? We could instrument the front-end and send data to Perses backend for metrics collection, then expose them so Perses can be added as a prometheus target. This might also be helpful in error detection. Very interesting.

AntoineThebaud:
Btw such feature would be a good plus I think for Perses promotion, since this kind of metrics/KPIs in Grafana are generally available only in its paid version.

nicolastakashi:
Indeed!
We can start easily like given metrics such as number of times a dashboard was viewed and would be amazing like if we have some metrics about time taken to render dashboard.

I'm not the most experienced in terms of FrontEnd Observbility but I'm available to help

celian-garcia:
My two cent here, I think this is called continuous profiling. Here is an interesting article on the topic. https://www.cncf.io/blog/2022/05/31/what-is-continuous-profiling/

I'm not sure this is really a monitoring use case as it is half metrics, half event here. I would add it in a fourth use case on top of classic monitoring, tracing & logging. 

I'm not experience at all on the topic and if anybody has enough experience to confirm/infirm or propose a tool that would be great.

sinkingpoint:
Gonna put down my thoughts here:

There's two real ways to do this - either metrics, or events, and which one we go with depends on what level of detail we want to capture, and the usecases for that data. 

Either way, at a high level, I'd like to introduce a new path to the backend that will get called once the UI has finished rendering a dashboard.

This means that the high level flow would be:

1. User clicks on dashboard
2. Frontend sends a request to `api/v1/projects/<project>/dashboards/<dashboard>` to fetch the dashboard contents
3. Frontend renders the dashboard
4. Frontend sends a request to `api/v1/projects/<project>/dashboards/<dashboard>/visit`

The new endpoint would take a payload that looks like this (for now, we could introduce more stats later): 
```
{
  "render_time_secs": <number>,
  "panel_errors": <number>
}
```

[...]

PR #1889, opened 2024-04-06
[FEATURE] Backend work for usage tracking

sinkingpoint:
# Description

Linked with #1170

Here we add the backend side of usage tracking. We introduce a new view service that will be used to
track the usage of the API, bound to the /api/v1/view endpoint.

We then add a new implementation of that service
that tracks views in Prometheus metrics, with the
flexibility to swap out the implementation later.

# Screenshots

# Checklist

### DIFF
--- a/internal/api/dependency/service_manager.go
+++ b/internal/api/dependency/service_manager.go
@@ -1,4 +1,4 @@
-// Copyright 2021 The Perses Authors
+// Copyright 2024 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -33,6 +33,7 @@ import (
 	secretImpl "github.com/perses/perses/internal/api/impl/v1/secret"
 	userImpl "github.com/perses/perses/internal/api/impl/v1/user"
 	variableImpl "github.com/perses/perses/internal/api/impl/v1/variable"
+	viewImpl "github.com/perses/perses/internal/api/impl/v1/view"
 	"github.com/perses/perses/internal/api/interface/v1/dashboard"
 	"github.com/perses/perses/internal/api/interface/v1/datasource"
 	"github.com/perses/perses/internal/api/interface/v1/ephemeraldashboard"
@@ -49,6 +50,7 @@ import (
 	"github.com/perses/perses/internal/api/interface/v1/secret"
 	"github.com/perses/perses/internal/api/interface/v1/user"
 	"github.com/perses/perses/internal/api/interface/v1/variable"
+	"github.com/perses/perses/internal/api/interface/v1/view"
 	"github.com/perses/perses/internal/api/migrate"
 	"github.com/perses/perses/internal/api/rbac"
 	"github.com/perses/perses/internal/api/schemas"
@@ -77,6 +79,7 @@ type ServiceManager interface {
 	GetSecret() secret.Service
 	GetUser() user.Service
 	GetVariable() variable.Service
+	GetView() view.Service
 }
 
 type service struct {
@@ -102,6 +105,7 @@ type service struct {
 	secret             secret.Service
 	user               user.Service
 	variable           variable.Service
+	view               view.Service
 }
 
 func NewServiceManager(dao PersistenceManager, conf config.Config) (ServiceManager, error) {
@@ -137,6 +141,7 @@ func NewServiceManager(dao PersistenceManager, conf config.Config) (ServiceManag
 	roleBindingService := roleBindingImpl.NewService(dao.GetRoleBinding(), dao.GetRole(), dao.GetUser(), rbacService, schemasService)
 	secretService := secretImpl.NewService(dao.GetSecret(), cryptoService)
 	userService := userImpl.NewService(dao.GetUser())
+	viewService := viewImpl.NewMetricsViewService()
 
 	svc := &service{
 		crypto:             cryptoService,
@@ -160,6 +165,7 @@ func NewServiceManager(dao PersistenceManager, conf config.Config) (ServiceManag
 		secret:             secretService,
 		user:               userService,
 		variable:           variableService,
+		view:               viewService,
 	}
 	return svc, nil
 }
@@ -246,4 +252,8 @@ func (s *service) GetUser() user.Service {
 
 func (s *service) GetVariable() variable.Service {
 	return s.variable
+}
+
+func (s *service) GetView() view.Service {
+	return s.view
 }
--- a/internal/api/impl/v1/view/endpoint.go
+++ b/internal/api/impl/v1/view/endpoint.go
@@ -17,20 +17,25 @@ import (
 	"fmt"
 
 	"github.com/labstack/echo/v4"
+	"github.com/perses/perses/internal/api/crypto"
 	apiInterface "github.com/perses/perses/internal/api/interface"
 	"github.com/perses/perses/internal/api/interface/v1/view"
+	"github.com/perses/perses/internal/api/rbac"
 	"github.com/perses/perses/internal/api/route"
 	"github.com/perses/perses/internal/api/utils"
 	v1 "github.com/perses/perses/pkg/model/api/v1"
+	"github.com/perses/perses/pkg/model/api/v1/role"
 )
 
 type endpoint struct {
 	service view.Service
+	rbac    rbac.RBAC
 }
 
-func NewEndpoint(service view.Service) route.Endpoint {
+func NewEndpoint(service view.Service, rbac rbac.RBAC) route.Endpoint {
 	return &endpoint{
 		service: service,
+		rbac:    rbac,
 	}
 }
 
@@ -46,6 +51,15 @@ func (e *endpoint) View(ctx echo.Context) error {
 
 	if err := view.Validate(); err != nil {
 		return apiInterface.HandleBadRequestError(err.Error())
+	}
+
+	claims := crypto.ExtractJWTClaims(ctx)
+	if claims == nil {
+		return apiInterface.HandleUnauthorizedError("missing claims")
+	}
+
+	if ok := e.rbac.HasPermission(claims.Subject, role.ReadAction, view.Project, role.DashboardScope); !ok {
+		return apiInterface.HandleUnauthorizedError(fmt.Sprintf("missing '%s' permission in '%s' project for '%s' kind", role.ReadAction, view.Project, role.DashboardScope))
 	}
 
 	return e.service.View(&view)
--- a/internal/api/impl/v1/view/endpoint.go
+++ b/internal/api/impl/v1/view/endpoint.go
@@ -1,0 +1,52 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


sinkingpoint:
Opening this up. The only bit I'm slightly unsure on here is the `CheckPermissions` change in the toolbox. It works for my purposes, but there might be a better way

internal/api/impl/v1/view/endpoint.go
```diff
@@ -0,0 +1,64 @@
...
+	view := v1.View{}
+	if err := ctx.Bind(&view); err != nil {
+		return apiInterface.HandleBadRequestError(err.Error())
+	}
+
+	if err := view.Validate(); err != nil {
+		return apiInterface.HandleBadRequestError(err.Error())
+	}
+
+	if err := e.toolbox.CheckPermission(ctx, nil, apiInterface.Parameters{
```
celian-garcia, line 56: As @Gladorme  suggested me, it's simpler to use directly RBAC service actually 

Here is how the code would looks like 
```go
	if claims == nil {
		// Claims can be nil with anonymous endpoint and unauthenticated users, no need to continue
		return nil
	}
	if ok := e.rbac.HasPermission(claims.Subject, role.ReadAction, view.Project, role.DashboardScope); !ok {
		return apiInterface.HandleUnauthorizedError(fmt.Sprintf("missing '%s' permission in '%s' project for '%s' kind", role.ReadAction, view.Project, role.DashboardScope))
	}
```

celian-garcia: The CheckPermissions from toolbox has a lot of complicate conditions that doesn't apply with your need

celian-garcia (changes requested):

internal/api/impl/v1/view/endpoint.go
sinkingpoint: Nice. That's a lot cleaner - thanks!

internal/api/impl/v1/view/endpoint_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/perses/perses/internal/api/crypto"
+	apiInterface "github.com/perses/perses/internal/api/interface"
+	"github.com/perses/perses/internal/api/rbac"
+	"github.com/perses/perses/pkg/model/api/v1/role"
+	promclient "github.com/prometheus/client_model/go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var _ = rbac.RBAC(&testRBAC{})
```
celian-garcia, line 33: Interesting, is this a way of testing that you copied from another place in Perses ? I would have used mockgen but I'm wondering what is the commonly used way since I didn't create some yet

pkg/model/api/v1/view.go
```diff
@@ -0,0 +1,46 @@
...
+
+	if v.Dashboard == "" {
+		return fmt.Errorf("dashboard cannot be empty")
+	}
+
+	if v.RenderErrors < 0 {
+		return fmt.Errorf("render_errors cannot be negative")
+	}
+
+	if v.RenderTimeSecs < 0 {
```
celian-garcia, line 41: Shouldn't we use a more precise unit ? Like ms ?

celian-garcia: ah ok it's a float then why not

internal/api/impl/v1/view/service.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"github.com/perses/perses/internal/api/interface/v1/view"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// A counter for the total number of views of a dashboard.
+var dashboardViewCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: "perses_dashboard_view_total",
```
celian-garcia, line 24:
```suggestion
	Name: "perses_dashboard_views_total",
```

internal/api/impl/v1/view/service.go
```diff
@@ -0,0 +1,66 @@
...
+	Help: "The total number of views of a dashboard",
+}, []string{"project", "dashboard"})
+
+// A counter for the total number of render errors of a dashboard.
+var dashboardRenderErrorCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: "perses_dashboard_render_error_total",
```
celian-garcia, line 30:
```suggestion
	Name: "perses_dashboard_render_errors_total",
```

celian-garcia (approved):
Overall LGTM little question and suggestions. 

Maybe we could have a mechanism to enable it or not by config but it can be discussed apart.

pkg/model/api/v1/view.go
```diff
@@ -0,0 +1,46 @@
...
+	// The resource being viewed.
+	Project   string `json:"project" yaml:"project"`
+	Dashboard string `json:"dashboard" yaml:"dashboard"`
+
+	// Stats about the view.
+	RenderTimeSecs float64 `json:"render_time" yaml:"render_time"`
+	RenderErrors   int     `json:"render_errors" yaml:"render_errors"`
+}
+
+func (v *View) Validate() error {
```
Nexucis, line 28: I would override the `Unmarshal` method for JSON and YAML to call this method. Like that when the struct is created from the body, we are sure the data are already good.

internal/api/impl/v1/view/service.go
Nexucis, line 24: Why not defining a constant `perses` for the metric namespace ?
We could also use it when defining the http server here: https://github.com/perses/perses/blob/main/internal/api/core/core.go#L50

internal/api/impl/v1/view/endpoint.go
```diff
@@ -0,0 +1,66 @@
...
+		service: service,
+		rbac:    rbac,
+	}
+}
+
+func (e *endpoint) CollectRoutes(g *route.Group) {
+	g.POST(fmt.Sprintf("/%s", utils.PathView), e.View, false)
+}
+
+func (e *endpoint) View(ctx echo.Context) error {
```
Nexucis, line 46: I don't think you need to expose this method outside of this package

internal/api/dependency/service_manager.go
```diff
@@ -102,6 +105,7 @@ type service struct {
 	secret             secret.Service
 	user               user.Service
 	variable           variable.Service
+	view               view.Service
```
Nexucis, line 108: from the issue, it mentioned an endpoint `visit` and here it's `view`. I don't really mind to change the name, just curious about the reason you changed it :).

internal/api/impl/v1/view/service.go
```diff
@@ -0,0 +1,66 @@
...
+// scraped and stored by Prometheus.
+type metricsViewService struct {
+}
+
+func NewMetricsViewService() view.Service {
+	return &metricsViewService{}
+}
+
+func (m *metricsViewService) View(view *v1.View) error {
+	dashboardViewCounter.WithLabelValues(view.Project, view.Dashboard).Inc()
```
Nexucis, line 57: Perhaps it's a good idea to check that the dashboard exists before changing the metrics. WDF ?

internal/api/impl/v1/view/service.go
```diff
@@ -0,0 +1,66 @@
...
+	Name: "perses_dashboard_render_error_total",
+	Help: "The total number of render errors of a dashboard",
+}, []string{"project", "dashboard"})
+
+// A histogram for the render time of a dashboard.
+var dashboardRenderTime = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+	Name:    "perses_dashboard_render_time_seconds",
+	Help:    "The render time of a dashboard",
+	Buckets: prometheus.DefBuckets,
+}, []string{"project", "dashboard"})
```
Nexucis, line 39: these labels are common for every metrics, it can be a good idea to put them as a constant

Nexucis (commented):
nice, thanks for the PR. Overall like @celian-garcia it looks already good to me. I have added some proposition to improve it a bit.
Let me know wdyf.

internal/api/impl/v1/view/endpoint_test.go
sinkingpoint: It's just a pattern to ensure that `testRBAC` implements the interface - https://go.dev/doc/faq#guarantee_satisfies_interface

internal/api/impl/v1/view/service.go
sinkingpoint: Done

pkg/model/api/v1/view.go
sinkingpoint: Done. Left one comment in the chat about this, but it seems to work

internal/api/impl/v1/view/service.go
sinkingpoint: Done. I've defined it as a namespace for these metrics - the actual flow of how we register metrics is a bit opaque, involving a magic import from perses/common, so we might want to rejig that flow in the future to make metrics registration a bit more explicit and consistent

internal/api/dependency/service_manager.go
sinkingpoint: TBH, no particular reason. I completely forgot I named it visit in the issue :|

internal/api/impl/v1/view/service.go
sinkingpoint: Sounds reasonable. Done

internal/api/impl/v1/view/service.go
Nexucis: actually that shouldn't be that magic.

in `perses/common` we are providing default metrics (`_info` and metrics for the http server). Then it's simply possible to provide a prometheus metric registry to the `perses/common` app. If a metric registry is not provided, then the default one (provided by Prometheus) is used.

We are using a different metric registry in the e2e test: https://github.com/perses/perses/blob/main/internal/api/e2e/framework/server.go#L103

So today if we want to make metrics registration a bit more explicit and consistent we just need to define a new one and then register every metric in the `main` and then inject it in the `perses/common` app

pkg/model/api/v1/view.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+)
+
+var _ = json.Unmarshaler(&View{})
+var _ = yaml.Unmarshaler(&View{})
```
Nexucis, line 24: we never needed to do that for other struct / resource definition. Why did you need it ?

sinkingpoint: It's not _strictly_ necessary. Just a habit to show the intention that `View` implements those interfaces. I can remove it if you want

Nexucis: oh I didn't know this trick. Interesting !
But yeah can you remove it as we don't do it across the whole code. (just to be consistent).

pkg/model/api/v1/view.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if err := tmp.Validate(); err != nil {
+		return err
+	}
+	*v = tmp
+	return nil
+}
+
+func (v *View) Validate() error {
```
Nexucis, line 66: and here looking at the rest of the code, you don't need to expose the method.

internal/api/impl/v1/view/service.go
Nexucis: I'm currently patching a bit `perses/common` that contains a bug about how we are using the prometheus registry.
And then I will integrate it in Perses. Let see if you find it more coherent !

Nexucis (approved):
once last comments are resolved, we should be good to merge !
Thank you for doing it @sinkingpoint !!

PR #1898, opened 2024-04-09
[BREAKINGCHANGE] Replace TimeSeriesQueryEditor and TraceQueryEditor by MultiQueryEditor

celian-garcia:
# Description
relates #1303

> [!NOTE]
Depends on new plugin editor https://github.com/perses/perses/pull/1887
(If you want to review this PR only, check only the latest commit diff)

Now that we modify the ``PluginEditor`` to support multiple query types let's extend this to the query editors.
We have two query editors containing a lot of duplications (`TimeSeriesQueryEditor` and `TraceQueryEditor` not to name them). 
I propose to have only one component ``MultiQueryEditor`` that can serve as editor for one or several query types.

# Screenshots

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [x] Visual tests are stable and unlikely to be flaky.
  See [Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests)
  and [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues
  include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

PR #1899, opened 2024-04-09
[FEATURE] Add `supportedQueryTypes` to the panel plugins definition

### LINKED ISSUE #1303: Add `supportedQueryInterfaces` support to PanelPlugin interface

celian-garcia:
# Description
fixes #1303 

> [!NOTE]
Depends on new multi query editor https://github.com/perses/perses/pull/1898
(If you want to review this PR only, check only the latest commit diff)

Last implementation of this thread ^^' 
Thanks to this PR, a panel have to declare the query types that it supports.

### Value
- clarify the contract between a panel and the queries that it can support. Nothing is guessed or under the hood. 
- fully ready to create panel supporting multiple query types

=> **this PR is clearly an enabler for the ``Table`` panel implementation.**

### Additional unexpected benefit: remove hard ScatterChart mention!
Before that, the ``PanelSpecEditor`` had to guess the query type taking the first of the existing queries. 
As you can see, at creation, when queries was empty this was leading to some problem (`default:` case, the `ScatterChart` is hardly mentioned whereas it's a plugin!). 
 ```ts
    // Get the corresponding queryEditor depending on the queryType
  const getQueryEditorComponent = () => {
    const queryType = getQueryType();
    // default case handles cause where there is no queryType yet (e.g. UI > 'editing' mode > 'Add Panel')
    switch (queryType) {
      case 'TimeSeriesQuery':
        return <TimeSeriesQueryEditor queries={panelDefinition.spec.queries ?? []} onChange={onQueriesChange} />;
      case 'TraceQuery':
        return <TraceQueryEditor queries={panelDefinition.spec.queries ?? []} onChange={onQueriesChange} />;
      default:
        // ScatterChart only handles trace queries for now
        if (kind === 'ScatterChart') {
          return <TraceQueryEditor queries={panelDefinition.spec.queries ?? []} onChange={onQueriesChange} />;
        }
        return <TimeSeriesQueryEditor queries={panelDefinition.spec.queries ?? []} onChange={onQueriesChange} />;
    }
 ```
=> This is old story, now the query editor component is always ``MultiQueryEditor`` and takes directly the ``supportedQueryTypes``!

# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/panels-plugin/src/plugins/scatterplot/ScatterChart.ts
+++ b/ui/panels-plugin/src/plugins/scatterplot/ScatterChart.ts
@@ -22,5 +22,6 @@ export const ScatterChart: PanelPlugin<ScatterChartOptions> = {
   PanelComponent: ScatterChartPanel,
   // TODO: add a chart options editor plugin, for example:
   // panelOptionsEditorComponents: [{ label: 'Settings', content: ScatterChartOptionsEditorSettings }],
+  supportedQueryTypes: ['TraceQuery'],
   createInitialOptions: createInitialScatterChartOptions,
 };
--- a/ui/plugin-system/src/components/PanelSpecEditor/PanelSpecEditor.tsx
+++ b/ui/plugin-system/src/components/PanelSpecEditor/PanelSpecEditor.tsx
@@ -12,13 +12,7 @@
 // limitations under the License.
 
 import { ErrorAlert, JSONEditor } from '@perses-dev/components';
-import {
-  PanelDefinition,
-  QueryDefinition,
-  UnknownSpec,
-  isValidQueryPluginType,
-  QueryPluginType,
-} from '@perses-dev/core';
+import { PanelDefinition, QueryDefinition, UnknownSpec } from '@perses-dev/core';
 import { usePlugin } from '../../runtime';
 import { PanelPlugin } from '../../model';
 import { OptionsEditorTabsProps, OptionsEditorTabs } from '../OptionsEditorTabs';
@@ -49,21 +43,6 @@ export function PanelSpecEditor(props: PanelSpecEditorProps) {
     throw new Error(`Missing implementation for panel plugin with kind '${kind}'`);
   }
 
-  // TODO: Every panel plugin should define which query type they support in their definition
-  const getQueryTypes = (): QueryPluginType[] => {
-    const firstQueryType = panelDefinition?.spec?.queries?.[0]?.kind;
-    if (!firstQueryType || firstQueryType === '' || !isValidQueryPluginType(firstQueryType)) {
-      // this case handles cause where there is no queryType yet (e.g. UI > 'editing' mode > 'Add Panel')
-      // ScatterChart only handles trace queries for now
-      // (this is needed as, otherwise, no way to know that ScatterChart support TraceQuery and not the default TimeSeriesQuery)
-      if (kind === 'ScatterChart') {
-        return ['TraceQuery'];
-      }
-      return ['TimeSeriesQuery'];
-    }
-    return [firstQueryType];
-  };
-
   const { panelOptionsEditorComponents, hideQueryEditor } = plugin as PanelPlugin;
   let tabs: OptionsEditorTabsProps['tabs'] = [];
 
@@ -72,7 +51,7 @@ export function PanelSpecEditor(props: PanelSpecEditorProps) {
       label: 'Query',
       content: (
         <MultiQueryEditor
-          queryTypes={getQueryTypes()}
+          queryTypes={plugin.supportedQueryTypes ?? []}
           queries={panelDefinition.spec.queries ?? []}
           onChange={onQueriesChange}
         />
--- a/ui/panels-plugin/src/plugins/bar-chart/BarChart.ts
+++ b/ui/panels-plugin/src/plugins/bar-chart/BarChart.ts
@@ -27,5 +27,6 @@ export const BarChart: PanelPlugin<BarChartOptions> = {
       content: BarChartOptionsEditorSettings,
     },
   ],
+  supportedQueryTypes: ['TimeSeriesQuery'],
   createInitialOptions: createInitialBarChartOptions,
 };
--- a/ui/panels-plugin/src/plugins/gauge-chart/GaugeChart.ts
+++ b/ui/panels-plugin/src/plugins/gauge-chart/GaugeChart.ts
@@ -21,6 +21,7 @@ import { GaugeChartPanel } from './GaugeChartPanel';
  */
 export const GaugeChart: PanelPlugin<GaugeChartOptions> = {
   PanelComponent: GaugeChartPanel,
+  supportedQueryTypes: ['TimeSeriesQuery'],
   panelOptionsEditorComponents: [
     {
       label: 'Settings',
--- a/ui/panels-plugin/src/plugins/markdown/Markdown.ts
+++ b/ui/panels-plugin/src/plugins/markdown/Markdown.ts
@@ -21,6 +21,7 @@ import { MarkdownPanelOptionsEditor } from './MarkdownPanelOptionsEditor';
  */
 export const Markdown: PanelPlugin<MarkdownPanelOptions> = {
   PanelComponent: MarkdownPanel,
+  supportedQueryTypes: [],
   panelOptionsEditorComponents: [
     {
       label: 'Markdown',
--- a/ui/panels-plugin/src/plugins/stat-chart/StatChart.ts
+++ b/ui/panels-plugin/src/plugins/stat-chart/StatChart.ts
@@ -21,6 +21,7 @@ import { StatChartPanel } from './StatChartPanel';
  */
 export const StatChart: PanelPlugin<StatChartOptions> = {
   PanelComponent: StatChartPanel,
+  supportedQueryTypes: ['TimeSeriesQuery'],
   panelOptionsEditorComponents: [
     {
       label: 'Settings',
--- a/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
+++ b/ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChart.ts
@@ -21,6 +21,7 @@ import { TimeSeriesChartPanel } from './TimeSeriesChartPanel';
  */
 export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
   PanelComponent: TimeSeriesChartPanel,
+  supportedQueryTypes: ['TimeSeriesQuery'],
   panelOptionsEditorComponents: [{ label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
   createInitialOptions: createInitialTimeSeriesChartOptions,
 };
--- a/ui/plugin-system/src/model/panels.ts
+++ b/ui/plugin-system/src/model/panels.ts
@@ -12,15 +12,16 @@
 // limitations under the License.
 
 import React from 'react';
-import { UnknownSpec, PanelDefinition } from '@perses-dev/core';
+import { UnknownSpec, PanelDefinition, QueryPluginType } from '@perses-dev/core';
 import { OptionsEditorTab } from '../components';
 import { OptionsEditorProps, Plugin } from './plugin-base';
 
 export type PanelOptionsEditorComponent<T> = Pick<OptionsEditorTab, 'label'> & {
   content: React.ComponentType<OptionsEditorProps<T>>;
 };
+
 /**
- * Plugin the provides custom visualizations inside of a Panel.
+ * Plugin the provides custom visualizations inside a Panel.
  */
 export interface PanelPlugin<Spec = UnknownSpec> extends Plugin<Spec> {
   PanelComponent: React.ComponentType<PanelProps<Spec>>;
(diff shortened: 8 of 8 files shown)


ui/dashboards/src/validation/panel.ts
```diff
@@ -17,7 +17,10 @@ export const panelEditorValidationSchema = z.object({
   name: z.string().nonempty('Required'),
   groupId: z.number(),
   description: z.string().optional(),
-  type: z.string().nonempty('Required'),
+  selection: z.object({
+    type: z.string(),
```
sjcobb, line 21: Having both kind and type is confusing, I thought it was `kind: ‘Panel’`? Is `type` going in the spec too or is this is something different?

ui/panels-plugin/src/plugins/scatterplot/ScatterChart.ts
```diff
@@ -22,5 +22,6 @@ export const ScatterChart: PanelPlugin<ScatterChartOptions> = {
   PanelComponent: ScatterChartPanel,
   // TODO: add a chart options editor plugin, for example:
   // panelOptionsEditorComponents: [{ label: 'Settings', content: ScatterChartOptionsEditorSettings }],
+  supportedQueryTypes: ['TraceQuery'],
```
sjcobb, line 25: Mentioned this on a sync a while back but I think this panel should be renamed to TraceChart or TraceViewer and ‘scatter’ should be a visual.display option on TimeSeriesChart (similar to the ‘bar’ option)

ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
```diff
@@ -15,36 +15,55 @@ import { useState } from 'react';
...
+  const defaultQueryType = queryTypes[0]!;
+
+  // Then the default plugin kind
   // Use as default the plugin kind explicitly set as default or the first in the list
+  const { data: queryPlugins } = useListPluginMetadata(queryTypes);
   const { defaultPluginKinds } = usePluginRegistry();
-  const { data: timeSeriesPlugins } = useListPluginMetadata(TIME_SERIES_QUERY_KEY);
-  const defaultTimeSeriesQueryKind = defaultPluginKinds?.[TIME_SERIES_QUERY_KEY] ?? timeSeriesPlugins?.[0]?.kind ?? '';
-  const { data: defaultQueryPlugin } = usePlugin(TIME_SERIES_QUERY_KEY, defaultTimeSeriesQueryKind, {
+  const defaultQueryKind = defaultPluginKinds?.[defaultQueryType] ?? queryPlugins?.[0]?.kind ?? '';
```
sjcobb, line 39: Still works for Markdown panel too?

ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.tsx
```diff
@@ -54,44 +71,52 @@ export const TimeSeriesQueryInput = ({
...
-  const {
-    spec: { plugin },
-  } = value;
+  const { value, onChange, queryTypes, ...others } = props;
 
   const handlePluginChange: PluginEditorProps['onChange'] = (next) => {
     onChange(
       produce(value, (draft) => {
-        draft.spec.plugin = next;
+        draft.kind = next.selection.type as QueryPluginType;
```
sjcobb, line 101: It miight be possible to not need the `as QueryPluginType` assertion if you inline this below instead of a separate handler function

ui/plugin-system/src/components/PanelSpecEditor/PanelSpecEditor.tsx
```diff
@@ -44,40 +43,19 @@ export function PanelSpecEditor(props: PanelSpecEditorProps) {
...
-    }
-    const queryType: string = queriesList[0]?.kind;
-    return queryType;
-  };
-
-  // Get the corresponding queryEditor depending on the queryType
-  const getQueryEditorComponent = () => {
-    const queryType = getQueryType();
-    // default case handles cause where there is no queryType yet (e.g. UI > 'editing' mode > 'Add Panel')
-    switch (queryType) {
```
sjcobb, line 60: 🎉 much cleaner now that these lines can be removed

ui/plugin-system/src/components/PluginEditor/plugin-editor-api.ts
```diff
@@ -50,90 +60,103 @@ export type UsePluginEditorProps = Pick<PluginEditorProps, 'pluginType' | 'value
...
       // Otherwise, kick off the async loading process
-      setPendingKind(nextKind);
+      setPendingSelection(nextSelection);
     }
 
     if (
-      pluginType === 'Panel' &&
-      hideQueryState.current[nextKind] !== undefined &&
-      hideQueryState.current[value.kind] !== hideQueryState.current[nextKind]
+      nextSelection.type === 'Panel' &&
```
sjcobb, line 155: Similar question as earlier, I’m not totally following when kind is used vs. type

I’m not against it, just curious what the guidelines are on when to use either

ui/plugin-system/src/components/PluginKindSelect/PluginKindSelect.tsx
```diff
@@ -12,31 +12,67 @@
...
  */
 export const PluginKindSelect = forwardRef((props: PluginKindSelectProps, ref) => {
-  const { pluginType, value: propValue, ...others } = props;
-  const { data, isLoading } = useListPluginMetadata(pluginType);
+  const { pluginTypes, value: propValue, onChange: propOnChange, ...others } = props;
+  const { data, isLoading } = useListPluginMetadata(pluginTypes);
 
   // Pass an empty value while options are still loading so MUI doesn't complain about us using an "out of range" value
-  const value = propValue !== '' && isLoading ? '' : propValue;
+  const value = propValue && isLoading ? '' : JSON.stringify(propValue);
```
sjcobb, line 38: I’m not totally following these JSON.stringify lines 🤔

sjcobb (approved):
Overall looks good! I left mostly nits and questions but nothing blocking

ui/plugin-system/src/components/PluginKindSelect/PluginKindSelect.tsx
celian-garcia: Well, you took the last PR of the thread, but there are three commits, each depending on the previous, and each commit having its own PR ^^'
This is being discussed here: https://github.com/perses/perses/pull/1887#discussion_r1557670040

ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
celian-garcia: MultiQueryEditor is not used in markdown panel as in this panel, there is a hideQueryEditor boolean. 
Oh! now I'm realizing that we might be able to remove this hideQueryEditor parameter to replace it by queryTypes is empty condition

celian-garcia: It makes me think as well that we might be able to remove the supportedQueryTypes field in favor of a regular tab 

from this 
```ts
export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
  PanelComponent: TimeSeriesChartPanel,
  panelOptionsEditorComponents: [{ label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }],
  supportedQueryTypes: ['TimeSeriesQuery'],
  createInitialOptions: createInitialTimeSeriesChartOptions,
};
```

To this (not sure it compiles but you see the idea)
```ts
function buildQueryEditor(queryTypes) {
  return function (props) {
    return <MultiQueryEditor queryTypes={queryTypes} {...props} />
  }

}
/**
 * The core TimeSeriesChart panel plugin for Perses.
 */
export const TimeSeriesChart: PanelPlugin<TimeSeriesChartOptions> = {
  PanelComponent: TimeSeriesChartPanel,
  panelOptionsEditorComponents: [
    { label: 'Query', content: buildQueryEditor(["TimeSeriesQuery"]) },
    { label: 'Settings', content: TimeSeriesChartOptionsEditorSettings }
  ],
  createInitialOptions: createInitialTimeSeriesChartOptions,
};
```

ui/dashboards/src/validation/panel.ts
celian-garcia: answered there https://github.com/perses/perses/pull/1899#discussion_r1560413645

ui/panels-plugin/src/plugins/scatterplot/ScatterChart.ts
celian-garcia: And so scatter would be a visual option on the TraceChart ? This is interesting because it rises the question of preferring to have one chart by query and support multiple visualization, or have one chart by visualization and support multiple queries. 

I'd prefer the latter as sounds more modular but I didn't know that already the TimeSeriesChart was supporting bar display! Is it reusing `BarChart` panel or a complete different implementation.

ui/plugin-system/src/components/PluginKindSelect/PluginKindSelect.tsx
sjcobb: 😅 lol sorry! Guess I should have read that matrix message before jumping in to review, I’ll try to find time to start at the beginning of the stack but definitely don’t wait on me, looks like Gabriel is giving a more thorough review anyways

sjcobb: I’ve started using Graphite lately btw, makes it easy to follow along with stacked PRs, although probably wouldn’t have helped me here since I should have just read your PR description 🤦‍♂️

ui/panels-plugin/src/plugins/scatterplot/ScatterChart.ts
sjcobb: Oh ECharts makes it super easy to switch the series type using their dataset concept, all you do is change between ‘line’, ‘bar’, and ‘scatter’ in that case since they’re all ‘time’ xAxis, basically a one line change plus tweaking their option to adjust styling: 

https://github.com/perses/perses/blob/187d88aa40a261f6d777ce72392222b90accd880/ui/panels-plugin/src/plugins/time-series-chart/utils/data-transform.ts#L148

Our BarChart uses ‘category’ xAxis though so it’s a bit different (more like Datadog’s “Top List” viz)

sjcobb: So wayy back TimeSeriesChart was LineChart, if sticking with ScatterChart approach, maybe it should be renamed back, here was that PR #575 and it links to a Matrix chat about it 

Definitely tradeoffs either way, I was one pushing for LineChart back then so it’s funny now that I’m on the other side. Whichever way we go, it should be all-in though and consistent across the core panel kinds

ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
sjcobb: Yeah that could be a lot to expose to plugin developers, although I guess we do already expose a lot as is

Now that I’m thinking about it, I wonder if putting supported data types on the actual metadata here where the plugin is registered would have made this any easier: https://github.com/perses/perses/blob/main/ui/panels-plugin/plugin.json#L10

celian-garcia: Yes actually this would even be complicate for me to manage the state from the spec editor if I'd do that. 

Regarding the place of supportedQueryTyprs we thought about it with Antoine and having it in the plugin.json might allow the backend to make some validation (if it exposes it, it can read it) that wouldn't be a bad idea

ui/panels-plugin/src/plugins/scatterplot/ScatterChart.ts
celian-garcia: Okay I slept on it and what I think is that finally the bar chart as it is right now is even completely timeseries tainted so at the end if we do that the bar chart alone does not make any sense? All of this will be improved anyway when we'll tackle the problem that I exposed on last sync meeting. Summarize it all is in my todo list. 

Waiting for that, Iwill do a separate Pr to align: ScatterChart to TraceChart as you suggested

ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.tsx
celian-garcia: I'm realizing that it's not needed at all. I'm not sure why I did it in the first place but I just remove it :)
Thanks for the trick though. I mark it as resolved

jgbernalp (approved):

celian-garcia:
SupportedQueryTypes in the ts file or in the plugin.json. Let's move this discussion to another day when the delocalized plugins system will be more mature. I'm merging it 

Thanks to the reviewers <3

PR #1901, opened 2024-04-12
[BREAKINGCHANGE] OAuth & OIDC /token endpoint should respect the oauth 2.0 RFC

celian-garcia:
# Description

This PR is following the idea that we should be able to use classic oauth libs to authenticate to Perses. 
It went from a robotic use cases that we have in Amadeus where we want to use client credentials flow and we would like to use the clientcredentials package from oauth2 golang library. 

Thanks to this PR, we can easily use the go perses client + oauth2 clientcredentials library to request Perses with clientcredentials authentication.
```go
// Initialize a perses client
client, err := perseshttp.NewFromConfig(perseshttp.RestConfigClient{
	URL: "http://localhost:8080",
})

// Initialize a client credentials config from standard go oauth2 library
oauthConfig := &clientcredentials.Config{
	ClientID:     "MyClientID",
	ClientSecret: "MyClientSecret",
	TokenURL:     "http://localhost:8080/api/v1/auth/providers/oauth/github/token",
	AuthStyle:    oauth2.AuthStyleInHeader,
}
// Update the perses http client under the hood to be the one generated from oauth2 library
// It will perform automatically the login requests when needed
client.Client = oauthConfig.Client(context.Background())

// Make any classic Perses query 
_, err = v1.NewWithClient(client).Project().List("")
```

I think it's legit and particularly if somebody want to request perses API with another language than go.

Changes:
- [breaking change] accessToken/refreshToken fields changed to access_token/refresh_token to comply with the oauth 2.0 specs https://datatracker.ietf.org/doc/html/rfc6749#section-5.1
  It's not giving token_type (REQUIRED) and expires_in (RECOMMENDED) but this is probably something we should think of. Token type should be always "bearer" in our case but seems not used by the lib and expires_in would serve to the libraries (For example currently the oauth2 clientcredentials library will make a GET /token *before each* HTTP query because no expires_in is provided)
- [e2e test] a simple nominal http server of an external oauth provider
- [e2e test] a "simple" (need to be simplified) nominal http server of an external OIDC provider 
- [e2e test] added some test of the endpoints + go client to login with an external oauth/OIDC provider
- [e2e test] a new parameter to testFunc allow to have access to the generated perses server (useful if we want to know the base URL)

# Screenshots

# Checklist

- [x] Add some tests with oauth/OIDC provider 
  - [x] API // Done but very complex. Specially the OIDC part ... I need to simplify
  - [x] Go client // Partially. Did only for the clientcredentials flow
  - [x] percli // Done introducing integration tests with CLI (meaning running the Perses backend)


pkg/client/api/auth.go
```diff
@@ -70,42 +75,63 @@ func (c *auth) Refresh(refreshToken string) (*api.AuthResponse, error) {
...
+	return &api.AuthResponse{
+		AccessToken:  token.AccessToken,
+		RefreshToken: token.RefreshToken,
+	}, nil
 }
 
 func (c *auth) ClientCredentialsToken(authKind, slugID, clientID, clientSecret string) (*api.AuthResponse, error) {
-	body := &api.TokenRequest{
-		GrantType:    api.GrantTypeClientCredentials,
+	config := &clientcredentials.Config{
```
celian-garcia, line 116: This method should not be used out of the percli context actually. 

If a user want to programmatically request perses (with the GO SDK for example), this should be something like that. 

For now this config is not integrated into the ``RestConfigClient`` struct but we might want to add it.
```go
  client, err := perseshttp.NewFromConfig(config.RestConfigClient)
	if err != nil {
		return nil, err
	}

  oauthConfig := &clientcredentials.Config{
		ClientID:     clientID,
		ClientSecret: clientSecret,
		TokenURL:     c.tokenURL(authKind, slugID),
		AuthStyle:    oauth2.AuthStyleInHeader,
	}
	client.Client = oauthConfig.Client(context.Background())
  return client, err
```

Nexucis:
I think overall the changes looks reasonable to me and respecting the RFC is indeed a good thing.

It would be great to have some tests here, just to be sure everything is working fine as it took us around 2 release to have something stable on the auth aspect (in the backend or in the CLI)

pkg/client/api/auth.go
Nexucis: that's where I'm struggling to understand your point.

the perses http client could be used (like you mention) with the usage of the GO SDK. Since this method is exposed I don't get why it is supposed to work only for the CLI.

If it is really the case, then we have an issue somewhere. the package `client/api` is supposed to be a standalone library

celian-garcia: It works for sure, but I'm telling it's designed for the CLI 

Because when we are in a golang code, that have a persistent context in memory we should use the golang oauth2 library to generate an http client that **automatically** manages the generating of the token when requesting. That's to comply with the specification of that lib that I'm doing these changes.

pkg/model/api/auth.go
```diff
@@ -43,12 +43,12 @@ func (u *Auth) UnmarshalJSON(data []byte) error {
 }
 
 type AuthResponse struct {
-	AccessToken  string `json:"accessToken"`
```
Nexucis, line 46: Could you add a note saying the json field name format is respecting the Oauth/OIDC Spec which is different from the convention we have for json field in Perses ?

Nexucis (approved):
excepting my last comment, it looks ok to me.

Thank you for the hard working on this topic @celian-garcia !
Will be cool to have it finally merged !

pkg/model/api/auth.go
celian-garcia: Commented the struct

Nexucis (approved):

Issue #1907, opened 2024-04-15
CLI: unable to use flag `--all` for the command `get` when no read permission is given


Nexucis:
### What did you do?

0. I'm using the Perses dev environment
1. Give default guest permission to be able to only create a project (so no read permission on anything)
2. login with the `guest` user

```bash
./bin/percli login --username=guest http://localhost:8080
```

3. Create a project for the user `guest`

```bash
project='{
  "kind": "Project",
  "metadata": {
    "name": "MyProject"
  }
}'
echo ${project} | ./bin/percli apply -f -
```

4. get the roles for the project `myproject`

```bash
 ./bin/percli get roles --project=myproject
   NAME  |  PROJECT  | AGE | ACTION |    SCOPE
---------+-----------+-----+--------+--------------
  editor | myproject | 24m | *      | Dashboard
         |           |     |        | Datasource
         |           |     |        | Folder
         |           |     |        | Secret
         |           |     |        | Variable
         |           |     | read   | Project
         |           |     |        | Role
         |           |     |        | RoleBinding
  owner  | myproject | 24m | *      | *
  viewer | myproject | 24m | read   | *

```

5. Do the same thing but with the flag `--all`

### What did you expect to see?

I would have expected to see the same thing than at the same step 4

### What did you see instead? Under which circumstances?

The CLI is returning the following error:

```bash
Error: something wrong happened with the request to the API.  
Message: unauthorized: missing 'read' permission in '' project for 'Role' kind StatusCode: 401
```

Of course it doesn't concern only the roles but any kind of resources. Option  `--all`` should return anything I have access to.

### System information

Darwin 23.4.0 arm64

### Perses version

v0.44.0

### Perses configuration file

```yaml
security:
  readonly: false
  encryption_key: "<secret>"
  enable_auth: true
  authorization:
    guest_permissions:
      - actions:
          - create
        scopes:
          - Project
  authentication:
    providers:
      enable_native: true
```

PR #1911, opened 2024-04-15
[ignore] use generic for every api services and the toolbox struct

Nexucis:
This PR is updating every services / endpoints to use generics.

It removes unnecessary cast / type check. Using direct type in the API services will be helpful for the issue #1907 (specially for the `List` method as we will be able to modify the query struct directly with the data coming from the parameter)


Gladorme (approved):
Cool!

AntoineThebaud (approved):

PR #1914, opened 2024-04-17
[ignore] provide a custom prometheus registry to register the metrics

Nexucis:
This should help to have a consistant way to register the metrics across the application and across the packages from `perses/common`.

Let's wait #1889 before merging this one.

### DIFF
--- a/internal/api/core/core.go
+++ b/internal/api/core/core.go
@@ -34,7 +34,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func New(conf config.Config, banner string, registry *prometheus.Registry) (*app.Runner, dependency.PersistenceManager, error) {
+func New(conf config.Config, enablePprof bool, registry *prometheus.Registry, banner string) (*app.Runner, dependency.PersistenceManager, error) {
 	persistenceManager, err := dependency.NewPersistenceManager(conf.Database)
 	if err != nil {
 		logrus.WithError(err).Fatal("unable to instantiate the persistence manager")
@@ -83,6 +83,7 @@ func New(conf config.Config, banner string, registry *prometheus.Registry) (*app
 
 	// register the API
 	runner.HTTPServerBuilder().
+		ActivatePprof(enablePprof).
 		APIRegistration(persesAPI).
 		GzipSkipper(func(c echo.Context) bool {
 			// let's skip the gzip compression when using the proxy and rely on the datasource behind.
--- a/internal/api/core/core.go
+++ b/internal/api/core/core.go
@@ -27,13 +27,14 @@ import (
 	"github.com/perses/perses/internal/api/provisioning"
 	"github.com/perses/perses/internal/api/rbac"
 	"github.com/perses/perses/internal/api/schemas"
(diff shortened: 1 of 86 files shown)


internal/api/core/core.go
```diff
@@ -47,7 +48,7 @@ func New(conf config.Config, banner string) (*app.Runner, dependency.Persistence
 	}
 	persesAPI := NewPersesAPI(serviceManager, persistenceManager, conf)
 	persesFrontend := ui.NewPersesFrontend()
-	runner := app.NewRunner().WithDefaultHTTPServer("perses").SetBanner(banner)
+	runner := app.NewRunner().WithDefaultHTTPServerAndPrometheusRegisterer(metricNamespace, registry, registry).SetBanner(banner)
```
celian-garcia, line 51: Now I'm realising that I might have been wrong in my perses/common review this is actually a registry and not registerer... Btw why are you giving it twice there ?

celian-garcia (commented):
I was just curious about usage of it. I think you should make a pass to the perses/common to rename all usages of registerer to registry actually WDYT?

internal/api/core/core.go
Nexucis: that's because I had to modify (again) the `perses/common` lib to add another parameter called `Gatherer`.
Here the prometheus registry is implementing the `Register` and the `Gatherer` interface. And it seems if you provide a different object of each interface, the metrics doesn't appear in `/metrics`.

It's really weird we need to do that. But meh

Nexucis:
@sinkingpoint let me know if you find the metrics registration flow clearer like that !

internal/api/core/core.go
```diff
@@ -29,10 +29,11 @@ import (
 	"github.com/perses/perses/internal/api/schemas"
 	"github.com/perses/perses/pkg/model/api/config"
 	"github.com/perses/perses/ui"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
-func New(conf config.Config, banner string) (*app.Runner, dependency.PersistenceManager, error) {
+func New(conf config.Config, banner string, metricNamespace string, registry *prometheus.Registry) (*app.Runner, dependency.PersistenceManager, error) {
```
sinkingpoint, line 36: Rather than passing `metricNamespace` here, would it make sense to just use `utils.MetricNamespace` directly? `view` already uses `utils.MetricNamespace` so it would be nice to be consistent

sinkingpoint:
This is much more explicit and much less opaque about how metrics get registered - thanks!

internal/api/core/core.go
Nexucis: ah yeah totally I forgot to change that ! Thanks :)

internal/api/core/core.go
Nexucis: well that's what I have observed, but if there is two different interfaces, that probably means you can provide two different implementations. I just didn't get then how to use them.
So for the moment I prefer to let `perses/common` the possibility to expose both interface.

PR #1920, opened 2024-04-18
[FEATURE] CLI: add cmd whoami

Nexucis:
can be used as follow:

```bash
./bin/percli whoami --show-token --show-url
Token used: eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiIsImV4cCI6MTcxMzI3OTg3MiwibmJmIjoxNzEzMjc4OTcyfQ.mecrHuc4FVMLLNcOfc_2IEPlZ85BZJrtO8FAWFKieZV4rg_b32iiLLcfcOEv36h6QL5EVizSg7XLA5UxCk-m2Q
Authenticated to the server: http://localhost:8080
User used: admin
```

### DIFF
--- a/internal/cli/cmd/whoami/whoami.go
+++ b/internal/cli/cmd/whoami/whoami.go
@@ -1,0 +1,92 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


AntoineThebaud (changes requested):
cli.md to update with the new command ;)

internal/cli/cmd/whoami/whoami.go
```diff
@@ -0,0 +1,92 @@
...
+	o := &option{}
+	cmd := &cobra.Command{
+		Use:   "whoami",
+		Short: "Display current user used",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return persesCMD.Run(o, cmd, args)
+		},
+	}
+	cmd.Flags().BoolVar(&o.showToken, "show-token", o.showToken, "Print the token the current session is using.")
+	cmd.Flags().BoolVar(&o.showURL, "show-url", o.showURL, "If true, print the current server's REST API URL.")
```
celian-garcia, line 90: Be consistent in the wording. Like choose between ``Print`` or ``If true, print``. I would be in favor of just `Print`

AntoineThebaud (approved):

PR #1927, opened 2024-04-19
[BUGFIX] Fix grafana migration failing for "custom" template variables

AntoineThebaud:
# Description

Rework the migration logic of StaticListVariable to avoid migration failure: relying on the `options` field was causing the following error when migrating dashboards retrieved directly from the Grafana database (not from UI):

```
error="something wrong happened with the request to the API.  Message: bad request: unable to convert to Perses dashboard: spec.variables.0.spec.plugin.spec.values: undefined field: options (and 1 more errors) StatusCode: 400"
```

This was happening because the `options` field is actually made-up by the frontend from the `query` field, it doesn't get saved in db.

Also I took advantage of this fix to enhance the StaticListVariable migration logic, to no longer lose the "key" when the `key : value` syntax is used.

# Checklist

### DIFF
--- a/docs/plugin/cue.md
+++ b/docs/plugin/cue.md
@@ -87,9 +87,7 @@ A variable migration file looks like the following:
 if #var.type == "custom" || #var.type == "interval" {
     kind: "StaticListVariable"
     spec: {
-        values: [ for _, option in #var.options if option.value !~ "^\\$__.*$" { // the if filters out the grafana global vars that'd be causing validation issues later (e.g "__auto_interval_sampling is used but not defined")
-            option.value
-        }]
+        values: strings.Split(#var.query, ",")
     }
 },
 ```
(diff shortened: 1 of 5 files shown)


docs/plugin/cue.md
```diff
@@ -87,9 +87,7 @@ A variable migration file looks like the following:
 if #var.type == "custom" || #var.type == "interval" {
     kind: "StaticListVariable"
     spec: {
-        values: [ for _, option in #var.options if option.value !~ "^\\$__.*$" { // the if filters out the grafana global vars that'd be causing validation issues later (e.g "__auto_interval_sampling is used but not defined")
-            option.value
-        }]
+        values: strings.Split(#var.query, ",")
```
AntoineThebaud, line 90: Here this is just an example to explain the plugin format, so we don't care if that's not the real logic used by the StaticListVariable plugin in reality (better to keep it simple in this part of the doc).

celian-garcia (approved):
Blind approved :D is a the end we don't have the issue I'm ok with that !

PR #1900, opened 2024-04-11
[FEATURE] Project can be renamed

Gladorme:
# Description

Allow users to rename projects using a display name as dashboards. Maybe we should implement this on all kinds 😄 

# Screenshots

// TODO

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [ ] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [ ] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [ ] Visual tests are stable and unlikely to be flaky.
  See [Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests)
  and [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues
  include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

PR #1928, opened 2024-04-19
[BUGFIX] Fix dashboard list with empty project on home page

Gladorme:
# Description

With PR #1900, empty projects are now listed on the home page. However if a project is empty, it will raise an error when opening the accordion as the API returns `null`. Idea for later: maybe returns an empty list? 😄 

# Screenshots

# Checklist


celian-garcia (approved):
That would be good if the backend was not returning null instead of []. This is coming from the way golang handle nil | empty list. I think we have fixed it in our home made tool.

Nexucis:
if the API is returning `null` instead of an empty array for a dashboard list there is a regression somewhere.

Gladorme:
@Nexucis I can't reproduce, I guess the issue was link to #1932

PR #1932, opened 2024-04-19
[BUGFIX] Fix project list endpoints returning wrong data

Gladorme:
# Description

Currently API is returning wrong data for `List` project requests. It's fixing the issue

# Checklist

### DIFF
--- a/internal/api/rbac/cache.go
+++ b/internal/api/rbac/cache.go
@@ -66,6 +66,10 @@ func (r *cacheImpl) IsEnabled() bool {
 }
 
 func (r *cacheImpl) GetUserProjects(user string, requestAction v1Role.Action, requestScope v1Role.Scope) []string {
+	if permissionListHasPermission(r.guestPermissions, requestAction, requestScope) {
+		return []string{GlobalProject}
+	}
+
 	projectPermission := r.cache.permissions[user]
 	if globalPermissions, ok := projectPermission[GlobalProject]; ok && permissionListHasPermission(globalPermissions, requestAction, requestScope) {
 		return []string{GlobalProject}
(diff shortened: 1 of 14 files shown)


internal/api/rbac/cache.go
```diff
@@ -66,6 +66,10 @@ func (r *cacheImpl) IsEnabled() bool {
 }
 
 func (r *cacheImpl) GetUserProjects(user string, requestAction v1Role.Action, requestScope v1Role.Scope) []string {
+	if permissionListHasPermission(r.guestPermissions, requestAction, requestScope) {
```
celian-garcia, line 69: What does it mean permissionListHasPermission ? is it a strings.Contains like ?

celian-garcia (commented):
Just dropped a comment for my personal knowledge but I feel like the toolbox is getting too complex for my brain. 

As I said you in private also a comment would be welcomed. 

Overall I'm not fan of these service changes. If that could be made on a upper level like toolbox that would be good?

internal/api/rbac/cache.go
Gladorme: Similar yes, however it check for wildcards too

Gladorme:
> Just dropped a comment for my personal knowledge but I feel like the toolbox is getting too complex for my brain.
> 
> As I said you in private also a comment would be welcomed.
> 
> Overall I'm not fan of these service changes. If that could be made on a upper level like toolbox that would be good?

Not possible to put at toolbox level as it's a generic. Would like a better solution too, but we only found this solution with @Nexucis

celian-garcia (approved):

Nexucis (approved):
thanks for fixing the hell I put there

PR #1933, opened 2024-04-19
[FEATURE] Activate pprof by flag

Nexucis:
Would avoid to modify and compile the code when trying to profile the application.


internal/api/schemas/loader.go
```diff
@@ -66,13 +66,14 @@ func (c *cueDefs) Load() error {
 		file := files[i]
 		if !file.IsDir() {
 			logrus.Warningf("Plugin %s will not be loaded: it is not a folder", file.Name())
+			i++
```
celian-garcia, line 69: 😱

celian-garcia (approved):

Issue #1923, opened 2024-04-18
Heap objects leak and memory usage


tristanmorgan:
### What did you do?

Hi,

I am running a single instance of perses with the default file database and a single dashboard and it is experiencing periodic termination due to OOM issues.

### What did you expect to see?

I expect to see a general flattening of the number heap objects and memory usage after a while with a subtle saw tooth pattern due to GC activity.

### What did you see instead? Under which circumstances?

I have noticed the number of objects on the heap correlated with the excess memory used.

![obligatory Perses screenshot](https://github.com/perses/perses/assets/6336384/d325cb48-142b-4705-96d3-e48be70f18c7)

[head of heap dump](https://github.com/perses/perses/files/15031380/perses-steaming-heap.txt)

### System information

Linux 6.1.21-v8+ aarch64

### Perses version

0.44.0

### Perses configuration file

```yaml
database:
  file:
    folder: "/local/perses"
    extension: "json"
        
schemas:
  panels_path: "/etc/perses/cue/schemas/panels"
  queries_path: "/etc/perses/cue/schemas/queries"
  datasources_path: "/etc/perses/cue/schemas/datasources"
  variables_path: "/etc/perses/cue/schemas/variables"
  interval: "5m"
```

### Logs

```text
2024-04-14T07:12:17.008+1000 [ERROR] client.driver_mgr.docker: OOM Killed: container_id=9149b04e72cc99c2943dc8ccd62d18b94e9a1890648bbb553c902fe2d519f830 driver=docker container_id=9149b04e72cc99c2943dc8ccd62d18b94e9a1890648bbb553c902fe2d519f830 container_image=sha256:90665c15d83725d62a8692933d2768d73a0d80d9ee7c70bbd866811491191745 nomad_job_name=perses nomad_task_name=perses nomad_alloc_id=22bcd979-81af-a709-a260-d7f5c5fb82ae
 client.driver_mgr.docker: OOM Killed: container_id=9149b04e72cc99c2943dc8ccd62d18b94e9a1890648bbb553c902fe2d519f830 driver=docker container_id=9149b04e72cc99c2943dc8ccd62d18b94e9a1890648bbb553c902fe2d519f830 container_image=sha256:90665c15d83725d62a8692933d2768d73a0d80d9ee7c70bbd866811491191745 nomad_job_name=perses nomad_task_name=perses nomad_alloc_id=22bcd979-81af-a709-a260-d7f5c5fb82ae
[...]

PR #1935, opened 2024-04-22
[BREAKINGCHANGE] Deactivate cron task that reloads the cue schemas

Nexucis:
Following the issue #1923 regarding the memory leaks we have when calling frequently the package cuelang to reload the schemas, I'm proposing to deactivate these cron.

As it's not possible to add at runtime a new plugin, likely we won't need to change the schemas at runtime as well. So there is no need to frequently reloading the schemas for the moment.

Besides we still have a file watching on the schema folder. So the functionality is partially degraded.

Once we have a proper plugin flow, these crons will be reactivated.

Monitoring is showing the memory leak is mitigated with this PR:

(for the same range of time (30min))
Before:

image

After:

image

### DIFF
--- a/internal/api/core/core.go
+++ b/internal/api/core/core.go
@@ -14,7 +14,6 @@
 package core
 
 import (
-	"context"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


Gladorme (approved):

celian-garcia (commented):
Just a comment, we probably need to ensure the watch works well with k8s env. By experience there is a special thing that needs to be done in that case as k8s have a way to update a file very specific with symlinks updates.

Nexucis:
hopefully that should work as well in k8s since we are watching the folder and the files.

Issue #1917, opened 2024-04-17
percli apply should not accept silently a different project in the json and in the -p


celian-garcia:
### What did you do?

Create a dashboard-with-project-b.json with project ProjectB
Run ``percli apply -d dashboard-with-project-b.json -p ProjectA``

### What did you expect to see?

I would expect an error saying that the projects are not aligned.

We could say that the project from the context takes precedence but it can be confusing if the project `ProjectA` is already saved in the context file and we do ``percli apply -f dashboard-with-project-b.json`` it would save in `ProjectA`. This is not better.

### What did you see instead? Under which circumstances?

Currently this is silently creating/updating the dashboard in the **ProjectB** (the one from the json file)

### Perses version

0.44.0

ying-jeanne:
hi @celian-garcia i opened a tantative PR regard this bug, if you could take a look. Thank you.

PR #1937, opened 2024-04-22
[ENHANCEMENT] Add --force option to allow applying resource despite project config inconsistencies

ying-jeanne:
[ENHANCEMENT] Add --force option to allow applying resource despite project config inconsistencies

# Description
Fix: https://github.com/perses/perses/issues/1917, take a look at the problem, i am uncertain if this solution aligns with our objectives. The proposed PR addition of a force flag aims to alert users to potential project inconsistencies while still permitting them to create based on input metadata.

# Screenshots

# Checklist

## UI Changes

- na Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- na Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- na Visual tests are stable and unlikely to be flaky.
  include:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/goreleaser/goreleaser/v2 v2.2.0
 	github.com/gorilla/securecookie v1.1.2
+	github.com/hashicorp/go-multierror v1.1.1
 	github.com/huandu/go-sqlbuilder v1.30.1
 	github.com/kylelemons/godebug v1.1.0
 	github.com/labstack/echo-jwt/v4 v4.2.0
@@ -102,6 +103,7 @@ require (
 	github.com/goreleaser/fileglob v1.3.0 // indirect
 	github.com/goreleaser/nfpm/v2 v2.39.0 // indirect
 	github.com/gorilla/websocket v1.5.1 // indirect
+	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/huandu/xstrings v1.4.0 // indirect
 	github.com/imdario/mergo v0.3.16 // indirect
 	github.com/imkira/go-interpol v1.1.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	cuelang.org/go v0.11.0-alpha.3
 	github.com/brunoga/deep v1.2.4
 	github.com/charmbracelet/huh v0.6.0
+	github.com/efficientgo/core v1.0.0-rc.3
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/gavv/httpexpect/v2 v2.16.0
 	github.com/go-jose/go-jose/v4 v4.0.4
@@ -14,7 +15,6 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/goreleaser/goreleaser/v2 v2.2.0
 	github.com/gorilla/securecookie v1.1.2
-	github.com/hashicorp/go-multierror v1.1.1
 	github.com/huandu/go-sqlbuilder v1.30.1
 	github.com/kylelemons/godebug v1.1.0
 	github.com/labstack/echo-jwt/v4 v4.2.0
@@ -103,7 +103,6 @@ require (
 	github.com/goreleaser/fileglob v1.3.0 // indirect
 	github.com/goreleaser/nfpm/v2 v2.39.0 // indirect
 	github.com/gorilla/websocket v1.5.1 // indirect
-	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/huandu/xstrings v1.4.0 // indirect
 	github.com/imdario/mergo v0.3.16 // indirect
 	github.com/imkira/go-interpol v1.1.0 // indirect
--- a/internal/cli/cmd/apply/apply.go
+++ b/internal/cli/cmd/apply/apply.go
@@ -72,7 +72,7 @@ func (o *option) validateProjectConsistency() error {
 	for _, entity := range o.entities {
 		kind := modelV1.Kind(entity.GetKind())
 		project := resource.GetProject(entity.GetMetadata(), o.Project)
-		if !o.forceCreate && len(o.Project) != 0 && project != o.Project {
+		if len(o.Project) != 0 && project != o.Project {
 			result.Add(fmt.Errorf("\ninconsistency has been detected for object %q %q: while the metadata suggests the project name %q, the currently selected project indicates %q, \nuse the '--force' flag to bypass consistency check", kind, entity.GetMetadata().GetName(), project, o.Project))
 		}
 	}
--- a/internal/cli/cmd/apply/apply.go
+++ b/internal/cli/cmd/apply/apply.go
@@ -16,7 +16,9 @@ package apply
 import (
 	"fmt"
 	"io"
+	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	persesCMD "github.com/perses/perses/internal/cli/cmd"
 	"github.com/perses/perses/internal/cli/config"
 	"github.com/perses/perses/internal/cli/file"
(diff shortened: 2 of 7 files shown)


celian-garcia (changes requested):
Hello @ying-jeanne, I'm not sure yet that we want to implement it like that. The force flag if used should be consistent in its behaviour on all the commands and here we have a behaviour for apply only. 

I'd see at a glance that we should reuse --all which is always used for "do this command for all projects" it would probably fit better the use case. @Nexucis WDYT ?

celian-garcia:
Also I think that an error message should be displayed at start of the command to prevent from applying half of the resources

Nexucis:
Hello @ying-jeanne !

Thank you for taking the time to fix this issue :).

> I'd see at a glance that we should reuse --all which is always used for "do this command for all projects" it would probably fit better the use case. @Nexucis WDYT ?

usage of `--all` in the command `apply` would be weird.

Reading `percli apply -f data.json --all`, for me means I want to apply all data coming from the file `data.json`. Which would mean we are partially applying the resources from the file.

From my perspective, introducing a flag like `--force` is a nice way to solve it. If we don't like the meaning of `--force`, we could have instead the flag `--ignore.project-set`

> Also I think that an error message should be displayed at start of the command to prevent from applying half of the resources

That's something not possible currently as to be able to detect the inconsistency we need to loop other all resources that the user want to apply. 

The only way to do that, is to loop a first time to search this issue.

celian-garcia:
I always see the ``force`` flag as something that you shouldn't do but as an unusual emergency case. Here it can be a valid daily case. To just to not have to manage your different projects separately.

I'm wondering after all if a warning wouldn't be sufficient.

Note for the overthinking buddies like me: Also the warnings in general could be hideable. Like if we had a common system of warnings in this case saying "project-inconsistency: you applied in the project 'hello' which is different from current selected project 'moto'" and a common --ignore flag that you could use to ignore that specific warning --ignore project-inconsistency, that would be good 
End of note for the overthinking buddies like me...

ying-jeanne:
Hi @celian-garcia @Nexucis both, thanks for the review 💟 . I would make it draft for the moment, since I realised later yesterday besides the --froce flag, It would be better to apply this consistency check in validate() function instead and introduce a multi-errors in order to go through all resources and return all inconsistencies in the response just as you described above. Will continue later this week and we can continue the discussion afterwards.

ying-jeanne:
I usually lean towards --force option since it's my go-to when I'm confident about my actions after reviewing what just happened/warned. However, I'm also receptive to considering --ignore.project-set. Personally, I don't have very strong preference here. In the updated code I've included the multierr functionality and relocated the validation logic to the validate() method. If you would like to review it again, just let me know wdyt.

internal/cli/cmd/apply/apply.go
```diff
@@ -65,7 +68,30 @@ func (o *option) Complete(args []string) error {
 }
 
 func (o *option) Validate() error {
-	return nil
+	var result *multierror.Error
+	for _, entity := range o.entities {
+		kind := modelV1.Kind(entity.GetKind())
+		project := resource.GetProject(entity.GetMetadata(), o.Project)
+		if !o.ForceCreate && len(o.Project) != 0 && project != o.Project {
+			result = multierror.Append(result, fmt.Errorf("inconsistency has been detected for object %q %q: While the JSON metadata suggests the project name %q, the command-line indicates %q,", kind, entity.GetMetadata().GetName(), project, o.Project))
```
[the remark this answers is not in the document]
ying-jeanne: Hi this makes sense, updated PR accordingly

celian-garcia:
Wow crazy I've been able to resolve the conflicts from Github UI directly despite the fact your PR comes from your fork :astonished: 
Github will always surprise me. <3

So @ying-jeanne thank you very much! Let's move forward and merge it, your changes are very nice.

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/goreleaser/goreleaser v1.26.2
 	github.com/gorilla/securecookie v1.1.2
+	github.com/hashicorp/go-multierror v1.1.1
```
Nexucis, line 18: Sorry for the long time it took me to go back at this PR.

The only thing I'm not ok with is the usage of library coming from hashicorp. (regarding the changes they made on their license), and besides introducing a new lib for managing an array of error, I find it a bit overkill.

internal/cli/opt/opt.go
```diff
@@ -105,3 +105,11 @@ func (o *ProjectOption) Complete() error {
 func AddProjectFlags(cmd *cobra.Command, o *ProjectOption) {
 	cmd.Flags().StringVarP(&o.Project, "project", "p", o.Project, "If present, the project scope for this CLI request")
 }
+
+type ForceCreateOption struct {
```
Nexucis, line 109: Do you see an usage of this flag in other command ? If no, probably it's not worth it to create a struct to handle this flag

ying-jeanne:
> Wow crazy I've been able to resolve the conflicts from Github UI directly despite the fact your PR comes from your fork 😲 Github will always surprise me. <3
> 
> So @ying-jeanne thank you very much! Let's move forward and merge it, your changes are very nice.

I signed share my fork with project maintainer! This is nice because back and force is so painful <3 thank you for the review @celian-garcia

go.mod
ying-jeanne: I hesitated creating a helper multierror in the project directly, look at go-multierror codebase, it is light with very few updates/CVEs. I understand each projects have their own preference, if you prefer less dependency, I could replace it.

internal/cli/cmd/apply/apply.go
```diff
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/efficientgo/core/merrors"
```
ying-jeanne, line 20: while doing something in prometheus go client, found this lib wrote by our dear bartek, efficient and lighter than multierror

internal/cli/cmd/apply/apply.go
```diff
@@ -65,7 +68,30 @@ func (o *option) Complete(args []string) error {
 }
 
 func (o *option) Validate() error {
-	return nil
+	var result *multierror.Error
```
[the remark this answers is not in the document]
ying-jeanne: changed in 7c25ff5396abc6515bf8fef305b78c5312981481

internal/cli/opt/opt.go
ying-jeanne: removed in 7c25ff5396abc6515bf8fef305b78c5312981481

internal/cli/cmd/apply/apply.go
```diff
@@ -65,7 +67,23 @@ func (o *option) Complete(args []string) error {
 	return o.setEntities()
 }
 
+func (o *option) validateProjectConsistency() error {
+	var result = merrors.New()
+	for _, entity := range o.entities {
+		kind := modelV1.Kind(entity.GetKind())
+		project := resource.GetProject(entity.GetMetadata(), o.Project)
+		if !o.forceCreate && len(o.Project) != 0 && project != o.Project {
```
celian-garcia, line 75: You can remove the !o.forceCreate condition from there.

celian-garcia (approved):
Thanks @ying-jeanne for reviving this! Just added a nit comment but not very important. 
I'm personally ok with this new lib, but I'd wait for the @Nexucis opinion as it was his concern.

Nexucis:
Looks better indeed.
Thank you for your permanent work on this PR @ying-jeanne and sorry for the long delay it took to review / gives opinion.

Once last comment from @celian-garcia are tackled I am fine to merge it.

Nexucis:
if by any chance you pass here to fix it soon @ying-jeanne , would be cool to merge it before making the new release !

ying-jeanne:
comments addressed in commit [8d2f589](https://github.com/perses/perses/pull/1937/commits/8d2f589c45771cd4731912a59c5cc1190c190d8f) (:

Nexucis (approved):
awesome thanks !

PR #1941, opened 2024-04-22
[ignore] update go deps

Nexucis:


Nexucis:
going forward on this one. It's super straightforward anyway

PR #1961, opened 2024-04-29
[BUGFIX] Reset the plaintext password on a failed create

### LINKED ISSUE #1960: Can't log in to provisioned users
### What did you do?

1. Clone the repo
2. Start the UI with `npm run start`
3. Start the backend with `./scripts/api_backend_dev.sh`
4. Access Perses at localhost:3000
5. Try to log in with username: `admin`, password: `password`
6. Observe that the log in fails

I should be able to log in

image

### System information

_No response_

### Perses version

0.44.0

### Perses configuration file

_No response_

### Logs

```text
INFO[2024-04-27T16:30:05+10:00]                                               method=POST status=400 uri=/api/auth/providers/native/login
WARN[2024-04-27T16:30:19+10:00] password do not match                         error="crypto/bcrypt: hashedPassword is not the hash of the given password"
```
```

sinkingpoint:
# Description

This fixes #1960

If the user creation fails, then the user
passed into the `Create` function is tainted because we've overridden the password with a hashed version.

This sets it back to the plaintext password
in the case of a failure, allowing whatever called `Create` to handle the error and potentially
retry.

# Checklist

### DIFF
--- a/internal/api/impl/v1/user/service.go
+++ b/internal/api/impl/v1/user/service.go
@@ -1,4 +1,4 @@
-// Copyright 2023 The Perses Authors
+// Copyright 2024 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


Nexucis:
I was looking how to solve this issue as well and it's interesting you didn't rollback the way the provisionner is working. Previously it was getting the data from the DB to know if it exists. And based on that it decided if it needs to update the data or to create.

If we want to keep the current provisionner implementation, and instead change how the services are working we would need to copy the entity to avoid changing it.
I'm saying that because this issue (i.e. the fact a service is changing / encrypting the data inside the entity) is occurring also in the secret/global secret service

On a smaller issue, every services are also changing the creation and update date of the entity.

Since in Golang we cannot copy an object easily, I'm thinking it might be easier to just call a `get` in the provisionner to know if we need to create or update the resource.

If we don't want that, probably then we should create a method `Copy` for every objects and call it in every service before starting anything.
Probably it can be useful for other usage to have this method Copy, I don't know. What do you think @sinkingpoint ?

Nexucis:
I'm trying this [lib](https://github.com/brunoga/deep) and it gaves nice result. It also make a bit better the code responsible of the query list. I think I will propose a PR with that

sinkingpoint:
I'm not attached to this tbh, it just seemed like the least disruptive fix - happy to close it in favor of yours which is more comprehensive anyway

PR #1967, opened 2024-04-29
[ignore] use deep copy in services

### LINKED ISSUE #1960: Can't log in to provisioned users

Nexucis:
Counter proposition to #1961.

I'm proposing to make a deep copy of an entity in every services before creating or updating the entity. Like that we ensure that the service can modify the object without modifying the initial reference.

As a side effect it also fixes #1960


Nexucis:
@sinkingpoint let me know what do you think about this way

internal/api/impl/v1/dashboard/service.go
```diff
@@ -42,47 +45,55 @@ func NewService(dao dashboard.DAO, globalVarDAO globalvariable.DAO, projectVarDA
 }
 
 func (s *service) Create(_ apiInterface.PersesContext, entity *v1.Dashboard) (*v1.Dashboard, error) {
+	copyEntity, err := deep.Copy(entity)
```
celian-garcia, line 48: To be safer and ensure we never use the original entity, you could encapsulate all the create logic into a create function and the copy would stay in the Create WDYT ?

Nexucis: ah yes good idea

celian-garcia (approved):

PR #1971, opened 2024-04-30
[BREAKINGCHANGE] Introduce a dedicated config for the frontend

Nexucis:
I'm introducing a dedicated config for the frontend.

Later it will be interesting as we want to provide a way to configure the au-refresh period value.

It's a breaking change as the config `info` and `important_dashboard` are moving under the new field `frontend`

### DIFF
--- a/ui/app/src/model/config-client.ts
+++ b/ui/app/src/model/config-client.ts
@@ -130,13 +130,17 @@ export interface SecurityConfig {
   authentication: AuthenticationConfig;
 }
 
+export interface FrontendConfig {
+  important_dashboards?: DashboardSelector[];
+  information?: string;
+}
+
 export interface ConfigModel {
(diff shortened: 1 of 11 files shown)


ui/app/src/model/config-client.ts
```diff
@@ -130,13 +130,17 @@ export interface SecurityConfig {
...
+}
+
 export interface ConfigModel {
   security: SecurityConfig;
   database: Database;
   schemas: ConfigSchemasModel;
-  important_dashboards?: DashboardSelector[];
-  information?: string;
   provisioning?: ProvisioningConfig;
+  Frontend: FrontendConfig;
```
Gladorme, line 143:
```suggestion
  frontend: FrontendConfig;
```

celian-garcia (approved):

PR #1974, opened 2024-05-03
[FEATURE] Explore traces support + introduce Graph tab

celian-garcia:
# Description

Getting my hands on the instant query to make two tabs Table / Graph like it's done in prometheus, I was coding already so many file moves that I wanted to create a PR to save it. 

It's been also easy to add a new tab for the switch between traces and timeseries so let's stamp it before continuing...

# Screenshots

image
image
image

# Checklist

## UI Changes

  include:


ui/explore/src/components/ExploreManager/TimeSeriesExplorer.tsx
```diff
@@ -0,0 +1,70 @@
+// Copyright 2024 The Perses Authors
```
AntoineThebaud, line 1: Since you named the other TracesExplorer, imho this one should be called, or at least contain "Metrics" in it. If I understood well you're preparing the ground to add a new tab for instant querying of metrics, so in the end I see two ways to go here:

1. 3 tabs: "Traces", "Metrics - range", "Metrics - instant"
2. 2 tabs: "Traces" and "Metrics" + a switch between range & instant for metrics within the tab panel

Personally I like the idea to have a first high-level choice to make between traces, metrics & logs, but it would be adding 1 click in the user's journey.

ui/explore/src/components/ExploreManager/ExploreManager.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { exploreTitleComponent } = props;
+  const [tabState, setTabState] = useState(0);
+
+  return (
+    <Stack sx={{ width: '100%' }} px={2} pb={2} pt={1.5} gap={3}>
+      <ExploreToolbar exploreTitleComponent={exploreTitleComponent} />
+
+      <Stack direction="row" gap={2}>
+        <Tabs
+          orientation="vertical"
```
AntoineThebaud, line 34: Not a fan of the vertical tabs here, we're losing space on the left & to display graphs it's always a good thing to be able to take more space 😄

ui/explore/src/components/ExploreManager/TimeSeriesExplorer.tsx
celian-garcia: Yes solution 2. There will be also some query params to prevent from having to select and allow to share

celian-garcia: I will change to metrics ;)

ui/explore/src/components/ExploreManager/ExploreManager.tsx
celian-garcia: In my opinion it lets a large space for the graph. I added a media adapter to be responsive in case we are in a mobile, but I don't know what I can do more. 

Maybe if we could reduce the tabs with a << that would change them to pictograms?

AntoineThebaud: Well I was thinking to go for horizontal tabs in all the cases. I also feel this would be better for overall UX consistency, since we went for this design for the admin & project pages. But having this kind of switch based on screen size is also fine for me.
Btw I'm just thinking about it now, what if you want to investigate both metrics & traces at the same time? Does it keep the other-tab's context when you are switching tabs?

celian-garcia: Funny that you talk about using the explorer for traces and metrics at the same time as it's been a reflex by Guillaume as well. 

To answer directly to your question, the context is lost if you switch from one tab to another, but it could be interesting to prevent that like saving the data in memory. We do this kind of thing in the query selection to not loose the previous query. I will think about it and make another PR for that.

celian-garcia: As discussed, updated with a card to highlight better the main content. Screenshots has been updated. 
I'm also adding to my todo list to save the last queries before switching tab.

AntoineThebaud (approved):
LGTM

Issue #1456, opened 2023-10-02
Prometheus Instant Query Support


nicolastakashi:
### Proposal

Currently, Perses only supports Prometheus Range Queries as part as the Explore Page and other on going discussions around the Perses ecosystem, we should provide the Instant Query capability.

## Dependencies

- [x] https://github.com/perses/perses/issues/1303

celian-garcia:
For reference, here is what would remain after this PR to consider the Prometheus Instant query is fully supported (be able to use it in all timeseries panels with a switch or something)
https://github.com/perses/perses/pull/1982#discussion_r1596726647

Nexucis:
> Just putting back porting my comment I made in the PR linked: 

I don't think it's a good idea to mixed panel in the same plugin.

What would be the benefit of doing that ? Excepting in an explorer mode, I don't see why in a dashboard you would need to switch from a range query to an instant query.

celian-garcia:
> back porting my answer ^^

This might be not useful in TimeSeriesChart yes, or at least I don't have an idea of the use case. But when you do use a gauge chart for example. You may want to have the choice between using a range query and display the average/last/min etc or make an instant query that would be more performant if you know you want the last.

celian-garcia:
I'm closing this issue as from the original description, the capability to make instant query is technically there now. Thanks to the supported query field and the "mode" instant or range, a plugin developer can make a switch if needed, putting the "mode" value in the panel specs if it needs to be saved

PR #1982, opened 2024-05-06
[FEATURE] Instant query table view

celian-garcia:
# Description

ref #1456 (not fix as it's not addressing all the problems. Only the explorer one)

Introduction of the ``TimeSeriesTable`` panel. It allow users to display "prometheus like" table results as a panel in a dashboard. It is also used in TimeSeries explorer.

image

This led to the creation of several new fields but no **breaking change**. 

### New ``mode`` field in ``DataQueriesProvider`` options
The new mode field is used to tell the DataQueriesProvider to execute the time series queries in `instant` mode (e.g /query endpoint used for prometheus)  or `range` mode (e.g /range endpoint used for prometheus)
Default mode is ``range``.

```html
<DataQueriesProvider definitions={definitions} options={{ mode: 'instant' }}>
```

### New ``queryOptions`` in ``PanelPlugin``
This allow to declare the mode at panel plugin level. Basically for now only the timeseries table panel will tell the queries to be done on instant mode.

```ts
// Example for timeseries table
export const TimeSeriesTable: PanelPlugin<TimeSeriesTableOptions> = {
  PanelComponent: TimeSeriesTablePanel,
  supportedQueryTypes: ['TimeSeriesQuery'],
  queryOptions: {
    mode: 'instant',
  },
  createInitialOptions: () => {
    return {};
  },
};
```

### A new ``hideHeader`` field in ``PanelOptions``
A minor change also is that you hide panel header which can be useful in embedded mode, and reduce the space taken by the panel in the explorer.
```html
<Panel
  panelOptions={{
    hideHeader: true,
  }}
  definition={{...}}
/>
```

### Explorer improvement
Used in explorer, the new timeseries table chart looks like this
image
image
image
image

# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/cue/schemas/panels/time-series-table/time-series-table.cue
+++ b/cue/schemas/panels/time-series-table/time-series-table.cue
@@ -1,4 +1,4 @@
-// Copyright 2023 The Perses Authors
+// Copyright 2024 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -11,10 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export * from './plugins/time-series-chart';
-export * from './plugins/gauge-chart';
-export * from './plugins/stat-chart';
-export * from './plugins/markdown';
-export * from './plugins/bar-chart';
-export * from './plugins/scatterplot';
-export * from './plugins/time-series-table';
+package model
+
+kind: "TimeSeriesTable"
+spec: close({})
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { useState, useMemo, memo } from 'react';
+import { useState, useMemo, memo, ReactNode } from 'react';
 import useResizeObserver from 'use-resize-observer';
 import { ErrorBoundary, ErrorAlert, combineSx, useId, useChartsTheme } from '@perses-dev/components';
 import { PanelDefinition } from '@perses-dev/core';
@@ -29,10 +29,15 @@ export interface PanelProps extends CardProps<'section'> {
 
 export type PanelOptions = {
   /**
-   * Content to render in the top-right corner of the panel. It will only be
-   * rendered when the panel is in edit mode.
+   * Allow you to hide the panel header if desired.
+   * This can be useful in embedded mode for example.
    */
-  extra?: (props: PanelExtraProps) => React.ReactNode;
+  hideHeader?: boolean;
+  /**
+   * Content to render in right of the panel header. (top right of the panel)
+   * It will only be rendered when the panel is in edit mode.
+   */
+  extra?: (props: PanelExtraProps) => ReactNode;
 };
 
 export type PanelExtraProps = {
@@ -95,15 +100,17 @@ export const Panel = memo(function Panel(props: PanelProps) {
       data-testid="panel"
       {...others}
     >
-      <PanelHeader
-        extra={panelOptions?.extra?.({ panelDefinition: definition, panelGroupItemId })}
-        id={headerId}
-        title={definition.spec.display.name}
-        description={definition.spec.display.description}
-        editHandlers={editHandlers}
-        links={definition.spec.links}
-        sx={{ paddingX: `${chartsTheme.container.padding.default}px` }}
-      />
+      {!panelOptions?.hideHeader && (
+        <PanelHeader
+          extra={panelOptions?.extra?.({ panelDefinition: definition, panelGroupItemId })}
+          id={headerId}
+          title={definition.spec.display.name}
+          description={definition.spec.display.description}
+          editHandlers={editHandlers}
+          links={definition.spec.links}
+          sx={{ paddingX: `${chartsTheme.container.padding.default}px` }}
+        />
+      )}
       <CardContent
         component="figure"
         sx={{
--- a/ui/explore/src/components/ExploreManager/MetricsExplorer.tsx
+++ b/ui/explore/src/components/ExploreManager/MetricsExplorer.tsx
@@ -15,10 +15,9 @@ import { useState } from 'react';
 import { QueryDefinition } from '@perses-dev/core';
 import { Box, Stack, Tab, Tabs } from '@mui/material';
 import { DataQueriesProvider, MultiQueryEditor, useSuggestedStepMs } from '@perses-dev/plugin-system';
-import { ErrorAlert, ErrorBoundary } from '@perses-dev/components';
-import { TimeSeriesChart } from '@perses-dev/panels-plugin';
 import useResizeObserver from 'use-resize-observer';
-import { PANEL_PREVIEW_DEFAULT_WIDTH, PANEL_PREVIEW_HEIGHT } from './constants';
+import { Panel } from '@perses-dev/dashboards';
+import { PANEL_PREVIEW_HEIGHT } from './constants';
 
 function TimeSeriesPanel({ queries }: { queries: QueryDefinition[] }) {
   const { width, ref: boxRef } = useResizeObserver();
@@ -38,10 +37,43 @@ function TimeSeriesPanel({ queries }: { queries: QueryDefinition[] }) {
 
   return (
     <Box ref={boxRef} height={height}>
-      <DataQueriesProvider definitions={definitions} options={{ suggestedStepMs }}>
-        <TimeSeriesChart.PanelComponent
-          contentDimensions={{ width: width ?? PANEL_PREVIEW_DEFAULT_WIDTH, height }}
-          spec={{}}
+      <DataQueriesProvider definitions={definitions} options={{ suggestedStepMs, mode: 'range' }}>
+        <Panel
+          panelOptions={{
+            hideHeader: true,
+          }}
+          definition={{
+            kind: 'Panel',
+            spec: { queries: queries, display: { name: '' }, plugin: { kind: 'TimeSeriesChart', spec: {} } },
+          }}
+        />
+      </DataQueriesProvider>
+    </Box>
+  );
+}
+
+function MetricDataTable({ queries }: { queries: QueryDefinition[] }) {
+  const height = PANEL_PREVIEW_HEIGHT;
+
+  // map TimeSeriesQueryDefinition to Definition<UnknownSpec>
+  const definitions = (queries ?? []).map((query) => {
+    return {
+      kind: query.spec.plugin.kind,
+      spec: query.spec.plugin.spec,
+    };
+  });
+
+  return (
+    <Box height={height}>
+      <DataQueriesProvider definitions={definitions} options={{ mode: 'instant' }}>
+        <Panel
+          panelOptions={{
+            hideHeader: true,
+          }}
+          definition={{
+            kind: 'Panel',
+            spec: { queries: queries, display: { name: '' }, plugin: { kind: 'TimeSeriesTable', spec: {} } },
+          }}
         />
       </DataQueriesProvider>
     </Box>
@@ -59,14 +91,14 @@ export function MetricsExplorer() {
         value={tabState}
         onChange={(_, state) => setTabState(state)}
         variant="scrollable"
-        sx={{ borderRight: 1, borderColor: 'divider' }}
+        sx={{ borderBottom: 1, borderColor: 'divider' }}
       >
+        <Tab label="Table" />
         <Tab label="Graph" />
       </Tabs>
       <Stack gap={1}>
-        <ErrorBoundary FallbackComponent={ErrorAlert}>
-          {tabState === 0 && <TimeSeriesPanel queries={queries ?? []} />}
-        </ErrorBoundary>
+        {tabState === 0 && <MetricDataTable queries={queries ?? []} />}
+        {tabState === 1 && <TimeSeriesPanel queries={queries ?? []} />}
       </Stack>
     </Stack>
   );
--- a/ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
@@ -1,0 +1,118 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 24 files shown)


ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
```diff
@@ -337,14 +337,9 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps) {
...
   // At this point, we have a response from the backend for all queries. (We're past loading === true).
-  // If we don't data from any of the queries, then check if we want to show an error.
-  // Put another way: If any queries have data, even if other queries failed, we will show the data (not the error).
-  // Unfortunately, partial errors get swallowed in this case.
-  // This could be refactored when someone takes a look at validation and error-handling.
-  if (!hasData) {
-    for (const result of queryResults) {
-      if (result.error) throw result.error;
-    }
+  // Then check if we want to show an error, no matter if we have data or not.
```
celian-garcia, line 340: This one is to be checked further. I had to do this because when switching from table to graph with a query valid for table but not for graph, the result was not showing the error. There was data and also error. 

Maybe a better investigation would be: why there was data if the result was an error, maybe a merge result that tried to merge from previous.

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -95,15 +95,17 @@ export const Panel = memo(function Panel(props: PanelProps) {
...
-      <PanelHeader
-        extra={panelOptions?.extra?.({ panelDefinition: definition, panelGroupItemId })}
-        id={headerId}
-        title={definition.spec.display.name}
-        description={definition.spec.display.description}
-        editHandlers={editHandlers}
-        links={definition.spec.links}
-        sx={{ paddingX: `${chartsTheme.container.padding.default}px` }}
-      />
+      {(panelOptions?.extra || definition.spec.display.name != '' || definition.spec.links) && (
```
celian-garcia, line 98: It would be probably better to have an extra option dedicated to hide the title

ui/panels-plugin/src/plugins/time-series-chart/TimeSeriesChartPanel.tsx
celian-garcia: I found out my problem I was not putting mode in the query key of the use queries. After adding it I can get back this code as it was.

ui/panels-plugin/src/plugins/time-series-table/TimeSeriesTable.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { PanelPlugin, PanelProps } from '@perses-dev/plugin-system';
+import { TimeSeriesTablePanel } from './TimeSeriesTablePanel';
+
+interface TimeSeriesTableOptions {}
+
+/**
+ * The core TimeSeriesTable panel plugin for Perses.
+ */
+export const TimeSeriesTable: PanelPlugin<TimeSeriesTableOptions> = {
```
celian-garcia, line 22: I need to find a way to set by default the TimeSeriesTable

Also, in other PRs, we can support the instant/range switch for other Charts like TimeSeriesChart, GaugeChart etc...

celian-garcia: First part solved introducing the queryOptions field here.

celian-garcia: Second part is more complex as we would need then to save the switch selection in query specs

ui/dashboards/src/components/Panel/Panel.tsx
celian-garcia: Fixed introducing the hideHeader panel option

nicolastakashi:
Congrats on this amazing work @celian-garcia 🥳

cue/schemas/panels/time-series-table/time-series-table.cue
```diff
@@ -0,0 +1,17 @@
+// Copyright 2023 The Perses Authors
```
nicolastakashi, line 1: Should we change to 2024?

celian-garcia: everywhere ? yeah probably... conflicts with pending PR incoming :smile:

celian-garcia: I will at least change this one which is new :+1:

ui/explore/src/components/ExploreManager/MetricsExplorer.tsx
```diff
@@ -38,10 +37,45 @@ function TimeSeriesPanel({ queries }: { queries: QueryDefinition[] }) {
...
+      </DataQueriesProvider>
+    </Box>
+  );
+}
+
+function MetricDataTable({ queries }: { queries: QueryDefinition[] }) {
+  const height = PANEL_PREVIEW_HEIGHT;
+
+  // map TimeSeriesQueryDefinition to Definition<UnknownSpec>
+  const definitions = queries.length
```
Gladorme, line 59: (queries ?? []).map((query) => {
    return {
        kind: query.spec.plugin.kind,
        spec: query.spec.plugin.spec,
    };
})

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+ * https://github.com/prometheus/prometheus/blob/2524a915915d7eb1b1207152d2e0ce5771193404/web/ui/react-app/src/pages/graph/DataTable.tsx
+ * @param result timeseries query result
+ * @constructor
+ */
+const DataTable = ({ result }: DataTableProps) => {
+  if (!result) {
+    return null;
+  }
+  const series = result.flatMap((d) => d.data).flatMap((d) => d?.series || []);
+  const maxFormattableSize = 1000;
```
Gladorme, line 38: Add comment explaining this value / var please

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+  let rows: ReactNode[] = [];
+  let limited = false;
+  const doFormat = series.length <= maxFormattableSize;
+  rows = series.map((s, seriesIdx) => {
```
Gladorme, line 42: Maybe use a dedicated function for readability

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+      </Table>
+    </>
+  );
+};
+
+const leftDelim = (br: number): string => (br === 3 || br === 1 ? '[' : '(');
+const rightDelim = (br: number): string => (br === 3 || br === 0 ? ']' : ')');
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const bucketRangeString = ([boundaryRule, leftBoundary, rightBoundary, _]: [
```
Gladorme, line 99: Because of the "\_" I guess we need to tweak the rule to skip check on "\_" vars

ui/plugin-system/src/model/panels.ts
```diff
@@ -34,6 +35,7 @@ export interface PanelPlugin<Spec = UnknownSpec> extends Plugin<Spec> {
    * @default [] (no query types supported) only relevant if hideQueryEditor is true
    */
   supportedQueryTypes?: QueryPluginType[];
+  queryOptions?: QueryOptions;
```
Gladorme, line 38: Small comment for explaining

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/get-time-series-data.ts
```diff
@@ -141,3 +145,67 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
...
+  }
+
+  switch (data.resultType) {
+    case 'vector':
+      return buildVectorData(data, query, seriesNameFormat);
+    case 'matrix':
+      return buildMatrixData(data, query, seriesNameFormat);
+    case 'scalar':
+      return buildScalarData(data, query, seriesNameFormat);
+    default:
```
Gladorme, line 208: Maybe log warning in console too?

Gladorme (commented):
Cool, not easy to review, I probably missed some stuff.

Btw, I found an issue when writing a wrong query on the explorer:
- Write wrong query => Graph error
- Fix query => Graph not updated still in error
When trying on variable form, the preview is well updated

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
celian-garcia: I have no idea, I copied that from Prometheus :D :D :D

celian-garcia:
> Cool, not easy to review, I probably missed some stuff.
> 
> Btw, I found an issue when writing a wrong query on the explorer:
> 
> * Write wrong query => Graph error
> * Fix query => Graph not updated still in error
>   When trying on variable form, the preview is well updated

Thank you Guillaume for the review! I will check the behavior described. I was sure it worked, but I focused so much on making it redraw when we switch tab that I might have forgot some simpler case... Let me try to draft a little test reproducing it as I feel it's easy to break.

Generally from your review, the Datatable is a copy-paste of Prometheus and should be probably reworked. I just didn't change anything but checkstyle matters, in order to easily backport any possible update on Prometheus side. I will write it down into a comment.

ui/plugin-system/src/model/panels.ts
celian-garcia: Updated wdyt?

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
```diff
@@ -0,0 +1,131 @@
+// Copyright 2024 The Perses Authors
```
AntoineThebaud, line 1: (Not sure where it would go in this file :D) Any chance to have some visual difference (e.g color and/or boldness) between label values and label names, like prometheus is doing? 
image

At least for me it's a good helper for readability 😄

ui/explore/src/components/ExploreManager/MetricsExplorer.tsx
```diff
@@ -38,10 +37,43 @@ function TimeSeriesPanel({ queries }: { queries: QueryDefinition[] }) {
...
+            spec: { queries: queries, display: { name: '' }, plugin: { kind: 'TimeSeriesChart', spec: {} } },
+          }}
+        />
+      </DataQueriesProvider>
+    </Box>
+  );
+}
+
+function MetricDataTable({ queries }: { queries: QueryDefinition[] }) {
+  const height = PANEL_PREVIEW_HEIGHT;
```
AntoineThebaud, line 56: IMHO the height should be increased in this case, we're quickly reaching the max height with just a few results here - and contrary to the Dashboard view, you cannot resize any panel height to adjust:

image

On prom UI they went for infinite expand, I dont know if we want to go this way (it causes other navigation issues when you have multiple queries returning multiple results), but I think as a first version we can go with the current PANEL_PREVIEW_HEIGHT as min height and at least the triple of that for max height

AntoineThebaud (commented):
Overall LGTM, just few remarks to improve the display (that could be tackled in a follow-up PR)

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
celian-garcia: Okay I see now, I took as input the formattedName which is just a string. I think I have the labels somewhere. Just need to do the same logic as them rebuilding the from labels

Thanks to that I discovered a bug in my implementation haha 
image

celian-garcia: argh no ok this is something new legend use the same than me..
That's interesting because I wouldn't expect to have this result in this case : 
image
I'd expect {} and not the repeat of the query. Strange

celian-garcia: As discussed in private, I will let it like that for now. And make a separate PR to handle that directly from labels. 

It can be fancy as well to extend this formatting to legends and tooltips. IMHO I don't really understand why this part is in the prometheus plugin. Is it really proper to prometheus to make such a display of labels? I would personally move it up to a shareable component. Free to use by any timeseries panel and taking labels + optional formatter to render a pretty line. So that Legend, Tooltip, and TimeSeriesTable could reuse it and have the same result.

celian-garcia: Regarding the default value to being the query. This is an assume choice made before us. https://github.com/perses/perses/pull/1045

I'd disagree on that but let's discuss in the future PR that will possibly change this behaviour.

ui/explore/src/components/ExploreManager/MetricsExplorer.tsx
celian-garcia: I'd like to have that control on the height of a panel but unfortunately the first element of the ``Panel`` is saying height=100% already. 

It forces me to give a value and I cannot know the height of the child component below ``Panel``. 

The best solution is to introduce some params to the ``Panel`` component to optionally disable this height 100% and give more control on min height / max height.

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
celian-garcia: Updated the eslint rules

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
celian-garcia: done

ui/panels-plugin/src/plugins/time-series-table/DataTable.tsx
celian-garcia: I did a pass on what we are using or not from the prom implem and I removed it, as well as the limit. 

Performance considerations can be tackled in a separate PR.

Gladorme (approved):
Cool! 🚀

ui/panels-plugin/src/plugins/time-series-table/TimeSeriesTable.ts
Nexucis: I don't think it's a good idea to mixed panel in the same plugin.

What would be the benefit of doing that ? Excepting in an explorer mode, I don't see why in a dashboard you would need to switch from a range query to an instant query.

celian-garcia: This might be not useful in TimeSeriesChart yes, or at least I don't have an idea of the use case. But when you do use a gauge chart for example. You may want to have the choice between using a range query and display the average/last/min etc or make an instant query that would be more performant if you know you want the last.

PR #1985, opened 2024-05-12
[DOC] Fix broken link to the contributing guide in the introduction page

Romain-Geissler-1A:
# Description

Fix broken link to the contributing guide in the introduction page.

[joke] Given that I contribute a very critical change in Perses, I guess I deserve my name in the MAINTAINERS file no ? :p [/joke]

# Checklist

### DIFF
--- a/docs/prologue/introduction.md
+++ b/docs/prologue/introduction.md
@@ -55,7 +55,7 @@ The project has guiding goals for its development:
 
 ## Chat with Perses project team?
 
-You can find the project team on Matrix in the #perses-dev channel, see the [contributing guide](../../CONTRIBUTING.md)
+You can find the project team on Matrix in the #perses-dev channel, see the [contributing guide](https://github.com/perses/perses/blob/main/CONTRIBUTING.md)
 for details.
 
 <img src="https://github-production-user-asset-6210df.s3.amazonaws.com/437001/280049367-00378850-a7a1-4176-9f71-1394f3ada470.png" alt="Matrix dev channel" style="width: 60%;"/>


celian-garcia (approved):

celian-garcia:
Haha thank you Romain! 😋
The maintainers file is not maintained. I'm wondering if we should maintain it... 

You for whom the opensources projects have no secret, this is useless no?

AntoineThebaud (approved):
LGTM, still I wonder if we could later make a relative link work here. It's quite convenient in the IDE when such link opens the local file instead of the origin's one, but I don't think we want to host CONTRIBUTING.md on the static website..

Romain-Geissler-1A:
> Haha thank you Romain! 😋 The maintainers file is not maintained. I'm wondering if we should maintain it...
> 
> You for whom the opensources projects have no secret, this is useless no?

(answering seriously for once ;) )

I have contributed (usually small useless trivial patches like this one) to about a hundred different repositories in diverse technologies, and the only thing I conclude is that there are as many open source policies as there are repositories. Everyone works differently and there is no "good" way which seems to be used by a majority. What we often see (including in our own proprietary software inside Amadeus) is that doc is often outdated the next day it was committed, so I personally trust only the source code. For the maintainers file, it shouldn't be hard to keep it up to date, I doubt it changes often anyway. Normally the time it was updated is when I passed by your office and laughed at @Gladorme because he we still not officially appointed maintainer :D So keep this file updated, it's not hard.

Finally: honestly people care much more about having the code maintained, and their issues fixed/investigated/replied to, and that you don't have a crazy huge contribution process, than looking at who maintains what ;) Just be friendly and care for your users, this is the most important thing, which I can tell you is lacking in many open source projects !

celian-garcia:
> Normally the last time it was updated is when I passed by your office and laughed at @Gladorme because he we still not officially appointed maintainer :D So keep this file updated, it's not hard.

That must be that yeah haha. Thanks for these wised words. We're making our best to make our new contributors feel good and eager to come back.

> LGTM, still I wonder if we could later make a relative link work here. It's quite convenient in the IDE when such link opens the local file instead of the origin's one, but I don't think we want to host CONTRIBUTING.md on the static website..

Sure! and you know I like this tool checking if the URL are still valid. Maybe there is an option to verify the relative links?

PR #1993, opened 2024-05-14
[ENHANCEMENT] Add query param 'metadata_only' on almost all query endpoints

Nexucis:
This PR is introducing the query param `metadata_only` on every endpoint excepting the `datasource` one.

On the datasource endpoint, the query param `default` and `kind` are used to filter the list. But for that we need to have access to the spec.

On the home page, when it's loading ~1k dashboards, it takes around 3s to load. With this PR it's taking 500ms.

### DIFF
--- a/internal/api/database/model/dao.go
+++ b/internal/api/database/model/dao.go
@@ -21,6 +21,7 @@ import (
 )
 
 type Query interface {
+	GetMetadataOnlyQueryParam() bool
 }
 
 type DAO interface {
--- a/ui/app/src/model/dashboard-client.ts
+++ b/ui/app/src/model/dashboard-client.ts
@@ -181,7 +181,9 @@ export function getDashboard(project: string, name: string) {
 }
 
 export function getDashboards(project?: string) {
-  const url = buildURL({ resource: resource, project: project });
+  const queryParams = new URLSearchParams();
+  queryParams.set('metadata_only', 'true');
+  const url = buildURL({ resource: resource, project: project, queryParams: queryParams });
   return fetchJson<DashboardResource[]>(url, {
     method: HTTPMethodGET,
     headers: HTTPHeader,
--- a/ui/app/src/model/dashboard-client.ts
+++ b/ui/app/src/model/dashboard-client.ts
@@ -57,9 +57,9 @@ export function useDashboard(project: string, name: string) {
  * Used to get dashboards in the API.
  * Will automatically be refreshed when cache is invalidated
  */
-export function useDashboardList(project?: string) {
+export function useDashboardList(project?: string, metadataOnly?: boolean) {
   return useQuery<DashboardResource[], Error>([resource, project], () => {
-    return getDashboards(project);
+    return getDashboards(project, metadataOnly);
   });
 }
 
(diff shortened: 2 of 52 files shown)


internal/api/database/model/dao.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 type Query interface {
+	GetMetadataOnlyQueryParam() bool
```
celian-garcia, line 24: Let's open the debate that we all expect: should it be ``metadata_only`` or something else? 

Food for think: 
What if the default was to actually give only the  metadata to prevent from dumb requests. You could have full=true for example in all the queries that actually want details

Gladorme: partial, full, skip_spec, ignore_spec, ...
Depends on the the default behavior too, "full=true" will not work with the current implem

celian-garcia: Yes that's what I'm saying. Changing the default behaviour to by default having full=false would in a sense give a message that if you use full=true you don't use default and don't complain if it fails or take time to answer, or is paginated.

Nexucis: When you get the list of data with no query parameter, you won't expect a partial response data.
What we can do is to provide a pagination so you cannot get everything excepting if you explicitly say it.

`metadata_only` is at least explicitly saying you will only get the metadata and nothing else.

Nexucis: but before including a pagination, I want to make everything to be able to get all data without crashing Perses.

ui/app/src/model/dashboard-client.ts
```diff
@@ -181,7 +181,9 @@ export function getDashboard(project: string, name: string) {
 }
 
 export function getDashboards(project?: string) {
```
Gladorme, line 183:
```suggestion
export function getDashboards(project?: string, metadata_only?: boolean) {
```

ui/app/src/model/dashboard-client.ts
```diff
@@ -181,7 +181,9 @@ export function getDashboard(project: string, name: string) {
...
-  const url = buildURL({ resource: resource, project: project });
+  const queryParams = new URLSearchParams();
+  queryParams.set('metadata_only', 'true');
+  const url = buildURL({ resource: resource, project: project, queryParams: queryParams });
```
Gladorme, line 186:
```suggestion
  let url = buildURL({ resource: resource, project: project });
  if (metadata_only) {
    const queryParams = new URLSearchParams();
    queryParams.set('metadata_only', 'true');
    url = buildURL({ resource: resource, project: project, queryParams: queryParams });
  }
```

Gladorme (commented):
Let's make UI functions more generic so we can request whatever we want

internal/api/database/model/dao.go
celian-garcia: Okay, just wanted to make sure we explore all the possibilities. 

I think I'd tend to be in favor of an API that **by default** gives everything and light than something very detailed and potentially paginated in the future. But at the end as long as the light mode is available I'm good with any of the default behaviour.

Gladorme (approved):
🚀

PR #1995, opened 2024-05-16
[ENHANCEMENT] on the backend side, query are parallelised when getting the list of resources

Nexucis:
I'm just parallelising the query when we get a full list of data per project. 

This drop again a bit the time it takes to get a response from the backend. When querying the metadata for the dashboard list we are around 50ms now for 1k dashboards

PR must be merged after #1994

### DIFF
--- a/internal/api/toolbox/list.go
+++ b/internal/api/toolbox/list.go
@@ -14,7 +14,11 @@
 package toolbox
 
 import (
+	"fmt"
+
+	"github.com/brunoga/deep"
 	"github.com/labstack/echo/v4"
+	"github.com/perses/common/async"
 	apiInterface "github.com/perses/perses/internal/api/interface"
 	"github.com/perses/perses/internal/api/rbac"
 	"github.com/perses/perses/pkg/model/api"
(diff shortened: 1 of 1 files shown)


internal/api/toolbox/list.go
```diff
@@ -81,12 +85,15 @@ func (t *toolbox[T, K, V]) listWhenPermissionIsActivated(ctx echo.Context, param
...
 
-	result := make([]any, 0)
+	result := make([]any, 0, len(projects))
+	asynchronousRequests := make([]async.Future[any], 0, len(projects))
 	for _, project := range projects {
-		parameters.Project = project
-		listResult, listErr := t.metadataOrFullList(persesContext, parameters, q)
-		if listErr != nil {
-			return nil, listErr
+		asynchronousRequests = append(asynchronousRequests, async.Async(t.asyncMetadataOrFullList(persesContext, parameters, project, q)))
```
AntoineThebaud, line 91: Did you check that it works well? I think it's typically the kind of code that was not working before Go 1.22 - after the loop is over `project` would have been equal to the latest iterated one. Now I guess it works like this, but just to be sure.

Nexucis: yeah I checked and I didn't get any issue !

AntoineThebaud (approved):

PR #1996, opened 2024-05-16
[BUGFIX] Fixing external ListVariable overriding dashboard ListVariable in query param

### LINKED ISSUE #1954: Overriden external ListVariable still takes precedence as query param
### What did you do?

Have an external (global or project, doesn't matter) ListVariable defined that return some results, then on a dashboard define a variable with the same name to override the previous one, choose a value for it, then press F5.

Since the local var overrides the external var, I expect the value in both the dropdown/input & the URL to remain identical & corresponding to what I chose.

The value in the dropdown/input is fine but at the end of the load the query param is taking the value of the first result returned by the external var.

_See var-greeting value changing to "bonjour", a value coming from the external var._ 

https://github.com/perses/perses/assets/7058693/f6f8d8c1-f598-44ed-8fd0-321a6be018ae

In the worst (maybe corner) case of overriding a list variable with a text variable, it can actually break a whole dashboard display, since the value in the query param takes precedence and replaces the one saved in the dashboard.

### System information

_No response_

### Perses version

0.45.0-rc0

### Perses configuration file

_No response_

### Logs

_No response_

Gladorme:
# Description

Doing a quick fix for issue #1954, I think there's still room for improvements. Lot of useEffects to remove in TemplateVariable.tsx + we should have a method returning all variables def a dashboard can use and remove variables definitions overrode. I will check when I have time 😄 

fixes #1954 
# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/dashboards/src/components/Variables/VariableList.tsx
+++ b/ui/dashboards/src/components/Variables/VariableList.tsx
@@ -13,6 +13,7 @@
 
 import { Box } from '@mui/material';
 import { VariableDefinition, VariableSpec } from '@perses-dev/core';
+import { useMemo } from 'react';
 import {
   ExternalVariableDefinition,
   useTemplateExternalVariableDefinitions,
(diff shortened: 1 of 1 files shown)


celian-garcia (commented):
I'm interested on that fix, can we call each other in 30min ?

celian-garcia (approved):

PR #2013, opened 2024-05-22
[ENHANCEMENT] Add formating to series name in explorer and use MUI Table

Gladorme:
# Description

Importing Prometheus series name formating to Perses explorer.
You can click on the label="value" for copying in the clipboard
Using same limit, if there is more than 1000 series, only show plain text time series name.
Use MUI Table + increase panel height

# Screenshots

image
image

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/explore/src/components/ExploreManager/constants.ts
+++ b/ui/explore/src/components/ExploreManager/constants.ts
@@ -11,5 +11,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export const PANEL_PREVIEW_HEIGHT = 300;
+export const PANEL_PREVIEW_HEIGHT = 700;
 export const PANEL_PREVIEW_DEFAULT_WIDTH = 840;
--- a/ui/panels-plugin/src/plugins/time-series-table/SeriesName.tsx
+++ b/ui/panels-plugin/src/plugins/time-series-table/SeriesName.tsx
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { Labels } from '@perses-dev/core';
-import { ReactElement, MouseEvent, ReactNode } from 'react';
+import { ReactElement, MouseEvent } from 'react';
 import { useSnackbar } from '@perses-dev/components';
 import { Typography } from '@mui/material';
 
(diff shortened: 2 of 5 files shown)


ui/panels-plugin/src/plugins/time-series-table/SeriesName.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+/*
+ * Display a series with labels in bold and with a copy to clipboard feature if isFormatted is enabled
+ * Else it will only display the series in plain text (mostly use for performance reasons)
+ */
+export function SeriesName({ name, labels, formattedName, isFormatted }: SeriesNameProps) {
+  if (isFormatted && labels) {
+    return <FormatedSeriesName labels={labels} />;
+  }
+  return <span>{formattedName ?? name}</span>;
```
celian-garcia, line 34: If possible, this should still be computed from labels as the formattedName is slightly different from your computation

Check ``sum by (notexisitinglabel) (up)`` result with and without formatting ;)

celian-garcia (changes requested):

ui/panels-plugin/src/plugins/time-series-table/SeriesName.tsx
Gladorme: Strange not the same behavior as prometheus 🤔

Gladorme: Hola @sjcobb I saw the PR #1085, any reason to not return `{}` as Prometheus when there is no label? 

I think it would be better to return the same result as Prometheus. And if user want to return the query, they can use the legend 😄 WDYT?

ui/explore/src/components/ExploreManager/constants.ts
```diff
@@ -11,5 +11,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export const PANEL_PREVIEW_HEIGHT = 300;
+export const PANEL_PREVIEW_HEIGHT = 700;
```
celian-garcia, line 14: This does not apply to graph ? isn't it weird in graph ?

Gladorme: I think it's okay: 
image

celian-garcia (approved):

Issue #74, opened 2021-08-20
Datasource discovery


Nexucis:
Instead of providing the direct URL to the Prometheus server, it could be interesting use a discovery (and so to have the corresponding config).

We can start to support k8s discovery. Others can come later

Nexucis:
This will modify the current `HTTPProxy` configuration by adding a new field `discovery`. It will also make the field `url` optional.

So the `HTTPProxy` definition is becoming:

```yaml
kind: "HTTPProxy"
spec:
  # URL is the url of datasource. It is not the url of the proxy.
  # Note: URL and discovery are mutually exclusive
  [ url: <url> ]

  # It is a tuple list of http methods and http endpoints that will be accessible.
  # Leave it empty if you don't want to restrict the access to the datasource.
  allowedEndpoints:
    - [ <Allowed Endpoints specification> ]

  # It can be used to provide additional headers that need to be forwarded when requesting the datasource
  headers:
    [ <string>: <string> ]
  # This is the name of the secret that should be used for the proxy or discovery configuration
  # It will contain any sensitive information such as password, token, certificate.
  # Please read the documentation about secrets to understand how to create one
  [ secret: <string> ]
  
  # This is the discovery configuration.
  [ discovery: <discovery_spec> ]
```

With the following `discovery_spec`:

```yaml
kind: "Kubernetes"
spec:
  # The namespace where the datasource is running
  namespace: <string>

  # The role Perses should use to contact the datasource
  role: <enum = "service" | "ingress">

  # The name of the service or of the ingress.
  name: <string>

  # The portName used when contacting the datasource
  # It's not necessary to define it
  [ portName: <string> ]
  [ portNumber: <number> ]
```

For the moment, I believe we don't need to support discovery by labels since we want to target a single service / ingress. So setting the name of the service will insure we are finding a single service and not multiple one, which is possible when finding service by labels.

Nexucis:
based on discussion on Matrix, there was the question about would it be possible to discover a list of datasource in a single configuration.

From my opinion, it makes more sense rather than discovering a single datasource with a discovery. But for this use case, (discovering a list of datasource), maybe we need to generate the datasource based on the discovery then.

Does it mean we need another object ? Like `DatasourceDiscovery` 

Does it matter to have a discovery configuration in the datasource object directly that can replace the http config ?

celian-garcia:
[...]

PR #2021, opened 2024-05-23
[FEATURE] Provide datasource discovery starting with http discovery

Nexucis:
This PR introduces the notion of datasource discovery.

I'm starting with a HTTP discovery so we have something generic. The idea is that you can expose an endpoint returning a list of global datasource and Perses will fetch it and inject the data in the database.

I think for the discovery it makes to create only global datasource because that's something you are doing in the backend config. So for me it doesn't really make sense to create data in a project using the discovery. But maybe I am wrong.

Linked to #74

### DIFF
--- a/internal/api/core/core.go
+++ b/internal/api/core/core.go
@@ -15,6 +15,7 @@ package core
 
 import (
 	"fmt"
+	"github.com/perses/perses/internal/api/discovery"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


Nexucis:
I'm thinking we should support at least also the Kubernetes discovery. So we have two services discovery and it will highlight what is missing in this current implementation and what is needed to be in common.

Also I need to find how to test this kind of service. It will be too hard to test manually

Gladorme (approved):
LGTM, however I think it missing some tests around the feature 😄

AntoineThebaud:
Relates to https://github.com/perses/perses/issues/74

PR #2033, opened 2024-05-28
[FEATURE] New TracingGanttChart panel based on Jaeger UI

andreasgerstmayr:
# Description

This PR adds a new `TracingGanttChart` panel, based on the Jaeger UI Gantt chart.

I'm currently working on:
- [ ] switch from Jaeger to OpenTelemetry data model (https://opentelemetry.io/docs/specs/otel/overview/#tracing-signal)
- [ ] fix the selection in the mini Gantt chart and the draggable divider (between span names and gantt chart)

### Open questions

**UI/UX**
We need to update the explore page to show a table of all matched traces/spans. After clicking on a trace, the Gantt chart should show up. Currently the Gantt Chart is embedded in a panel and occupies only half the browser's height because of the query editors on top - ideally there should be more space for the Gantt chart (in Jaeger UI, it's displayed fullscreen).

**CSS**
Jaeger UI uses global CSS files. This can be a problem when other Perses components use the same `id`s or CSS classes. Ideally we'll switch to CSS modules.
Currently I'm `rsync`'ing the css files to the `dist` folder in the `build:css` npm script. This is a problem however when embedding components (see https://github.com/perses/perses/issues/894) - ideally the CSS gets compiled and embedded in the JS files in the `dist` folder of the plugin. This means we need a webpack step for each (or only this?) plugin :|.

**Styling**
Jaeger UI uses the Ant Design Library, Perses uses Material UI. Probably we should replace all components with their Material UI counterparts.
Additionally, it should be possible to customize the styling when this component gets embedded into another application (e.g. OpenShift Console).

**State Management**
Jaeger UI uses Redux (without Redux Toolkit), Perses uses zustand. Should we migrate everything?

**Tests**
Some tests use the `enzyme` library, which is deprecated and only supported until React 17. It still works with React 18 though, if you run `npm install --legacy-peer-deps`, which is obviously a hack. Moving to React Testing Library is not straightforwarded, because the philosophies between these testing libraries are quite different (`enzyme` exposes all internal details (e.g. props) vs. RTL is more a blackbox testing with `queryByRole` etc.).
Currently 6/597 tests fail, mainly because of some issue with `jest.spyOn()` on the default export. Probably can be worked around by removing the default export and using a named export instead.

**Dependencies**
I already tried to remove all unreferences dependencies, possibly we can bring down the number of dependencies even more.

**MVP**
Can we discuss a set of minimum requirements to get an initial version of this PR merged?

# Screenshots

grafik

# Checklist

## UI Changes

  include:


celian-garcia (commented):
Hi @andreasgerstmayr first of all I would like to thank you for working on this. That's amazing to see a Gantt chart in Perses!

Let me divide my answer in two sections, what does not regard the Jaeger UI copy paste, and what does.

# General questions

### UI/UX
Regarding height of query section, we definitely agree on the fact that query takes far too much space. This is a problem that will be addressed at the same time for traces explorer, timeseries explorer, and even panel spec editor. Fortunately the component is the same one

There is also the problem of displaying traces with table or with gantt chart. In grafana this is a two step flow. First you search for a trace with some filters and it displays a table with a list of trace, then you click on the trace and it open a side explore with a query with only the trace id, the result is the Trace directly in gantt chart format.
This is interesting as we'll have to do something similar, probably. 
I am wondering if you have already any display idea. I see in your screen that the trace seems to be collapsable. Does it mean that this is designed to display a list of trace in accordion with gantt chart inside each? If yes not bad idea.

Thirdly, we have a feature of panel fullscreen developed by Guillaume recently, we can maybe do something similar to have a full screen toggle in explorer.

### CSS
I'm not expert in CSS module but you're definitely true that we'll want to avoid having CSS conflicts. This is a general topic for all the plugins that we'll implement. So if you find a solution I guess this is a rule that we'll want to enforce to the plugin developers. Except if the plugin solution invalidate the need of separating the classes? ping @AntoineThebaud @jgbernalp 

### Styling
I'm not sure that migrating to Material UI would be a strong requirement. I think the most important would be to use the same theme, so maybe a translation from the material theme to the whatever theme the library is using would be easier for you and already customizable enough to change colors/typography? WDYT? We're doing already such generation to translate our theme to echart https://github.com/perses/perses/blob/e9dddceb0fdf640e9b0ebdb278b0dde8b5f49216/ui/components/src/utils/theme-gen.ts#L23

# Copy Jaeger UI ? Really ?

Now before trying to answer the rest of the Open Questions, I am wondering if it's really a good solution to copy paste the code like that. There is a ton of code that we'll have to maintain :astonished:. I'm a bit afraid of such a work to be honest.

But let's assume we go this way, just to answer the questions.

### State Management
Jaeger UI seems to use redux far more heavily than we are using zustand. In Perses, we use zustand store for some specific use cases but generally, simple properties and callback do the job. In the Jaeger code you copied, it looks like redux is the heart of any single property change. 
I feel like there's no other way than cope with redux if we use this copy paste solution.

### Tests
A bit the same answer than state management

### Dependencies
Nothing to say. For sure remove useless deps. For the necessary ones, that's alright. As long as they'll be encapsulated in the plugin.

# MVP 
This is indeed very important. I'll conclude by this.
First we need to decide if the copy paste of Jaeger solution is the right one. I'd like to see that you searched for potential Gantt libraries and that you considered maybe some other opensource projects ? Ok maybe Grafana we don't have rights but do we have rights to fork from before the new licence ? :smiling_imp:

andreasgerstmayr:
I agree, it's a lot of new code, and some dependencies and code patterns (Redux, enzyme, etc.) don't align well with Perses. Probably we don't need the entire feature set of the Jaeger UI Gantt chart.

I'm closing this in favor of a new, basic Gantt chart implementation in #2047.

PR #2045, opened 2024-06-07
[BUGFIX] DaC Go SDK: Fix unwanted intermediary field append in YAML output

AntoineThebaud:
# Description

YAML marshalling was causing such unexpected `pluginspec` field to appear in the final output:
```yaml
queries:
  - kind: TimeSeriesQuery
    spec:
      plugin:
          kind: PrometheusTimeSeriesQuery
          spec:
              pluginspec: # invalid !
                  query: sum by (app_name) (jvm_info_total{stack="$stack"})
```

To solve this I added an explicit `inline` marshall instruction for this field + same for any similar cases in other builders, just in case.

# Checklist

### DIFF
--- a/docs/dac/go/README.md
+++ b/docs/dac/go/README.md
@@ -1,4 +1,4 @@
-# CUE SDK for Dashboard-as-Code
+# Go SDK for Dashboard-as-Code
 
 > [!NOTE]
 > To get started with Dashboard-as-Code in Perses, have a look at the [DaC user guide](../../user-guides/dashboard-as-code.md) first.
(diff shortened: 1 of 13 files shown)


AntoineThebaud:
> I'm always cautious with ,inline and json and marshal/unmarshal does it work for json ?

Well we have at least the DaC unit tests that validate the JSON output, and as you can see they are passed succesfully. Actually if you look at the rest of the existing code there are already DaC libraries for which the Builder struct already has `json:",inline" yaml:",inline"` and it's not causing issues (e.g [here](https://github.com/perses/perses/blob/main/go-sdk/prometheus/variable/promql/promql.go#L60), or [here](https://github.com/perses/perses/blob/main/go-sdk/variable/text-variable/text-variable.go#L25)), so this PR just makes it consistent everywhere.

celian-garcia (approved):
Lgtm

PR #2046, opened 2024-06-07
[FEATURE] TimeSeries chart: add query-level settings for color override

### LINKED ISSUE #557: Allow set of a color when you have only one timeseries for query

AntoineThebaud:
# Description

- Closes #557.
- Also aims to fulfill use cases mentioned in #1524 (for TimeSeries chart only).

This PR is basically a second version of PR https://github.com/perses/perses/pull/1678, revamped to match what was discussed and summed up [here](https://github.com/perses/perses/pull/1678#discussion_r1442975526).
It adds a new field to the spec of the TimeSeriesChart plugin, to allow defining style overrides for the provided queries.

Here's how it looks like in the panel editor:

image

For now only color settings can be provided, but this could be easily extended if need be. _For this reason the color attributes are mandatory for now because they're the only settings available. These could be made optional once we have more settings implemented._

For color override, 2 different modes are available: `fixed` and `fixed-single`. Their behaviors are explained in the user doc. See below screenshot for a showcase:

2024-06-07_14h12_26

A new builder was added to the DaC Go SDK to support the new query settings.

### Additional changes:
- Rename the ThresholdColorPicker component to OptionsColorPicker & reuse it for query color overrides
- from the original work of @sjcobb:
  - Change default color palette to 'auto' (see comment [here](https://github.com/perses/perses/issues/1524#issuecomment-1868190329)). _The happo diff is only about this_.
  - VisualOptionsEditor refactored to use immer
  - Refactor useMemo logic in TimeSeriesChartPanel

# Checklist

## UI Changes

  include:


docs/plugin/panel.md
```diff
@@ -118,6 +120,19 @@ values:
 mode: <enum = "auto" | "categorical">
 ```
 
+### Query Settings specification
+
+```yaml
+# queryIndex is an unsigned integer that should match an existing index in the panel's `queries` array
+queryIndex: <number>
```
AntoineThebaud, line 127: FYI @Nexucis in the end we dropped using a JSON reference here, a basic number to save only the index part was enough on the technical side. Are you fine with this?

ui/panels-plugin/src/plugins/time-series-chart/utils/palette-gen.ts
```diff
@@ -12,46 +12,53 @@
...
-    return getAutoPaletteColor(seriesName, fallbackColor);
-  }
-
   // Explicit way to always cycle through classical palette instead of changing when based on number of series.
   if (visual.palette?.mode === 'categorical') {
     return getCategoricalPaletteColor(categoricalPalette, seriesIndex, fallbackColor);
   }
 
-  // Decide which palette to use based on total series returned.
-  // When series number exceeds number of colors in palette, use generative colors instead.
```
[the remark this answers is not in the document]
AntoineThebaud: No because as you can see the default color palette is now 'auto' in all cases, and this mode "_Get color from generative color palette, this approaches uses series name as the seed and allows for consistent colors across panels (when all panels use this approach)_"

celian-garcia (approved):
LGTM functionally :) thanks !

docs/plugin/panel.md
Nexucis: Yes I am fine with that

Nexucis (approved):

Nexucis:
Amazing work @AntoineThebaud thank you so much for doing it !!!

PR #2047, opened 2024-06-07
[FEATURE] New TracingGanttChart panel

### LINKED ISSUE #978: Distributed tracing dashboard / Gantt chart
Hello, first thanks for the awesome initiative. 

I would like to explore the possibility if the perses project will provide any distributed tracing visualization capabilities - e.g. trace search and Gantt chart/timeline view. 

Screenshot of Jaeger UI

andreasgerstmayr:
# Description
This PR adds a new `TracingGanttChart` panel. This is built from scratch, based on React and Material UI.

Resolves #978
Obsoletes #2033

**Features**
* colors are generated deterministically (same as the `TimeSeriesChart`)
* spans with errors have a red hue and an error icon next to the name
* span collapsing
* span indention guide highlighting
* resizable divider (between span name and span bars)
* dynamic indention width based on the spans in viewport (making it easier to navigate deeply nested spans)
* ticks at 0%, 25%, 50%, 75% and 100%
* mini Gantt chart overview to select a time frame of the trace to zoom into
* sidepanel: span attributes and events are visible in a side panel (after clicking on a span)
* performance: tested with traces up to 800 spans; `react-virtuoso` only renders rows which are visible
* unit tests

**Alternative Gantt chart libraries considered**
* https://scheduler.bitnoise.pl/
* https://frappe.io/gantt
* https://github.com/MaTeMaTuK/gantt-task-react
* https://www.npmjs.com/package/vis-timeline
* https://anovokmet.github.io/svelte-gantt
* https://thetechcompany.github.io/react-timeline-gantt/
* https://github.com/robicch/jQueryGantt

Some of them look nice, but all of them are basically intended for project planning purposes (using days and weeks, drag 'n' drop support, etc.). Tweaking one of these to match our use case is likely more effort than starting from scratch.

# Screenshots

grafik

https://github.com/perses/perses/assets/538011/f7e2692c-1ee4-4fc0-b2b4-d612ba702a2e

grafik

# Checklist

## UI Changes

  include:


ui/panels-plugin/src/plugins/trace-table/DataTable.tsx
```diff
@@ -70,21 +73,34 @@ export function DataTable({ result }: DataTableProps) {
...
   let totalSpanCount = 0;
   let totalErrorCount = 0;
   for (const stats of Object.values(trace.serviceStats)) {
     totalSpanCount += stats.spanCount;
     totalErrorCount += stats.errorCount ?? 0;
   }
 
+  const traceLinkParams = new URLSearchParams({
+    explorer: '1',
+    queries: `[{"kind":"TraceQuery","spec":{"plugin":{"kind":"TempoTraceQuery","spec":{"query":"${trace.traceId}"}}}}]`,
```
andreasgerstmayr, line 86: This seems hacky, is there a better way to do it?
I could use `onClick`/`setQueries` and handle it client side, but then I'll loose the browser features of real links, like open a link in a new (background) tab, etc.

ui/explore/src/components/ExploreManager/TracesExplorer.tsx
```diff
@@ -32,55 +90,21 @@ function TracePanel({ queries }: { queries: QueryDefinition[] }) {
...
-      </DataQueriesProvider>
-    </Box>
-  );
-}
-
-export function TracesExplorer() {
-  const { tab, queries, setTab, setQueries } = useExplorerManagerContext();
+  // TODO: how to access query string here?
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const isSingleTrace = definitions.length === 1 && isValidTraceId((definitions[0]?.spec as any).query || '');
```
andreasgerstmayr, line 95: How can I access the `query` here? It's required to determine whether to show the trace table or Gantt chart.

ui/panels-plugin/src/plugins/trace-table/DataTable.tsx
Gladorme: Hmm, I don't think of a better solution :/

However, I think we should remove this magic number (1) and put it behind a variable 😄

ui/explore/src/components/ExploreManager/TracesExplorer.tsx
Gladorme: How to access `query` without the `as any` ?

andreasgerstmayr: I think it's not possible, because the required type `TempoTraceQuerySpec` is in `tempo-plugin`, which is not (and should not be) a dependency of `@perses-dev/explore`. I've extended the condition to check the `kind`s, but I guess that's all I can do here :)

... unless we would move the `query` one level up to the `TraceQuery` instead of `TempoTraceQuery`. But I guess it's ok as-is now.

AntoineThebaud (changes requested):
Could you please add the associated cuelang schema on the backend side, just like you did for TraceTable in https://github.com/perses/perses/pull/2061? Thanks in advance :)

ui/panels-plugin/src/plugins/trace-table/trace-table-model.ts
```diff
@@ -11,16 +11,29 @@
...
+import { MouseEvent } from 'react';
+
 /**
  * The Options object type supported by the TraceTable panel plugin.
  */
 // Note: The interface attributes must match cue/schemas/panels/trace-table/trace-table.cue
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface TraceTableOptions {}
+export interface TraceTableOptions {
+  onTraceClick?: (e: MouseEvent, traceId: string) => void;
```
andreasgerstmayr, line 22: Is this fine? I think in general the `spec` of a Panel should be JSON-serializable, but in this case I need functions for
* running `setQueries()` on the explore page
* creating a link to the trace detail page (useful when the panel gets embedded outside of perses)

ui/panels-plugin/src/plugins/trace-table/DataTable.tsx
andreasgerstmayr: In which package should I define this constant? It belongs to the `explore` package, however we don't have a dependency from `panels-plugin` on the `explore` package (I'm not sure if I should add one? Doesn't seem right).

Maybe we can change `explorer=1` to `explorer=traces`, to avoid having a magic number in the code, and not need a dependency from `panels-plugin` to the `explore` package?

Gladorme:
Hey, sorry for the delay. Thanks for the local env setup 👍 
However, I tested / played with our PRD Tempo. It's really great! 😄 

I spotted few issues:
- Dark theme have some CSS color issue
  - On the trace time line: 
image
  - Spans char on span hover:
image
  - Traces table on trace hover: 
image
- Some data are truncated on the right and bottom of spans table:
image
- When users do not chose the default datasource and click on a trace on trace table. It will redirect to the span panel with the default datasource

And some futur improvements:
- Trace table loading: once the "search" request is done maybe we could show the table, instead of waiting all traces requests. For example, the search request was finished at 1.5s, but the all traces requests took 9s to finish
image
- Do not show "Events" tab if there is no event and/or show a chip with the number of events in the span detail view
image
- Allow users to resize span detail view

ui/explore/src/components/ExploreManager/TracesExplorer.tsx
```diff
@@ -32,55 +105,28 @@ function TracePanel({ queries }: { queries: QueryDefinition[] }) {
...
-          />
-        </Stack>
-      </DataQueriesProvider>
-    </Box>
-  );
-}
-
-export function TracesExplorer() {
-  const { tab, queries, setTab, setQueries } = useExplorerManagerContext();
+  // Cannot cast to TempoTraceQuerySpec because 'tempo-plugin' types are not accessible in @perses-dev/explore
```
Gladorme, line 108: It will be easier when the explorer components will be located in plugin too.

ui/panels-plugin/src/plugins/trace-table/trace-table-model.ts
```diff
@@ -11,16 +11,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { MouseEvent } from 'react';
+
 /**
  * The Options object type supported by the TraceTable panel plugin.
  */
 // Note: The interface attributes must match cue/schemas/panels/trace-table/trace-table.cue
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
```
Gladorme, line 20: `// eslint-disable-next-line` can be remove if we keep it like that (depends on what is done for next comment 😄 )

ui/panels-plugin/src/plugins/trace-table/trace-table-model.ts
Gladorme: Maybe it's better to do it in `TraceTableProps` and keep `TraceTableOptions` interface to model stuff?
```ts
export interface TraceTableProps extends PanelProps<TraceTableOptions> {
  onTraceClick?: (e: MouseEvent, traceId: string) => void;
  traceLink?: (traceId: string) => string;
}
```

andreasgerstmayr: I like this idea :+1:

If these properties should be set, this panel (React component) must be embedded from code anyway, and it's not possible to set this property from a JSON dashboard specification.

andreasgerstmayr:
Thank you for the review!

> Dark theme have some CSS color issue

Ah, good catch. I forgot to test the dark theme.
Fixed.

>  Some data are truncated on the right and bottom of spans table:

Should be fixed now.
 
>  When users do not chose the default datasource and click on a trace on trace table. It will redirect to the span panel with the default datasource

Fixed.

> And some futur improvements:
> 
>  Trace table loading: once the "search" request is done maybe we could show the table, instead of waiting all traces requests. For example, the search request was finished at 1.5s, but the all traces requests took 9s to finish

This will be resolved once Tempo moves to vParquet 4. It's already the default block format upstream, but there's no release yet including this change.

>  Not show "Events" tab if there is no event and/or show a chip with the number of events in the span detail view

Done.

>  Allow users to resize span detail view

Done.

andreasgerstmayr:
There's one minor issue remaining: when you click on a link in the trace table, it does a full page refresh.
If I remove `reloadDocument` from the `<Link>`, it updates the URL but doesn't update the content due to how the state is propagated from `<ExplorerManagerProviderWithQueryParams>` to `<ExplorerManagerProvider>`. I'll create a separate PR for it.

Gladorme (approved):
LGTM

Gladorme:
Minor issue: if you go to event tab with a span that have an event and click on a span without event. There is nothing in the view and no tab selected. Maybe we should still display event tab if it's the current tab even if there is no event OR switch tab to attributes if there is no event 🤔

andreasgerstmayr:
> Minor issue: if you go to event tab with a span that have an event and click on a span without event. There is nothing in the view and no tab selected. Maybe we should still display event tab if it's the current tab even if there is no event OR switch tab to attributes if there is no event 🤔

Good catch! Fixed, I'm switching the tab state to attributes now if there's no event.

PR #2060, opened 2024-06-12
[FEATURE] Provide a way by config to activate the ephemeral dashboard feature

Nexucis:
Not everybody needs the ephemeral dashboard. So having a config that explicitly activate this feature let people choose what feature they would like to use.

### DIFF
--- a/pkg/model/api/config/config.go
+++ b/pkg/model/api/config/config.go
@@ -18,11 +18,24 @@ import (
 
 	"github.com/perses/common/config"
 	"github.com/prometheus/common/model"
+	"github.com/sirupsen/logrus"
 )
 
 const (
 	defaultEphemeralDashboardsCleanupInterval = 24 * time.Hour
 )
+
+type EphemeralDashboard struct {
+	Activate        bool           `json:"activate" yaml:"activate"`
+	CleanupInterval model.Duration `json:"cleanup_interval" yaml:"cleanup_interval"`
+}
+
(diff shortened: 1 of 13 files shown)


pkg/model/api/config/config.go
```diff
@@ -40,14 +52,22 @@ type Config struct {
...
+	EphemeralDashboard EphemeralDashboard `json:"ephemeral_dashboard,omitempty" yaml:"ephemeral_dashboard,omitempty"`
 	// Frontend contains any config that will be used by the frontend itself.
 	Frontend Frontend `json:"frontend,omitempty" yaml:"frontend,omitempty"`
 }
 
 func (c *Config) Verify() error {
-	if c.EphemeralDashboardsCleanupInterval <= 0 {
-		c.EphemeralDashboardsCleanupInterval = model.Duration(defaultEphemeralDashboardsCleanupInterval)
+	if c.EphemeralDashboardsCleanupInterval > 0 {
+		// This is to avoid an immediate breaking change. This code will be removed for the version v0.49.0
```
celian-garcia, line 66: if possible, a warn would be cool

celian-garcia (approved):

pkg/model/api/config/config.go
Nexucis: done

AntoineThebaud (commented):
too late :(

PR #2061, opened 2024-06-12
[FEATURE] TraceTable panel

andreasgerstmayr:
# Description

Adds a new table showing all traces matching a TraceQL query, similar to the TimeSeriesTable Panel but for traces.

# Screenshots

grafik

grafik

# Checklist

## UI Changes

  include:


celian-garcia:
Woaw you did something like a legend but completely boosted. I like what I see! 

Didn't check the code yet but looks really cool thanks! I will try to have a look this week.

ui/panels-plugin/src/plugins/trace-table/TraceTablePanel.tsx
```diff
@@ -0,0 +1,59 @@
...
+          height={contentDimensions.height / 2}
+          aria-label="Loading..."
+        />
+      </Box>
+    );
+  }
+
+  const tracesFound = queryResults.some((traceData) => (traceData.data?.traces ?? []).length > 0);
+  if (!tracesFound) {
+    return <ErrorAlert error={{ name: 'No traces found.', message: 'No traces found.' }} />;
```
celian-garcia, line 51: I believe no traces found should not be an error alert but ok for the DataTable. A "No traces found" message centered in regular typography would be preferred. 
Like we have in the dashboards table 
image

ui/panels-plugin/src/plugins/trace-table/DataTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export function DataTable({ result }: DataTableProps) {
+  const theme = useChartsTheme();
+
+  if (!result) {
+    return null;
+  }
+
+  const traces = result.flatMap((d) => d.data).flatMap((d) => d?.traces || []);
+  const rows = traces.map(buildRow.bind(null, theme));
```
celian-garcia, line 46: I´d find more readable ``traces.map((trace) => buildRow(theme, trace))``

celian-garcia (commented):
Functionally approved. Some technical nit but overall looks good. 

Maybe some additional approval would be needed to see if this is the kind of table we want for traces explorer but to me this is better. 

Also if you can reduce a bit the height taken by each table line and maybe increase the overall height of the table that would be perfect

Nexucis:
Oh whaou, this is so cool ! Thank you @andreasgerstmayr 🙏

andreasgerstmayr:
Thanks for the review! I've applied all requested changes.

> Also if you can reduce a bit the height taken by each table line and maybe increase the overall height of the table that would be perfect

Regarding the table height: The table uses 100% of the available space, should I increase `PANEL_PREVIEW_HEIGHT` of the explore page? Or reduce the height of the scatterplot?

ui/panels-plugin/src/plugins/trace-table/TraceTablePanel.tsx
celian-garcia: IMO that should be the same message as when no timeseries found for graphs, for consistency. Ideally this would even be something a plugin developer should not have to deal with. 


In the specific case of the traces explorer now.
In case there's no data, I would see a message no data in the scatter plot and no table panel at all WDYT ? 
Same for errors and loading.
Since the beginning I feel like what you're doing is kind of a legend and that's what a legend would do

sjcobb:
Long term, should the Table panel be updated to support multiple query types?

Or could go the route I suggested here: https://github.com/perses/perses/pull/1672#discussion_r1435464852
Where ScatterChart becomes something like TraceViewer and then have a `visual.display: 'table' | 'scatter'` similar to how [TimeSeriesChart](https://github.com/perses/perses/blob/main/cue/schemas/panels/time-series/time-series.cue#L38) supports `visual.display: 'bar'`

celian-garcia:
> Long term, should the Table panel be updated to support multiple query types?
> 
> Or could go the route I suggested here: [#1672 (comment)](https://github.com/perses/perses/pull/1672#discussion_r1435464852) Where ScatterChart becomes something like TraceViewer and then have a `visual.display: 'table' | 'scatter'` similar to how [TimeSeriesChart](https://github.com/perses/perses/blob/main/cue/schemas/panels/time-series/time-series.cue#L38) supports `visual.display: 'bar'`

I personally think that a generic table panel should be created independently from this one. 

It's good to have a dedicated "table" panel for traces with all the chips and fancy things, and on which you can take design decisions that you don't have to backport for timeseries / logs. Easier for development lifecycle.

Although the generic table would be very interesting as well to display information based on ts/traces ... Where we don't necessarily need to know it is ts/traces behind. (E.g all status pages, a list of k8s resources, ...) 
On the contrary the dedicated tables TracesTable TimeSeriesTable would display Traces and Timeseries as they are, with all the subtleties of their respective data models. (E.g traces list of jaeger, timeseries table of Prometheus)

celian-garcia:
> Or could go the route I suggested here: [#1672 (comment)](https://github.com/perses/perses/pull/1672#discussion_r1435464852) Where ScatterChart becomes something like TraceViewer and then have a `visual.display: 'table' | 'scatter'` similar to how [TimeSeriesChart](https://github.com/perses/perses/blob/main/cue/schemas/panels/time-series/time-series.cue#L38) supports `visual.display: 'bar'`

Why not,
Functionally  there is the benefit of having less panels in the drop-down with the price of knowing that you have to go in the settings to change display

Technically this is ofc a benefit because less code to maintain. 
I'm wondering if we couldn't have a system having the two panels reusing each other. Like the traces table would exist as a panel but actually instantiating a trace chart with display: table

celian-garcia:
> Thanks for the review! I've applied all requested changes.
> 
> > Also if you can reduce a bit the height taken by each table line and maybe increase the overall height of the table that would be perfect
> 
> Regarding the table height: The table uses 100% of the available space, should I increase `PANEL_PREVIEW_HEIGHT` of the explore page? Or reduce the height of the scatterplot?

I'm not sure let's not block the PR for that. 

LGTM I approve. Although I like the idea of Steven to actually have a TraceChart with ``visual.display: 'table' | 'scatter'``. I like also to have a TraceTable that you can just select in the dropdown. Both would be great like the ``BarChart`` yeah but the bar chart was simple, this was nothing more than a flag in echarts so no duplication. Here to avoid duplication it would be more complex. It would need either a dependency between the two panels or a third party module.

celian-garcia (approved):

andreasgerstmayr:
My thoughts on supporting two visualizations in the same panel:

For ScatterPlot and TraceTable:
I think the ScatterPlot is generic enough to get support for logs or time series in the future, so I would not merge ScatterPlot and TraceTable into a single panel.

For TimeSeriesTable and TraceTable:
Sounds fine to me, we could even detect the table variant based on the query type (and show an error if queries are of mixed type).
However, if we add timeseries or tracing specific panel options in the future, it'll get awkward imho.
Regarding code duplication I don't see much benefit, both tables are quite simple, they print the header (type specific) and then the rows (again type specific), so I don‘t see much code reduction except for the panel boilerplate code.

ui/panels-plugin/src/plugins/trace-table/TraceTablePanel.tsx
andreasgerstmayr: > IMO that should be the same message as when no timeseries found for graphs, for consistency. Ideally this would even be something a plugin developer should not have to deal with.

I like the idea, I've created https://github.com/perses/perses/issues/2077 to track this.

> In the specific case of the traces explorer now. In case there's no data, I would see a message no data in the scatter plot and no table panel at all WDYT ? Same for errors and loading. 

If #2077 gets implemented, maybe the panel in the explorer page could be a "meta-panel", containing the ScatterPlot and the TraceTable, and handling the loading/error/no data state at the parent Panel.

celian-garcia:
> My thoughts on supporting two visualizations in the same panel:
> 
> For ScatterPlot and TraceTable: I think the ScatterPlot is generic enough to get support for logs or time series in the future, so I would not merge ScatterPlot and TraceTable into a single panel.
> 
> For TimeSeriesTable and TraceTable: Sounds fine to me, we could even detect the table variant based on the query type (and show an error if queries are of mixed type). However, if we add timeseries or tracing specific panel options in the future, it'll get awkward imho. Regarding code duplication I don't see much benefit, both tables are quite simple, they print the header (type specific) and then the rows (again type specific), so I don‘t see much code reduction except for the panel boilerplate code.



> My thoughts on supporting two visualizations in the same panel:
> 
> For ScatterPlot and TraceTable: I think the ScatterPlot is generic enough to get support for logs or time series in the future, so I would not merge ScatterPlot and TraceTable into a single panel.
> 
> For TimeSeriesTable and TraceTable: Sounds fine to me, we could even detect the table variant based on the query type (and show an error if queries are of mixed type). However, if we add timeseries or tracing specific panel options in the future, it'll get awkward imho. Regarding code duplication I don't see much benefit, both tables are quite simple, they print the header (type specific) and then the rows (again type specific), so I don‘t see much code reduction except for the panel boilerplate code.

Yeah I think all the debate is should we make component vertically generic (support several display for same query type) or horizontally (support several query type for the same display). We started with first approach but since I enabled the multi query types support this is changing the game haha 😅 

For the code duplication I was meaning between an eventual TraceChart with ``display: table`` and TraceTable. And potentially then if scatter plot is generic enough between TraceChart with ``display: scatter plot`` and ScatterPlot.
TraceTable and TimeseriesTable are too different and indeed makes no sense to merge.

We're going too far anyway, I'd be personally in favour of keeping everything as modular as possible. 
Let's keep ScatterPlot, TimeseriesTable, TracesTable, and a future GenericTable. They're all design to display different things.

PR #2073, opened 2024-06-14
[ENHANCEMENT] Schemas load: avoid breaking too early in case of error + add monitoring for it

AntoineThebaud:
# Description

Currently if we have e.g 10 panels plugin installed and the 4th one in alphabetical order is faulty/corrupted, the 6 following plugins wont be loaded even though they'd be valid, which is too bad..
This PR aims to change that + take advantage of changing this part of the code to add instrumentation of schemas load.

New metrics overview:
```
# HELP perses_plugin_schemas_load_attempts Amount of attempts to load plugin schemas
# TYPE perses_plugin_schemas_load_attempts gauge
perses_plugin_schemas_load_attempts{load="change",schema="model",status="error"} 0
perses_plugin_schemas_load_attempts{load="change",schema="model",status="success"} 4
perses_plugin_schemas_load_attempts{load="full",schema="migration",status="error"} 0
perses_plugin_schemas_load_attempts{load="full",schema="migration",status="success"} 9
perses_plugin_schemas_load_attempts{load="full",schema="model",status="error"} 0
perses_plugin_schemas_load_attempts{load="full",schema="model",status="success"} 15
```
Just in case, because I assume it may not be fully self-explanatory:
- `load="change"` tracks the case of dynamic changes detected by file watch
- `load="full"` tracks the case of full schemas load (= at start + cron task)
- `schema="migration"` tracks the migration schemas only (= `migrate.cue` files). _Btw the error timeserie is always 0 here because we have no relevant error case to detect for these in the current state._
- `schema="model"` tracks the "real" plugin schema to be used for backend validation.

# Checklist

### DIFF
--- a/internal/api/migrate/loader.go
+++ b/internal/api/migrate/loader.go
@@ -65,13 +65,15 @@ func (c *migCuePart) GetSchemaPath() string {
 	return c.schemasPath
 }
 
-func (c *migCuePart) Load() error {
-	conditions, err := c.buildListOfConditions()
+// NB There are no relevant case of "failed loads" in the case of migration schemas at the moment
+func (c *migCuePart) Load() (successfulLoadsCount, _ int, err error) {
+	var conditions string
+	conditions, successfulLoadsCount, err = c.buildListOfConditions()
 	if err != nil {
-		return err
+		return
 	}
 	c.listOfConditions.Store(&conditions)
-	return nil
+	return
 }
 
 func (c *migCuePart) getConditions() string {
@@ -82,15 +84,16 @@ func (c *migCuePart) getPlaceholder() string {
 	return c.placeholderText
 }
 
-func (c *migCuePart) buildListOfConditions() (string, error) {
+func (c *migCuePart) buildListOfConditions() (string, int, error) {
+	successfulLoadsCount := 0
+
 	files, err := os.ReadDir(c.schemasPath)
 	if err != nil {
-		return "", err
+		return "", successfulLoadsCount, err
 	}
 
+	// gather the content of all migration files found
 	var listOfConditions strings.Builder
-
-	// process each schema plugin to convert it into a CUE Value
 	for _, file := range files {
 		if !file.IsDir() {
 			logrus.Tracef("file %s is ignored since we are looking for directories", file.Name())
@@ -105,6 +108,7 @@ func (c *migCuePart) buildListOfConditions() (string, error) {
 
 		listOfConditions.WriteString(string(contentStr))
 		listOfConditions.WriteString("\n")
+		successfulLoadsCount++
 	}
 
 	// append a default conditional for any Grafana plugin that has no corresponding Perses plugin
@@ -114,5 +118,5 @@ func (c *migCuePart) buildListOfConditions() (string, error) {
 	}`, c.defaultValue))
 	listOfConditions.WriteString("\n")
 
-	return listOfConditions.String(), nil
+	return listOfConditions.String(), successfulLoadsCount, nil
 }
--- a/internal/api/migrate/loader.go
+++ b/internal/api/migrate/loader.go
@@ -65,15 +65,16 @@ func (c *migCuePart) GetSchemaPath() string {
 	return c.schemasPath
 }
 
-// NB There are no relevant case of "failed loads" in the case of migration schemas at the moment
-func (c *migCuePart) Load() (successfulLoadsCount, _ int, err error) {
+// Load will load the migration schemas from the filesystem
+func (c *migCuePart) Load() (int, int, error) {
 	var conditions string
-	conditions, successfulLoadsCount, err = c.buildListOfConditions()
-	if err != nil {
-		return
+	conditions, successfulLoadsCount, err := c.buildListOfConditions()
+	if err == nil {
+		c.listOfConditions.Store(&conditions)
 	}
-	c.listOfConditions.Store(&conditions)
-	return
+	// NB: There are no relevant case of "failed loads" to monitor in the case of migration schemas
+	// at the moment, so we always return 0 for the failed loads count
+	return successfulLoadsCount, 0, err
 }
 
 func (c *migCuePart) getConditions() string {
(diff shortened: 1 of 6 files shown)


celian-garcia:
I believe it's worth it to put the explanation of different load and schema types in the documentation alongside the code.

internal/api/schemas/schemas.go
```diff
@@ -269,10 +269,20 @@ func (s *sch) validatePlugin(plugin common.Plugin, modelKind string, modelName s
 }
 
 func (s *sch) init() error {
-	for _, l := range s.loaders {
-		if err := l.Load(); err != nil {
+	return RunLoaders(s.loaders, "model", "full")
+}
+
+func RunLoaders(loaders []Loader, schemaType, loadType string) error {
```
celian-garcia, line 275: ``schemaType`` and ``loadType`` enums would be more future proof.

celian-garcia: That would be the opportunity to document it on the enum type :)

internal/api/migrate/loader.go
```diff
@@ -65,13 +65,15 @@ func (c *migCuePart) GetSchemaPath() string {
 	return c.schemasPath
 }
 
-func (c *migCuePart) Load() error {
-	conditions, err := c.buildListOfConditions()
+// NB There are no relevant case of "failed loads" in the case of migration schemas at the moment
+func (c *migCuePart) Load() (successfulLoadsCount, _ int, err error) {
```
celian-garcia, line 69: doc should start with Load :D 
this can be the opportunity to tell what this function is doing before adding a NB :P
```
// Load the migration schema ?
// NB: ...
```

internal/api/migrate/loader.go
```diff
@@ -65,13 +65,15 @@ func (c *migCuePart) GetSchemaPath() string {
...
+	var conditions string
+	conditions, successfulLoadsCount, err = c.buildListOfConditions()
 	if err != nil {
-		return err
+		return
 	}
 	c.listOfConditions.Store(&conditions)
-	return nil
+	return
```
celian-garcia, line 76: More readable if you keep not named returns no ? 

```go
	conditions, successfulLoadsCount, err := c.buildListOfConditions()
	if err == nil {
    c.listOfConditions.Store(&conditions)
	}
  return successfulLoadsCount, 0, err
```

celian-garcia: btw if I understand well the second returns is the failure counts? 
Migration can't fail?

celian-garcia (changes requested):
Great to see this new interesting metric! Mainly docs and nits for me to understand

internal/api/migrate/loader.go
AntoineThebaud: > More readable if you keep not named returns no ?
> 
> ```go
> 	conditions, successfulLoadsCount, err := c.buildListOfConditions()
> 	if err == nil {
>           c.listOfConditions.Store(&conditions)
> 	}
>         return successfulLoadsCount, 0, err
> ```

I was hesitating to do this, invert the if condition so that we only have 1 return clause, but it may be confusing to have a `err == nil` check while 99% of the time the usual pattern is `err != nil`. I had the feeling that could be the kind of thing one could miss at a first glance. But maybe I'm worrying too much, WDYT?

AntoineThebaud: > btw if I understand well the second returns is the failure counts? Migration can't fail?

Yes kinda, that's what I put in comment above (and in PR desc too)
> // NB There are no relevant case of "failed loads" in the case of migration schemas at the moment

Basically regarding migration schema, since they are not mandatory we just skip if we don't find one. In the current state of the code there's no relevant error case to monitor. I suppose that we could some validation based on the format we expect (explained in the doc https://github.com/perses/perses/blob/main/docs/plugin/cue.md#variable-1), at least to validate that we indeed have a list of conditions for instance.
I wont tackle this for now but I'll add a comment to mention the future work at least.

celian-garcia: > I was hesitating to do this, invert the if condition so that we only have 1 return clause, but it may be confusing to have a err == nil check while 99% of the time the usual pattern is err != nil. I had the feeling that could be the kind of thing one could miss at a first glance. But maybe I'm worrying too much, WDYT?

I still find more readable my way but I understand it might be subjective. As you wish!

internal/api/schemas/monitoring.go
```diff
@@ -0,0 +1,58 @@
...
+	Help:      "Amount of attempts to load plugin schemas",
+}, labelNames)
+
+func RegisterMetrics(reg prometheus.Registerer) {
+	reg.MustRegister(loadAttempts)
+}
+
+func MonitorLoadAttempts(successCount, errorCount int, schema SchemaType, load LoadType) {
+	loadAttempts.WithLabelValues(successStatus, string(schema), string(schema)).Set(float64(successCount))
+	loadAttempts.WithLabelValues(errorStatus, string(schema), string(schema)).Set(float64(errorCount))
```
celian-garcia, line 57:
```suggestion
	loadAttempts.WithLabelValues(successStatus, string(schema), string(load)).Set(float64(successCount))
	loadAttempts.WithLabelValues(errorStatus, string(schema), string(load)).Set(float64(errorCount))
```

internal/api/schemas/loader.go
```diff
@@ -213,15 +225,15 @@ func (r *Reloader) String() string {
...
 	select {
 	case <-ctx.Done():
 		logrus.Infof("canceled %s", r.String())
 		break
 	default:
-		for _, l := range r.Loaders {
-			if err := l.Load(); err != nil {
-				logrus.WithError(err).Errorf("unable to load a schema")
-			}
+		err := RunLoaders(r.Loaders, "model", "full")
```
celian-garcia, line 234: It works to set directly the string instead of the enum? cool!

celian-garcia (changes requested):
Slight copy paste forget but otherwise it will be fine

internal/api/schemas/schemas.go
```diff
@@ -269,10 +269,20 @@ func (s *sch) validatePlugin(plugin common.Plugin, modelKind string, modelName s
 }
 
 func (s *sch) init() error {
-	for _, l := range s.loaders {
-		if err := l.Load(); err != nil {
+	return RunLoaders(s.loaders, "model", "full")
```
Nexucis, line 272: it doesn't seem there is something that prevents you to use the constant you defined in the file `monitoring` :)

internal/api/schemas/loader.go
AntoineThebaud: yes but we want to pass the enum type no? I just forgot actually 😅

celian-garcia (approved):
perfect !

PR #2080, opened 2024-06-18
[BREAKINGCHANGE] Reword "Template Variable" as "Variable" everywhere

AntoineThebaud:
# Description

Get rid of the "Template Variable" wording in the frontend code, that most probably came from Grafana vocabulary while we simply called this object "Variable" in Perses.

As a consequence, the current `<TemplateVariableProvider>` is renamed `<VariableProvider>`, hence the breaking change for the embed use case.

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/plugin-system/src/runtime/variables.ts
+++ b/ui/plugin-system/src/runtime/variables.ts
@@ -1,4 +1,4 @@
-// Copyright 2023 The Perses Authors
+// Copyright 2024 The Perses Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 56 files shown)


ui/plugin-system/src/runtime/variables.ts
```diff
@@ -124,22 +124,22 @@ export class VariableStoreStateMap {
...
   if (ctx === undefined) {
-    throw new Error('No TemplateVariableContext found. Did you forget a Provider?');
+    throw new Error('No VariableContext found. Did you forget a Provider?');
   }
   return ctx;
 }
 
-export function useTemplateVariableValues(names?: string[]): VariableStateMap {
-  const { state } = useTemplateVariableContext();
+export function useVariableValues(names?: string[]): VariableStateMap {
```
celian-garcia, line 141: useTemplateVariableValues was used only in that file so maybe this is better that you rename this one into something else or even merge the two functions what do you think ?

celian-garcia: This is the same for the useTemplateVariableValues inside the TemplateVariableProvider.tsx

image

AntoineThebaud: 2nd PR coming for this ;) I really wanted to focus only on the "Template" word removal since there are a lot of files modified

celian-garcia (approved):
I can only like that as I tried to do the same actually haha

PR #2099, opened 2024-07-01
[ENHANCEMENT] Avoid refetching data when re-opening the search bar

Nexucis:
Like suggested in #2090 by @celian-garcia, I'm adding the usage of `refetchOnMount` to avoid refetching data when re-opening the search bar.

Note: should be merged after #2097


ui/app/src/model/admin-client.ts
```diff
@@ -89,10 +89,14 @@ export function useGlobalDatasource(name: string) {
  * Used to get global datasources from the API.
  * Will automatically be refreshed when cache is invalidated
  */
-export function useGlobalDatasourceList() {
-  return useQuery<GlobalDatasourceResource[], Error>([globalDatasourceResource], () => {
-    return getGlobalDatasources();
-  });
+export function useGlobalDatasourceList(refetchOnMount = true) {
```
Gladorme, line 92: I think it would be better to have all hooks aligned with `useProjectList`
```suggestion
type GlobalDatasourceListOptions = Omit<UseQueryOptions<GlobalDatasourceResource[], Error>, 'queryKey' | 'queryFn'>; 
export function useGlobalDatasourceList(options: GlobalDatasourceListOptions) {
```

Gladorme (approved):

PR #2104, opened 2024-07-02
Add a button to see more data and hide scroll bar

Nexucis:
I have added a button to see more result in case the search is returning more than 10 results. I have also hidden the scroll bar, I found sexier ;)

image


Gladorme:
IMO, we should show the scrollbar (on desktop view) if the element can be scrolled. Maybe the default one is too ugly, but it should be displayed (maybe with custom css).

Nexucis:
moving forward on this one as there are other things to do on the search-bar and I think people needs to try it to see how good is it and how much we need to adapt / change it to have something smoother

PR #2113, opened 2024-07-11
[FEATURE] Add Table panel

Gladorme:
# Description

Add a basic table panel based on the existing Table component developed by Julie (didn't know there was already some stuff before starting 😄). Currently only supporting timeseries. The column editor in the screenshot may change

# Screenshots

image

# Checklist

## UI Changes

  include:


cue/schemas/panels/table/table.cue
```diff
@@ -0,0 +1,33 @@
...
+#columns: {
+	name:               strings.MinRunes(1)
+	header?:            string
+	headerDescription?: string
+	cellDescription?:   string
+	align?:             "left" | "center" | "right"
+	enableSorting?:     bool
+	width?:             number | "auto"
+	hide?:              bool
+}
```
AntoineThebaud, line 33: (Putting this unrelated comment here just to ease the discussion, as github doesn't allow to create threads from standalone comments...)

As discussed offline, it's too bad that in the case of this panel (& others actually) we have some fields of the query that are useless but still displayed: 
2024-07-24_10h25_50

- **Legend name**: ideally it should either serve a purpose (e.g as a default name for the "value" column generated for this query) or hide it if not applicable (but may be not so simple to implement + doesnt solve the problem to have a "zombie" attribute in the model)
- **Min step**: This I guess the proper solution is out of scope of this panel solely, we already discussed the fact to split the current Prometheus timeseries query plugin in 2 different query plugins, one for instant queries & one for range queries, this way the instant query one could just not have this attribute in its model.

ui/components/src/ColumnsEditor/ColumnEditorContainer.tsx
```diff
@@ -0,0 +1,77 @@
...
+              </span>
+            ) : (
+              <strong>{column.name}</strong>
+            )}
+          </Typography>
+        </Stack>
+
+        <Stack direction="row" gap={1}>
+          <IconButton size="small" sx={{ marginLeft: 'auto' }} onClick={handleHideColumn}>
+            {column.hide ? <EyeOffIcon /> : <EyeIcon />}
```
AntoineThebaud, line 67: As already discussed personally I'm not fan of this show/hide button being at the top-right
image

To me it goes against an implicit convention here,
- the trashbin icon seems to mean "remove this column settings" and that's actually what it does: it removes the column settings, not the column itself in the generated table.
- following the same logic, the eye icon seems to mean "hide this column settings", which doesn't make sense, and it actually hides the column in the generated table. 

Thus I'd rather have a show/hide toggle within the column settings part (in first or second position, as if you hide a column you don't care about providing descriptions/headers that will never get displayed)

I know the main advantage to have this eye icon there is to have access to this show/hide feature even if the corresponding column settings is collapsed:
image

Still, I value more UX consistency than saving clicks here, so I still prefer my proposed design 🤓 

But I guess we could manage to find a solution satisfying both, e.g move the controls to the row when it's collapsed / to the form when it's expanded? I can't come up with a design idea right now..

AntoineThebaud (commented):
Just me being annoying on UX as usual 🤓

ui/panels-plugin/src/plugins/table/TablePanel.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import { PanelProps, QueryData, useDataQueries } from '@perses-dev/plugin-system';
+import { LoadingOverlay, Table, TableColumnConfig } from '@perses-dev/components';
+import { useMemo, useState } from 'react';
+import { ColumnDefinition, TimeSeries, TimeSeriesData } from '@perses-dev/core';
+import { SortingState } from '@tanstack/react-table';
+import { TableOptions } from './table-model';
+
+/*
+ * Generate column config from column definitions, if a column has multiple definitions, the first one will be used.
```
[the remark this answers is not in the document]
Gladorme: Not easy: panel Settings Editor, do not have access to query result :/

ui/components/src/ColumnsEditor/ColumnEditorContainer.tsx
Gladorme: I will replicate it in the editor

AntoineThebaud: > But I guess we could manage to find a solution satisfying both, e.g move the controls to the row when it's collapsed / to the form when it's expanded? I can't come up with a design idea right now..

Just for context, a draft of what I had in mind (collapsed vs expanded view)

Propal table

Nexucis:
Would you mind to complete the Go SDK with this new panel ?

Nexucis:
and the file migrate.cue, would be nice to add it

ui/panels-plugin/src/plugins/table/Table.ts
```diff
@@ -0,0 +1,34 @@
...
+ * The core TimeSeriesTable panel plugin for Perses.
+ */
+export const Table: PanelPlugin<TableOptions> = {
+  PanelComponent: TablePanel,
+  supportedQueryTypes: ['TimeSeriesQuery'],
+  queryOptions: {
+    mode: 'instant',
+  },
+  panelOptionsEditorComponents: [
+    { label: 'Columns Settings', content: TableColumnsEditor },
```
AntoineThebaud, line 30:
```suggestion
    { label: 'Column Settings', content: TableColumnsEditor },
```

ui/panels-plugin/src/plugins/table/TablePanel.tsx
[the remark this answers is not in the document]
AntoineThebaud: Yeah true.. for #2046 we added a dedicated custom hook to be able to access the information from the panel settings, but there it was just about the amount of queries defined in the spec, while here it's about a dynamic list of results returned by the query(ies), so indeed not so easy..
Let's keep this for a possible follow-up PR.

ui/components/src/ColumnsEditor/ColumnEditorContainer.tsx
AntoineThebaud: I'm fine with your latest change (the eye icon is hidden when the row is expanded) as a quickwin mitigation, let's go with this for now.

PR #2127, opened 2024-07-26
[BUGFIX] explore page: sync query parameters to explorer state

andreasgerstmayr:
# Description

Previously, clicking on a `<Link>` (without `reloadDocument` attribute) to the explore page with queries in the URL did update the query parameters, but the updated state was not reflected in the `ExplorerManagerProvider`. This is because the query parameters were only considered for the initial state of the `ExplorerManagerProvider`, but other changes of the query parameters were not considered until a full page refresh.

Changes:
* switched the `explorer` type from a number to a string to improve the permalink, i.e. remove the magic number
* keep the URL changes in the browser history, to allow easy navigation from e.g the trace table to gantt chart and back via the browser's back button
* refactored the `ExplorerManagerProvider` to use the state from a single place (either from a parent component like `ExplorerManagerProviderWithQueryParams` or a local state)

# Screenshots

no visible UI changes

# Checklist

## UI Changes

  include:


Gladorme:
There were two providers in case some users would embed it and don't want the component to change url query. I guess not a lot of people will embed it, but I think we should to keep this feature 🤔 
About magic numbers, it was to limit hardcoded plugin stuff in `components` package. For now, I guess it's okay. In the future, it will be computed from installed plugins.

andreasgerstmayr:
> There were two providers in case some users would embed it and don't want the component to change url query. I guess not a lot of people will embed it, but I think we should to keep this feature 🤔

Ok, I'll create a v2 of this PR.

> About magic numbers, it was to limit hardcoded plugin stuff in `components` package. For now, I guess it's okay. In the future, it will be computed from installed plugins.

I think it'd be cool to make it as easy as possible to share links and generate links for a specific query or trace from outside of Perses.

For example, a current link looks like this:
http://localhost:3000/explore?explorer=1&queries=%5B%7B%22kind%22%3A%22TraceQuery%22%2C%22spec%22%3A%7B%22plugin%22%3A%7B%22kind%22%3A%22TempoTraceQuery%22%2C%22spec%22%3A%7B%22query%22%3A%227962171c6d5d15add82a08b937435f54%22%7D%7D%7D%7D%5D&start=1h&refresh=0s


What about a link like this:
http://localhost:3000/explore?explorer=traces&queryKind=TempoTraceQuery&query=7962171c6d5d15add82a08b937435f54&start=1h&refresh=0s

(can we infer `TraceQuery` from `TempoTraceQuery`?)

We can also support multiple queries this way:
http://localhost:3000/explore?explorer=traces&queryKind=TempoTraceQuery&query=7962171c6d5d15add82a08b937435f54&queryKind=TempoTraceQuery&query=684566796876457568&start=1h&refresh=0s

What do you think?

Gladorme:
> 
> > About magic numbers, it was to limit hardcoded plugin stuff in `components` package. For now, I guess it's okay. In the future, it will be computed from installed plugins.
> 
> I think it'd be cool to make it as easy as possible to share links and generate links for a specific query or trace from outside of Perses.

> For example, a current link looks like this: http://localhost:3000/explore?explorer=1&queries=%5B%7B%22kind%22%3A%22TraceQuery%22%2C%22spec%22%3A%7B%22plugin%22%3A%7B%22kind%22%3A%22TempoTraceQuery%22%2C%22spec%22%3A%7B%22query%22%3A%227962171c6d5d15add82a08b937435f54%22%7D%7D%7D%7D%5D&start=1h&refresh=0s
> 
> What about a link like this: http://localhost:3000/explore?explorer=traces&queryKind=TempoTraceQuery&query=7962171c6d5d15add82a08b937435f54&start=1h&refresh=0s
> 
> (can we infer `TraceQuery` from `TempoTraceQuery`?)
> 
> We can also support multiple queries this way: http://localhost:3000/explore?explorer=traces&queryKind=TempoTraceQuery&query=7962171c6d5d15add82a08b937435f54&queryKind=TempoTraceQuery&query=684566796876457568&start=1h&refresh=0s
> 
> What do you think?


I agree that current links are not easily readable. It would be cool to have something better, but I don't know if it will be easy todo with plugins. AFAIK, `QueryType` should still be kept in `core` (not something a plugin can add). 
So maybe we can do something around it, however "TempoTraceQuery" is coming from a plugin and we should not know anything about it in `component/explorer` package level. The only thing I see would be naming convention 🤔

andreasgerstmayr:
Agree, the explorer package should not now about the `TempoTraceQuery` type. Effectively I'll need a way to serialize the generic QueryDefinition to query params (key/value pairs) instead of JSON.

I'll give it a try :)

andreasgerstmayr:
I've added `<ExplorerManagerProviderWithQueryParams> ` again, so it can be used with or without queryParams as the state storage.

I also noticed a bug when opening a gantt chart and then switching to metrics tab and back. tanstack/query issued a re-fetch, which gets stuck in `fetching` state, showing the loading indicator forever. This only happened with the gantt chart.
Turns out there's a bug in tanstack/query, resulting in queries which return data with circular dependencies (which is only the case for a full `Trace`, used by the gantt chart) to be stuck in `fetching` forever: https://github.com/TanStack/query/issues/6954 (fix/workaround is to disable the `structuralSharing` feature).

andreasgerstmayr:
I've some preliminary code regarding readable URL query strings, but I'll create a separate PR for that to keep the current PR simpler.

Gladorme (approved):
Cool! LGTM 😄

PR #2132, opened 2024-07-30
[BREAKINGCHANGE] GO-SDK: remove the possibility to create the plugin builder

Nexucis:
In the different plugin supported, it was possible to create the builder for each plugin which is not the way the SDK has been thought.

On each plugin, there is only one method exposed that should be used to create the plugin. Builder is just the internal way we used to create the plugin


Gladorme (approved):
LGTM 😄

PR #2133, opened 2024-07-30
[ENHANCEMENT] GO-SDK: add missing chart definition on tracing subject and for the time-series-table

Nexucis:


Gladorme (approved):
good when linter fixed :)

PR #2165, opened 2024-08-13
Fix markdown plugin

jgbernalp:
Fix Markdown plugin

### DIFF
--- a/dev/config.yaml
+++ b/dev/config.yaml
@@ -1,7 +1,7 @@
 security:
   readonly: false
   encryption_key: "=tW$56zytgB&3jN2E%7-+qrGZE?v6LCc"
-  enable_auth: true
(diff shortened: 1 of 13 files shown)


dev/config.yaml
```diff
@@ -1,7 +1,7 @@
 security:
   readonly: false
   encryption_key: "=tW$56zytgB&3jN2E%7-+qrGZE?v6LCc"
-  enable_auth: true
+  enable_auth: false
```
Nexucis, line 4:
```suggestion
    enable_auth: true
```

jgbernalp: yes this is always changed when running the e2e locally, need to adjust that. I reverted it

Nexucis (approved):

PR #2167, opened 2024-08-13
simplifying the launch of the e2e tests

Nexucis:


Nexucis:
it reduces the time to run the e2e by 2min. Not that bad

jgbernalp (approved):

PR #2174, opened 2024-08-14
[DOC] Add tshoot section with TLS/HTTP mixed datasource

indigo423:
# Description

Introducing a troubleshooting section using the initial article [Using Perses with TLS and Prometheus with plain HTTP](https://github.com/perses/perses/discussions/2171).

Resolve: #2196

# Reviewer Hints

I've created a directory collecting troubleshooting articles. I have added the screenshots in the same directory. There is an "images" directory, I wasn't sure if it's a good idea to get all images in a similar spot. Just thinking out loud, do you want screenshots here? They come with a maintenance cost and they get outdated soon.

# Checklist


indigo423:
@Nexucis here is a proposal for the troubleshooting section. Feel free ripping it apart :)

Nexucis:
Yeah that's ok for me. Thank you for doing it @indigo423.

Regarding the images, I would prefer to use the github storage for that. A tips is to open an issue, then copy and past the image that will give you a link, and then use this link in the mardown file. Like that we don't store any images in git

indigo423:
> Regarding the images, I would prefer to use the github storage for that. A tips is to open an issue, then copy and past the image that will give you a link, and then use this link in the mardown file. Like that we don't store any images in git

I've removed the images from the repository, created an issue with the screenshots for reference and used the links instead.

Nexucis:
nice thanks @indigo423. Actually you didn't need to create an issue. Just going to the template to be able to copy and past the image to generate the link. And then you can forgot the issue once you have the links. But no worries it's just a trick.

Thank you for doing it !

Nexucis (approved):

PR #2184, opened 2024-08-19
[ENHANCEMENT] Ignore Not Found errors in percli delete command

kamijin-fanta:
# Description

The current `percli delete` command outputs a Not Found error when a resource is not found. This is inconvenient when using a definition file that includes multiple existing resources for deletion.

For example, the command would result in the following output.

```
$ # add new resources
$ percli apply -f sandbox.json
object "Project" "project-1" has been applied
object "Role" "defaultrole" has been applied in the project "project-1"
object "User" "user1" has been applied
object "RoleBinding" "projectrolebinding" has been applied in the project "project-1"

$ # delete resources
$ percli delete -f sandbox.json
object "Project" "project-1" has been deleted
Error: something wrong happened with the request to the API.  Message: document not found StatusCode: 404
```

In this PR, I will make changes to ignore Not found errors.

This is an implementation idea. Feedback and discussion are welcome.

# Screenshots

not ui changes

# Checklist

## UI Changes

  include:

### DIFF
--- a/internal/cli/cmd/remove/remove.go
+++ b/internal/cli/cmd/remove/remove.go
@@ -14,6 +14,7 @@
 package remove
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"sort"
@@ -25,6 +26,7 @@ import (
 	"github.com/perses/perses/internal/cli/resource"
 	"github.com/perses/perses/internal/cli/service"
 	"github.com/perses/perses/pkg/client/api"
+	"github.com/perses/perses/pkg/client/perseshttp"
 	modelAPI "github.com/perses/perses/pkg/model/api"
 	modelV1 "github.com/perses/perses/pkg/model/api/v1"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


Nexucis:
While I agree it can be a good idea to be more permissive in this command, the example you are showing shouldn't give you this error.

If you apply a file and then use the same file to delete the matching document, it shouldn't return an error and should work like that.
Is there a bug somewhere ? Why the command doesn't find one resource ?

Nexucis:
regarding the way to implement it, another idea could be to collect all errors and then throw it at the end. So the command can go through all resources you wanted to delete, but at the end it gives you a full result on what happens and can fail (with error code).
What do you think ?

kamijin-fanta:
Thank you for the review! When a project is deleted, the associated Roles and RoleBindings, etc... are also deleted. Therefore, even if you use the JSON file that was used for creation, there might be cases where deletion fails. If necessary, I can prepare the file tomorrow.

Returning an error only after all resources have been added or deleted is a good idea to enhance usability.

Nexucis:
ah yeah you are right, I forgot that deleting the project is also deleting every resources related to the project. That makes more sense to ignore the 404 errors indeed mmm.

The issue with that is in other cases, you may want to show and return the error even if it's a 404.

celian-garcia:
My two cent here. I agree with @kamijin-fanta that as long as it displays what it is deleting that's okay to not return error. 

Also giving a full of not existing resource file, a message "Nothing found to be deleted" could be also good instead of empty answer.

If we want to go in error, then it makes me think a bit of the #1917 issue where it seems we're going for a --force flag. That could also be another option.

Nexucis (approved):
alright, you convinced me :).

LGTM

kamijin-fanta:
Thank you for the quick review and merge!

PR #2211, opened 2024-08-30
[FEATURE] Usage Tracking Frontend Work

sinkingpoint:
# Description

This is the other half of https://github.com/perses/perses/issues/1170

This adds in a provider that tracks metrics about a dashboard that is being loaded. Each individual panel uses that provider to store the state of it's queries, and then once all the queries are finished, we
ship the metrics off to the previously created /visit api.

# Screenshots

https://github.com/user-attachments/assets/f693294a-73ef-4dc5-a2a8-9d28ea68fdc7

# Checklist

## UI Changes

  include:

### DIFF
--- a/internal/api/impl/v1/view/endpoint.go
+++ b/internal/api/impl/v1/view/endpoint.go
@@ -52,13 +52,15 @@ func (e *endpoint) view(ctx echo.Context) error {
 		return apiInterface.HandleBadRequestError(err.Error())
 	}
 
-	claims := crypto.ExtractJWTClaims(ctx)
-	if claims == nil {
-		return apiInterface.HandleUnauthorizedError("missing claims")
-	}
+	if e.rbac.IsEnabled() {
+		claims := crypto.ExtractJWTClaims(ctx)
+		if claims == nil {
+			return apiInterface.HandleUnauthorizedError("missing claims")
+		}
 
-	if ok := e.rbac.HasPermission(claims.Subject, role.ReadAction, view.Project, role.DashboardScope); !ok {
-		return apiInterface.HandleUnauthorizedError(fmt.Sprintf("missing '%s' permission in '%s' project for '%s' kind", role.ReadAction, view.Project, role.DashboardScope))
+		if ok := e.rbac.HasPermission(claims.Subject, role.ReadAction, view.Project, role.DashboardScope); !ok {
+			return apiInterface.HandleUnauthorizedError(fmt.Sprintf("missing '%s' permission in '%s' project for '%s' kind", role.ReadAction, view.Project, role.DashboardScope))
+		}
 	}
 
 	if _, err := e.dashboardService.Get(apiInterface.NewPersesContext(ctx), apiInterface.Parameters{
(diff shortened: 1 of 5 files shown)


Nexucis:
@Gladorme or @AntoineThebaud if you can give to @sinkingpoint some feedback if you have a bit of time, that would be cool !

It's just to let him know if the overall approach looks ok to you. Thank you 😊

Gladorme:
I was wondering if we could use browser native APIs , but I never used this one or did something similar 🤔https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings. 

I think this feature should be optional and could be disabled by config, it should be optional too.

sinkingpoint:
I'm happy to put this behind a config flag, but the timing source is a bit difficult here - there's actually a big disconnect because all the panels function independently. That means there's no way to tie a given panel to an entry in the resource timings API, nor a way to look at that API directly and know which requests make up panel requests, and whether the requests you see from that API are actually all the panels - some may not have submitted their requests.

sinkingpoint:
@Gladorme do we have an existing framework for passing configs to the frontend?

Gladorme:
> @Gladorme do we have an existing framework for passing configs to the frontend?

We using the backend to pass the data to the frontend.
Using the [frontend config](https://github.com/perses/perses/blob/bbd58fa5c1cfcf2925416f1c8a27bcf474bfaf64/ui/app/src/model/config-client.ts#L142) + [backend config](https://github.com/perses/perses/blob/bbd58fa5c1cfcf2925416f1c8a27bcf474bfaf64/pkg/model/api/config/frontend.go#L50C6-L50C14).

It can be used after with [ConfigContextProvider](https://github.com/perses/perses/blob/bbd58fa5c1cfcf2925416f1c8a27bcf474bfaf64/ui/app/src/context/Config.tsx#L29) 😄

Gladorme:
In e2e, auth is disabled. It has found an issue with backend endpoint => [it should not check claim if auth is disabled  ](https://github.com/perses/perses/blob/7b7af5efb98090c32ee5016fc003a33a7d4cff32/internal/api/impl/v1/view/endpoint.go#L55)

Gladorme:
```go
func (e *endpoint) view(ctx echo.Context) error {
	view := v1.View{}
	if err := ctx.Bind(&view); err != nil {
		return apiInterface.HandleBadRequestError(err.Error())
	}

	if e.rbac.IsEnabled() {
		if claims == nil {
			return apiInterface.HandleUnauthorizedError("missing claims")
		}

		if ok := e.rbac.HasPermission(claims.Subject, role.ReadAction, view.Project, role.DashboardScope); !ok {
			return apiInterface.HandleUnauthorizedError(fmt.Sprintf("missing '%s' permission in '%s' project for '%s' kind", role.ReadAction, view.Project, role.DashboardScope))
		}
	}

	if _, err := e.dashboardService.Get(apiInterface.NewPersesContext(ctx), apiInterface.Parameters{
		Project: view.Project,
		Name:    view.Dashboard,
	}); err != nil {
		return apiInterface.HandleNotFoundError(err.Error())
	}

	return e.service.View(&view)
}

```

Is fixing the issue 😄

Gladorme:
And for info, the is auth is disabled here: https://github.com/perses/perses/blob/7b7af5efb98090c32ee5016fc003a33a7d4cff32/scripts/api_backend_dev.sh#L15

Can be tricky to find 😅

sinkingpoint:
Ah, good spot 🙏 Lemme fix that

Gladorme (approved):
LGTM! 
Maybe we should add more context like if it's has been loaded on a phone or laptop in a future PR. I think it should have a big impact on the renderDuration 🤔

nicolastakashi:
Amazing 🥳 
LGTM

nicolastakashi (approved):

sinkingpoint:
Thanks folks! Apologies it took so long 😅

PR #2218, opened 2024-09-03
[BUGFIX] migrate: ignore adhoc variables

7840vz:
# Description

migrate: ignore [adhoc](https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#add-ad-hoc-filters) variables not supported by Perses:

Error without ignoring it :

```
percli migrate -f dash.json -o yaml --online
Error: something wrong happened with the request to the API.  Message: bad request: unable to convert to Perses dashboard: spec.variables.2.spec: undefined field: hide StatusCode: 400
```
Example variable in grafana:

```
...
            {
                "datasource": {
                    "type": "prometheus",
                    "uid": "PC7A11E9A55DE2B10"
                },
                "description": "Metrics filter.",
                "label": "mfilter",
                "name": "mfilter",
                "type": "adhoc"
            },
...
```

# Screenshots

# Checklist

### DIFF
--- a/internal/api/migrate/mapping.cuepart
+++ b/internal/api/migrate/mapping.cuepart
@@ -70,7 +70,7 @@ spec: {
     display: {
         name: #grafanaDashboard.title
     }
-    variables: [ for _, grafanaVar in #grafanaDashboard.templating.list {
+    variables: [ for _, grafanaVar in #grafanaDashboard.templating.list if grafanaVar.type != "adhoc" {
         _displayConversion: {
             #var: _
 


7840vz:
sorry, I think adhoc is handled, it is that hide field is not always present.

AntoineThebaud:
@7840vz actually we have checks against null fields to avoid such errors in many cases like
```cue
if #var.hide != _|_
```
But here it's missing at line 117
https://github.com/perses/perses/blob/4ffc25a5f77a6c32fcc611ac1932c53d86fb5bd5/internal/api/migrate/mapping.cuepart#L117
We are assuming `hide` is always present, but it looks like as most fields it's not always the case..

There's indeed something to improve here to avoid the error your encountered but there's 2 lines to change actually.. I will open a PR asap

PR #2237, opened 2024-09-09
Perses channels are moving to CNCF slack

Nexucis:
Part of https://github.com/cncf/toc/issues/1411 tasks


celian-garcia (approved):
Too bad you can't have a direct link to the channel. I guess it's not possible. Approved

PR #2246, opened 2024-09-12
[ENHANCEMENT] Return invalid_request oauth2 error when user sync failed

celian-garcia:
# Description

People used to login with different providers but have the same email address. This is failing on purpose for security reasons but with an Internal Server Error which is inappropriate. Let's give them some clues on what the failure was.

# Screenshots

# Checklist

## UI Changes

  include:


internal/api/impl/auth/service.go
```diff
@@ -53,7 +54,7 @@ func saveProviderInfo(old v1.UserSpec, uInfoProvider v1.OAuthProvider) (v1.UserS
 	foundPerfectMatch := false
 	for _, provider := range old.OauthProviders {
 		if provider != uInfoProvider {
-			return old, false, errors.New("this user is already registered with a different oauth provider context")
+			return old, false, fmt.Errorf("this user is already registered with a different oauth provider context (%s)", provider)
```
celian-garcia, line 57: I'm realizing I forgot provider.Name

celian-garcia: I'm using provider.Issuer as we don't save the provider name in the user. This will be the public URL of the provider. The one the user is redirected to when login so nothing secret.

Gladorme (approved):

PR #2259, opened 2024-09-19
[ENHANCEMENT] Migrate: rename Value to value in table migrations

7840vz:
# Description

Small improvement to table migrations:
Special 'Value' column from Grafana should be 'value' in order to match Perses 'value' column with actual values.

# Screenshots
old migration:
image
after migration:
image

# Checklist

### DIFF
--- a/cue/schemas/panels/table/migrate.cue
+++ b/cue/schemas/panels/table/migrate.cue
@@ -18,7 +18,11 @@ if #panel.type != _|_ if #panel.type == "table" {
 		// We use the future 'header' information as a key for both maps here, because this is the common denominator between the two sources
 		// Indeed in grafana the fieldconfig's overrides are matched against the final column name (thus potentially renamed))
 		_renamedMap: {if #panel.transformations != _|_ for transformation in #panel.transformations if transformation.id == "organize" for technicalName, prettyName in transformation.options.renameByName {
-			"\(prettyName)": technicalName
+			"\(prettyName)": [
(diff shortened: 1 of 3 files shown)


AntoineThebaud (approved):
LGTM! Thank you for this fix 🙏 

NB: I just noticed that column renaming can also be achieved via field overrides, where this special case of Value should also be taken into account. I will tackle this in another PR.

7840vz:
Thank you! Do you mean this https://github.com/perses/perses/issues/2260 can be fixed by overrides?

> LGTM! Thank you for this fix 🙏
> 
> NB: I just noticed that column renaming can also be achieved via field overrides, where this special case of Value should also be taken into account. I will tackle this in another PR.

AntoineThebaud:
Well regarding excluding columns, the migration logic already supports show/hide from the "Organize fields" transformation:

https://github.com/perses/perses/blob/b3957aa918467e5bbff23d615553f865c433fc9d/cue/schemas/panels/table/migrate.cue#L12-L15

But regarding the "Filter fields by name" transformation you mentioned in #2260, I need to think about it but I don't see an easy way to achieve the migration for now...

AntoineThebaud:
> LGTM! Thank you for this fix 🙏
> 
> NB: I just noticed that column renaming can also be achieved via field overrides, where this special case of Value should also be taken into account. I will tackle this in another PR.

This has been tackled in https://github.com/perses/perses/pull/2273.

PR #2273, opened 2024-09-26
[ENHANCEMENT] Table panel migration: support column rename from field overrides

AntoineThebaud:
# Description

This PR improves the table migration logic by supporting migrating column renames from both the "Organize" transformation & the "field overrides" options from Grafana.

NB1: this remains a best effort migration. In the new test file added, the 2 last panels are not accurately migrated, but these are tackling some corner cases that's not worth investing on imho.

NB2: the migration logic was quite heavily revamped here, hopefully the new code should allow for easier extension in the future: table panels in Grafana provides many customization capabilities, so we'll most likely have to enrich this code again later..

# Checklist

### DIFF
--- a/cue/schemas/panels/table/migrate.cue
+++ b/cue/schemas/panels/table/migrate.cue
@@ -9,37 +9,72 @@ if #panel.type != _|_ if #panel.type == "table" {
 			][0]
 		}
 
-		_excludedColumns: [if #panel.transformations != _|_ for transformation in #panel.transformations if transformation.id == "organize" for excludedColumn, value in transformation.options.excludeByName if value {
-			name: excludedColumn,
-			hide: true
-		}]
-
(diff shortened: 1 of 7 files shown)


celian-garcia (approved):
LGTM
I'm wondering, do you think it's possible to monitor the memory usage of each unit test? I mean as we had some memory issues, that would be cool to ensure that when we modify the cuelang schemas, we don't introduce a new issue like this.

AntoineThebaud:
> LGTM I'm wondering, do you think it's possible to monitor the memory usage of each unit test? I mean as we had some memory issues, that would be cool to ensure that when we modify the cuelang schemas, we don't introduce a new issue like this.

I have no idea right now how to simply put that in place but I guess it can be achieved, indeed that's a good remark & idea... I will think about it.

PR #2275, opened 2024-09-27
[DOC] Table: Add datamodel documentation

AntoineThebaud:
NB: I wonder if we could just somehow include/redirect to the cuelang files instead of maintaining this doc on the side, I think the CUE syntax is a bit clearer than the one we use in this doc, it basically describes the same thing & it would avoid misalignment between code & doc. 

# Checklist

### DIFF
--- a/docs/plugins/panels.md
+++ b/docs/plugins/panels.md
@@ -32,6 +32,13 @@ spec:
   text: <string>
 ```
 
+## PieChart
+
+```yaml
+kind: "PieChart"
+# TODO document the spec of PieChart
+```
+
 ## StatChart
 
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/plugins/panels.md
```diff
@@ -58,6 +65,28 @@ kind: "ScatterChart"
...
+### Column Settings specification
+
+```yaml
+name:  <string>
+[ header:  <string> ]
+[ headerDescription:  <string> ]
+[ cellDescription: <string> ]
+[ align: <enum = "left" | "center" | "right"> ]
+[ enableSorting: <boolean> ]
+[ width: <number | "auto"> ]
```
AntoineThebaud, line 86: not sure for the syntax here...

nicolastakashi: I think we can just remove the square brackets

AntoineThebaud: Sorry I should have precised, I meant the syntax for the disjunction here, I think the "enum" we usually use (like for `align` right above) doesnt apply here because in the case of `width` the user can either provide a number or the "auto" string. Any other string is invalid.
In both CUE & typescript the syntax is as simple as `width?: number | 'auto';`

Issue #2269, opened 2024-09-25
percli dac build -d should return a non-zero exit code in case of build failure


flazarus1A:
### What did you do?

I have a project with 2 dashboards located in the same folder. I build them with the -d flag for convenience.

### What did you expect to see?

In case of build failure of any dashboard in the folder, the exit code should be non-zero for scripting consistency. This is already the case for the -f flag for a single dashboard.

### What did you see instead? Under which circumstances?

If any dashboard (or all) in the specified folder does not build, percli successfully logs the problem(s) in the console, but the exit code always remains 0.

### System information

Windows / git bash

### Perses version

0.47.1

### Perses configuration file

```yaml
N/A
```

### Logs

```text
N/A
```

PR #2279, opened 2024-09-30
[BUGFIX] Fix exit code when error occured in cmd dac build

Nexucis:
This PR is introducing a way to send error message on stderr.

Having access to the stderr channel is helping to fix https://github.com/perses/perses/issues/2269. Because now we can send multiple message to stderr without stopping processing the files in a directory.

If one file failed, it will afterward return a non 0 exit code.


AntoineThebaud (approved):
Thanks! My code was quite dumb here 🙈

Issue #2195, opened 2024-08-25
Create and GET project causes 500


agardnerIT:
### What did you do?

1. Spin up perses
2. Create a project via ui: `proj1`
3. `http://localhost:8080/api/v1/projects` now gives `{"message":"internal server error"}`

```
ERRO[2024-08-25T14:47:03+10:00] unexpected error not handle                   error="json: error calling MarshalJSON for type json.RawMessage: invalid character 'k' looking for beginning of value"
INFO[2024-08-25T14:47:03+10:00]                                               method=GET status=500 uri=/api/v1/projects
```

### What did you see instead? Under which circumstances?

500 error

### System information

Darwin 23.6.0 arm64

### Perses version

0.47.1

Nexucis:
Could you share the server config please ? 

I just tried with the demo instance and I have no issue creating and getting a project.

agardnerIT:
Sure, I downloaded and extracted the `0.47.1` release and running from that extracted directory:

```
% pwd 
~/Downloads/perses_0.47.1_darwin_arm64

% ls
CHANGELOG.md    README.md       cue             data            percli
LICENSE         config.yaml     cue.mod           perses


% ./perses --config=config.yaml
```

## config.yaml
Built from instructions [here](https://perses.dev/docs/perses/v0.47.0/user-guides/configuration.md/)
```
database:
  file: # the configuration when you want to use the filesystem as a database. Note that you can configure it using the flags, which gives you the choice to not create a configuration file just for that.
    folder: "./data" # It's the path where the file will be read / stored
    extension: "yaml" # The extension of the files read / stored. "yaml" or "json" are the only extension accepted. Yaml is the default one
schemas:
  panels_path: "cue/schemas/panels"
  queries_path: "cue/schemas/queries"
  datasources_path: "cue/schemas/datasources"
  variables_path: "cue/schemas/variables"
```

Nexucis:
This is really a mystery for me.

- I have download the exact same binary/archive (`perses_0.47.1_darwin_arm64`)

- I have the same config

```yaml
database:
  file:
    folder: "./data"
    extension: "json"

schemas:
  panels_path: "./cue/schemas/panels"
  queries_path: "./cue/schemas/queries"
  datasources_path: "./cue/schemas/datasources"
  variables_path: "./cue/schemas/variables"
  interval: "6h"
```

And still I am not able to reproduce the issue you are facing :(

[...]

PR #2283, opened 2024-10-01
[BUGFIX] Fix database folder in default config file for the archives

Nexucis:
in the archive the database folder cannot be `perses` since it's the name of the executable.


Nexucis:
coming from this issue: https://github.com/perses/perses/issues/2195#issuecomment-2358031535

AntoineThebaud (approved):

PR #2291, opened 2024-10-01
[IGNORE] DaC CUE SDK: refactor/simplify dac-utils code for variables (part 2)

AntoineThebaud:
# Description

Same refactor as done in https://github.com/perses/perses/pull/2289 but for TextVariable & StaticListVariable.
Also took the opportunity to cover StaticListVariable in the DaC unit tests.

# Checklist

### DIFF
--- a/cue/dac-utils/variable/text/text.cue
+++ b/cue/dac-utils/variable/text/text.cue
@@ -16,14 +16,17 @@ package text
 import (
 	v1Dashboard "github.com/perses/perses/cue/model/api/v1/dashboard"
 	varBuilder "github.com/perses/perses/cue/dac-utils/variable"
-	v1Variable "github.com/perses/perses/cue/model/api/v1/variable"
 )
 
-#kind: varBuilder.#kind & "TextVariable"
-#name: varBuilder.#name
-#display?: v1Variable.#Display & {
-	hidden: bool | *false
-}
+// include the definitions of varBuilder at the root
+varBuilder
+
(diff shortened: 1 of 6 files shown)


cue/dac-utils/variable/text/text.cue
```diff
@@ -16,14 +16,17 @@ package text
 import (
 	v1Dashboard "github.com/perses/perses/cue/model/api/v1/dashboard"
 	varBuilder "github.com/perses/perses/cue/dac-utils/variable"
-	v1Variable "github.com/perses/perses/cue/model/api/v1/variable"
 )
 
-#kind: varBuilder.#kind & "TextVariable"
-#name: varBuilder.#name
-#display?: v1Variable.#Display & {
```
celian-garcia, line 24: Voodoo magic or did you remove something here with the remove of v1Variable.#Display ? 
I definitely need to ramp up on that language. I would like to make an accelerated course one day of we can program that together :)

AntoineThebaud: Nope it's just that now it gets included by the new code you see on the right
```cue
// include the definitions of varBuilder at the root
varBuilder
```

with varBuilder defining exactly the same for Display: https://github.com/perses/perses/blob/main/cue/dac-utils/variable/variable.cue#L22

AntoineThebaud: And ofc would be happy to share knowledge with you anytime ;)

PR #2297, opened 2024-10-02
[ENHANCEMENT] Provide a way to set auth in Perses http client

Nexucis:


pkg/client/perseshttp/client.go
```diff
@@ -102,7 +103,23 @@ type RESTClient struct {
...
+	authMutex sync.RWMutex
+}
+
+func (c *RESTClient) SetBasicAuth(username, password string) {
+	c.authMutex.Lock()
+	c.basicAuth = &secret.BasicAuth{
+		Username: username,
+		Password: password,
+	}
+	c.authMutex.Unlock()
```
AntoineThebaud, line 116: any reason to put the unlock at the end in these setters, vs in a defer in newRequest()?

Nexucis: that's because here there is no return and it's a linear code. So there is only one way to leave it.

In the newRequest, there is something to return, so defer is necessary

celian-garcia (approved):

pkg/client/perseshttp/client.go
```diff
@@ -102,7 +103,23 @@ type RESTClient struct {
 	// base is the root URL for all invocations of the client
 	BaseURL *url.URL
 	// Set specific behavior of the client.  If not set http.DefaultClient will be used.
-	Client *http.Client
+	Client    *http.Client
+	authMutex sync.RWMutex
+}
+
+func (c *RESTClient) SetBasicAuth(username, password string) {
```
celian-garcia, line 110: What is the difference from having it configured in the rest config client ? I mean for the basic Auth in particular 
For authorization I would maybe understand even if it is for client credentials, I'd prefer to have a way to configure my own client coming from oAuth2 library like I did in my Pr

Nexucis: it will avoid recreating entirely the RESTClient.

PR #2300, opened 2024-10-03
[BREAKINGCHANGE] Refactor package pkg/client to handle native/oauth auth natively

Nexucis:
This PR is changing the http client of Perses. The main goal is to handle the Oauth and native auth easily.

For that I am introducing a custom round tripper in the package `transport`. That's the part that is handling automatically the native authentification.

We are using the package `oauth2` to handle oauth authentication.

As a side effect, the login requests are now returning the object `oauth2.Token`. Doing that help a lot to handle the different cases (i.a, does the token expire, the type of the token ...etc.) 

With this PR handling configuring oauth with the perses client will be like that:

```go
client, err := config.NewFromConfig(config.RestConfigClient {
	URL: "http://localhost:8080",
        Auth: &config.AuthConfig {
		OauthConfig: &config.OauthConfig{
		        ClientID:     "MyClientID",
		        ClientSecret: "MyClientSecret",
		        TokenURL:     "http://localhost:8080/api/v1/auth/providers/oauth/github/token",
		        AuthStyle:    oauth2.AuthStyleInHeader,
		}
	}

})

// Make any classic Perses query 
_, err = v1.NewWithClient(client).Project().List("")

### DIFF
--- a/docs/user-guides/configuration.md
+++ b/docs/user-guides/configuration.md
@@ -494,17 +494,38 @@ discovery_name: <string>
 # URL of the HTTP server exposing the global datasource list to retrieve.
 url: <url>
 
-[ basicAuth: <Basic Auth specification> ]
+[ auth: <Auth specification> ]
 
 # The HTTP authorization credentials for the targets.
 # Basic Auth and authorization are mutually exclusive. Use one or the other not both at the same time.
 [ authorization: <Authorization specification> ]
 
 # Config used to connect to the targets.
-[ tlsConfig: <TLS Config specification> ]
+[ tls_config: <TLS Config specification> ]
 
 headers:
   [<string>:<string>]
+```
+
+##### Auth
+
+```yaml
+[ basic_auth: <Basic Auth specification> ]
+
+[ oauth_config: <Oauth specification> ]
+```
+
+##### Oauth specification
+
+```yaml
+# ClientID is the application's ID.
+client_id: <string>
+
+# ClientSecret is the application's secret.
+client_secret: <string>
+
+# TokenURL is the resource server's token endpoint URL. This is a constant specific to each server.
+token_url: <string>
 ```
 
 ##### Basic Auth specification
--- a/docs/user-guides/configuration.md
+++ b/docs/user-guides/configuration.md
@@ -512,7 +512,7 @@ headers:
 ```yaml
 [ basic_auth: <Basic Auth specification> ]
 
-[ oauth_config: <Oauth specification> ]
+[ oauthg: <Oauth specification> ]
 ```
 
 ##### Oauth specification
--- a/internal/api/e2e/api/auth_test.go
+++ b/internal/api/e2e/api/auth_test.go
@@ -16,7 +16,6 @@
 package api
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -33,7 +32,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 	"golang.org/x/oauth2"
-	"golang.org/x/oauth2/clientcredentials"
 )
 
 func TestAuth(t *testing.T) {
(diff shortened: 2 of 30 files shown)

### CODE REVIEW

File: docs/user-guides/configuration.md [hunk 1]
```diff
@@ -494,19 +494,40 @@ discovery_name: <string>
...
 headers:
   [<string>:<string>]
 ```
 
+##### Auth
+
+```yaml
+[ basic_auth: <Basic Auth specification> ]
+
+[ oauth_config: <Oauth specification> ]
```
celian-garcia, line 515: I fill ``oauth_config```heavy and not consistent with ``basic_auth`` 
Could you change to ``oauth``?

pkg/client/config/config.go
```diff
@@ -0,0 +1,189 @@
...
+	Auth *AuthConfig `json:"auth,omitempty" yaml:"auth,omitempty"`
+	// The HTTP authorization credentials for the targets.
+	Authorization *secret.Authorization `json:"authorization,omitempty" yaml:"authorization,omitempty"`
+	// TLSConfig to use to connect to the targets.
+	TLSConfig *secret.TLSConfig `json:"tls_config,omitempty" yaml:"tls_config,omitempty"`
+	Headers   map[string]string `json:"headers,omitempty" yaml:"headers,omitempty"`
+}
+
+// NewFromConfig create an instance of RESTClient using the config passed as parameter
+func NewFromConfig(config RestConfigClient) (*perseshttp.RESTClient, error) {
```
celian-garcia, line 134: Not possible to have the client ctor inside the perseshttp package? 
To me it should be either ``perseshttp.NewFromConfig``, or eventually ``config.NewRESTClient``

pkg/client/config/config.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/perses/perses/pkg/model/api"
+	"github.com/perses/perses/pkg/model/api/v1/common"
+	"github.com/perses/perses/pkg/model/api/v1/secret"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+const connectionTimeout = 30 * time.Second
+
+type PublicOauthConfig struct {
```
celian-garcia, line 34: now that it's in a ``config`` package, you don't need all these Config suffixes

internal/api/e2e/client/test_utils.go
```diff
@@ -22,11 +22,21 @@ import (
...
+	"github.com/perses/perses/pkg/client/config"
 	"github.com/perses/perses/pkg/model/api"
 	"github.com/perses/perses/pkg/model/api/v1/common"
 )
 
+func withAuthClient(t *testing.T, testFunc func(v1.ClientInterface, dependency.PersistenceManager) []api.Entity) {
+	server, _, persistenceManager := e2eframework.CreateServer(t, e2eframework.DefaultAuthConfig())
+	defer server.Close()
+	persesClient := createAuthClient(t, server)
+	usrEntity := e2eframework.NewUser("foo", "$2a$10$iCemKjvjN.ieqJOPlEL5keGRLAGKRGBNjph2N8uY.4XFPZYwcBT8K") // the encrypted value for password
```
celian-garcia, line 34: too precise to deserve a test_utils function in my opinion. the withClient as well. 

You can directly use the withServer config. It provides now the URL in the ``testFunc``
e.g https://github.com/perses/perses/blob/a403f6433958d96f85d5e80c8492c4bc03fb484f/internal/api/e2e/api/auth_test.go#L437-L450

pkg/client/config/config.go
Nexucis: Ih had to do this to avoid a circular dependencies.

Yeah I think renaming it `config.NewRESTClient` makes sense. Thanks

internal/api/e2e/client/test_utils.go
celian-garcia: I would even remove the ``withClient`` function

celian-garcia (changes requested):
That's amazing to see how the interface is improved. Thank you @Nexucis, this will improve a lot our codes that request Perses.

I have some suggestion about code readability mainly WDYT?

internal/api/e2e/client/test_utils.go
Nexucis: withClient simplify the creation of the client to avoid to do it for every tests.

Nexucis: I have simplifies the oauth tests ;)

internal/api/e2e/api/auth_test.go
```diff
@@ -436,28 +436,29 @@ func TestAuth_OIDCProvider_Token_WithLib(t *testing.T) {
 	// Server with OIDC provider configured.
 	e2eframework.WithServerConfig(t, conf, func(server *httptest.Server, expect *httpexpect.Expect, manager dependency.PersistenceManager) []modelAPI.Entity {
 		usersCreatedByTheSuite := []modelAPI.Entity{
-			e2eframework.NewUser("MyClientID"),
+			e2eframework.NewUser("MyClientID", "password"),
```
celian-garcia, line 439: to be very accurate, in this case password would be ""

celian-garcia (approved):
Cool, a slight comment to use password = "" when you init fake user that will be cleaned in oauth/oidc tests. This is to have it closer to what it's effectively created by the server. But in the end it serves just to not confuse reader.

PR #2311, opened 2024-10-07
[ENHANCEMENT] Configuration of the HTTP client for oauth/oidc providers

### LINKED ISSUE #2302: Allowing for non tls communication
### Proposal

Hello,

Recently, I have been using Perses and had to configure authentication to access my dashboards. So I set up a keycloak server and parametered Perses to have my dashboards not accessible from random people. The configuration has been working fine since I had a session registered and logged in in Keycloak, but Perses keeps saying it can't verify certificates authenticity. Which I guess is because I have self signed certificates from keycloak. I keep having a message saying "Internal error server" instead of the page that is configured after redirection. I assume that Perses can't process authentication without tls enabled ?

So my request is the following :

Could it be possible to add non tls authentication/communication between Perses and other services ? More generally, it would be nice to have the possibility to enable or disable tls directly in the configuration of Perses. Also would be nice to have some configuration to generate self signed certificates when using tls, and/or have the possibility to install our own certificates easily through configuration.

What do you think ? Thanks for your work so far.

celian-garcia:
# Description

This PR allows us to set a timeout and various TLS configurations into the HTTP client that is used to make the various queries to external auth providers.

Hopefully fixes #2302 

It makes a little cleanup as well in the auth config documentation that was missing some elements.

# Screenshots

# Checklist

### DIFF
--- a/internal/api/impl/auth/auth.go
+++ b/internal/api/impl/auth/auth.go
@@ -16,8 +16,10 @@ package auth
 import (
 	"errors"
 	"fmt"
+	promConfig "github.com/prometheus/common/config"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/labstack/echo/v4"
 	"github.com/perses/perses/internal/api/crypto"
@@ -58,6 +60,23 @@ func getRedirectURI(r *http.Request, authKind string, slugID string) string {
 	return rd.String()
 }
 
+// newHTTPClient is a simple http client builder designed to be used for the queries to external authentication providers.
+func newHTTPClient(httpConfig config.HTTP) (*http.Client, error) {
+	httpClient := &http.Client{
+		Timeout: time.Duration(httpConfig.Timeout),
+	}
+	if httpConfig.TLSConfig != nil {
+		tlsConfig, err := promConfig.NewTLSConfig(httpConfig.TLSConfig)
+		if err != nil {
+			return nil, err
+		}
+		if tlsConfig != nil {
+			httpClient.Transport = &http.Transport{TLSClientConfig: tlsConfig}
+		}
+	}
+	return httpClient, nil
+}
+
 type endpoint struct {
 	endpoints       []route.Endpoint
 	jwt             crypto.JWT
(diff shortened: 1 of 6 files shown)


internal/api/impl/auth/auth.go
```diff
@@ -58,6 +60,23 @@ func getRedirectURI(r *http.Request, authKind string, slugID string) string {
...
+	httpClient := &http.Client{
+		Timeout: time.Duration(httpConfig.Timeout),
+	}
+	if httpConfig.TLSConfig != nil {
+		tlsConfig, err := promConfig.NewTLSConfig(httpConfig.TLSConfig)
+		if err != nil {
+			return nil, err
+		}
+		if tlsConfig != nil {
+			httpClient.Transport = &http.Transport{TLSClientConfig: tlsConfig}
```
Nexucis, line 74: I think something like that will be cleaner:

https://github.com/perses/perses/blob/release/v0.48/pkg/client/config/config.go#L157-L165

pkg/model/api/config/authentication.go
```diff
@@ -16,6 +16,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"github.com/prometheus/common/config"
```
Nexucis, line 19: import not in the right place, shouldn't be mixed with native import

celian-garcia: New workstation, have to configure again my IDE hehe

internal/api/impl/auth/auth.go
Nexucis: would be great to be able to use TLS configuration for oauth authentication in the client as well. If you can find the way, that would be cool !

Also I think it will be nice to have it in v0.48.0

celian-garcia: The client doesn't make direct queries to the oauth/oidc provider. These queries are done by the backend exactly in the code I modified. 

Or are you taking about a TLS config to query the perses backend? I think this is already configurable no?

Nexucis: ah interesting, I forgot Perses was doing the request to the provider.

But yeah also the TLS config for querying the perses backend. And actually when we are using oauth, I didn't find the way to inject the tls config into the http client of the oauth client.

So if you know how to do it, that would be great !

celian-garcia: This is done in the newQueryContext function of that PR. I attached in the comment a GitHub issue for reference.

celian-garcia: Here is the ref 
https://github.com/golang/oauth2/issues/187

Nexucis: ah yeah I see. Thanks, I will send a PR to modify the perses client

Nexucis (approved):

Issue #2039, opened 2024-06-05
Memory leak with /migrate


celian-garcia:
### What did you do?

I made a script to migrate all dashboards from Amadeus Grafana instance to Perses instance. It works well but at some point, when it arrives to some dashboard there is a big memory leak.

I'm doing the migrate for around 1k dashboards. 
And it seems to leak when it arrives to some dashboards with row repeat. I'm not sure yet it's linked but I will dig further to isolate the dashboards and see if I can reproduce with a few examples. 

Of course I cannot share the dashboards

### What did you expect to see?

Memory correctly released by the Garbage Collector

### What did you see instead? Under which circumstances?

![image](https://github.com/perses/perses/assets/10258608/90716d04-63f1-4d57-8734-1e98804d926f)

### System information

Linux 6.5.0-28-generic x86_64

### Perses version

e9dddceb0fdf640e9b0ebdb278b0dde8b5f49216

celian-garcia:
@luweglarz generated some flamegraph for us with pyroscope and at the moment where the memory leaks. 
https://flamegraph.com/share/996769bd-2e45-11ef-ab1b-b6898751f631
![image](https://github.com/perses/perses/assets/10258608/8762a218-493e-4229-a43d-bd71ce7a4acb)

CPU as well
https://flamegraph.com/share/bdbbd4ec-2e45-11ef-aba3-6a3d1814cbe4
![image](https://github.com/perses/perses/assets/10258608/9c0c2945-396f-4e53-aa2f-50f1703d1150)

I think it can be interesting.

celian-garcia:
![Capture d’écran du 2024-07-10 10-55-49](https://github.com/perses/perses/assets/10258608/ec21ede4-0954-47eb-bda2-b99d5d375a14)

khimick:
Hello, I can confirm this behavior. When perses try to migrate some grafana dashboards it return errors and continue consume memory.  It take all accessible memory and kills by OOM killer.

Nexucis:
Thank you for your feedback @khimick that's really useful you confirm you have the same behaviour : a.k.a when there are errors when migrating, it's starting to leak.

I think to move on this topic, we will need to have a little set of dashboards that allow to reproduce the issue.

On Amadeus side our set of dashboard is too big (~5k), and we didn't find exactly what is the exact dashboard that start the leak. For sure we will need more time and work to isolate the one that caused it.

[...]

PR #2312, opened 2024-10-08
[ENHANCEMENT] Grafana migration: use new CUE evaluator to improve memory usage

AntoineThebaud:
# Description

This PR bumps the CUE dep to v0.11.0-alpha3 to allow enabling the new "EvalV3" evaluator from CUE with the latest improvements brought to it.

By doing so we were hoping to solve the memory leak issue we encounter when migrating dashboards (https://github.com/perses/perses/issues/2039). Unfortunately it doesn't, the memory usage still increase after each migration & doesnt drop before several minutes.

Still, with evalv3 the memory used is much less than before (you also see the difference in response time in the UI of course):
- When migrating a big dashboard (23k lines-long json): **~15GB** with old evaluator **vs ~3GB** with the new one.
- When migrating the blackbox exporter dashboard [attached here](https://github.com/perses/perses/issues/2039#issuecomment-2355022470): **~3.7GB** with the old evaluator **vs ~175MB** with the new one.
- When migrating the haproxy dashboard [attached here](https://github.com/perses/perses/issues/2039#issuecomment-2355022470): **~7.2GB** with the old evaluator **vs ~150MB** with the new one.

> [!WARNING]  
> **Side effect of this change**: the new evaluator is still under development & got some bugs. Regarding our usage of CUE for grafana migration, using it unfortunately breaks some valid logic we have for the Table panel migration. Therefore I brought back the migration logic we had prior from https://github.com/perses/perses/pull/2273 to still get the benefits of the memory usage reduction - at the cost of losing the table's field overrides previously supported.

# Checklist


Nexucis:
maybe we should keep the dashboard that contains the table to use them to test it back the migration when it will be fixed for the table ?

AntoineThebaud:
> maybe we should keep the dashboard that contains the table to use them to test it back the migration when it will be fixed for the table ?

yeah ok I can try to bring the old logic that was revamped with https://github.com/perses/perses/pull/2273

Nexucis (approved):

PR #2313, opened 2024-10-08
[ENHANCEMENT] Use TLS config in oauth client config

Nexucis:
use TLS config when configuring the oauth client in `pkg/client`

### DIFF
--- a/pkg/client/config/config.go
+++ b/pkg/client/config/config.go
@@ -15,6 +15,7 @@ package config
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
@@ -119,6 +120,27 @@ type Auth struct {
 	BasicAuth  *secret.BasicAuth `json:"basic_auth,omitempty" yaml:"basic_auth,omitempty"`
 }
 
+func (a *Auth) Validate() error {
+	if a == nil {
+		return nil
+	}
+	nbAuthConfigured := 0
+	if a.NativeAuth != nil {
+		nbAuthConfigured++
+	}
+	if a.Oauth != nil {
+		nbAuthConfigured++
+	}
+	if a.BasicAuth != nil {
+		nbAuthConfigured++
+	}
+
+	if nbAuthConfigured > 1 {
+		return fmt.Errorf("only one type of authentication should be configured")
+	}
+	return nil
+}
+
 // RestConfigClient defines all parameters that can be set to customize the RESTClient
 type RestConfigClient struct {
 	URL  *common.URL `json:"url" yaml:"url"`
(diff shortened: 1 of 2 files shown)


Nexucis:
part of #2311

pkg/client/config/config.go
```diff
@@ -164,19 +159,32 @@ func NewRESTClient(config RestConfigClient) (*perseshttp.RESTClient, error) {
...
 		}
+		if config.Auth.Oauth != nil {
+			oauthConfig := &clientcredentials.Config{
+				ClientID:     config.Auth.Oauth.ClientID,
+				ClientSecret: config.Auth.Oauth.ClientSecret,
+				TokenURL:     config.Auth.Oauth.TokenURL,
+				Scopes:       config.Auth.Oauth.Scopes,
+				AuthStyle:    config.Auth.Oauth.AuthStyle,
+			}
+			ctx := context.WithValue(context.Background(), oauth2.HTTPClient, roundTripper)
```
celian-garcia, line 178: I think this the http client that should go under this context value, not the transport

Nexucis: oh yeah good call, thanks !

Nexucis: done

pkg/client/config/config.go
```diff
@@ -119,6 +120,18 @@ type Auth struct {
 	BasicAuth  *secret.BasicAuth `json:"basic_auth,omitempty" yaml:"basic_auth,omitempty"`
 }
 
+func (a *Auth) Validate() error {
+	if a == nil {
+		return nil
+	}
+	if (a.NativeAuth != nil && (a.Oauth != nil || a.BasicAuth != nil)) ||
```
AntoineThebaud, line 127: I'd replace

```go
if (a.NativeAuth != nil && (a.Oauth != nil || a.BasicAuth != nil)) ||
	(a.Oauth != nil && (a.NativeAuth != nil || a.BasicAuth != nil)) ||
	(a.BasicAuth != nil && (a.NativeAuth != nil || a.Oauth != nil)) {
	return fmt.Errorf("only one type of authentication should be configured")
}
```

by

```go
nbAuthConfigured := 0
if a.NativeAuth != nil {
	nbAuthConfigured++
}
if a.Oauth != nil {
	nbAuthConfigured++
}
if a.BasicAuth != nil {
	nbAuthConfigured++
}

if nbAuthConfigured > 1 {
	return fmt.Errorf("only one type of authentication should be configured")
}
```

celian-garcia (approved):

AntoineThebaud (approved):

pkg/client/config/config.go
Nexucis: oh yeah I like that. Thanks !

PR #2332, opened 2024-10-11
[BUGFIX] Migration: Fix table migration fail when excludeByName is not present in transformation="organize"

7840vz:
# Description
Fix table migration fail when excludeByName is not present in transformation="organize".
Note: Logic is currently disable altogther.
# Checklist


AntoineThebaud (approved):
Thanks for this (future) fix! Too bad we have to disable this whole part for now :l

Issue #1419, opened 2023-09-08
Show error message in query editor when query is invalid


AntoineThebaud:
### Proposal

This is something lacking when you have more than 1 query defined in the considered panel: if you have 2+ queries and one has invalid syntax, the trends are all displayed except the invalid one, without any error showing.
Thus it would be good to display explicitly what's wrong with each invalid query independtly (e.g just aside/below each query input)

andreasgerstmayr:
This is also relevant for the explore page.

+1 on showing the error next to the query instead (or additional?) to the panel component.

AntoineThebaud:
FYI I'm currently tackling this one - ⚠️ in the case of the PrometheusTimeSeriesQuery plugin only - with the coming PromQL debugger feature.

AntoineThebaud:
> FYI I'm currently tackling this one - ⚠️ in the case of the PrometheusTimeSeriesQuery plugin only - with the coming PromQL debugger feature.

So, even though we now have the PromQL debugger for Prometheus—which can help in some cases (e.g., syntax errors are reported the same way both with the parsing query and the data fetching query)—it doesn't cover all scenarios and, of course, doesn't apply to other datasources. Therefore, we still want a generic solution to display any error from any data source, returned by a given query, alongside that query.

Gladorme:
Taking this one

AntoineThebaud:
Closed by https://github.com/perses/perses/pull/3532

PR #2344, opened 2024-10-15
[FEATURE] PromQL Tree view / Debugger

AntoineThebaud:
# Description

This PR enrich the PromQL editor with a toggleable Tree view that helps debugging your prom queries. This tree is a read-only view that gets computed/refreshed after each query update (when the input loses focus).

It is based on the Tree view available in the new UI of Prometheus 3.0, itself inspired by Promlens (more about it [here](https://promlabs.com/blog/2024/09/11/a-look-at-the-new-prometheus-3-0-ui/#promlens-style-query-tree-view)). Most of the underlying logic was actually copy-pasted from Prometheus repo.

> [!IMPORTANT]  
> This feature is only available when using a datasource compatible with Prometheus 3.0 specs. Indeed it depends on a new endpoint `api/v1/parse_query` that is only available in the new version. FYI Thanos team planned to [implement the same in Thanos](https://github.com/thanos-io/thanos/issues/7727).

image

It is available in all the places where users can provide promQL expressions (metrics explorer, panel editor & variable editor)

image

> [!NOTE]  
> This PR partially tackles https://github.com/perses/perses/issues/2355. It doesn't implement all the features from prom's Tree view. This is a first step and I aim to complete it with the missing parts with follow-up PRs. Current status:
> &nbsp;&nbsp;✔️ Query break-down in a Tree view
> &nbsp;&nbsp;✔️ Variables interpolation
> &nbsp;&nbsp;❌ Selectable tree nodes, that update the current query fired
> &nbsp;&nbsp;❌ Sub-queries aside each node, with details about each label/dimension available & eventual error state
> &nbsp;&nbsp;❌ "Explain" tab

Besides, as a positive side-effect this feature partially solves https://github.com/perses/perses/issues/1419 - for promQL queries only of course:

image

# Checklist

## UI Changes

  include:


ui/prometheus-plugin/src/components/PromQL.tsx
```diff
@@ -14,22 +14,62 @@
...
   const isDarkMode = theme.palette.mode === 'dark';
+  const [isTreeViewVisible, setTreeViewVisible] = useState(false);
 
   const promQLExtension = useMemo(() => {
     return new PromQLExtension().activateLinter(false).setComplete(completeConfig).asExtension();
   }, [completeConfig]);
 
+  const queryExpr = useReplaceVariablesInString(rest.value);
+
+  // Specific errors that user should be able to hide
```
celian-garcia, line 43: I'm not fan of that part, it seems you automatically hide the tree view in case of some errors happen, checking the error message but error messages is not something on which we could rely. Possibly did you checked if you couldn't rely on something more fixed like status code or something? Also I believe that some proxy could send different things. Let's say prometheus is behind a proxy that want to secure changing the 404 to a 403, with a message different than this one, then you're screwed.

celian-garcia: What if we just keep the button, tree view not visible by default. And if it is clicked, wrap any error message like that:

```
Looks like the tree view failed with following message:
---
<any error message as it comes>
---
if it is a Not Found or Permission Denied message, it might be because you're not targetting a Prometheus 3.0 datasource.
```

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -81,7 +81,8 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
       </FormControl>
       <PromQLEditor
         completeConfig={{ remote: { url: promURL } }}
-        value={query}
+        value={value.query}
```
celian-garcia, line 84: This change sounds to counter a mechanism of ``useQueryState``? can you explain? At a glance it seems this trick can introduce confusion.

celian-garcia (changes requested):
Cool to see it coming in Perses! I reviewed what I knew. Didn't check the new code as I believe it's moreorless copy pasting from promlens / prometheus? 
Regarding that, did you copy pasted from Prometheus? I remember Julius was saying it'd be better as it's more up-to-date and cleaner. @juliusv if by any chance you see this message do you confirm?

AntoineThebaud:
> Cool to see it coming in Perses! I reviewed what I knew. Didn't check the new code as I believe it's moreorless copy pasting from promlens / prometheus? Regarding that, did you copy pasted from Prometheus? I remember Julius was saying it'd be better as it's more up-to-date and cleaner. @juliusv if by any chance you see this message do you confirm?

Yes I've copy pasted from Prometheus not Promlens, I indicated it in the PR desc and also pointed to the precise file+version in the files concerned, e.g:
https://github.com/perses/perses/blob/a899dc753e84a32097baeb2d3f3e16c9c8038d12/ui/prometheus-plugin/src/components/promql/ast.ts#L14

;)

ui/prometheus-plugin/src/components/PromQL.tsx
AntoineThebaud: Yeah I know this is not ideal and actually I didn' t manage to base the behavior on error status code instead, I dont have access to such attribute it seems...

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
AntoineThebaud: I want the tree view to be (re)computed when the input loses focus, and not after each change, to avoid spamming API calls - just like we do for the queries to retrieve the data. Hence the trick here, `value.query` gets updated only after `handleQueryBlur` from `useQueryState` is called, while the `query` object provided by `useQueryState` gets updated after each call to  `handleQueryChange`.

ui/prometheus-plugin/src/components/PromQL.tsx
AntoineThebaud: Tackled offline, agreed to let it like that for now & simplify the logic the day we have a real PR dedicated to https://github.com/perses/perses/issues/1419. Added comment in the code to add this context.

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
AntoineThebaud: Added comment for clarification.

celian-garcia (approved):
Lgtm now, this is clearer with the comment thanks 😊 I'm still not fan of the error handling but if it's removed when we tackle the "1 error by query" issue then ok. I hope it won't stay too long like this ^^'

ui/prometheus-plugin/src/components/PromQLEditor.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const [isTreeViewVisible, setTreeViewVisible] = useState(false);
+
+  const promQLExtension = useMemo(() => {
+    return new PromQLExtension().activateLinter(false).setComplete(completeConfig).asExtension();
+  }, [completeConfig]);
+
+  const queryExpr = useReplaceVariablesInString(rest.value);
+
+  const { data: parseQueryResponse, isLoading, error } = useParseQuery(queryExpr ?? '', datasource);
+  const { errorMessage, isGenericError } = processError(error);
```
Gladorme, line 72:
```suggestion
  const { errorMessage, isGenericError } = useMemo(() => processError(error), [error])
```

ui/prometheus-plugin/src/components/PromQLEditor.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  const treeViewText = 'Tree View';
```
Gladorme, line 74: Can be moved outside of the function

Gladorme (approved):
Awesome! ❤️

PR #2351, opened 2024-10-17
[BUGFIX] TracingGanttChart: calculate start and end time from all spans

andreasgerstmayr:
# Description

In some cases, the trace start and end time does not match the start and end time of the root span (in other words, some spans are in progress or started after the root span has ended).

* compute and memoize start and end time of the trace by scanning all spans of the trace
* introduce a GanttTrace data type which holds these pre-computed fields
* update all React props `rootSpan` --> `trace` and all references `rootSpan.startTimeUnixMs` --> `trace.startTimeUnixMs` (and `.endTime`)

# Screenshots

The trace contains spans starting after the root span has ended:
image

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/panels-plugin/src/test/get-trace-data-jaeger.ts
+++ b/ui/panels-plugin/src/test/get-trace-data-jaeger.ts
@@ -1,0 +1,147 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Span, Trace, TraceAttribute, TraceAttributeValue, TraceResource } from '@perses-dev/core';
+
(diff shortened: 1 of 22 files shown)


ui/panels-plugin/src/test/get-trace-data-jaeger.ts
```diff
@@ -0,0 +1,147 @@
+// Copyright 2023 The Perses Authors
```
celian-garcia, line 1:
```suggestion
// Copyright 2024 The Perses Authors
```

celian-garcia (commented):
Just a nit. 
Apart from that I went through the files and didn' t see anything shocking. 

I was wondering if it couldn' t be considered as a breaking change to modify the TracingGanttChart component interface but I guess one should use the TracingGanttCharPanel component right? On which the interface didn' t change.

And by the way that' s cool as I think we have some slight time issues with our spans and sometimes they are no completely align. I guess it will fix som display for us as well.

andreasgerstmayr:
> I was wondering if it couldn' t be considered as a breaking change to modify the TracingGanttChart component interface but I guess one should use the TracingGanttCharPanel component right? On which the interface didn' t change.

Yes, `TracingGanttChart.PanelComponent` should be used.
In the `index.ts` file we only export types from `TracingGanttChart.ts`, so I'd consider everything which is not exported as an internal implementation detail.

> And by the way that' s cool as I think we have some slight time issues with our spans and sometimes they are no completely aligned. I guess it will fix som display for us as well.

:+1: let me know if you find other issues :)

celian-garcia (approved):
This freaking DCO on suggestion.. Sorry you'll have to fix it manually I believe... 
Otherwise LGTM

PR #2354, opened 2024-10-18
[BUGFIX] Redirect to the desired page after login success

### LINKED ISSUE #2284: After sign-in, Perses should redirect to the initial URL used
### Proposal

When you are sharing a link like to the explorer and you need to sign-in, the initial URL is lost. Once signed, it only redirects to the home page and not to the initial URL shared.

For example : https://demo.perses.dev/explore?start=1h&refresh=0s&explorer=metrics&data=%7B%22tab%22%3A%22finder%22%2C%22queries%22%3A%5B%5D%7D 

1. click on the link
2. You are redirected to the sign-in page
3. You are signing
4. Then you are redirected to the home page and not the explorer

celian-garcia:
# Description

This PR aims to fix #2284

This is two main changes:

## Redirection query params (UI + API change)
Make sure that the original route is never lost, passing it everywhere. To the Perses backend and to the OIDC/Oauth providers.

## Refactor how the UI refresh the access token presence (UI only change)
The main idea is to simplify relying on cookie presence only to know if we need to redirect or not. 
=> All routes/redirections are in the Router 💪 
From what I understood, the standard way appears to be creating a guard component in the Router which will react on the cookie presence/absence. And to place all the routes component under that guard.

### Problem 1
The main tricky point resided then in the fact that **when we refresh the access token with the refresh token, it was removing the cookie and recreated it**. The guard was then too reactive and redirected directly to signin page. 😢 
=> Solution to that is to use a debounce mechanism that you will see in the ``useIsAccessTokenExist`` function, completely copy pasted from there https://dev.to/vikas2426/debouncing-in-react-a-custom-hook-example-3i4n
Thanks to this debounce mechanism, the user is not redirected to sign in page when the refresh mechanism happens, because the guard will wait for 500ms before react on a cookie removal.

### Problem 2
Second tricky point: when a query get a 401 and try refresh a token, it's very likely that a lot of other queries are also doing the same. They can then enter in conflicts with each other. We need a solution to prevent this from happening.
=> It seems that it's sufficient to additionally refresh the token at the time it disappear, in the auth guard (debounce will make sure it's waiting the token update)

# Screenshots

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/app/src/model/auth-client.ts
+++ b/ui/app/src/model/auth-client.ts
@@ -15,11 +15,15 @@ import { useMutation, UseMutationResult, useQuery, useQueryClient, UseQueryResul
 import { fetch, fetchJson } from '@perses-dev/core';
 import { useCookies } from 'react-cookie';
 import { decodeToken } from 'react-jwt';
+import { useQueryParam } from 'use-query-params';
+import { useEffect, useState } from 'react';
 import buildURL from './url-builder';
 import { HTTPHeader, HTTPMethodPOST } from './http';
 
 const authResource = 'auth';
 const jwtPayload = 'jwtPayload';
+const redirectQueryParam = 'rd';
+const cookieRefreshTime = 500;
 
 export interface NativeAuthBody {
   login: string;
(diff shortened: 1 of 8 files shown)


ui/app/src/model/auth-client.ts
```diff
@@ -31,6 +32,14 @@ export function useIsAccessTokenExist() {
   return cookies[jwtPayload] !== undefined;
 }
 
+export function useRedirectQueryParam() {
+  const [redirectURI] = useQueryParam<string>('rd');
```
celian-garcia, line 36: change to path to avoid confusion

ui/app/src/views/auth/SignInView.tsx
```diff
@@ -12,30 +12,30 @@
...
 
 function SignInView() {
   const isSignUpDisable = useIsSignUpDisable();
-  const isAccessTokenExist = useIsAccessTokenExist();
   const authMutation = useNativeAuthMutation();
   const navigate = useNavigate();
   const { successSnackbar, exceptionSnackbar } = useSnackbar();
   const [login, setLogin] = useState<string>('');
   const [password, setPassword] = useState<string>('');
+  const redirect = useRedirectQueryParam();
```
celian-garcia, line 30: change to path to avoid confusion

Gladorme (approved):
Cool !

Issue #2345, opened 2024-10-15
Panic with percli offline migrate


sunng87:
### What did you do?

I'm trying to migrate the grafana NodeExporterFull dashboard to Perses using percli.

### What did you see instead? Under which circumstances?

I'm getting this panic error

```
./bin/percli migrate -f ~/tmp/node-exporter-full.json 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xc3129b]

goroutine 1 [running]:
github.com/perses/perses/internal/cli/cmd/migrate.(*option).offlineExecution(0xc00032e8f0, {0xc000706000?, 0xcf2d80?, 0xc000300150?})
        /home/nsun/workspace/perses/internal/cli/cmd/migrate/migrate.go:132 +0x5b
github.com/perses/perses/internal/cli/cmd/migrate.(*option).Execute(0xc00032e8f0)
        /home/nsun/workspace/perses/internal/cli/cmd/migrate/migrate.go:115 +0x10a
github.com/perses/perses/internal/cli/cmd.Run({0xf5d4a0, 0xc00032e8f0}, 0xc000334608, {0xc000308440, 0x0, 0x2})
        /home/nsun/workspace/perses/internal/cli/cmd/cmd.go:42 +0xdd
github.com/perses/perses/internal/cli/cmd/migrate.NewCMD.func1(0xc00032d300?, {0xc000308440?, 0x4?, 0xdd8ed7?})
        /home/nsun/workspace/perses/internal/cli/cmd/migrate/migrate.go:153 +0x32
github.com/spf13/cobra.(*Command).execute(0xc000334608, {0xc000308420, 0x2, 0x2})
        /home/nsun/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:985 +0xaca
github.com/spf13/cobra.(*Command).ExecuteC(0xc000328008)
        /home/nsun/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        /home/nsun/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
main.main()
        /home/nsun/workspace/perses/cmd/percli/main.go:97 +0x105

```

The json file is downloaded from Grafana website:
[1860_rev37.json](https://github.com/user-attachments/files/17385175/1860_rev37.json)

[...]

PR #2363, opened 2024-10-22
[BUGFIX] Fix migrate cmd by making the flag schemas.<schema_kind> mandatory when being offline

Nexucis:
fix https://github.com/perses/perses/issues/2345

When using the offline migrate command, you need to provide the schemas to be able to perform it. The nil pointer sharing in the issue is occurring because the flags are not provided and then the schemas are not loaded locally.

### DIFF
--- a/internal/cli/cmd/migrate/migrate.go
+++ b/internal/cli/cmd/migrate/migrate.go
@@ -58,7 +58,7 @@ func (o *option) Complete(args []string) error {
 		return outputErr
 	}
 	o.completeInput()
-	if (len(o.chartsSchemas) > 0 && len(o.queriesSchemas) > 0) || len(o.variablesSchemas) > 0 {
+	if len(o.chartsSchemas) > 0 && len(o.queriesSchemas) > 0 && len(o.variablesSchemas) > 0 {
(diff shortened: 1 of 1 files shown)


internal/cli/cmd/migrate/migrate.go
```diff
@@ -162,7 +165,7 @@ percli migrate -f ./dashboard.json --input=DS_PROMETHEUS=PrometheusDemo --online
 	cmd.Flags().StringVar(&o.variablesSchemas, "schemas.variables", "", "Path to the CUE schemas for the dashboard variables")
 	cmd.Flags().BoolVar(&o.online, "online", false, "When enable, it can request the API to use it to perform the migration")
 
-	// when online flag is used, the CLI will call the endpoint /migrate that will then use the schema from the server.
+	// When an online flag is used, the CLI will call the endpoint /migrate that will then use the schema from the server.
```
AntoineThebaud, line 168:
```suggestion
	// When the online flag is used, the CLI will call the endpoint /migrate that will then use the schema from the server.
```

AntoineThebaud (approved):

PR #2370, opened 2024-10-24
[ENHANCEMENT]: Custom http request headers for autocompletion of Promql editor

ZonaHex:
# Description

This PR is to allow adding http request headers for autocompletion of codemirror promql editor.

Starting from [codemirror-promql v0.49.0](https://www.npmjs.com/package/@prometheus-io/codemirror-promql/v/0.49.0), they have allowed adding specific HTTP headers along with the requests to Prometheus. Related PR is [here](https://github.com/prometheus/prometheus/pull/13118). This can be useful for those who use a custom prometheus url, require specific http headers, and hope to see the autocompletion of codemirror editor works.

I've tried to make some adjustments to implement this feature for the promql editor of our Prometheus-plugin. This dependency upgration may cause some problems and require a discussion.

Looking forward to getting some feedback :)

# Screenshots

# Checklist

## UI Changes

  include:


ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -41,9 +43,18 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
   const datasourceSelectLabelID = `prom-datasource-label-${selectedDatasource.name || 'default'}`;
 
   const { data: client } = useDatasourceClient<PrometheusClient>(selectedDatasource);
-  const promURL = client?.options.datasourceUrl;
   const { data: datasourceResource } = useDatasource(selectedDatasource);
 
+  const promURL =
+    (datasourceResource?.plugin.spec as PrometheusDatasourceSpec).directUrl ?? client?.options.datasourceUrl;
+  let completeConfig: CompleteConfiguration;
+  if (datasourceResource && (datasourceResource.plugin.spec.proxy as HTTPProxy)?.spec.headers) {
```
Nexucis, line 51: when the HTTP proxy is defined in the datasource, then it's the Perses backend server that is modifying the request before forwarding it to Prometheus.

So actually it shouldn't be necessary to do that. Injection of header should be done through the Proxy for security matter

ZonaHex: Hi, thank u for the review! This PR actually comes from a production problem in our web. We embedded perses panels in our workbench with datasource using directUrl and proxy of our http headers as a plugin spec. The proxy works fine with prometheus api for query_range, but didn't work with codemirror-promql's autocompletion (no http headers appear in the request). As I understand from this [line of code](https://github.com/perses/perses/pull/2370/files#diff-88f464b7edf2c44c2db875af4196defd6fe5fa870981627e195d6d35dfb2cd6eL83), the http request initiating from codemirror only have url specified from config, and didn't support the use of headers before the release of [codemirror-promql v0.49.0](https://www.npmjs.com/package/@prometheus-io/codemirror-promql/v/0.49.0).
Perhaps there are also other problems with the using of embedded panels and configuration of datasource using directUrl and proxy. Besides this PR, right now I can't figure out why the proxy only partially works.

ui/prometheus-plugin/package.json
```diff
@@ -34,7 +34,7 @@
     "@perses-dev/components": "0.48.0",
     "@perses-dev/core": "0.48.0",
     "@perses-dev/plugin-system": "0.48.0",
-    "@prometheus-io/codemirror-promql": "^0.43.0",
+    "@prometheus-io/codemirror-promql": "^0.49.0",
```
Nexucis, line 37:
```suggestion
    "@prometheus-io/codemirror-promql": "^0.55.0",
```

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
Nexucis: ah yeah ok I missed the part it was for the embedded part.
sounds good to me then. Thanks for the details and explanation :)

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -41,9 +43,18 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
   const datasourceSelectLabelID = `prom-datasource-label-${selectedDatasource.name || 'default'}`;
 
   const { data: client } = useDatasourceClient<PrometheusClient>(selectedDatasource);
-  const promURL = client?.options.datasourceUrl;
   const { data: datasourceResource } = useDatasource(selectedDatasource);
 
+  const promURL =
+    (datasourceResource?.plugin.spec as PrometheusDatasourceSpec).directUrl ?? client?.options.datasourceUrl;
```
Nexucis, line 49: I am wondering if `client?.options.datasourceUrl` would be enough. I am pretty sure it's the case, but I can't find where it is set.

I have this feeling because I know based on the proxy configuration we are giving to the client, what URL it should use so `directUrl` or `plugin.spec.proxy.spec.url`

Nexucis:
@ZonaHex  can you fix the conflict. I think that is the reason the CI doesn't start

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
Nexucis: @Gladorme or @celian-garcia `client?.options.datasourceUrl;` should be enough to get the accurate URL right ? Otherwise it wouldn't work today whether the proxy is used or not

ZonaHex: Hi, I just finished testing it. I think `client?.options.datasourceUrl;` should be enough and I've made the commit.

Nexucis:
cool, thank you for the changes. Hopefully the e2e tests should pass now.

I still have a weird feeling about passing the proxy header config like that. Because well it's part of the proxy config, so the frontend shouldn't use it.
Maybe later we should allow using headers when using `directURL`. That would be cleaner from my point of view even for the embedded use case.

If in the embedded use case you are using the proxy config, then you should have a proxy somewhere implemented, but the frontend component shouldn't use it.

Nexucis:
aaand the e2e tests didn't pass :( with a weird error. @Gladorme can you advise why there is this error please ?

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -41,9 +43,18 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
   const datasourceSelectLabelID = `prom-datasource-label-${selectedDatasource.name || 'default'}`;
 
   const { data: client } = useDatasourceClient<PrometheusClient>(selectedDatasource);
-  const promURL = client?.options.datasourceUrl;
   const { data: datasourceResource } = useDatasource(selectedDatasource);
 
+  const promURL = client?.options.datasourceUrl;
+  let completeConfig: CompleteConfiguration;
+  if (datasourceResource && (datasourceResource.plugin.spec.proxy as HTTPProxy)?.spec.headers) {
```
galangel, line 50: reuse

```suggestion
const promHeaders: TYPE = (datasourceResource?.plugin.spec.proxy as HTTPProxy)?.spec.headers;
  if (datasourceResource && promHeaders) {
```

ui/prometheus-plugin/src/plugins/prometheus-time-series-query/PrometheusTimeSeriesQueryEditor.tsx
```diff
@@ -41,9 +43,18 @@ export function PrometheusTimeSeriesQueryEditor(props: PrometheusTimeSeriesQuery
   const datasourceSelectLabelID = `prom-datasource-label-${selectedDatasource.name || 'default'}`;
 
   const { data: client } = useDatasourceClient<PrometheusClient>(selectedDatasource);
-  const promURL = client?.options.datasourceUrl;
   const { data: datasourceResource } = useDatasource(selectedDatasource);
 
+  const promURL = client?.options.datasourceUrl;
```
galangel, line 48: It seems that the client and the following logic only serve the config/URL 
I would extract all related lines to improve readability and perhaps test if necessary.

jgbernalp:
@ZonaHex plugins have now its own repo, can you move this PR to https://github.com/perses/plugins?

ZonaHex:
> @ZonaHex plugins have now its own repo, can you move this PR to https://github.com/perses/plugins?

Okay, I will take a look at this. Recently I haven't got time to follow up on repo updates.

ZonaHex:
This PR is no longer needed with the latest Perses version.

PR #2383, opened 2024-10-29
[ENHANCEMENT] migrate current.value to defaultValue

7840vz:
# Description
Migrates current value to defaultValue in dashboard variables. This specifically useful for special values like `$__all` which can be set by default in Grafana.

# Checklist

### DIFF
--- a/internal/api/migrate/mapping.cuepart
+++ b/internal/api/migrate/mapping.cuepart
@@ -165,7 +165,7 @@ spec: {
                 if grafanaVar.sort != _|_ if #mapping.sort[grafanaVar.sort] != _|_ {
                     sort: #mapping.sort[grafanaVar.sort]
                 }
-                if grafanaVar.current != _|_ if grafanaVar.current.value != _|_ {
+                if (*(grafanaVar.current.value) | _|_ ) != _|_ {
                     defaultValue: grafanaVar.current.value
                 }
                 if (*(grafanaVar.current.value) | _|_ ) == _|_ if *(grafanaVar.includeAll) | false  {
--- a/internal/api/migrate/mapping.cuepart
+++ b/internal/api/migrate/mapping.cuepart
@@ -165,7 +165,9 @@ spec: {
                 if grafanaVar.sort != _|_ if #mapping.sort[grafanaVar.sort] != _|_ {
                     sort: #mapping.sort[grafanaVar.sort]
                 }
-                // defaultValue: nothing to map to this field
+                if grafanaVar.current != _|_ if grafanaVar.current.value != _|_ {
+                    defaultValue: grafanaVar.current.value
+                }
                 #var: grafanaVar
                 plugin: [ // switch
                     %(conditional_variables)
--- a/internal/api/migrate/mapping.cuepart
+++ b/internal/api/migrate/mapping.cuepart
@@ -168,6 +168,9 @@ spec: {
                 if grafanaVar.current != _|_ if grafanaVar.current.value != _|_ {
                     defaultValue: grafanaVar.current.value
                 }
+                if (*(grafanaVar.current.value) | _|_ ) == _|_ if *(grafanaVar.includeAll) | false  {
+                    defaultValue: "$__all"
+                }
                 #var: grafanaVar
                 plugin: [ // switch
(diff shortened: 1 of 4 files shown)


Nexucis:
Thank you for the enhancement! Do you mind to target the branch release/v0.49 @7840vz ? Like that we can include it in the next rc for the v0.49.0 :)

7840vz:
done:)

Nexucis:
would you mind to rebase your PR base on the new branch ? There are some commit that are not related to your PR here

7840vz:
oops, rebased :)

Nexucis:
perfect ! Thank you @7840vz :)

@AntoineThebaud can you confirm it's ok here as well ? Thank you

internal/api/migrate/mapping.cuepart
```diff
@@ -165,7 +165,12 @@ spec: {
                 if grafanaVar.sort != _|_ if #mapping.sort[grafanaVar.sort] != _|_ {
                     sort: #mapping.sort[grafanaVar.sort]
                 }
-                // defaultValue: nothing to map to this field
+                if grafanaVar.current != _|_ if grafanaVar.current.value != _|_ {
+                    defaultValue: grafanaVar.current.value
+                }
+                if (*(grafanaVar.current.value) | _|_ ) == _|_ if *(grafanaVar.includeAll) | false  {
```
AntoineThebaud, line 171: oh interesting I never thought of this possibility to check the existence of nested fields directly

AntoineThebaud: But if there is no current value defined on Grafana side I think we shouldnt set defaultValue then?

internal/api/migrate/mapping.cuepart
```diff
@@ -165,7 +165,12 @@ spec: {
                 if grafanaVar.sort != _|_ if #mapping.sort[grafanaVar.sort] != _|_ {
                     sort: #mapping.sort[grafanaVar.sort]
                 }
-                // defaultValue: nothing to map to this field
+                if grafanaVar.current != _|_ if grafanaVar.current.value != _|_ {
```
AntoineThebaud, line 168:
```suggestion
                if (*(grafanaVar.current.value) | _|_ ) != _|_ {
```
reusing your below trick. I guess that should work here too?

internal/api/migrate/mapping.cuepart
7840vz: Idea is to "mimic" Grafana default behaviour, if current is not set at all, 'All is used by default; when dashboard is loaded with includeAll: true. 

Without explicitly setting defaultValue: `$__all` in here,  Perses sets first value of the list as the defaultValue instead when dashboard is loaded. 

I think using `All` is better default when current is not set.

internal/api/migrate/mapping.cuepart
7840vz: I guess :)

internal/api/migrate/mapping.cuepart
AntoineThebaud: Okay so actually on our side we have some issues with this behavior of grafana, it doesn't scale well for variables that return a lot of values, on the initial load of the dashboard if no defaultValue is defined then the fact to take All by default requires a lot of time to compute & load the dashboard, and this for any user, even the ones only interested in visualizing a subset of objects.

If you want All to be the default you can always save the dashboard with this explicit default value set, and this is already covered by your previous conditional statement here.

So yeah, we're not really in favor of the proposed migration logic here. And if ever this should change in the future, it should be done in the Perses's code responsible of that behavior: https://github.com/perses/perses/blob/e2c3831331c62a01947791e1d576d2b5dfea6807/ui/dashboards/src/context/VariableProvider/hydrationUtils.ts#L32-L39

Nexucis:
ah I am so sorry I didn't want to close this PR. It seems it has been closed because the release branch has been deleted.
Can you reopen it to tackle (again) the main branch. Sorry for that

7840vz:
> ah I am so sorry I didn't want to close this PR. It seems it has been closed because the release branch has been deleted. Can you reopen it to tackle (again) the main branch. Sorry for that

np, reopened

PR #2388, opened 2024-10-31
[BREAKINGCHANGE] Refactor the way to manage authentication in the perses http client

Nexucis:
I am breaking (again) the RestConfig for the perses http client regarding the way the authentication is managed.

I believe it will simplify a bit the config. It also simplify the way we are managing the auth in the client. Now everything is handled inside the `http.Client` and not anymore partially in the object `Request`

### DIFF
--- a/docs/user-guides/configuration.md
+++ b/docs/user-guides/configuration.md
@@ -522,9 +522,8 @@ discovery_name: <string>
 ```yaml
 # URL of the HTTP server exposing the global datasource list to retrieve.
 url: <url>
-
-[ auth: <Auth specification> ]
-
+[ basic_auth: <Basic Auth specification> ]
+[ oauth: <Oauth specification> ]
 # The HTTP authorization credentials for the targets.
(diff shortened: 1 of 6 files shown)


celian-garcia (approved):
LGTM even if yet another breaking change 🙄

docs/user-guides/configuration.md
```diff
@@ -522,9 +522,8 @@ discovery_name: <string>
 ```yaml
 # URL of the HTTP server exposing the global datasource list to retrieve.
 url: <url>
-
-[ auth: <Auth specification> ]
-
+[ basic_auth: <Basic Auth specification> ]
+[ oauth: <Oauth specification> ]
 # The HTTP authorization credentials for the targets.
 # Basic Auth and authorization are mutually exclusive. Use one or the other not both at the same time.
```
celian-garcia, line 528: This comment seems to apply to the moved config fields ?

PR #2392, opened 2024-11-04
[FEATURE] Table panel: columns can be reordered

### LINKED ISSUE #2353: Table: Add parameter to set the order of table columns
### Proposal
There's no way to set order of columns now.
In Grafana we can set an order by transformation.
I guess `columnSettings` in migration and `columnConfig` in UI allows to add extra atribute for column's order definition.

FYI. 
In my current case i was trying to move `value` column to the last position in table. 
I resolve it such way: change this string 
https://github.com/perses/perses/blob/feda4f2de387d39c17908b4edebe3fff58d659ad/ui/panels-plugin/src/plugins/table/TablePanel.tsx#L122
on this:
`        return { timestamp: ts.values[0][0], ...ts.labels, value: ts.values[0][1] }; // TODO: support multiple values and timestamps`

Gladorme:
# Description

Allow user to reorder columns in the table panel. It's based on the index of the column settings. If a column doesn't have a column setting, it will be use default ordering afters columns with columns settings. Closes #2353

# Screenshots

image

Drag and drop support:
image
output

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/panels-plugin/src/plugins/table/ColumnsEditor/ColumnEditorContainer.tsx
+++ b/ui/panels-plugin/src/plugins/table/ColumnsEditor/ColumnEditorContainer.tsx
@@ -52,7 +52,13 @@ export function ColumnEditorContainer({
         gap={4}
       >
         <Stack direction="row" gap={1}>
-          <DragButton onMoveUp={onMoveUp} onMoveDown={onMoveDown} />
+          <DragButton
+            onMoveUp={onMoveUp}
+            onMoveDown={onMoveDown}
+            menuSx={{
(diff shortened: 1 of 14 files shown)


ui/panels-plugin/src/plugins/table/ColumnsEditor/ColumnEditorContainer.tsx
```diff
@@ -47,6 +52,14 @@ export function ColumnEditorContainer({
         gap={4}
       >
         <Stack direction="row" gap={1}>
+          <DragButton
```
AntoineThebaud, line 55: I'd rather see this button on the right side, to align with the drag&drop UX of panels

Issue #2346, opened 2024-10-16
Update project introduction (Coredash->CNCF)


AntoineThebaud:
### Proposal

https://github.com/perses/perses/blob/main/docs/prologue/introduction.md should mention the CNCF sandbox instead of the deprecated CoreDash initiative.

Nexucis:
We could remove / merge this page with the home page. WDYT?

PR #2394, opened 2024-11-05
remove perses intro

Nexucis:
removing the intro that is not relevant anymore and won't be needed regarding the last changes of the website.

fix https://github.com/perses/perses/issues/2346


Nexucis:
moving forward on this one so I can release the v0.49

PR #2405, opened 2024-11-07
[BUGFIX] Fix tree view broken when using prometheus built-in vars

AntoineThebaud:
# Description

With https://github.com/perses/perses/pull/2344 I missed the fact that prometheus-specific builtin variables (`$__interval`, `$__rate_interval`..) aren't getting replaced by proper values with the current logic.. Thus any query containing such variable is producing an error (prometheus cannot parse this perses-specific syntax). This PR aims to fix that.

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/prometheus-plugin/src/components/PromQLEditor.tsx
+++ b/ui/prometheus-plugin/src/components/PromQLEditor.tsx
@@ -21,7 +21,7 @@ import { ErrorAlert } from '@perses-dev/components';
 import CloseIcon from 'mdi-material-ui/Close';
 import { useReplaceVariablesInString } from '@perses-dev/plugin-system';
 import { PrometheusDatasourceSelector } from '../model';
-import { replacePromBuiltinVariables } from '../utils';
+import { replacePromBuiltinVariables } from '../plugins/prometheus-time-series-query/replace-prom-builtin-variables';
 import { useParseQuery } from './parse';
 import TreeNode from './TreeNode';
 
--- a/ui/prometheus-plugin/src/components/PromQLEditor.tsx
+++ b/ui/prometheus-plugin/src/components/PromQLEditor.tsx
@@ -21,6 +21,7 @@ import { ErrorAlert } from '@perses-dev/components';
 import CloseIcon from 'mdi-material-ui/Close';
 import { useReplaceVariablesInString } from '@perses-dev/plugin-system';
 import { PrometheusDatasourceSelector } from '../model';
+import { replacePromBuiltinVariables } from '../utils';
 import { useParseQuery } from './parse';
 import TreeNode from './TreeNode';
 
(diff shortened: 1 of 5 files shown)


ui/prometheus-plugin/src/components/PromQLEditor.tsx
```diff
@@ -72,7 +73,10 @@ export function PromQLEditor({ completeConfig, datasource, ...rest }: PromQLEdit
     return new PromQLExtension().activateLinter(false).setComplete(completeConfig).asExtension();
   }, [completeConfig]);
 
-  const queryExpr = useReplaceVariablesInString(rest.value);
+  let queryExpr = useReplaceVariablesInString(rest.value);
+  if (queryExpr) {
+    queryExpr = replacePromBuiltinVariables(queryExpr, 12345, 12345); // TODO placeholder values to replace
```
AntoineThebaud, line 78: Looks like providing proper values for the step parameters here wont be easy, currently I'd need to rely on the timeseries query context but actually these step values are computed independently / before the queries are getting fired, so maybe we should extract them to independant hooks to be reused here? Need to think about it...

ui/prometheus-plugin/src/utils/utils.ts
```diff
@@ -11,8 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { replaceVariable } from '@perses-dev/plugin-system';
```
AntoineThebaud: This import was causing an issue with Jest:

```
FAIL src/utils/utils.test.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'install')

      24 | import { getFormattedAxis } from '../utils';
      25 |
    > 26 | use([EChartsBarChart, GridComponent, DatasetComponent, TitleComponent, TooltipComponent, CanvasRenderer]);
         |    ^
      27 |
      28 | const BAR_WIN_WIDTH = 14;
      29 | const BAR_GAP = 6;

      at use (../node_modules/echarts/lib/extension.js:109:7)
      at ../node_modules/echarts/lib/extension.js:96:7
          at Array.forEach (<anonymous>)
      at each (../node_modules/echarts/node_modules/zrender/lib/core/util.js:205:13)
      at use (../node_modules/echarts/lib/extension.js:95:5)
      at Object.<anonymous> (../components/src/BarChart/BarChart.tsx:26:4)
      at Object.<anonymous> (../components/src/BarChart/index.ts:17:14)
      at Object.<anonymous> (../components/src/index.ts:18:14)
      at Object.<anonymous> (../plugin-system/src/components/CalculationSelector/CalculationSelector.tsx:24:21)
      at Object.<anonymous> (../plugin-system/src/components/CalculationSelector/index.ts:17:14)
      at Object.<anonymous> (../plugin-system/src/components/index.ts:17:14)
      at Object.<anonymous> (../plugin-system/src/index.ts:17:14)
      at Object.<anonymous> (src/utils/utils.ts:37:23)
      at Object.<anonymous> (src/utils/utils.test.ts:17:16)
```

This was due to the fact that Jest doesn't support tree shaking, about that see
- https://github.com/jestjs/jest/issues/11230
- https://dev.to/fogel/potential-issues-with-barrel-files-in-jest-1nkl

The location change here is just a way to work around the issue, it's actually surprising that we didnt face it earlier already..

Nexucis (approved):

PR #2410, opened 2024-11-08
[FEATURE] new CLI command for plugins development

AntoineThebaud:
# Description

This PR adds a new `percli plugin` command group and a first `percli plugin update` sub-command that can be used to (re)install the [common](https://github.com/perses/perses/tree/main/cue/schemas/common) CUE package as a dependency to the local CUE module.

The logic is very similar to `percli dac setup` (hence the big refactoring part) and same as for this one ultimately we should get rid of this command in favor of relying on the native dependency management of CUE.

Ref discussion https://github.com/perses/perses/discussions/2085#discussioncomment-11078190

# Checklist


Nexucis:
don't you want to start by the command `percli plugin new` ?

Nexucis (approved):

AntoineThebaud:
> don't you want to start by the command `percli plugin new` ?

update is kinda a subset of the future "new" command and is actually standalone, it (re)creates the deps so you are not obliged to have run "new" before this one

PR #2416, opened 2024-11-12
[BUGFIX] Fix login command avoiding reusing previous config

Nexucis:
Currently if you log using the CLI in the demo and then you are waiting for the token to expire, you won't be able to log again. Instead you will the following error:

```bash
$ ./bin/percli login http://demo.perses.dev/
Error: something wrong happened with the request to the API. Error: Get “http://demo.perses.dev/api/config”: oauth2: token expired and refresh token is not set
```

Also you cannot log into a Perses instance where the auth is disabled. You will face the same issue. 

All of that is because when using the command `percli login` we are creating the Perses HTTP client based on a previous config where potentially token is expired or doesn't exist.

Creating from scratch the Perses HTTP client is fixing the issues mentioned above.

### DIFF
--- a/internal/cli/cmd/login/login.go
+++ b/internal/cli/cmd/login/login.go
@@ -90,20 +90,22 @@ func (o *option) Complete(args []string) error {
 		return fmt.Errorf("no URL has been provided neither found in the previous configuration")
 	}
 
-	// create a new apiClient
-	o.restConfig = config.Global.RestClientConfig
-	o.restConfig.URL = o.url
-	if o.restConfig.TLSConfig == nil {
-		o.restConfig.TLSConfig = &secret.TLSConfig{}
+	// Create a new apiClient from scratch.
+	// We shouldn't use the previous context as for the moment we have a single config.
+	// So, switching from a Perses instance to another one without restarting from scratch the context/ the Perses client doesn't make sense.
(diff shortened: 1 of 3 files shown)


internal/cli/cmd/login/login.go
```diff
@@ -224,9 +227,9 @@ func (o *option) newLoginOption() (loginOption, error) {
 
 func (o *option) selectAndSetProvider() error {
 	providers := o.remoteConfig.Security.Authentication.Providers
-	// The first step is to collect the different providers and store it into items + modifiers.
-	// items will be the selection items to display to users.
-	// modifiers will be the action to save the different user input into option struct.
+	// The first step is to collect the different providers and store it into items and modifiers.
+	// Items will be the selection items to display to users.
```
celian-garcia, line 231: Options now apparently

celian-garcia (approved):
Just a nit in comments but approved !

internal/cli/cmd/login/login.go
Nexucis: I don't get it 😅

celian-garcia: The variable has been renamed from ``items`` to ``options``, but not the comments

PR #2426, opened 2024-11-21
test

Nexucis:
testing e2e issue by moving back from 3 commits ahead


Nexucis:
alright it's related to an upgrade of Prometheus 3.0 and we have to change the way to load the PromQL tree view. We shouldn't send a request directly to the endpoint `parse_query` but only if the user is clicking on the tree icon

PR #2427, opened 2024-11-21
[BUGFIX] PromQL Tree view: stop firing the parse query before the show button gets clicked

AntoineThebaud:
# Description

Change the PromQL tree view behavior to only fire the request to the parse_query endpoint if the user decided to show the Tree view.
Previous design was to prevent relevant syntax errors to be hidden by default to the user, but in the end it causes other issues / unwanted console errors so in the end 'getting rid of that.
This should also fix the issue we have with e2e tests currently.

# Checklist

## UI Changes

  include:


Nexucis (approved):
❤️

PR #2431, opened 2024-11-22
[BUGFIX] User with multiple providers shouldn't be created

### LINKED ISSUE #2429: Users with both nativeProvider and oauthProviders can be created.
### What did you do?

This issue has been reported by one of our user: ref https://github.com/perses/perses/issues/2428

Creating this kind of user passes 
```json
{  
  "kind": "User",  
  "metadata": {  
    "name": "alice"  
  },  
  "spec": {  
    "nativeProvider": {  
        "password": "password"  
    },  
    "oauthProviders": [  
      {  
        "issuer": "http://localhost/openid",  
        "email": "alice@example.com",  
        "subject": "123456789"  
      }  
    ]  
  }  
}  
```

This should not pass because we decided for now to not let a user create an account with different kind of providers. This is to limit the risks to impersonate.

Also creation of a user with oauthProviders should be done internally only, at first login time. This is because we have a not abvious internal structure that is not documented.

User has been created and we cannot login with it (on purpose)

### System information

_No response_

### Perses version

0.49.0

### Perses configuration file

_No response_

### Logs

_No response_

Nexucis:
fix #2429

### DIFF
--- a/cue/model/api/v1/user_go_gen.cue
+++ b/cue/model/api/v1/user_go_gen.cue
@@ -15,9 +15,9 @@ package v1
 }
 
 #UserSpec: {
-	firstName?:      string          @go(FirstName)
-	lastName?:       string          @go(LastName)
-	nativeProvider?: #NativeProvider @go(NativeProvider)
+	firstName?:      string                 @go(FirstName)
+	lastName?:       string                 @go(LastName)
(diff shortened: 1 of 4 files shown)


Nexucis:
probably need more thoughts here to move forward on this PR and to be able to merge it.

PR #2440, opened 2024-11-26
copy duration type from prometheus/common

Nexucis:
We are copying the duration type from prometheus/common to be able to add the kubebuilder annotation as it doesn't seem we will be able to add it in prometheus/common.

See https://github.com/prometheus/common/pull/708

We have to do it also because the annotation doesn't work when we have an array of duration.

See also https://github.com/perses/perses-operator/issues/20

@ibakshay @jgbernalp @nicolastakashi


Nexucis:
that should also help to fix the PR: https://github.com/perses/perses-operator/pull/21 and solve the issue https://github.com/perses/perses-operator/issues/20

pkg/model/api/v1/common/duration.go
```diff
@@ -0,0 +1,202 @@
...
+// This code has been copied entirely from prometheus/common.
+// We did it to be able to add the kubebuilder annotation
+
+// Duration wraps time.Duration. It is used to parse the custom duration format
+// from YAML.
+// This type should not propagate beyond the scope of input/output processing.
+// +kubebuilder:validation:Schemaless
+// +kubebuilder:validation:Type=string
+// +kubebuilder:validation:Format=duration
+type Duration time.Duration
```
jgbernalp, line 33: Now that we control the type we can add the proper validation as:

```suggestion
// +kubebuilder:validation:Schemaless
// +kubebuilder:validation:Type=string
// +kubebuilder:validation:Format=duration
// +kubebuilder:validation:Pattern=`^(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?$`
type Duration time.Duration
```

ibakshay: I don't think we require **// +kubebuilder:validation:Schemaless** annotation now :) 
```suggestion
// +kubebuilder:validation:Type=string
// +kubebuilder:validation:Format=duration
// +kubebuilder:validation:Pattern=`^(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?$`
type Duration time.Duration
```

Nexucis: all done @ibakshay @jgbernalp :)

jgbernalp (approved):

PR #2454, opened 2024-12-03
[BUGFIX] Fix joinByField migration

7840vz:
Fix joinByField migration in table. List is expected for `columns`.

### DIFF
--- a/cue/schemas/panels/table/migrate.cue
+++ b/cue/schemas/panels/table/migrate.cue
@@ -132,7 +132,7 @@ if (*#panel.type | null) == "table" {
 					if transformation.id == "joinByField" {
 						kind: "JoinByColumnValue"
 						spec: {
-							columns: *transformation.options.byField | []
+							columns: [*transformation.options.byField | _|_]
 							if transformation.disabled != _|_ {
 								disabled: transformation.disabled
 							}


cue/schemas/panels/table/migrate.cue
```diff
@@ -132,7 +132,7 @@ if (*#panel.type | null) == "table" {
 					if transformation.id == "joinByField" {
 						kind: "JoinByColumnValue"
 						spec: {
-							columns: *transformation.options.byField | []
+							columns: [*transformation.options.byField | _|_]
```
AntoineThebaud, line 135: Hmm weird that it doesnt complain with `explicit error (_|_ literal) in source`.
e.g https://cuelang.org/play/?id=sZfe0B6lj5f#w=function&i=cue&f=eval&o=cue

I still think we should avoid such assignment, but I'm not sure to understand the change here, can you explain what it fixes?

AntoineThebaud (changes requested):

cue/schemas/panels/table/migrate.cue
AntoineThebaud: Oh ok simply that my bad.. in that case I think we should go for
```suggestion
							columns: *[transformation.options.byField] | []
```

AntoineThebaud (changes requested):

PR #2461, opened 2024-12-04
[ENHANCEMENT] Add yellow mapping from grafana

7840vz:
Migratoin: Add missing color aliases from grafana pallete (yellow).
image


Gladorme (approved):
Thanks 😄

PR #2462, opened 2024-12-04
[FEATURE] Value mapping settings for Status History and Stat Chart

### LINKED ISSUE #2337: Common value mapping feature for panels
### Proposal

Based on cell settings of the table, create a common "value mappings" editor / spec, that panel can implements. Based on [@7840vz suggestion](https://github.com/perses/perses/pull/2298#issuecomment-2410941711)

abelyakin:
# Description
This PR adds a new value mapping feature, letting users define custom mappings for specific values. This makes data visualization more flexible and customizable by transforming raw values into more meaningful representations.

This functionality is designed to be reusable by any panel and is currently implemented for status history and stat chart components.

Resolves #2337

# Screenshots

Screenshot 2024-12-04 at 22 34 04
image

# Checklist

## UI Changes

  include:

### DIFF
--- a/cue/schemas/common/mappings.cue
+++ b/cue/schemas/common/mappings.cue
@@ -28,8 +28,8 @@ import (
 #rangeCondition: {
 	kind: "Range"
 	spec: {
-		from?: number
-		to?: number
+		from?:  number
+		to?:    number
 		result: #mappingResult
 	}
 }
@@ -38,21 +38,21 @@ import (
 	kind: "Regex"
 	spec: {
 		pattern: strings.MinRunes(1)
-		result: #mappingResult
+		result:  #mappingResult
 	}
 }
 
 #miscCondition: {
 	kind: "Misc"
 	spec: {
-		value: "empty" | "null" | "NaN" | "true" | "false"
+		value:  "empty" | "null" | "NaN" | "true" | "false"
 		result: #mappingResult
 	}
 }
 
 #mappingResult: {
-	value:            string
-  	color?:           =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
+	value:  string
+	color?: =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
 }
 
 #mappings: #valueCondition | #rangeCondition | #regexCondition | #miscCondition
--- a/cue/schemas/common/mappings.cue
+++ b/cue/schemas/common/mappings.cue
@@ -1,0 +1,58 @@
+// Copyright 2024 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"strings"
+)
+
+#valueCondition: {
+	kind: "Value"
+	spec: {
+		value:  strings.MinRunes(1)
+		result: #mappingResult
+	}
+}
+
+#rangeCondition: {
+	kind: "Range"
+	spec: {
+		from?: number
+		to?: number
+		result: #mappingResult
+	}
+}
+
+#regexCondition: {
+	kind: "Regex"
+	spec: {
+		pattern: strings.MinRunes(1)
+		result: #mappingResult
+	}
+}
+
+#miscCondition: {
+	kind: "Misc"
+	spec: {
+		value: "empty" | "null" | "NaN" | "true" | "false"
+		result: #mappingResult
+	}
+}
+
+#mappingResult: {
+	value:            string
+  	color?:           =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
+}
+
+#mappings: #valueCondition | #rangeCondition | #regexCondition | #miscCondition
--- a/ui/components/src/StatChart/StatChart.tsx
+++ b/ui/components/src/StatChart/StatChart.tsx
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { ReactElement, useMemo } from 'react';
-import { formatValue, FormatOptions } from '@perses-dev/core';
+import { useMemo } from 'react';
+import { FormatOptions } from '@perses-dev/core';
 import { Box, Typography, styled } from '@mui/material';
 import merge from 'lodash/merge';
 import { use, EChartsCoreOption } from 'echarts/core';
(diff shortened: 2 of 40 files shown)


ui/components/src/StatusHistoryChart/StatusHistoryChart.tsx
```diff
@@ -153,7 +147,7 @@ export function StatusHistoryChart(props: StatusHistoryChartProps) {
   };
 
   return (
-    <Box style={{ height: height }} sx={{ overflow: 'auto' }}>
+    <Box sx={{ height: height, overflow: 'auto' }}>
```
Gladorme, line 150: We do not but high dynamic value in sx props: https://mui.com/system/getting-started/the-sx-prop/#dynamic-values. This change need to be reverted

ui/components/src/StatChart/StatChart.tsx
```diff
@@ -33,7 +34,8 @@ const SERIES_NAME_FONT_WEIGHT = 400;
 const VALUE_FONT_WEIGHT = 700;
 
 export interface StatChartData {
-  calculatedValue?: number | null;
+  color: string;
+  calculatedValue?: string | number | null | undefined;
```
Gladorme, line 38:
```suggestion
  calculatedValue?: string | number | null;
```
or 
```suggestion
  calculatedValue: string | number | null | undefined;
```

ui/components/src/ValueMappingEditor/ValueMappingsEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import { useState } from 'react';
+import AddIcon from 'mdi-material-ui/Plus';
+import { ValueMapping } from '@perses-dev/core';
+import { ValueMappingEditor } from './ValueMappingEditor';
+
+export interface ValueMappingsEditorProps {
+  mappings: ValueMapping[];
+  onChange: (valueMappings: ValueMapping[]) => void;
+}
```
Gladorme, line 25: Would be cool if table cell settings could use this mapping editor, in order to not have code duplication. AFAIK only colors change (2 instead of 1).
I propose something like this:
```suggestion
  hasSecondaryColor?: bool
}
```

If it's enabled, we could use "Primary color" and "Secondary color", so it can be used everywhere 🤔 Or if it's to abstract to users and we could add an option to rename headers:

```suggestion
  hasSecondaryColor?: bool
  colorHeaders?: string[]
}
```

cue/schemas/common/mappings.cue
```diff
@@ -0,0 +1,58 @@
...
+	kind: "Misc"
+	spec: {
+		value:  "empty" | "null" | "NaN" | "true" | "false"
+		result: #mappingResult
+	}
+}
+
+#mappingResult: {
+	value:  string
+	color?: =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
```
Gladorme, line 55: In line with something more generic with my previous comments:
```suggestion
	color?: =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
	secondaryColor?: =~"^#(?:[0-9a-fA-F]{3}){1,2}$"
```

Gladorme (commented):
Cool!
Would be nice if table cell settings could leverage this mapping editor 😄

ui/panels-plugin/src/plugins/stat-chart/utils/data-transform.ts
```diff
@@ -12,48 +12,54 @@
...
-      } else {
-        // thresholds.steps should be in ascending order, so return if value is less than step.value
-        return;
-      }
-    });
+  // Check thresholds first (they take priority)
+  if (thresholds?.steps && typeof value === 'number') {
+    const matchingColors = thresholds.steps
+      .map((step, index) => {
+        if (value > step.value) {
```
lomobot, line 37: I guess, it'll be right to include threshold value to range, which is colored
```suggestion
        if (value >= step.value) {
```

ui/components/src/StatusHistoryChart/StatusHistoryChart.tsx
abelyakin: Thank you! Brought old code by mistake.

ui/components/src/ValueMappingEditor/ValueMappingsEditor.tsx
abelyakin: I agree, and it was my original plan, but since there are options on how to implement it, I would discuss it with others. 
I would move tables to value mapping as well since it is a successor of cell settings and overall functionality duplication. To keep the value mapping API consistent across all panels, I would take inspiration from Grafana, where they let you select one color but also let you choose where it is going to be applied. If it is text color, it applies as is, and if it is background, it calculates the text color for perfect contrast.

ui/panels-plugin/src/plugins/stat-chart/utils/data-transform.ts
abelyakin: fixed!

Gladorme (commented):
Using default Node Exporter dashboard, I think this mapping should be working:
image
And it made me think for stat chart, we will a conflict with thresholds + value mappings colors

abelyakin:
> I am not a huge fan of `React.FC` for component not using `children` prop, it can be misleading. (https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/). Is there a particular reason you use it?

Recently introduced eslint rule https://github.com/perses/perses/blob/f5ad1265b5ef73559441aa82ff1ad091bff98b8d/ui/.eslintrc.base.js#L55 forces us to add return types to functions (and React components), which looks a bit verbose in my opinion. Since it's going to be JSX.Element in most cases for React components, this is just a matter of internal project convention.

Starting React 18 previous issues are fixed with implicit children typing. It provides cleaner code while maintaining type safety, satisfies our eslint rules without verbose JSX.Element declarations, and creates a consistent typing pattern across the codebase. The difference in readability is clear:

Before: `const Component = ({ title }: Props): JSX.Element => <div>{title}</div>;`

After: `const Component: FC<Props> = ({ title }) => <div>{title}</div>;`

Personally, I'm more familiar with and lean towards using React.FC type, as it provides the same type safety with cleaner syntax while still satisfying the eslint rule. However, I'm open to discussion on this convention.

Gladorme:
Fine for FC then, I didn't know it has been fixed in React 18 👍

Gladorme (approved):
LGTM, thanks! 😄

Nexucis:
@jgbernalp in case you didn't see this PR that changes various thing in the different panels / plugins

AntoineThebaud:
Looking back at this, for StatChart if I'm not wrong "Thresholds" & "Value mappings" are overlapping features for the colors (since defining mappings for a precise value or a range can be easily defined similarly on thresholds side), and would be 100% overlapping if we had a display text option on the Thresholds side (which could be considered).

Having multiple ways to achieve the same customization is the kind of situation that we wanted to avoid as much as possible with Perses compared to Grafana, because it adds visual clutter & cognitive load for the end user (e.g when using both customizations, which one takes precedence?)

IMHO "thresholds" is more intuitive/a self-explanatory term that makes more sense for the StatChart, so I'd propose to get rid of Value Mappings here & add a display text customization to the thresholds. WDYT? ping @abelyakin @Nexucis @Gladorme

PR #2463, opened 2024-12-05
[ENHANCEMENT] Add sparkline migration

7840vz:
Adds sparkline migration to StatChart.
image

### DIFF
--- a/cue/schemas/panels/stat/migrate.cue
+++ b/cue/schemas/panels/stat/migrate.cue
@@ -1,4 +1,4 @@
-if (*#panel.type | null) == "stat" {
+ if (*#panel.type | null) == "stat" {
 	kind: "StatChart"
 	spec: {
 		calculation: *#mapping.calc[#panel.options.reduceOptions.calcs[0]] | #defaultCalc // only consider [0] here as Perses's GaugeChart doesn't support individual calcs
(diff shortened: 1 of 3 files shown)


cue/schemas/panels/stat/migrate.cue
```diff
@@ -1,4 +1,4 @@
-if (*#panel.type | null) == "stat" {
+ if (*#panel.type | null) == "stat" {
```
AntoineThebaud, line 1:
```suggestion
if (*#panel.type | null) == "stat" {
```

cue/schemas/panels/stat/migrate.cue
```diff
@@ -27,6 +27,9 @@ if (*#panel.type | null) == "stat" {
...
 		}
 
-		// nothing to map to the `sparkline` field yet
+		#sparkline: *#panel.options.graphMode | "none"
+		if #sparkline == "area" {
+			sparkline: {}
+		}
 	}
-},
+}
```
AntoineThebaud, line 35:
```suggestion
},
```

I know it's kinda counter-intuitive but the coma is actually mandatory here 😅 https://github.com/perses/perses/blob/main/docs/plugins/cue.md#panel-1

AntoineThebaud (changes requested):

AntoineThebaud (approved):

PR #2465, opened 2024-12-05
[ENHANCEMENT] Update Markdown migration

7840vz:
It is possible to get Grafana Text panel without content field, if text panel is generated like via grafonnet library.
This small change covers it and provides empty panel in those cases.


AntoineThebaud (approved):
👍

PR #2467, opened 2024-12-05
[IGNORE] `dac build`: remove useless check for output flag

AntoineThebaud:
# Description

The removed check was redundant with the base validation behavior coming with that flag:
https://github.com/perses/perses/blob/1f0d7cc7b2ef419592df52d1bbbdcabc1e4461cc/internal/cli/output/output.go#L33-L43

(Some commands (e.g `describe`) were already fine)

# Checklist

### DIFF
--- a/internal/cli/cmd/get/get.go
+++ b/internal/cli/cmd/get/get.go
@@ -58,6 +58,8 @@ func (o *option) Complete(args []string) error {
 	}
 
 	// Complete the output only if it has been set by the user
+	// NB: In the case of the `get` command, the default output format is/should be a table, not json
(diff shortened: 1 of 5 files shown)


internal/cli/cmd/get/get.go
```diff
@@ -57,12 +57,10 @@ func (o *option) Complete(args []string) error {
 		return err
 	}
 
-	// Complete the output only if it has been set by the user
-	if len(o.Output) > 0 {
```
Nexucis, line 61: you shouldn't do that as the default format for the `get` command is a table (that's the only excepting existing in the CLI about the format supported)

Nexucis (approved):

PR #2471, opened 2024-12-06
[FEATURE] add plugin build command

Nexucis:
Adding a new command `percli plugin build` that is running `npm run build` and then is creating the archive with the expecting tree folder.


internal/cli/cmd/plugin/build/build.go
```diff
@@ -0,0 +1,143 @@
...
+	frontendFolder string
+	schemaFolder   string
+	distFolder     string
+	writer         io.Writer
+	errWriter      io.Writer
+}
+
+func (o *option) Complete(args []string) error {
+	if len(args) > 0 {
+		return fmt.Errorf("no args are supported by the command 'update'")
```
AntoineThebaud, line 45:
```suggestion
		return fmt.Errorf("no args are supported by the command 'build'")
```

internal/cli/cmd/plugin/build/build.go
```diff
@@ -0,0 +1,143 @@
...
+	// Add the dist content at the root of the archive (saying differently, dist folder should not appear in the archive)
+	distFiles, err := os.ReadDir(o.distFolder)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read 'dist' directory: %w", err)
+	}
+	for _, f := range distFiles {
+		list[path.Join(o.distFolder, f.Name())] = f.Name()
+	}
+
+	// Do the same for the Cuelang schemas
```
AntoineThebaud, line 106: We also need to embed the `cue.mod` folder here.

And (maybe for a followup PR?) I guess we should run `percli plugin update` before that in order to ensure that `cue.mod/pkg` is populated with the common deps.

Nexucis: I was thinking that as well and then I'm thinking that perhaps in some case, you don't necessary need to get the cue dependencies. For example the markdown panel doesn't depend on anything and as such doesn't require to run `percli plugin update`

AntoineThebaud (approved):
Noice

PR #2481, opened 2024-12-10
[ENHANCEMENT] Improve the migration

Nexucis:
The main idea is to offload as much as possible of the migration logic to Golang, and delegate to Cuelang only when necessary (= at plugin model level), in order to solve #2039.

That's a big change so let's keep an eye on that, but tests are passing & memory usage + computing time is much much less than it was.


AntoineThebaud:
https://github.com/perses/perses/blob/main/scripts/check_license.sh#L6 "migrate.cue" can be removed there

cue/schemas/panels/bar/migrate.cue
```diff
@@ -1,16 +1,36 @@
...
-		#decimal: *#panel.fieldConfig.defaults.decimal | null
-		if #decimal != null {
-			format: decimalPlaces: #decimal
-		}
+import (
+	commonMigrate "github.com/perses/perses/cue/schemas/common/migrate"
+)
+
+#grafanaType: "bargauge"
+#panel:       _
```
AntoineThebaud: technically would be more accurate to put 
```suggestion
#panel:       {...}
```

internal/api/impl/migrate/migrate.go
```diff
@@ -48,8 +49,12 @@ func (e *endpoint) Migrate(ctx echo.Context) error {
 	if err := ctx.Bind(body); err != nil {
 		return apiinterface.HandleBadRequestError(err.Error())
 	}
-	grafanaDashboard := migrate.ReplaceInputValue(body.Input, string(body.GrafanaDashboard))
-	persesDashboard, err := e.migrationService.Migrate([]byte(grafanaDashboard))
+	rowGrafanaDashboard := []byte(migrate.ReplaceInputValue(body.Input, string(body.GrafanaDashboard)))
```
AntoineThebaud, line 52:
```suggestion
	rawGrafanaDashboard := []byte(migrate.ReplaceInputValue(body.Input, string(body.GrafanaDashboard)))
```

😉

Nexucis:
PR is now ready. After running this PR on the different dashboards community provides that was showing the memory issue, I am not able to reproduce the issue.

Let's see if on a long running memory is kept at a low level.

internal/api/migrate/grafana.go
```diff
@@ -0,0 +1,245 @@
...
+	type plain SimplifiedDashboard
+	if err := json.Unmarshal(data, (*plain)(tmp)); err != nil {
+		return err
+	}
+	tmp.rearrangeGrafanaPanelsWithinExpandedRows()
+	*d = *tmp
+	return nil
+}
+
+func (d *SimplifiedDashboard) rearrangeGrafanaPanelsWithinExpandedRows() {
```
AntoineThebaud, line 188: let's keep the intro comment too?

```suggestion
/**
 * This function addresses an issue we have with Grafana datamodel when it comes to migrating dashboards to Perses: When
 * a row is expanded in Grafana, its children panels are moved up in the main panels list, thus become siblings of the row.
 * When it comes to Perses migration we need to recompose the parent->children relationships. However in its current state
 * the CUE language doesn't permit us to achieve this recomposition, hence this processing in the backend code.
 *
 * So what this function does is basically the following: whenever such pattern is encountered in the panels list:
 * ...
 * row1,
 * panelA,
 * panelB,
 * panelC,
 * row2,
 * ...
 * the objects get rearranged like:
 * ...
 * row1: {
 *   panelA,
 *   panelB,
 *   panelC,
 * },
 * row2,
 * ...
 */
func (d *SimplifiedDashboard) rearrangeGrafanaPanelsWithinExpandedRows() {
```

internal/api/migrate/loader.go
```diff
@@ -14,112 +14,89 @@
...
 	for _, file := range files {
 		if !file.IsDir() {
 			logrus.Tracef("file %s is ignored since we are looking for directories", file.Name())
 			continue
 		}
-		migFilePath := filepath.Join(c.schemasPath, file.Name(), "migrate.cue")
-		contentStr, readErr := os.ReadFile(migFilePath)
-		if readErr != nil {
-			logrus.WithError(readErr).Debugf("No migration file found at %s, plugin %s will be skipped", migFilePath, file.Name())
+		schemaPath := filepath.Join(panelSchemaPath, file.Name(), "migrate")
```
AntoineThebaud, line 69: to be adjusted with the new `migrate` folder added

internal/api/migrate/migrate.go
```diff
@@ -81,250 +46,159 @@ func ReplaceInputValue(input map[string]string, grafanaDashboard string) string
...
+	if err := grafanaValue.Validate(cue.Final()); err != nil {
+		logrus.WithError(err).Trace("Unable to wrap the received json into a CUE definition")
+		return nil, true, apiinterface.HandleBadRequestError(err.Error())
+	}
 
-	newGrafanaDashboardRaw, err := json.Marshal(grafanaDashboard)
-	if err != nil {
-		logrus.WithError(err).Error("Unable to marshal the modified dashboard")
-		return grafanaDashboardRaw
+	// Finally, execute the cuelang script with the static mapping and the Grafana Panel as a scope.
```
AntoineThebaud, line 182: wrong/deprecated comment

internal/api/migrate/testdata/simple_perses_dashboard.json
```diff
@@ -452,20 +455,23 @@
...
                 },
                 {
                   "header": "API server",
                   "name": "apiserver"
                 },
                 {
-                  "header": "ValueCustomHeader",
-                  "name": "value"
+                  "hide": true,
+                  "name": "timestamp"
```
AntoineThebaud, line 470: weird that the migration is worse here (merged settings getting split), are you sure your maim branch was up to date here?

internal/cli/cmd/migrate/migrate.go
```diff
@@ -131,8 +131,12 @@ func (o *option) onlineExecution(grafanaDashboard json.RawMessage) (*modelV1.Das
 }
 
 func (o *option) offlineExecution(grafanaDashboard json.RawMessage) (*modelV1.Dashboard, error) {
-	finalGrafanaDashboard := migrate.ReplaceInputValue(o.input, string(grafanaDashboard))
-	return o.mig.Migrate([]byte(finalGrafanaDashboard))
+	rowGrafanaDashboard := []byte(migrate.ReplaceInputValue(o.input, string(grafanaDashboard)))
```
AntoineThebaud, line 134:
```suggestion
	rawGrafanaDashboard := []byte(migrate.ReplaceInputValue(o.input, string(grafanaDashboard)))
```

pkg/model/api/v1/common/jsonref.go
```diff
@@ -29,7 +29,7 @@ var (
 
 type JSONRef struct {
 	// Ref is the JSON reference.
-	// That's the only thing used during the marshalling / unmarshalling process.
+	// That's the only thing used during the marshaling / unmarshalling process.
```
AntoineThebaud, line 32: learned sth today:

> Marshaling is predominantly used in 🇺🇸 American (US) English ( en-US ) while marshalling is predominantly used in 🇬🇧 British English (used in UK/AU/NZ) ( en-GB )

Then I guess you should either change nothing here OR change unmarsha**ll**ing too ;D

AntoineThebaud (changes requested):
Nice! since we are merging this on main directly I guess the PR should also update the plugins docs accordingly following the structure changes ;)

internal/api/migrate/testdata/simple_perses_dashboard.json
Nexucis: and it doesn't look like I am outdated regarding the table plugin

internal/api/migrate/loader.go
Nexucis: that's already the case ;)

internal/api/migrate/loader.go
```diff
@@ -14,112 +14,89 @@
...
-	var conditions string
-	conditions, successfulLoadsCount, err := c.buildListOfConditions()
-	if err == nil {
-		c.listOfConditions.Store(&conditions)
+	for _, file := range files {
+		if !file.IsDir() {
+			logrus.Tracef("file %s is ignored since we are looking for directories", file.Name())
+			continue
+		}
+		schemaPath := filepath.Join(path, file.Name(), "migrate")
```
AntoineThebaud, line 43:
```suggestion
		schemaPath := filepath.Join(path, file.Name(), migrationFolder)
```

internal/api/migrate/loader.go
AntoineThebaud, line 69:
```suggestion
		schemaPath := filepath.Join(panelSchemaPath, file.Name(), migrationFolder)
```

AntoineThebaud (approved):
Awesome, 'actually jealous that I didn't have this idea before you. 

🚀

PR #2500, opened 2024-12-12
[ENHANCEMENT] `percli dac diff`: add output + avoid early return when processing a directory

AntoineThebaud:
# Description

- **add output**: `diff` now produces an output similar to what `preview` is doing, with project + dashboard name + status (NEW/UPDATED/ERROR) + Diff file path if applicable:
  ```bash
  $ percli dac diff -d testdata/
  time="2024-12-13T22:23:11+01:00" level=info msg="No dashboard brand-new found in project perses-github-actions, skipping diff generation"
  time="2024-12-13T22:23:11+01:00" level=info msg="built/perses-github-actions-empty_dashboard.diff successfully generated"
  - project: perses-github-actions
    dashboard: brand-new
    status: NEW
  - project: perses-github-actions
    dashboard: empty_dashboard
    status: UPDATED
    diff: built/perses-github-actions-empty_dashboard.diff
  ```
- **avoid early return when processing a directory**: This change is needed to correctly handle the common case where a user of DaC adds a new dashboard: in that case today the diff command fails, thus any further dashboard to process is ignored. This PR fixes that by not stopping the directory processing on individual file error.

# Checklist

### DIFF
--- a/internal/cli/cmd/dac/diff/diff.go
+++ b/internal/cli/cmd/dac/diff/diff.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"path"
 
@@ -127,16 +126,11 @@ func (o *option) Execute() error {
 func (o *option) processDashboardDiff(updatedDashboard *modelV1.Dashboard, project string) (string, string) {
 	currentDashboard, err := o.apiClient.V1().Dashboard(project).Get(updatedDashboard.Metadata.Name)
 	if err != nil {
-		var reqErr *perseshttp.RequestError
-		if errors.As(err, &reqErr) {
-			if reqErr.StatusCode == http.StatusNotFound {
-				logrus.Infof("No dashboard %s found in project %s, skipping diff generation", updatedDashboard.Metadata.Name, project)
-				return statusNew, ""
-			}
-			logrus.WithError(err).Errorf("Unexpected error while fetching dashboard %s in project %s", updatedDashboard.Metadata.Name, project)
-		} else {
-			logrus.WithError(err).Errorf("Unknown error while fetching dashboard %s in project %s", updatedDashboard.Metadata.Name, project)
+		if errors.Is(err, perseshttp.RequestNotFoundError) {
+			logrus.Infof("No dashboard %s found in project %s, skipping diff generation", updatedDashboard.Metadata.Name, project)
+			return statusNew, ""
 		}
+		logrus.WithError(err).Errorf("Unknown error while fetching dashboard %s in project %s", updatedDashboard.Metadata.Name, project)
 		return statusError, ""
 	}
 
--- a/internal/cli/cmd/dac/diff/diff.go
+++ b/internal/cli/cmd/dac/diff/diff.go
@@ -15,11 +15,12 @@ package diff
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


internal/cli/cmd/dac/diff/diff.go
```diff
@@ -82,34 +83,36 @@ func (o *option) Validate() error {
...
+		return fmt.Errorf("error creating the output folder: %v", err)
+	}
+
+	var execResult strings.Builder
 	for _, preview := range o.dashboards {
 		project := resource.GetProject(preview.GetMetadata(), o.Project)
 		dashboard, err := o.apiClient.V1().Dashboard(project).Get(preview.Metadata.Name)
 		if err != nil {
-			return err
+			execResult.WriteString(fmt.Sprintf("No dashboard %s found in project %s, skipping diff generation: %v\n", preview.Metadata.Name, project, err))
```
Nexucis, line 97: you could just write the message directly into the output. Would be simpler don't you think ? And also it will show the progress of the command

internal/cli/cmd/dac/diff/diff.go
```diff
@@ -82,34 +83,36 @@ func (o *option) Validate() error {
...
+	// Create the output folder (+ any parent folder if applicable) where to store the diff files
+	err := os.MkdirAll(config.Global.Dac.OutputFolder, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("error creating the output folder: %v", err)
+	}
+
+	var execResult strings.Builder
 	for _, preview := range o.dashboards {
 		project := resource.GetProject(preview.GetMetadata(), o.Project)
 		dashboard, err := o.apiClient.V1().Dashboard(project).Get(preview.Metadata.Name)
```
Nexucis, line 95: I thought one of the idea was to generate the diff file if the there is no previous version. Would be less miss-leading if we don't display an error in case previous dashboard is not found (since it's a new dasbhoard)

internal/cli/cmd/dac/diff/diff.go
Nexucis: For example, in case of error, you could use the logger in this case instead of using the output handler. Logger will write the log on the error output

internal/cli/cmd/dac/diff/diff.go
AntoineThebaud: we didnt discuss that case afaik but good idea indeed, 'need to see how to cleanly generate such diff (it should full of "+" lines then)

internal/cli/cmd/dac/diff/diff.go
```diff
@@ -82,34 +98,61 @@ func (o *option) Validate() error {
...
-		if outputErr := output.HandleString(o.writer, fmt.Sprintf("%s successfully created", filePath)); outputErr != nil {
-			return outputErr
+	return output.Handle(o.writer, o.Output, result)
+}
+
+func (o *option) processDashboardDiff(updatedDashboard *modelV1.Dashboard, project string) (string, string) {
+	currentDashboard, err := o.apiClient.V1().Dashboard(project).Get(updatedDashboard.Metadata.Name)
+	if err != nil {
+		var reqErr *perseshttp.RequestError
+		if errors.As(err, &reqErr) {
```
Nexucis, line 131: Instead of casting the err, you could simply do :

```suggestion
 if errors.Is(createErr, perseshttp.RequestNotFoundError) {
}
```

Nexucis (approved):

PR #2502, opened 2024-12-13
[IGNORE] Upgrade TypeScript to v5.4

Gladorme:
# Description

Upgrading TypeScript to v5.4, upgrading to a higher version will need to update storybook deps. Checking, but not sure it will as fast/easy as this update 😄

### DIFF
--- a/ui/explore/src/components/PrometheusMetricsFinder/filter/FilterInputs.tsx
+++ b/ui/explore/src/components/PrometheusMetricsFinder/filter/FilterInputs.tsx
@@ -215,7 +215,7 @@ export function RawFilterInput({
             />
           );
         }}
-        onChange={(_: SyntheticEvent, newValue: string[] | string | null) => {
+        onChange={(_, newValue) => {
           if (typeof newValue === 'string') {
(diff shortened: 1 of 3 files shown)


ui/explore/src/components/PrometheusMetricsFinder/filter/FilterInputs.tsx
```diff
@@ -215,7 +215,7 @@ export function RawFilterInput({
             />
           );
         }}
-        onChange={(_: SyntheticEvent, newValue: string[] | string | null) => {
+        onChange={(_, newValue) => {
```
Nexucis, line 218: why did you need to remove every type here ? I thought the previous PRs you made was to add them

Nexucis: after talking with Guillaume, it seems for obscure reason, we cannot keep the type for this particular one as it doesn't compile afterward.

So let's drop it

Nexucis (approved):
noice. Thanks !

PR #2505, opened 2024-12-13
[ENHANCEMENT] `percli dac preview`: add current dashboard URL to the output when applicable

AntoineThebaud:
# Description

Adding the URL of the existing dashboard, when applicable, corresponding to the preview helps the user to quickly see the before & after states, using "current" link + preview link.

# Checklist

### DIFF
--- a/internal/cli/cmd/dac/preview/preview.go
+++ b/internal/cli/cmd/dac/preview/preview.go
@@ -14,8 +14,10 @@
 package preview
 
 import (
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"strings"
 
 	"github.com/perses/perses/internal/api/utils"
@@ -27,6 +29,7 @@ import (
 	"github.com/perses/perses/internal/cli/resource"
 	"github.com/perses/perses/internal/cli/service"
 	"github.com/perses/perses/pkg/client/api"
+	"github.com/perses/perses/pkg/client/perseshttp"
 	modelV1 "github.com/perses/perses/pkg/model/api/v1"
 	"github.com/perses/perses/pkg/model/api/v1/common"
 	"github.com/sirupsen/logrus"
@@ -37,6 +40,7 @@ type previewResponse struct {
 	Project   string `json:"project" yaml:"project"`
 	Dashboard string `json:"dashboard" yaml:"dashboard"`
 	Preview   string `json:"preview" yaml:"preview"`
+	Current   string `json:"current,omitempty" yaml:"current,omitempty"`
 }
 
 func newEphemeralDashboard(project string, name string, ttl common.Duration, dashboard *modelV1.Dashboard) *modelV1.EphemeralDashboard {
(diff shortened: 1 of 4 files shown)


internal/cli/cmd/dac/preview/preview.go
```diff
@@ -146,13 +150,32 @@ func (o *option) computeEphemeralDashboardName(dashboardName string) string {
...
+	name := dashboard.Metadata.Name
+
 	finalURL := &url.URL{}
 	*finalURL = *o.apiClient.RESTClient().BaseURL
-	finalURL.Path = fmt.Sprintf("/%s/%s/%s/%s", utils.PathProject, tmpDashboard.Metadata.Project, utils.PathEphemeralDashboard, tmpDashboard.Metadata.Name)
+	finalURL.Path = fmt.Sprintf("/%s/%s/%s/%s", utils.PathProject, project, utils.PathEphemeralDashboard, tmpDashboard.Metadata.Name)
+
+	// Check if the dashboard already exists to provide the "current" URL
+	currentURL := &url.URL{}
+	_, err := o.apiClient.V1().Dashboard(project).Get(name)
```
AntoineThebaud, line 162: @Nexucis it looks like I shouldn't query the API this way if I want the unit tests to pass.. any idea how to fix that?
Btw we are using this syntax for other commands (`dac diff` & `project`..) but they either have no unit tests, or tests that don't cover such instruction ^^'

Nexucis: actually you can, but you need to inject the fake client available in the unit test with the attribute `APIClient`:

```go
APIClient:       fakeapi.New(),
```

AntoineThebaud: It's already the case :( https://github.com/perses/perses/blob/e31e673e1d28e1999b5d99cd81df290ebf4d224f/internal/cli/cmd/dac/preview/preview_test.go#L43

Nexucis: Ah yeah sorry I jump a bit too quickly to this assumption (that the fake client wasn't used).

Looking at the CI error:

```
    --- FAIL: TestPreviewCMD/preview_with_no_prefix (0.03s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x9ba603]

goroutine 44 [running]:
testing.tRunner.func1.2({0xa45600, 0x1026610})
	/opt/hostedtoolcache/go/1.23.4/x64/src/testing/testing.go:1632 +0x230
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.23.4/x64/src/testing/testing.go:1635 +0x35e
panic({0xa45600?, 0x1026610?})
	/opt/hostedtoolcache/go/1.23.4/x64/src/runtime/panic.go:785 +0x132
github.com/perses/perses/pkg/client/fake/api/v1.(*client).Dashboard(0xafdbd2?, {0xc00051e7b0?, 0xc000099a10?})
```

It panics because the interface `Dashboard` for the fake client is not implemented. You would have to implement it and then tests should work

Gladorme (approved):

internal/cli/cmd/dac/preview/preview.go
```diff
@@ -145,11 +149,29 @@ func (o *option) computeEphemeralDashboardName(dashboardName string) string {
...
 func (o *option) buildPreviewResponse(dashboard *modelV1.Dashboard, tmpDashboard *modelV1.EphemeralDashboard) previewResponse {
-	previewURL := common.NewURL(o.apiClient.RESTClient().BaseURL, utils.PathProject, tmpDashboard.Metadata.Project, utils.PathEphemeralDashboard, tmpDashboard.Metadata.Name)
+	project := tmpDashboard.Metadata.Project
+	name := dashboard.Metadata.Name
+
+	previewURL := common.NewURL(o.apiClient.RESTClient().BaseURL, utils.PathProject, project, utils.PathEphemeralDashboard, tmpDashboard.Metadata.Name)
+
+	// Check if the dashboard already exists to provide the "current" URL
+	currentURL := &common.URL{}
+	_, err := o.apiClient.V1().Dashboard(project).Get(name)
```
Nexucis, line 159:
```suggestion
if _, err := o.apiClient.V1().Dashboard(project).Get(name); err != nil {
      if errors.Is(err, perseshttp.RequestNotFoundError) {
      } else {
      }
}

Nexucis (approved):
excepting my last suggestions, looks good to me. Thank you for finishing this part !

internal/cli/cmd/dac/preview/preview.go
AntoineThebaud: oops 🙊

PR #2514, opened 2024-12-17
[FEATURE] add remote loader for runtime plugins

jgbernalp:
# Description

This PR uses the remote plugin loader to load all types of plugins at runtime, based on the installed plugins fetched from `api/v1/plugins`

This will work in conjunction with https://github.com/perses/plugins/pull/16 and https://github.com/perses/plugins/pull/18

### DIFF
--- a/internal/api/plugin/npm.go
+++ b/internal/api/plugin/npm.go
@@ -25,7 +25,7 @@ const (
 )
 
 type NPMPerses struct {
-	PluginType string `json:"pluginType"`
+	Plugins []PluginMetadata `json:"plugins"`
 }
 
 type NPMPackage struct {
@@ -39,14 +39,14 @@ type BuildInfo struct {
 	Name    string `json:"buildName"`
 }
 
-type Metadata struct {
+type ManifetsMetadata struct {
 	BuildInfo BuildInfo `json:"buildInfo"`
 }
 
 type NPMManifest struct {
-	ID       string   `json:"id"`
-	Name     string   `json:"name"`
-	Metadata Metadata `json:"metaData"`
+	ID       string           `json:"id"`
+	Name     string           `json:"name"`
+	Metadata ManifetsMetadata `json:"metaData"`
 }
 
 func ReadManifest(pluginPath string) (*NPMManifest, error) {
--- a/internal/api/plugin/npm.go
+++ b/internal/api/plugin/npm.go
@@ -17,6 +17,8 @@ import (
 	"encoding/json"
 	"os"
 	"path"
+
+	v1 "github.com/perses/perses/pkg/model/api/v1"
 )
 
 const (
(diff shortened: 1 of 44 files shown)


ui/plugin-runtime/package.json
```diff
@@ -28,7 +28,8 @@
     "lint:fix": "eslint --fix src --ext .ts,.tsx"
   },
   "dependencies": {
-    "@module-federation/enhanced": "^0.1.21"
+    "@module-federation/enhanced": "^0.1.21",
+    "@perses-dev/plugin-system": "0.49.0"
```
Nexucis, line 32: do you foresee in a long term that `plugin-system` and `plugin-runtime` are merged and there is only one package for the plugin system ?

jgbernalp: yes I started as its own package to avoid breaking the current system, prior an initial release we can merge them

ui/plugin-runtime/src/remotePluginLoader.ts
```diff
@@ -0,0 +1,60 @@
...
+export const remotePluginLoader = (baseURL?: string): PluginLoader => {
+  return {
+    getInstalledPlugins: async (): Promise<PluginModuleResource[]> => {
+      const pluginsResponse = await fetch(`${baseURL ? baseURL : ''}/api/v1/plugins`);
+
+      const plugins = await pluginsResponse.json();
+
+      const pluginModules = plugins.map((plugin: Record<string, unknown>) => {
+        return {
+          kind: 'PluginModule',
```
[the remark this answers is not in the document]
jgbernalp: true, let me make the changes.

internal/api/plugin/plugin.go
```diff
@@ -17,18 +17,39 @@ import (
...
 	"github.com/sirupsen/logrus"
 )
 
-const pluginFileName = "plugin.json"
+const (
+	PluginModuleKind = "PluginModule"
+	pluginFileName   = "plugin-modules.json"
+)
+
+type PluginMetadata struct {
```
Nexucis, line 31: I propose to put that in `pkg/model/api/v1` and then to use the classic `Kind, Metadata, Spec`

Then in spec, you can have the Display struct and the list of the plugins

jgbernalp: I tried to use the default Metadata but the version there is numeric, and for the case of the plugins it will use semver. Should we change it there? Am I missing something?

Nexucis: ah yeah I forgot it contains a version in number.

So yeah let's create another metadata for the plugin only. But I would still consider to use the classic struct `Kind`, `Metadata`, `Spec` even if `Metadata` is another one.

Something like that

```go
type PluginSpec {
  Display *common.Display
  Plugins []Plugin
}

type Plugin struct {
  Kind Kind
  Metadata PluginMetadata
  Spec PluginSpec
}
```

internal/api/plugin/npm.go
```diff
@@ -39,14 +39,14 @@ type BuildInfo struct {
 	Name    string `json:"buildName"`
 }
 
-type Metadata struct {
+type ManifetsMetadata struct {
```
AntoineThebaud, line 42:
```suggestion
type ManifestMetadata struct {
```

jgbernalp:
I adjusted the plugin module structure to match better our types conventions as discussed. I added the changes from the plugins side here: https://github.com/perses/plugins/pull/18

jgbernalp:
yes exactly. I'll merge it then. Thnx for checking.

PR #2493, opened 2024-12-12
[BUGFIX] Fix for migration of decimal units

lomobot:
# Description
Fix for migration of decimal units. 
Name of **Decimals** field in json was fixed in migration scripts. (Was **panel.fieldConfig.defaults.decimal**, changed to **panel.fieldConfig.defaults.decimals**)
Example from Grafana dashboard:
```
  "fieldConfig": {
    "defaults": {
      "mappings": [
        {
          "options": {
            "match": "null",
            "result": {
              "text": "N/A"
            }
          },
          "type": "special"
        }
      ],
      "thresholds": {
        "mode": "absolute",
        "steps": [
          {
            "color": "green",
            "value": null
          }
        ]
      },
      "color": {
        "fixedColor": "rgb(31, 120, 193)",
        "mode": "fixed"
      },
      "decimals": 0,
      "unit": "short"
    },
    "overrides": []
  },
```

+ Add unit mapping for "none" to"decimal" and "short" to "decimal".

# Screenshots
image

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

PR #2519, opened 2024-12-19
[BUGFIX] Update unit's migration

lomobot:
# Description
After my previous [PR](https://github.com/perses/perses/pull/2493) problem was discovered, which caused during migration, when "decimals" parameter created, but "unit" is not, because of it's custom-created and not defined in mapping list. As result, such migrated dash can't work in perses, because in panels' schemas defined "unit" attribute as mandatory.
Part of my Grafana dash:
```
          "fieldConfig": {
            "defaults": {
              "custom": {
                "axisCenteredZero": false,
                "axisLabel": "out(-) | in(+)",
                "fillOpacity": 30,
                "gradientMode": "opacity",
                "lineInterpolation": "smooth",
                "lineWidth": 2,
                "showPoints": "never"
              },
              "decimals": 1,
              "unit": "seg/s"
            },
```
Migration result for yAxis attributes:
```
              "yAxis": {
                "format": {
                  "decimalPlaces": 1
                }
              }
```
Error when I tried to import this dash:
```
bad request: invalid panel 7_10: spec.yAxis.format: incomplete value {decimalPlaces:1,unit:"milliseconds" | "seconds" | "minutes" | "hours" | "days" | "weeks" | "months" | "years"} | {decimalPlaces:1,unit:"percent" | "percent-decimal"} | {decimalPlaces:1,unit:"decimal",shortValues?:bool} | {decimalPlaces:1,unit:"bytes",shortValues?:bool} | {decimalPlaces:1,unit:"bits/sec" | "bytes/sec" | "counts/sec" | "events/sec" | "messages/sec" | "ops/sec" | "packets/sec" | "reads/sec" | "records/sec" | "requests/sec" | "rows/sec" | "writes/sec",shortValues?:bool}
```

To resolve this issue I offer to set "decimal" as default unit if another isn't defined after mapping.

# Checklist

### DIFF
--- a/cue/schemas/panels/bar/migrate/migrate.cue
+++ b/cue/schemas/panels/bar/migrate/migrate.cue
@@ -28,7 +28,7 @@ spec: {
 	if #unit != null {
 		format: unit: #unit
 	}
-	#decimal: *#panel.fieldConfig.defaults.decimal | null
+	#decimal: *#panel.fieldConfig.defaults.decimal | *#panel.fieldConfig.defaults.decimals | null
 	if #decimal != null {
 		format: {
 			decimalPlaces: #decimal
--- a/cue/schemas/panels/bar/migrate/migrate.cue
+++ b/cue/schemas/panels/bar/migrate/migrate.cue
@@ -24,8 +24,8 @@ kind: "BarChart"
 spec: {
 	calculation: *commonMigrate.#mapping.calc[#panel.options.reduceOptions.calcs[0]] | commonMigrate.#defaultCalc // only consider [0] here as Perses's GaugeChart doesn't support individual calcs
 
-	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | null
-	if #unit != null {
+	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | "decimal"
+	{
 		format: unit: #unit
 	}
 
--- a/cue/schemas/panels/bar/migrate/migrate.cue
+++ b/cue/schemas/panels/bar/migrate/migrate.cue
@@ -24,13 +24,17 @@ kind: "BarChart"
 spec: {
 	calculation: *commonMigrate.#mapping.calc[#panel.options.reduceOptions.calcs[0]] | commonMigrate.#defaultCalc // only consider [0] here as Perses's GaugeChart doesn't support individual calcs
 
-	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | "decimal"
-	{
+	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | null
+	if #unit != null {
 		format: unit: #unit
 	}
-
-	#decimal: *#panel.fieldConfig.defaults.decimal | *#panel.fieldConfig.defaults.decimals | null
+	#decimal: *#panel.fieldConfig.defaults.decimal | null
 	if #decimal != null {
-		format: decimalPlaces: #decimal
+		format: {
+			decimalPlaces: #decimal
+			if #unit == null {
+				unit: "decimal"
+			}
+		}
 	}
 }
--- a/cue/schemas/panels/pie/migrate/migrate.cue
+++ b/cue/schemas/panels/pie/migrate/migrate.cue
@@ -25,8 +25,8 @@ spec: {
 
 	calculation: *commonMigrate.#mapping.calc[#panel.options.reduceOptions.calcs[0]] | commonMigrate.#defaultCalc // only consider [0] here as Perses's PieChart doesn't support individual calcs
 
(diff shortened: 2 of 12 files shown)


cue/schemas/panels/bar/migrate/migrate.cue
```diff
@@ -24,8 +24,8 @@ kind: "BarChart"
 spec: {
 	calculation: *commonMigrate.#mapping.calc[#panel.options.reduceOptions.calcs[0]] | commonMigrate.#defaultCalc // only consider [0] here as Perses's GaugeChart doesn't support individual calcs
 
-	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | null
-	if #unit != null {
+	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | "decimal"
```
AntoineThebaud, line 27: this fix works however it would be better to keep not creating a format when there's no unit nor decimal

Thus please consider this fix instead:
```suggestion
	#unit: *commonMigrate.#mapping.unit[#panel.fieldConfig.defaults.unit] | null
	if #unit != null {
		format: unit: #unit
	}

	#decimal: *#panel.fieldConfig.defaults.decimal | null
	if #decimal != null {
		format: {
			decimalPlaces: #decimal
			if #unit == null {
				unit: "decimal"
			}
		}
	}
```

internal/api/migrate/testdata/barchart_perses_dashboard.json
```diff
@@ -1,68 +1,68 @@
 {
-  "kind": "Dashboard",
-  "metadata": {
+    "kind": "Dashboard",
```
AntoineThebaud, line 2: For all the unit tests files, please tune your IDE to not alter the indentation already in place (which is 2 Spaces)

AntoineThebaud (changes requested):

cue/schemas/panels/bar/migrate/migrate.cue
lomobot: This fix was created because of we met the case, when custom unit in grafana can’t be mapped by current migration instrument, but other attributes migrates. And we have optional parameters without mandatory, as I show in pr description. Perses can’t import dashboard with such definition.

AntoineThebaud: Yes I see which case you faced, basically `format` is an optional attribute but if ever it is present, a value for `unit` has to be provided. My proposed change also fixes your issue, but enforces this "decimal" default value only when there's something to map to decimalPlaces - instead of systematically with your proposed logic.

lomobot: Oh, I'm terrible sorry=) I saw your reply on smartphone and suggestion window was displayed incorrectly. Now I got it, thank you.

internal/api/migrate/testdata/barchart_perses_dashboard.json
Nexucis: still accurate, @lomobot can you reformat the JSON so it respects the initial indentation ? Thank you by advance

cue/schemas/panels/pie/migrate/migrate.cue
```diff
@@ -29,10 +29,14 @@ spec: {
 	if #unit != null {
 		format: unit: #unit
 	}
-
-	#decimal: *#panel.fieldConfig.defaults.decimal | *#panel.fieldConfig.defaults.decimals | null
```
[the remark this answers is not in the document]
lomobot: oh, thank you so much! my fault, fixed it

internal/api/migrate/testdata/barchart_perses_dashboard.json
lomobot: output migration tests' files was returned to source state (used sources from previous commits)

AntoineThebaud (changes requested):
please get rid of the remaining reindents, see diff for tests

AntoineThebaud:
> > please get rid of the remaining reindents, see diff for tests
> 
> Removed it for barchart dashboard. It was removed at another dashboards earlier. Check it, pls

you still have massive reindents on 3 test files..

image

AntoineThebaud:
> But I removed reindents from these files. Doesn't it?

Don't you see the [+229 −207] at the top hinting that you still have a lot of reindents, & the precise diffs in files changed list? My screenshot above was taken with the latest status of your PR

AntoineThebaud:
> Don't you see the [+229 −207] at the top hinting that you still have a lot of reindents, & the precise diffs in files changed list? My screenshot above was taken with the latest status of your PR

ok so actually for the remaining file you actually fixed the wrong reident done in your previous PR. I would have prefer to fix the indents misalignement issue in a dedicated PR to avoid polluting the diff here but in the end I will take care of it in a follow-up PR.

AntoineThebaud (approved):
Thanks for the fix! 🙏

PR #2565, opened 2025-01-17
sync plugin-runtime after 0.50 release

jgbernalp:


Nexucis:
awesome thank you very much @jgbernalp for doing it. Evertime I am doing it, I am failing on the UI side 😅

PR #2567, opened 2025-01-17
[FEATURE] Load plugin migration files

Nexucis:
This PR introduces a new package `plugin/migrate`  that will replace the current `migrate` package.

It basically just changes the way to load the `migrate.cue` files. With the new implementation, we are opening the file `migrate.cue` to know which kind of plugin it is supporting (panels, queries, variables).


Nexucis:
we will need the v0.4 of the various plugins to makes the CI green again.

internal/api/plugin/migrate/grafana.go
```diff
@@ -0,0 +1,269 @@
...
+	"encoding/json"
+
+	"github.com/perses/perses/pkg/model/api/v1/variable"
+)
+
+const (
+	grafanaPanelRowType = "row"
+)
+
+type GrisPosition struct {
```
AntoineThebaud, line 26:
```suggestion
type GridPosition struct {
```

internal/api/plugin/migrate/migrate.go
```diff
@@ -0,0 +1,302 @@
...
+
+// TODO Put in common with percli plugin lint command.
+func isPackageMigrate(file string) (bool, error) {
+	data, err := os.ReadFile(file)
+	if err != nil {
+		return false, err
+	}
+	return strings.Contains(string(data), "package migrate"), nil
+}
+
```
AntoineThebaud, line 69: maybe a comment to explain the "guessing" logic of this function?

AntoineThebaud (approved):

internal/api/plugin/migrate/migrate.go
AntoineThebaud: my remark was actually for getPluginKind 😄

Nexucis: ah. Well I will put a comment when coming back to this part, which will be soon enough to not forget haha

PR #2549, opened 2025-01-10
[ENHANCEMENT] Unify loading and error handling for all panels, pass data as props to panels

andreasgerstmayr:
# Description

* **unify loading and error handling for all panels:** Move the loading and error handling from panel plugins to parent `<Panel>` component. This reduces the responsibility of the panel plugins to only render the chart, and not having to handle the loading and error states individually.
* **add loading and error indicator to panel header:** This resolves the issue of swallowed errors if one query succeeds and another query fails (see removed comment in `TimeSeriesChartPanel.tsx`) for all panels. Now the successful query is rendered, and a warning indicator shows the error of the failing query.
* **pass query data as props to panel plugins:** This enables embedding a Panel in another Panel, for example the table panel can embed the TimeSeriesChartPanel in its cells. It also simplifies testing of panel plugins.

It uses the following algorithm to determine the panel content:
1. If any query has data available, render the panel.
2. Otherwise, if any query is in loading state, show a loading overlay.
3. Otherwise, if any query has an error, show the error.
4. (afaics this state cannot happen) show an empty chart.

The panel header contains a loading or error indicator:
1. If any query is loading, and the panel content does not already show a loading overlay, show a circular loading indicator.
2. Otherwise, if any query has an error, show a warning icon with the error message in the tooltip.

This is a non-breaking change except for panel plugins which have a different, custom loading state.

# Screenshots

1/2 queries are loading:
Bildschirmfoto vom 2025-01-10 13-05-55

1/2 queries has an error:
Bildschirmfoto vom 2025-01-10 13-04-50

# Checklist

- [X] Pull request has a descriptive title and context useful to a reviewer.
- [X] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [X] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [X] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [X] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [X] Visual tests are stable and unlikely to be flaky.
  See [Storybook](https://github.com/perses/perses/tree/main/ui/storybook#visual-tests)
[...]

PR #2574, opened 2025-01-20
[FEATURE] Label metrics in stat chart

### LINKED ISSUE #2340: StatChart: Show label value instead of timeseries value
### Proposal

Add support for showing metrics labels values instead of value itself.
Useful to show "info-style metrics":
https://www.robustperception.io/why-info-style-metrics-have-a-value-of-1/

For example:
Grafana panel with DB version:
image

it is set in property:       `"fields": "/^version$/"` :
```
{
  "description": "Greenplum database version.",
  "fieldConfig": {
    "defaults": {
      "color": {
        "fixedColor": "text",
        "mode": "fixed"
      },
      "unit": "short"
    },
    "overrides": []
  },
  "gridPos": {
    "h": 4,
    "w": 6,
    "x": 18,
    "y": 1
  },
  "id": 5,
  "options": {
    "reduceOptions": {
      "values": false,
      "calcs": [
        "lastNotNull"
      ],
      "fields": "/^version$/"
    },
    "orientation": "auto",
    "textMode": "auto",
    "wideLayout": true,
    "colorMode": "fixed",
    "graphMode": "none",
    "justifyMode": "auto",
    "showPercentChange": false
  },
  "pluginVersion": "11.0.0",
  "targets": [
    {
      "expr": "gp_version_info{instance=\"$instance\", version!=\"\"}",
      "format": "table",
      "legendFormat": "{{ instance }}",
      "refId": "Database version"
    }
  ],
  "title": "Database version",
  "type": "stat"
}
```

Perses now:
image

abelyakin:
# Description
This PR introduces a new feature that allows the use of label metrics as values in the Stat Chart.

Resolves #2340 

# Screenshots

Screenshot 2025-01-20 at 17 28 45

# Checklist

## UI Changes

  include:


cue/schemas/panels/stat/migrate/migrate.cue
```diff
@@ -34,7 +34,9 @@ import "list"
 kind: "StatChart"
 spec: {
 	calculation: *commonMigrate.#mapping.calc[#panel.options.reduceOptions.calcs[0]] | commonMigrate.#defaultCalc // only consider [0] here as Perses's GaugeChart doesn't support individual calcs
-
+	if (*#panel.options.reduceOptions.fields | _|_) != _|_ {
+		metricLabel: #panel.options.reduceOptions.fields
+	}
```
AntoineThebaud, line 39:
```suggestion
        #metricLabel: *#panel.options.reduceOptions.fields | null
	if #metricLabel != null {
		metricLabel: #metricLabel
	}
```

abelyakin: thanks! accepted it

Nexucis:
perhaps we could avoid the regexp syntax ? Since we want to display a single string coming from a single label. It would simpler to propose instead the label to catch. And in the UI it could be a dropdown if this the chosen source

abelyakin:
> perhaps we could avoid the regexp syntax ? Since we want to display a single string coming from a single label. It would simpler to propose instead the label to catch. And in the UI it could be a dropdown if this the chosen source

It might be confusing on the screenshot but the regexp syntax is optional - for exact matches, you can simply specify the label directly. The regexp option is available for more advanced use cases, such as:
- When label existence isn't guaranteed
- When you need to specify fallbacks
- When you need pattern matching

For example, in our dashboard we use patterns like `^(version|version2)$` to match either "version" or "version2" labels.

Nexucis:
Alright, thanks for the clarification @abelyakin ! Sounds good to me.

Will see what others will have to say on this PR !

7840vz:
> perhaps we could avoid the regexp syntax ? Since we want to display a single string coming from a single label. It would simpler to propose instead the label to catch. And in the UI it could be a dropdown if this the chosen source

Grafana does it as regex, to support catching values from multiple labels, therefore it would be easier to migrate it that way from grafana if regex is supported here as well. in 99% it is single label though.

Gladorme (approved):
LGTM :)

ui/plugin-system/src/components/MetricLabelInput/MetricLabelInput.tsx
```diff
@@ -0,0 +1,36 @@
+// Copyright 2023 The Perses Authors
```
Nexucis, line 1:
```suggestion
// Copyright 2025 The Perses Authors
```

It's a new year ;)

ui/plugin-system/src/components/MetricLabelInput/index.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2023 The Perses Authors
```
Nexucis, line 1:
```suggestion
// Copyright 2025 The Perses Authors
```

cue/schemas/common/metricLabel.cue
```diff
@@ -0,0 +1,16 @@
+// Copyright 2023 The Perses Authors
```
Nexucis, line 1:
```suggestion
// Copyright 2025 The Perses Authors
```

Nexucis:
> Grafana does it as regex, to support catching values from multiple labels, therefore it would be easier to migrate it that way from grafana if regex is supported here as well. in 99% it is single label though.

If 99% of the time, it's a single label, maybe we could have an autocompletion of the labels. Would be nice to see the available labels when typing in this input if we are not using the Regexp syntax.

But for sure it can be done later. WDYT @abelyakin ?

Otherwise, minus my last suggestions to change the year of licensing,  like I said I am fine with the changes and ok to merge it since @Gladorme gave a go for that.

Nexucis:
FYI @jgbernalp I think for this particular panel, we will have some trouble to migrate it since it depends on packages that are not yet released.

Gladorme:
> If 99% of the time, it's a single label, maybe we could have an autocompletion of the labels. Would be nice to see the available labels when typing in this input if we are not using the Regexp syntax.

No possible for the moment, we want to have autcomplete for table too. But data is not passed to editor for now. Maybe it will be easier with #2549? 🤔

Nexucis:
ah yeah I forgot the panel doesn't have access to the data coming from the query. My bad.
Yeah we will need to discuss probably about #2549. That will be a big change to pass the data to the panel

Nexucis (approved):
moving forward on this PR as I would like to see it merged before we are merging the new plugin architecture. I will make the changes myself.

Thank you for the enhancement @abelyakin !

PR #2577, opened 2025-01-22
[BUGFIX] Login from percli (method not allowed)

celian-garcia:
# Description

``percli login <url>``  fails now with following log in the backend 

time="2025-01-22T08:30:19Z" level=info func="github.com/perses/common/echo.(*Builder).build.Logger.LoggerWithConfig.func2.1" file="/home/runner/go/pkg/mod/github.com/perses/common@v0.26.0/echo/middleware/logger.go:63" method=POST status=405 uri=//api/auth/providers/oidc/azure/device/code

I suspected the // and this code is making sure that we don't create a URL with //api when calling the device auth. 

As it's something that we do several time in the code, I created a function to mutualize. 

Now it works well we can login with ``percli login <url>``

# Screenshots

# Checklist

## UI Changes

  include:


pkg/client/api/auth/auth.go
```diff
@@ -129,9 +130,9 @@ func (c *auth) ClientCredentialsToken(authKind, slugID, clientID, clientSecret s
 }
 
 func (c *auth) deviceAuthURL(authKind, authProvider string) string {
-	return fmt.Sprintf("%s%s/%s/%s/%s/%s", c.client.BaseURL.String(), utils.APIPrefix, utils.PathAuthProviders, authKind, authProvider, utils.PathDeviceCode)
+	return urlutils.CloneURLAddingPath(c.client.BaseURL, utils.APIPrefix, utils.PathAuthProviders, authKind, authProvider, utils.PathDeviceCode).String()
```
Nexucis, line 133: I don't get why it is fixing the issue ? For me it doesn't change anything excepting adding more complexity on how we are managing URL currently.

pkg/client/url/url.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package url
+
+import (
+	"net/url"
+	"path"
+)
+
+func CloneURLAddingPath(u *url.URL, sections ...string) *url.URL {
```
Nexucis, line 21: I would put that in `pkg/api/v1/common/url.go` instead and rename it by

```suggestion
func NewURL(url *URL, paths ...string) *URL {
```

Nexucis:
strange things the e2e tests are not covering this issue. Do you know why ?

pkg/client/api/auth/auth.go
celian-garcia: .String() was giving a trailing slash, and APIPrefix was containing a starting slash so it was doubling the slash. In the clone function I'm using a mechanism used heavily already in the rest of the code so I mutualized

pkg/client/url/url.go
Nexucis: I did the changes as Celian doesn't have more time to spend on this fix.

Nexucis (approved):

PR #2576, opened 2025-01-21
[BUGFIX] Inject API prefix in all UI assets

Nexucis:
This PR is adding a placeholder in every route contain the JS files built for the Perses app.

Then at runtime, we are opening the JS files to replace this placeholder with the actual API prefix. This should fix #2575.

Why we didn't do that since the beginning ? Excellent question, let me answer it right away : 

The backend contains actually a middleware that redirect any requests that doesn't contain the API prefix to the same route containing this API prefix. I have assumed (wrongly) it will allow to serve every assets without modifying/injecting the API prefix in every JS files. 

It works if the requests to the root is actually send to the backend. Which is not the case when it's behind a reverse proxy because well the reverse proxy need the API prefix to know where to redirect the requests.

Anyway, it was the opportunity to rewrite the code serving the frontend and clarifying the steps with more comments.

/cc @ibakshay 

Edit: 

The internal proxy in Perses backend wasn't supporting the API proxy. The last commit is tackling this issue.

PR #2586, opened 2025-01-23
sync main

Nexucis:
This PR is synchronising the main branch with `feature/plugin-runtime`.

It also upgrade `rspack` to the last version (v1.2.2) and fixing the `index.html` files that has been changed from main.

When upgrading `rspack`  I have noticed a strange warning:

```bash
@perses-dev/app:build: WARNING in ../plugin-system/dist/remote/PluginRuntime.js
@perses-dev/app:build:   ⚠ ESModulesLinkingWarning: export 'default' (imported as 'ReactRouterDOM') was not found in 'react-router-dom' (possible exports: AbortedDeferredError, Await, BrowserRouter, Form, HashRouter, Link, MemoryRouter, NavLink, Navigate, NavigationType, Outlet, Route, Router, RouterProvider, Routes, ScrollRestoration, UNSAFE_DataRouterContext, UNSAFE_DataRouterStateContext, UNSAFE_ErrorResponseImpl, UNSAFE_FetchersContext, UNSAFE_LocationContext, UNSAFE_NavigationContext, UNSAFE_RouteContext, UNSAFE_ViewTransitionContext, UNSAFE_useRouteId, UNSAFE_useScrollRestoration, createBrowserRouter, createHashRouter, createMemoryRouter, createPath, createRoutesFromChildren, createRoutesFromElements, createSearchParams, defer, generatePath, isRouteErrorResponse, json, matchPath, matchRoutes, parsePath, redirect, redirectDocument, renderMatches, replace, resolvePath, unstable_HistoryRouter, unstable_usePrompt, unstable_useViewTransitionState, useActionData, useAsyncError, useAsyncValue, useBeforeUnload, useBlocker, useFetcher, useFetchers, useFormAction, useHref, useInRouterContext, useLinkClickHandler, useLoaderData, useLocation, useMatch, useMatches, useNavigate, useNavigation, useNavigationType, useOutlet, useOutletContext, useParams, useResolvedPath, useRevalidator, useRouteError, useRouteLoaderData, useRoutes, useSearchParams, useSubmit)
```

And yes indeed there is no function or constant called `ReactRouterDOM` coming from the package `react-router-dom`. Is it normal @jgbernalp ?


jgbernalp:
Let me check.

Nexucis:
@jgbernalp it looks like your last commit break the e2e tests :(

jgbernalp:
@Nexucis they were already broken, but the last commit should fix it.

Nexucis:
ah yeah sorry about that @jgbernalp

ui/app/rspack.config.ts
```diff
@@ -23,7 +23,6 @@ module.exports = {
   context: __dirname,
   output: {
     path: path.resolve(__dirname, './dist'),
-    publicPath: 'PREFIX_PATH_PLACEHOLDER/',
```
Nexucis, line 26: mmm that would be annoying to remove it as it is necessary for the `prefix_api` things. See #2576

Nexucis: and it doesn't seem it fixed the e2e issue @jgbernalp.

But what we could do, is to rollback to remove the upgrade of rspack so it's not blocking the merge of this PR and the merge of the plugin.

What do you think ?

Nexucis: but removing the `PREFIX_PATH_PLACEHOLDER` is fixing the issue that `npm start` doesn't work anymore.

Probably a compromise would be to do `publicPath: isDev? '': 'PREFIX_PATH_PLACEHOLDER/'`

jgbernalp: Ok then it needs to be only when building, but not on dev.

jgbernalp: this missing prefix also caused also an issue with the favicon so the e2e tests failed. But it should be good now. I applied the publicPath for non dev environments.

Nexucis: unfortunately it seems there is something else

jgbernalp (approved):

PR #2593, opened 2025-01-27
[FEATURE] Perses server is able to load plugin from dev environment server

Nexucis:
Here a proposition to let the Perses server to read plugin serving by the dev server from `rsbuild`.

You just need to configure in Perses the dev environment and then it will read the schema and refresh the list of module available on `/api/v1/plugins`.

It will also forward the request to the dev server when needed. 

To be used like that: 

```yaml
plugins:
  dev_environment:
    url: "http://localhost:3005"
    plugins:
      - name: TimeSeriesChart
        absolute_path: "/Users/ahusson/workspace/go/src/perses/plugins/TimeSeriesChart"
```

I am just wondering if it's a good idea to have this configuration in the global config of Perses or if we should have it in a different file.
On the other hand, if someone want to split the Perses config in different files, there are tools for that that will compute and merge multiple files to a single one. That's what we are using to build the perses demo config. 

There is a based config file versioned on the repo `perses/website` and then there is another file directly on the VM where you have all the secret. Then these files are merged to give the final config.

So probably it's fine.

Related to the issue https://github.com/perses/plugins/issues/34

### DIFF
--- a/dev/config.yaml
+++ b/dev/config.yaml
@@ -26,13 +26,6 @@ provisioning:
   folders:
     - "dev/data"
 
-schemas:
-  panels_path: "cue/schemas/panels"
-  queries_path: "cue/schemas/queries"
-  datasources_path: "cue/schemas/datasources"
-  variables_path: "cue/schemas/variables"
-  interval: "5m"
-
 ephemeral_dashboard:
   enable: true
   cleanup_interval: 1h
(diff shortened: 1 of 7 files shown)


jgbernalp:
@Nexucis do we need the `absolute_path`? IIUC is loading the schema and migration. But we can also include those files as part of the rsbuild dev server. That way the configuration will be just:

```
plugins:
  dev_environment:
    url: "http://localhost:3005"
    plugins:
      - name: TimeSeriesChart
```

Also, for a future enhancement, we could provide a url for each plugin.

```
plugins:
  dev_environment:
    plugins:
      - name: TimeSeriesChart
        url: "http://localhost:3005"
      - name: Prometheus
        url: "http://localhost:3002"
```

jgbernalp:
> Should we support a global URL that can be overridden in each plugin configuration ?

Yes I think that will be great. Feel free to do it in a following PR.

jgbernalp (approved):

Issue #1543, opened 2023-10-25
Real plugin architecture & clean installation flow


AntoineThebaud:
### Proposal

Our current plugin architecture technically works, still it's tightly coupled with Prometheus datasource and installing a new plugin is not possible as is (it's possible to hot-load the backend part of it, but for the frontend part doing 99% of the job it actually has to be part packaged with Perses).

The idea would be to really have externalized plugins & provide a single entry point to install them, that would take as input a package containing both the backend & frontend parts of the plugin, be able to "install" it & have Perses updated live accordingly, with as less amount of actions as possible:
- using the UI, ideally drag & drop + submit, that's it?
- using the CLI, ideally single command?
- provisioned set of plugins when starting Perses that would work using the same package format?

AntoineThebaud:
FYI starting a POC around this topic here https://github.com/perses/poc-plugin-v2

jgbernalp:
In addition to this:

What pain points did we have?
- Too many coupled providers make the plugin usage cumbersome when using them in embedding environments and also in Perses UI, maybe creating more reusable components and let the consumer, including Perses UI, connect them with providers can be more flexible.
- Although There is inherent coupling between queries, datasources and variables. In the UI the panel plugins and the queries are also coupled, this creates the challenge to adjust every panel to a specific query/datasource. Maybe query or datasource plugins should generate a common format (DataFrame) so they can be connected with several panels without a re-implementation
- (same as point 1) The plugin needs to be in the core source code, so it's not really pluggable at runtime.

What would an easy plugin usage look like?
- The plugin(s) should be self contained as a single folder / file that can be installed on runtime or pre-packaged in Perses
- Panel plugin components could be dependent only on ECharts easing the embedded usage, reducing the footprint the reusable components should have.
- Have a Perses plugin SDK (point 5) so boilerplates can be generated and packaged for distribution.
- If plugins can be added at runtime we need to make sure we isolate their execution (sandboxing) so an error wont crash the UI / Backend, in the frontend this is already considered using error boundaries.
- We can define a PluginManager that takes care of the discovery, installation and plugin lifecycle: installation, activation, deactivation, and cleanup.
- It would be great to have an official plugin marketplace/catalog where users can discover and install plugins easily

Nexucis:
on top of what have been said, I think it will be cool if we are able to include also this issue https://github.com/perses/perses/issues/1546.

Would be cool to be able to provide an icon in the plugin package !

jgbernalp:
[...]

PR #2598, opened 2025-02-03
[BREAKINGCHANGE] remove explorer references to local plugins and allow them to load from remote plugins

jgbernalp:
# Description

This PR:

- Removes prometheus and tempo local plugins, to remove their references from the explore page. 
- Changes the explore package to load "Explore" plugins using remote plugins.
- Adds the plugin module info into the plugin metadata so plugins know from which module to load
- Exports the ExploreManagerProvider and query-params as this needs to be consumed by the implementations of "Explore" plugins.

# Screenshots

No UI changes

# Checklist


AntoineThebaud:
> @AntoineThebaud on this PR mainly the cue files related to DaC are failing as they are not anymore part of the perses repo. Should these files also be part of individual plugins?

Oh indeed we should extract these dac-utils files for prometheus out of this repo.. IMHO this kind of utils for a plugin should be part of the same repo as the plugin itself, however it should not get packaged in the plugin archive (as it has no purpose for that)

Nexucis:
@jgbernalp unfortunately I think we should not remove the schema yet and do it in a dedicated PR once we have migrated the various SDK to the plugin repo.
I will update the issue https://github.com/perses/perses/issues/1543 to reflect this issue

Nexucis:
I guess this PR is ready to be reviewed and merged right @jgbernalp ?

jgbernalp:
> I guess this PR is ready to be reviewed and merged right @jgbernalp ?

yes it is now

Nexucis (approved):
awesome work as usual @jgbernalp. Thank you for being so reactive on this topic 🙏

PR #2613, opened 2025-02-07
[ENHANCEMENT] detect if Perses is used in container to add default path

Nexucis:
I am proposing to detect if Perses is used in a container, so we can inject the default value where the plugins are stored


Gladorme (approved):
cool! blind 😄

PR #2617, opened 2025-02-08
[ENHANCEMENT] Disable schema validation by config for plugin in development

Nexucis:
This PR is introducing a way to disable the schema validation for the plugin in development. It is just to aknowledge the fact, when implementing a new plugin, we don't necessary have yet the associated schema and as such we shouldn't ask to have it immediately.

Note that for the moment, if this option is activated, then you won't be able to save any data as the schema has not been loaded.

I also took the opportunity to update the configuration doc about the plugin config.

Finally I have changed the attribute for the plugin config. It's not anymore `plugins` but simply `plugin` without the `s`


docs/configuration/configuration.md
```diff
@@ -666,3 +684,48 @@ project: <string>
...
+
+##### PluginInDevelopment config
+
+```yaml
+# The name of the plugin in development
+name: <string>
+
+# A way to disable the schema validation of the plugin in development.
+# It can be useful when you are developing a plugin and, you don't want to have the schema validation. 
+# For example, because you didn't yet define the schema.
```
AntoineThebaud, line 722:
```suggestion
# It can be useful for example when you are developing a plugin and you didn't define the schema yet.
```

AntoineThebaud (approved):

PR #2631, opened 2025-02-11
[FEATURE] Add OAuth secret type for Datasources

### LINKED ISSUE #2535: Configure a datasource with Oauth
### Proposal

Today it's not possible to provide an Oauth configuration for the datasource and as such, it's not possible to connect Perses to a datasource behind Oauth.

Would be great to have it. For that we will have to modify the proxy to handle such use case.

dougkirkley:
# Description

Adds an OAuth secret for use in perses Datasources via the http proxy

closes #2535

# Screenshots

# Checklist

## UI Changes

  include:


pkg/model/api/v1/secret/oauth.go
```diff
@@ -0,0 +1,154 @@
...
+		Scopes:           oauth.Scopes,
+		EndpointParams:   oauth.EndpointParams,
+		AuthStyle:        oauth.AuthStyle,
+	}
+}
+
+type OAuth struct {
+	token            *oauth2.Token
+	tlsConfig        *TLSConfig
+	ClientID         string           `json:"client_id" yaml:"client_id"`
```
Nexucis, line 48: when we are putting resources in pkg/model/api, we are enforcing the usage of the camelCase for YAML and JSON tags.

```suggestion
ClientID string `json:"clientID" yaml:"clientID"`
```

pkg/model/api/v1/secret/oauth.go
```diff
@@ -0,0 +1,154 @@
...
+func (o *OAuth) GetToken(ctx context.Context) (*oauth2.Token, error) {
+	httpClient, err := newHTTPClient(o.tlsConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// add out http client with tls config
+	context.WithValue(ctx, oauth2.HTTPClient, httpClient)
+
+	if o.ClientSecret == "" && o.ClientSecretFile != "" {
```
Nexucis, line 94: for this part, I think it won't hurt if `oauth` has a function to get the `clientSecret`, like what we are providing in the secret.Authorization

```go
func (o *OAuth) GetClientSecret() (string, error) {
	if len(o.ClientSecretFile) > 0 {
		data, err := os.ReadFile(o.ClientSecretFile)
		if err != nil {
			return "", err
		}
		return string(data), nil
	}
	return a.ClientSecret, nil
}
```

pkg/model/api/v1/secret/oauth.go
```diff
@@ -0,0 +1,154 @@
...
+	"golang.org/x/oauth2/clientcredentials"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+)
+
+type PublicOAuth struct {
+	httpClient       *http.Client
+	tlsConfig        *TLSConfig
```
Nexucis, line 18: there is a public TLSConfig, you should use it

pkg/model/api/v1/secret.go
```diff
@@ -25,6 +24,8 @@ type SecretSpec struct {
 	BasicAuth *secret.BasicAuth `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
 	// The HTTP authorization credentials for the targets.
 	Authorization *secret.Authorization `json:"authorization,omitempty" yaml:"authorization,omitempty"`
+	// Oauth configuration for the targets.
+	OAuth *secret.OAuth
```
Nexucis, line 28:
```suggestion
OAuth *secret.OAuth `json:"oauth,omitempty" yaml:"oauth,omitempty"`
```

Nexucis: And since you are updating the `SecretSpec`, can you also update the `Encrypt` and `Decrypt` function here: https://github.com/perses/perses/blob/main/internal/api/crypto/crypto.go#L65

Any secret field identified by the usage of `secret.Hidden` when it goes to public, should be encrypted

Nexucis (changes requested):
Really cool to see this PR @dougkirkley, thank you for doing that. It is going on the right directing I believe.

I have added few comments to help you moving forward on this work.

internal/api/impl/proxy/proxy.go
```diff
@@ -286,6 +286,15 @@ func (h *httpProxy) setupAuthentication(req *http.Request) error {
 		}
 		req.Header.Set("Authorization", fmt.Sprintf("%s %s", auth.Type, credential))
 	}
+	oauth := h.secret.OAuth
+	if oauth != nil {
+		token, err := oauth.GetToken(req.Context())
+		if err != nil {
+			return err
+		}
+		req.Header.Set("Authorization", fmt.Sprintf("%s %s", "Bearer", token.AccessToken))
```
Nexucis, line 295: It's simpler like that:

```suggestion
req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token.AccessToken))
```

cue/model/api/v1/dashboard_go_gen.cue
```diff
@@ -4,7 +4,7 @@
 
 package v1
 
-import "github.com/perses/perses/cue/model/api/v1/common"
+import "github.com/perses/perses/pkg/model/api/v1/common"
```
Nexucis, line 7: this change shouldn't exist. Are you on a Mac ? If it's the case I guess the `sed` command didn't work.

pkg/model/api/v1/secret/oauth.go
```diff
@@ -0,0 +1,111 @@
...
+	"encoding/json"
+	"fmt"
+	"golang.org/x/oauth2"
+	"net/http"
+	"net/url"
+	"os"
+)
+
+type PublicOAuth struct {
+	httpClient       *http.Client
```
Nexucis, line 26: It's not used right ? Can you remove it ?

internal/api/crypto/crypto.go
```diff
@@ -80,6 +80,14 @@ func (c *crypto) Encrypt(spec *modelV1.SecretSpec) error {
 		}
 		authorization.Credentials = encryptedCredentials
 	}
+	oauth := spec.OAuth
+	if oauth != nil {
+		encryptedSecret, err := c.encrypt(oauth.ClientSecret)
```
Nexucis, line 85: clientID and clientSecret should be encrypted

cue/model/api/v1/secret/oauth_go_gen.cue
```diff
@@ -0,0 +1,22 @@
...
+	clientID:         #Hidden @go(ClientID)
+	clientSecret:     #Hidden @go(ClientSecret)
+	clientSecretFile: string  @go(ClientSecretFile)
+	tokenURL:         string  @go(TokenURL)
+	scopes: [...string] @go(Scopes,[]string)
+	endpointParams: url.#Values       @go(EndpointParams)
+	authStyle:      oauth2.#AuthStyle @go(AuthStyle)
+}
+
+#OAuth: _
```
Nexucis, line 22: @AntoineThebaud we should have the new OAuth struct here. I don't remember what you told me to get it or if there was a solution. Can you help us for this issue please ?

Nexucis (changes requested):
This is in a good shape, I have added more comments, but they are not severe.

cue/model/api/v1/dashboard_go_gen.cue
dougkirkley: I am, and I did see some sed errors, I thought it was fishy. I'll fix that.

cue/model/api/v1/secret/oauth_go_gen.cue
```diff
@@ -0,0 +1,22 @@
...
+)
+
+#PublicOAuth: {
+	clientID:         #Hidden @go(ClientID)
+	clientSecret:     #Hidden @go(ClientSecret)
+	clientSecretFile: string  @go(ClientSecretFile)
+	tokenURL:         string  @go(TokenURL)
+	scopes: [...string] @go(Scopes,[]string)
+	endpointParams: url.#Values       @go(EndpointParams)
+	authStyle:      oauth2.#AuthStyle @go(AuthStyle)
```
Nexucis, line 19: Cuelang package `golang.or/x/oauth2` does not exist and so the cuelang tests are failing. 

`@oauth2.#AuthStyle is a number.` @AntoineThebaud what should we do to fix this issue ?

pkg/client/config/config.go
```diff
@@ -73,40 +49,25 @@ func NewPublicRestConfigClient(config *RestConfigClient) *PublicRestConfigClient
...
 	}
+
+	tlsConfig := secret.NewPublicTLSConfig(config.TLSConfig)
+
 	return &PublicRestConfigClient{
 		URL:           config.URL,
 		NativeAuth:    api.NewPublicAuth(config.NativeAuth),
 		BasicAuth:     secret.NewPublicBasicAuth(config.BasicAuth),
-		Oauth:         newPublicOauth(config.OAuth),
+		Oauth:         secret.NewPublicOAuth(config.OAuth, tlsConfig),
```
Nexucis, line 59:
```suggestion
secret.NewPublicOAuth(config.OAuth),
```

internal/api/impl/proxy/proxy.go
```diff
@@ -14,9 +14,13 @@
 package proxy
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
```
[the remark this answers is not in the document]
Nexucis: Expected result will be

```go
import (
	"context"
	"crypto/tls"
	"encoding/json"
	"errors"
	"fmt"
	"io"
	"net"
	"net/http"

	"golang.org/x/oauth2"
	"golang.org/x/oauth2/clientcredentials"
)
```

pkg/model/api/v1/secret.go
```diff
@@ -16,7 +16,6 @@ package v1
 import (
 	"encoding/json"
 	"fmt"
-
```
Nexucis, line 19: this empty line shouldn't be removed. See my comment regarding the import

pkg/model/api/v1/secret/oauth.go
```diff
@@ -0,0 +1,106 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package secret
+
+import (
+	"encoding/json"
+	"fmt"
+	"golang.org/x/oauth2"
```
Nexucis, line 19: please sort the import like assess in a previous comment

internal/api/e2e/api/auth_test.go
```diff
@@ -375,7 +375,7 @@ func TestAuth_OAuthProvider_Token_WithLib(t *testing.T) {
 
 		authenticatedClient, err := config.NewRESTClient(config.RestConfigClient{
 			URL: persesBaseURL,
-			OAuth: &config.OAuth{
+			OAuth: &secret.OAuth{
```
Nexucis, line 378: you need to import the `secret` package.

Nexucis: Also, you will have to fix the usage of the struct `OAuth` that you moved from the `config` package to the `secret` package in the whole file

internal/api/impl/proxy/proxy.go
```diff
@@ -286,9 +291,60 @@ func (h *httpProxy) setupAuthentication(req *http.Request) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := http.Client{
+		Transport: transport,
+	}
+
+	// add our http client with tls config
+	context.WithValue(ctx, oauth2.HTTPClient, httpClient)
```
Nexucis, line 319:
```suggestion
newContext := context.WithValue(ctx, oauth2.HTTPClient, httpClient)
```

Then you should use this new context when calling `conf.Token`

cue/model/api/v1/secret/oauth_go_gen.cue
Nexucis: ah yes indeed. We missed that. We will have to create patch files like `authorization_patch.cue`, but that's not under the scope of this PR.

So it's good here.

Nexucis:
Almost good @dougkirkley :). Can you update the doc https://github.com/perses/perses/blob/main/docs/api/secret.md ? 

Did you test at least manually it's working ?

dougkirkley:
@Nexucis I have not fully tested it yet, I wanted to make sure we had a schema that made sense first. Is there a recommended path to test the API without UI changes other than go tests?

Nexucis:
@dougkirkley the code in the current shape looks reasonable to me and I will be happy to merge it.

Regarding the tests, yeah excepting by go tests there is no other automatic way to do it currently. For such use case, we would need some tests with real deployment I believe. But it's not in place and it's quite heavy to do it.

But for sure it would be great to cover it with e2e tests. Perhaps we can start a prometheus in a container in the github action and then setup some simple e2e tests here: https://github.com/perses/perses/tree/main/internal/api/e2e/api.

I don't know if it's easy to setup a Prometheus protected by OAuth in a Github action, but at least it will be interesting just to start covering the proxy implementation. But this is out of the scope of this PR. That's something that should be done first in a different PR to setup simple test, and then on this one, see how we can have an OAuth setup.

dougkirkley:
@Nexucis so I did a little test using the `percli` and was able to apply the dataSource but when getting it afterwards it doesn't return the new fields, so it's not getting set somewhere or possibly the struct field tags maybe are missing something?

dougkirkley:
@Nexucis So thinking about it more, should `OAuth` be a type instead of its own thing? I don't think I can see how someone would need either `Basic Auth` or `Authorization` with `OAuth`, I feel like maybe all three should be mutually exclusive auth options.

docs/api/secret.md
```diff
@@ -61,6 +64,27 @@ credentials: <string> # Optional
...
+# ClientID is the application's ID.
+clientID: <string>
+# ClientSecret is the application's secret.
+clientSecret: <string>
+clientSecretFile: <filename> # Optional
+# TokenURL is the resource server's token endpoint URL. 
+# This is a constant specific to each server.
+tokenURL: <string> 
+# Scope specifies optional requested permissions.
+scopes: <array> # Optional
```
Nexucis, line 79:
```suggestion
scopes:
- <string> # Optional
```

Nexucis:
> @Nexucis So thinking about it more, I don't think I can see how someone would need either `Basic Auth` or `Authorization` with `OAuth`, I feel like maybe all three should be mutually exclusive auth options. I guess I was thinking that based on updating the docs.

yes this is entirely true, you cannot setup a basic auth at the same time than other, and same apply with others

Nexucis:
tests are failing @dougkirkley because of the cast missing.

Nexucis (approved):
awesome work! Thank you for making this happen @dougkirkley !

PR #2661, opened 2025-02-18
[ignore] use function LoadModelSchema in the discovery tests

Nexucis:


AntoineThebaud (approved):
Oh ok that's what you had in mind I see.. thanks

PR #2663, opened 2025-02-18
[IGNORE] add <PanelPluginLoader> component to simplify embedding panels

andreasgerstmayr:
# Description

This component wraps loading a panel plugin from the registry and rendering its PanelComponent. Example usage:

```
<PanelPluginLoader
  kind="StatChart"
  spec={{
    calculation: 'last-number',
    format: {
      unit: 'decimal',
    },
    sparkline: {},
  }}
  contentDimensions={{ width: 300, height: 100 }}
  queryResults={[
    { definition: { kind: 'TimeSeriesQuery', spec: { plugin: { kind: '', spec: {} } } }, data: tsData },
  ]}
/>
```

This is useful e.g. in the Table panel, when rendering Panels in a table cell.

# Screenshots

no UI changes

# Checklist

## UI Changes

  include:

### DIFF
--- a/ui/dashboards/src/components/Panel/PanelContent.tsx
+++ b/ui/dashboards/src/components/Panel/PanelContent.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { usePlugin, PanelProps, QueryData, PanelPlugin } from '@perses-dev/plugin-system';
+import { usePlugin, PanelProps, QueryData, PanelPlugin as IPanelPlugin } from '@perses-dev/plugin-system';
 import { UnknownSpec, PanelDefinition, QueryDataType } from '@perses-dev/core';
 import { ReactElement } from 'react';
 import { LoadingOverlay } from '@perses-dev/components';
@@ -80,7 +80,7 @@ export function PanelContent(props: PanelContentProps): ReactElement {
 }
 
 interface PanelLoadingProps extends Pick<PanelContentProps, 'spec' | 'definition' | 'contentDimensions'> {
-  plugin?: PanelPlugin<UnknownSpec, PanelProps<UnknownSpec, QueryDataType>>;
+  plugin?: IPanelPlugin<UnknownSpec, PanelProps<UnknownSpec, QueryDataType>>;
 }
 
 function PanelLoading({ plugin, spec, definition, contentDimensions }: PanelLoadingProps): ReactElement {
--- a/ui/dashboards/src/components/Panel/PanelContent.tsx
+++ b/ui/dashboards/src/components/Panel/PanelContent.tsx
@@ -11,12 +11,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { usePlugin, PanelProps, QueryData, PanelPlugin as IPanelPlugin } from '@perses-dev/plugin-system';
+import { usePlugin, PanelProps, QueryData, PanelPlugin } from '@perses-dev/plugin-system';
 import { UnknownSpec, PanelDefinition, QueryDataType } from '@perses-dev/core';
 import { ReactElement } from 'react';
 import { LoadingOverlay } from '@perses-dev/components';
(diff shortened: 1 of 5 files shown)


ui/dashboards/src/components/Panel/PanelPlugin.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface PanelPluginProps extends PanelProps<UnknownSpec, QueryDataType> {
+  kind: string;
+}
+
+/**
+ * PanelPlugin loads the panel plugin specified by the 'kind' prop from the plugin registry and
+ * renders its PanelComponent.
+ */
+export function PanelPlugin(props: PanelPluginProps): ReactElement {
```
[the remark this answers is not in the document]
andreasgerstmayr: I went with `PanelPluginLoader` for now, a bit more descriptive than `PanelPlugin` and it doesn't collide with the `PanelPlugin` type from `plugin-system`.

ui/dashboards/src/components/Panel/PanelContent.tsx
```diff
@@ -24,16 +25,13 @@ export interface PanelContentProps extends Omit<PanelProps<UnknownSpec>, 'queryR
...
 
 /**
- * A small wrapper component that renders the appropriate PanelComponent from a Panel plugin based on the panel
- * definition's kind. Used so that an ErrorBoundary can be wrapped around this.
+ * Based on the status of the queries (loading, error or data available), this component renders a
+ * loading overlay, throws an error, or renders the panel content.
  */
 export function PanelContent(props: PanelContentProps): ReactElement {
   const { panelPluginKind, definition, queryResults, spec, contentDimensions } = props;
   const { data: plugin, isLoading: isPanelLoading } = usePlugin('Panel', panelPluginKind, { throwOnError: true });
```
Gladorme, line 33: Little bit sad this call is made twice (here and in PanelPluginLoader), do you think we can make a unique call?
+
Now, PanelPluginLoader should be responsible load plugin but the PanelContent do it too :/

Not super clear to me, what do you think about it?

Gladorme: This call usePlugin is only useful for showing custom plugin loader, but no plugin are using currently, should we keep this feature? 🤔

andreasgerstmayr: `PanelContent` needs a `Plugin` instance to show a custom loading overlay, in case the panel plugin defines a custom overlay (for example the GaugeChart had a circle skeleton, and the other plugins have the default rectangular skeleton). Except for that, `PanelContent` doesn't need the `Plugin` instance.

`PanelPluginLoader` also loads the plugin, because it should be used standalone (`<PanelPluginLoader kind="StatChart" queryResults={[...]}/>`).

I think it should not matter performance-wise, because `usePlugin()` first gets a value from the context, and then runs a `@tanstack/query` with a `queryKey` set. So the second call should get the cached result.

I don't see a way how we could combine them, except for dropping support for a custom loading overlay (afaics only the GaugeChart used a custom overlay, I'll have to update the plugin in the `perses/plugins` repo to restore the custom loading overlay).

andreasgerstmayr: > This call usePlugin is only useful for showing custom plugin loader, but no plugin are using it currently, should we keep this feature? 🤔

It was used in the `GaugeChart`, I'll need to update the `GaugeChart` in the `plugins` repo to reinstate this custom loading overlay. But we could also drop it, not sure how important a custom loading screen is.

Gladorme (approved):
LGTM, let's keep the custom loader feature

ui/dashboards/src/components/Panel/PanelContent.tsx
andreasgerstmayr: For reference, this is the PR in the plugins repo to re-add the circular loading screen: https://github.com/perses/plugins/pull/68.

PR #2665, opened 2025-02-19
[BREAKINGCHANGE] Rename field in the service discovery config and add a concept doc about SD

Nexucis:


docs/datasource-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+    kubernetes_sd:
+      datasource_plugin_kind: "PrometheusDatasource"
+      namespace: "my-namespace"
+      service_configuration:
+        enable: true
+        port_name: "http"
+        serviceType: "ClusterIP"
+      labels:
+        app: prometheus
+```
```
ibakshay, line 64: This is a really cool feature. We should also add this config in the [values file of the perses/helm-charts](https://github.com/perses/helm-charts/blob/218d2a42a609eec237ae880a0aeca9b7a149dc06/charts/perses/values.yaml#L100), right?

Nexucis: yeah :)

PR #2680, opened 2025-02-24
[DOC] Add documentation about how to migrate from Grafana

Nexucis:


docs/migration.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+- Then, you can use the `migrate` command to migrate your dashboard. We recommend using the `--online` flag to be sure
+  that the migration is done with the latest version of the plugins.
+
+```bash
+percli migrate -f grafana-dashboard.json --online -o json > perses-dashboard.json
+```
+
+- As a tips, you may want to open the file and remove any reference to the previous datasource used in Grafana. This
```
jgbernalp, line 78:
```suggestion
- As a tip, you may want to open the file and remove any reference to the previous datasource used in Grafana. This
```

docs/migration.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Context
+
+Migrating from Grafana to Perses means to be able to translate the Grafana dashboards to Perses dashboards (in terms of
+data-model).
+
+*Note that we are not supporting any other Grafana resources like alerts, users, etc. We are only focusing on the
+dashboards.*
+
+The challenge around this process is actually to be able to translate the Grafana various plugin to the one supported by
```
AntoineThebaud, line 15:
```suggestion
The challenge around this process is actually to be able to translate the Grafana various plugins to the ones supported by
```

docs/migration.md
```diff
@@ -0,0 +1,127 @@
...
+
+```json
+{
+  "datasource": {
+    "kind": "PrometheusDatasource"
+  }
+}
+```
+
+- You should check the unsupported migrations. For example for the variable, you will have a static variable like this:
```
AntoineThebaud, line 102:
```suggestion
- You should check the unsupported migrations. For example in the case of a variable, you will get a static variable like this:
```

AntoineThebaud (approved):

docs/migration.md
```diff
@@ -0,0 +1,127 @@
...
+
+- Then, you can use the `migrate` command to migrate your dashboard. We recommend using the `--online` flag to be sure
+  that the migration is done with the latest version of the plugins.
+
+```bash
+percli migrate -f grafana-dashboard.json --online -o json > perses-dashboard.json
+```
+
+- As a tip, you may want to open the file and remove any reference to the previous datasource used in Grafana. This
+  will allow the dashboard to use the default datasource. Do that only if you want to use the default datasource.
```
ibakshay, line 79: @Nexucis Does it make sense to introduce a flag `--use-default-datasource=true` and if this flag is set then we don't add the datasource config at all in the dashboard manifest?

Nexucis: That looks difficult to me to achieve that. The usage of the datasource is handled by Cuelang and by the cuelang migration script (so basically on every plugin).
Handling the Cuelang script was already something quite hard and it's executed more or less as a close box. And I'll  like to keep it like that.

Issue #1984, opened 2024-05-08
DaC with CUE: rely on native dependency management


AntoineThebaud:
### Proposal

Rework the DaC CUE SDK to rely on the new dependency management now natively supported in CUE since [v0.8.0](https://github.com/cue-lang/cue/releases/tag/v0.8.0) (still experimental for now). More precisely this involves:
- Removal of the  `dac setup` command, as it was a workaround solution in absence of a proper dep management. In the end we should `cue get` the dependencies just like we `go get` them when using the Go SDK.
- Update of the user doc.

AntoineThebaud:
Some news about this activity: CUE's new module implementation (that enables proper dependency management) is now the default one since v0.9. However it seems things wont be as simple as running a command like `cue get github.com/perses/perses` just like we are doing `go get github.com/perses/perses` with the Go SDK, we'll have to publish our schemas to an OCI registry (hopefully they'd be accepted in the Central registry that CUE org aims to provide). 
About this topic see https://github.com/cue-lang/cue/wiki/FAQ:-Modules-and-v0.9#what-is-the-central-registry

AntoineThebaud:
Issue rescoped to consider the move to CUE's dependency management for both the plugin datamodel side & DaC SDK side at the same time since it's quite intricated.

Ongoing work on [feature/remove-plugin-schemas](https://github.com/perses/perses/tree/feature/remove-plugin-schemas) branch here and also on the [feature/rely-on-cue-dep-mgt](https://github.com/perses/plugins/tree/feature/rely-on-cue-dep-mgt) branch on perses/plugins side.

Things to be done:
- perses/perses: 
  - [x] new module + rely on CUE's dep mgt in test files + remove plugin-specific schemas -> done with https://github.com/perses/perses/pull/2645
  - [x] remove `percli plugin update` -> done with https://github.com/perses/perses/pull/2660
  - [ ] rework `percli plugin build` command as explained in https://github.com/perses/plugins/pull/55
  - [ ] change back `/cuelang` to `/cue` in the module name when possible
  - [ ] publish schemas to central registry as part of CI (= integrate this as part of our release process)
  - [ ] rework user docs (breaking change in import paths etc, redirect to CUE dep mgt doc...)
- perses/plugin: 
  - [x] Get rid of the usage of `percli plugin update` command and rely instead on CUE's native dependency management -> done with https://github.com/perses/plugins/pull/55
[...]

PR #2695, opened 2025-03-01
[ENHANCEMENT] rework `percli plugin build` to vendor CUE dependencies

AntoineThebaud:
# Description

Add a step of dependencies vendoring to the `percli plugin build` command.

Why: Following the move to CUE native dependency management, the deps of the plugin are no longer vendored under `cue.mod/pkg`. Letting things in this state would imply that, at runtime, the Perses server should have access to an OCI registry - thus either access to internet OR to a custom registry set up by users in their private network - to be able to evaluate a plugin schema that imports packages - like the `common` package of Perses.
As we consider this constraint a no-go in our case, We looked for solutions to that problem... After asking to CUE
maintainers it seems our long term solution would be this native vendor command, yet to be implemented. https://github.com/cue-lang/cue/issues/3328.
Thus for the time being we have to reproduce the vendoring structure of the "[old-modules](https://cuelang.org/docs/concept/faq/new-modules-vs-old-modules/#what-do-you-mean-by-old-and-new-modules)" with custom code.

Ref https://github.com/perses/perses/issues/1984

# Checklist


internal/cli/cmd/plugin/build/build.go
```diff
@@ -81,22 +95,169 @@ func (o *option) Execute() error {
...
+		// Vendor the dependencies
+		if err := o.vendorCueDependencies(); err != nil {
+			return fmt.Errorf("failed to vendor CUE dependencies: %w", err)
+		}
+		// restore the original state on exit
+		defer func() {
+			if err := os.RemoveAll(vendorDir); err != nil {
+				logrus.Printf("failed to cleanup %s: %v", vendorDir, err)
+			}
+		}()
```
AntoineThebaud, line 121: Technically a plugin could mix both old dependency management & new, which means that in this case this piece of code will NOT restore the original state. People would still be able to manually `git checkout --` afterwards though. IMHO this is a sufficiently acceptable impediment + corner case anyway to not go through the hassle of making the logic smarter here (would require to backup the whole cue.mod/pkg folder & restore it at the end). WDYT?

Nexucis: perhaps we could simply check if the folder `cue.mode/pkg` already exists ?

A more complex but not horrible solution would be:
- verify with git there is no file that are not tracked
- run the build command
- remove the files that git returned as not tracked in `cue.mod/pkg`, so the build command doesn't seem to alter the system.

The second proposal wouldn't be that horrible from my opinion. What do you prefer ?

internal/cli/cmd/plugin/build/build.go
```diff
@@ -126,7 +287,7 @@ func (o *option) computeArchiveFiles() ([]archives.FileInfo, error) {
 		return nil, fmt.Errorf("unable to read 'schema' directory: %w", err)
 	}
 	for _, f := range cueFiles {
-		list[filepath.Join(o.cfg.SchemasPath, f.Name())] = path.Join("schemas", f.Name())
+		list[filepath.Join(o.cfg.SchemasPath, f.Name())] = filepath.Join("schemas", f.Name())
```
Nexucis, line 290: I have intentionally let the usage of the pkg `path` here because the value of the map is what will be created in the archive. And I am pretty sure, the path shouldn't be OS relative but the regular `/schemas/cue.mod` one

Nexucis: excepting if you found an issue when creating the archive, I believe you should let it like that

AntoineThebaud: Ok indeed I thought that was a miss, I'll check if this is not causing any issue

internal/cli/cmd/plugin/build/build.go
AntoineThebaud: Let's go for first option as agreed to not be opinionated on VCS usage

internal/cli/cmd/plugin/build/build.go
AntoineThebaud: seems fine, reverted

internal/cli/cmd/plugin/build/build.go
```diff
@@ -22,16 +22,33 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
+const (
+	moduleFileBackup = "module.cue.bak"
+	vendorDirBackup  = "pkg.bak"
+)
+
+var (
+	vendorDir  = filepath.Join("cue.mod", "pkg")
```
Nexucis, line 48: since my last PR #2700, there is a constant `plugin.CuelangModuleFolder` that represent `cue.mod`

internal/cli/cmd/plugin/build/build.go
```diff
@@ -81,22 +98,185 @@ func (o *option) Execute() error {
 		return fmt.Errorf("unable to read manifest: %w", err)
 	}
 
-	// Then we need to create the archive.
-	// The archive contains the following file:
+	// Check if we need to vendor the dependencies:
+	// - Parse the module file into a CUE instance
+	ctx := cuecontext.New()
```
Nexucis, line 103: I think this change should go to a dedicated function. It will isolate it from the rest of the code.

Regarding the defer function, I would say the code can return the function to be defer. And then, in the function `Execute` you can call the defer function returned in the `defer` state.

Nexucis: Also, you will see with the changes from #2700 that you can now check if `schema` are required.

So all of that is under the condition that schemas are required. You should add this check at the beginning.

AntoineThebaud: Actually I wanted to encapsulate that in a dedicated function but was kinda stuck because of the defer, I didnt think about this trick thanks!

internal/cli/cmd/plugin/build/build.go
```diff
@@ -94,12 +108,94 @@ func (o *option) Execute() error {
 		return fmt.Errorf("unable to read manifest: %w", err)
 	}
 
-	// Then we need to create the archive.
-	// The archive contains the following file:
+	moduleFilePath := filepath.Join(o.pluginPath, plugin.CuelangModuleFolder, moduleFile)
+	moduleFileBackupPath := filepath.Join(o.pluginPath, moduleFileBackup)
+	vendorDirPath := filepath.Join(o.pluginPath, plugin.CuelangModuleFolder, vendorDir)
+	vendorDirBackupPath := filepath.Join(o.pluginPath, vendorDirBackup)
```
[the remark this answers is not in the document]
Nexucis: depends if you need these values in other functions than the one involved for cue dependencies, otherwise I don't think it's necessary

internal/cli/file/file.go
```diff
@@ -118,6 +119,43 @@ func visit(dir string) ([]string, error) {
...
+
+		if info.IsDir() {
+			return os.MkdirAll(targetPath, os.ModePerm)
+		}
+		return CopyFile(path, targetPath)
+	})
+}
+
+// Copy a file
+func CopyFile(src, dst string) error {
```
Nexucis, line 142: instead of `CopyFile` it can simply be `Copy`. So when used `file.Copy` you can understand you are copying a file

Nexucis (approved):
This is pure gold ! Well done @AntoineThebaud !

PR #2700, opened 2025-03-03
[ENHANCEMENT] Fix various issues in the CLI command plugin build and lint

Nexucis:
this PR is adding unit tests for the CLI command `plugin build` and `plugin lint`.

As a side effect, I have fixed various issues in both command such as:
- in the command `build`:
   - checking if the required files exist can only be done after building the frontend
   - the command `npm` must be ran into the frontend folder
   - in case the folder `node_modules` does not exist, the command will install de dependencies by running the command `npm ci`
   - the `schemas` folder set in the `package.json` is used now when building the archive so it can be found back by the backend.
- `cue.mod` folder shouldn't be checked if the Cuelang schemas are not required (depends of the plugin type)
- command lint cannot verify if the manifest exists as it required to build the plugin for that.

It also introduced a new flag `plugin.path` to be able to run the command outside of the plugin repository / folder.

### DIFF
--- a/internal/cli/cmd/plugin/build/build_test.go
+++ b/internal/cli/cmd/plugin/build/build_test.go
@@ -1,0 +1,33 @@
+// Copyright 2025 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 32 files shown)


internal/cli/cmd/plugin/build/build_test.go
```diff
@@ -0,0 +1,33 @@
...
+	cmdTest "github.com/perses/perses/internal/cli/test"
+)
+
+func TestPluginBuildCMD(t *testing.T) {
+	testSuite := []cmdTest.Suite{
+		{
+			Title:           "Build plugin",
+			Args:            []string{"--plugin.path", filepath.Join("testdata", "barchart")},
+			IsErrorExpected: false,
+			ExpectedMessage: "Building BarChart successfully\n",
```
AntoineThebaud, line 29: Btw could you change Building in the output by Built?

AntoineThebaud (approved):
🚀

PR #2701, opened 2025-03-03
[IGNORE] run all tests on windows

Nexucis:
It's been a while there are issues raised time to time on windows. I think it's time to run the tests to see if Perses is really working on windows


AntoineThebaud (approved):
Much love from a windaube user 🙃

Issue #2711, opened 2025-03-05
Datasource autodiscovery not working as expected


ibakshay:
### What did you do?

1. Deploy Perses with the following datasource discovery configuration in ConfigMap:
```yaml
datasource:
  global:
    discovery:
      - name: "prometheus-k8s-discovery"
        kubernetes_sd:
          datasource_plugin_kind: "PrometheusDatasource"
          namespace: "kube-monitoring"
          service_configuration:
            enable: true
            port_name: "http-web"
            service_type: "ClusterIP"
          labels:
            app: prometheus-prometheus
```

2. Start the Perses pod.
3. Observe the pod logs, which show the fatal error preventing startup.

### What did you expect to see?

Perses should successfully register discovered Prometheus services as global datasources.
- The pod should start without fatal errors related to datasource discovery tasks.

### What did you see instead? Under which circumstances?

The pod fails to start and logs indicate that the task is not recognized as a valid SimpleTask or Task.
```
time="2025-03-05T16:01:17Z" level=fatal msg="unable to instantiate the tasks for datasource discovery: task is not a SimpleTask or a Task"
```

### System information

local kind cluster

### Perses version

persesdev/perses:main-2025-03-04-682a13f7-distroless-debug

### Perses configuration file

```yaml
api_prefix: /perses
    datasource:
      global:
        discovery:
          - name: "prometheus-k8s-discovery"
            kubernetes_sd:
              datasource_plugin_kind: "PrometheusDatasource"
              namespace: "kube-monitoring"
              service_configuration:
                enable: true
                port_name: "http-web"  # Matches the Prometheus service port name
                service_type: "ClusterIP"  # Fixed field name
              labels:
                app: prometheus-prometheus
    database:
      file:
        extension: json
        folder: /perses
    frontend:
      important_dashboards: ]
    provisioning:
      folders:
      - /etc/perses/provisioning
      interval: 1m
    schemas:
      datasources_path: /etc/perses/cue/schemas/datasources
      interval: 5m
      panels_path: /etc/perses/cue/schemas/panels
      queries_path: /etc/perses/cue/schemas/queries
      variables_path: /etc/perses/cue/schemas/variables
    security:
      cookie:
        same_site: lax
        secure: false
      enable_auth: false
      readonly: false
```

### Logs

```text
time="2025-03-05T16:01:17Z" level=fatal msg="unable to instantiate the tasks for datasource discovery: task is not a SimpleTask or a Task"
```

ibakshay:
[...]

PR #2712, opened 2025-03-05
[ignore] move datasource discovery test to the new plugin system

Nexucis:
that should fix the windows tests as a side effect.


AntoineThebaud (approved):

ibakshay:
> that should fix the windows tests as a side effect.

Thank you. You are so fast!🚀🚀 I will test again tomorrow 😄

Nexucis:
@ibakshay that's not this PR that is fixing #2711 it's #2714 ;)

PR #2721, opened 2025-03-06
[DOC] Refactor and improve the doc about how to load a plugin

Nexucis:
I have split the initial doc about how a plugin is loaded. I think it makes more sense like that.

### DIFF
--- a/docs/configuration/load-plugin.md
+++ b/docs/configuration/load-plugin.md
@@ -1,0 +1,50 @@
+How to load a plugin
+====================
+
+Before being able to use a plugin, it must be recognized and loaded by the Perses backend.
+
+When the Perses backend is starting, it will look at a specific folder that should contain any plugin archive file. This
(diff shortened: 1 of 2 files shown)


docs/configuration/load-plugin.md
```diff
@@ -0,0 +1,50 @@
...
+
+Perses will extract every archive contained in this folder and will put the data into another folder. This folder can be
+set using the following configuration:
+
+```yaml
+plugin:
+  path: /path/to/plugin/folder
+```
+
+By default, the plugin folder is set to `plugins` or to `/etc/perses/plugins` if it's running in a container.
```
ibakshay, line 27: Do you think it makes to add a line when Perses runs in a k8s pod? 
@Nexucis

Nexucis: well in a k8s pod, Perses is still running in a container, so that should be fine I guess

PR #2722, opened 2025-03-06
[BUGFIX] Table: fix reorder columns button tooltip

andreasgerstmayr:
# Description

The children component of a `<Tooltip>` must accept a ref and spread the passed props:
https://mui.com/material-ui/react-tooltip/#custom-child-element

Fixes the following error in the JS console when editing a Table Panel and opening the "Column Settings" tab:
```
ColumnEditorContainer.tsx:91 Warning: Failed prop type: Invalid prop `children` supplied to `ForwardRef(Tooltip)`. Expected an element that can hold a ref. Did you accidentally use a plain function component for an element instead? For more information see https://mui.com/r/caveat-with-refs-guide
    at Tooltip (http://localhost:3000/plugins/Table/static/js/async/vendors-node_modules_mui_material_Button_Button_js-node_modules_mui_material_Divider_Divider_-a0c33b.js:14539:97)
    at ColumnEditorContainer (http://localhost:3000/plugins/Table/static/js/async/__federation_expose_Table.js:1387:11)
```

# Screenshots

no UI changes

# Checklist

## UI Changes

  include:


Nexucis:
@AntoineThebaud the CI is failing here because (I think) when the PR is coming from a fork, github-action does not inject any secret (for security matter)

Nexucis:
Perhaps that's a use case that should be shared with the Cuelang team. We cannot import the Cuelang dependencies in a CI/CD as it require a token to get them and when the PR is coming from a fork, the token cannot be used.

AntoineThebaud:
> Perhaps that's a use case that should be shared with the Cuelang team. We cannot import the Cuelang dependencies in a CI/CD as it require a token to get them and when the PR is coming from a fork, the token cannot be used.

I [asked](https://cuelang.slack.com/archives/C063DSMKS56/p1741345131840009) on CUE's slack about this, let's see..

AntoineThebaud:
> > Perhaps that's a use case that should be shared with the Cuelang team. We cannot import the Cuelang dependencies in a CI/CD as it require a token to get them and when the PR is coming from a fork, the token cannot be used.
> 
> I [asked](https://cuelang.slack.com/archives/C063DSMKS56/p1741345131840009) on CUE's slack about this, let's see..

This is what they replied:

_We plan on supporting [OIDC Connect](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect) in Github Actions, which I think would solve this use case. For now, I'm afraid that people will probably have to provide their own secrets, otherwise there's a clear security risk: anyone contributing to a fork could steal the token because they control the action script._

-> I wonder if there is a way to avoid requesting every contributor to set up their own token. Like, could we isolate these tests in a dedicated job workflow/job that would somehow require a maintainer's approval to be triggered?

Nexucis (approved):
otherwise, once the CI is green, approved !

andreasgerstmayr:
> @andreasgerstmayr can you update your PR with `main` to see if the CI is green again please ?

done
CI is green now, thanks!

I pressed the merge button :)

PR #2723, opened 2025-03-06
[BUGFIX] Discovery: Fix label selector builder and empty value in cuelang

Nexucis:
This PR aims to tackle the various issues raised in https://github.com/perses/perses/issues/2711.

I am adding more unit tests also to covering more use case.

One mistake I did when writing the discovery was to think an empty string and a nil value is the same for a string in Cuelang. And obviously it's not the case.

So I think the safest way to avoid the case raised in the last comment of the PR is to avoid to set a concrete value if the string is empty.


internal/api/discovery/kubernetes/kubernetes_test.go
ibakshay, line 1: I love the all the tests that you have added! 🚀

AntoineThebaud (approved):

PR #2724, opened 2025-03-07
[BREAKINGCHANGE] `percli dac setup`: rely on CUE's new modules for deps installation

AntoineThebaud:
# Description

With the move to CUE dependency management, it is no longer required to manage ourselves the installation of perses SDK (via release archive download + extract etc...), we can now rely on `cue mod tidy` for that - also valid for plugins that now each have an independent module and no longer come with perses/perses repo.

The doc got reworked accordingly. Btw the Go SDK part was not up to date (percli dac setup was totally ignored there 😅 ) so I adjusted this part too.

Also added some tests (still quite some cases not tested though..).

Ref https://github.com/perses/perses/issues/1984.

# Checklist


docs/dac/getting-started.md
```diff
@@ -7,49 +7,70 @@
...
 
 - The [official website](https://cuelang.org/) of Cuelang.
 - [Cuetorials](https://cuetorials.com/), a 3rd party source of information that is a very good complement.
 
-You should then have a look at the [CUE SDK documentation](../dac/cue/README.md) to better understand how to use the framework.
+You should then have a look at the [CUE SDK documentation](../dac/cue/README.md) to better understand how to use it.
 
 You can also check an example of DaC usage [here](https://github.com/perses/perses/blob/main/internal/test/dac/input.cue).
 
+For each [plugin](../concepts/plugins.md) you would like to use in your DaC, it is strongly recommended to import its module so that you benefit from its schema validation locally. Check the list of available plugins & their corresponding module name at https://github.com/perses/plugins.
```
AntoineThebaud, line 56: _"Check the list of available plugins & their corresponding module name at https://github.com/perses/plugins"_ to be improved later when plugin docs will be also be published to the documentation website

docs/dac/getting-started.md
```diff
@@ -91,6 +113,11 @@ You can also check an example of DaC usage [here](https://github.com/perses/pers
 !!! warning
 	Do not log / print on the standard stdout! It would break the output of the `dac build` command.
 
+For each [plugin](../concepts/plugins.md) you would like to use in your DaC, you will have to import the corresponding dependency in order to be able to use the piece of SDK they provide. Check the list of available plugins & their corresponding module name at https://github.com/perses/plugins.
```
AntoineThebaud, line 116: NB: same as above

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -23,17 +23,23 @@ import (
...
 
 const (
-	cueLanguage    = "cue"
-	goLanguage     = "go"
-	cueSchemasPath = "cue/"
+	cueLanguage      = "cue"
+	goLanguage       = "go"
+	cueSchemasPath   = "cue/"
+	minVersionForGo  = "v0.44.0" // Release that introduced the Go SDK
+	minVersionForCue = "v0.47.0" // TODO should be changed to v0.51.0 once released (Release that brought the move to CUE's new modules)
```
Nexucis, line 37: I think you can already set `v0.51.0-beta.0`.

Because with the changes proposed, if I want to use cuelang, it won't be possible to use the v0.47.0 as it's not available from the registry

internal/cli/cmd/dac/setup/setup.go
```diff
@@ -128,23 +169,42 @@ func (o *option) Execute() error {
...
+			return err
+		}
+		// Resolve the sdk dependency (= write an example file
+		// and run 'cue mod tidy')
+		exampleFilePath := "example.cue"
+		if err := os.WriteFile(exampleFilePath, []byte(exampleCUEDac), 0600); err != nil {
+			return fmt.Errorf("failed to create example.cue: %w", err)
+		}
+		logrus.Debugf("example.cue file created successfully")
+		cmd := exec.Command("cue", "mod", "tidy")
```
Nexucis, line 185: I hope we won't need anymore to use a token to get the dependencies

internal/cli/cmd/dac/setup/setup.go
AntoineThebaud: Yes but this is actually to have the test passing, since v0.51.0-beta.0 doesnt exist yet :D So either we go with this "lie" for now, or we release a beta without this PR and it would be part of the next pre-release. WDYT?

internal/cli/cmd/dac/setup/setup.go
AntoineThebaud: Well yes you have to be logged in everytime you need to resolve dependencies.. AFAIS when working with dac it's not that annoying, the authentication persist for quite a long time.

internal/cli/cmd/dac/setup/setup.go
Nexucis: yeah maybe let's wait the release before merging this PR. That would avoid an issue if someone want to try it already

Nexucis: should be good to update the version now

internal/cli/cmd/dac/setup/testdata/.gitignore
```diff
@@ -0,0 +1,3 @@
+go.sum
```
Nexucis, line 1: any reason to ignore the file `go.sum` ?

Nexucis (approved):
lgtm otherwise

internal/cli/cmd/dac/setup/testdata/.gitignore
AntoineThebaud: to me we don't care of tracking such file in the case of test data, but I dont mind adding it back

PR #2748, opened 2025-03-10
[BUGFIX] CLI: plugin build should consider all subfolder when getting the cue dependencies

Nexucis:
that should fix the issue found in https://github.com/perses/plugins/pull/75.

I would like to add a test to cover the case when the folder `schemas` contained multiple sub-folder and nothing at the root like it is the case for the Prometheus or Tempo plugin.

### DIFF
--- a/internal/cli/cmd/plugin/build/build.go
+++ b/internal/cli/cmd/plugin/build/build.go
@@ -53,6 +53,7 @@ type option struct {
 	skipNPMInstall            bool
 	archiveFormat             archive.Format
 	cfg                       config.PluginConfig
+	initialCFG                config.PluginConfig
 	cfgPath                   string
 	pluginPath                string
 	isSchemaRequired          bool
@@ -74,6 +75,7 @@ func (o *option) Complete(args []string) error {
 		return fmt.Errorf("unable to resolve the configuration: %w", err)
 	}
 	o.cfg = cfg
+	o.initialCFG = cfg
 	// Overriding the path with the plugin path
 	o.cfg.DistPath = filepath.Join(o.pluginPath, o.cfg.DistPath)
 	o.cfg.FrontendPath = filepath.Join(o.pluginPath, o.cfg.FrontendPath)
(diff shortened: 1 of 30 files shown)


internal/cli/cmd/plugin/build/build.go
```diff
@@ -214,21 +218,21 @@ func (o *option) prepareCueFiles() (func(), error) {
 	// managed the old way. This is technically possible & supported by CUE, thus we want to handle this case.
 	var restoreVendorDir func()
 	if _, err := os.Stat(o.vendorDirPath); err == nil {
-		if err := file.CopyDir(o.vendorDirPath, o.vendorDirBackupPath); err != nil {
-			return nil, fmt.Errorf("failed to backup original %s: %w", o.vendorDirPath, err)
+		if copyErr := file.CopyDir(o.vendorDirPath, o.vendorDirBackupPath); copyErr != nil {
```
Nexucis, line 221: these renaming of the error variable are here to avoid the shadowing variable issue

internal/cli/cmd/plugin/build/build.go
```diff
@@ -120,11 +122,13 @@ func (o *option) Execute() error {
 
 	// If the schema is required, we need to prepare the CUE files for the archive
 	if o.isSchemaRequired {
-		if restoreInitialState, err := o.prepareCueFiles(); err != nil {
-			return fmt.Errorf("unable to prepare the CUE files for the archive: %w", err)
-		} else if restoreInitialState != nil {
+		restoreInitialState, cueErr := o.prepareCueFiles()
```
Nexucis, line 125: this change is convering the fact even if there is an issue when vendoring the dependencies, we are still removing any backup folder created.

internal/cli/cmd/plugin/build/build.go
AntoineThebaud: I don't think the shadowing causes an issue here though?

Nexucis: no it doesn't, for this part it's more a good practice to avoid the shadowing variable. It avoid any confusion while reading the code.

PR #2752, opened 2025-03-11
[ENHANCEMENT] Use cue mod tidy to download cue deps

Nexucis:


AntoineThebaud (approved):
I confirm it works well like this!

PR #2757, opened 2025-03-12
Revert "[BREAKINGCHANGE] `percli dac setup`: rely on CUE's new modules for deps installation"

Nexucis:
Reverts perses/perses#2724

It looks like somehow this change breaks the e2e tests


Nexucis:
or maybe it doesn't which is even stranger ....

Nexucis:
ok closing this one, the e2e tests are flaky ....

PR #2765, opened 2025-03-13
[IGNORE] Fix dev dashboard with new prom demo url

Nexucis:
let's see if it's enough to stabilise the e2e tests

### DIFF
--- a/dev/data/9-dashboard.json
+++ b/dev/data/9-dashboard.json
@@ -272,7 +272,7 @@
                   "plugin": {
                     "kind": "PrometheusTimeSeriesQuery",
                     "spec": {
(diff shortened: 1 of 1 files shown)


Nexucis:
weell not at all ...

ddanielcruzz:
I noticed some flakiness with [this part of e2e tests](https://github.com/perses/perses/blob/main/ui/e2e/src/pages/DashboardPage.ts#L298-L313) which seems unrelated to Prometheus. I had Playwright UI open and I could see the edit button but Playwright couldn't locate/click it.

Nexucis:
yeah I came to the same conclusion than you @ddanielcruzz.

I am also wondering if the PR #2770 makes the tests less flaky

PR #2770, opened 2025-03-13
[ENHANCEMENT] Improve dashboard performance by reducing re-renders

andreasgerstmayr:
# Description

Reduce unnecessary re-renders of `<DashboardApp>` component.
`DashboardApp` calls the `useDashboard()` hook, which calls `useVariableDefinitionActions()`. This function uses the store and returns **a new object every time**, invalidating the hook. By using `useShallow()`, it checks if the object actually changed or not.

However, in my opinion, it'd be best to remove this function and use `useStore(store, (s) => s.setVariableDefinitions)` instead when accessing `setVariableDefinitions` for example. What do you think?

I also updated `VariableProvider` and `VariableProviderWithQueryParams` to not create a new store every time for the initial value.

# Screenshots

Screenshots from the React Profiler while loading the node exporter dashboard:

Before: The component was rendered 9 times.
Bildschirmfoto vom 2025-03-13 19-16-13

After: The component is rendered 3 times.
Bildschirmfoto vom 2025-03-13 19-17-35

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/context/VariableProvider/VariableProvider.tsx
+++ b/ui/dashboards/src/context/VariableProvider/VariableProvider.tsx
@@ -37,6 +37,7 @@ import {
   TextVariableDefinition,
   ListVariableDefinition,
 } from '@perses-dev/core';
+import { useShallow } from 'zustand/react/shallow';
 import { checkSavedDefaultVariableStatus, findVariableDefinitionByName, mergeVariableDefinitions } from './utils';
 import { hydrateVariableDefinitionStates as hydrateVariableDefinitionStates } from './hydrationUtils';
 import { getInitalValuesFromQueryParameters, getURLQueryParamName, useVariableQueryParams } from './query-params';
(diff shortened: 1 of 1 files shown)


Nexucis:
> However, in my opinion, it'd be best to remove this function and use useStore(store, (s) => s.setVariableDefinitions) instead when accessing setVariableDefinitions for example. What do you think?

I don't have much an opinion about that part. Perhaps re-creating from scratch can help to reinitialise the dashboard ?

Otherwise thanks a lot for finding this optimisation @andreasgerstmayr ! I have wonder if it helps to make the e2e tests less flaky

andreasgerstmayr:
> > However, in my opinion, it'd be best to remove this function and use useStore(store, (s) => s.setVariableDefinitions) instead when accessing setVariableDefinitions for example. What do you think?
> 
> I don't have much an opinion about that part. Perhaps re-creating from scratch can help to reinitialise the dashboard ?

As far as I see it, `useVariableDefinitionActions()` is a utility function to write for example
```
const { getSavedVariablesStatus, setVariableDefaultValues } = useVariableDefinitionActions();
```
instead of
```
const store = useVariableDefinitionStoreCtx();
const getSavedVariablesStatus = useStore(store, (s) => s.getSavedVariablesStatus);
const setVariableDefaultValues = useStore(store, (s) => s.setVariableDefaultValues);
```

so it shouldn't affect any actual UI functionality either way.

We already use `useStoreWithEqualityFn()` in other places, so I updated the code to use the same.

> Otherwise thanks a lot for finding this optimisation @andreasgerstmayr ! I have wonder if it helps to make the e2e tests less flaky

:crossed_fingers:

Nexucis (approved):
awesome changes @andreasgerstmayr thank you 🙏

PR #2795, opened 2025-03-18
[ENHANCEMENT] CLI/plugin: Improve the way to vendor cue dependencies and rely on cue default caching

Nexucis:
This PR is changing the way we are vendoring the Cuelang dependency.

Before we were setting a temporary folder that was used as cache folder to download the Cuelang dependency.

Now we are not altering the default caching folder of Cuelang. The code is looking at the dependency file and then based on them, it will copy one by one the dependency from the caching folder to the vendor folder.

If one dependency is not found, then the code is calling cuelang to download it.

### DIFF
--- a/internal/cli/cmd/plugin/build/build.go
+++ b/internal/cli/cmd/plugin/build/build.go
@@ -14,37 +14,21 @@
 package build
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
-	"strings"
 
-	"cuelang.org/go/cue"
-	"cuelang.org/go/cue/cuecontext"
-	"cuelang.org/go/cue/format"
-	"cuelang.org/go/cue/load"
 	"github.com/mholt/archives"
 	"github.com/perses/perses/internal/api/archive"
 	"github.com/perses/perses/internal/api/plugin"
 	persesCMD "github.com/perses/perses/internal/cli/cmd"
 	"github.com/perses/perses/internal/cli/cmd/plugin/config"
-	"github.com/perses/perses/internal/cli/file"
 	"github.com/perses/perses/internal/cli/output"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-)
-
-const (
-	vendorDir        = "pkg"
-	vendorDirBackup  = vendorDir + ".bak"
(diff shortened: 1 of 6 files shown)


Nexucis:
this change should help to solve the issue raised in the PR https://github.com/perses/plugins/pull/90

internal/cli/cmd/plugin/build/cue.go
```diff
@@ -0,0 +1,259 @@
...
+// As we consider this constraint a no-go in our case, we asked to CUE maintainers.
+// It seems our long-term solution would be https://github.com/cue-lang/cue/issues/3328. For the time being, we thus have to
+// reproduce the vendoring structure of the "old modules" (https://cuelang.org/docs/concept/faq/new-modules-vs-old-modules/) with
+// custom code.
+//
+// More precisely, the steps are:
+// - Parse the module.cue file containing the dependencies
+// - if no `deps` are present, skip vendoring
+// - if there is already a vendor directory, skip vendoring
+// - Loop other the dependencies and for each:
```
AntoineThebaud, line 100:
```suggestion
// - Loop over the dependencies and for each:
```

internal/cli/cmd/plugin/build/build.go
```diff
@@ -148,242 +131,6 @@ func (o *option) Execute() error {
...
-	// - if no `deps` are present, skip vendoring
-	deps := module.LookupPath(cue.ParsePath("deps"))
-	if deps.Err() != nil {
-		logrus.Debugf("`deps` not found, no CUE dependencies to vendor")
-		return nil, nil
-	}
-
-	// Backup the original state of cue.mod/pkg if it exists*, and restore the original state on exit.
-	// *: even if a given plugin relies on new dependency management, it may also rely on other dependencies
-	// managed the old way. This is technically possible & supported by CUE, thus we want to handle this case.
```
AntoineThebaud, line 218: So looking at the new code we lost this capability to support such "hybrid" setup with your change. Given the situation it's an acceptable drawback imho, just wanted your ack that this is done on purpose?

AntoineThebaud (approved):

internal/cli/cmd/plugin/build/build.go
Nexucis: yeah I did it on purpose because I saw some issue with the hybrid mode, and I felt for the moment removing this possibility will simplify the situation and the issues that can come

PR #2797, opened 2025-03-19
[ENHANCEMENT] allow scoped datasource proxies to use global secrets

jgbernalp:
# Description

When using datasources with TLS, CAs can be defined globally to verify datasource backends. This PR allows proxies to use global secrets in addition to project scoped secrets

# Checklist


Nexucis:
Well I would foresee a security issue for this one. Because it means you can do the following scenario:
- Create a datasource at project level overriding the global one
- Use the global secret to be able to contact the datasource
- Potentially do write action on the datasource with no more restriction.

jgbernalp:
Yes you are right, maybe I need to use a GlobalDatasource instead or create the secrets when the datasource is created. (from the perses-operator side).

PR #2798, opened 2025-03-19
[ignore] bump go deps

Nexucis:
As always there are issues with the inner `go.mod`, so this PR should be able to upgrade the dependencies that dependabot is stuck to do it properly.

related to PR https://github.com/perses/perses/pull/2784 and https://github.com/perses/perses/pull/2785


nicolastakashi:
I'm wondering if eventually we want to use renovate as well as we're using on the chart.

Nexucis:
maybe renovate is clever than dependabot for this kind of update yeah.

/cc @celian-garcia

celian-garcia:
> maybe renovate is clever than dependabot for this kind of update yeah.
> 
> /cc @celian-garcia 
 
Renovate intelligence will be to be able to run go mod tidy as post command. Is that something that dependabot is doing?

Nexucis:
well here it's more to run go mod tidy on different inner `go.mod` files. And I didn't really check if dependabot is able to do it, but I doubt a bit. dependabot is really basic

PR #2801, opened 2025-03-20
[ENHANCEMENT] adding extra args support dac build

nicolastakashi:
# Description

This pull request introduces changes to the `internal/cli/cmd` package to support passing arguments to commands in the `build` subcommand. The most important changes include adding a new method to set arguments, updating the `option` struct to store these arguments, and modifying the command execution to include the arguments.

Enhancements to argument handling:

* [`internal/cli/cmd/cmd.go`](diffhunk://#diff-ff081a1614e8adc291f4c20213d31463eb3b63cb0ea04a80dfe3626458464973R31-R38): Added the `SetArgs` method to the `Option` interface to allow setting command arguments.

Updates to the `build` command:

* [`internal/cli/cmd/dac/build/build.go`](diffhunk://#diff-97a1ca64aa554e8fc0b8b0f14879390430207a930004b32489d5191fc381e0c4R47-L53): Updated the `option` struct to include an `args` field for storing command arguments.
* [`internal/cli/cmd/dac/build/build.go`](diffhunk://#diff-97a1ca64aa554e8fc0b8b0f14879390430207a930004b32489d5191fc381e0c4L119-R132): Modified the `processFile` method to include the arguments when executing `go run` and `cue eval` commands.
* [`internal/cli/cmd/dac/build/build.go`](diffhunk://#diff-97a1ca64aa554e8fc0b8b0f14879390430207a930004b32489d5191fc381e0c4R186-R189): Added the `SetArgs` method to the `option` struct to set the command arguments.
* [`internal/cli/cmd/dac/build/build.go`](diffhunk://#diff-97a1ca64aa554e8fc0b8b0f14879390430207a930004b32489d5191fc381e0c4R222-R224): Updated the command usage documentation to include an example of passing extra arguments to the `build` command.

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/internal/cli/cmd/dac/build/build.go
+++ b/internal/cli/cmd/dac/build/build.go
@@ -131,7 +131,7 @@ func (o *option) processFile(file string, extension string) error {
 		//      However, the cue code is (for now at least) not well packaged for such an external reuse.
 		//      See https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/eval.go#L87
 		// NB3: #nosec is needed here even if the user-fed parts of the command are sanitized upstream
-		cmd = exec.Command("cue", "eval", file, "--out", o.Output, "--concrete") // #nosec
+		cmd = exec.Command("cue", "eval", file, "--out", o.Output, "--concrete", strings.Join(o.args, " ")) // #nosec
 	} else {
 		return output.HandleString(o.writer, fmt.Sprintf("skipping %q because it is neither a `cue` or `go` file", file))
 	}
--- a/internal/cli/cmd/dac/build/build.go
+++ b/internal/cli/cmd/dac/build/build.go
@@ -44,13 +44,12 @@ type option struct {
 	opt.OutputOption
 	writer    io.Writer
 	errWriter io.Writer
+	args      []string
 	Mode      string
 }
 
 func (o *option) Complete(args []string) error {
-	if len(args) > 0 {
-		return fmt.Errorf("no args are supported by the command 'build'")
-	}
+	o.args = args
 	if outputErr := o.OutputOption.Complete(); outputErr != nil {
 		return outputErr
 	}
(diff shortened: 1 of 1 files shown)


Nexucis:
I don't understand why you need to modify the interface `option` to set the args. `args` are already available in the method `complete`

internal/cli/cmd/dac/build/build.go
```diff
@@ -218,6 +217,9 @@ percli dac build -d my_dashboards
 
 # build all the files under a given directory & deploy the resulting resources right away
 percli dac build -d my_dashboards && percli apply -d built
+
+# build a given file as JSON providing extra arguments to the Go program
+percli dac build -f main.go -ojson -- --arg1=value1 --arg2=value2
```
Nexucis, line 222: perhaps we can have extra check to verify the extra args are following this struct ?

nicolastakashi: @Nexucis sorry I'm afraid I didn't follow your idea, could you come again?

Nexucis: I mean, does it makes sense to check that the syntax to catch the extra args start with `--` and then followed by `--arg1=value` ?

nicolastakashi: I don't think so, because the structure of args will depend on the framework used to define flags on the rendered, which might be different between frameworks.

Wdyt @Nexucis ?

internal/cli/cmd/dac/build/build.go
```diff
@@ -124,7 +123,7 @@ func (o *option) processFile(file string, extension string) error {
 		// we must be in the closest directory to the file.
 		folder := filepath.Dir(file)
 		extractedFile := filepath.Base(file)
-		cmd = exec.Command("go", "run", extractedFile, "--output", o.Output) // #nosec
+		cmd = exec.Command("go", "run", extractedFile, "--output", o.Output, strings.Join(o.args, " ")) // #nosec
```
Nexucis, line 126: probably you can add the same changes for cuelang ?

nicolastakashi: Ups, missed this.

PR #2802, opened 2025-03-20
[ENHANCEMENT] make proxy authorization optional

jgbernalp:
# Description

This PR makes the authorization options for a secret optional. This will allow clients using the proxy to pass their own Authorization header in cases this Header is available from the client or is dynamic and cannot be set to a specific value.

# UI changes

updated the secrets table to include the "No Authorization" option

Screenshot 2025-03-20 at 10 59 02

updated the secrets form to include the "No Authorization" option

Screenshot 2025-03-20 at 10 59 10

# Checklist

### DIFF
--- a/pkg/model/api/v1/secret.go
+++ b/pkg/model/api/v1/secret.go
@@ -58,11 +58,8 @@ func (s *SecretSpec) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 func (s *SecretSpec) validate() error {
-	if s.BasicAuth == nil && s.Authorization == nil {
-		return fmt.Errorf("at most one of basicAuth and authorization must be configured")
-	}
-	if s.BasicAuth != nil && s.Authorization != nil {
-		return fmt.Errorf("basicAuth and authorization are mutually exclusive, use one of them")
+	if s.BasicAuth != nil && s.Authorization != nil && s.OAuth != nil {
+		return fmt.Errorf("basicAuth, authorization and oauth are mutually exclusive, use one of them")
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


pkg/model/api/v1/secret.go
```diff
@@ -58,9 +58,6 @@ func (s *SecretSpec) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 func (s *SecretSpec) validate() error {
-	if s.BasicAuth == nil && s.Authorization == nil {
-		return fmt.Errorf("at most one of basicAuth and authorization must be configured")
-	}
 	if s.BasicAuth != nil && s.Authorization != nil {
```
Nexucis, line 61: Can you modify this condition, oauth is missing in this check.

```suggestion
if s.BasicAuth != nil && s.Authorization != nil && s.OAuth != nil {
}
```

pkg/model/api/v1/secret/authorization.go
```diff
@@ -91,6 +91,9 @@ func (a *Authorization) GetCredentials() (string, error) {
 }
 
 func (a *Authorization) validate() error {
+	if a == nil {
```
Nexucis, line 94: Excepting if you face an issue, I don't see why it would be necessary. If `Authorization` is not set, then yaml or json shouldn't call it.

Nexucis (approved):
Yes that makes sense it's optional in the secret spec specially because you could also want to just set the TLS config and nothing else

LGTM minus the comments I raised above

pkg/model/api/v1/secret/authorization.go
jgbernalp: it was setting a default value for the token type

jgbernalp: checking this again, is not needed.

jgbernalp:
@Nexucis I missed the UI changes for this configuration, it should be all set now!

Nexucis (approved):
lgtm

PR #2803, opened 2025-03-20
[ENHANCEMENT] allow insecureSkipVerify to be optional

jgbernalp:
# Description

This PR makes `insecureSkipVerify` optional as it already defaults to false.

# Checklist

### DIFF
--- a/cue/model/api/v1/secret/tls_config_go_gen.cue
+++ b/cue/model/api/v1/secret/tls_config_go_gen.cue
@@ -7,16 +7,16 @@ package secret
 // PublicTLSConfig is the public struct of TLSConfig.
 // It's used when the API returns a response to a request
 #PublicTLSConfig: {
-	ca?:                #Hidden @go(CA)
-	cert?:              #Hidden @go(Cert)
-	key?:               #Hidden @go(Key)
-	caFile?:            string  @go(CAFile)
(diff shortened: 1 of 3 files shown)


Nexucis:
well since it's a boolean, it's already optional by default 🤔. And removing it in the json makes de boolean undefined in Javascript. That's quite annoying to have a boolean undefined don't you think ?

jgbernalp:
Is not optional by default because you have to include the value, for example in a secret, even if you are using the default.

```yaml
secret:
    tlsConfig:
      caFile: /ca/service-ca.crt
      insecureSkipVerify: false
```

From the frontend side, undefined is falsy so it will mean the same as are already checking the value with `!!field.value` like:
```typescript
<Controller
    control={form.control}
    name="spec.tlsConfig.insecureSkipVerify"
    render={({ field }) => (
      <FormControlLabel
        label="Insecure Skip Verify"
        control={
          <Switch
            {...field}
            checked={!!field.value} // <---
            readOnly={action === 'read'}
            onChange={(event) => {
              if (action === 'read') return; 
              field.onChange(event);
            }}
          />
        }
      />
    )}
  />
```

jgbernalp:
or maybe we just need the `omitempty`?

Nexucis (approved):
yeah ok, fair enough

PR #2810, opened 2025-03-21
[FEATURE] Add zoom in/zoom out buttons

lucrus73:
# Description

This PR implements a "Zoom out x2" menu item in the time range menu. See https://github.com/perses/perses/discussions/2773

# Screenshots

image


lucrus73:
@Nexucis can you please help me understand what's the problem with my code? Gofmt is complaining about a file I changed (namely pkg/model/api/config/frontend.go), but I can't see what I did wrong.

Gladorme:
Strange to see this "Zoom out" here, not really fan 😅 
Some questions in my head, that could maybe improve the feature:
Why x2 and not x0.5, x3, ...? Why no "Zoom In"? Maybe it need it's own button?  🤔

lucrus73:
> Strange to see this "Zoom out" here, not really fan 😅 Some questions in my head, that could maybe improve the feature: Why x2 and not x0.5, x3, ...? 

I borrowed the idea from Munin, which only has a "Zoom out x2" button and it does the job pretty well. I imagine the use case like when you're looking at some  historical interval and you'd like to see more around it, without having to edit the absolute time range. A simple "Zoom out x2" is easier and faster to understand than a "Zoom out x0.5" or "Zoom out x3", because it's just the double of the currently shown interval. 

> Why no "Zoom In"? 

Because there already exists an easy to use "Zoom In" action: just drag the mouse over the chart.
  
> Maybe it need it's own button? 

I think dragging the mouse is just what people expect.

> Why Prometheus / Grafana UIs did not implement it? 🤔

I don't know Grafana UI, even less I know why it did not implement it. I think we should do what we think is useful, regardless of what Grafana did or will do. Perses goals does not include being a Grafana UI clone.

jgbernalp:
I think it will be better to have both zoom in and zoom out options, I find the Prometheus + and - very clear and useful with their own dedicated buttons. WDYT?

lucrus73:
> I think it will be better to have both zoom in and zoom out options, I find the Prometheus + and - very clear and useful with their own dedicated buttons. WDYT?

I think this is the way to go. On my first try I did not consider this solution, because it all started from my Munin experience and the context of the discussion was the missing zoom out feature only, but now I agree the two buttons would be a way better solution.

I'm going to edit my code as soon as I find the time, but I'm leaving for a cruise on sunday, so I suspect I'll be AFK for a few days...

jgbernalp:
@lucrus73 I think it will be desirable that these buttons have the timerange in between so is clear thew can be used to zoom in our out time.

lucrus73:
> @lucrus73 I think it will be desirable that these buttons have the timerange in between so is clear thew can be used to zoom in our out time.

I think this wouldn't make for a great user experience, because the timerange changes its width based on the selected value, and having one zoom button at its left and the other at its right would effectively change the distance between them, forcing the user, just after having clicked the right one and maybe desiring to go back, to look for the place where the left one ended up. That would happen because the zoom buttons always change the selected timerange, and, consequently, they almost always change the width of the timerange control. Zooming out (assuming zoom out button would be the one on the right) and zooming back in would become a somewhat cumbersome gesture.

That being said, if you still think that's the better solution, I can easily edit my PR, because it's a trivial change. Please let me know.

AntoineThebaud:
> I think this wouldn't make for a great user experience, because the timerange control changes its width based on the selected value, and having one zoom button at its left and the other at its right would effectively change the distance between them, forcing the user, just after having clicked the right one and maybe desiring to go back, to look for the place where the left one ended up. That would happen because the zoom buttons always change the selected timerange, and, consequently, they almost always change the width of the timerange control. Zooming out (assuming zoom out button would be the one on the right) and zooming back in would become a somewhat cumbersome gesture.

Good point, so what about having something like this?

image

lucrus73:
> Good point, so what about having something like this?
> 

I had the same idea when I first started to code this patch and I still believe it would be the best solution (zoom-actions wise), but I didn't code my patch that way because I suspect that the reload and the refresh delay controls should even more rightfully be placed there for the same reasons, because the zoom in and out icons make their own meaning very clear regardless of where they are placed, while I find the round arrow is not so self explaining if it's not placed aside the timerange control... or maybe it's just my habit of seeing it there, I don't know.

If you think that the reload and refresh delay controls keep their clear meaning even when placed aside the zoom buttons, I'm more than willing to change the position of the zoom buttons the way you suggest.

jgbernalp:
I think @AntoineThebaud suggestion would work best based on all the arguments presented.

lucrus73:
> I think @AntoineThebaud suggestion would work best based on all the arguments presented.

Ok, I'm going to follow his suggestion then. 

Just a last question: would it be better to place the zoom in button (plus) on the left, and the zoom out button (minus) on the right, like depicted by @AntoineThebaud, or the other way around? Just asking because I think the zoom out would be better placed on the left, since the reading direction (left-to-right) makes users feel like things "go ahead" (or "grow") when going towards the right, and they "go back" (or "shrink") when going towards the left. I hope I made what I mean clear enough.

jgbernalp:
yeah I agree it feels more natural to have the zoom out on the left and zoom in on the right.

lucrus73:
Can you please help me understand what `make checkformat` is complaining about? I've already tried to indent that code in a few ways that seemed reasonable to me, but I can't get past this error.

jgbernalp:
It seems is just a matter of indenting using either spaces or tabs, depends on your IDE setup, but that might be the reason. You can try to run `make checkformat` locally to debug.

jgbernalp:
check: https://github.com/lucrus73/perses/pull/1

jgbernalp (approved):
Great addition, thnx!

PR #2859, opened 2025-04-09
[IGNORE] Add integration tests for SQL database

pvcnt:
# Description

While getting familiar the code, I saw that there were no tests for the SQL-based database. I am hence opening this PR with some basic tests (copied from the file-based database), to reduce the risk of a breaking change when modifying this layer. I am relying on [testcontainers](https://golang.testcontainers.org/modules/mysql/) to get a working database.

# Checklist


pvcnt:
I don't understand the failures.
- "tests with file DB on windows": test failing is seemingly unrelated to this change (comes from `internal/api/e2e/client/project_test.go`).
- "check code format": when I run `make checkunused` locally it does not modify any files.

pvcnt:
My bad, I was on the wrong PR. 🤦 I should have fixed the "check code format" check, still unsure about the other one.

celian-garcia:
Very interesting! I'm wondering if you couldn't make parametrized tests here in a database_test.go file upper, with one parameter by DB. and executing the tests for both DB each time. WDYT? It would prevent from future desync

pvcnt:
You're right, that would be ideal. I'm still new to Go development though, so I'm not sure if there is a recommended pattern to implement this? Or some example in the codebase? Here it's somewhat "complex" because the SQL-based DAO needs to be initialized only once before all tests but data is cleaned-up after every test. Initializing one per test would be too slow.

celian-garcia:
> You're right, that would be ideal. I'm still new to Go development though, so I'm not sure if there is a recommended pattern to implement this? Or some example in the codebase? Here it's somewhat "complex" because the SQL-based DAO needs to be initialized only once before all tests but data is cleaned-up after every test. Initializing one per test would be too slow.

The idea I'd have in mind would be to have database_test.go like this (sorry if it's  not syntactically correct I just wrote it in the comment directly)

```go

var sqlDAO = databadeModel.DAO
var filedAO = databaseModel.DAO

func TestMain(m *testing.M) {
  sqlDAO = sql.SetupTest()
  defer sql.TeardownTest()
  fileDAO = file.SetupTest()
  
  exitCode := m.Run()
  os.Exit(exitCode)
}

func removeAllResources(t *testing.T) {
	if err := dao.DeleteByQuery(&project.Query{}); err != nil {
		t.Fatal(err)
	}
}

func TestDAO_Create(t *testing.T) {
        testSuite := []struct {
		title  string
		dao  databaseModel.DAO
	}{
        for _, test := range testSuite {
		t.Run(test.title, func(t *testing.T) {
	              defer removeAllResources(t)
	              projectEntity := &modelV1.Project{
		              Kind: modelV1.KindProject,
		              Metadata: modelV1.Metadata{
			              Name: "perses",
		              },
	              }
	              assert.NoError(t, test.dao.Create(projectEntity))
	              assert.True(t, databaseModel.IsKeyConflict(dao.Create(projectEntity)))
	         }
	}
}

...

```

celian-garcia:
mmh this testcontainers library fails on windows. I'm not used to that lib. If you are, can you check the reason?

Nexucis:
Hello,

The issue related to windows is simply there is no docker layer on windows and as such you cannot ask or run a container on windows in a GitHub action.

```
2025/04/15 12:10:44 github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 27.5.1
  API Version: 1.47
  Operating System: Microsoft Windows Server Version 21H2 (OS Build 20348.3328)
  Total Memory: 16383 MB
  Testcontainers for Go Version: v0.36.0
  Resolved Docker Host: npipe:////./pipe/docker_engine
  Resolved Docker Socket Path: //./pipe/docker_engine
  Test SessionID: 8acb1731c054c06e6a2ec7200ccbaed1ed0503369a90527bea7cec6e34c1dcc2
  Test ProcessID: 97ff39b1-d033-489b-a989-85a8217c0154
2025/04/15 12:10:59 generic container: create container: ensure default network: network create: Error response from daemon: could not find plugin bridge in v1 plugin registry: plugin not found
```

There is an issue opened on GitHub side about that and still not implemented.

As a side note, all e2e tests are running with the file database and the SQL database. So at least the whole system is tested using both database

celian-garcia:
> Hello,
> 
> The issue related to windows is simply there is no docker layer on windows and as such you cannot ask or run a container on windows in a GitHub action.
> 
> ```
> 2025/04/15 12:10:44 github.com/testcontainers/testcontainers-go - Connected to docker: 
>   Server Version: 27.5.1
>   API Version: 1.47
>   Operating System: Microsoft Windows Server Version 21H2 (OS Build 20348.3328)
>   Total Memory: 16383 MB
>   Testcontainers for Go Version: v0.36.0
>   Resolved Docker Host: npipe:////./pipe/docker_engine
>   Resolved Docker Socket Path: //./pipe/docker_engine
>   Test SessionID: 8acb1731c054c06e6a2ec7200ccbaed1ed0503369a90527bea7cec6e34c1dcc2
>   Test ProcessID: 97ff39b1-d033-489b-a989-85a8217c0154
> 2025/04/15 12:10:59 generic container: create container: ensure default network: network create: Error response from daemon: could not find plugin bridge in v1 plugin registry: plugin not found
> ```
> 
> There is an issue opened on GitHub side about that and still not implemented.
> 
> As a side note, all e2e tests are running with the file database and the SQL database. So at least the whole system is tested using both database

Oh ok sure. If I understand, that is kind of invalidating the usage of that library in the windows tests then.
The question @Nexucis is would we accept to make a system with some flag for Linux only tests? It's probably cumbersome knowing that this is covered in e2e part.

Nexucis:
> The question @Nexucis is would we accept to make a system with some flag for Linux only tests

Well, it's always a good idea to add specific tests for the SQL database, and we could add them with a special `go:build` directive like we did for the e2e tests or the specific tests relative to Cuelang. And then run it only on linux.

So yeah no worries on that. 

> It's probably cumbersome knowing that this is covered in e2e part.

yeah the e2e tests are covering already many uses cases and indirectly are testing the database as well, whether it's a mysql database or the filesystem database

pvcnt:
To clarify why I opened this PR: While reading the code, I had the impression that the SQL database was never tested. It seems that I missed the fact that e2e tests also test that path (even though I do not know how the database is run then?). I still believe it might make sense to have more granular tests testing the database itself, rather than relying on running the entire e2e suite in two modes, but I'll let you be the judges of that!

Nexucis:
> even though I do not know how the database is run then?

The database is running in Github Action service container (it's defined in the CI files : https://github.com/perses/perses/blob/main/.github/workflows/go.yml#L93-L110) 

> I still believe it might make sense to have more granular tests testing the database itself,

Oh yeah I agree as well, I was just saying the pressure of doing it is not that high thanks to the e2e tests.

So if you would like to continue the effort, I am just suggestion to add a directive `go:build` on the tests files so we can run it only on a linux machine.

Nexucis:
Hey @pvcnt any plan on your side to continue to work on this topic ?
Or you don't have time anymore and it's better to close this PR ? And come back later on that topic on a fresher clone and PR ?

pvcnt:
Yes, I believe we can close the PR, I do not have bandwidth to work on it
at the moment.  And after previous comments, I was not even convinced
anymore this would bring a lot of values, since the database is already
tested via integration tests.

Le lun. 11 août 2025 à 10:14, Augustin Husson ***@***.***> a
écrit :

PR #2861, opened 2025-04-10
[FEATURE] Allow to configure CORS

pvcnt:
# Description

CORS is needed in order to host the frontend independently from the backend. This change adds the ability to configure the [CORS middleware](https://echo.labstack.com/docs/middleware/cors) in the config file.

# Checklist

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -151,9 +151,12 @@ encryption_key: <secret> # Optional
 
 # The path to the file containing the secret key.
 encryption_key_file: <filename> # Optional
+
+# Configuration for CORS (cross-origin resource sharing).
+cors: <CORS config> # Optional
 ```
 
-### Cookie config
+#### Cookie config
 
 ```yaml
 # Set the same_site cookie attribute and prevents the browser from sending the cookie along with cross-site requests.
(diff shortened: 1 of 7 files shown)


docs/configuration/configuration.md
```diff
@@ -151,9 +151,12 @@ encryption_key: <secret> # Optional
 
 # The path to the file containing the secret key.
 encryption_key_file: <filename> # Optional
+
+# Configuration for CORS (cross-origin resource sharing).
+cors: <CORS config> # Optional
 ```
 
-### Cookie config
+#### Cookie config
```
pvcnt, line 159: Indentation was wrong.

internal/api/core/core_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		resp = sendRequest(expect, "https://github.com")
+		resp.Status(http.StatusOK)
+		resp.Header("Access-Control-Allow-Origin").IsEqual("https://github.com")
+
+		resp = sendRequest(expect, "https://google.com")
+		resp.Status(http.StatusOK)
+		resp.Header("Access-Control-Allow-Origin").IsEmpty()
+	})
+}
```
jgbernalp, line 126: Maybe we are missing a test in which the preflight request fails, WDYT?

internal/api/core/core_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/gavv/httpexpect/v2"
+	"github.com/perses/perses/internal/api/utils"
+	"github.com/perses/perses/internal/test"
+	apiConfig "github.com/perses/perses/pkg/model/api/config"
+	"github.com/perses/perses/pkg/model/api/v1/secret"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/assert"
+)
+
+// We cannot reuse in this package what is inside internal/api/e2e because it would create a
```
celian-garcia, line 33: It would be very disappointing :/ If it's an e2e test, it should be located in e2e folder

celian-garcia (changes requested):

internal/api/core/core_test.go
pvcnt: I added an addition test for a preflight request with an invalid origin, but I'm not a CORS expert to be certain of what exactly is expected (the CORS middleware should be right about it though!). My understanding is that the request itself succeeds (i.e., it does return a 2xx code), that is the browser that is ultimately failing it if some headers are missing/incorrect.

internal/api/core/core_test.go
celian-garcia: Well e2e is properly badly named but this is definetely the kind of tests that we do in that folder. What we do in the WithServer of the "e2eframework" we've built is not very different from what you are doing in this withServer function. Considering the way we architectured, the place of that test is in a cors_test.go in e2e /api folder.

internal/api/core/core_test.go
celian-garcia: Yes would be hard to make a failing cors request actually as it's something forbidden only in a browser. Maybe creating a selenium test and I'm not even sure that localhost wouldn't be just whitelisted. Ok to not have failing tests here as impossible to me.

jgbernalp: We can do a [preflight request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#preflighted_requests) that uses the OPTIONS method and assert that it returns a 204 or not.

internal/api/core/core_test.go
pvcnt: Sounds good, I moved and renamed the file.

internal/api/core/core_test.go
celian-garcia: @jgbernalp you mean make a preflight request when CORS are disabled ?

jgbernalp: I meant to test that having CORS disabled does not change the current behavior. But I think the happy path would be sufficient as with CORS disabled the middleware is not added.

jgbernalp:
@pvcnt can you check the test failures, it seems related to missing configuration with the new `cors.enabled` field

pvcnt:
@jgbernalp Thanks for the callout, that's fixed now.

celian-garcia (approved):
ok to me  🙂 @jgbernalp I give you time to review.

pvcnt:
Failing with:
```
error: 		ui/ui-guidelines.md:20: "https://eslint.org/" not accessible even after retry; status code 0: net/http: request canceled (Client.Timeout or context cancellation while reading body)
```
I suppose it's a transient error?

Interestingly, I tried to run `make checkdocs` locally but it changes every single file.

celian-garcia:
Let's try again

celian-garcia:
dammit so flaky !

PR #2873, opened 2025-04-17
[BREAKING] core: switch the trace data model to OpenTelemetry (OTLP)

andreasgerstmayr:
# Description

Previously, the Tempo plugin received traces in OTLP format, transformed it into a custom Perses data model (a span tree), and passed it to the TracingGanttChart plugin.

Retrospectively, I think this wasn't ideal, and the data model transformation should be moved from the data source (Tempo) to the panel plugin (e.g. TracingGanttChart), because:

* using a standard input format like OTLP is better for adoption by other panel plugins
* some plugins might not need a span tree, i.e. computing the span tree is a waste of compute resources
* some edge cases (partial traces) are better handled at the panel plugin (e.g. TracingGanttChart)

This is the change for the core data model, I already prepared the change for the Tempo and TracingGanttChart plugins in a separate PR.

It is a breaking change for plugins which use the trace data model, not for end users, if the respective plugins are updated at the same time.

# Screenshots

no UI changes

# Checklist

## UI Changes

### DIFF
--- a/ui/core/src/model/index.ts
+++ b/ui/core/src/model/index.ts
@@ -23,6 +23,7 @@ export * from './kind';
 export * from './layout';
 export * from './legend';
 export * from './notice';
+export * from './otlp';
 export * from './panels';
 export * from './project';
 export * from './query';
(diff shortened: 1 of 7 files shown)


Gladorme (approved):
Make sense to me 👍 Trusting on the model, did not check

AntoineThebaud (approved):

Issue #2856, opened 2025-04-09
HTTP datasource discovery exposed the secured fields in the config endpoint


celian-garcia:
### What did you do?

https://github.com/perses/perses/blob/main/pkg/model/api/config/datasourcediscovery.go#L27

Should use ``PublicRestConfigClient`` here, otherwise if someone put oauth or basic auth for example, the secrets are leaked in the config endpoint.

### What did you see instead? Under which circumstances?

xxx

### Perses configuration file

```yaml

```

### Logs

```text

```

dougkirkley:
I can work on this issue, do we just want to create a separate `NewPublicRESTClient` function that takes in the `PublicRestConfigClient`?

celian-garcia:
I didn't check yet, as I was afraid of the consequences and the rabbit hole it could be 😄 
But I guess this would be a part of the fix yes. Or something that would convert config the other way around, to not duplicate code maybe.

celian-garcia:
But if you're motivated, yes you can work on it! nobody is on it.

dougkirkley:
Sure, I can look into it, I followed the rabbit hole a bit the other day, maybe there is some interface way to fix the duplication of the client function.

celian-garcia:
I was checking what prometheus was doing the other day, but it might be a big refactoring to make like them. 
https://github.com/prometheus/common/blob/main/config/http_config.go

Nexucis:
I am proposing to simply override the method `MarshalJSON` and `MarshalYAML` to use the public config instead in https://github.com/perses/perses/pull/2888

PR #2888, opened 2025-05-05
[BUGFIX] Config: hide secret in the datasource discovery

Nexucis:
fix https://github.com/perses/perses/issues/2856

### DIFF
--- a/pkg/model/api/config/datasourcediscovery.go
+++ b/pkg/model/api/config/datasourcediscovery.go
@@ -14,6 +14,7 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/model/api/config/datasourcediscovery.go
```diff
@@ -27,6 +28,16 @@ type HTTPDiscovery struct {
 	config.RestConfigClient `json:",inline" yaml:",inline"`
 }
 
+func (d HTTPDiscovery) MarshalYAML() (interface{}, error) {
+	cfg := config.NewPublicRestConfigClient(&d.RestConfigClient)
+	return cfg, nil
+}
+
+func (d HTTPDiscovery) MarshalJSON() ([]byte, error) {
```
celian-garcia, line 36: Isn't it dangerous to do that? I mean never saved in db this ?

Nexucis: the config is never saved in the database, so we should be good.

Besides in the config the usage of the type `secret.Hide` is working exactly like that: by overriding these methods

celian-garcia: sure sure. Ok if never saved then let's do that

celian-garcia (approved):

Issue #2938, opened 2025-05-21
ListVariable Infinite Rerender When Depending On Other Variables


OrReuben:
### What did you do?

- Create a variable plugin with the following `dependsOn` function:
```typescript
  dependsOn: options => {
    const variables = parseVariables(options.values.map(v => (typeof v === 'string' ? v : v.value)).join(' '));

    return { variables };
  }
```
- Create a variable in the dashboard that uses its own variable name
- Select Allow Multiple Values and Allow All option
- Select all in the variable
- See it crashing

### What did you expect to see?

When a list-type variable depends on other variables, it’s possible for it to indirectly or mistakenly depend on itself. If that variable is also configured as multi-select with the "Select All" option enabled, this self-dependency can cause an infinite re-render loop, eventually crashing the application.

The core issue is that the dependsOn method in the variable plugin has no awareness of the variable's own name, so it cannot filter itself out from its list of dependencies. This makes it impossible to prevent self-dependency purely within the plugin logic.

### What did you see instead? Under which circumstances?

**Proposed Solution**
To avoid this issue, the responsibility of filtering out self-references should lie in the plugin system layer, not in individual plugins. Specifically:

Before passing the dependency list to a variable plugin, the system should always exclude the variable itself from its own dependency list.

This ensures stability and prevents infinite loops, even if a variable is misconfigured or inadvertently set to depend on itself.

attaching a video:
![Image](https://github.com/user-attachments/assets/ac92bbcd-53b8-436a-9fa3-b2a16f13c703)

### Perses configuration file

```yaml

```

### Logs

```text

```

Nexucis:
closed with #2939

PR #2939, opened 2025-05-21
[BUGFIX] ListVariable Infinite Re-render

OrReuben:
solves https://github.com/perses/perses/issues/2938
# Description
When a list-type variable depends on other variables, it’s possible for it to indirectly or mistakenly depend on itself. If that variable is also configured as multi-select with the "Select All" option enabled, this self-dependency can cause an infinite re-render loop, eventually crashing the application.

The core issue is that the dependsOn method in the variable plugin has no awareness of the variable's own name, so it cannot filter itself out from its list of dependencies. This makes it impossible to prevent self-dependency purely within the plugin logic.

# Screenshots
Bug

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/Variables/variable-model.ts
+++ b/ui/plugin-system/src/components/Variables/variable-model.ts
@@ -51,11 +51,13 @@ export function useListVariablePluginValues(definition: ListVariableDefinition):
   const capturingRegexp =
     definition.spec.capturingRegexp !== undefined ? new RegExp(definition.spec.capturingRegexp, 'g') : undefined;
 
-  let dependsOnVariables: string[] | undefined;
+  let dependsOnVariables: string[] = Object.keys(allVariables); // Default to all variables
   if (variablePlugin?.dependsOn) {
     const dependencies = variablePlugin.dependsOn(spec, variablePluginCtx);
-    dependsOnVariables = dependencies.variables;
+    dependsOnVariables = dependencies.variables ? dependencies.variables : dependsOnVariables;
   }
+  // Exclude self variable to avoid circular dependency
+  dependsOnVariables = dependsOnVariables.filter((v) => v !== definition.spec.name);
 
   const variables = useAllVariableValues(dependsOnVariables);
 


ui/plugin-system/src/components/Variables/variable-model.ts
```diff
@@ -51,11 +51,13 @@ export function useListVariablePluginValues(definition: ListVariableDefinition):
   const capturingRegexp =
     definition.spec.capturingRegexp !== undefined ? new RegExp(definition.spec.capturingRegexp, 'g') : undefined;
 
-  let dependsOnVariables: string[] | undefined;
+  let dependsOnVariables: string[] = Object.keys(allVariables); // Default to all variables
   if (variablePlugin?.dependsOn) {
     const dependencies = variablePlugin.dependsOn(spec, variablePluginCtx);
-    dependsOnVariables = dependencies.variables;
+    dependsOnVariables = dependencies.variables ? dependencies.variables : dependsOnVariables;
```
jgbernalp, line 57:
```suggestion
    dependsOnVariables = dependencies.variables;
```

ui/plugin-system/src/components/Variables/variable-model.ts
```diff
@@ -51,11 +51,13 @@ export function useListVariablePluginValues(definition: ListVariableDefinition):
   const capturingRegexp =
     definition.spec.capturingRegexp !== undefined ? new RegExp(definition.spec.capturingRegexp, 'g') : undefined;
 
-  let dependsOnVariables: string[] | undefined;
+  let dependsOnVariables: string[] = Object.keys(allVariables); // Default to all variables
```
AntoineThebaud, line 54: Are you sure here? Where previously a "standalone" variable would have no dependency as expected, here it will be considered as dependent of all the other variables & would wait for them to load for no benefit.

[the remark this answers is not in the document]
OrReuben: This is the default behavior as for now.
If a user does not pass a dependsOn function the deps will be undefined and therefore you will get the full list of all variables as deps

OrReuben: If deps are undefined, by default you will get all of them

ui/plugin-system/src/components/Variables/variable-model.ts
```diff
@@ -51,11 +51,13 @@ export function useListVariablePluginValues(definition: ListVariableDefinition):
...
   if (variablePlugin?.dependsOn) {
     const dependencies = variablePlugin.dependsOn(spec, variablePluginCtx);
-    dependsOnVariables = dependencies.variables;
+    dependsOnVariables = dependencies.variables ? dependencies.variables : dependsOnVariables;
   }
+  // Exclude self variable to avoid circular dependency
+  dependsOnVariables = dependsOnVariables.filter((v) => v !== definition.spec.name);
```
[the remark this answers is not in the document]
OrReuben: i dont want them to ever be undefined:

see example
```typescript
export function useVariableValues(names?: string[]): VariableStateMap {
  const { state } = useVariableContext();

  const values = useMemo(() => {
    const values: VariableStateMap = {};
    names?.forEach((name) => {
      const s = state[name];
      if (s) {
        values[name] = s;
      }
    });
    return values;
  }, [state, names]);

  if (names === undefined) {
    return state;
  }

  return values;
}
```

OrReuben:
@AntoineThebaud @jgbernalp just making sure we understand the issue - 
if the deps are undefined, by default, the function will retrieve everything from the variableState.
Until now, if u didn't pass a `dependsOn` method - the deps would be undefined and therefore you would be depending on ALL variables (including self).

my fix here is not regarding default depending to all variables (as it was before), it's just to make sure we never depend on the self variable.

OrReuben:
@jgbernalp I believe something is wrong at the core implementation of the dependencies logically -
- if the user does not pass a dependsOn function, he will depend on all variables (he won't wait for them to load, but will depend on them regardless)
- If the user passes a dependsOn, he cannot be filtering it's self variable since he does not have the context to what is himself.


If I were to choose one solution for all plugin kinds, I would say we need to create a custom function that goes over the user's spec, retrieves all variables and automatically depend on them.

The dependsOn method is optional - if a user passes it, we need to depend on whatever he returns

Special case for variable dependencies- can never depend on its self variable.

OrReuben:
@jgbernalp also, it sounds to me more of a bug that we depend on other variables and not wait for them to load.

OrReuben:
@jgbernalp Another thing regarding the `dependsOn` method, for variables we do not have access to the `spec` property for some reason. is it intended?

jgbernalp:
> @jgbernalp also, it sounds to me more of a bug that we depend on other variables and not wait for them to load.

I agree that depending on other variables without waiting is also a bug. 

> @jgbernalp Another thing regarding the dependsOn method, for variables we do not have access to the spec property for some reason. is it intended?

Not sure what you mean, the spec is passed as first parameter to the `dependsOn` method.

I think the best way would be to create a test for this so we can validate regular and edge cases.

OrReuben:
> > @jgbernalp also, it sounds to me more of a bug that we depend on other variables and not wait for them to load.
> 
> I agree that depending on other variables without waiting is also a bug.
> 
> > @jgbernalp Another thing regarding the dependsOn method, for variables we do not have access to the spec property for some reason. is it intended?
> 
> Not sure what you mean, the spec is passed as first parameter to the `dependsOn` method.
> 
> I think the best way would be to create a test for this so we can validate regular and edge cases.

yeah my bad, it is just named differently so i was a bit confused with it (named as options while other plugins call it spec)

Regarding the changes in this PR, I just want to ensure we're aligned: I did not change the previous default behavior of the variable plugin.

As we discussed, we should avoid relying on variables that we don’t explicitly wait for. Prior to this change, the system defaulted to depending on all variables without actually waiting for them.

The primary change introduced in this PR is the removal of the self variable from the dependency list, in order to prevent self-dependency and avoid the infinite re-render issue.

jgbernalp:
I think we agree on the points that the initial implementation might be adding more things that needed and that a self dependency causes an infinite render issue. The best course of action would be to create a test on this PR for these cases.

OrReuben:
@jgbernalp anything waiting from my side?

jgbernalp:
No @OrReuben, thnx for your patience and contributions. I'll merge it now!

PR #2940, opened 2025-05-22
[FEATURE] add kubernetes authn + authz provider

PeterYurkovich:
# Description
This PR looks to add kubernetes authorization and authentication to perses. Documentation has been updated in the Authentication.md and Authorization.md files which explains more in detail, but on a surface level, the Perses backend will expect an "Authorization" header which it will use to perform a TokenReview request to kubernetes to determine the user, and then SubjectAccessReview to determine RBAC controls for namespaces, PersesDashboards and PersesDatasources. 

# Screenshots

Error on login screen when login fails while kubernetes authn + authz are enabled. 
image

# Checklist

## UI Changes

# TODO
These are the remaining items to complete for this PR. However I wanted to put it up early to begin the review and discussion process.

- [ ] Update percli to accept kubeconfig locations (or the `KUBECONFIG` env variable)
	- [ ] Update the flow diagram to reflect the change
- [ ] Update the client API package with the new options
- [ ] Support special characters in usernames
- [ ] Further in-cluster testing and integrations with the `perses/perses-operator`

While most kubernetes usernames will follow the current username convention of `^[a-zA-Z0-9_.-]+$`, some usernames fall outside of the convention. For example `kube:admin`, `kube-admin`, or `system:account`. The first step to supporting these usernames is allowing them to be saved as files. This is completed through hashing the file names before saving them. Since all queries are completed on the internal data of the files, this change should be cosmetic while allowing for unique characters and preventing collisions. Finally the verification function will need to know if it is executing within a kubernetes auth format. Unfortunately the verification is quite deeply nested within the object oriented structure, and so a "kubenetes" value may have to be passed down to a number of location, similar to how `case_sensitive` is. This isn't an ideal solution though, and it would be best to be able to avoid it if possible. Feedback on this would be appreciated.Signed-off-by: Peter Yurkovich <pyurkovi@redhat.com>


internal/api/rbac/k8s.go
```diff
@@ -0,0 +1,460 @@
...
+
+	for _, namespace := range k8sNamespaces.Items {
+		namespaces = append(namespaces, namespace.Name)
+	}
+
+	return namespaces
+}
+
+func (r *k8sImpl) getUser(ctx echo.Context) (*k8suser.Info, error) {
+	ctx.Request().Header.Set("Authorization", crypto.GetAuthnHeaderFromClient(ctx, r.kubeconfig))
```
Nexucis, line 384: First of all, thank you for taking the time to take care about this feature.

Overall I think the changes are more or less ok. What is bothering me currently is this particular line actually.

Initially the RBAC system is a pure in memory system and doesn't modify anything specially the HTTP request.
And ideally I would prefer it is kept like that.

Besides setting the `Authorization` Header for the whole request feels weird for me as in the JWT middleware it is actually set with the `Perses-Authorization` Header.

So here we have a kind of conflict. Perhaps we should fork the request in that case.

I understand the requirement to actually rely on an external permission system than the internal one, I am just wondering if we really need to use the initial request coming to Perses. 

In my head, if we want to support another external auth and permission layer, I would just use the http client to contact the remote auth server without touching the initial request.

Now, do I have a better solution ? Not currently unfortunately, I need to think more about that and to read again and again what you coded

PeterYurkovich:
Sorry for the delay in updating, I needed to take some time to try to figure out the best way to structure all of the changes and abstractions. In the most recent set of commits I have added two new abstractions:

1. A request "security" parameter, which contains the jwtImpl as well as a new "GetUser" function which is able to retrieve the username for a user from the `echo.context`. Many places which previously had the jwtImpl passed down to them and then exacted and checked the claim for the subject now instead use the "GetUser" function instead, which is able to handle both requests which contain a jwt header/cookie as well as a request which just contains a k8s Authorization header.
2. An authorizationMiddleware abstraction, previously was the jwtMiddleware. Allow the checking of the Authorization header for both jwt and k8s.

Finally some other work was completed to allow the usage of special characters in usernames, like what kubernetes often has. This was completed by checking if the DAO has the HashFileNames config parameter set to true, however I could easily see an argument that it should be a separate config parameter specifically for ignoring validating the usernames.

As always let me know what you think, I'm happy to make changes where you think would be best. I'm also happy to squash the changes down to a single commit if you would like, I just thought it may be easier to review with the changes chunked out like this

pkg/client/config/config.go
```diff
@@ -32,6 +32,8 @@ import (
 
 const connectionTimeout = 30 * time.Second
 
+// TODO update pkg/client/api with new kubernetes authentication options. Unclear if this is needed,
+// Maybe for the perses operator?
```
jgbernalp: currently we just pass the Authorization header directly: https://github.com/perses/perses-operator/blob/main/internal/perses/common/perses_client_factory.go#L66

PeterYurkovich: Yes, I had tested this and and the client (which is used in the percli) and had forgotten to remove the comment after I had done so. I've cleaned up a lot of the extraneous comments in the last push

PR #2948, opened 2025-05-27
[ENHANCEMENT] pass rest of options from queryOptions

ddanielcruzz:
# Description

`<DataQueriesProvider />` accepts a `queryOptions` prop that is intended to match the options used in `@tanstack/react-query` query hooks. 

Although it appears that all options can be passed in, only a limited subset is actually forwarded to the query configuration. This can create a false impression that all fetching options are customizable. This PR updates the implementation to pass all options using the spread operator. The spread is applied as the first argument to avoid accidentally overriding custom logic such as `enabled` or `queryFn`.

# Screenshots

N/A

# Checklist

## UI Changes

N/A


Nexucis (approved):
ah yeah, indeed it won't hurt to do that

PR #2958, opened 2025-06-02
[DOC] Update doc about how to load a plugin in dev environment

Nexucis:
I am just updating the instruction to load a plugin from a dev environment

### DIFF
--- a/docs/configuration/load-plugin.md
+++ b/docs/configuration/load-plugin.md
@@ -10,7 +10,8 @@ plugin:
   archive_path: /path/to/archive/folder
 ```
 
-Note: to know how to build a plugin archive, please refer to the [plugin creation documentation](../plugins/creation.md).
+Note: to know how to build a plugin archive, please refer to
+the [plugin creation documentation](../plugins/creation.md).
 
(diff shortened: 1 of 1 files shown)


docs/configuration/load-plugin.md
```diff
@@ -34,16 +35,15 @@ plugins to be loaded.
...
-```yaml
-plugin:
-  dev_environment:
-    plugins:
-      - name: <plugin_module_name> # Like Prometheus
-        url: http://localhost:3005 # The URL of the rsbuild dev server
-        disable_schema: false # If you want to disable the schema loading
-        absolute_path: "/path/to/plugin/folder" # The absolute path of the plugin folder. Required if disable_schema is false
+```bash
+percli plugin start ./path/to/plugin
```
jgbernalp, line 47: maybe we should mention that percli login must be executed first to a running perses backend. Or does the start command executes a login first?

Nexucis: ah yeah we should mention it.

No the start command does not execute the login first. I think it's better it does not do it

jgbernalp (approved):

PR #2959, opened 2025-06-02
[FEATURE] Core: Add ProfileQuery

Stevelening:
# Description

Add new query type ProfileQuery for supporting profiling data. Related to PR https://github.com/perses/perses/pull/2917 .

I split the previous PR into two PR perses/core and perses/plugin-system. This first PR (perses/core) need to be done before the second one with only perses/plugin-system.

# Checklist

### DIFF
--- a/internal/api/plugin/schema/schema.go
+++ b/internal/api/plugin/schema/schema.go
@@ -135,6 +135,8 @@ func (s *sch) Load(pluginPath string, module v1.PluginModule) error {
 			s.datasources[schema.Name] = schema.Instance
 		case plugin.KindTimeSeriesQuery, plugin.KindTraceQuery:
 			s.queries[schema.Name] = schema.Instance
+		case plugin.KindProfileQuery:
+			//todo: profile don't need a query but maybe something else
 		case plugin.KindVariable:
 			s.variables[schema.Name] = schema.Instance
 		case plugin.KindPanel:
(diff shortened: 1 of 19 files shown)


ui/core/src/model/units/index.ts
```diff
@@ -15,3 +15,4 @@ export * from './utils';
 export * from './units';
 export * from './constants';
 export * from './types';
+export * from './bytes';
```
Gladorme: Should not be necessary. If it's because of `formatBytes` func, check `formatValue` func already exported 😄

ui/plugin-system/src/runtime/time-series-queries.ts
```diff
@@ -132,7 +132,6 @@ export function useTimeSeriesQueries(
       const plugin = pluginLoaderResponse[idx]?.data;
       const { queryEnabled, queryKey } = getQueryOptions({ plugin, definition, context });
       return {
-        ...queryOptions,
```
Gladorme: Do not remove 😄

internal/api/plugin/schema/schema.go
```diff
@@ -135,6 +135,8 @@ func (s *sch) Load(pluginPath string, module v1.PluginModule) error {
 			s.datasources[schema.Name] = schema.Instance
 		case plugin.KindTimeSeriesQuery, plugin.KindTraceQuery:
 			s.queries[schema.Name] = schema.Instance
+		case plugin.KindProfileQuery:
```
Gladorme: Move with other queries case

PR #3020, opened 2025-06-13
[ENHANCEMENT] stronger plugin format validation

AntoineThebaud:
# Description

When loading a plugin model schema, the presence & the type of the `spec` attribute are now checked just like `kind` was already. The error feedback was also improved. 

# Checklist


internal/api/plugin/schema/cue.go
```diff
@@ -47,14 +45,27 @@ func LoadModelSchema(schemaPath string) (string, *build.Instance, error) {
...
+	}
+	// Check the presence & type of `spec`
+	specValue := schema.LookupPath(cue.ParsePath("spec"))
+	if specValue.Err() != nil {
+		return "", nil, fmt.Errorf("invalid schema at %s: required `spec` field is missing: %w", schemaPath, specValue.Err())
 	}
+	specKind := specValue.Kind()
+	// NB: not perfect check. We accept BottomKind as it is returned in valid cases (empty struct, disjunction..),
+	// but it is also returned in invalid cases (spec: number, spec: string..). Unfortunately we can't make this smarter here.
+	if specKind != cue.StructKind && specKind != cue.BottomKind {
```
Nexucis: actually you can make it smarter if you do that: 

```go
if specKind == cue.BottomKind && specValue.IncompleteKind() != cue.StructKind {
  // Invalid case
}
```

Nexucis: if it's a disjunction, you could even check what type each value is in the disjunction is. Like that for example: https://github.com/perses/perses/blob/main/internal/api/plugin/migrate/migrate.go#L305-L317

Nexucis: so if in the disjunction you have a string or any native type (bool, float, int), then you could reject it

AntoineThebaud: Oh interesting, 'might not have dig enough the library then.. thanks I will check this

AntoineThebaud: so in the end went for
```go
if specKind != cue.StructKind && specValue.IncompleteKind() != cue.StructKind {
```
which matches exactly what we wanted here 🥳

PR #3027, opened 2025-06-17
[IGNORE] RBAC interface is using the perses context

Nexucis:
This PR is changing the rbac interface in order to use the Perses context instead of the login of the user.

This is a first PR to make the authorization layer more generic and later to be able to improve and accept the PR #2940.

Other PRs will follow changing step by step the authorization layer

/cc @PeterYurkovich


Nexucis:
moving forward on this one as it goes to a feature branch, so review can be done later

PR #3031, opened 2025-06-19
[BUGFIX] Fixing fetch + JWT token expired + replacing datasource LRU cache by Tanstack query

### LINKED ISSUE #3029: Having no read access on GlobalDatasources prevents to use Datasources
### What did you do?

- Have a Perses server running without the default `read *` guest permission.
- create a project, a datasource and a dashboard
- edit the dashboard to start adding a panel

In my panel I should be able to use the (project) datasource I defined earlier

the dropdown is empty

Image

In the console logs we can see this

Image

This is due to a design issue in the Datasource provider, that tries to retrieve the Global datasources and the project ones, no matter what.

### System information

_No response_

### Perses version

0.51.0

### Perses configuration file

```yaml
[...]
guest_permissions:
  - actions:
      - create
    scopes:
      - Project
[...]
```

### Logs

```text
GET http://localhost:3000/api/v1/globaldatasources?kind=PrometheusDatasource&default=true 403 (Forbidden)
query.mjs:294 UserFriendlyError: forbidden access: missing 'read' global permission for 'GlobalDatasource' kind
    at fetch (fetch.ts:27:1)
    at async fetchJson (fetch.ts:33:1)
    at async Object.current (DatasourceStoreProvider.tsx:108:1)
    at async Promise.all (:3000/projects/hello/dashboard/index 0)
    at async Object.getClient [as getDatasourceClient] (DatasourceStoreProvider.tsx:134:1)
```

Gladorme:
# Description

This PR is fixing lot of stuff around fetch.

Fix #3029 => fetch was not throwing error if response was not OK

Fixing access token refreshing:
- access token cookie max age aligned with refresh token max age => in order to have 401 status code when sending expired cookie
- add queryresult as reset key for panel error boundary
- override native fetch for Perses  with refetch capabilities, when authorization is enabled

LRU Cache was bothering me (spamming in dashboard loading) and could be incompatible with jwt fix, so I changed to use Tanstack Query on the way too (no more spam) 🔥 

# Checklist


dev/config.yaml
```diff
@@ -13,6 +13,7 @@ security:
         scopes:
           - Project
   authentication:
+    access_token_ttl: 5s
```
Nexucis: should be removed

ui/app/src/model/auth-client.ts
```diff
@@ -32,11 +32,15 @@ export interface NativeAuthBody {
...
+  const accessToken = useAuthToken();
 
   // Warm the access token request cache back
   // If the refresh token is not expired, the debounce mechanism will get the refreshed accedd token.
   // Otherwise, debounce will let pass the empty access token and auth guard will redirect to sign in.
-  if (isAuthEnabled && cookies[jwtPayload] === undefined) {
+  if (
+    isAuthEnabled &&
+    (!accessToken || !accessToken.data || !accessToken.data.exp || accessToken.data.exp > new Date())
+  ) {
```
shahrokni: I think this condition can be shortened a bit. Would that work for you? 

```javaScript
/*This is shorter and uses .? operator*/
if (isAuthEnabled && (!accessToken?.data?.exp || accessToken.data.exp > new Date())) {
    refreshToken();
}
```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
+  const basePath = getBasePathName();
+  let url = `${!project && !dashboard ? 'globaldatasources' : 'datasources'}/${encodeURIComponent(name)}`;
+  if (dashboard) {
+    url = `dashboards/${encodeURIComponent(dashboard)}/${url}`;
+  }
+  if (project) {
+    url = `projects/${encodeURIComponent(project)}/${url}`;
+  }
+  return `${basePath}/proxy/${url}`;
 }
```
shahrokni, line 39: **_Improvement, (Your call)_**
Although the logic of the function is correct, **from the consumers' perspective** the function may misbehave, if both the project and dashboard are passed down. The reason is that the second condition overrides the first one in case both parameters are available. One easy solution would be in-line return statement. However, the ambiguity would still exist. The second solution which  would be better is: 

```Typescript
export const proxyUrlBuilder = (
  type: "global" | "project" | "dashboard",
  params: { project?: string; dashboard?: string; name: string }
): string => {
  const basePath = getBasePathName();

  const strategies: Record<string, () => string> = {
    global: () =>
      `${basePath}/proxy/globaldatasources/${encodeURIComponent(params.name)}`,
    project: () =>
      `${basePath}/proxy/projects/${encodeURIComponent(
        params.project!
      )}/datasources/${encodeURIComponent(params.name)}`,
    dashboard: () =>
      `${basePath}/proxy/dashboards/${encodeURIComponent(
        params.dashboard!
      )}/datasources/${encodeURIComponent(params.name)}`,
  };

  if (!strategies[type]) {
    throw new Error(`Invalid type: ${type}`);
  }

  return strategies[type]();
};

```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
-        // Like that, next time another panel ask for the exact same datasource (with the same selector), then we won't query the server to try it again.
-        // It's ok to do it as the cache has an expiration of 5min.
-        // We have the same logic for the globalDatasources.
-        this.cache.setUndefinedDatasource(project, selector);
-      } else {
-        this.cache.setDatasource(result);
+  const getDatasource = useCallback(
+    async (project: string, selector: DatasourceSelector) => {
+      if (isDatasourcesPending || datasources === undefined) {
+        return undefined;
```
shahrokni: **_Code-style (Your call)_**

Just a suggestion,
wouldn't this a bit cleaner? 

```javaScript
isDatasourcesPending || !datasources

/*OR, If it makes sense and datasources is an array, you can even do this*/
isDatasourcesPending || !datasources?.length
```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
+      return datasources.find((datasource) => {
+        if (datasource.metadata.project !== project) {
+          return false;
+        }
+        if (selector.kind !== datasource.spec.plugin.kind) {
+          return false;
+        }
+        if (selector.name === undefined) {
+          return datasource.spec.default;
+        }
```
shahrokni: **_Code-style suggestion (Your call)_**

```javaScript
if (!selector.name)
```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
+        return datasource.metadata.name.toLowerCase() === selector.name.toLowerCase();
+      });
+    },
+    [datasources, isDatasourcesPending]
+  );
+
+  const getGlobalDatasource = useCallback(
+    async (selector: DatasourceSelector) => {
+      if (isGlobalDatasourcesPending || globalDatasources === undefined) {
+        return undefined;
```
shahrokni: **_Code-style suggestion (Your call)_**

```javaScript
 if (isGlobalDatasourcesPending || !globalDatasources)
```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
-      return list;
-    });
-  }
+      return globalDatasources.find((datasource) => {
+        if (selector.kind !== datasource.spec.plugin.kind) {
+          return false;
+        }
+        if (selector.name === undefined) {
+          return datasource.spec.default;
+        }
```
shahrokni: **_Code-style suggestion (Your call)_**

```javaScript
if (!selector.name)
```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
+      });
+    },
+    [globalDatasources, isGlobalDatasourcesPending]
+  );
+
+  const listDatasources = useCallback(
+    async (project: string, pluginKind?: string) => {
+      if (isDatasourcesPending || datasources === undefined) {
+        return [];
+      }
```
shahrokni: **_Code-style suggestion (Your call)_**

```javaScript
 if (isDatasourcesPending || !datasources) {
        return [];
      }
```

And also this one

```javaScript
if (pluginKind && datasource.spec.plugin.kind !== pluginKind)
```

ui/app/src/model/datasource-api.ts
```diff
@@ -11,223 +11,114 @@
...
+      return globalDatasources.filter((datasource) => {
+        if (pluginKind !== undefined && datasource.spec.plugin.kind !== pluginKind) {
+          return false;
+        }
+        return true;
+      });
+    },
+    [globalDatasources, isGlobalDatasourcesPending]
+  );
```
shahrokni: **_Code-style suggestion (Your call)_**

```javaScript
if (isGlobalDatasourcesPending || !globalDatasources)
```

And this one

```javaScript
  if (pluginKind && datasource.spec.plugin.kind !== pluginKind)
```

ui/app/src/model/datasource-client.ts
```diff
@@ -20,130 +20,78 @@ import {
...
-    },
-    onSuccess: () => {
-      return queryClient.invalidateQueries({ queryKey: key });
-    },
+export function updateDatasource(entity: DatasourceResource): Promise<DatasourceResource> {
+  const name = entity.metadata.name;
+  const project = entity.metadata.project;
+  const url = buildURL({ resource, project, name });
+  return fetchJson<DatasourceResource>(url, {
+    method: HTTPMethodPUT,
```
shahrokni, line 64: Just a quick Question @Nexucis and @Gladorme 

Wouldn't it be better to expose a patch instead of put?
Are we replacing the entire resource or is it a partial update?

ui/app/src/model/global-datasource-client.ts
```diff
@@ -10,22 +10,169 @@
...
+
+export function updateGlobalDatasource(entity: GlobalDatasourceResource): Promise<GlobalDatasourceResource> {
+  const name = entity.metadata.name;
+  const url = buildURL({ resource, name });
+  return fetchJson<GlobalDatasourceResource>(url, {
+    method: HTTPMethodPUT,
+    headers: HTTPHeader,
+    body: JSON.stringify(entity),
+  });
+}
```
shahrokni, line 67: **_Just a quick Question_** @Nexucis and @Gladorme

Wouldn't it be better to expose a patch instead of put?
Are we replacing the entire resource or is it a partial update?

shahrokni (changes requested):

ui/app/src/model/global-datasource-client.ts
Nexucis: We are replacing the entire resource. So a `PUT` is accurate in that case.

ui/app/src/model/datasource-client.ts
Gladorme: We are replacing entire resource, this endpoint do not support partial update

ui/app/src/model/datasource-client.ts
```diff
@@ -156,44 +104,78 @@ export function useDatasourceList(options: DatasourceListOptions): UseQueryResul
...
+    mutationFn: (datasource: DatasourceResource) => {
+      return updateDatasource(datasource);
+    },
+    onSuccess: (entity: DatasourceResource) => {
+      return Promise.all([
+        queryClient.invalidateQueries({ queryKey: [...queryKey, entity.metadata.name] }),
+        queryClient.invalidateQueries({ queryKey }),
+      ]);
+    },
   });
```
shahrokni, line 154: @Gladorme I meant handling the exceptions using onError here. But now I see probably  those exceptions are handled by an ErrorBoundry in the code. **So, ignore this comment.**

ui/app/src/model/datasource-api.ts
Gladorme: I don't want to change the signature of this function and we want to override dashboard if project is provided 😄

shahrokni (approved):

AntoineThebaud (approved):

PR #3033, opened 2025-06-19
[IGNORE] Replace RBAC package by authorization and provide a new interface

Nexucis:
I am replacing the package `rbac` by `authorization`. This new package also provide a better and more complete interface. This new interface should be more generic and less relative to how we implement our rbac system.

I am also starting to cleanup the package crypto. In the next PR, I will tackle the usage of the interface `JWT` from the package `crypto`.

/cc @celian-garcia @PeterYurkovich

### DIFF
--- a/internal/api/authorization/authorization.go
+++ b/internal/api/authorization/authorization.go
@@ -30,6 +30,11 @@ type Authorization interface {
 	// IsEnabled returns true if the authorization is enabled, false otherwise.
 	IsEnabled() bool
 	// GetUser returns the user information from the context. The user information will depend on the implementation.
+	// While implementing this method, consider that the user information is not guaranteed to be set in the context.
+	// You should consider the case where the context can be empty and that the function can be called from an anonymous endpoint.
+	// To check if it is called from an anonymous endpoint, you can use the function utils.IsAnonymous.
+	// In case the context is not empty, and it is not an anonymous endpoint, the user information should be set in the context.
+	// If it is not the case, you should return an error.
 	GetUser(ctx echo.Context) (any, error)
 	// GetUsername returns the username/the login of the user from the context.
 	GetUsername(ctx echo.Context) (string, error)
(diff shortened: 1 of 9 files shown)


internal/api/authorization/authorization.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/perses/perses/internal/api/interface/v1/globalrole"
+	"github.com/perses/perses/internal/api/interface/v1/globalrolebinding"
+	"github.com/perses/perses/internal/api/interface/v1/role"
+	"github.com/perses/perses/internal/api/interface/v1/rolebinding"
+	"github.com/perses/perses/internal/api/interface/v1/user"
+	"github.com/perses/perses/pkg/model/api/config"
+	v1Role "github.com/perses/perses/pkg/model/api/v1/role"
+)
+
+type Authorization interface {
```
PeterYurkovich, line 29: I think my only real concern with the interface is its interaction with anonymous endpoints. For example the `GetUserProjects` and `GetPermissions` native implementations will error if they are called via an anonymous endpoint, but the `HasPermission` implementation won't. 

Furthermore the interaction between the middleware and the user/permission is reinforced only through comments saying "if we are at this point without a user it is because it is an anonymous endpoint". I don't wonder if rather than needing to make that assumption, we can instead look to only check for the user and the permissions if the endpoints aren't anonymous. That way we could have a direct line of "this is an anonymous endpoint" parameters rather than needing to rely on the endpoint not enabling the middleware which in turn doesn't add the header, which in turn is attempted to be found. For example we could put the anonymous endpoint check [here](https://github.com/perses/perses/blob/main/internal/api/toolbox/list.go#L47) so that we just return the full list if the anonymous endpoint is just going to return eveything anyways

With that change we could then say that all functions under the Authorization interface should only be called under non-anonymous endpoints and remove the differing behavior between them. If that isn't desired the I think it would be good to at least note in the comments here on the implementation which functions can and can't be called by anonymous endpoints.

Nexucis: yes you are right.
I will try to find a way to clarify when we are dealing with an anonymous endpoint. There is also the provisioning use case that is creating an empty context and the `authorisation` service is called to get the current user.

In the current implementation, everything is a bit mixed and I guess that's where the confusion is coming from.
Let see if I can make it better, but I don't make any promises on that

Nexucis: I think a way for handling the anonymous endpoint is to inject in the context the information whether it is an anonymous endpoint or not.

It's again something that will rely on an another middleware and it won't solve everything but that's a step in the right direction.

Nexucis: I have introduced a new middleware and modify a bit the `Authorisation` interface. I have also added more comments to give more details about what case the implementation should consider.

@PeterYurkovich let me know what do you think about it.

PeterYurkovich: I think that these changes are fine. The extra errors and boolean being returned should tip off anyone looking to use them that there is reason for them to potentially fail, and then check the comments left on them as to why. I think it should be fine to merge with this

Nexucis:
As @PeterYurkovich agreed with the changes, I considered it as an approval. I am merging this PR then

PR #3035, opened 2025-06-20
[ENHANCEMENT] make `unit` field in `format` optional

coleenquadros:
When creating panels using go sdk and format is not provided, cue validation complains while importing that the panel is invalid 

```
              "columnSettings": [
                {
                  "name": "cluster",
                  "header": "Cluster",
                  "format": {
                    "unit": ""
                  },
                  "align": "right"
                },
```

`bad request: invalid panel 0_0: spec.columnSettings.0.format: 11 errors in empty disjunction: spec.columnSettings.0.format.unit: 8 errors in empty disjunction: spec.columnSettings.0.format.unit: conflicting values "bytes" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:35:18 spec.columnSettings.0.format.unit: conflicting values "days" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:69 spec.columnSettings.0.format.unit: conflicting values "decimal" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format`

https://github.com/perses/plugins/pull/170

### DIFF
--- a/ui/components/src/FormatControls/FormatControls.tsx
+++ b/ui/components/src/FormatControls/FormatControls.tsx
@@ -12,16 +12,16 @@
 // limitations under the License.
 import { Switch, SwitchProps } from '@mui/material';
 import {
-  shouldShortenValues,
   FormatOptions,
   UNIT_CONFIG,
   UnitConfig,
   isUnitWithDecimalPlaces,
   isUnitWithShortValues,
+  shouldShortenValues,
 } from '@perses-dev/core';
 import { ReactElement } from 'react';
 import { OptionsEditorControl } from '../OptionsEditorLayout';
-import { SettingsAutocomplete, SettingsAutocompleteOption } from '../SettingsAutocomplete';
+import { SettingsAutocomplete } from '../SettingsAutocomplete';
 
 export interface FormatControlsProps {
   value: FormatOptions;
@@ -29,18 +29,16 @@ export interface FormatControlsProps {
   disabled?: boolean;
 }
 
-type AutocompleteUnitOption = Omit<SettingsAutocompleteOption, 'id'> &
-  UnitConfig & {
-    id: NonNullable<FormatOptions['unit']>;
-  };
+type AutocompleteUnitOption = UnitConfig & {
+  id: NonNullable<FormatOptions['unit']>;
+};
 
 const KIND_OPTIONS: readonly AutocompleteUnitOption[] = Object.entries(UNIT_CONFIG)
-  .map(([id, config]) => {
-    const unit = id as NonNullable<FormatOptions['unit']>;
+  .map<AutocompleteUnitOption>(([id, config]) => {
     return {
       ...config,
-      id: unit,
-      group: config.group ?? 'Decimal',
+      id: id as AutocompleteUnitOption['id'],
+      group: config.group || 'Decimal',
     };
   })
   .filter((config) => !config.disableSelectorOption);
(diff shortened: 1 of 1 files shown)


coleenquadros:
@saswatamcode @AntoineThebaud for review

go-sdk/common/format.go
```diff
@@ -50,7 +50,7 @@ const (
 )
 
 type Format struct {
-	Unit          string `json:"unit" yaml:"unit"`
+	Unit          string `json:"unit,omitempty" yaml:"unit,omitempty"`
 	DecimalPlaces int    `json:"decimalPlaces,omitempty" yaml:"decimalPlaces,omitempty"`
 	ShortValues   bool   `json:"shortValues,omitempty" yaml:"shortValues,omitempty"`
 }
```
AntoineThebaud, line 56: Btw in the same file you should also add this at the beginning of `validate()`:

```go
if f.Unit == nil {
    return nil
}
```

Otherwise your initial issue wont be solved. Btw maybe worth updating the unit tests to match this case then.

go-sdk/common/format.go
```diff
@@ -50,7 +50,7 @@ const (
 )
 
 type Format struct {
-	Unit          string `json:"unit" yaml:"unit"`
+	Unit          string `json:"unit,omitempty" yaml:"unit,omitempty"`
```
AntoineThebaud:
```suggestion
	Unit          *string `json:"unit,omitempty" yaml:"unit,omitempty"`
```

coleenquadros:
I do not have expertise in typescript and the UI fixes have been suggested by AI. Can someone with typescript knowledge review the UI changes?

ui/components/src/FormatControls/FormatControls.tsx
```diff
@@ -21,21 +21,26 @@ import {
...
-type AutocompleteUnitOption = UnitConfig & { id: FormatOptions['unit'] };
+type AutocompleteUnitOption = Omit<SettingsAutocompleteOption, 'id'> &
+  UnitConfig & {
+    id: NonNullable<FormatOptions['unit']>;
+  };
 
-const KIND_OPTIONS: AutocompleteUnitOption[] = Object.entries(UNIT_CONFIG)
+const KIND_OPTIONS: readonly AutocompleteUnitOption[] = Object.entries(UNIT_CONFIG)
   .map(([id, config]) => {
+    const unit = id as NonNullable<FormatOptions['unit']>;
```
jgbernalp: All FormatOptions have a value, so from typescript view this is never null
```suggestion
    const unit = id as FormatOptions['unit'];
```

ui/components/src/FormatControls/FormatControls.tsx
```diff
@@ -61,7 +66,7 @@ export function FormatControls({ value, onChange, disabled = false }: FormatCont
 
   const handleKindChange = (_: unknown, newValue: AutocompleteUnitOption): void => {
     onChange({
-      unit: newValue.id,
+      unit: newValue.id ?? 'decimal', // Fallback to 'decimal' if no unit is selected
```
jgbernalp: This will mean that when we try to edit a dashboard that was created with go, then we will always change it to decimal. Maybe what we need to support here is to be able to select and undefined/empty value to the unit. This will match what the SDK does.

jgbernalp:
@coleenquadros I created a PR on your fork with the type suggestions, PTAL: https://github.com/coleenquadros/perses/pull/1 otherwise this LGTM

PR #3042, opened 2025-06-24
split plugin loaded from dev and from the regular way

Nexucis:
Goal of this PR is to be able to unregister a plugin from the dev environment. 
It will happen when you stop the command `percli plugin start`

To be able to achieve that, we need to make a clear separation between plugins loaded from the dev and plugins loaded from the archives.

For the moment, the split is done for the struct handling the schema, but for the struct handling the migration script.


Gladorme (approved):
lgtm

PR #3045, opened 2025-06-25
[FEATURE] Time Series Chart CSV Export

ericahinkleRH:
Signed-off-by: Erica Hinkle <ehinkle@redhat.com>

## Explanation: 
Created a button within the hamburger option on each Perses graph panel. This feature takes the graph data from the specified panel and exports it to your computer as a CSV file named "'YourProjectName'_'YourPanelName'_data.csv". The first line of the CSV has important column name information. Then, the rest of the file is all of the data based on times and then commas separating the data itself. Each line has "TimeSeries Information, data1, data2(if applicable), etc" and all data is comma separated. I only made substantial changes to the "PanelActions.tsx", "PanelHeader.tsx", and "Panel.tsx" files.

#### Red Hat JIRA: 
https://issues.redhat.com/browse/OU-839

#### The illuminated three stacked lines is where my "Export as CSV" option lives:
image

#### What now pops up once you press the three lines "hamburger menu". It is a button that you can press:
image

#### Once you press the export button, it triggers a download on your device with all of the CSV data in the graph that you specified. The name will be "'YourProjectName'_'YourPanelName'_data.csv" and then when you open the file the first line is describing the names for the row setup, like the names and context for each of the numbers and datetimes. the format of the first line/data is "dateTime, data1, data2, etc.":
image

# Checklist

## UI Changes


Nexucis:
Hello @ericahinkleRH, thank you for the PR.

Do you mind to explain why do we need to export the data from a panel ? Personally I am struggling to see a good purpose about it.

On a different subject, the CSV export is currently behind a hamburger button. Do you see other things that would go behind this button ?

Nexucis:
Besides since this is a feature at panel level, it should work for every panel plugin. So probably if there is a good reason to do that, the system plugin should provide somehow an interface to implement regarding the CSV export. Then in the panel plugin like TimeSeriesChart, it will have to implement this interface to propose a way to export the timeseries data.

When a plugin doesn't implement this interface, the button should not appear

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -91,6 +135,9 @@ export const Panel = memo(function Panel(props: PanelProps) {
 
   const { queryResults } = useDataQueriesContext();
 
+  // Extract project name from URL - memoized to prevent unnecessary re-calculations
+  const projectName = useMemo(() => extractProjectNameFromUrl(), []);
```
jgbernalp: I'm not sure this is the best way, we are making the assumption that this panel would be used on a specific URL pattern. Maybe would be better to pas this as an optional prop.

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -60,6 +62,129 @@ const ConditionalBox = styled(Box)({
   justifyContent: 'flex-end',
 });
 
+// Function to check if the data is time series data
```
jgbernalp: The logic below seems specific to map time series to CSV, it might be better to extract this in a new file

ericahinkleRH:
> @Nexucis 
I was told by the team that at Red Hat, we've had users request the ability to export data to a CSV file so they can perform additional analysis outside the application. Additionally, I see what you are saying about the hamburger menu so for now I'll just remove it and replace it with the export button instead and we can always change it in the future if more functionality is added! Thank you for your comments and I will look into plugins and the development environment.

jgbernalp:
@ericahinkleRH I created a discussion to align on the interface design for this: https://github.com/perses/perses/discussions/3053

PR #3046, opened 2025-06-26
[ignore] Introduce a new interface `Authorisation`

Nexucis:
This new interface will help to put in place a way to delegate the authorisation layer to a third party.


Nexucis:
this will help to move forward on the PR #2940

Issue #2840, opened 2025-04-01
"Execute" button in Explore page


ibakshay:
### What did you do?

When using the Explore interface, there is no dedicated "Execute" button to run queries. Currently, queries only execute when clicking on random parts of the page, which is unintuitive and difficult for users.

![Image](https://github.com/user-attachments/assets/0b873e8f-6c59-4812-8200-23603f031d54)

### What did you expect to see?

There should be a clearly visible "Execute" button similar to Prometheus's interface, allowing users to explicitly run their queries after entering them.
I will be happy to contribute if this indeed is a bug :) 

### What did you see instead? Under which circumstances?

There is no `Execute button`.I would be happy to create a PR for this issue if it's confirmed as a bug.

### Perses configuration file

```yaml

```

### Logs

```text

```

AntoineThebaud:
I wouldn't classify this as a bug, we reuse the same editor component here than for panels in dashboards where there's also no Execute button.
I agree such button would be a good addition for the Explore view, not so sure about dashboard panel editor but, to be analyzed

sinkingpoint:
I'm looking to get back into Perses stuff, so I'll take a look at this @AntoineThebaud

sinkingpoint:
OK, this might b e a bit more difficult than I imagined. Because the explore page/query editor is actually defined in the individual plugins, there's a few approaches here:

- Make every query editor have an Execute button
- Add in a prop to the queryeditor component that adds the explicit run button, which then has to be enabled in each plugin
- Add in a new provider (the existing `ExploreManagerProvider` from the UI package can't be accessed from the components package without a gnarly circular dependency) that can be accessed by the `QueryEditor` to add an explicit run button only on Explore pages

Thoughts @AntoineThebaud ?

AntoineThebaud:
Well if that makes things easier I guess having an Execute button everywhere (thus in the dashboard panel editor too) is fine, I had no strong opinion on that and it looks like nobody else had this concern 😄

PR #3049, opened 2025-06-27
[ENHANCEMENT] Add Execute Query Button to query editors

sinkingpoint:
# Description

Solves #2840

Currently when you edit a query it runs whenever there is a re-render done (generally when you click off of the editor box). This leads to a weird experience if you're not used to that, where your query seemingly runs at random. Here we add a new button to Query Editors that gives users the ability to manually run their query, making this flow more explicit.

# Screenshots

image
image

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
+++ b/ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
@@ -11,7 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Box } from '@mui/material';
+import { Box, Button } from '@mui/material';
+import Reload from 'mdi-material-ui/Reload';
 import { ErrorAlert, ErrorBoundary } from '@perses-dev/components';
 import { ReactElement } from 'react';
 import { PluginKindSelect } from '../PluginKindSelect';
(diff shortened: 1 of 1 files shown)


Nexucis:
@Stevelening this button will be on every explorer. It will be interesting to know if it will be fine / if it makes sense to have it in the profiling explorer.

Stevelening:
It does not make sense to have this button in the profiling explorer because there is no query to edit or run.

andreasgerstmayr (approved):

andreasgerstmayr:
Oh sorry, I didn't read the last comment before clicking the approve button.

> It does not make sense to have this button in the profiling explorer because there is no query to edit or run.

If there is no query, how does the data source (pyroscope, etc.) know which profile to fetch?

Gladorme (approved):
> It does not make sense to have this button in the profiling explorer because there is no query to edit or run.

You probably mean freetext query. So yeah, every change trigger the query.

I am good with this feature, what about next: should we force user to use this button to run the query and if they change something in plugin editor, it will not retrigger query until this button clicked or we keep the current hybrid implem: where you can run query with the button and when the plugin editor has changes

nicolastakashi (approved):

Stevelening:
> > It does not make sense to have this button in the profiling explorer because there is no query to edit or run.
> 
> You probably mean freetext query. So yeah, every change trigger the query in Pyroscope datasources.
> 
> I am good with this feature, what about next: should we force user to use this button to run the query and if they change something in plugin editor, it will not retrigger query until this button clicked or we keep the current hybrid implem: where you can run query with the button and when the plugin editor has changes

Exactly 👍

Nexucis (approved):
great. Thank you all for your feedback !

Merging this one since we already have 4 approvals that's more than enough :)

Nexucis:
thanks you @sinkingpoint for doing it ! Cool to see you back around :)

PR #3061, opened 2025-07-01
[FEATURE] Add SQLProxy Datasource proxy kind 

### LINKED ISSUE #3047: Add SQL Proxy for Datasources
### Proposal

The idea is to add the ability to proxy SQL queries like we do for HTTP requests now.

dougkirkley:
# Description

Adds an SQLProxy Datasource proxy kind to allow SQL datasources to be proxied using the Perses backend.

Closes #3047

# Checklist


cue/common/proxy/sql.cue
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package proxy
+
+#SQLProxy: {
+	kind: "SQLProxy"
+	spec: {
+		driver: "mysql" | "mariadb" | "postgres"
```
Nexucis: I would assume that the driver `mysql` and `mariadb` is the same. At least I know that the driver `mysql` is working when connecting Perses to `mariadb`

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,112 @@
...
+	if s.Host == "" {
+		return errors.New("host cannot be empty")
+	}
+
+	if s.Database == "" {
+		return errors.New("database cannot be empty")
+	}
+
+	if s.Username == "" {
+		return errors.New("username cannot be empty")
```
Nexucis: actually the username can be empty. Protecting a database is not mandatory even if it will be crazy not do it

pkg/model/api/v1/datasource/http/http_test.go
```diff
@@ -1,285 +0,0 @@
...
-	"net/http"
-	"net/url"
-	"testing"
-
-	"github.com/perses/perses/pkg/model/api/v1/common"
-	"github.com/stretchr/testify/assert"
-	"gopkg.in/yaml.v3"
-)
-
-func TestUnmarshalJSONConfig(t *testing.T) {
```
Nexucis: I don't think these tests are not relevant anymore, on the contrary. Can you restore them please ? And have the same kind for the SQL proxy please ?

internal/api/impl/proxy/proxy.go
```diff
@@ -16,22 +16,31 @@ package proxy
...
 	"strings"
 	"time"
 
+	"github.com/go-sql-driver/mysql"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/jackc/pgx/v5/stdlib"
+
 	"github.com/perses/perses/internal/api/authorization"
 	"github.com/perses/perses/pkg/model/api/config"
+
```
Nexucis: the import are not sorted properly. We should only have two blocs, one for the native import, one for others

internal/api/impl/proxy/proxy.go
```diff
@@ -161,24 +171,22 @@ type proxy interface {
 	serve(c echo.Context) error
 }
 
-func newProxy(spec v1.DatasourceSpec, path string, crypto crypto.Crypto, retrieveSecret func(name string) (*v1.SecretSpec, error)) (proxy, error) {
-	cfg, err := datasourceHTTP.ValidateAndExtract(spec.Plugin.Spec)
+func newHTTPProxy(spec v1.DatasourceSpec, path string, crypto crypto.Crypto, retrieveSecret func(name string) (*v1.SecretSpec, error)) (proxy, error) {
```
Nexucis: if you keep the generic function `newProxy`, and if the function `ValidateAndExtract` is returning the config as an interface and the kind of the proxy, then you can simplify a bunch of things.

- It would avoid to modify the files `projectdatasource.go` `localdatasource.go` and `globaldatasource.go`
- It will avoid to change too much the extraction logic from the function `ValidateAndExtract`.

pkg/model/api/v1/secret/tls_config.go
```diff
@@ -52,9 +52,9 @@ func NewPublicTLSConfig(t *TLSConfig) *PublicTLSConfig {
 	}
 }
 
-func BuildTLSConfig(cfg *TLSConfig) (*tls.Config, error) {
+func BuildPromTLSConfig(cfg *TLSConfig) (*promConfig.TLSConfig, error) {
```
Nexucis: Why do you need to add this function ?

pkg/model/api/v1/datasource/http/test/password_file.txt
```diff
@@ -1 +0,0 @@
-my awesome and unnecessary long password
```
Nexucis, line 1: This file was not used anymore ?

internal/cli/cmd/dac/setup/testdata/go.mod
```diff
@@ -1,3 +1,5 @@
 module example.com
 
 go 1.23.3
+
+require github.com/perses/perses v0.50.0 // indirect
```
Nexucis: this change is weird

internal/api/validate/validate.go
```diff
@@ -109,8 +110,11 @@ func validateVariableNames(variables []dashboard.Variable) error {
 }
 
 func validateDatasourcePlugin(plugin common.Plugin, name string, sch schema.Schema) error {
-	if _, err := http.ValidateAndExtract(plugin.Spec); err != nil {
```
Nexucis, line 112: if `ValidateAndExtract` is still returning the config + kind like I have suggested in a different comment, then you don't need to change this function. And it will avoid to have a switch everywhere managing every datasource proxy plugin kind.

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+type Config struct {
+	Driver Driver `json:"driver" yaml:"driver"`
+	// Host is the hostname required to contact the datasource
+	Host string `json:"host" yaml:"host"`
+	// Database is the database for the datasource
+	Database string `json:"database" yaml:"database"`
+	// Username is the username for the datasource
+	Username string `json:"username" yaml:"username"`
```
Nexucis: since the username is already setup in the secret with `basic_auth`. We should use this one there.

docs/api/datasource.md
```diff
@@ -117,6 +117,16 @@ datasource is used.
       url= '/proxy/globaldatasources/' + datasource.metadata.name 
   ```
 
+When using the SQLProxy proxy kind:
+
+The method needs to be `POST` and the path needs to be `/query`
```
Nexucis: it's more the body rather than the path that should be used to pass the query right ?

docs/concepts/proxy.md
```diff
@@ -28,6 +28,14 @@ Depending on the datasource scope (Global, Project, Local), the URL to access th
 /proxy/projects/<project-name>/dashboards/<dashboard-name>/datasources/<datasource-name>/<datasource_api_path>
 ```
 
+4. For the SQLProxy proxy kind, the URL to access the datasource is:
+
+The method needs to be `POST` and the path needs to be `/query`
+
+```text
+/proxy/projects/<project-name>/dashboards/<dashboard-name>/datasources/<datasource-name>/query
```
Nexucis: that's not the case since the query has to be in the body

Nexucis: and here the documentation should make a different between Perses serving as HTTP proxy, and then making the SQL requests for the frontend. That's quite different. In that last case, Perses is not used as an actual proxy since it doesn't forward the request based on a configuration. It actually performs the query

internal/api/discovery/cuetils/node.go
```diff
@@ -20,6 +20,8 @@ import (
 	"strings"
 
 	"cuelang.org/go/cue"
+
```
Nexucis: this new line should not exist. We only make a difference between the native package and others

internal/api/e2e/api/proxy_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/gavv/httpexpect/v2"
+
```
Nexucis: same here, resort the imports

internal/api/impl/proxy/proxy.go
```diff
@@ -379,3 +387,250 @@ func (h *httpProxy) prepareTLSConfig() (*tls.Config, error) {
...
+	config   *datasourceSQL.Config
+	secret   *v1.SecretSpec
+	path     string
+	password string
+}
+
+func (s *sqlProxy) serve(c echo.Context) error {
+	r := c.Request()
+
+	if s.path != "query" {
```
Nexucis: since the query is found in the body, we don't need to check the path. And besides it will simplify the pre-requisite when calling Perses.

internal/api/impl/proxy/proxy.go
```diff
@@ -379,3 +387,250 @@ func (h *httpProxy) prepareTLSConfig() (*tls.Config, error) {
...
+		return apiinterface.HandleBadRequestError("SQL proxy only supports the '/query' path")
+	}
+
+	// if this isn't a POST request don't perform the SQL query
+	if r.Method != http.MethodPost {
+		return echo.NewHTTPError(http.StatusMethodNotAllowed, fmt.Sprintf("you are not allowed to use this endpoint %q with the HTTP method %s", s.path, r.Method))
+	}
+
+	q := &sqlQuery{}
```
Nexucis, line 442: probably at some point we will need to read/parse/analyse the query to ensure Perses cannot be used in order to write data in the targeted database.

That's not something we should do in this PR, but something to think for the next one

internal/api/impl/proxy/proxy.go
```diff
@@ -379,3 +387,250 @@ func (h *httpProxy) prepareTLSConfig() (*tls.Config, error) {
...
+		logrus.WithError(err).Error("unable to execute the query")
+		return apiinterface.InternalError
+	}
+	defer func(rows *sql.Rows) {
+		if err = rows.Close(); err != nil {
+			logrus.WithError(err).Error("unable to close rows")
+		}
+	}(rows)
+
+	cols, err := rows.Columns()
```
Nexucis: for readability, would be nice to put the rest of the code starting here in a dedicated function like `decodeQuery(c echo.Context, rows)`. Maybe not the best name, any suggestions would likely be better

internal/api/impl/proxy/proxy.go
```diff
@@ -379,3 +387,250 @@ func (h *httpProxy) prepareTLSConfig() (*tls.Config, error) {
...
+	if s.secret == nil {
+		return nil, nil
+	}
+	return secretModel.BuildTLSConfig(s.secret.TLSConfig)
+}
+
+// SQLOpen opens a database specified by its database driver in the address
+func (s *sqlProxy) sqlOpen(tlsConfig *tls.Config) (*sql.DB, error) {
+	switch s.config.Driver {
+	case datasourceSQL.DriverMySQL, datasourceSQL.DriverMariaDB:
```
Nexucis: For each driver, it will be nice to have the creation of the database connection in a dedicated function

internal/api/impl/proxy/proxy.go
```diff
@@ -379,3 +387,250 @@ func (h *httpProxy) prepareTLSConfig() (*tls.Config, error) {
...
+			Addr:   s.config.Host,
+			DBName: s.config.Database,
+		}
+
+		if s.password != "" {
+			mysqlConfig.Passwd = s.password
+		}
+
+		if tlsConfig != nil {
+			tlsConfigName := "perses-tls"
```
Nexucis: I would not name it like that. Would be better to replace `perses` with the `{datasource_kind}-{datasource-project-name-?{datasource-name}-tls`

Nexucis:
Hey @dougkirkley.

Thank you for taking care about this part. I think I have reviewed most of the codes and raised various and more or less serious points.

Let me know if you have questions.

internal/api/impl/proxy/proxy.go
dougkirkley: @Nexucis the reason I added this piece was because if you don’t specify a path then you get a 404 at the datasource path, so I was just thinking that this could just have implementations be the same. Maybe this could be fixed by changing the Echo endpoints instead?

Nexucis: indeed it returns a 404 which shouldn't be the case ... I looked at why it returned this error and it seems it's not routed to the correct endpoint. I am trying to understand why.

internal/api/impl/proxy/proxy.go
dougkirkley: I created a function for each driver and then just had the existing `sqlOpen` function call those inside the switch statement. Let me know if that is not what you wanted.

internal/api/impl/proxy/proxy.go
dougkirkley: I was able to get this to work by adding an extra POST endpoint pointing to just the datasource URL without any `*`

pkg/model/api/v1/datasource/proxy.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (c *configFinder) unmarshalConfig(spec reflect.Value) {
+	if !spec.IsValid() {
+		return
+	}
+
+	spec = getNextElem(spec)
+	var data []byte
+	data, c.err = json.Marshal(spec.Interface())
```
Nexucis, line 114: here we know the kind of config we need to un-marshal. So instead of marshalling and un-marshalling a generic struct, we can have the switch based on kind and unmarshal the correct config.

Doing that will also avoid to marshal and unmarshal twice the same data. A process that might have a cost when requesting 50 times per second the datasources from the server.

pkg/model/api/v1/datasource/proxy.go
```diff
@@ -0,0 +1,166 @@
...
+	return "", false
+}
+
+// getKindFromMap extracts the string value of the 'kind' field from a map, if it exists.
+func getKindFromMap(v reflect.Value) (string, bool) {
+	for _, key := range v.MapKeys() {
+		if strings.ToLower(key.String()) == ProxyKindField {
+			value := v.MapIndex(key)
+			value = getNextElem(value)
+			if value.Kind() == reflect.String {
```
Nexucis, line 151:
```suggestion
if value.Kind() == reflect.String {
  kindValue := strings.ToLower(value.String())
  if kindValue == http.HTTPProxyKindName || kindValue == sql.SQLProxyKindName {
    return kindValue, true
  }
}
```

Nexucis: otherwise, there is no guarantee we are looking at the correct proxy config

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+type Proxy struct {
+	Kind string `json:"kind" yaml:"kind"`
+	Spec Config `json:"spec" yaml:"spec"`
+}
+
+const (
+	SQLProxyKindName = "sqlproxy"
```
Nexucis:
```suggestion
ProxyKindName = "sqlproxy"
```

I am suggesting that as the prefix `SQL` is redundant because the package is named like that. And calling `sql.ProxyKindName` is more elegant :)

pkg/model/api/v1/datasource/http/http_test.go
```diff
@@ -19,9 +19,10 @@ import (
 	"net/url"
 	"testing"
 
-	"github.com/perses/perses/pkg/model/api/v1/common"
 	"github.com/stretchr/testify/assert"
 	"gopkg.in/yaml.v3"
+
```
Nexucis: no new line in import

pkg/model/api/v1/datasource/http/http.go
```diff
@@ -124,145 +122,5 @@ type Proxy struct {
 }
 
 const (
-	HTTPProxyKindField = "kind"
-	HTTPProxyKindName  = "httpproxy"
-	HTTPProxySpec      = "spec"
+	HTTPProxyKindName = "httpproxy"
```
Nexucis:
```suggestion 
ProxyKindName = "httpproxy"
```
same reason than for the package sql

internal/api/impl/proxy/proxy.go
```diff
@@ -379,3 +427,252 @@ func (h *httpProxy) prepareTLSConfig() (*tls.Config, error) {
...
+		return nil
+	}
+
+	basicAuth := s.secret.BasicAuth
+	if basicAuth != nil {
+		password, err := basicAuth.GetPassword()
+		if err != nil {
+			return err
+		}
+		s.password = password
```
Nexucis, line 505: you need to add `s.username= basicauth.Username` since the username is now set thanks to basic_auth

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	*s = tmp
+	return nil
+}
+
+func (s *Config) validate() error {
+	if s.Driver == "" {
+		return errors.New("driver is required")
+	}
+
```
Nexucis, line 151: here you need to check if the driver is known / supported

For example: 

```go
if s.Driver != DriverMySQL && s.Driver != DriverPostgreSQL {
   return fmt.Errorf("driver %s is not supported", s.Driver)
}
```

pkg/model/api/v1/datasource/sql/sql_test.go
```diff
@@ -0,0 +1,82 @@
+// Copyright 2021 The Perses Authors
```
Nexucis:
```suggestion
// Copyright 2025 The Perses Authors
```

internal/api/impl/proxy/proxy.go
Nexucis: yes that's exactly what I meant. That's really better, thank you!

internal/api/impl/proxy/proxy.go
Nexucis: oh yeah, great idea ! I was thinking the same last night and I was wondering if it would work. That's perfect !

Nexucis:
@dougkirkley thank you for considering my suggestions. I have reviewed again the changes and added few more comments (nothing critical). It is far better and we are in the great direction to be able to merge it !

dougkirkley:
@Nexucis They are all good suggestions! Especially after being too close to it to see the flaws, it has been great.

pkg/model/api/v1/datasource/proxy.go
```diff
@@ -0,0 +1,177 @@
...
+		c.err = json.Unmarshal(data, c.config)
+	}
+}
+
+// getKindFromStruct extracts the string value of the 'kind' field from a struct, if it exists.
+func getKindFromStruct(v reflect.Value) (string, bool) {
+	field := v.FieldByNameFunc(func(name string) bool {
+		return strings.ToLower(name) == ProxyKindField
+	})
+	if field.IsValid() && field.CanInterface() && field.Kind() == reflect.String {
```
Nexucis, line 136: same as what you change in `getKindFromMap`

```go
if field.IsValid() && field.CanInterface() && field.Kind() == reflect.String {
	kindValue := strings.ToLower(field.String())
	if kindValue == http.ProxyKindName || kindValue == sql.ProxyKindName {
		return kindValue, true
	}
}
```

pkg/model/api/v1/datasource/proxy.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	switch c.foundKind {
+	case http.ProxyKindName:
+		c.config = &http.Config{}
+		c.err = json.Unmarshal(data, c.config)
+	case sql.ProxyKindName:
+		c.config = &sql.Config{}
+		c.err = json.Unmarshal(data, c.config)
+	}
```
Nexucis, line 128: I think it worths to add a default cases: 

```go
default:
   c.err = fmt.Errorf("proxy %q not managed", c.foundKind)
```

internal/api/impl/proxy/proxy.go
```diff
@@ -161,36 +173,57 @@ type proxy interface {
...
+
 	var scrt *v1.SecretSpec
-	if len(cfg.Secret) > 0 {
-		scrt, err = retrieveSecret(cfg.Secret)
-		if err != nil {
-			return nil, err
+
+	switch kind {
+	case datasourceHTTP.ProxyKindName:
+		if len(cfg.(*datasourceHTTP.Config).Secret) > 0 {
```
Nexucis: instead of casting everytime the config, you should create a new variable containing the casted config. Will be easier to read

internal/api/discovery/cuetils/node.go
```diff
@@ -110,10 +111,10 @@ func (n *Node) doesKindExistWithSpecIndex() (bool, int, int) {
 	kindIndex := -1
 	specIndex := -1
 	for i, node := range n.Nodes {
-		if node.FieldName == http.HTTPProxyKindField && strings.ToLower(node.ConcreteValue) == http.HTTPProxyKindName {
+		if node.FieldName == datasource.ProxyKindField && strings.ToLower(node.ConcreteValue) == http.ProxyKindName {
```
Nexucis, line 114: for the datasource discovery, it requires more changes than that. But I think it's fine for the moment, I will handle this part on my side as it's not that easy to handle.

Nexucis:
Last thing: I am wondering if we should already define a config for postgresql and one for mysql. The SSL mode is only used for postgresql, and we are missing a lot of option for mysql and probably also for postgresql.

dougkirkley:
@Nexucis Are you thinking just adding specific fields for `MySQL` and `Postgres` or different `Config` structs completely?

something like this maybe?

```
spec:
  host: localhost:5432
  database: mydb
  postgres:
    ssl_mode: verify
```

Nexucis:
yeah that sounds right @dougkirkley 🙏

dougkirkley:
@Nexucis I added the driver specific configs, let me know what you think.

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,125 @@
...
+type PostgresConfig struct {
+	// MaxConns is the maximum size of the pool
+	MaxConns int32 `json:"max_conns,omitempty" yaml:"max_conns,omitempty"`
+	// ConnectTimeout the timeout value used for socket connect operations.
+	ConnectTimeout time.Duration `json:"connect_timeout,omitempty" yaml:"connect_timeout,omitempty"`
+	// PrepareThreshold specifies the number of PreparedStatement executions that must occur before the driver begins using server-side prepared statements.
+	PrepareThreshold *int `json:"prepare_threshold,omitempty" yaml:"prepare_threshold,omitempty"`
+	// SSLMode to use when connecting to postgres
+	SSLMode SSLMode `json:"ssl_mode,omitempty" yaml:"ssl_mode,omitempty"`
```
Nexucis: I think it worths to add a check on the value of `SSLMode` when it is no empty. Just to ensure there is no other value than the one listed above.

pkg/model/api/v1/datasource/proxy_test.go
```diff
@@ -0,0 +1,126 @@
...
+// limitations under the License.
+
+package datasource
+
+import (
+	"github.com/perses/perses/pkg/model/api/v1/common"
+	"github.com/perses/perses/pkg/model/api/v1/datasource/http"
+	"github.com/perses/perses/pkg/model/api/v1/datasource/sql"
+	"net/url"
+	"testing"
```
Nexucis, line 18: import need to be sorted here

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,125 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sql
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/go-sql-driver/mysql"
```
Nexucis: import need to be sorted here

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,125 @@
...
+	SSLModeAllow      SSLMode = "allow"
+	SSLModePreferable SSLMode = "prefer"
+	SSLModeRequire    SSLMode = "require"
+	SSLModeVerifyFull SSLMode = "verify-full"
+	SSLModeVerifyCA   SSLMode = "verify-ca"
+)
+
+type PostgresConfig struct {
+	// MaxConns is the maximum size of the pool
+	MaxConns int32 `json:"max_conns,omitempty" yaml:"max_conns,omitempty"`
```
Nexucis: in the datamodel, we are following the Kubernetes convention, so every key should follow the camel case convention

For example:

```suggestion
MaxConns int32 `json:"maxConns,omitempty" yaml:"maxConns,omitempty"`
```

It's different in the configuration, we are following the snake_case because it helps to determinate the environment variable that can be used to override the configuration.

pkg/model/api/v1/datasource/proxy.go
```diff
@@ -0,0 +1,183 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package datasource
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/perses/perses/pkg/model/api/v1/datasource/http"
+	"github.com/perses/perses/pkg/model/api/v1/datasource/sql"
```
Nexucis: import needs to be sorted

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,125 @@
...
+	Driver Driver `json:"driver" yaml:"driver"`
+	// Host is the hostname required to contact the datasource
+	Host string `json:"host" yaml:"host"`
+	// Database is the database for the datasource
+	Database string `json:"database" yaml:"database"`
+	// Secret is the name of the secret that should be used for the proxy or discovery configuration
+	// It will contain any sensitive information such as username, password, token, certificate.
+	Secret string `json:"secret,omitempty" yaml:"secret,omitempty"`
+	// MySQL specific driver config
+	MySQL *mysql.Config `json:"mysql,omitempty" yaml:"mysql,omitempty"`
```
Nexucis: Actually you can't use directly the `mysql.Config` struct as it doesn't provide the JSON and YAML tags. You will need to provide another one like we did for the config.

You can copy the one from the config and remove `username` `password` and that should do the work

Nexucis: you will have to change the case of the tags too to follow the camelCase convention

Nexucis:
almost good @dougkirkley ! Once my last comments are fixed, I think we will be good to merge this one :)

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,145 @@
...
+	SSLModeAllow      SSLMode = "allow"
+	SSLModePreferable SSLMode = "prefer"
+	SSLModeRequire    SSLMode = "require"
+	SSLModeVerifyFull SSLMode = "verify-full"
+	SSLModeVerifyCA   SSLMode = "verify-ca"
+)
+
+type MYSQLConfig struct {
+	Params           map[string]string `json:"params,omitempty" yaml:"params,omitempty"`
+	MaxAllowedPacket int               `json:"maxAllowedPacket_packet,omitempty" yaml:"maxAllowedPacket,omitempty"`
```
Nexucis:
```suggestion
MaxAllowedPacket int `json:"maxAllowedPacket,omitempty" yaml:"maxAllowedPacket,omitempty"`
```

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,145 @@
...
+
+	if s.Host == "" {
+		return errors.New("host cannot be empty")
+	}
+
+	if s.Database == "" {
+		return errors.New("database cannot be empty")
+	}
+
+	if s.Postgres.SSLMode != "" {
```
Nexucis: this thing should be done in the Unmarshal (for YAML and JSON) function of the`PostgresConfig`

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -0,0 +1,145 @@
...
+const (
+	SSLModeDisable    SSLMode = "disable"
+	SSLModeAllow      SSLMode = "allow"
+	SSLModePreferable SSLMode = "prefer"
+	SSLModeRequire    SSLMode = "require"
+	SSLModeVerifyFull SSLMode = "verify-full"
+	SSLModeVerifyCA   SSLMode = "verify-ca"
+)
+
+type MYSQLConfig struct {
```
Nexucis:
```suggestion

type MySQLConfig struct {
```

pkg/model/api/v1/datasource/sql/sql.go
Nexucis: when I said that, I wasn't saying to put it in the Unmarshal function related to the `Config` struct, but actually related to the `PostgresConfig`

So actually you need to implement the function for the struct `PostgresConfig` and add in the validate method what you did. Exactly like you did for the Config but for PostgresConfig.

Does it make sense for you @dougkirkley ?

dougkirkley: yep, that makes sense to me

dougkirkley: @Nexucis okay, updated

PR #3068, opened 2025-07-07
[FEATURE]: Decouple datasource api from datasource datasource provider

shahrokni:
# Description

This is the **first step** towards decoupling the data source API from its provider. 

## Background
As you know Perses dashboards, panels, components, etc. could be used by other applications. This means although they can be fed by Perses dedicated data sources, their independence from Perses (core notions like Global, users, projects, etc) should be fully maintained. One of the problems that we have at the moment is that the **ds provider** and the **Perses APP API** are coupled. While it makes sense for APP API to be aware of Perses internal notions, the provider should not know about its consumer. 

## Solution - First Step

1.  Introduce a new generic data source interface which could support the existing ones
2. Cut off the dependency from API and provider 
3. Pass the result of API directly to the provider

## Acceptance Criteria
If everything goes well, the following views should work as before, but independent to the APP API

1. Explore View
2. Dashboard View
3. Variable View

image

image

image

# Checklist

## UI Changes


ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -49,22 +49,11 @@ export type BuildDatasourceProxyUrlParams = {
...
- * The external API for fetching datasource resources
- */
-export interface DatasourceApi {
-  buildProxyUrl?: BuildDatasourceProxyUrlFunc;
-  getDatasource: (project: string, selector: DatasourceSelector) => Promise<DatasourceResource | undefined>;
-  getGlobalDatasource: (selector: DatasourceSelector) => Promise<GlobalDatasourceResource | undefined>;
-  listDatasources: (project: string, pluginKind?: string) => Promise<DatasourceResource[]>;
-  listGlobalDatasources: (pluginKind?: string) => Promise<GlobalDatasourceResource[]>;
-}
-
```
shahrokni, line 62: Dear reviewer,

Please make sure this is a safe change. (Some dependency somewhere in other repos that I am not aware of?)
I tested it myself and everything was fine. Please check if I am missing something.
This block should be moved to app. From now on every consumer is responsible to fetch the data from the relevant ds and feed it to the DS provider.

shahrokni:
@Nexucis :)

ui/app/src/model/datasource-api.ts
```diff
@@ -114,11 +128,17 @@ export function useDatasourceApi(): DatasourceApi {
     [globalDatasources, isGlobalDatasourcesPending]
   );
 
+  const listAllDatasources = useCallback(() => {
```
Gladorme: I don't think we need to keep DatasourceApi anymore. And it will reduce:
```tsx
const { listAllDatasources, buildProxyUrl } = useDatasourceApi();

  const allResources = useMemo(() => {
    return listAllDatasources();
  }, [listAllDatasources]);
```

to

```tsx
const {data: datasources} = useDatasourceList()
const {data: globalDatasources} = useGlobalDatasourcesList()
const allDatasources = useMemo(() => [...datasources, ...globalDatasources], [datasources, globalDatasources])
```

and buildProxyUrl can be imported directly

Gladorme: If used at multiple place => create new hook in datasource-api 😄

ui/app/src/model/datasource-api.ts
```diff
@@ -38,87 +37,22 @@ export function buildProxyUrl({
   return `${basePath}/proxy/${url}`;
 }
 
-export function useDatasourceApi(): DatasourceApi {
-  const { data: globalDatasources, isLoading: isGlobalDatasourcesPending } = useGlobalDatasourceList();
-  const { data: datasources, isLoading: isDatasourcesPending } = useDatasourceList({});
+export const useAllDatasourceResources = (): GenericDatasourceResource[] => {
```
Gladorme: I think we can add project as optional option and use it in `useDatasourceList`, because we don't want all datasources for Perses `HelperDasboardView` 😄

ui/app/src/views/projects/explore/ProjectExploreView.tsx
```diff
@@ -20,7 +20,7 @@ import React, { ReactElement, useMemo } from 'react';
 import { useGlobalVariableList } from '../../../model/global-variable-client';
 import { useVariableList } from '../../../model/variable-client';
 import { buildGlobalVariableDefinition, buildProjectVariableDefinition } from '../../../utils/variables';
-import { useDatasourceApi } from '../../../model/datasource-api';
+import { buildProxyUrl, useAllDatasourceResources } from '../../../model/datasource-api';
 
 export interface ProjectExploreViewProps {
```
Gladorme, line 25: In a next PR (if you want, else I will do it), you can move `ProjectExploreView` out of `views/project` to `views/explore`, because it's not related to a particular project

shahrokni: Sure. I will do it in Step 2.

Gladorme (approved):
LGTM

ui/dashboards/src/context/DatasourceStoreProvider.tsx
```diff
@@ -49,22 +49,11 @@ export type BuildDatasourceProxyUrlParams = {
...
 /**
  * A `DatasourceContext` provider that uses an external API to resolve datasource selectors.
  */
 export function DatasourceStoreProvider(props: DatasourceStoreProviderProps): ReactElement {
-  const { projectName, datasourceApi, onCreate, children } = props;
+  const { projectName, datasources, buildProxyUrl, onCreate, children } = props;
```
jgbernalp, line 56: Why did we change the datasourceApi injection into a collection of datasources? The API allows to query the datasources lazily and uses the filtering on request rather than having to load all the datasources and do the filtering in the frontend.

Gladorme: We wanted to simplify embedded usage for be able to use it for on single panel or not dashboard resources (like explorer) + remove Perses specific notions. IMO, it's easier for to give provider the definition directly and did not think about lazy, mb here 😓

Issue #3059, opened 2025-07-01
Data store provider re-factor


shahrokni:
### Proposal

We should refactor the Data Store Provider due to the following issue.

Right now the Data Store Provider  is tightly coupled to the Perses. **The question is, what if we want to reuse some Perses components which need to be fed by a data source in other projects?**  #3058 

Some of Perses components may be used by other projects. Components wrapped by Data Store Provider should not be tightly coupled with Perses. Otherwise, they are not reusable. Notions including user, projects, global, etc are only recognized by Perses, and not the consumer projects. Therefore, Data store provider should be independent and be fed by the consumer itself.

![Image](https://github.com/user-attachments/assets/fda3f045-e3f5-45a6-a4a9-2d6df2f5179e)

### Example

1. TODO APP is using Perses Bar chart
2. Bar chart needs data source to render information
3. TODO provides the data source which can be recognized by Perses Bar chart
4. Bar chart is already wrapped by the data source provider
5. TODO passes down the data source to the provider
6. Provider gives them to the bar chart

### Tricky part

Explained by @Gladorme 
There should be a data source priority logic.
If there are data source with identical name:

1. Project data source overrides Global one
2. Local data source overrides the Project one 

(Lowest Priority) Global > Project > Local (Highest Priority)

> Yep, this is the logic I would like to have for the provider. The only thing tricky is: we still want to keep Perses app datasource "priority". When you select a datasource with DatasourceSelector in a panel query. If a datasource have a same name, a global datasource can be overridden by a project datasource and a project datasource can be overridden by a local (dashboard) datasource. So we will need a generic logic to handle this too

shahrokni:
@Gladorme and @Nexucis 

Please assign this issue to me. Thank you.

jgbernalp:
Not sure if this is an issue. The data store provider receives an API that make requests to a backend (perses API) to list types of datasources. Components are meant to be reusable for frontends with different requirements and structure, but having always a Perses backend. If only the data visualization part is needed to be reusable, the panel components have a property to add the raw data. But other components like datasources are meant to know the global, project or local scope concepts.

PR #3094, opened 2025-07-15
[BREAKINGCHANGE] DatasourceProvider: decouple datasource provider and dashboard datasource

shahrokni:
# Description

Relates to #3059 
This PR removes the the `dashboardResource` prop from the data source provider. The `dashboardResource` prop used to be a way to override the persisted datasources which would be served from the provider.  This idea used to be useful, because it gave the developers the opportunity to temporarily override the current datasources and test and use them in the dashboards. 

However, from the provider point of view, a datasource should always be a datasource regardless of its origin. That is why we came up with the idea of the Generic Data Source (@Gladorme) This idea keep the provider super flexible and this means the components which are wrapped by this provider can be used in other applications. Therefore, the notion of dashboardResource should be removed. 

To achieve this goal, the PR removes the mentioned prop, but it provides a functionality by which a temporary set of data sources can be served along the persisted datasouces. 

# Test

> :warning:
>Please Test Carefully!

Please check the following components or views and make sure the data source provider is working fine

- [ ] Explorer View - You should be able to run your query over a specific valid datasource
- [ ] Variable Drawers - You should be able to see all valid datasource and select them
- [ ] Dashboard View - Open a valid dashboard connected to a valid ds and check if your panels are working as expected
- [ ] Dashboard => Data source Edit mode - You should be able to edit local ds, if they are valid, they would override others
- [ ] Dashboard => Data source Edit mode => Apply Local Data source
- [ ] Dashboard => Data source Edit mode => Apply Local Data source => Then remove it
- [ ] Dashboard => Data source Edit mode => Apply Local Data source => Save (Should reload if the save is successful)
- [ ]  Dashboard => Data source Edit mode => Apply Local Data source => Save and fully refresh

# Screenshots

image

image

image

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/model/datasource-api.ts
+++ b/ui/app/src/model/datasource-api.ts
@@ -12,19 +12,34 @@
 // limitations under the License.
 
 import { useMemo } from 'react';
-import { GenericDatasourceResource, GenericMetadata } from '@perses-dev/core';
+import {
+  DashboardResource,
+  DatasourceResource,
+  EphemeralDashboardResource,
+  GenericDatasourceResource,
+  GenericMetadata,
+  GlobalDatasourceResource,
+} from '@perses-dev/core';
 import { useDatasourceList } from './datasource-client';
 import { useGlobalDatasourceList } from './global-datasource-client';
 import { getBasePathName } from './route';
+import { useDashboard } from './dashboard-client';
+import { useEphemeralDashboard } from './ephemeral-dashboard-client';
 
 interface DataSourceFilter {
   project?: string;
+  dashboard?: string;
 }
 
 interface ProxyBuilderParam {
   project?: string;
   dashboard?: string;
   name: string;
+}
+
+export interface AllDatasources {
+  datasources: GenericDatasourceResource[];
+  isLoading: boolean;
 }
 
 export const buildProxyUrl = ({ project, dashboard, name }: ProxyBuilderParam): string => {
@@ -44,8 +59,8 @@ export const buildProxyUrl = ({ project, dashboard, name }: ProxyBuilderParam): 
  * to filter the projects datasources
  * @param filter Filters the project datasources
  */
-export const useAllDatasourceResources = (filter?: DataSourceFilter): GenericDatasourceResource[] => {
-  const { project } = filter || {};
+export const useAllDatasourceResources = (filter?: DataSourceFilter): AllDatasources => {
+  const { project, dashboard } = filter || {};
 
   const { data: datasources, isLoading: isDatasourceLoading, error: datasourceError } = useDatasourceList({ project });
   const {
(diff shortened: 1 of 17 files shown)


Nexucis:
@jgbernalp & @andreasgerstmayr perhaps you will be interested by this PR :)

shahrokni:
@Nexucis I would like to have @Gladorme opinion as well, if it is possible :)

ui/app/src/model/datasource-api.ts
```diff
@@ -54,27 +69,96 @@ export const useAllDatasourceResources = (filter?: DataSourceFilter): GenericDat
...
   } = useGlobalDatasourceList();
 
-  if (globalDatasourceError || datasourceError) {
-    throw new Error('Could not fetch all datasources');
-  }
+  const {
+    data: dashboardResource,
+    isLoading: isDashboardRecourseLoading,
+    error: dashboardResourceError,
+  } = useDashboard(project || '', dashboard || '');
```
jgbernalp, line 76: Not sure if we are really decoupling as consumers now need to know that they need a dashboard store first to be able to create a datasource-api.

jgbernalp: injecting a datasource-api allows to lazy load the specific datasources only when they are needed. This approach seems to need all datasources upfront.

shahrokni: @jgbernalp 
By injecting a `DatasourceApi`  into the provider, we risk overlooking an important aspect: the API (in this case, a REST API) communicates with the concept of a "Resource." This raises a critical question: _**what resource are we referring to in this context?**_

Consider the following interface, which is expected from the provider:

```typescript
export interface DatasourceApi {
  buildProxyUrl?: BuildDatasourceProxyUrlFunc;
  getDatasource: (project: string, selector: DatasourceSelector) => Promise<DatasourceResource | undefined>;
  getGlobalDatasource: (selector: DatasourceSelector) => Promise<GlobalDatasourceResource | undefined>;
  listDatasources: (project: string, pluginKind?: string) => Promise<DatasourceResource[]>;
  listGlobalDatasources: (pluginKind?: string) => Promise<GlobalDatasourceResource[]>;
}
```

**_What happens if, in the future, we need to support different types of resources? Or if the concept of "Global" no longer applies in certain cases?_** 

Injecting a `DatasourceApi`  would tightly couple the provider to a specific API implementation, making it harder to adapt to such changes. This coupling would also make maintenance more challenging as the provider would need to evolve alongside the API.

The goal of this provider, along with dashboards and panels, is to remain flexible and reusable. By allowing consumers to inject generic datasources that can be recognized and used by anyone, we ensure that the provider is not tied to a specific API or resource structure. This approach promotes scalability and makes the provider more adaptable to future requirements.

Lazy loading has nothing to do with provider. Provider has a very specific task. 

1. It is a wrapper which is fed by proper data initially 
2. Exposes data to the lower components
3. If lazy loading is a concern, it should be handled by the one who is providing the datasources accordingly
4. For sure, if the datasources  is updated, the provider is updated, and eventually the exposed datasources is updated for everyone

PR #3105, opened 2025-07-21
[ignore] fixing usage of pointer and some neats in the k8s authz implementation

Nexucis:
- all method from the interface `Authorisation` implemented by the struct `k8simpl` should be implemented with a pointer to the struct.
- replacing `user` by `usr` to avoid package collision. Same for `authenticator`
- fixing some neats in the comments.

/cc @PeterYurkovich


PeterYurkovich:
Changes look good to me

PR #3111, opened 2025-07-22
[BUGFIX]: update datasource selector to include datasourcevariable

ibakshay:
This Pull request fixes https://github.com/perses/perses/issues/3084

cc: @OrReuben 

Signed-off-by: Akshay Iyyadurai Balasundaram <akshay.iyyadurai.balasundaram@sap.com>

# Description

# Screenshots

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [ ] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [ ] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [ ] Visual tests are stable and unlikely to be flaky.
  See [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

PR #3112, opened 2025-07-23
[ignore] avoid checking datatracker url in doc as it returns a 403

Nexucis:
datatracker is the website referencing all RFC and it now returns a 403 when GitHub is checking the URL. Perphaps GitHub has been blacklisted


Nexucis:
should help to merge the pr #3111 and #3104

Gladorme (approved):

PR #3115, opened 2025-07-23
[BUGFIX] UI: remove cookies on token refresh fail

Gladorme:
# Description

With #3031 fix, we are refetching access token when back returns code 401. For this we need to keep expired access token in browser to send it, else we have a 400 error. For this the access token cookie max age is equal to refresh token cookie max age.
However, if by any chance the refresh token cookie is deleted and not the access token cookie, the UI will be stuck in bad state (all fetch failing and not redirecting to login page).

This PR force the delete of the cookie in case of refresh token is failing, forcing the user to be redirected to login page.

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/model/fetch.ts
+++ b/ui/app/src/model/fetch.ts
@@ -14,9 +14,11 @@
 import { StatusError } from '@perses-dev/core';
 import { refreshToken } from './auth-client';
 
+const JWT_COOKIES = ['jwtPayload', 'jwtSignature', 'jwtRefreshToken'];
+
 // Delete a cookie by setting its expiration date to the past
 function deleteCookie(name: string): void {
-  document.cookie = name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+  document.cookie = `${name}=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;`;
 }
 
 export function enableRefreshFetch(): void {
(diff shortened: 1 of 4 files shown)


ui/app/src/model/fetch.ts
```diff
@@ -14,9 +14,14 @@
 import { StatusError } from '@perses-dev/core';
 import { refreshToken } from './auth-client';
 
+// Delete a cookie by setting its expiration date to the past
+function deleteCookie(name: string): void {
+  document.cookie = name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+}
```
shahrokni: Using JS literals is recommended. 
 
```typescript
document.cookie = `${name}=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;`;
```

ui/app/src/model/fetch.ts
```diff
@@ -26,6 +31,13 @@ export function enableRefreshFetch(): void {
                 return target.apply(that, args);
               })
               .catch((refreshError: StatusError) => {
+                if (refreshError.status === 400) {
+                  // If refresh token fails, remove jwt cookies
+                  // This will force the user to be redirected to the login page
+                  deleteCookie('jwtPayload');
+                  deleteCookie('jwtSignature');
+                  deleteCookie('jwtRefreshToken');
+                }
```
shahrokni: We can I have a bit of improvement here

```typescript
const JWT_COOKIES = ['jwtPayload', 'jwtSignature', 'jwtRefreshToken'];

export function enableRefreshFetch(): void {
  global.fetch = new Proxy(global.fetch, {
    apply: async function (target, that, args: Parameters<typeof global.fetch>): Promise<Response> {
      return target
        .apply(that, args)
        .then((res) => {
          if (res.status === 401) {
            return refreshToken()
              .then(() => {
                return target.apply(that, args);
              })
              .catch((refreshError: StatusError) => {
                if (refreshError.status === 400) {
                  JWT_COOKIES.forEach(deleteCookie);
                }
                throw refreshError;
              });
          }
          return res;
        })
        .catch((error: StatusError) => {
          throw error;
        });
    },
  });
}
```

Nexucis (approved):

PR #3117, opened 2025-07-24
[FEATURE] Enable role, rolebinding and user only when native authz is used

Nexucis:
Role, RoleBinding and user are here to make the native authorisation working. As such when a third-party authorisation service is used, we don't need to expose these resources

Edit: As a side effect, I had to rework the e2e tests. Indeed, by deactivating the possibility to create role/roleBinding when the native authorisation is not activated, then current tests were not passing anymore as we are creating the tests without any authentication.

This is not the case anymore, every standard tests scenarios are using a server with auth activated. A default user is created "alice" with default role and roleBiding to give her full right on any object provided by the API.

### DIFF
--- a/cmd/perses/main.go
+++ b/cmd/perses/main.go
@@ -57,12 +57,12 @@ func main() {
 	promRegistry := prometheus.NewRegistry()
 	registerMetrics(promRegistry)
 
-	runner, persistentManager, err := core.New(conf, *pprof, promRegistry, banner)
+	runner, dependencyManager, err := core.New(conf, *pprof, promRegistry, banner)
 	if err != nil {
 		logrus.Fatal(err)
 	}
 	defer func() {
-		if daoCloseErr := persistentManager.GetPersesDAO().Close(); daoCloseErr != nil {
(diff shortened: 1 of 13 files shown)


internal/api/impl/v1/globalrole/endpoint.go
```diff
@@ -42,6 +43,11 @@ func NewEndpoint(service globalrole.Service, authz authorization.Authorization,
 func (e *endpoint) CollectRoutes(g *route.Group) {
 	group := g.Group(fmt.Sprintf("/%s", utils.PathGlobalRole))
 
+	if !e.authorizationIsNative {
```
PeterYurkovich: I believe that the globalrole endpoints being disabled has now been done in the `server.go` file so the changes in the globalrole endpoint are no longer needed

Nexucis: ah yes totally. I thought I rollbacked these changes, but it seems not. Thank you for highlighting it

AntoineThebaud (approved):
LGTM

PR #3128, opened 2025-07-25
[BUGFIX] Handle Piechart tooltip overlap with legend

### LINKED ISSUE #2772: PieChart: tooltip is hidden behind the legend.
### What did you do?

I created a Pie Chart Panel in the UI. 

1. The content of the tooltip should not be hiding behind the legend. 

The content of the tooltip is hiding behind the legend. 

Image

### System information

_No response_

### Perses version

0.50.1

### Perses configuration file

```yaml

```

### Logs

```text

```

shahrokni:
Closes #2772 

# Description
For details of the bug please take a look at the related issue
https://github.com/perses/perses/issues/2772#issue-2919900618

## The Bug
image

# About this change

This fix 

1. Removes Overflow hidden behavior of the chart container, **if there is enough room for both legend and the container** (This is the improvement)
2. If there is an overlap, the legend covers the chart. **This has been the default behavior**
3. If there is no enough room for both, the legend is removed. **This also has been the default behavior** 

## Normal Browser size

image

## Smallest possible

image

## Smallest - Legend Scroll

image

## Overlap

> :warning: If overlap happens the legend covers the chart. (This has been the default behavior). Under these circumstances, for sure the tool tip also will be covered!

image

# Checklist

## UI Changes

### DIFF
--- a/ui/components/src/ContentWithLegend/ContentWithLegend.tsx
+++ b/ui/components/src/ContentWithLegend/ContentWithLegend.tsx
@@ -65,7 +65,6 @@ export function ContentWithLegend({
         sx={{
           marginRight: `${margin.right}px`,
           marginBottom: `${margin.bottom}px`,
-          overflow: 'hidden',
         }}
       >
         {typeof children === 'function' ? children({ width: content.width, height: content.height }) : children}
--- a/ui/.prettierrc.json
+++ b/ui/.prettierrc.json
@@ -1,5 +1,5 @@
 {
   "printWidth": 120,
   "singleQuote": true,
-  "trailingComma": "es5"  
+  "trailingComma": "es5"
 }
--- a/ui/components/src/Legend/Legend.tsx
+++ b/ui/components/src/Legend/Legend.tsx
@@ -149,6 +149,7 @@ export function Legend({
           position: 'absolute',
           top: 0,
           right: 0,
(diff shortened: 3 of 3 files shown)


Nexucis (commented):
so cool, thank you for fixing that !

ui/components/src/ContentWithLegend/ContentWithLegend.tsx
```diff
@@ -65,11 +94,12 @@ export function ContentWithLegend({
         sx={{
           marginRight: `${margin.right}px`,
           marginBottom: `${margin.bottom}px`,
-          overflow: 'hidden',
```
Gladorme, line 68: Let's just remove this field `overflow` and it will fix the issue. I don't really see a case where we want tooltip to be hidden.

Gladorme: For the overlap, set a background color to legend (ui/components/src/Legend/Legend.tsx) and it will do the work:
```tsx
  if (options.position === 'right') {
    return (
      <Box
        style={{
          width: width,
          height: height,
        }}
        sx={(theme) => ({
          position: 'absolute',
          top: 0,
          right: 0,
          backgroundColor: theme.palette.background.default,
        })}
      >
        {legendContent}
      </Box>
    );
  }
```
The bg can be set in all position i think 🤔

PR #3129, opened 2025-07-25
[FEATURE] - TimeSeries Export Functionality w/ Plugins

ericahinkleRH:
This PR adds CSV export functionality to time series panels, enabling users to download time series data in CSV format. The implementation includes core changes to support CSV export capabilities and integrates with the CSV export plugin functionality. 

An "Export as CSV" button appears in the top right of any panel that holds appropriate time series charts/data and when clicked downloads a file on your computer with the name "panelName_data.csv". The first line of the CSV holds labels for the data columns and if the graph has legends then those labels are used. Under that are all the rows of the actual data.

## Integration Testing Setup
* This PR is to be used to create a snapshot branch
* Related plugin PR: TBD
* Branch: TimeSeriesFixed2 contains updated changes

## Changes Made
* Added core CSV export functionality for time series data
* Adjusted Panel Tests to reflect core coding changes
* Implemented integration with plugin system for CSV export capabilities

## Screenshots
**Export Button:**
image

**File Name and Information Format:**
image

## Checklist
- [x] Pull request title follows the [<catalog_entry>] <commit message> naming convention using one of the following catalog_entry values: FEATURE, ENHANCEMENT, BUGFIX, BREAKINGCHANGE, DOC, IGNORE.

## UI Changes
- [ ] Code follows the UI guidelines.
- [ ] Visual tests are stable and unlikely to be flaky. See e2e docs for more details. Common issues include:
   * Is the data inconsistent? You need to mock API requests.
   * Does the time change? You need to use consistent time values or mock time utilities.
   * Does it have loading states? You need to wait for loading to complete.

### DIFF
--- a/dev/config.yaml
+++ b/dev/config.yaml
@@ -1,7 +1,7 @@
 security:
   readonly: false
   encryption_key: "=tW$56zytgB&3jN2E%7-+qrGZE?v6LCc"
-  enable_auth: true
+  enable_auth: false
   authorization:
     guest_permissions:
       - actions:
--- a/dev/config.yaml
+++ b/dev/config.yaml
@@ -1,7 +1,7 @@
 security:
   readonly: false
   encryption_key: "=tW$56zytgB&3jN2E%7-+qrGZE?v6LCc"
-  enable_auth: false
+  enable_auth: true
   authorization:
     guest_permissions:
       - actions:
--- a/ui/dashboards/src/components/Panel/Panel.test.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.test.tsx
@@ -73,7 +73,11 @@ jest.mock('@perses-dev/plugin-system', () => {
         PanelComponent: (): JSX.Element => <div>TimeSeriesChart panel</div>,
         actions: [
           {
-            component: (): JSX.Element => <button aria-label="Export CSV">Export</button>,
+            component: (): JSX.Element => (
+              <button aria-label="Export CSV" data-testid="export-action">
+                Export
+              </button>
+            ),
             location: 'header',
           },
         ],
@@ -145,11 +149,9 @@ describe('Panel', () => {
     userEvent.hover(panel);
 
     await waitFor(() => {
-      const exportButton = screen.getAllByLabelText('Export CSV');
-      expect(exportButton.length).toBeGreaterThan(0);
-      exportButton.forEach((button) => {
-        expect(button).toHaveTextContent('Export');
-      });
+      const exportButtons = screen.getAllByTestId('export-action');
+      expect(exportButtons.length).toBeGreaterThan(0);
+      expect(exportButtons[0]).toBeInTheDocument();
     });
   });
 
@@ -272,8 +274,6 @@ describe('Panel', () => {
 
   it('shows loading indicator if 1/2 queries are loading', () => {
     (useDataQueriesContext as jest.Mock).mockReturnValue({
-      isFetching: true,
-      errors: [],
       queryResults: [{ data: { series: [{ name: 'test', values: [[1, 2]] }] }, isFetching: true }],
     });
 
@@ -283,8 +283,6 @@ describe('Panel', () => {
 
   it('does not show a loading indicator if 2/2 queries are loading', () => {
     (useDataQueriesContext as jest.Mock).mockReturnValue({
-      isFetching: true,
-      errors: [],
       queryResults: [],
     });
 
@@ -294,8 +292,6 @@ describe('Panel', () => {
 
   it('shows query errors in the tooltip', () => {
     (useDataQueriesContext as jest.Mock).mockReturnValue({
-      isFetching: false,
-      errors: ['test error'],
       queryResults: [{ error: 'test error' }],
     });
 
--- a/ui/dashboards/src/components/Panel/Panel.test.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.test.tsx
@@ -15,14 +15,66 @@ import { screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { PanelDefinition } from '@perses-dev/core';
 import { DataQueriesProvider, TimeRangeProvider, useDataQueriesContext } from '@perses-dev/plugin-system';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
 import { renderWithContext } from '../../test';
 import { VariableProvider } from '../../context';
 import { Panel, PanelProps } from './Panel';
 
+const testTheme = createTheme({
+  components: {
+    MuiButtonBase: {
+      defaultProps: {
+        disableRipple: true,
+        disableTouchRipple: true,
+      },
+    },
+    MuiButton: {
+      defaultProps: {
+        disableRipple: true,
+        disableTouchRipple: true,
+      },
+    },
+    MuiIconButton: {
+      defaultProps: {
+        disableRipple: true,
+        disableTouchRipple: true,
+      },
+    },
+  },
+  transitions: {
+    create: () => 'none',
+  },
+});
+
+jest.mock('@perses-dev/components', () => ({
+  ...jest.requireActual('@perses-dev/components'),
+  InfoTooltip: ({ children, description }: { children: React.ReactNode; description: string }): JSX.Element => (
+    <>
+      {children}
+      <div role="tooltip">{description}</div>
+    </>
+  ),
+}));
+
+jest.mock('@mui/material/CircularProgress', () => {
+  return function MockCircularProgress(props: { 'aria-label'?: string }): JSX.Element {
+    return <div aria-label={props['aria-label'] || 'loading'} />;
+  };
+});
+
 jest.mock('@perses-dev/plugin-system', () => {
   return {
     ...jest.requireActual('@perses-dev/plugin-system'),
-    useDataQueriesContext: jest.fn(() => ({ queryResults: [] })),
+    useDataQueriesContext: jest.fn(() => ({
+      queryResults: [],
+      isFetching: false,
+      errors: [],
+    })),
+    usePluginRegistry: jest.fn(() => ({
+      getPlugin: jest.fn((): { PanelComponent: () => JSX.Element } => ({
+        PanelComponent: (): JSX.Element => <div>TimeSeriesChart panel</div>,
+      })),
+    })),
   };
 });
 
@@ -51,7 +103,6 @@ describe('Panel', () => {
     },
   });
 
-  // Helper to render the panel with some context set
   const renderPanel = (
     definition?: PanelDefinition,
     editHandlers?: PanelProps['editHandlers'],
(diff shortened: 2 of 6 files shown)


ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -60,23 +116,26 @@ describe('Panel', () => {
...
     renderWithContext(
-      <TimeRangeProvider timeRange={{ pastDuration: '1h' }}>
-        <VariableProvider
-          initialVariableDefinitions={[
-            {
-              kind: 'TextVariable',
-              spec: {
-                name: 'foo',
-                value: 'bar ',
+      // Using your existing renderWithContext helper
```
jgbernalp:
```suggestion
```

ui/dashboards/src/components/Panel/PanelHeader.tsx
```diff
@@ -13,23 +13,31 @@
...
+import { Link, TimeSeriesData, TimeSeries, TimeSeriesMetadata } from '@perses-dev/core';
 import { QueryData, useReplaceVariablesInString } from '@perses-dev/plugin-system';
-import { ReactElement, ReactNode } from 'react';
+import { ReactElement, ReactNode, useMemo } from 'react';
 import { HEADER_ACTIONS_CONTAINER_NAME } from '../../constants';
 import { PanelActions, PanelActionsProps } from './PanelActions';
 
 type OmittedProps = 'children' | 'action' | 'title' | 'disableTypography';
 
+interface AbsoluteTimeRange {
```
jgbernalp: This interface is already present in `@perses-dev/core` we don't need to redefine it here

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -50,7 +51,9 @@ export interface PanelActionsProps {
     isPanelViewed?: boolean;
     onViewPanelClick: () => void;
   };
-  queryResults: QueryData[];
+  queryResults: TimeSeriesData | undefined;
```
jgbernalp: We cannot change the queryResults to be always TimeSeriesData, as the plugins are the ones driving the QueryData.

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -148,7 +218,7 @@ export const Panel = memo(function Panel(props: PanelProps) {
         }}
         ref={setContentElement}
       >
-        <ErrorBoundary FallbackComponent={ErrorAlert} resetKeys={[definition.spec, queryResults]}>
+        <ErrorBoundary FallbackComponent={ErrorAlert} resetKeys={[definition.spec]}>
```
jgbernalp: The error boundary should also reset when there are new queryResults

ui/dashboards/src/components/Panel/Panel.tsx
```diff
@@ -53,6 +58,50 @@ export type PanelExtraProps = {
   panelGroupItemId?: PanelGroupItemId;
 };
 
+// Function to extract project name from URL (kept as fallback)
+const extractProjectNameFromUrl = (): string => {
```
jgbernalp: We cannot rely that the URL structure will be the same as panels might be embedded in other apps. It would be better pass this as a prop and let the parent component provide it.

ui/dashboards/src/components/Panel/PanelHeader.tsx
```diff
@@ -50,6 +61,48 @@ export function PanelHeader({
   const title = useReplaceVariablesInString(rawTitle) as string;
   const description = useReplaceVariablesInString(rawDescription);
 
+  const timeSeriesDataForExport = useMemo(() => {
```
jgbernalp: We don't always have time series as panels can contain different kinds of data. So the export function should be defined in the plugin side and not here.

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -90,18 +176,24 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
   const linksAction = links && links.length > 0 && <PanelLinks links={links} />;
   const extraActions = editHandlers === undefined && extra;
 
-  const queryStateIndicator = useMemo(() => {
-    const hasData = queryResults.some((q) => q.data);
-    const isFetching = queryResults.some((q) => q.isFetching);
-    const queryErrors = queryResults.filter((q) => q.error);
+  const queryStateIndicator = useMemo((): ReactNode | undefined => {
+    const hasData = queryResults && queryResults.series && queryResults.series.length > 0;
```
jgbernalp: not sure why these checks were removed.

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -41,6 +41,7 @@ export interface PanelActionsProps {
   descriptionTooltipId: string;
   links?: Link[];
   extra?: React.ReactNode;
+  panelPluginKind: string;
```
jgbernalp: We probably don't need to pass the panelPluginKind, as is already known in the Panel component. From the Panel component we could fetch the actions defined in the plugin and pass them as `pluginActions`

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -50,7 +51,9 @@ export interface PanelActionsProps {
     isPanelViewed?: boolean;
     onViewPanelClick: () => void;
   };
-  queryResults: QueryData[];
+  queryResults: TimeSeriesData | undefined;
+  projectName?: string;
+  panelProps?: Record<string, unknown>;
```
jgbernalp: We can use the `PanelContentProps` interface here. this contains already the pluginKind and the data so the plugin can use it to build actions.

jgbernalp (changes requested):

ui/core/src/model/panels.ts
```diff
@@ -54,3 +54,31 @@ export interface PanelEditorValues {
   groupId: PanelGroupId;
   panelDefinition: PanelDefinition;
 }
+
+export interface ExportFormat {
```
jgbernalp: I think these interfaces are not used here as the export actions are specific to every plugin

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -87,21 +95,53 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
...
-    const isFetching = queryResults.some((q) => q.isFetching);
+  const queryStateIndicator = useMemo((): ReactNode | undefined => {
+    // Check if any query is fetching
+    const isQueryFetching = queryResults.some((q) => q.isFetching);
+    // Get query-specific errors
     const queryErrors = queryResults.filter((q) => q.error);
-    if (isFetching && hasData) {
+
+    // Convert QueryData[] to TimeSeriesData format for data availability check
+    const timeSeriesData: TimeSeriesData | undefined =
```
jgbernalp: As stated before we don't need a specific format for the data as plugins contain more than time series data. I don't think we need to change this part of the code, as the logic should be present on the plugin

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -68,9 +70,15 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
   description,
   descriptionTooltipId,
   links,
-  queryResults,
+  panelContentProps,
```
jgbernalp: as we are only using the queryResults, we probably don't need to pass it wrapped into the panelContentProps

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -15,14 +15,66 @@ import { screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { PanelDefinition } from '@perses-dev/core';
 import { DataQueriesProvider, TimeRangeProvider, useDataQueriesContext } from '@perses-dev/plugin-system';
-import { renderWithContext } from '../../test';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
+import { renderWithContext } from '../../test'; // Assuming this correctly sets up the test environment
```
jgbernalp:
```suggestion
import { renderWithContext } from '../../test';
```

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -69,8 +70,11 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
   descriptionTooltipId,
   links,
   queryResults,
+  pluginActions = [],
 }) => {
-  const descriptionAction = useMemo(() => {
+  const { errors } = useDataQueriesContext();
```
jgbernalp: I'm not sure why do we need to pull the errors from the queries context here

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -87,21 +89,31 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
...
-      // If the panel has no data, the panel content will show the loading overlay.
-      // Therefore, show the circular loading indicator only in case the panel doesn't display the loading overlay already.
       return <CircularProgress aria-label="loading" size="1.125rem" />;
-    } else if (queryErrors.length > 0) {
-      const errorTexts = queryErrors
-        .map((q) => q.error)
-        .map((e: any) => e?.message ?? e?.toString() ?? 'Unknown error') // eslint-disable-line @typescript-eslint/no-explicit-any
+    }
+
+    const allErrors = queryErrors.map((q) => q.error).filter(Boolean);
```
jgbernalp: Do we need to change these logic, it seems before we were using a `else if`?

ericahinkleRH: the previous code was perfectly fine and I can keep it if you think that is better but the new way I am doing it is just more robust in checking for errors. Do you think I should still revert it? both methods work

jgbernalp: I'd say we don't need to change right now, because the error handling in this case is not related to your changes.

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -15,14 +15,66 @@ import { screen, waitFor } from '@testing-library/react';
...
+  return function MockCircularProgress(props: { 'aria-label'?: string }): JSX.Element {
+    return <div aria-label={props['aria-label'] || 'loading'} />;
+  };
+});
+
 jest.mock('@perses-dev/plugin-system', () => {
   return {
     ...jest.requireActual('@perses-dev/plugin-system'),
-    useDataQueriesContext: jest.fn(() => ({ queryResults: [] })),
+    useDataQueriesContext: jest.fn(() => ({
```
jgbernalp, line 68: we are not using anymore a `useDataQueriesContext`, this seems to be part of the initial solution.

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -15,14 +15,66 @@ import { screen, waitFor } from '@testing-library/react';
...
+      queryResults: [],
+      isFetching: false,
+      errors: [],
+    })),
+    usePluginRegistry: jest.fn(() => ({
+      getPlugin: jest.fn((): { PanelComponent: () => JSX.Element } => ({
+        PanelComponent: (): JSX.Element => <div>TimeSeriesChart panel</div>,
```
jgbernalp, line 73: Since we are mocking the plugin, we can also mock the pluginActions to test the new feature.

jgbernalp (changes requested):

ui/dashboards/src/components/Panel/Panel.test.tsx
jgbernalp: Correction, we need the `useDataQueriesContext` but not the additional parameters

dev/config.yaml
```diff
@@ -1,7 +1,7 @@
 security:
   readonly: false
   encryption_key: "=tW$56zytgB&3jN2E%7-+qrGZE?v6LCc"
-  enable_auth: true
+  enable_auth: false
```
jgbernalp: This might have happened when running the server in e2e mode
```suggestion
  enable_auth: true
```

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -215,6 +294,8 @@ describe('Panel', () => {
 
   it('shows query errors in the tooltip', () => {
     (useDataQueriesContext as jest.Mock).mockReturnValue({
+      isFetching: false,
+      errors: ['test error'],
```
jgbernalp:
```suggestion
```

ui/dashboards/src/components/Panel/Panel.test.tsx
```diff
@@ -60,40 +115,53 @@ describe('Panel', () => {
...
+    const panel = getPanel();
+    userEvent.hover(panel);
+
+    await waitFor(() => {
+      const exportButton = screen.getAllByLabelText('Export CSV');
+      expect(exportButton.length).toBeGreaterThan(0);
+      exportButton.forEach((button) => {
+        expect(button).toHaveTextContent('Export');
+      });
+    });
```
jgbernalp: Maybe this could be simplified to:
```suggestion
    const exportButton = screen.getByLabelText('Export CSV');
    expect(exportButton).toBeInTheDocument();
```

Issue #2225, opened 2024-09-04
Currency value formatting


bripkens:
### Proposal

Hey folks,

we have found that people want to have a formatter for currencies. This is a proposal to add this to the formatting system.

 - Proposed unit: `currency-${currencyCode}`, e.g., `currency-usd` or `currency-jpy`. 
 - The currency code can then be placed neatly into the `Intl.NumberFormat` APIs ([see docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#currency_2))

The implementation is very similar to the one for decimals in the end. I'd be happy to contribute this part.

Nexucis:
ah interesting.

I guess that won't hurt to support it. I'm just wondering if it means we need to list every existing currency ?

bripkens:
Eventually, the answer is probably "yes". Though I would suppose that a certain subset could be sufficient for now. Even a Grafana limits itself to a subset.

<img width="410" alt="image" src="https://github.com/user-attachments/assets/c60dbdf2-04ca-449b-ab1e-6970f199dad4">

Aurifi:
Hey, @Nexucis , just gave this Issue a shot. :) Please let me know if I'm missing something in the process or any adaptions are still necessary.

PR #3132, opened 2025-07-27
[FEATURE] Add currency value formatting (perses#2225)

aurifi:
# Description

Solves #2225 

The changes in this PR add "Currency" as additional Unit for formatting. Initially the [15 most traded currencies](https://en.wikipedia.org/wiki/Foreign_exchange_market#Most_traded_currencies_by_value) were added. The currency units follow the ISO 4217 currency codes, thus addition or removal of units should be trivial. 

# Screenshots

image

image

# Checklist

## UI Changes

### DIFF
--- a/go-sdk/common/format.go
+++ b/go-sdk/common/format.go
@@ -18,9 +18,12 @@ import (
 	"fmt"
 )
 
-type TimeUnit string
-type PercentageUnit string
-type ThroughputUnit string
+type (
+	TimeUnit       string
+	PercentageUnit string
+	ThroughputUnit string
+	CurrencyUnit   string
+)
 
 const (
 	MilliSecondsUnit       TimeUnit       = "milliseconds"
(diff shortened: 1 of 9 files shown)


Gladorme:
Hello, thanks for the contribution 👍 

Format need to be added in `cue/common/format.cue` too, in order to have the validation working

Aurifi:
Thanks @Gladorme ! Good point, I'll add those to cue aswell as the go-validation! :)

Edit: After looking into it there's even more to consider. E.g. migration of Grafana-Dashboards.

Nexucis:
@aurifi I think it will be fine if having the logic to migrate is done in another PR. Supporting the currency will be already something.

If you are willing to continue this PR, it will be great to edit the file https://github.com/perses/perses/blob/main/go-sdk/common/format.go as well. Like that, our Dashboard as Code SDKs will be up to date :).

ui/core/src/model/units/units.ts
```diff
@@ -47,21 +48,24 @@ export const UNIT_GROUP_CONFIG: Readonly<Record<UnitGroup, UnitGroupConfig>> = {
...
 } as const;
 
 export type FormatOptions =
   | TimeFormatOptions
   | PercentFormatOptions
   | DecimalFormatOptions
   | BytesFormatOptions
-  | ThroughputFormatOptions;
+  | ThroughputFormatOptions
+  | CurrencyFormatOptions;
```
AntoineThebaud, line 68: this should be replicated on the CUE side as well https://github.com/perses/perses/blob/main/cue/common/format.cue#L16

Aurifi:
Hey @AntoineThebaud & @Nexucis ,
i've added the types to format.go & format.cue. However while looking into the migration of dashboards I've noticed that dashboards which contain Currencies are not able to be saved, as the type is not yet present in the plugins for e.g. TimeseriesCharts etc, so the version in the individual plugins will need to be bumped also.

go-sdk/common/format.go
```diff
@@ -47,12 +50,24 @@ const (
...
+	SwissFrancUnit         CurrencyUnit   = "chf"
+	RenminbiUnit           CurrencyUnit   = "cny"
+	EuroUnit               CurrencyUnit   = "eur"
+	PoundUnit              CurrencyUnit   = "gbp"
+	HongKongDollarUnit     CurrencyUnit   = "hkd"
+	IndianRupeeUniit       CurrencyUnit   = "inr"
+	YenUnit                CurrencyUnit   = "jpy"
+	SouthKoreanWonUnit     CurrencyUnit   = "krw"
+	NorwegianKroneUnit     CurrencyUnit   = "nok"
+	NewZealandDollarUnit   CurrencyUnit   = "nzd"
```
Nexucis, line 64: it seems it is missing the currency `"usd" | "sek" | "sgd"`. Can you add them please ?

Nexucis: and since you started to sort them in alphabetic order (at least for the first twelve), can you do the same for the `"usd" | "sek" | "sgd"`

Nexucis:
> Hey @AntoineThebaud & @Nexucis , i've added the types to format.go & format.cue. However while looking into the migration of dashboards I've noticed that dashboards which contain Currencies are not able to be saved, as the type is not yet present in the plugins for e.g. TimeseriesCharts etc, so the version in the individual plugins will need to be bumped also.

yeah once this is merged and released, we should update all plugins to support the new format and then released all of them (once again).

/cc @andreasgerstmayr @jgbernalp

go-sdk/common/format.go
Aurifi: Added and sorted those :)

Nexucis (approved):
awesome, thank you @aurifi !

PR #3138, opened 2025-07-30
[DOC] Add concept doc for Variable + few patches

AntoineThebaud:
# Description

New simple concept doc for the variables (see rendered [here](https://github.com/perses/perses/blob/antoinethebaud/doc-var/docs/concepts/variable.md)) + few renaming here and there to remove plural.

# Checklist


docs/concepts/variable.md
```diff
@@ -0,0 +1,23 @@
+# Variable
+
+A **variable** in Perses is a dynamic configuration element that allows you to create interactive and reusable dashboards. Variables act as parameters that can be modified by users to filter data, change query scopes, or customize dashboard views without editing the underlying dashboard configuration.
+
+![variables](https://github.com/user-attachments/assets/62d5d94b-7ceb-4bfc-ad38-a577a724b419)
```
jgbernalp, line 5: This file was not found from my side, not sure if we should add it as part of the repo or the link was meant for something different

AntoineThebaud: Oh strange I do see it on my side.. for such images until now we used the github storage that way, so I just did the same as usual.. I saw that there can be restrictions on image visibility (like it should be referenced by code from the main branch in order to work) so maybe merging the PR will solve it? Anyway we can try and fix afterwards if needed, the doc is not yet "released"

PR #3140, opened 2025-07-30
[ENHANCEMENT] create RouterProvider to remove hard dependency on react-router in plugins

andreasgerstmayr:
# Description

Some panel plugins (TraceTable, ScatterPlot, TracingGanttChart) support linking to other pages, e.g. clicking on a trace in the TraceTable should navigate to the TracingGanttChart.

We can't use react-router in the panel, because panels might be embedded into React applications which use a different routing library, or a different major version of react-router.

This new provider abstracts the basic routing functionality, to remove the dependency on the exact version of react-router.

# Screenshots

no UI changes

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/views/projects/dashboards/HelperDashboardView.tsx
+++ b/ui/app/src/views/projects/dashboards/HelperDashboardView.tsx
@@ -18,6 +18,7 @@ import { ExternalVariableDefinition, OnSaveDashboard, ViewDashboard } from '@per
 import {
   PluginRegistry,
   UsageMetricsProvider,
+  ReactRouterProvider,
   ValidationProvider,
   remotePluginLoader,
 } from '@perses-dev/plugin-system';
@@ -82,39 +83,41 @@ export function HelperDashboardView(props: GenericDashboardViewProps): ReactElem
       }}
     >
       <ErrorBoundary FallbackComponent={ErrorAlert}>
-        <PluginRegistry
-          pluginLoader={remotePluginLoader()}
-          defaultPluginKinds={{
-            Panel: 'TimeSeriesChart',
-            TimeSeriesQuery: 'PrometheusTimeSeriesQuery',
-          }}
-        >
-          <ValidationProvider>
-            <ErrorBoundary FallbackComponent={ErrorAlert}>
-              <UsageMetricsProvider project={project.metadata.name} dashboard={dashboardResource.metadata.name}>
-                <ViewDashboard
-                  dashboardResource={dashboardResource}
-                  datasourceApi={datasourceApi}
-                  externalVariableDefinitions={externalVariableDefinitions}
-                  dashboardTitleComponent={
-                    <ProjectBreadcrumbs dashboardName={getResourceDisplayName(dashboardResource)} project={project} />
-                  }
-                  emptyDashboardProps={{
-                    additionalText: 'In order to save this dashboard, you need to add at least one panel!',
-                  }}
-                  onSave={onSave}
-                  onDiscard={onDiscard}
-                  initialVariableIsSticky={true}
-                  isReadonly={isReadonly}
-                  isVariableEnabled={isLocalVariableEnabled}
-                  isDatasourceEnabled={isLocalDatasourceEnabled}
-                  isEditing={isEditing}
-                  isCreating={isCreating}
-                />
-              </UsageMetricsProvider>
-            </ErrorBoundary>
-          </ValidationProvider>
-        </PluginRegistry>
+        <ReactRouterProvider>
+          <PluginRegistry
+            pluginLoader={remotePluginLoader()}
+            defaultPluginKinds={{
+              Panel: 'TimeSeriesChart',
+              TimeSeriesQuery: 'PrometheusTimeSeriesQuery',
+            }}
+          >
+            <ValidationProvider>
+              <ErrorBoundary FallbackComponent={ErrorAlert}>
+                <UsageMetricsProvider project={project.metadata.name} dashboard={dashboardResource.metadata.name}>
+                  <ViewDashboard
+                    dashboardResource={dashboardResource}
+                    datasourceApi={datasourceApi}
+                    externalVariableDefinitions={externalVariableDefinitions}
+                    dashboardTitleComponent={
+                      <ProjectBreadcrumbs dashboardName={getResourceDisplayName(dashboardResource)} project={project} />
+                    }
+                    emptyDashboardProps={{
+                      additionalText: 'In order to save this dashboard, you need to add at least one panel!',
+                    }}
+                    onSave={onSave}
+                    onDiscard={onDiscard}
+                    initialVariableIsSticky={true}
+                    isReadonly={isReadonly}
+                    isVariableEnabled={isLocalVariableEnabled}
+                    isDatasourceEnabled={isLocalDatasourceEnabled}
+                    isEditing={isEditing}
+                    isCreating={isCreating}
+                  />
+                </UsageMetricsProvider>
+              </ErrorBoundary>
+            </ValidationProvider>
+          </PluginRegistry>
+        </ReactRouterProvider>
       </ErrorBoundary>
     </Box>
   );
--- a/ui/plugin-system/src/runtime/RouterProvider.tsx
+++ b/ui/plugin-system/src/runtime/RouterProvider.tsx
@@ -1,0 +1,65 @@
+// Copyright 2025 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React, { createContext, useContext, ReactNode, ReactElement } from 'react';
+import { Link as RouterLink, useNavigate } from 'react-router-dom';
+
+interface LinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+  to: string;
+}
+
+export interface RouterContextType {
+  RouterComponent: (props: LinkProps & React.RefAttributes<HTMLAnchorElement>) => ReactNode;
+  navigate: (to: string) => void;
+}
+
+export const RouterContext = createContext<RouterContextType | undefined>(undefined);
+
+export function useRouterContext(): RouterContextType {
+  const ctx = useContext(RouterContext);
+  if (ctx === undefined) {
+    throw new Error('No RouterContext found. Did you forget a <RouterProvider>?');
+  }
+  return ctx;
+}
+
+interface RouterProviderProps {
+  RouterComponent: RouterContextType['RouterComponent'];
+  navigate: RouterContextType['navigate'];
+  children?: React.ReactNode;
+}
+
+/**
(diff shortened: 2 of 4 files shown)


ui/app/src/views/projects/dashboards/HelperDashboardView.tsx
```diff
@@ -82,39 +83,41 @@ export function HelperDashboardView(props: GenericDashboardViewProps): ReactElem
...
-                  isVariableEnabled={isLocalVariableEnabled}
-                  isDatasourceEnabled={isLocalDatasourceEnabled}
-                  isEditing={isEditing}
-                  isCreating={isCreating}
-                />
-              </UsageMetricsProvider>
-            </ErrorBoundary>
-          </ValidationProvider>
-        </PluginRegistry>
+        <ReactRouterProvider>
```
andreasgerstmayr: Not sure at what level in the hierarchy I should add it, I added it close to the `PluginRegistry` for now because it's related to the panel plugins. Alternative I could also add it to `app/src/render-app.tsx` or somewhere else.

ui/plugin-system/src/runtime/RouterProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Link as RouterLink, useNavigate } from 'react-router-dom';
+
+interface LinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+  to: string;
+}
+
+export interface RouterContextType {
+  RouterComponent: (props: LinkProps & React.RefAttributes<HTMLAnchorElement>) => ReactNode;
+  navigate: (to: string) => void;
+}
```
andreasgerstmayr, line 24: Should the router provider also support adding routes?

**Use Case:** When clicking on a trace in the TraceTable or a bubble in the ScatterChart, I want to navigate to the Gantt chart.

Unfortunately the final route is only known when the Trace Table is shown in e.g. the Tempo Explorer, because technically the Trace Table can also be used with results from a Jaeger data source (doesn't exist yet, but maybe in the future), so I cannot hardcode a link to the Tempo Explorer in the Trace Table panel.

In https://github.com/perses/plugins/pull/211 I support configuring this link in the Trace Table panel spec, but another way would be to add it to this router, and embed this provider in the Tempo Explorer. What do you think?

OTOH, this doesn't help much when embedding the Trace Table in a dashboard. In this case, clicking on a link in the trace table should jump to the right explorer view, with the right trace query 🤔

jgbernalp: I think the routes are responsibility of the host application, and the dashboards panels loaded in the host application should be configured to link to whatever route the dashboard creator intends within the host application. This means it wont be also hardcoded for Tempo but will be a dashboard configuration.

ui/app/src/views/projects/dashboards/HelperDashboardView.tsx
jgbernalp: probably it will be better to add it on the application, as the `HelperDashboardView` could be used directly in a Host application, if we leave it here then we are locking the Provider for users who want to embed a whole dashboard rather than specific panels.

ui/plugin-system/src/runtime/RouterProvider.tsx
andreasgerstmayr: > I think the routes are responsibility of the host application, and the dashboards panels loaded in the host application should be configured to link to whatever route the dashboard creator intends within the host application. This means it wont be also hardcoded for Tempo but will be a dashboard configuration.

Ok, so in other words, that means the links should be configured in the panel spec? i.e. what https://github.com/perses/plugins/pull/211 does?

ui/app/src/views/projects/dashboards/HelperDashboardView.tsx
andreasgerstmayr: great point! I'll move it.

ui/plugin-system/src/runtime/RouterProvider.tsx
jgbernalp: yes exactly, I meant to say that the links should not be hardcoded also in plugins or dashboards but configurable.

jgbernalp (approved):

Nexucis:
@shahrokni could you please review this PR as well ? I am curious about your thought, specially since you are taking care about re-organising the embedded API.

ui/plugin-system/src/runtime/RouterProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const RouterContext = createContext<RouterContextType | undefined>(undefined);
+
+export function useRouterContext(): RouterContextType {
+  const ctx = useContext(RouterContext);
+  if (ctx === undefined) {
+    throw new Error('No RouterContext found. Did you forget a <RouterProvider>?');
+  }
```
shahrokni, line 32: A more common JS syntax would be 

```javaScript
if (!ctx) {
  throw new Error('No RouterContext found. Did you forget a <RouterProvider>?');
}
```

ui/app/src/render-app.tsx
```diff
@@ -56,13 +57,15 @@ export function renderApp(container: Element | null): void {
...
-                  </NavHistoryProvider>
+                  <ReactRouterProvider>
+                    <NavHistoryProvider>
+                      <SnackbarProvider anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}>
+                        <AuthorizationProvider>
+                          <App />
+                        </AuthorizationProvider>
+                      </SnackbarProvider>
+                    </NavHistoryProvider>
+                  </ReactRouterProvider>
```
shahrokni, line 68: Could you please explain the role of this provider?

**Let me explain myself,**

I am wondering if the `ReactRouterProvider` would affect the Perses app. 
After removing this provider (ReactRouterProvider ), the router was still working, and I could simply navigate through the application. It seems the `ReactRouterProvider` is only a wrapper on top of the `RouterProvider` which borrows the navigate functionality from `react-router-dom`. OK, makes sense. But then, let's take a deeper look. The mentioned navigate functionality can only be accessed through `useRouterContext`. Here, I am a bit confused, because there is no usage of this hook in the Perses app, or even Perses/Perses repo. So, why would we wrap the app with `ReactRouterProvider`?

 **_Am I missing something?_**

ui/plugin-system/src/runtime/RouterProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Link as RouterLink, useNavigate } from 'react-router-dom';
+
+interface LinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+  to: string;
+}
+
+export interface RouterContextType {
+  RouterComponent: (props: LinkProps & React.RefAttributes<HTMLAnchorElement>) => ReactNode;
+  navigate: (to: string) => void;
+}
```
shahrokni, line 24: I like the abstraction idea,
But I am wondering if this abstraction covers all the functionalities. For example, the navigate functionality has been provided. Good.  But, what about the rest of the functionalities? I guess (if I am not mistaken) you would add functionalities as we go along. But what if, some libraries offer more than the others, and do not  fit this abstraction? Then would we have some `undefined` functionalities? 

**_Am I missing something?_**

```typescript
/**
 * Returns an imperative method for changing the location. Used by `<Link>`s, but
 * may also be used by other elements to change the location.
 *
 * @see https://reactrouter.com/v6/hooks/use-navigate
 */
export declare function useNavigate(): NavigateFunction;
/**
 * Returns the context (if provided) for the child route at this level of the route
 * hierarchy.
 * @see https://reactrouter.com/v6/hooks/use-outlet-context
 */
export declare function useOutletContext<Context = unknown>(): Context;
/**
 * Returns the element for the child route at this level of the route
 * hierarchy. Used internally by `<Outlet>` to render child routes.
 *
 * @see https://reactrouter.com/v6/hooks/use-outlet
 */
export declare function useOutlet(context?: unknown): React.ReactElement | null;
/**
 * Returns an object of key/value pairs of the dynamic params from the current
 * URL that were matched by the route path.
 *
 * @see https://reactrouter.com/v6/hooks/use-params
 */
export declare function useParams<ParamsOrKey extends string | Record<string, string | undefined> = string>(): Readonly<[
    ParamsOrKey
] extends [string] ? Params<ParamsOrKey> : Partial<ParamsOrKey>>;
/**
 * Resolves the pathname of the given `to` value against the current location.
 *
 * @see https://reactrouter.com/v6/hooks/use-resolved-path
 */
export declare function useResolvedPath(to: To, { relative }?: {
    relative?: RelativeRoutingType;
}): Path;
/**
 * Returns the element of the route that matched the current location, prepared
 * with the correct context to render the remainder of the route tree. Route
 * elements in the tree must render an `<Outlet>` to render their child route's
 * element.
 *
 * @see https://reactrouter.com/v6/hooks/use-routes
 */
export declare function useRoutes(routes: RouteObject[], locationArg?: Partial<Location> | string): React.ReactElement | null;
export declare function useRoutesImpl(routes: RouteObject[], locationArg?: Partial<Location> | string, dataRouterState?: RemixRouter["state"], future?: RemixRouter["future"]): React.ReactElement | null;

/* ... etc*/
```

ui/app/src/render-app.tsx
andreasgerstmayr: You're right, this provider is not used in the core app at all, it's only required for panel plugins.

I added this comment on top of the provider which should clarify its usage:
```
/**
 * Some panel plugins (TraceTable, ScatterPlot, TracingGanttChart) support linking to other pages,
 * e.g. clicking on a trace in the TraceTable should navigate to the TracingGanttChart.
 *
 * We can't use react-router in the panel, because panels might be embedded into React applications
 * which use a different routing library, or a different major version of react-router.
 *
 * This provider abstracts the basic routing functionality, to remove the dependency on the exact version of react-router.
 */
```

ui/plugin-system/src/runtime/RouterProvider.tsx
andreasgerstmayr: I kept the interface as minimal as possible, to make it easier to switch out the router implementation (not every React app uses the same routing library and library version as Perses does).

Let's say you're an observability SaaS and would like to embed Perses panels into your own application, which uses `react-router-dom` version 5, or some other routing library altogether (Perses uses `react-router-dom` version 6). That's where this provider comes handy.

The panel can have this code (example from https://github.com/perses/plugins/pull/211/files#diff-e7d0d0cd983f1a7c6f5f7718b09c3d5499c27ae7cab36f624c74a9b77dc0f26bR197)
```
const { RouterComponent } = useRouterContext();
return (
  <Link variant="body1" color="inherit" underline="hover" component={RouterComponent} to={trace.traceLink}>
);
```

This decouples the routing library from the panels.

Without this provider, every app which embeds Perses panels containing links must use the same `react-router-dom` version as Perses itself.

ui/app/src/render-app.tsx
shahrokni: So, do we need to use  ` <ReactRouterProvider>` in `render-app`? (if its only used and required panel plugins)
Because, it seems, it would not affect the Perses app routing process.

shahrokni: @andreasgerstmayr ?

andreasgerstmayr: It is a React context, the provider must be declared higher up in the hierarchy than the component which reads the context (the panel plugin). If I declare the context in the same React component which uses it, it'll fail because the context is undefined.

Initially I had it declared in the `<HelperDashboardView />` component, but @jgbernalp suggested to move it to the application, because if the host application (e.g. OpenShift Console, or some observability SaaS app) includes `<HelperDashboardView />` directly than the app cannot overwrite the `RouterProvider` anymore (see the first resolved thread in this PR).

shahrokni (approved):

PR #3141, opened 2025-07-31
[BUGFIX] Avoid dashboard name duplication from the front-end side

### LINKED ISSUE #1526: Prevent providing an already-used dashboard name when creating a new one
### Proposal

When you create a new dashboard, there are no checks to verify if the provided name (more precisely, the name computed from the provided display name) is already taken by another dashboard in the same project. Thus you're able to land on the dashboard page & add things to it, and once you'll try to save it you will get a conflict error. There's a workaround here, since you can go to dashboard JSON and change the metadata.name to solve the conflict _(changing metadata this way is possible when you create a new dashboard)_.

To improve this we could have a check when providing the dashboard name in the creation modal, to avoid ending up in such situation.

shahrokni:
Closes #1526 

# Description
Details of the bug can be found here #1526 
There was a small bug with the `useDashboardValidationSchema` function. 

1. The bug was resolved
2. The function was improved a bit to handle corner cases including error and loading states

image

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/dialogs/CreateDashboardDialog.tsx
+++ b/ui/app/src/components/dialogs/CreateDashboardDialog.tsx
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Dispatch, DispatchWithoutAction, ReactElement, useState } from 'react';
-import { Button, FormControlLabel, MenuItem, Stack, Switch, TextField } from '@mui/material';
+import { Dispatch, DispatchWithoutAction, ReactElement, useCallback, useState } from 'react';
+import { Button, CircularProgress, FormControlLabel, MenuItem, Stack, Switch, TextField } from '@mui/material';
 import { Dialog } from '@perses-dev/components';
 import { Controller, FormProvider, SubmitHandler, useForm } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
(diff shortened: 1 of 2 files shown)


shahrokni:
@AntoineThebaud  and @Gladorme

jgbernalp (approved):

ui/app/src/components/dialogs/CreateDashboardDialog.tsx
```diff
@@ -92,32 +92,52 @@ interface DuplicationFormProps {
...
     defaultValues: { dashboardName: '', projectName: projects[0]?.metadata.name ?? '' },
   });
 
-  const processDashboardForm: SubmitHandler<CreateDashboardValidationType> = (data) => {
-    onClose();
-    if (onSuccess) {
-      onSuccess({ project: data.projectName, dashboard: data.dashboardName } as DashboardSelector);
-    }
-  };
+  const getProcessDashboardForm = useCallback((): SubmitHandler<CreateDashboardValidationType> => {
```
AntoineThebaud: why rename the function here? After the changes I think `processDashboardForm` still makes more sense than `getProcessDashboardForm`

shahrokni: So, the function itself has slightly changed. Now the function is returing another function. (The old one)

The new one is kind of JS clouser, and for that, since it is returning another function, get+ the old function name may make more sense. But if you prefer we can change it to the old one. No worries. Just let me know. 
@AntoineThebaud

AntoineThebaud: Ok my bad, I read to fast here... in such case better to name it `handleProcessDashboardForm` I guess, this `handle` terminology is what we usually use in the code when defining such callbacks.

shahrokni: Understood. I ll change it and let you know.

Issue #3084, opened 2025-07-11
[datasourceVariable] Saving a dashboard with datasourceVariable does not work


ibakshay:
### What did you do?

I created a datasource variable and tried to use it in a panel on a dashboard. It works during edit mode, but I'm not able to save the dashboard. I'm getting the following error when I save the dashboard.

```
Request URL
https://demo.perses.dev/api/v1/projects/akshay/dashboards/nodeexporterfull
Request Method
PUT
Status Code
400 Bad Request

Response: 
{
    "message": "bad request: invalid query n°1: spec.datasource: conflicting values \"$testPrometheus\" and {kind:string,name?:string} (mismatched types string and struct):\n    1:58\n    ../etc/perses/plugins/Prometheus-0.51.0-rc.3/cue.mod/pkg/github.com/perses/perses/cue/common/datasource.cue:17:15\n"
}
```

### Screenshot
<img width="1509" height="746" alt="Image" src="https://github.com/user-attachments/assets/c981d218-c5c1-42f2-98db-425dad3042d9" />

### What did you expect to see?

The dashboard is being saved without any error. 

### What did you see instead? Under which circumstances?

```
Request URL
https://demo.perses.dev/api/v1/projects/akshay/dashboards/nodeexporterfull
Request Method
PUT
Status Code
400 Bad Request

Response: 
{
    "message": "bad request: invalid query n°1: spec.datasource: conflicting values \"$testPrometheus\" and {kind:string,name?:string} (mismatched types string and struct):\n    1:58\n    ../etc/perses/plugins/Prometheus-0.51.0-rc.3/cue.mod/pkg/github.com/perses/perses/cue/common/datasource.cue:17:15\n"
}
```

### Perses configuration file

```yaml

```

### Logs

```text

```

ibakshay:
@OrReuben , Could you please check and tell me if this is a valid issue?

OrReuben:
Hmm I can't confirm this issue happening to me as we are using a fixed version of perses and did not update lately.
Can you show a video of the exact process you did that made it happen?

ibakshay:
Sure, The perses version is `v0.51.1`.  @OrReuben 

I have added the screen recording to reproduce this issue below. 

## Screen Recording
https://github.com/user-attachments/assets/88d5501a-5c63-4d57-ae03-c8a07c89b335

OrReuben:
Seems to me like its a validation error on the backend side - probably because of recent change.
[...]

PR #3151, opened 2025-08-05
[ignore] bump prometheus, tempo, statchart and piechart version

Nexucis:


scripts/plugin/plugin.yaml
```diff
@@ -11,13 +11,13 @@
 - name: "Markdown"
   version: "0.8.0"
 - name: "PieChart"
-  version: "0.8.0"
+  version: "0.8.1"
 - name: "Prometheus"
-  version: "0.51.0-rc.3"
+  version: "0.52.0"
```
ibakshay: FYI: This new Prometheus version also will have this bug - https://github.com/perses/perses/issues/3084

PR #3160, opened 2025-08-07
[BUGFIX] Explorer: remove the dependency to the app

Nexucis:
it makes the explorer package unusable because it does not compile anymore as the package @perses-dev/app is not published

### DIFF
--- a/ui/explore/src/components/ExploreManager/ExploreManager.tsx
+++ b/ui/explore/src/components/ExploreManager/ExploreManager.tsx
@@ -13,10 +13,9 @@
 
 import { Box, Button, Card, Stack, Tab, Tabs, useMediaQuery } from '@mui/material';
 import { PluginLoaderComponent, useListPluginMetadata } from '@perses-dev/plugin-system';
-import { ReactElement, ReactNode, useEffect, useMemo } from 'react';
+import { ReactElement, ReactNode, useEffect, useMemo, useState } from 'react';
 import ChevronRight from 'mdi-material-ui/ChevronRight';
 import ChevronLeft from 'mdi-material-ui/ChevronLeft';
-import { useLocalStorage } from '@perses-dev/app/src/utils/browser-storage';
 import { ExploreToolbar } from '../ExploreToolbar';
 import { useExplorerManagerContext } from './ExplorerManagerProvider';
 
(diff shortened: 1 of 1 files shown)


shahrokni:
⚠️  **_I am testing this, Please be patient ..._**

ui/explore/src/components/ExploreManager/ExploreManager.tsx
```diff
@@ -31,7 +30,7 @@ export function ExploreManager(props: ExploreManagerProps): ReactElement {
   const plugins = useListPluginMetadata(['Explore']);
 
   const isSmallScreen = useMediaQuery('(max-width: 768px)');
-  const [isCollapsed, setIsCollapsed] = useLocalStorage<boolean>('explore-tabs-collapsed', false);
+  const [isCollapsed, setIsCollapsed] = useState<boolean>(false);
```
shahrokni, line 33: @Nexucis 

This change may cause a bug. I will explain my concern, and then let me know you are still confident with the change. 
If you are aware of the mentioned issues, and confident with them, let me know to resolve this comment. 

## Why this may have a consequence

There is a difference between React useState and local storage. 


While useState is effective for managing state in memory during the component's lifecycle, useLocalStorage takes state management a step further by allowing data persistence in the browser's local storage, so if you refresh and go back to the site, the state will persist.

https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

So, I see two bug-prone areas here.

1. If someone other part of the code is counting on the explore-tabs-collapse
2. This behavior would be in memory and not persisted. As soon as you close the browser, it is gone

## More context if you are interested
As you can see this is a wrapper on top of the browser local-storage. 

```typescript
export function useLocalStorage<T>(key: string, initialValue: T): StorageTuple<T> {
  const { value, setValueAndStore } = useStorage(global.localStorage, key, initialValue);
  return [value, setValueAndStore];
}

function useStorage<T>(
  storage: Storage,
  key: string,
  initialValue: T
): {
  setValueAndStore: (value: T) => void;
  setValue: (value: T) => void;
  value: T;
} {
  // Use state so that changes cause the page to re-render
  const [value, setValue] = useState<T>(() => {
    try {
      const json = storage.getItem(key);
      if (json !== null) {
        return JSON.parse(json);
      }
    } catch {
      // No-op
    }

    // Either the value isn't in storage yet or JSON parsing failed, so
    // set to the initial value in both places
    storage.setItem(key, JSON.stringify(initialValue));
    return initialValue;
  });

  // Set in both places
  const setValueAndStore = useCallback(
    (val: T) => {
      setValue(val);
      storage.setItem(key, JSON.stringify(val));
    },
    [setValue, storage, key]
  );

  return { value, setValue, setValueAndStore };
}

```

Nexucis: yes I am aware of that and I believe since the code is relatively new, there is no code that is based on the state of the `explore-tabs-collapse`. And if it is the case, then whether the state is persistent or not, should not be a factor for the code depending of it.

PR #3162, opened 2025-08-07
[BUGFIX] Fix Piechart legend value

shahrokni:
# Description

# Screenshots

# Checklist

## UI Changes


cue/common/calculation.cue
```diff
@@ -13,4 +13,4 @@
 
 package common
 
-#calculation: "first" | *"last" | "first-number" | "last-number" | "mean" | "sum" | "min" | "max"
+#calculation: "first" | *"last" | "first-number" | "last-number" | "mean" | "sum" | "min" | "max" | "abs" | "relative"
```
AntoineThebaud: I dont think we should mix the two here, they're wont be a case where all these values would be applicable I guess.
Thus we should go with sth like:

```suggestion
#calculationAggregation: "first" | *"last" | "first-number" | "last-number" | "mean" | "sum" | "min" | "max"

#calculationComparison: "abs" | "relative"
```

NB: if ever in the future we need to mix both for a new plugin, there'd always be the option of adding
`#calculation: #calculationAggregation | #calculationComparison`

Nexucis:
@sjcobb if you have time, perhaps you will be interested by this PR :)

ui/plugin-system/src/components/LegendOptionsEditor/LegendOptionsEditor.tsx
```diff
@@ -54,23 +54,19 @@ const SIZE_OPTIONS: LegendSizeOption[] = Object.entries(LEGEND_SIZE_CONFIG).map(
...
   onChange: (legend?: LegendSpecOptions) => void;
   showValuesEditor?: boolean;
+  legendValueOptionsFilter?: { filterType: 'include' | 'exclude'; legendValueOptions: LegendValue[] };
 }
 
 export function LegendOptionsEditor({
   value,
   onChange,
   showValuesEditor = true,
+  legendValueOptionsFilter,
```
AntoineThebaud: Instead of a filter I think (a bit linked to my above suggestion for CUE) that the new prop should be sth like
```ts
calculation?: "aggregation" | "comparison"
```
and based on the passed string the control would either use [max, min, avg...] or [abs, relative]

Your approach gives more flexibility but again I think aggr & comparison operators are really different things that are not going to be used together, so having a strict split between the 2 makes more sense imho

WDYT @Nexucis?

ui/core/src/model/calculations.ts
```diff
@@ -24,6 +38,8 @@ export const CalculationsMap = {
   sum: sum,
   min: min,
   max: max,
+  abs: NOP,
```
jgbernalp: Not sure why `abs` would be a no op in time series. It could mean all the values in the timeseries should be positive (reflected on the y axis) Maybe a different name would make more sense?.

Even if this calculations are used by a single plugin through filtering. I wonder if the plugins should then be able to provide their own calculations based on how they want to present the data. IIUC this calculations are for time series and not specific to a plugin.

shahrokni: @jgbernalp 
The nature of abs and relative is different from the other functionalities. If you take a look at the existing calculations you see that their function signature would not fit the abs and relative.
Fortunately the abs and relative have been already calculated on the pie-chart plugin, and since the information is already available, we do not need to do extra calculations. Those NOP functions are only playing a marker role. Something like a marker interface in OOP

@jgbernalp 
**The name somehow should convey this message that the functionality is provided by the plugin itself**, and is not common calculation, would you agree? Maybe I can simply set it to undefined? (Instead of NOP)



```typeScript
/* Important: Absolute and relative operations are not time-series operations 
   Why? 
   Fact #1: At the moment absolute and relative calculations are only used by pie-chart
   Fact #2: Time-series data (values), which is a sequence of [timestamp, value] pairs.
   Pie charts are typically used to represent proportions or aggregated values at a single point in time,
   not over a time range.
   So, for now we only use abs and relative notions as markers for pie-chart calculations.
*/
```

ui/plugin-system/src/components/LegendOptionsEditor/LegendOptionsEditor.tsx
shahrokni: I like the idea, but why did you pick the name comparison? Is comparison a proper name? 
Let's say I have a pie-chart called fruit. 
In my basket there are 5 apples and 5 oranges. So, in the aggregate have 10 items, 50% apples, and 50% oranges
So, here for instance,  abs for the apples would be 5, and the relative would be 50%. 
Where is the comparison part here?

AntoineThebaud: Feel free to propose another name if you have, I didnt find a better one on my side 😅 PieChart (and by extension both the relatvie & abs displays) vizualization is sth that allows for easy comparison of data (e.g which subset takes the biggest part of the Pie) so imho "comparison" works here, but I'm opened to suggestions

shahrokni: Then, comparison it is 😈 👍🏼

ui/core/src/model/calculations.ts
AntoineThebaud: My [suggestion](https://github.com/perses/perses/pull/3162#discussion_r2262422100) may remove the need of this trick? You could have different algorithms for the different calculation "modes"

shahrokni: > My [suggestion](https://github.com/perses/perses/pull/3162#discussion_r2262422100) may remove the need of this trick? You could have different algorithms for the different calculation "modes"

True, your idea may help us get rid of the trick. 👍🏼

jgbernalp: Agree, if the nature is different we should not mix it. But we should not add logic specific to a plugin into the time series, that would break the whole idea of keeping this as a generic interface. We probably need to adjust the cue model so we have predefined calculations and plugins can add their own. So we don't have to add placeholders.

cue/common/calculation.cue
Nexucis: if we change the field name `calculation` to another thing, that will be an annoying breaking change on every plugin

AntoineThebaud: Well, when we would upgrade the perses/perses dep of these (5) plugins in their respective cue.mod, we'd also have to do this rename.. no big deal IMHO.
But if you prefer we could do sth like this to avoid a breaking change:

```cue
#calculation: #calculationAggregation // kind-of default value defined for retro-compatibility. Please rather use below types directly for new plugins.

#calculationAggregation: "first" | *"last" | "first-number" | "last-number" | "mean" | "sum" | "min" | "max"

#calculationComparison: "abs" | "relative"
```

Nexucis: yes I would prefer we do that, and hopefully we can put a `deprecated` comment on this particular field

ui/core/src/model/calculations.ts
shahrokni: @jgbernalp 
I will wait for your approval as well. Take a look whenever you have time

jgbernalp (approved):

Issue #2999, opened 2025-06-10
Add Loki datasource plugin and related visualizations


jgbernalp:
### Proposal

Create a datasource for Loki that allows users to configure:

- Tenant configuration, by using the `X-Scope-OrgID` header or a configurable API path
- Add Query plugins for
  - Log Queries
  - Metric Queries
- Reuse the histogram chart to display log volume

jgbernalp:
Additional requirements important for users:

- Saving Complex Queries and reusing
- Displaying total number of query results
- Listing Distinct values (ie repeated errors)
- Plotting recurrences of specific errors

PR #3163, opened 2025-08-07
[FEATURE] Add LogQuery support and related data structures and queries

abelyakin:
# Description
Adds necessary models and queries for LogQuery type

We need it for [#perses/plugins/178](https://github.com/perses/plugins/pull/178)
relates to #2999 

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/ui/core/src/model/log-data.ts
+++ b/ui/core/src/model/log-data.ts
@@ -30,11 +30,6 @@ export interface LogData {
   direction?: 'forward' | 'backward';
 }
 
-export interface LogStream {
-  labels: Labels;
-  entries: LogEntry[];
-}
-
 export interface LogMetadata {
   notices?: Notice[];
 
--- a/ui/core/src/model/log-data.ts
+++ b/ui/core/src/model/log-data.ts
@@ -25,7 +25,7 @@ export interface LogData {
   timeRange?: AbsoluteTimeRange;
   entries: LogEntry[];
   metadata?: LogMetadata;
-  totalCount: number;
+  totalCount?: number;
   hasMore?: boolean;
   direction?: 'forward' | 'backward';
 }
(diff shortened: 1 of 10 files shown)


ui/core/src/model/log-data.ts
```diff
@@ -0,0 +1,74 @@
...
+  timestamp: number;
+  line: string;
+  labels: Labels;
+}
+
+export interface LogData {
+  timeRange?: AbsoluteTimeRange;
+  entries: LogEntry[];
+  metadata?: LogMetadata;
+  totalCount: number;
```
jgbernalp: not sure if all the clients will provide this value by default, in some cases might be an estimation, it will require a new query or if the pagination is cursor based this probably doesn't exist. Maybe we can make it optional

```suggestion
  totalCount?: number;
```

jgbernalp:
@abelyakin small comment other than that LGTM.

Nexucis:
@abelyakin regarding your last comment for the Loki & ClickHouse plugin, do you need to update this PR ?

ui/core/src/model/log-data.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface LogData {
+  timeRange?: AbsoluteTimeRange;
+  entries: LogEntry[];
+  metadata?: LogMetadata;
+  totalCount?: number;
+  hasMore?: boolean;
+  direction?: 'forward' | 'backward';
+}
+
+export interface LogStream {
```
Gladorme: `LogStream` correspond to what? I don't see it referenced or called anywhere 🤔

abelyakin: Good catch! deleted

abelyakin:
> @abelyakin regarding your last comment for the Loki & ClickHouse plugin, do you need to update this PR ?

@Nexucis  Yes, this PR would need updates, since logQuery response should be converted into a unified logs format. But as I mentioned in the Loki datasource plugin PR, my suggestion is to handle that as a separate follow-up PR (new feature), rather than mixing it into the current changes.

Is that okay?

Nexucis:
yeah it's ok @abelyakin let's start by that and we will iterate base on that. Thank you for your help already !

Issue #3018, opened 2025-06-13
`percli plugin lint`: add support for unit tests


AntoineThebaud:
### Proposal

Currently, `percli plugin lint` just ensures that the plugin schema(s) have the right format (= `kind` is defined, files belong to the `model` package) and that's it. We could enhance it to optionally accept unit tests, in the form of JSON files providing a payload that should match (or not?) the corresponding schema. This would allow plugin developers to validate their schemas against sample inputs, not just against the (very permissive) plugin format.

PR #3168, opened 2025-08-10
[FEATURE] percli: new command `plugin test-schemas`

AntoineThebaud:
# Description

`percli plugin test-schemas` allows plugin developers to define unit tests for their CUE files (both model & migration schemas). More info in the helper of the command.

Example of output:
```bash
$ ./bin/percli plugin test-schemas --plugin.path internal/cli/cmd/plugin/testschemas/testdata/my-datasource-plugin
✓ basic-datasource.json (model-valid) [my-datasource\tests\valid\basic-datasource.json]
✗ empty-url.json (model-invalid) [my-datasource\tests\invalid\empty-url.json]: Expected invalid data but validation succeeded
✓ basic-migration (migrate) [my-query\migrate\tests\basic-migration]
✗ basic-migration (migrate) [my-variable\migrate\tests\basic-migration]: Migration output mismatch.
Diff:  {
   "kind": "MyVariable",
   "spec": {
     "options": [
       "value1",
       "value2",
-      "value3"
+      "valueTypo"
     ]
   }
 }
✓ basic-variable.json (model-valid) [my-variable\tests\valid\basic-variable.json]
✓ invalid-options.json (model-invalid) [my-variable\tests\invalid\invalid-options.json]

Test Results: 4 passed, 2 failed
Error: 2 test(s) failed

```

Closes https://github.com/perses/perses/issues/3018 (after discussion we decided to make this a new command instead of adding stuff to `lint`).

NB: I'm planning to use this new command to break down the [big migration tests](https://github.com/perses/perses/tree/main/internal/api/plugin/migrate/testdata) we have currently in this repo into smaller pieces in perses/plugins.

# Checklist

### DIFF
--- a/internal/api/plugin/migrate/migrate.go
+++ b/internal/api/plugin/migrate/migrate.go
@@ -134,8 +134,8 @@ func GetPluginKind(migrateFile string) (plugin.Kind, error) {
 	return plugin.KindQuery, nil
 }
 
-// ExecuteCuelangScript executes a CUE migration script against grafana data
-func ExecuteCuelangScript(cueScript *build.Instance, grafanaData []byte, defID string, typeOfDataToMigrate string) (*common.Plugin, bool, error) {
+// executeCuelangScript executes a CUE migration script against grafana data
+func executeCuelangScript(cueScript *build.Instance, grafanaData []byte, defID string, typeOfDataToMigrate string) (*common.Plugin, bool, error) {
 	ctx := cuecontext.New()
 	grafanaValue := ctx.CompileString(fmt.Sprintf("%s: _", defID))
 	grafanaValue = grafanaValue.FillPath(
--- a/internal/api/plugin/migrate/panel.go
+++ b/internal/api/plugin/migrate/panel.go
@@ -145,9 +145,9 @@ func migrateQuery(queries map[string]*queryInstance, target json.RawMessage, res
 }
 
 func ExecuteQueryScript(cueScript *build.Instance, grafanaQueryData []byte) (*common.Plugin, bool, error) {
-	return ExecuteCuelangScript(cueScript, grafanaQueryData, "#target", "query")
+	return executeCuelangScript(cueScript, grafanaQueryData, "#target", "query")
 }
 
 func ExecutePanelScript(cueScript *build.Instance, grafanaPanelData []byte) (*common.Plugin, bool, error) {
-	return ExecuteCuelangScript(cueScript, grafanaPanelData, "#panel", "panel")
+	return executeCuelangScript(cueScript, grafanaPanelData, "#panel", "panel")
 }
--- a/internal/api/plugin/migrate/variable.go
+++ b/internal/api/plugin/migrate/variable.go
@@ -177,5 +177,5 @@ func migrateTextVariable(v TemplateVar) *dashboard.Variable {
 }
 
 func ExecuteVariableScript(cueScript *build.Instance, grafanaVariableData []byte) (*common.Plugin, bool, error) {
-	return ExecuteCuelangScript(cueScript, grafanaVariableData, "#var", "variable")
+	return executeCuelangScript(cueScript, grafanaVariableData, "#var", "variable")
 }


internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,523 @@
...
+	}
+	outputBuilder.WriteString(fmt.Sprintf("\nTest Results: %d passed, %d failed\n", passed, failed))
+
+	if failed > 0 {
+		outputBuilder.WriteString(fmt.Sprintf("%d test(s) failed", failed))
+	} else {
+		outputBuilder.WriteString("All schema tests passed\n")
+	}
+
+	return output.HandleString(o.writer, outputBuilder.String())
```
AntoineThebaud: Should Execute() return an error if some tests failed unexpectedly?

Nexucis: yes definitely it should return an error

internal/api/plugin/migrate/migrate.go
```diff
@@ -133,7 +134,8 @@ func getPluginKind(migrateFile string) (plugin.Kind, error) {
 	return plugin.KindQuery, nil
 }
 
-func executeCuelangMigrationScript(cueScript *build.Instance, grafanaData []byte, defID string, typeOfDataToMigrate string) (*common.Plugin, bool, error) {
+// ExecuteCuelangMigrationScript executes a CUE migration script against grafana data
+func ExecuteCuelangMigrationScript(cueScript *build.Instance, grafanaData []byte, defID string, typeOfDataToMigrate string) (*common.Plugin, bool, error) {
```
Nexucis: since it's in the migration package and it is now exposed, perhaps it makes sense to rename it by `ExecuteCulangScript`.  WDYT ?

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+			failed++
+			outputBuilder.WriteString(fmt.Sprintf("✗ %s (%s) [%s]: %s\n", result.TestName, result.TestType, result.TestPath, result.Error))
+		}
+	}
+	outputBuilder.WriteString(fmt.Sprintf("\nTest Results: %d passed, %d failed\n", passed, failed))
+
+	if failed > 0 {
+		outputBuilder.WriteString(fmt.Sprintf("%d test(s) failed", failed))
+	} else {
+		outputBuilder.WriteString("All schema tests passed\n")
```
Nexucis: I think these messages are not necessary as you already printed the result.
I think you can just return an error in case there is at least one test failure and otherwise nothing.

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+			TestName: testName,
+			TestPath: relPath,
+			TestType: TestTypeMigrate,
+		}
+
+		// Look for input.json and expected.json in this directory
+		inputPath := filepath.Join(currentPath, "input.json")
+		expectedPath := filepath.Join(currentPath, "expected.json")
+
+		if _, err := os.Stat(inputPath); os.IsNotExist(err) {
```
Nexucis: `os.IsNotExist` will only check if the error is matching the "not existing file". But if there is any other error, this check will be false but you will still have an issue with the file.

An alternative would be to do something like that:

```go
if _, err := os.Stat(inputPath); err != nil {
	result.Error = fmt.Sprintf("issue while finding the file input.json: %s", err),
	results = append(results, result)
	return filepath.SkipDir
}
```

Nexucis: actually I would just remove these pre-checks. Reading the file will fail immediately if the file does not exist with an accurate error. So it's enough

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+			results = append(results, result)
+			return filepath.SkipDir
+		}
+
+		// Determine the definition ID and type based on the plugin kind
+		var defID, typeOfData string
+		// Set default definition ID and type based on the plugin kind from the loaded schema
+		switch pluginKind {
+		case v1plugin.KindVariable:
+			defID = "#var"
```
Nexucis: I think here you should run the method `migrate.ExecuteVariableMigrationScript` instead. Like that the logic of the defID will remain in the package `plugin/migrate` and won't be spread / copy in another package

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+		logrus.Debugf("Migration results - success: %v, empty: %v, error: %v", resultPlugin != nil, resultIsEmpty, err)
+
+		if err != nil {
+			result.Error = fmt.Sprintf("Migration execution failed: %v", err)
+			results = append(results, result)
+			return filepath.SkipDir
+		}
+
+		if resultIsEmpty || resultPlugin == nil {
+			result.Error = "Migration resulted in ignored or nil result"
```
Nexucis:
```suggestion
result.Error = "Migration resulted is ignored or is empty"
```

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+		actualBytes, _ := json.Marshal(resultValue)
+
+		// Compare JSON outputs (normalized)
+		var actualNormalized, expectedNormalized interface{}
+		if err := json.Unmarshal(actualBytes, &actualNormalized); err != nil {
+			result.Error = fmt.Sprintf("Failed to parse actual migration output: %v", err)
+			results = append(results, result)
+			return filepath.SkipDir
+		}
+		if err := json.Unmarshal(expectedData, &expectedNormalized); err != nil {
```
Nexucis, line 391: you don't need to unmarshal the expected result as you already have the data in memory.

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+		// Convert back to map for comparison
+		resultValue := map[string]interface{}{
+			"kind": resultPlugin.Kind,
+			"spec": resultPlugin.Spec,
+		}
+		actualBytes, _ := json.Marshal(resultValue)
+
+		// Compare JSON outputs (normalized)
+		var actualNormalized, expectedNormalized interface{}
+		if err := json.Unmarshal(actualBytes, &actualNormalized); err != nil {
```
Nexucis: same for the actualNormalized, I don't see the value of Unmarshalling it into a generic struct since you already have the data in memory in a slice of byte

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -0,0 +1,531 @@
...
+		}
+
+		actualNormalizedBytes, _ := json.Marshal(actualNormalized)
+		expectedNormalizedBytes, _ := json.Marshal(expectedNormalized)
+
+		if string(actualNormalizedBytes) == string(expectedNormalizedBytes) {
+			result.Success = true
+		} else {
+			result.Error = fmt.Sprintf("Migration output mismatch.\nExpected: %s\nActual: %s",
+				string(expectedNormalizedBytes), string(actualNormalizedBytes))
```
Nexucis: perhaps here we can do a diff or something like that

Nexucis (changes requested):
So cool you made it @AntoineThebaud.
I have added a couple of suggestions to simplify the code.

But other-all it looks more than ok

internal/cli/cmd/plugin/testschemas/test-schemas.go
AntoineThebaud: Indeed, idk why I didnt think about that 😅

internal/cli/cmd/plugin/testschemas/test-schemas.go
AntoineThebaud: alright, btw in the end I'm going for writing bit by bit to the output directly, instead of building & printing the full report string at once

internal/cli/cmd/plugin/testschemas/test-schemas.go
AntoineThebaud: Actually after simplification of this part I still need a normalization of expectedData here for proper diff computation, for the expected data only, otherwise e.g you get false positive with same fields not in the same order

internal/api/plugin/migrate/migrate.go
```diff
@@ -134,8 +134,8 @@ func GetPluginKind(migrateFile string) (plugin.Kind, error) {
 	return plugin.KindQuery, nil
 }
 
-// ExecuteCuelangMigrationScript executes a CUE migration script against grafana data
-func ExecuteCuelangMigrationScript(cueScript *build.Instance, grafanaData []byte, defID string, typeOfDataToMigrate string) (*common.Plugin, bool, error) {
+// ExecuteCuelangScript executes a CUE migration script against grafana data
+func ExecuteCuelangScript(cueScript *build.Instance, grafanaData []byte, defID string, typeOfDataToMigrate string) (*common.Plugin, bool, error) {
```
Nexucis: I suppose you don't need to expose this function since you are using the more specifics one

AntoineThebaud: indeed!

Nexucis (approved):
nice ! Thank you !

PR #3190, opened 2025-08-14
[ENHANCEMENT] CLI/PLUGIN: schema files are watched and reloaded when modified

Nexucis:
The CLI is watching any Cuelang files contained in the `schema` folder. 
Anytime a change occurred on any files watched, the CLI is calling the API to reload the schema and the migration file.

This should complete the plugin development and help to load or reload a plugin in development without having to re-start the Perses server.

### DIFF
--- a/internal/api/impl/v1/plugin/endpoint.go
+++ b/internal/api/impl/v1/plugin/endpoint.go
@@ -44,7 +44,7 @@ func (e *endpoint) CollectRoutes(g *route.Group) {
 	if e.enableDev {
 		group.POST("", e.PushDevPlugin, true)
 		group.DELETE(fmt.Sprintf("/:%s", utils.ParamName), e.DeleteDevPlugin, true)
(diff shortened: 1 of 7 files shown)


internal/api/impl/v1/plugin/endpoint.go
```diff
@@ -44,6 +44,7 @@ func (e *endpoint) CollectRoutes(g *route.Group) {
 	if e.enableDev {
 		group.POST("", e.PushDevPlugin, true)
 		group.DELETE(fmt.Sprintf("/:%s", utils.ParamName), e.DeleteDevPlugin, true)
+		group.POST(fmt.Sprintf("/:%s/refresh", utils.ParamName), e.PushDevPlugin, true)
```
Gladorme:
```suggestion
		group.POST(fmt.Sprintf("/:%s/refresh", utils.ParamName), e.RefreshDevPlugin, true)
```

Gladorme (approved):

Issue #3187, opened 2025-08-13
Panel Form: button `run query` does not work


Nexucis:
### What did you do?

1. Edit a panel.
2. Enter a query
3. Click on the button `run query`

### What did you expect to see?

The query is ran and the preview is refreshed.

### Perses version

v0.52.0-beta.3

PR #3192, opened 2025-08-14
[BUGFIX] Add watch_query_changes to fix run query button issue

shahrokni:
Relates to #3187 

> 🔍  Read the PR before you review anything.

> ⚠️ I would like to have @jgbernalp  and @Nexucis  approval as well. Do not merge unless they have tested and approved this fix

> ⚠️  This change will affect Explorer, Edit Panel View, (**Prometheus and Tempo**)

> 🧪  Please test before merge

> 💀   This change will not cover similar onBlur issues of Legend and Min Step. They will have their own dedicated issues

## The Bug 🐛 

`Run Query Button` as its name suggests is responsible for running an **_updated_** query. At the moment, this button does not function properly, and the query-execution is triggered by onBlur event. 

## The Change 🔧 

This change is tracking the changes of the query using a watch function. When, the button is pressed, the spec is updated, and subsequently the query is executed. I would like to mention that the button and query editor are disconnected, and the query editor is a plugin which is deduced on run-time. 

## Test 🧪 

1. In any Edit Panel > Change the query, trigger the onBlur > The Query Should NOT run
2. In any Edit Panel > Change the query, press the button > The Query Should run
3. Please repeat the same steps in Explorer View for both Prometheus and Tempo / Update Plugins-System manually first
4. If the query has not changed, pressing button will not do anything
5. @Nexucis Running query on enter is not possible, because multi-line query is possible. Enter should simply break the line 

the_button

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.test.tsx
+++ b/ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.test.tsx
@@ -1,0 +1,78 @@
+// Copyright 2025 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { screen, cleanup } from '@testing-library/react';
+import { QueryDefinition, QueryPluginType } from '@perses-dev/core';
+import { renderWithContext } from '../../test';
+import { QueryEditorContainer } from './QueryEditorContainer';
+
+describe('QueryEditorContainer', () => {
+  beforeEach(() => {
+    global.fetch = jest.fn(() =>
+      Promise.resolve({
+        json: () => Promise.resolve({ success: true }),
+      })
+    ) as jest.Mock;
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    cleanup();
+  });
+
+  const querySpec = { query: 'any-fake-query', limit: 20, datasource: undefined };
+  const renderingProps: Record<string, { queryTypes: string[]; value: object }> = {
+    TimeSeriesQuery: {
+      queryTypes: ['TimeSeriesQuery'],
+      value: {
+        kind: 'TimeSeriesQuery',
+        spec: {
+          plugin: {
+            kind: 'PrometheusTimeSeriesQuery',
+            spec: querySpec,
+          },
+        },
+      },
+    },
+    TraceQuery: {
+      queryTypes: ['TraceQuery'],
+      value: {
+        kind: 'TraceQuery',
+        spec: {
+          plugin: {
+            kind: 'TempoTraceQuery',
+            spec: querySpec,
+          },
+        },
+      },
+    },
+  };
+
+  Object.keys(renderingProps).forEach((key) => {
+    it(`should render ${key} with run query button`, () => {
+      renderWithContext(
+        <QueryEditorContainer
+          queryTypes={renderingProps[key]?.queryTypes as QueryPluginType[]}
+          index={1}
+          query={renderingProps[key]?.value as QueryDefinition}
+          isCollapsed={false}
+          onDelete={jest.fn()}
+          onChange={jest.fn()}
+          onCollapseExpand={jest.fn()}
+        />
+      );
+      const runQuerybutton = screen.getByTestId('run_query_button');
+      expect(runQuerybutton).toBeInTheDocument();
+    });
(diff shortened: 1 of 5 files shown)


jgbernalp:
With changes that affect several views I think it will be useful to add an e2e test or unit tests in the code, whichever applies best to the changes, following the UI checklist. So the manual testing the reviewer does is additional.

ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
```diff
@@ -31,6 +31,31 @@ export function PluginEditor(props: PluginEditorProps): ReactElement {
...
+
+  const {
+    selection: { kind, type },
+  } = value;
+
+  let queryHandlerSettings = undefined;
+
+  if (
+    (kind === 'PrometheusTimeSeriesQuery' && type === 'TimeSeriesQuery') ||
+    (kind === 'TempoTraceQuery' && type === 'TraceQuery')
```
jgbernalp: Is there any way we can avoid hardcoding the query and datasource kinds?

shahrokni: @jgbernalp 
I am not sure what you mean by hardcoding. I was also thinking about a more innovative idea than a simple string comparison (without making things too much complex), but couldn't find a better way. Maybe you can suggest something. 

**_I just made it a bit more organized and readable_**

**Update**

```
Technically, we don't need to have that conditional statement,
I think we can simply get rid of it, watching a query is not dependent upon the implementations.
Let's say, as long as we can edit a query expression, we should be able to execute it using the button.
```

shahrokni:
> With changes that affect several views I think it will be useful to add an e2e test or unit tests in the code, whichever applies best to the changes, following the UI checklist. So the manual testing the reviewer does is additional.

I added 2 unit tests to make sure the button is rendered and functional accordingly. I see room for e2e, though first we need to have the data-testid merged, to get to the PromQL Expression. The data-testid is in the plugin repo. 

Potential Test Scenario

1. Change the Expression,
2. Trigger onBlur
3. Nothing should happen
4. Click on the button
5. Something should happen

Also, I am thinking about to write some comprehensive tests for that area of the repo (all together), because it seems we are not covering that part if I am not msitaken. 

```typescript
 {/* TODO: We need to wait for this to be merged, then we need to add proper e2e for some scenarios */}
      <CodeMirror
        data-testid="promql_expression_editor"
```

ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
```diff
@@ -31,6 +31,32 @@ export function PluginEditor(props: PluginEditorProps): ReactElement {
...
+  const runQueryHandler = useCallback((): void => {
+    onSpecChange({ ...value.spec, query: watchedQuery });
+  }, [value.spec, onSpecChange, watchedQuery]);
+
+  const {
+    selection: { kind, type },
+  } = value;
+
+  const withRunQueryButtonPairs: Record<string, string> = {
+    PrometheusTimeSeriesQuery: 'TimeSeriesQuery',
```
AntoineThebaud: @jgbernalp's [remark](https://github.com/perses/perses/pull/3192#discussion_r2276970804) is still valid here. perses/perses should be agnostic to the "concrete" implementations like **Prometheus**TimeSeriesQuery, **Tempo**TraceQuery etc because the goal is to support plugins even unofficial ones we may not be aware of

shahrokni: Technically, we don't need to have that conditional statement, 
I think we can simply get rid of it, watching a query is not dependent upon the implementations. 
Let's say, as long as we can edit a query expression, we should be able to execute it using the button. 

Would you agree? @AntoineThebaud

Nexucis:
> @Nexucis Running query on enter is not possible, because multi-line query is possible. Enter should simply break the line

But we could do what any tchat is doing. To do multi-line query you need to press `maj + enter` and then if you press `enter`  it triggers the query.
I think that's a pretty standard usage don't you think ?

shahrokni:
> > @Nexucis Running query on enter is not possible, because multi-line query is possible. Enter should simply break the line
> 
> But we could do what any tchat is doing. To do multi-line query you need to press `maj + enter` and then if you press `enter`  it triggers the query. I think that's a pretty standard usage don't you think ?

@Nexucis Indeed it is, thank you for the suggestion,
I am just wondering, if we could move this to its own dedicated issue, and let this one to deal with Run_Query_Button only,
Is it possible?

Nexucis:
yup that's totally a different change and it will change the behaviour of the query field. So it makes totally sense to do it in another PR.

ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.test.tsx
```diff
@@ -0,0 +1,78 @@
+// Copyright 2024 The Perses Authors
```
Nexucis: does not really matter, but just for the sanity we are in 2025 :)

```suggestion
// Copyright 2025 The Perses Authors
```

Nexucis (approved):
excepting my last comment, lgtm

PR #3194, opened 2025-08-18
[ENHANCEMENT] CLI/PLUGIN: improve output of the cmd test-schemas

Nexucis:
I am adding green or red color on each test result depending if the test failed or is a success.
The message summarising the tests result is also improved.

Edit: I have also added a color for the type of the test and for the path.

Before:
image

After:
image

### DIFF
--- a/internal/cli/cmd/plugin/testschemas/test-schemas.go
+++ b/internal/cli/cmd/plugin/testschemas/test-schemas.go
@@ -22,6 +22,7 @@ import (
 
 	"cuelang.org/go/cue/build"
 	"cuelang.org/go/cue/cuecontext"
+	"github.com/fatih/color"
 	"github.com/kylelemons/godebug/diff"
 	"github.com/perses/perses/internal/api/plugin"
 	"github.com/perses/perses/internal/api/plugin/migrate"
@@ -84,6 +85,10 @@ func (o *option) Validate() error {
 }
 
 func (o *option) Execute() error {
+	red := color.New(color.FgRed)
+	green := color.New(color.FgGreen)
+	cyan := color.New(color.FgCyan)
+	magenta := color.New(color.FgHiMagenta)
 	npmPackageData, readErr := plugin.ReadPackage(o.cfg.FrontendPath)
 	if readErr != nil {
 		return fmt.Errorf("unable to read plugin package.json: %w", readErr)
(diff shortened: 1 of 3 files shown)


internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -107,31 +110,25 @@ func (o *option) Execute() error {
...
-			if err := output.HandleString(o.writer, fmt.Sprintf("✓ %s (%s) [%s]", result.TestName, result.TestType, result.TestPath)); err != nil {
-				return err
+			if outputErr := output.HandleString(o.writer, fmt.Sprintf("%s %s (%s) [%s]", green.Sprint("✓"), result.TestName, result.TestType, result.TestPath)); outputErr != nil {
+				return outputErr
 			}
 		} else {
 			failed++
-			if err := output.HandleString(o.writer, fmt.Sprintf("✗ %s (%s) [%s]: %s", result.TestName, result.TestType, result.TestPath, result.Error)); err != nil {
-				return err
+			if outputErr := output.HandleString(o.writer, fmt.Sprintf("%s %s (%s) [%s]: %s", red.Sprint("x"), result.TestName, result.TestType, result.TestPath, result.Error)); outputErr != nil {
```
AntoineThebaud:
```suggestion
			if outputErr := output.HandleString(o.writer, fmt.Sprintf("%s %s (%s) [%s]: %s", red.Sprint("✗"), result.TestName, result.TestType, result.TestPath, result.Error)); outputErr != nil {
```

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -107,31 +110,25 @@ func (o *option) Execute() error {
...
-		return err
-	}
-
 	if failed > 0 {
-		return fmt.Errorf("%d test(s) failed", failed)
+		return fmt.Errorf("%s: %d out of the %d test(s) failed", red.Sprint("ERROR"), failed, len(results))
 	}
 
-	return output.HandleString(o.writer, "All schema tests passed")
+	return output.HandleString(o.writer, fmt.Sprintf("%s: All of the %d schema tests passed\n", green.Sprint("SUCCESS"), len(results)))
```
AntoineThebaud:
```suggestion
	return output.HandleString(o.writer, fmt.Sprintf("%s: All of the %d schema test(s) passed\n", green.Sprint("SUCCESS"), len(results)))
```

internal/cli/cmd/plugin/testschemas/test-schemas_test.go
```diff
@@ -28,13 +28,13 @@ func TestPluginTestSchemasCMD(t *testing.T) {
 			Title:                "Schema tests for a panel plugin, all succeeding",
 			Args:                 []string{"--plugin.path", filepath.Join(projectPath, "internal", "cli", "cmd", "plugin", "testschemas", "testdata", "my-panel-plugin")},
 			IsErrorExpected:      false,
-			ExpectedRegexMessage: "Test Results: 6 passed, 0 failed\nAll schema tests passed",
+			ExpectedRegexMessage: "SUCCESS: All of the 6 schema tests passed\n",
```
AntoineThebaud:
```suggestion
			ExpectedRegexMessage: "SUCCESS: All of the 6 schema tests passed",
```

not needed as this is an Expected**Regex**Message ;) (to just check the last line)

Nexucis: Actually I think I would prefer to keep this end of line in the test. So it's not removed by mistake next time we are modifying the code.
Because I have added it a new end of line on purpose (in the code I mean)

AntoineThebaud: alright

PR #3208, opened 2025-08-20
[IGNORE] make <RouterProvider> optional

andreasgerstmayr:
# Description

Log a warning instead of throwing an exception if `<RouterProvider>` (see https://github.com/perses/perses/pull/3140 for more details) is not present.

# Screenshots

no UI changes

# Checklist

## UI Changes


Nexucis:
well if there is no RouteProvider, I believe nothing will work, right ? So if it's the case, throwing an error is the correct way to go.

andreasgerstmayr:
> well if there is no RouteProvider, I believe nothing will work, right ? So if it's the case, throwing an error is the correct way to go.

It does work, but the panels won't show any clickable links, and instead show only the link text. For example in the TraceTable panel, it will show all traces, but you can't click on them.

PR #3214, opened 2025-08-21
[FEATURE]: Add Quick Query Viewer

### LINKED ISSUE #3071: Panels should have the option to see the queries being executed from the UI
### Proposal

When troubleshooting why a panel has no data, it would be useful to see the queries that are executed without entering an editing mode in the panel, to then fix the query if necessary.

shahrokni:
Closes #3071 

## Description 🖊️ 

This is the Quick Query Viewer **kick-off** . 
> ⚠️ More interesting ideas can be added later on top. Let's have a scaffold first ...

## Test 🧪 

1. Open a project
2. Then go to a dashboard
3. All panel headers should have an extra functionality (Query Viewer)
4. Click on the icon button
5. Do you see the queries? (It should show all quires)
6. You can also open Tree View

## Screenshots 🖼️ 

queryviewer

image

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/components/Panel/PanelActions.tsx
+++ b/ui/dashboards/src/components/Panel/PanelActions.tsx
@@ -15,6 +15,7 @@ import { Stack, Box, Popover, CircularProgress, styled, PopoverPosition } from '
 import { isValidElement, PropsWithChildren, ReactNode, useMemo, useState } from 'react';
 import { InfoTooltip } from '@perses-dev/components';
 import { QueryData } from '@perses-dev/plugin-system';
+import DatabaseSearch from 'mdi-material-ui/DatabaseSearch';
 import ArrowCollapseIcon from 'mdi-material-ui/ArrowCollapse';
 import ArrowExpandIcon from 'mdi-material-ui/ArrowExpand';
 import PencilIcon from 'mdi-material-ui/PencilOutline';
@@ -50,6 +51,9 @@ export interface PanelActionsProps {
     isPanelViewed?: boolean;
     onViewPanelClick: () => void;
   };
+  viewQueriesHandler?: {
+    onClick: () => void;
+  };
   queryResults: QueryData[];
   pluginActions?: ReactNode[];
 }
@@ -64,6 +68,7 @@ const ConditionalBox = styled(Box)({
 export const PanelActions: React.FC<PanelActionsProps> = ({
   editHandlers,
   readHandlers,
+  viewQueriesHandler,
   extra,
   title,
   description,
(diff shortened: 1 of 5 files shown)


ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -236,7 +253,9 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
         </OnHover>
         {divider} {queryStateIndicator}
         <OnHover>
-          {extraActions} {readActions}
+          {extraActions}
+          {viewQueryAction}
```
jgbernalp: This probably should be inside the overflow menu, we have a lot of actions and they are included in a menu for smaller screens

shahrokni: > This probably should be inside the overflow menu, we have a lot of actions and they are included in a menu for smaller screens

Sorry, I did not get the point,
It is already in the OverFlowMenu. Are you suggesting, it should be removed from here?

jgbernalp:
@shahrokni awesome, this is definitely useful but I see we are reusing the plugin spec editor, and passing a no op when changing values to make it read only. The problem I see is with UX, users would think they can adjust the query here and change values like the datasource, legend, etc. But nothing will happen, and it might be considered an error. Should we at least have an attribute to mark the plugin editor as readonly? this will allow plugin developers to disable/hide controls when using a readonly mode.

ui/dashboards/src/components/Panel/PanelActions.tsx
jgbernalp: It is not, it should be inside `<OverflowMenu>`

andreasgerstmayr: If you go to the "Node Explorer" dashboard, and hover over the "Sys Load (5m avg)" panel, it looks like this in Chrome:
image

(this is a "medium panel width" panel). Moving it to `<OverflowMenu>` should fix it.

shahrokni:
> @shahrokni awesome, this is definitely useful but I see we are reusing the plugin spec editor, and passing a no op when changing values to make it read only. The problem I see is with UX, users would think they can adjust the query here and change values like the datasource, legend, etc. But nothing will happen, and it might be considered an error. Should we at least have an attribute to mark the plugin editor as readonly? this will allow plugin developers to disable/hide controls when using a readonly mode.

Indeed this is misleading. That is my concern too. The point is that there is a `isReadOnly` field which doesn't disable the controls. I think, since this component is used in Edit Forms, and in Edit forms you can change the query and see results quickly (without saving), the flag doesn't disable the controls **but simply disable the `Save Button`**. (Or something like that)

To resolve this issue, if I am not mistaken, we need to change some stuff around plugins first. Because, PluginSpecEditor is only  a wrapper on top of the 

```typescript
const { data: plugin, isLoading, error } = usePlugin(pluginType, pluginKind);
/* Some code here*/
 const { OptionsEditorComponent } = plugin;
  return OptionsEditorComponent ? <OptionsEditorComponent {...others} /> : null;
```

If you allow me we resolve this matter, first with a separate PR for plugins, and then again we get back here and add some improvements.

jgbernalp:
sure if there is a property already available, then is responsibility of the plugin to do it. Thnx for the clarification

Issue #3188, opened 2025-08-13
Variable Form: Refresh button for the preview does not work


Nexucis:
### What did you do?

1. Edit a variable
2. Click on the refresh button

### What did you expect to see?

I was expecting to see the query matching the variable configuration is re-sent and the preview is refreshed with the result

### Perses version

v0.52.0-beta.3

rakshitjain23:
@Nexucis Can i work on this?

Nexucis:
hello @rakshitjain23 @shahrokni is working on this issue so excepting if he is moving on something else, it's better to let him fixing this issue :)

PR #3231, opened 2025-08-26
[BUGFIX] Add refresh token to the variable fetch

shahrokni:
Closes https://github.com/perses/perses/issues/3188

## Description 🖊️ 

The refresh token is not included in the query  key , so the result is cached, and with every refresh action the cached result is returned. To solve this problem, the refresh token should be added to the query key. 

## Test 🧪 

1. Go to the Edit/Add Variables
2. Find preview labels
3. Set the type to the list
4. Set the type on Prometheus Name Variables
5. Press the refresh button
6. Check the network traffic from the developer tool
7. With every single button press, a request (technically 2, because of an existing bug) should be made.

> http://localhost:3000/proxy/projects/quick_query_view/datasources/newdatasource/api/v1/labels

## Warning - Existing Bug ⚠️ 🐞 

There has been already a bug which makes an extra request call to fetch the labels. 
Commented in the code to be addressed later. 

```typescript
 {/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
          {/* 
              TODO: What is cool about this? This makes an extra request to the server
              This is already a bug. Should be removed after investigation
          */}
          <ClickAwayListener onClickAway={() => refreshPreview()}>
            <Box />
          </ClickAwayListener>
```

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
+++ b/ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
@@ -29,9 +29,10 @@ import { DiscardChangesConfirmationDialog, ErrorAlert, ErrorBoundary, FormAction
 import { Control, Controller, FormProvider, SubmitHandler, useForm, useFormContext, useWatch } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { getSubmitText, getTitleAction } from '../../../utils';
-import { VARIABLE_TYPES } from '../variable-model';
 import { PluginEditor } from '../../PluginEditor';
 import { useValidationSchemas } from '../../../context';
+import { VARIABLE_TYPES } from '../variable-model';
+import { useTimeRange } from '../../../runtime';
 import { VariableListPreview, VariablePreview } from './VariablePreview';
 
 function FallbackPreview(): ReactElement {
@@ -108,7 +109,9 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
    * having to re-fetch the values when the user is still editing the spec.
    */
   const [previewSpec, setPreviewSpec] = useState<ListVariableDefinition>(form.getValues() as ListVariableDefinition);
+  const { refresh } = useTimeRange();
   const refreshPreview = (): void => {
+    refresh();
     setPreviewSpec(form.getValues() as ListVariableDefinition);
   };
 
@@ -150,10 +153,13 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
 
         <Stack>
           {/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
+          {/* 
+              TODO: What is cool about this? This makes an extra request to the server
+              This is already a bug. Should be removed after investigation
+          */}
           <ClickAwayListener onClickAway={() => refreshPreview()}>
             <Box />
           </ClickAwayListener>
-          {/** **/}
           <ErrorBoundary FallbackComponent={ErrorAlert}>
             <Controller
               control={control}
--- a/ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
+++ b/ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useMemo, useState } from 'react';
 import { Alert, Box, Card, Chip, CircularProgress, IconButton, Stack, Typography } from '@mui/material';
 import { InfoTooltip, useSnackbar } from '@perses-dev/components';
 import Refresh from 'mdi-material-ui/Refresh';
@@ -90,16 +90,21 @@ interface VariableListPreviewProps {
 }
 
 export function VariableListPreview(props: VariableListPreviewProps): ReactElement {
-  const { definition, onRefresh } = props;
+  const { onRefresh, definition } = props;
   const { data, isFetching, error } = useListVariablePluginValues(definition);
   const errorMessage = (error as Error)?.message;
 
-  return (
-    <VariablePreview
-      values={data?.map((val) => val.value) || []}
-      onRefresh={onRefresh}
-      isLoading={isFetching}
-      error={errorMessage}
-    />
+  const variablePreview = useMemo(
+    () => (
+      <VariablePreview
+        values={data?.map((val) => val.value) || []}
+        onRefresh={onRefresh}
+        isLoading={isFetching}
+        error={errorMessage}
+      />
+    ),
+    [errorMessage, isFetching, onRefresh, data]
   );
+
+  return variablePreview;
 }
--- a/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
+++ b/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
@@ -73,6 +73,13 @@ export function useSuggestedStepMs(width?: number): number {
 export function TimeRangeProvider(props: TimeRangeProviderProps): ReactElement {
   const { timeRange, refreshInterval, children, setTimeRange, setRefreshInterval } = props;
 
+  /**
+   * TODO: The hook needs refactor. There is a bug here with refreshKey. If the refreshInterval is not set,
+   * refreshKey string includes an undefined xx:yy:zz:undefined:0
+   * I think exposing refresh functionality also is very risky. A careless usage of refresh may cause
+   * infinite rendering loop.
+   */
+
   const [localTimeRange, setLocalTimeRange] = useState<TimeRangeValue>(timeRange);
   const [localRefreshInterval, setLocalRefreshInterval] = useState<DurationString | undefined>(refreshInterval);
 


ui/plugin-system/src/components/Variables/variable-model.ts
```diff
@@ -39,11 +39,14 @@ export function filterVariableList(data: VariableOption[], capturedRegexp: RegEx
...
 
-export function useListVariablePluginValues(definition: ListVariableDefinition): UseQueryResult<VariableOption[]> {
+export function useListVariablePluginValues(
+  definition: ListVariableDefinition,
+  refreshToken?: string
+): UseQueryResult<VariableOption[]> {
   const { data: variablePlugin } = usePlugin('Variable', definition.spec.plugin.kind);
   const datasourceStore = useDatasourceStore();
   const allVariables = useAllVariableValues();
-  const { absoluteTimeRange: timeRange, refreshKey } = useTimeRange();
```
jgbernalp: why do we remove this? wouldn't be better to actually check why this refreshKey is not being set with a value that can invalidate cache?

ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -107,9 +107,14 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
    * spec that will be used for preview. The reason why we do this is to avoid
    * having to re-fetch the values when the user is still editing the spec.
    */
-  const [previewSpec, setPreviewSpec] = useState<ListVariableDefinition>(form.getValues() as ListVariableDefinition);
+  const [previewSpec, setPreviewSpec] = useState<ListVariableDefinition & { refreshToken: string }>({
+    ...form.getValues(),
+    refreshToken: String(Date.now()),
```
jgbernalp: I'd try to avoid mixing specs with other values that are specific of the fetching logic. If we are just passing just the current date all the time, then this can be also be moved to the part that does the actual fetching.

ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
```diff
@@ -85,21 +85,26 @@ export function VariablePreview(props: VariablePreviewProps): ReactElement {
 }
 
 interface VariableListPreviewProps {
-  definition: ListVariableDefinition;
+  definition: ListVariableDefinition & { refreshToken: string };
```
jgbernalp: If we really need a refreshToken I'd create a prop for it, why merge it with an unrelated definition?

ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -107,9 +107,14 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
...
+  } as ListVariableDefinition & { refreshToken: string });
+
   const refreshPreview = (): void => {
-    setPreviewSpec(form.getValues() as ListVariableDefinition);
+    const refreshToken = String(Date.now());
```
andreasgerstmayr: Can you call `refresh()` which is returned from the `useTimeRange()` hook here? This should reset the time range and re-fetch all queries.

ui/plugin-system/src/components/Variables/variable-model.ts
shahrokni: @jgbernalp 

It seems refreshKey already has a bug. 






> The point is useTimeRange is used in the body useListVariablePluginValues, meaning you can have access to the refresh functionality there. Calling the refresh, due to existing bug which has been mentioned already, causes an infinite loop of rendering. If you can suggest where it should actually be placed without the mentioned problem I will put it there. **The last part of refresh key is completed calling the refresh functionality**

```typescript
{/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
          {/* 
              TODO: What is cool about this? This makes an extra request to the server
              This is already a bug. Should be removed after investigation
          */}
          <ClickAwayListener onClickAway={() => refreshPreview()}>
            <Box />
          </ClickAwayListener>
```

ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
shahrokni: @andreasgerstmayr 
Actually, I tried to use the refresh function.
 
The point is useTimeRange is used in the body useListVariablePluginValues, meaning you can have access to the refresh functionality there. Calling the refresh, due to an existing bug which has been mentioned already, causes an infinite loop of rendering. 

If you can suggest where it should actually be placed without the mentioned problem I will put it there. 

And if I am not mistaken the root of the problem (infinite loop) is here

```typescript
{/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
          {/* 
              TODO: What is cool about this? This makes an extra request to the server
              This is already a bug. Should be removed after investigation
          */}
          <ClickAwayListener onClickAway={() => refreshPreview()}>
            <Box />
          </ClickAwayListener>
```

ui/plugin-system/src/components/Variables/variable-model.ts
jgbernalp: ok, thnx for clarifying. In that case I wonder if we need this refreshKey at all. Shouldn't the actual timeRange change as it uses the current time already? This will effectively invalidate the cache for a selected time range, without causing rendering loops. In the case of variable editor I'd assume the current time minus a predefined range is used.

ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -107,8 +107,14 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
    * spec that will be used for preview. The reason why we do this is to avoid
    * having to re-fetch the values when the user is still editing the spec.
    */
-  const [previewSpec, setPreviewSpec] = useState<ListVariableDefinition>(form.getValues() as ListVariableDefinition);
+  const [previewSpec, setPreviewSpec] = useState<ListVariableDefinition>({
+    ...form.getValues(),
+  } as ListVariableDefinition);
```
jgbernalp:
```suggestion
const [previewSpec, setPreviewSpec] = useState<ListVariableDefinition>(form.getValues() as ListVariableDefinition);
```

ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
andreasgerstmayr: I just added `refresh()` from `useTimeRange()` to the `refreshPreview` handler, I don't see any infinite loop. There are many network requests though, because it refreshes the queries of every single panel on the dashboard.

Which is kind of expected, but probably not what we want in this case. I would add another `<TimeRangeProvider>` wrapping only the variable editor, to make it independent of the TimeRange of the entire dashboard, and call the `refresh()` on it.

ui/plugin-system/src/components/Variables/variable-model.ts
shahrokni: @jgbernalp 
I think refreshKey could be used and would work just fine, if the refresh function was called carefully. 

Generally, I see some unnecessary re-rendering phases around the application by profiling the app. Therefore, refresh, if not used carefully causes a lot of issues. 

The only reason I preferred to provide a new refreshToken from outside was to avoid touching the following block, because I am not confident if my assumption about this block is correct or not. 

**_Assumption: This immediate function call probably was used to force an update, when something is changed around the form._**

```typescript
/* Hack?: Cool technique to refresh the preview to simulate onBlur event */

 /* 
 TODO: What is cool about this? This makes an extra request to the server
 This is already a bug. Should be removed after investigation
 */

<ClickAwayListener onClickAway={() => refreshPreview()}>
<Box />
</ClickAwayListener>
```   

## What is refreshKey

refreshKey is

```typescript
/* This is exposed through refresh */
const [refreshCounter, setRefreshCounter] = useState(0);
/*Some Code... */
refreshKey: `${absoluteTimeRange.start}:${absoluteTimeRange.end}:${localRefreshInterval}:${refreshCounter}`,
```

## What about timeRange? Shouldn't timeRange do the job?

Good point, but, either it is a bug or it has been intentional, it seems it is returning the same value. Same value, same query key, and eventually cached result

image

ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
shahrokni: @andreasgerstmayr 

As you mentioned correctly, pressing refresh should not cause many network requests. This is not efficient. 
I see the current fix already makes 2 (one extra because of the mentioned bug), which is ok, and the extra one would be resolved soon as well. 

image


Now the solution you came up with is interesting and can be considered, but I believe it should be resolved in a higher scope. 

> I would add another <TimeRangeProvider> wrapping only the variable editor, to make it independent of the TimeRange of the entire dashboard, and call the refresh() on it.

I am personally a bit conservative about Big-Bang changes under a small scope. However, if you are not convinced, I can think about it. Not biased at all.

andreasgerstmayr: I don't see a bug with the current behavior. If we call `refresh()` of the `<TimeSeriesProvider>`, which is a parent of the dashboard, it syncs the relative time ("Last 1 hour") to an absolute time. Naturally, all components which use the time range (all panels on the dashboard), will refresh. _This is exactly how it should work and not a bug._ It has the same effect as clicking on the refresh button on top of the dashboard.

Now, in our case, we only want to refresh the query for the variable editor. Therefore, I think it makes sense to wrap the variable editor in a different `<TimeSeriesProvider>`. Adding a custom refresh key seems like a workaround to me. What do you think?

shahrokni: @andreasgerstmayr 

All good and done. only refresh is used. No new refresh token. 

Tnx for the review and suggestions.

jgbernalp (approved):

andreasgerstmayr (approved):

PR #3235, opened 2025-08-26
BUGFIX: Remove unnecessary network call on variable refresh

### LINKED ISSUE #3234: Extra network call when the refresh button is pressed (Variable lables)
### What did you do?

- Open a dashboard
- Try to create a variable
- Select name variables 
- Clear your developer tool network tab
- Press refresh button
- There are two requests made by the client

It should only make a single request. 

The root of this issue is an immediate function call for some unknown reasons. This is probably a workaround. 
The comments already mention it might be a hack. 

```typescript
          {/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
          {/* 
              TODO: What is cool about this? This makes an extra request to the server
              This is already a bug. Should be removed after investigation
          */}
          <ClickAwayListener onClickAway={() => refreshPreview()}>
            <Box />
          </ClickAwayListener>
```

Image

### System information

_No response_

### Perses version

0.52

### Perses configuration file

```yaml

```

### Logs

```text

```

AnandRaj2224:
This PR removes the `ClickAwayListener` component that was causing a second, unnecessary network request when refreshing variable values in the editor.

Closes #3234

Hi @shahrokni, I've submitted the fix for review. Please let me know if any changes are needed. Thanks!

### DIFF
--- a/ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
+++ b/ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
@@ -152,14 +152,6 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
         )}
 
         <Stack>
-          {/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
-          {/* 
-              TODO: What is cool about this? This makes an extra request to the server
-              This is already a bug. Should be removed after investigation
-          */}
-          <ClickAwayListener onClickAway={() => refreshPreview()}>
-            <Box />
-          </ClickAwayListener>
           <ErrorBoundary FallbackComponent={ErrorAlert}>
             <Controller
               control={control}


ui/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx
```diff
@@ -152,14 +152,6 @@ function ListVariableEditorForm({ action, control }: KindVariableEditorFormProps
         )}
 
         <Stack>
-          {/** Hack?: Cool technique to refresh the preview to simulate onBlur event */}
-          {/* 
-              TODO: What is cool about this? This makes an extra request to the server
-              This is already a bug. Should be removed after investigation
-          */}
-          <ClickAwayListener onClickAway={() => refreshPreview()}>
```
shahrokni, line 160: @AnandRaj2224 

Thank you for the PR.

Sharing your ideas and assumptions is as valuable as your contribution. Therefore, I would like to ask you to share 

- How you would test this and challange the idea of removing that block of code
- What you guess about the original code, maybe the  developer had some challenges back then
- What would be your anticipation about any likely consequences.

Please add the mentioned to the PR.

AnandRaj2224: My initial anticipation was that this was dead code and its removal would simplify the codebase with no side effects. This was wrong.

The direct consequence, as shown by the failed tests, is a regression. this has now re-introducing a bug that the original developer had fixed. The system is now failing in at least two specific scenarios that this code was designed to prevent or maybe more scenarios not covered in test cases.

I will now investigate the failed tests to understand the specific scenarios I missed and Update the PR with a more targeted change that still achieves the original goal 

thank u for guiding me on the right approach as this is my first ever try at contributing to opensource.

AnandRaj2224:
After investigating further, it's clear this change is more complex than I initially anticipated and involves deep backend logic in Go. but I'm not the right person to implement the fix.

I'm closing this PR to allow a developer with more Go experience to tackle the problem correctly. Thanks for all the guidance!

PR #3244, opened 2025-08-27
[IGNORE] add panelOptions.showIcons setting

andreasgerstmayr:
# Description

Currently, we show most of the panel icons only on hover, unless the dashboard is in editing mode, or the panel is in fullscreen mode.

In the Tempo explore view, the panel is in fullscreen mode, but doesn't have `readHandlers.isPanelViewed` set (because it's always fullscreen, there is no option to change the panel size), therefore the icons (among others, the download icon) is only shown on hover, which is bad UX.

This PR adds an explicit `showIcons` setting which can be either `hover` or `always`, and by default uses the current behavior (`always` if the dashboard is in editing mode or panel is in fullscreen mode, `hover` otherwise).

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/components/Panel/Panel.tsx
+++ b/ui/dashboards/src/components/Panel/Panel.tsx
@@ -37,6 +37,11 @@ export type PanelOptions = {
    */
   hideHeader?: boolean;
   /**
+   * Whether to show panel icons always, or only when hovering over the panel.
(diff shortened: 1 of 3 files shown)


Gladorme (approved):
lgtm, I would have put enhancement tag instead of ignore 🤔

andreasgerstmayr:
> lgtm, I would have put enhancement tag instead of ignore 🤔

done :+1:

PR #3247, opened 2025-08-28
[ignore] Apply analyzer modernize to the whole codebase

Nexucis:


andreasgerstmayr:
Looks like the bulk of the changes come from updating `interface{}` to `any`.
Would it be possible to split these changes in a separate PR? I think that'd make the review of the other changes much easier.

andreasgerstmayr:
> Looks like the bulk of the changes come from updating `interface{}` to `any`. Would it be possible to split these changes in a separate PR? I think that'd make the review of the other changes much easier.

done in https://github.com/perses/perses/pull/3251

Gladorme (approved):

Issue #3209, opened 2025-08-20
PromQL tree view: placeholders instead of proper values are displayed for built-in variables


AntoineThebaud:
### What did you do?

- Use prometheus built-in interval variables such has `$__interval` or `$__rate_interval` in a promQL expression.
- Toggle the Tree view.
- Notice that instead of proper values, these variables get interpolated to "12345".
<img width="2130" height="708" alt="Image" src="https://github.com/user-attachments/assets/ab4a0f24-52d6-437c-84a3-7beae98dfa91" />

NB: this is a known issue, due to a workaround, see TODO comment https://github.com/perses/plugins/blob/30ff725d227ac5afd7dd7c234cb037e39d9b6a08/prometheus/src/components/PromQLEditor.tsx#L48-L52

### What did you expect to see?

Built-in variables such has `$__interval`, `$__rate_interval` etc should get replaced by proper values in the tree view.

### Perses version

0.52.0-beta.3

### Perses configuration file

```yaml

```

### Logs

```text

```

shahrokni:
⚠️  **_I am working on this_**

After a discussion with @jgbernalp we decided to continue with a completely new Provider to share the information of the editing panel down the React hierarchy. For this feature specifically, we are talking about the panel's width. 

## Why are we doing this? 

To calculate the Suggested Step MS we need to have absolute time range and panel's width.
Suggested Step MS  itself is the key to calculate the  built-in interval variables.

PR #3248, opened 2025-08-28
[BUGFIX] Pass the suggestedStemp MS to the QueryEditor

shahrokni:
Relates to https://github.com/perses/perses/issues/3209

> ⚠️  THERE WILL BE ALSO A PR FOR PERSES/PLUGINS AFTER THIS ONE IS MERGED

## The Bug 🐞 

In tree view instead of showing the actual values of `minStep` and `intervalMs` fake place holders are used. 
As an old comment in the code-base suggests, these values are actually calculated just before the query is executed. 

```typescript
    // TODO placeholder values for steps to be replaced with actual values
    // Looks like providing proper values involves some refactoring: currently we'd need to rely on the timeseries query context,
    // but these step values are actually computed independently / before the queries are getting fired, so it's useless to fire
    // queries here, so maybe we should extract this part to independant hook(s), to be reused here?
    
    queryExpr = replacePromBuiltinVariables(queryExpr, 12345, 12345);
```

## How these values are calculated ❓ 

An existing function located in `plugins\prometheus-time-series-query\get-time-series-data.ts` does the job. Therefore, we just need to call it with the proper parameters to get the right answer. 

## What parameters are required ❓ 

1. timeRange
2. suggestedStepMs. This is already calculated based on the width (pixels) of the Panel

**_Good news, we already have access to timeRange, we just need to get suggestedStepMs._** 

## Long story short 📖 

This PR passes down suggestedStepMs for calculation of `minStep` and `intervalMs`

image

# Checklist

## UI Changes


AntoineThebaud:
@shahrokni I will take time to review this ASAP, but just one thing rn: it's not accurate to say the current placeholder concern `minStep` and `intervalMs`, the current limitation concerns the **Prometheus Built-in Variables** which are these ones (visible in the Edit Variables drawer): 
image
Just sharing this in case this would change something in what you coded, anyway I will check soon

shahrokni:
@AntoineThebaud 

just updated the description

jgbernalp:
I wonder if instead of passing the props to many components, would it be better to use this values from the existing providers. If we see the preview as a panel in the current dashboard being edited, these default variables values should be filled out based on the current time range selection. This will avoid doing special logic for one case, if a panel is being edited and makes use of variables only defined in the dashboard it should also make the replacements.

shahrokni:
> I wonder if instead of passing the props to many components, would it be better to use this values from the existing providers. If we see the preview as a panel in the current dashboard being edited, these default variables values should be filled out based on the current time range selection. This will avoid doing special logic for one case, if a panel is being edited and makes use of variables only defined in the dashboard it should also make the replacements.

Hey @jgbernalp 
I am not sure if I got your suggestion. Could you explain your solution based on the current structure.  I share the structure here

image

Are you suggesting to put the value into one of the existing Context Providers instead of passing it down step by step? 
 **Could you suggest one? (One of the existing providers)**

jgbernalp:
From my search, we can use the `useSuggestedStepMs` hook. That requires the TimeRangeProvider in the tree. This gets the time range from the context and receives a width option to do the calculation.

shahrokni:
> From my search, we can use the `useSuggestedStepMs` hook. That requires the TimeRangeProvider in the tree. This gets the time range from the context and receives a width option to do the calculation.

@jgbernalp But then you need to pass down the width, no?
Because, you need it to call useSuggestedStepMs. We are in the Spec Editor, and we need something from Preview. Two different parallel branches.

image

```typescript
export function useSuggestedStepMs(width?: number): number {
  const { absoluteTimeRange } = useTimeRange();
  if (width === undefined) return 0;
  return getSuggestedStepMs(absoluteTimeRange, width);
}
```

Let me ask this way to be more clear: When you are calling the hook in PrometheusTimeSeriesQueryEditor, or anywhere under the SpecEditor, how would you get the width from preview? You need to lift it up from preview and somhow pass it down. Agree?

jgbernalp:
Even if we lift this state we will have prop drilling, passing down the width to the component tree until we reach the plugin spec editor (PromQL tree), so we will have the same issue. I think a cleaner solution would be to include the `suggestedStepMs` in the time range context, allow the panel preview to set it as is the component that is displaying the query results and is aware of the display width. Then, plugin developers can choose to use the `suggestedStepMs` value coming from the TimeRange context or calculate a new one, using the `useSuggestedStepMs` hook,  if the want to provide a different width, for example having a different kind of preview. WDYT?

shahrokni:
> Even if we lift this state we will have prop drilling, passing down the width to the component tree until we reach the plugin spec editor (PromQL tree), so we will have the same issue. I think a cleaner solution would be to include the `suggestedStepMs` in the time range context, allow the panel preview to set it as is the component that is displaying the query results and is aware of the display width. Then, plugin developers can choose to use the `suggestedStepMs` value coming from the TimeRange context or calculate a new one, using the `useSuggestedStepMs` hook, if the want to provide a different width, for example having a different kind of preview. WDYT?

This sounds reasonable and a potential solution. OK, let's see if I can prepare something. Thank you Gabriel.

PR #3250, opened 2025-08-28
[FEATURE] enhance variable replacement with custom formats

jgbernalp:
# Description

This PR adds support for custom formats for variable replacements using the `${varName:format}` syntax. For example a query defined as:

```
hello ${var1:json}'
```

with `var1` being an array with `perses` and `prometheus` will be processed as:

```
hello ["perses","prometheus"]
```

more formats and examples are included in the tests.

A new `parseVariablesAndFormat` function was created as the `parseVariables` is being used by plugins to get only variable names, so it was left unchanged.

# Checklist

### DIFF
--- a/ui/plugin-system/src/utils/variables.ts
+++ b/ui/plugin-system/src/utils/variables.ts
@@ -15,7 +15,8 @@ import { VariableValue } from '@perses-dev/core';
 import { VariableStateMap } from '@perses-dev/plugin-system';
 
 export function replaceVariables(text: string, variableState: VariableStateMap): string {
-  const variables = parseVariables(text);
+  const variablesMap = parseVariablesAndFormat(text);
+  const variables = Array.from(variablesMap.keys());
   let finalText = text;
   variables
     // Sorting variables by their length.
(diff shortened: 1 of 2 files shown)


ui/plugin-system/src/utils/variables.ts
```diff
@@ -25,23 +26,108 @@ export function replaceVariables(text: string, variableState: VariableStateMap):
...
-        finalText = replaceVariable(finalText, v, variable?.value);
+        finalText = replaceVariable(finalText, v, variable?.value, variablesMap.get(v));
       }
     });
 
   return finalText;
 }
 
-export function replaceVariable(text: string, varName: string, variableValue: VariableValue): string {
+export type InterpolationFormat =
```
andreasgerstmayr: nit: This could be an enum, and to get the type we could use
```
type LogLevelStrings = keyof typeof LogLevel;
```
(example from https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time)

Then we could verify the variable format at runtime and drop the big switch statement in `stringToFormat`.

andreasgerstmayr (approved):
LGTM, great work :+1:

ui/plugin-system/src/utils/variables.ts
```diff
@@ -25,23 +26,108 @@ export function replaceVariables(text: string, variableState: VariableStateMap):
...
+      return `{${values.join(',')}}`;
+    case 'json':
+      return JSON.stringify(values);
+    case 'lucene':
+      return `(${values.map((v) => `"${v}"`).join(' OR ')})`;
+    case 'percentencode':
+      return encodeURIComponent(values.join(','));
+    case 'pipe':
+      return values.join('|');
+    case 'raw':
```
andreasgerstmayr: nit: this is the same as `csv`, could use a fall-through `case`

AntoineThebaud (approved):
Awesome! We should draft a doc for this feature btw

jgbernalp:
> Awesome! We should draft a doc for this feature btw

Will do it in a separate PR. It seems the variable section is not visible in the current docs.

Issue #3266, opened 2025-08-30
[Panel]: PromQL changes not applied when clicking "Apply" without "Run Query"


ibakshay:
### What did you do?

I edited the PromQL expression by adding an extra label
I clicked the "Apply" button without pressing "Run Query"
The new label I added in the PromQL editor is not reflected in the visualization

Steps to reproduce: Edit query → Click Apply (skip Run Query) → Changes not visible

https://github.com/user-attachments/assets/7d3cea2c-69bf-41f0-92dc-0678bc033814
 

### What did you expect to see?

**Expected behavior**: Changes should apply when clicking "Apply"
**Actual behavior**: Need to click "Run query" everytime before I click "Apply" for changes to take effect

### Perses version

v0.52.0-rc.0

### Perses configuration file

```yaml

```

### Logs

```text

```

shahrokni:
@ibakshay  This is because of a recent change by which onBlur event is removed. So, as you mentioned you are forced to click on Run Query. The point you made about Apply and capturing the changes before the actual save was supposed to come later with an improvement. **_But now since you catch it quickly, I will create a PR soon._** 

_**Working on it. Coming soon!**_

ibakshay:
I was testing today and I found a bug related to this fix. So, I am reopening the issue. You can find the bug details in the below gif. 

https://github.com/user-attachments/assets/46130cf7-b7fb-46ac-b05c-5f4f354ddfff

PR #3269, opened 2025-09-01
[BREAKINGCHANGE] Cleanup spec editor before the form is submitted

shahrokni:
Closes https://github.com/perses/perses/issues/3266

# Description

Recently, the onBlur event was removed from the query editor. When a query (text) is updated, the Run Query button must be pressed; otherwise, the specs are not updated. To ensure that query changes are saved and persisted when the Apply button is pressed before the Run Query button, a cleanup request is made. This notifies all components subscribed to requestCleanupAt, allowing them to clean up their state and update any dangling state.

@ibakshay 

I would like to ask you to test it yourself. 

# Checklist

## UI Changes


shahrokni:
@andreasgerstmayr 
I have an idea instead of lifting up, 
just give some time, and then then if not convinced we continue with lifting up

https://github.com/perses/perses/pull/3269#pullrequestreview-3173950726

shahrokni:
@andreasgerstmayr 

https://github.com/perses/perses/pull/3269#pullrequestreview-3173950726


Sorry I am bit confused how the proposed comment would resolve the current issue. 
The current issue happens because,

1. The on-blur event to capture the changes is not used anymore. So, (2)
2. To capture the query changes Run Query Button should be pressed first. (this calls onSpecChange) If not pressed, then (3)
3. Specs are not fully updated, so Apply would miss the remaining changes



## Solution

Apply (from the dashboard) needs to tell all components under the PanelSpecEditor, to flush their remaining changes (using onSpecChange) before it proceeds.  This should not be restricted to the query text. Today we lift the query state up, but as the application grows, this approach may become a problem. Introducing a so-called clean up request through a provider gives the opportunity to all nested components which deal with spec to settle down all remaining changes at any level.

andreasgerstmayr:
> Sorry I am bit confused how the proposed comment would resolve the current issue.

If we lift the state of the draft/pending query up, the parent component is always aware of the (currently unsaved) query spec and we don't need `watchQueryChanges` or the changes in this PR.

On the other hand, it'll re-render the editor component on every change/keypress :/

> ## Solution
> 
> Apply (from the dashboard) needs to tell all components under the PanelSpecEditor, to flush their remaining changes (using onSpecChange) before it proceeds. This should not be restricted to the query text. Today we lift the query state up, but as the application grows, this approach may become a problem. Introducing a so-called clean up request through a provider gives the opportunity to all nested components which deal with spec to settle down all remaining changes at any level.

If we go down this route, that the parent component calls a callback (clean up request) on the child component, I would suggest using [useImperativeHandle](https://react.dev/reference/react/useImperativeHandle#usage) instead of the current workaround (updating a context, which triggers useEffect, which calls the callback. Everything inside a `setTimeout` with `0`, which may introduce race conditions). I would also avoid creating a new context if possible (maybe we could add the ref to `queryHandlerSettings`?)

ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
```diff
@@ -60,6 +61,8 @@ export function PluginEditor(props: PluginEditorProps): ReactElement {
         }
       : undefined;
   }, [withRunQueryButton]);
+  console.log({ ref });
```
andreasgerstmayr:
```suggestion
```

andreasgerstmayr (approved):
LGTM, thanks for the changes!

I already approved it, but please remove the `console.log` before merging

shahrokni:
> LGTM, thanks for the changes!
> 
> I already approved it, but please remove the `console.log` before merging

Yes sure

PR #3271, opened 2025-09-01
[BUGFIX] Grafana migration: fix invisible group generated for orphan panels

AntoineThebaud:
# Description

Migrating a Grafana dashboard with "orphan" panels (aka not belonging to a row) leads to such result on Perses:

image

Which looks fine at first glance, but if you go to the edit view for the first panels, this is what you see 

image (14)

There's a deeper fix/change to consider here (see https://github.com/perses/perses/issues/3272), this PR aims just to fix the current migration result by adding a placeholder title to the grid getting generated for orphan panels, so that it properly shows on the UI.

# Checklist

## UI Changes

### DIFF
--- a/internal/api/plugin/migrate/migrate.go
+++ b/internal/api/plugin/migrate/migrate.go
@@ -260,9 +260,6 @@ func (m *completeMigration) migrateGrid(grafanaDashboard *SimplifiedDashboard) [
 	orphansGridSpec := &dashboard.GridLayoutSpec{
 		Display: &dashboard.GridLayoutDisplay{
 			Title: "Panel Group 1", // placeholder value since we don't want an empty string.
-			Collapse: &dashboard.GridLayoutCollapse{
-				Open: true, // because orphan panels are always visible in Grafana
-			},
 		},
 	}
 	orphansGrid := dashboard.Layout{
--- a/internal/api/plugin/migrate/testdata/dashboards/panel_with_links_perses_dashboard.json
+++ b/internal/api/plugin/migrate/testdata/dashboards/panel_with_links_perses_dashboard.json
@@ -90,10 +90,7 @@
         "kind": "Grid",
         "spec": {
           "display": {
-            "title": "Panel Group 1",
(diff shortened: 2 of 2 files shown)


internal/api/plugin/migrate/migrate.go
```diff
@@ -254,17 +254,27 @@ func (m *completeMigration) Migrate(grafanaDashboard *SimplifiedDashboard) (*v1.
...
 	var result []dashboard.Layout
-	defaultSpec := &dashboard.GridLayoutSpec{}
-	defaultLayout := dashboard.Layout{
+	// This is not allowed in Perses to have "orphan" panels (a.k.a panels that don't belong to a group).
+	// Thus we have to create a first grid to gather the eventual orphans found at the beginning of the
+	// Grafana dashboard.
+	orphansGridSpec := &dashboard.GridLayoutSpec{
+		Display: &dashboard.GridLayoutDisplay{
+			Title: "Panel Group 1", // placeholder value since we don't want an empty string.
+			Collapse: &dashboard.GridLayoutCollapse{
```
Nexucis: So actually, you don't need to define the `Collapse`. `Collapse` can be undefined and then that means the group is not rendered as collapsible group.

Nexucis:
I have added a short comment to explain a bit how it is supposed to work https://github.com/perses/perses/pull/3300

internal/api/plugin/migrate/migrate.go
AntoineThebaud: Well I just tried and I cant confirm if that works effectively that way, whenever `collapse` is undefined on frontend side it gets defaulted to `collapse: open: true`. Anyway in the current case I can remove this piece of code but in the scope of https://github.com/perses/perses/issues/3272 we'll have to do a 2nd pass later probably.

Nexucis: mmm then I would say this is a bug unfortunately :/.

Nexucis (approved):

Issue #3242, opened 2025-08-27
Refreshing the query results regardless of internal state changes


shahrokni:
### What did you do?

In dashboards, when interacting with panels, repeatedly refreshing in quick succession (e.g., using the Run Query button) may not yield updated information.

## Reason
Tanstack Query may have caused this issue. It seems, when the spec is intact, by pressing the button only the cached value is returned. So, you need to change something in the Spec Editor!

> ℹ️ It was discussed here https://github.com/perses/perses/discussions/3201 in details

### What did you expect to see?

When exploring dashboards, it is logical to send GET requests whenever a panel is refreshed, even if the internal state remains unchanged. This behavior is an essential and intrinsic feature of dashboards, as time—one of the most critical parameters—is constantly progressing. Therefore, regardless of whether the internal state of the dashboard has changed, every refresh should trigger a new GET request to retrieve the most up-to-date information.

<img width="1325" height="524" alt="Image" src="https://github.com/user-attachments/assets/093dcf57-3303-4e8e-9faf-ec808130e699" />

## Details

Details can be found here

https://github.com/perses/perses/discussions/3201

### Perses version

> 0.52

PR #3287, opened 2025-09-03
[BUGFIX] Force a refresh time range when the Run Query Button is clicked everytime

shahrokni:
Closes https://github.com/perses/perses/issues/3242
Closes https://github.com/perses/perses/discussions/3201

## Description

Discussed in details https://github.com/perses/perses/issues/3242 with @andreasgerstmayr 

## The Fix 🔧 

This PR wraps the PanelEditorForm with a new Time Range. Whenever the Run Query Button is pressed a fresh time range is generated, forcing  the query to be executed. So, whether the spec has changed or not, the query is executed, and no cached result will be returned. 

## Test 🧪 

The following picture is self-explanatory 

image

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
@@ -13,7 +13,7 @@
 
 import { ReactElement, useCallback, useEffect, useRef, useState } from 'react';
 import { Box, Button, Grid, MenuItem, Stack, TextField, Typography } from '@mui/material';
-import { Action, PanelDefinition, PanelEditorValues } from '@perses-dev/core';
+import { Action, DEFAULT_DASHBOARD_DURATION, PanelDefinition, PanelEditorValues } from '@perses-dev/core';
 import { DiscardChangesConfirmationDialog, ErrorAlert, ErrorBoundary } from '@perses-dev/components';
 import {
   PluginKindSelect,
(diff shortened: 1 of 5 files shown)


ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
```diff
@@ -30,8 +30,16 @@ import { PluginEditorProps, PluginEditorRef, usePluginEditor } from './plugin-ed
...
+    value,
+    withRunQueryButton = true,
+    pluginTypes,
+    pluginKindLabel,
+    onChange: _,
+    isReadonly,
+    postExecuteRunQuery: refresh,
+    ...others
+  } = props;
   const { pendingSelection, isLoading, error, onSelectionChange, onSpecChange } = usePluginEditor(props);
```
andreasgerstmayr, line 44: Can we add
```
const { refresh } = useTimeRange();
```
here (in `PluginEditor`), instead of passing it down via props?

Or is this `PluginEditor` component also used in places where we don't have a `<TimeRangeProvider>`?

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
```diff
@@ -112,162 +119,165 @@ export function PanelEditorForm(props: PanelEditorFormProps): ReactElement {
   const watchedName = useWatch({ control: form.control, name: 'panelDefinition.spec.display.name' });
   const watchedDescription = useWatch({ control: form.control, name: 'panelDefinition.spec.display.description' });
   const watchedPluginKind = useWatch({ control: form.control, name: 'panelDefinition.spec.plugin.kind' });
+  const { timeRange } = useTimeRangeParams(initialTimeRange);
```
andreasgerstmayr, line 122: What do you think about the following suggestion?

```suggestion
  const { timeRange } = useTimeRange();
```

this would re-use the same time range which the user choose.

ui/plugin-system/src/components/PluginEditor/PluginEditor.tsx
shahrokni: @andreasgerstmayr You got it right. It is in some places without the provider. Example: datasource and variable editor. Both don't deal with the time range provider

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
andreasgerstmayr: on the other hand, maybe using the default time range, as you have it currently, is better. approving this PR :)

andreasgerstmayr (approved):

PR #3301, opened 2025-09-08
[IGNORE] tune workflows to trigger only when relevant

AntoineThebaud:
# Description

- Don't trigger frontend workflow if only backend or doc files were modified.
- Don't trigger backend workflows if only frontend or doc files were modified.
- Trigger only the `doc` workflow if only doc files were modified.

# Checklist

### DIFF
--- a/.github/workflows/doc.yaml
+++ b/.github/workflows/doc.yaml
@@ -1,0 +1,33 @@
+name: doc
+on:
+  push:
+    branches:
+      - main
+      - release/*
+      - snapshot/*
+    tags:
+      - v*
+  pull_request:
+    paths:
+      - '**.md'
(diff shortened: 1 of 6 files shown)


.github/workflows/doc.yaml
```diff
@@ -0,0 +1,33 @@
+name: cue
```
Gladorme:
```suggestion
name: doc
```

Gladorme (approved):
lgtm

AntoineThebaud:
Unfortunately we can't do that yet, required checks are not compatible with conditional workflows, you end up with PR stuck :( see https://github.com/orgs/community/discussions/13690

Issue #3237, opened 2025-08-27
Handle international characters in UI


noxyu-17:
Currently, upon creating a new object (Project, Dashboard, etc) the name provided by the user is _silently_ passed through `generateMetadataName()` defined `utils/metadata.ts`. Thus, all non-Latin characters are silently converted to underscores.

It may work for some scripts with occasional non-Latin character, but, e.g., Ukrainian words are converted to `"______"` as a whole, with no warning or error. Also, it leads to unexpected behaviour - names of the same length are converted to identical `"____"` strings, so, out of the blue, you cannot create a project named "Русні" if you already have one named "Пизда".

Of course the workaround is to create everything in Latin script and then "rename" to its intended representation, but separate `metadata.name` and `spec.display.name` would be better UX.

Nexucis:
We are using latin character for the name of the dashboard because we had in mind at that time to be able to support CRDs and I think Kubernetes allows only name with latin character.

But if it's not the case anymore (which would be great for Kubernetes), then we could change that as well.

Nexucis:
@jgbernalp @ibakshay do you know if UTF-8 is supported in k8s name ?

ibakshay:
> [@jgbernalp](https://github.com/jgbernalp) [@ibakshay](https://github.com/ibakshay) do you know if UTF-8 is supported in k8s name ?

k8s doesn't support all UTF-8 characters. The k8s only support the names defined in RFC 1123. The rules for k8s name is documented [here](https://kubernetes-docsy-staging.netlify.app/docs/concepts/overview/working-with-objects/names/).

noxyu-17:
The problem is mixing `metadata.name` - which indeed needs to be plain old "DNS name" as per K8S API concepts, no problem here -  with `spec.display.name` - which can be a combination of Unicode emojis if the user wants it to.

It would be much better to have a *UI* thing that allows editing `metadata.name` and `spec.display.name` _separately_.

Nexucis:
yes you are right @noxyu-17 that would be better indeed.

/cc @Gladorme @shahrokni FYI

AntoineThebaud:
About "UI thing that allows editing `metadata.name` and `spec.display.name` separately", this can probably be tackled as part of https://github.com/perses/perses/issues/1343.

PR #3330, opened 2025-09-11
[BUGFIX] Exclude the special charachters form the panel name

### LINKED ISSUE #3285: Panel Name - Validation trap
### What did you do?

1. Choose a invalid name for your panel. For example Panel@
2. Try to save the panel
3. A validation exception is thrown (So far so good)
4. Try to fix the name of the panel using a proper name. For instance, MyPanel
5. Try to save again. You will get the same exception saying you are not allowed to save the Panel@

The updated panel name should be applied and saved

I am not sure, if it is coming from UI. Needs some investigation. (90% UI)

Image

shahrokni:
Closes #3285 

## Description 🖊️ 
A regex has been added to the panel name validation schema to avoid special characters. 
The following Co-Pilot instruction explains why we should avoid the following characters. 

image

## Test 🧪 

Simply try to go against the validation

```typescript
export const panelDisplaySpec: z.ZodSchema<PanelDisplay> = z.object({
  name: z
    .string()
    .min(1, { message: 'Required' })
    .regex(/^[a-zA-Z0-9 _-]+$/, { message: 'Name must only contain alphanumeric characters, underscores, or dashes' }),
  description: z.string().optional(),
});
```

## Special Characters with Reserved Meanings:

- @ (used in email addresses, URLs, or as a delimiter in some systems)
- % (used for URL encoding)
- $ (used in some programming languages or templates for variable substitution)
- & (used in query strings for separating parameters)
- * (used as a wildcard in some systems)
- # (used for fragments in URLs)
- : (used in JSON and URLs)
- / (used in paths or as a delimiter)
- \ (escape character in many languages)
- ? (used in query strings)
- = (used in key-value pairs in query strings)
- + (used for space encoding in URLs)
- <, > (used in HTML and XML)
- {, } (used in JSON and templates)
- [ , ] (used in JSON arrays or other contexts)
- " (used for strings in JSON)
- ' (can cause issues in some contexts, like SQL injection)
- Tab (\t)
- Newline (\n, \r)

## Control Characters:
Any non-printable ASCII characters (e.g., \0, \b, etc.).

## Other Problematic Characters:
! (may have special meanings in some systems)
| (used in some shells or as a delimiter)
~ (used in some systems for home directories or special meanings)

# Checklist

## UI Changes

### DIFF
--- a/ui/core/src/schema/panel.ts
+++ b/ui/core/src/schema/panel.ts
@@ -16,7 +16,10 @@ import { Link, PanelDefinition, PanelDisplay, PanelEditorValues, PanelSpec, Quer
 import { PluginSchema, pluginSchema } from './plugin';
 
 export const panelDisplaySpec: z.ZodSchema<PanelDisplay> = z.object({
-  name: z.string().min(1, { message: 'Required' }),
+  name: z
+    .string()
+    .min(1, { message: 'Required' })
+    .regex(/^[a-zA-Z0-9 _-]+$/, { message: 'Name must only contain alphanumeric characters, underscores, or dashes' }),
   description: z.string().optional(),
 });
 
--- a/ui/plugin-system/src/context/ValidationProvider.tsx
+++ b/ui/plugin-system/src/context/ValidationProvider.tsx
@@ -39,7 +39,7 @@ export const ValidationSchemasContext = createContext<ValidationSchemas | undefi
 
 export function useValidationSchemas(): ValidationSchemas {
   const ctx = useContext(ValidationSchemasContext);
-  if (ctx === undefined) {
+  if (!ctx) {
     throw new Error('No ValidationSchemasContext found. Did you forget a Provider?');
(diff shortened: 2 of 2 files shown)


ui/core/src/schema/panel.ts
```diff
@@ -16,7 +16,10 @@ import { Link, PanelDefinition, PanelDisplay, PanelEditorValues, PanelSpec, Quer
 import { PluginSchema, pluginSchema } from './plugin';
 
 export const panelDisplaySpec: z.ZodSchema<PanelDisplay> = z.object({
-  name: z.string().min(1, { message: 'Required' }),
+  name: z
```
Nexucis, line 19: actually like raised in https://github.com/perses/perses/issues/3237 the `display.name` should allow using all characters.

We should make the difference between `metadata.name` and `display.name` in the UI as the constraints are not the same for each.

PR #3332, opened 2025-09-11
[BUGFIX] Simplify the panels key management

shahrokni:
Closes https://github.com/perses/perses/issues/3285

## Description 🖊️ 
This change decouples the display name of a panel from its so-called metadata key. 
With this fix you should be able to use characters like !@#$%^&*()_+=-/.,? in the display name. 

## What does this change do ❓ 

This change simply uses the randomUUID from browser crypto API to generate a  UUID V4
This value is used to maintain an object of panels. 

## What is Random UUID

crypto.randomUUID() generates a UUID v4, which is statistically unique. UUIDs (Universally Unique Identifiers) are designed to be globally unique, and the randomness in UUID v4 ensures that the likelihood of a collision is extremely low.

## Test 🧪 

1. You should be able to create a panel with a display name which may include !@#$%^&*()_+=-/.,:;|}{[]\
2. You should be able to edit your panels with no issue
3. You should be able to clone your panels with no issue

image

## Final note ℹ️ 

Still it is not clear why such a complex login was used to generate a key for the panels object, while it could be any unique generated id. I talked to some colleagues, though no one knew if there was any valid reason behind such a complex logic. 

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/context/DashboardProvider/duplicate-panel-slice.ts
+++ b/ui/dashboards/src/context/DashboardProvider/duplicate-panel-slice.ts
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { StateCreator } from 'zustand';
-import { getValidPanelKey, insertPanelInLayout, UnpositionedPanelGroupItemLayout } from '../../utils/panelUtils';
+import { insertPanelInLayout, UnpositionedPanelGroupItemLayout } from '../../utils/panelUtils';
 import { generateId, Middleware } from './common';
 import { PanelGroupSlice, PanelGroupItemId } from './panel-group-slice';
 import { PanelSlice } from './panel-slice';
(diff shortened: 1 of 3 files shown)


Nexucis:
noice, that should fix also https://github.com/perses/perses/issues/3237

shahrokni:
> noice, that should fix also #3237

For panels, yes, but that issue also mentions projects and dashboars, if I am not mistaken

Nexucis:
ah yes indeed you are right

Nexucis:
did you try to edit an existing dashboard and an existing panel to see what is happening ?

shahrokni:
> did you try to edit an existing dashboard and an existing panel to see what is happening ?

Ye worked just fine, with their old generated id. If you have any specific scenario, I ll test it
The old items which currently exist, keep their own id.

PR #3346, opened 2025-09-15
[FEATURE] Add Panel Editor Context

shahrokni:
Relates to https://github.com/perses/perses/issues/3209

# Description 🖊️ 

This PR helps us to **bridge the gap between the panel preview and pane spec editor**. This is important, because some spec editor calculations are dependent upon preview information. 

## Example 

To calculate the built-in Prometheus variables like $__interval or $__rate_interval we need to have `Suggested Step MS` which is calculated based on the panel's `width` and `absolute time range.` 

## Current implementation and the issue of drilling the props down to the plugin

As you can see in the following diagram, the information and its consumer  are located in two parallel sub-trees. 
This means to gain access to the panel preview width, wen need to lift the prop up and drill it down to the sub-tree, which is not ideal.

image

## Solution 

As it has been discussed with @jgbernalp to bridge the gap, it makes sense to have a dedicated provider to share some information between parallel sub-trees of the Edit Panel. This idea resolves the issue of drilling the props down. 

image

## Quick Test of the idea

image

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/components/PanelDrawer/PanelPreview.tsx
+++ b/ui/dashboards/src/components/PanelDrawer/PanelPreview.tsx
@@ -11,11 +11,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { ReactElement, useRef } from 'react';
+import { ReactElement, useContext, useRef } from 'react';
 import { Box } from '@mui/material';
 import { DataQueriesProvider, usePlugin, useSuggestedStepMs } from '@perses-dev/plugin-system';
 import { PanelEditorValues } from '@perses-dev/core';
 import { Panel } from '../Panel';
+import { PanelEditorContext } from '../../context';
 
 const PANEL_PREVIEW_HEIGHT = 300;
 const PANEL_PREVIEW_DEFAULT_WIDTH = 840;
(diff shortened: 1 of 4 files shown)


ui/dashboards/src/components/PanelDrawer/PanelPreview.tsx
```diff
@@ -26,6 +27,10 @@ export function PanelPreview({ panelDefinition }: Pick<PanelEditorValues, 'panel
   if (boxRef.current !== null) {
     width = boxRef.current.getBoundingClientRect().width;
   }
+
+  const { preview } = useContext(PanelEditorContext) ?? {};
+  preview?.setPreviewPanelWidth?.(width);
+
```
jgbernalp: We should probably set the width when the boxRef is present. Also if the provider is not defined the initial state of the context can be undefined, so no destructure possible. Just coding defensively for contexts in which the components are used directly and not from the Perses UI
```suggestion
 const panelEditorContext = useContext(PanelEditorContext);
 
if (boxRef.current !== null) {
width = boxRef.current.getBoundingClientRect().width;
panelEditorContext?.preview?.setPreviewPanelWidth?.(width);
}

```

jgbernalp (approved):

PR #3353, opened 2025-09-15
[ignore] set a deadline for dropping react 17 support

Nexucis:
as agreed during the Perses sync meeting, we are setting a deadline for dropping react 17 support.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -2,7 +2,8 @@
 
 ## More data sources
 
-With the new plugin system now available, we have unblocked the possibility to support other data sources besides Prometheus
+With the new plugin system now available, we have unblocked the possibility to support other data sources besides
+Prometheus
 and Tempo.
 
 1. We are working on supporting Logs Datasource, starting with OpenSearch and Loki.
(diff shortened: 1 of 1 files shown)


jgbernalp (approved):

Gladorme (approved):

ROADMAP.md
```diff
@@ -2,7 +2,8 @@
 
 ## More data sources
 
-With the new plugin system now available, we have unblocked the possibility to support other data sources besides Prometheus
+With the new plugin system now available, we have unblocked the possibility to support other data sources besides
+Prometheus
```
AntoineThebaud, line 6: This section can be further updated (Loki is now supported + Pyroscope), maybe starting with something like

```suggestion
Since the release v0.51 and its revamped plugin architecture, we have unblocked the possibility to support other data sources. New plugins to handle Loki for logs & Pyroscope for profiles have been released in v0.52 and we are still considering other technologies like Opensearch & Jaeger

(also a paragraph to mention/update the status around Clickhouse)
```

Nexucis: ok will do in another PR then

Nexucis: done in https://github.com/perses/perses/pull/3354

PR #3358, opened 2025-09-17
[ENHANCEMENT] Improve percli migrate command to support default datasource

ibakshay:
Enhance the migration command to utilize a default datasource when specified with `--use-default-datasource` flag.

This change ensures that the migration process can clean datasource references in migrated dashboards.

## Usage Example

`percli migrate -f ./grafana.json --plugin.path ./plugins --use-default-datasource=true`

## Notes
No breaking changes; flag is opt-in. Behavior unchanged when flag is omitted or false

### DIFF
--- a/internal/cli/cmd/migrate/migrate.go
+++ b/internal/cli/cmd/migrate/migrate.go
@@ -73,16 +73,17 @@ type option struct {
 	persesCMD.Option
 	opt.FileOption
 	opt.OutputOption
-	writer          io.Writer
-	errWriter       io.Writer
-	rowInput        []string
-	input           map[string]string
-	project         string
-	pluginPath      string
-	online          bool
-	mig             migrate.Migration
-	apiClient       api.ClientInterface
-	migrationFormat migrationFormat
+	writer               io.Writer
+	errWriter            io.Writer
+	rowInput             []string
+	input                map[string]string
+	project              string
+	pluginPath           string
+	online               bool
+	useDefaultDatasource bool
+	mig                  migrate.Migration
+	apiClient            api.ClientInterface
+	migrationFormat      migrationFormat
 }
 
 func (o *option) Complete(args []string) error {
@@ -135,6 +136,10 @@ func (o *option) Validate() error {
 	if o.migrationFormat != nativeFormat && o.migrationFormat != customResourceFormat && o.migrationFormat != customResourceShortFormat {
 		return fmt.Errorf("invalid value for flag --format: %s", o.migrationFormat)
 	}
+
+	if !o.online && o.mig == nil {
+		return fmt.Errorf("offline migration requires --plugin.path to be specified, or use --online for server-side migration")
+	}
 	return nil
 }
 
@@ -154,6 +159,22 @@ func (o *option) Execute() error {
 		return err
 	}
 	persesDashboard.Metadata.Project = o.project
+
+	// Apply datasource cleaning if the flag is set
+	if o.useDefaultDatasource {
+		// Clean datasource references on all queries in all panels
+		for _, panel := range persesDashboard.Spec.Panels {
+			for _, query := range panel.Spec.Queries {
+				if pluginSpec, ok := query.Spec.Plugin.Spec.(map[string]any); ok {
+					if datasourceRef, ok := pluginSpec["datasource"].(map[string]any); ok {
+						// Remove the "name" property to use default datasource
+						delete(datasourceRef, "name")
+					}
+				}
+			}
+		}
+	}
+
 	if o.migrationFormat == customResourceFormat || o.migrationFormat == customResourceShortFormat {
 		customResource := createCustomResource(persesDashboard)
 		return output.Handle(o.writer, o.Output, customResource)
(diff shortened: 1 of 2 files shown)


ibakshay:
The e2etest failure seems to be unrelated to my changes. 

```yaml
time="2025-09-17T12:07:30Z" level=error msg="Failed to sync user in database." error="Error 1062 (23000): Duplicate entry 'john.doe' for key 'PRIMARY'" provider=bb1659e1-a9c1-4c01-a794-1647d20fb0c1
time="2025-09-17T12:07:30Z" level=info method=POST status=400 uri=/api/auth/providers/oauth/bb1659e1-a9c1-4c01-a794-1647d20fb0c1/token
time="2025-09-17T12:07:30Z" level=info method=POST status=200 uri=/api/auth/providers/oauth/bb1659e1-a9c1-4c01-a794-1647d20fb0c1/token
=== RUN   TestLoginOAuth/Nominal_Client_Credentials_flow
time="2025-09-17T12:07:30Z" level=info method=GET status=200 uri=/api/config
time="2025-09-17T12:07:30Z" level=info method=POST status=200 uri=/api/auth/providers/oauth/bb1659e1-a9c1-4c01-a794-1647d20fb0c1/token
=== NAME  TestLoginOAuth
    server.go:98: ErrorCode: 404, key: john.doe
--- FAIL: TestLoginOAuth (7.68s)
    --- PASS: TestLoginOAuth/Nominal_Device_Code_flow (5.02s)
    --- PASS: TestLoginOAuth/Nominal_Client_Credentials_flow (0.01s)
    ```

File: internal/cli/cmd/migrate/migrate.go [hunk 1]
```diff
@@ -154,6 +159,22 @@ func (o *option) Execute() error {
 		return err
 	}
 	persesDashboard.Metadata.Project = o.project
+
+	// Apply datasource cleaning if the flag is set
+	if o.useDefaultDatasource {
```
Nexucis: I think this should go in the migration logic directly.

Then you will have to expose a query parameter in the migrate endpoint for the online migration. For the offline migration you will have to call the migrate method with this parameter

pkg/client/api/client.go
```diff
@@ -51,17 +53,33 @@ func (c *client) V1() v1.ClientInterface {
...
+	if useDefaultDatasource {
+		request = request.Query(&migrateQuery{values: map[string][]string{"default-datasource": {"true"}}})
+	}
+
+	err := request.Do().Object(result)
 	return result, err
 }
 
+type migrateQuery struct {
+	values map[string][]string
```
Nexucis: you could do that instead :

```go

func (c *client) Migrate(body *api.Migrate, useDefaultDatasource bool) (*modelV1.Dashboard, error) {
	result := &modelV1.Dashboard{}
	err := c.restClient.Post().
		APIVersion("").
		Resource("migrate").
		Body(body).
        Query(&migrateQuery{defaultDatasource: useDefaultDatasource}).
		Do().
		Object(result)
	return result, err
}

type migrateQuery struct {
  defaultDatasource bool
}

func (q *migrateQuery) GetValues() url.Values {
  return url.Values{
    "default-datasource": []string{strconv.FormatBool(q.defaultDatasource)}
  }
}
```

ibakshay: thank you very much for the suggestion! I have modified in 7f9e4a2ba044d25265cbd80772498d59886ea881

pkg/client/api/client.go
```diff
@@ -51,17 +54,33 @@ func (c *client) V1() v1.ClientInterface {
...
 
-func (c *client) Migrate(body *api.Migrate) (*modelV1.Dashboard, error) {
+func (c *client) Migrate(body *api.Migrate, useDefaultDatasource bool) (*modelV1.Dashboard, error) {
 	result := &modelV1.Dashboard{}
-	err := c.restClient.Post().
+	request := c.restClient.Post().
 		APIVersion("").
 		Resource("migrate").
-		Body(body).
-		Do().
```
Nexucis, line 60: can't you just add `Query()` directly in the chain like suggested ?

ibakshay: Sorry, I have overlooked your suggestion. 
Here we go 1a1e65204d54818e5682220850205a2aa49879f6 :)

Nexucis:
sorry I forgot also how the migration is working on server side and actually there is already a body that contains some parameter.
I think to be coherent with the current code, we should add this new parameter to the body actually instead of adding it as a query parameter

Nexucis:
sorry for misleading here

ibakshay:
No problem. The query parameter should be removed and  the `default-datasource` property should be added to this struct right? 
https://github.com/perses/perses/blob/1a1e65204d54818e5682220850205a2aa49879f6/pkg/model/api/migrate.go#L21-L24

Nexucis:
yes exactly. And in the body you can even rename the parameter by `UseDefaultDatasource` that would be more explanatory

ibakshay:
> yes exactly. And in the body you can even rename the parameter by `UseDefaultDatasource` that would be more explanatory

Done in 162179f83f82c65e80b06b4b1053f321a9ca33c2 :)

Nexucis (approved):
awesome, thank you @ibakshay !

ibakshay:
> awesome, thank you @ibakshay !

Thank you for reviewing! 
I will create a follow-up PR in the next coming days with a checkbox to select the `defaultDatasource` in the UI here. 
image
WDYT?

PR #3372, opened 2025-09-18
[BUGFIX] use globalThis object over global

### LINKED ISSUE #3361: Error using packages with vite: "global is not defined"
### What did you do?

When building with vite and making use of any packages that call `fetch` or `fetchJson` from the `@perses-dev/core` package, an error is thrown that prevents execution:

```
ReferenceError: global is not defined
    at fetch (fetch.ts:18:20)
    at fetchResults (prometheus-client.ts:225:26)
    at fetchWithPost (prometheus-client.ts:183:10)
    at rangeQuery (prometheus-client.ts:88:10)
    at Object.rangeQuery (prometheus-datasource.tsx:52:38)
    at Object.getTimeSeriesData (get-time-series-data.ts:124:31)
    at async queryFn (time-series-queries.ts:145:24)
```

This is because the `ui/core/utils/fetch.ts` file references the `global` object which is only defined in node, webpack, or webpack-compatible build tools. If possible, I'd like to propose changing this so that this file switches to the [`globalThis` object](https://tc39.es/ecma262/#sec-global-object) instead of the node/webpack-only `global` object.

I can submit a PR to resolve this issue if interested.

https://github.com/perses/perses/blob/fc44f896292fe64e02c78886da1509af19cc71b7/ui/core/src/utils/fetch.ts#L18

### Perses version

0.52.0

### Temporary Fix

As a temporary solution, it is possible to redefine `global`, either through the vite [`define`](https://vite.dev/config/shared-options.html#define) configuration, or through a shim (below) however these are not preferred solutions as global could be redefined by any other library.

```ts
global ??= globalThis ?? window
```

lachieh:
# Description

Fixes #3361

Uses `globalThis` instead of `global` to ensure compatibility across different JavaScript environments. `global` is node specific, while `globalThis` is a standard that works in both Node.js and browser environments. The globalThis object is already in use in other parts of the codebase ([here](https://github.com/perses/perses/blob/9b21ce5364088dc58464c93ef62cf1875ce197b5/ui/dashboards/src/context/DashboardProvider/common.ts#L32-L36) and [here](https://github.com/perses/perses/blob/9b21ce5364088dc58464c93ef62cf1875ce197b5/ui/components/src/utils/component-ids.ts#L25-L27)) and the tests pass with this change.

# Screenshots

N/A

# Checklist

## UI Changes

- [x] N/A ~~Changes that impact the UI include screenshots and/or screencasts of the relevant changes.~~

### DIFF
--- a/ui/app/src/utils/browser-storage.ts
+++ b/ui/app/src/utils/browser-storage.ts
@@ -21,7 +21,7 @@ type StorageTuple<T> = [T, (next: T) => void];
  * storage does not have any data yet.
  */
 export function useLocalStorage<T>(key: string, initialValue: T): StorageTuple<T> {
-  const { value, setValueAndStore } = useStorage(global.localStorage, key, initialValue);
+  const { value, setValueAndStore } = useStorage(window.localStorage, key, initialValue);
   return [value, setValueAndStore];
 }
 
--- a/ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.test.tsx
+++ b/ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.test.tsx
@@ -18,7 +18,7 @@ import { QueryEditorContainer } from './QueryEditorContainer';
 
 describe('QueryEditorContainer', () => {
   beforeEach(() => {
-    global.fetch = jest.fn(() =>
+    globalThis.fetch = jest.fn(() =>
       Promise.resolve({
         json: () => Promise.resolve({ success: true }),
       })
--- a/ui/app/src/model/fetch.ts
+++ b/ui/app/src/model/fetch.ts
@@ -22,8 +22,8 @@ function deleteCookie(name: string): void {
 }
 
 export function enableRefreshFetch(): void {
-  global.fetch = new Proxy(global.fetch, {
-    apply: async function (target, that, args: Parameters<typeof global.fetch>): Promise<Response> {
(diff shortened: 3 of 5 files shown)


ui/plugin-system/src/components/MultiQueryEditor/QueryEditorContainer.test.tsx
```diff
@@ -18,7 +18,7 @@ import { QueryEditorContainer } from './QueryEditorContainer';
 
 describe('QueryEditorContainer', () => {
   beforeEach(() => {
-    global.fetch = jest.fn(() =>
+    globalThis.fetch = jest.fn(() =>
```
lachieh, line 21: I confirmed that the tests pass with and without this change

ui/app/src/utils/browser-storage.ts
```diff
@@ -21,7 +21,7 @@ type StorageTuple<T> = [T, (next: T) => void];
  * storage does not have any data yet.
  */
 export function useLocalStorage<T>(key: string, initialValue: T): StorageTuple<T> {
-  const { value, setValueAndStore } = useStorage(global.localStorage, key, initialValue);
+  const { value, setValueAndStore } = useStorage(window.localStorage, key, initialValue);
```
lachieh, line 24: This is the only instance of `global` that uses a browser-only API so I've switched this to use `window` instead.

PR #3373, opened 2025-09-18
[ENHANCEMENT] Allow UI to select default datasource during migration

ibakshay:
UI implementation of https://github.com/perses/perses/pull/3358
- Added a checkbox for using default datasource in Perses

# Description

# Screenshots

https://github.com/user-attachments/assets/38e4d7db-5ce7-4a64-9d0b-87ce11320b6f

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/model/migrate-client.ts
+++ b/ui/app/src/model/migrate-client.ts
@@ -21,6 +21,7 @@ const resource = 'migrate';
 export interface MigrateBodyRequest {
   input?: Record<string, string>;
   grafanaDashboard: Record<string, unknown>;
+  useDefaultDatasource?: boolean;
 }
 
 export function useMigrate(): UseMutationResult<DashboardResource, StatusError, MigrateBodyRequest> {
@@ -33,7 +34,7 @@ export function useMigrate(): UseMutationResult<DashboardResource, StatusError, 
         headers: HTTPHeader,
         body: `{"input":${body.input ? JSON.stringify(body.input) : '{}'}, "grafanaDashboard": ${JSON.stringify(
           body.grafanaDashboard
-        )}}`,
+        )}, "useDefaultDatasource": ${body.useDefaultDatasource ? 'true' : 'false'}}`,
       });
     },
   });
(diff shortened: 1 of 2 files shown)


ui/app/src/model/migrate-client.ts
```diff
@@ -33,7 +34,7 @@ export function useMigrate(): UseMutationResult<DashboardResource, StatusError,
         headers: HTTPHeader,
         body: `{"input":${body.input ? JSON.stringify(body.input) : '{}'}, "grafanaDashboard": ${JSON.stringify(
           body.grafanaDashboard
-        )}}`,
+        )}, "useDefaultDatasource": ${body.useDefaultDatasource ? 'true' : 'false'}}`,
```
shahrokni: The body. useDefaultDatasource is an optional Boolean itself (am I right?), so maybe you could simply do something like this? 
Would that make sense? 


```typescript
"useDefaultDatasource": !!body.useDefaultDatasource
```

ui/app/src/views/import/GrafanaFlow.tsx
```diff
@@ -48,6 +58,7 @@ function GrafanaFlow({ dashboard }: GrafanaFlowProps): ReactElement {
   const { exceptionSnackbar } = useSnackbar();
   const [projectName, setProjectName] = useState<string>('');
   const [grafanaInput, setGrafanaInput] = useState<Record<string, string>>({});
+  const [useDefaultDatasource, setUseDefaultDatasource] = useState<boolean>(false);
   const { data, isLoading, error } = useProjectList();
```
shahrokni: useState with an initial value would not need to have the explicit type. 
So, the type is deduced based on the given initial value. 

Could simply be 

```typescript
const [useDefaultDatasource, setUseDefaultDatasource] = useState(false);
```

shahrokni:
Apart from those tiny comments, it looks beautiful ❤️

ui/app/src/model/migrate-client.ts
ibakshay: fixed in 63c1dac4f8afd149b7d095d9690bd626c52a9a0b

ui/app/src/views/import/GrafanaFlow.tsx
ibakshay: thank you. fixed in 63c1dac4f8afd149b7d095d9690bd626c52a9a0b

PR #3377, opened 2025-09-19
[BREAKINGCHANGE] Improve support for reverse proxy configuration

### LINKED ISSUE #2579: api_prefix: Not applied when front is sending data to /api/v1/views
### What did you do?

1. Similar to #2575 I have put Perses behind a reverse proxy that redirect every request starting with a prefix. 
2. Configure the api_prefix in the Perses application.
3. Access a dashboard through the reverse proxy in a browser.
4. Observe the network requests in the browser's developer tools.

All requests are performed correctly

The dashboards are loaded correctly and every data going to /proxy are correctly sent and processed. 

The only thing that remains is when the frontend is trying to send the time the dashboard has taken to load. 
The frontend at this moment is sending the data to the endpoint `api/v1/views`

That's a remaining issue raised during the bugfix #2576 

### System information

all system.

### Perses version

v0.50.0

### Perses configuration file

```yaml
N/A
```

### Logs

```text
N/A
```

lachieh:
# Description

This PR adds some necessary changes to enable reverse proxy support, allowing the application to work correctly when deployed behind a reverse proxy with a custom base path.

There is still an issue with the plugin files having been compiled with a specific base path (`/`) but this will be tackled in a future PR.

Closes #2579

## Summary of Changes

**Plugin System Updates:**
- **`remotePluginLoader.ts`**: Modified to accept an options object in addition to a single string so that the baseUrl is handled correctly while maintaining the original behavior of the original implementation.
- **Multiple React components**: Updated to pass `getBasePathName()` to remotePluginLoader for consistent base path handling

**API Request Updates:**
- **`UsageMetricsProvider.tsx`**: Added `baseUrl` parameter to ensure API requests use the correct base path when behind a proxy

# Screenshots

N/A

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/model/config-client.ts
+++ b/ui/app/src/model/config-client.ts
@@ -215,6 +215,6 @@ export function useConfig(options?: ConfigOptions): UseQueryResult<ConfigModel, 
 }
 
 export function fetchConfig(): Promise<ConfigModel> {
-  const url = buildURL({ resource: resource, apiPrefix: `${PERSES_APP_CONFIG.api_prefix}/api` });
+  const url = buildURL({ resource: resource, apiUrl: '/api', apiPrefix: PERSES_APP_CONFIG.api_prefix });
   return fetchJson<ConfigModel>(url);
 }
--- a/ui/app/src/model/config-client.ts
+++ b/ui/app/src/model/config-client.ts
@@ -190,6 +190,7 @@ export interface DatasourceConfig {
 }
 
 export interface ConfigModel {
+  api_prefix?: string;
   security: SecurityConfig;
   database: Database;
   schemas: ConfigSchemasModel;
--- a/ui/app/src/model/config-client.ts
+++ b/ui/app/src/model/config-client.ts
@@ -14,6 +14,7 @@
 import { useQuery, UseQueryOptions, UseQueryResult } from '@tanstack/react-query';
 import { DashboardSelector, DurationString, fetchJson, Permission, StatusError } from '@perses-dev/core';
 import { Duration } from 'date-fns';
+import { PERSES_APP_CONFIG } from '../config';
 import buildURL from './url-builder';
 
 const resource = 'config';
@@ -214,6 +215,6 @@ export function useConfig(options?: ConfigOptions): UseQueryResult<ConfigModel, 
 }
 
 export function fetchConfig(): Promise<ConfigModel> {
-  const url = buildURL({ resource: resource, apiPrefix: '/api' });
+  const url = buildURL({ resource: resource, apiPrefix: `${PERSES_APP_CONFIG.api_prefix}/api` });
   return fetchJson<ConfigModel>(url);
 }
--- a/ui/plugin-system/src/remote/remotePluginLoader.test.ts
+++ b/ui/plugin-system/src/remote/remotePluginLoader.test.ts
@@ -49,36 +49,27 @@ describe('remotePluginLoader', () => {
   });
 
   describe('getInstalledPlugins', () => {
-    it('should fetch plugins from correct endpoint without baseURL', async () => {
-      const mockResponse = { json: jest.fn().mockResolvedValue([MOCK_VALID_PLUGIN_MODULE_RESOURCE]) };
-      mockFetch.mockResolvedValue(mockResponse as unknown as Response);
-
-      const loader = remotePluginLoader();
-      await loader.getInstalledPlugins();
-
-      expect(mockFetch).toHaveBeenCalledWith('/api/v1/plugins');
-    });
-
-    it('should fetch plugins from correct endpoint with baseURL', async () => {
-      const baseURL = 'https://example.com';
-      const mockResponse = { json: jest.fn().mockResolvedValue([MOCK_VALID_PLUGIN_MODULE_RESOURCE]) };
-      mockFetch.mockResolvedValue(mockResponse as unknown as Response);
-
-      const loader = remotePluginLoader(baseURL);
-      await loader.getInstalledPlugins();
-
-      expect(mockFetch).toHaveBeenCalledWith(`${baseURL}/api/v1/plugins`);
-    });
-
-    it('should return valid plugin modules when fetch succeeds', async () => {
+    it('should fetch plugins from correct endpoint without options', async () => {
       const mockResponse = { json: jest.fn().mockResolvedValue([MOCK_VALID_PLUGIN_MODULE_RESOURCE]) };
       mockFetch.mockResolvedValue(mockResponse as unknown as Response);
 
       const loader = remotePluginLoader();
       const result = await loader.getInstalledPlugins();
 
+      expect(mockFetch).toHaveBeenCalledWith('/api/v1/plugins');
       expect(result).toEqual([MOCK_VALID_PLUGIN_MODULE_RESOURCE]);
       expect(mockConsoleError).not.toHaveBeenCalled();
+    });
+
+    it('should fetch plugins from correct endpoint with apiPrefix', async () => {
+      const apiPrefix = 'https://example.com';
+      const mockResponse = { json: jest.fn().mockResolvedValue([MOCK_VALID_PLUGIN_MODULE_RESOURCE]) };
+      mockFetch.mockResolvedValue(mockResponse as unknown as Response);
+
+      const loader = remotePluginLoader({ apiPrefix });
+      await loader.getInstalledPlugins();
+
+      expect(mockFetch).toHaveBeenCalledWith(`${apiPrefix}/api/v1/plugins`);
     });
 
     it('should filter out invalid plugin modules and return only valid ones', async () => {
@@ -137,40 +128,38 @@ describe('remotePluginLoader', () => {
       const loader = remotePluginLoader();
       const result = await loader.importPluginModule(MOCK_VALID_PLUGIN_MODULE_RESOURCE);
 
-      expect(mockLoadPlugin).toHaveBeenCalledWith('test-module', 'testPlugin', undefined);
+      expect(mockLoadPlugin).toHaveBeenCalledWith('test-module', 'testPlugin', '/plugins');
       expect(result).toEqual(MOCK_REMOTE_PLUGIN_MODULE);
       expect(mockConsoleError).not.toHaveBeenCalled();
     });
 
-    it('should pass baseURL to loadPlugin when provided', async () => {
-      const baseURL = 'https://example.com';
-      mockLoadPlugin.mockResolvedValue(MOCK_REMOTE_PLUGIN_MODULE);
-
-      const loader = remotePluginLoader(baseURL);
-      const result = await loader.importPluginModule(MOCK_VALID_PLUGIN_MODULE_RESOURCE);
-
-      expect(mockLoadPlugin).toHaveBeenCalledWith('test-module', 'testPlugin', baseURL);
-      expect(result).toEqual(MOCK_REMOTE_PLUGIN_MODULE);
-    });
-
-    it('should handle different baseURL formats', async () => {
-      const testCases = [
-        { baseURL: 'https://example.com', expected: 'https://example.com/api/v1/plugins' },
-        { baseURL: 'https://example.com/', expected: 'https://example.com//api/v1/plugins' },
-        { baseURL: 'http://localhost:3000', expected: 'http://localhost:3000/api/v1/plugins' },
-        { baseURL: '', expected: '/api/v1/plugins' },
+    it('should handle options object variations', async () => {
+      const testCases: Array<{
+        options?: { apiPrefix?: string; baseUrl?: string };
+        expected: string;
+      }> = [
+        // object with baseUrl only
+        { options: { baseUrl: 'https://example.com' }, expected: 'https://example.com/plugins' },
+        { options: { baseUrl: 'https://example.com/' }, expected: 'https://example.com//plugins' },
+        { options: { baseUrl: 'http://localhost:3000' }, expected: 'http://localhost:3000/plugins' },
+        { options: { baseUrl: '' }, expected: '/plugins' },
+        // object with nothing or undefined
+        { options: {}, expected: '/plugins' },
+        { options: undefined, expected: '/plugins' },
       ];
 
+      mockLoadPlugin.mockResolvedValue(MOCK_REMOTE_PLUGIN_MODULE);
+
       for (const testCase of testCases) {
-        const mockResponse = { json: jest.fn().mockResolvedValue([]) };
-        mockFetch.mockResolvedValue(mockResponse as unknown as Response);
+        const loader = remotePluginLoader(testCase.options);
+        const result = await loader.importPluginModule(MOCK_VALID_PLUGIN_MODULE_RESOURCE);
 
-        const loader = remotePluginLoader(testCase.baseURL || undefined);
-        await loader.getInstalledPlugins();
-
-        expect(mockFetch).toHaveBeenCalledWith(testCase.expected);
-        mockFetch.mockClear();
+        expect(mockLoadPlugin).toHaveBeenCalledWith('test-module', 'testPlugin', testCase.expected);
+        expect(result).toEqual(MOCK_REMOTE_PLUGIN_MODULE);
+        expect(mockConsoleError).not.toHaveBeenCalled();
       }
+
+      mockLoadPlugin.mockClear();
     });
 
     it('should handle multiple plugins in a module', async () => {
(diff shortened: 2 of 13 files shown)


ui/endpoint.go
```diff
@@ -97,6 +119,15 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 		// In that case, we need to read the requested files from the file system.
 		// The First thing to do is to replace the URL path with the local path of the plugin.
 		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
+		// If the request is for the mf-manifest.json file, then we need to replace the value of "publicPath" with the
+		// proper API prefix so the React app is able to find the static files of the plugin.
```
jgbernalp: @Nexucis I remember we talked about this replacement before. ~IIUC this endpoint is only for when using the plugins in dev "mode", is that correct?~

jgbernalp:
A couple of things:

## Alternative setting of the public path from the frontend
We could replace the values when serving them from the backend, but the RsBuild plugin for module federation allows to pass an inline function to define the public path. This could be an alternate solution in which Hosts, such as the Perses UI, set the value so the Perses plugin runtime can consume it at runtime:


```typescript
// Plugin rs build configuraion

...

export default defineConfig({
  ...
  plugins: [
    pluginReact(),
    pluginModuleFederation({
      ...
      getPublicPath: 'function() { return window.PERSES_PLUGIN_ASSETS_PATH || ""; }', // <<-
    }),
  ],
});
```

## Configuration and initialization of the PERSES_PLUGIN_ASSETS_PATH
Adding `pluginAssetsPath` to the configuration will let clients load Perses plugins into environments that embed Perses as well as through reverse proxies. If `prefixAPI` is already set, this value can default to the prefixAPI value.

## Sharing the value through the app
If we set this value in the configuration, we can add it to the `ConfigContextProvider` this will allow the component tree and other providers to use it. If not set it should use the one derived from the `window.location.pathname` for backwards compatibility.

WDYT?

lachieh:
Yeah, that's a great solution. I was looking for something that didn't require updating all the plugins, but only ended up with brittle solutions like changes to the go backend like this one.

I'll remove the go backend changes and leave the react ones since they are still required for the feature.

lachieh:
Actually, there's a small issue with `remotePluginLoader`. The baseURL parameter is used to construct the apiUrl for plugins, as well as the whole path to the plugins folder once it is passed to `loadPlugin`:
1. Constructing the API path to plugins:
   https://github.com/perses/perses/blob/91784be3da2b8ceda79e040d1b1b89c3926bcded/ui/plugin-system/src/remote/remotePluginLoader.ts#L47

2. As the whole path to plugins:
   https://github.com/perses/perses/blob/91784be3da2b8ceda79e040d1b1b89c3926bcded/ui/plugin-system/src/remote/remotePluginLoader.ts#L71
The issue is that when `baseUrl` is provided, the `/plugins` prefix gets removed making it not a baseUrl in the same definition in both places.
   https://github.com/perses/perses/blob/91784be3da2b8ceda79e040d1b1b89c3926bcded/ui/plugin-system/src/remote/PluginRuntime.tsx#L207

I've updated the PR with a solution that is backwards compatible, though I could also do a version that adds the breaking behavior.

ui/plugin-system/src/remote/remotePluginLoader.ts
```diff
@@ -41,10 +41,96 @@ const isPluginModuleResource = (pluginModule: unknown): pluginModule is PluginMo
...
+ * ```
+ */
+export function remotePluginLoader(baseUrl: string): PluginLoader;
+export function remotePluginLoader(options?: RemotePluginLoaderOptions): PluginLoader {
+  const { apiUrl, pluginsUrl } = paramToOptions(options);
+
   return {
     getInstalledPlugins: async (): Promise<PluginModuleResource[]> => {
-      const pluginsResponse = await fetch(`${baseURL ? baseURL : ''}/api/v1/plugins`);
+      const pluginsResponse = await fetch(apiUrl);
```
lachieh: It might actually be worth pulling the [`buildURL` function](https://github.com/lachieh/perses/blob/7ca38f394615ef0e28659464f022c552fa09ccc1/ui/app/src/model/url-builder.ts) from `app` into `plugin-system` as a helper function so it can be used here.

ui/plugin-system/src/remote/remotePluginLoader.test.ts
```diff
@@ -137,7 +137,7 @@ describe('remotePluginLoader', () => {
       const loader = remotePluginLoader();
       const result = await loader.importPluginModule(MOCK_VALID_PLUGIN_MODULE_RESOURCE);
 
-      expect(mockLoadPlugin).toHaveBeenCalledWith('test-module', 'testPlugin', undefined);
+      expect(mockLoadPlugin).toHaveBeenCalledWith('test-module', 'testPlugin', '/plugins');
```
lachieh, line 131: Passing the `'/plugins'` value into loadPlugin is the same behavior as `undefined` since the it replaces the same value:
https://github.com/perses/perses/blob/91784be3da2b8ceda79e040d1b1b89c3926bcded/ui/plugin-system/src/remote/PluginRuntime.tsx#L207

ui/plugin-system/src/remote/remotePluginLoader.ts
jgbernalp: yes indeed.

jgbernalp:
Thnx for the update @lachieh , I think a breaking change will make the function API more readable about the options. For this case we might have different values for base url for the plugins assets or prefix for the API, for example:

```typescript
type RemotePluginLoaderOptions = {
   /**
   * The API path for fetching available Perses plugins. It's used to construct the full URL to the `/api/v1/plugins` endpoint.
   * @default ''
   **/
   apiPrefix?: string;
   /**
   * The base URL for loading plugin assets (e.g., JavaScript files). It's used to construct the full URL to the `/plugins` directory
   * @default ''
   **/
   baseUrl?: string;
}

type PluginsFullPaths = {
  pluginsApiPath: string;
  pluginsAssetsPath: string;
}


const DEFAULT_PLUGINS_API_PATH = "/api/v1/plugins"
const DEFAULT_PLUGINS_ASSETS_PATH = "/plugins"

const paramToOptions = (options?: RemotePluginLoaderOptions): PluginsFullPaths => {
   // default behavior: no args uses defaults
    if (options === undefined) {
      return {
        pluginsApiPath: PLUGINS_API_PATH,
        pluginsAssetsPath: DEFAULT_PLUGINS_ASSETS_PATH,
      };
    }
  
    return {
      pluginsApiPath: options.apiPrefix ? `${options.apiPrefix}${DEFAULT_PLUGINS_API_PATH}` : DEFAULT_PLUGINS_API_PATH,
      pluginsAssetsPath: options.baseUrl ? `${options.baseUrl}${DEFAULT_PLUGINS_ASSETS_PATH}` : DEFAULT_PLUGINS_ASSETS_PATH,
   };
}

remotePluginLoader(options?: RemotePluginLoaderOptions) {
   const { pluginsApiPath, pluginsAssetsPath } = paramToOptions(options);
   
   ...
}

```

In the case Perses is behind a reverse proxy, the values of `apiPrefix` and `baseUrl` will be the same. For users embedding Perses in their UIs, the api and baseURL locations might differ.

WDYT?

ui/plugin-system/src/runtime/UsageMetricsProvider.tsx
```diff
@@ -23,11 +23,13 @@ interface UsageMetrics {
   renderDurationMs: number;
   renderErrorCount: number;
   pendingQueries: Map<string, QueryState>;
+  baseUrl?: string;
 }
 
 interface UsageMetricsProps {
   project: string;
   dashboard: string;
+  baseUrl?: string;
```
jgbernalp: I think in this case is the `apiPrefix`

```suggestion
  apiPrefix?: string;
```

ui/plugin-system/src/remote/remotePluginLoader.ts
```diff
@@ -41,10 +41,51 @@ const isPluginModuleResource = (pluginModule: unknown): pluginModule is PluginMo
...
+const paramToOptions = (options?: RemotePluginLoaderOptions): ParsedPluginOptions => {
+  if (options === undefined) {
+    return {
+      pluginsApiPath: DEFAULT_PLUGINS_API_PATH,
+      pluginsAssetsPath: DEFAULT_PLUGINS_ASSETS_PATH,
+    };
+  }
+
+  return {
+    pluginsApiPath: `${options?.apiPrefix ?? ''}${DEFAULT_PLUGINS_API_PATH}`,
```
lachieh, line 74: @jgbernalp should this also attempt to use the `baseUrl` if no `apiPrefix` is provided?

```suggestion
    pluginsApiPath: `${options?.apiPrefix ?? options?.baseUrl ?? ''}${DEFAULT_PLUGINS_API_PATH}`,
```

I think I prefer the explicit behavior but it does mean that the usage looks a little more verbose (not necessarily an issue), for example [here](https://github.com/perses/perses/pull/3377/files#diff-f4e96f71ee886ea641bc5d25c08b85739a3f85a93a45a3d3820c95d452781adaR87) becomes:
```diff

+  const pluginLoader = remotePluginLoader({ baseUrl: getBasePathName(), apiPrefix: getBasePathName() });
// ...
- <PluginRegistry pluginLoader={remotePluginLoader({ baseUrl: getBasePathName() })}>
+ <PluginRegistry pluginLoader={remotePluginLoader(pluginLoader)}>
```

This is how I have it in the latest push 👆🏻

jgbernalp: I don't think so. The `baseUrl` and `apiPrefix` can be different, so the plugin loader should be something like:
```
const pluginLoader = remotePluginLoader({ baseUrl: getBasePathName(), apiPrefix: getApiPrefix() });
```

we don't have the `getApiPrefix` function yet, but these values should come from the [perses config](https://perses.dev/perses/docs/configuration/configuration/#:~:text=path%20%3Capi_prefix%3E/api-,api_prefix,-%3A%20%3Cstring%3E) using the [ConfigProvider](https://github.com/perses/perses/blob/main/ui/app/src/context/Config.tsx)

ui/plugin-system/src/remote/remotePluginLoader.ts
lachieh: I'll handle this one as a separate PR

ui/plugin-system/src/remote/remotePluginLoader.ts
lachieh: Ok cool, added a `useRemotePluginLoader` hook inside app since the usage was the same across the app.

ui/app/src/model/config-client.ts
```diff
@@ -190,6 +190,7 @@ export interface DatasourceConfig {
 }
 
 export interface ConfigModel {
+  api_prefix?: string;
```
jgbernalp, line 194: Sorry I overlooked the fact that this uses a fetch to get the config. so if the API has a prefix fetching the config won't work as at that moment the prefix is not known. One option is to set this value globally on the window object. And serve a Javascript from the backend with the value. From react we can read that value, same as the plugins will do for the assets path.

lachieh: Oh, I missed this too. Ok, yeah, let me take a look at something.

I'm thinking that if the go backend pulls the values from the config and injects them into the HTML then they'll be up to date on reload. The alternative is replacing them at runtime, but that means the container needs a restart if anything changes.

What do you think?

jgbernalp: I'd say we can add the minimal params so the config can be fetched, in this case we just need the apiPrefix. If the config changes the frontend will just do a new request. Maybe is better to serve a Javascript file rather than parse the whole HTML. WDYT?

lachieh: Yep, minimal coverage works for me.

I think js works, but I think it's still going to require some templating/replacement in the html if the basePath changes. I suppose we could use a relative path to load the file?

Injecting it straight into the HTML means that it's always available and won't rely on any extra network requests to get the values.

lachieh: Oh nevermind, basePath is handled with the `PREFIX_PATH_PLACEHOLDER/` replacement. Ok, I'm good with the js approach. Can even handle it in the same manner as that.

lachieh: Ended up using rspack since it's already being used to do the same thing for the public path assets:
[Just these changes are here](https://github.com/perses/perses/pull/3377/commits/f4d02cbd454e490ddc5bbfe53367133d09c9837e)

ibakshay:
Thank you @lachieh for this contribution! 
Can you tell if this PR also solves this issue https://github.com/perses/perses/issues/3384?

lachieh:
I don't think it will, though it could have something to do with the way the oidc handling is managing the state that is passed through to the provider.

The more I did into this, I think a lot of the misconfiguration and misunderstanding (including my own) comes from the fact that `api_prefix` isn't only used as the _API_ prefix but sometimes is used as the base application path (see the publicPath setting in rspack.config.ts) which is relevant for path-based reverse proxy configuration. In the React application, the `getBasePathName` function is used to determine this in most places but not all. That function also isn't available when using `@perses/plugin-system` as a dependency, though there is a a plan to solve this one.

All that to say, I'm wondering if there should either be a limit on changing the api_prefix and base_path as the one setting or making both available to configure. With either usage, I think it's important to audit the usage and be consistent across both front and backend.

ui/app/src/model/config-client.ts
```diff
@@ -213,6 +215,6 @@ export function useConfig(options?: ConfigOptions): UseQueryResult<ConfigModel,
 }
 
 export function fetchConfig(): Promise<ConfigModel> {
-  const url = buildURL({ resource: resource, apiPrefix: '/api' });
+  const url = buildURL({ resource: resource, apiPrefix: `${PERSES_APP_CONFIG.api_prefix}/api` });
```
jgbernalp: this results in the prefixApi added 2 times, one from the config and another one from the getBasePathName. Agree that is very convoluted the handling of apiPrefix and basePath. Looking deep into the code the apiPrefix is not only used for APIs but for general assets. Also in this case the buildURL function prefixAsset has a different meaning, describing which api to use as we have `/api` and `/api/v1` endpoints both serving valid APIs.

Maybe the url-builder should look like:
```diff
import { getBasePathName } from './route';

-const apiPrefix = '/api/v1';
+const apiUrl = '/api/v1';

export type URLParams = {
  resource: string;
  name?: string;
  project?: string;
  pathSuffix?: string[];
  queryParams?: URLSearchParams;
- apiPrefix?: string;
+ apiUrl?: string;
+ apiPrefix?: string;
};

export default function buildURL(params: URLParams): string {
-  const basePath = getBasePathName();
+  const basePath = params.apiPrefix !== undefined ? params.apiPrefix : getBasePathName();
-  let url = params.apiPrefix === undefined ? apiPrefix : params.apiPrefix;
+  let url = params.apiUrl === undefined ? apiUrl : params.apiUrl;
  if (params.project !== undefined && params.project.length > 0) {
    url = `${url}/projects/${encodeURIComponent(params.project)}`;
  }
  url = `${url}/${params.resource}`;
  if (params.name !== undefined && params.name.length > 0) {
    url = `${url}/${encodeURIComponent(params.name)}`;
  }

  if (params.pathSuffix && params.pathSuffix.length > 0) {
    url = `${url}/${params.pathSuffix.join('/')}`;
  }

  if (params.queryParams !== undefined) {
    url = `${url}?${params.queryParams.toString()}`;
  }
  return basePath + url;
}
```

Then the config client will be:

```typescript
...
const url = buildURL({ resource: resource, apiUrl: '/api', apiPrefix: PERSES_APP_CONFIG.api_prefix });
```

lachieh: > Looking deep into the code the apiPrefix is not only used for APIs but for general assets. Also in this case the buildURL function prefixAsset has a different meaning, describing which api to use as we have /api and /api/v1 endpoints both serving valid APIs.

Yeah, agreed. Updated!

jgbernalp:
LGTM, awesome work!. I think the only remaining part will be https://github.com/perses/plugins/pull/391 improvement and add tests specific to reverse proxies to check that the plugins also work correctly. But the current change can be merged as it was made in a non-breaking fashion.

jgbernalp:
Any thoughts @Gladorme @AntoineThebaud @shahrokni @Nexucis ?

lachieh:
@jgbernalp there is the small breaking change for `@perses/plugin-system`'s `remotePluginLoader`:

```ts
// before
function remotePluginLoader(baseURL?: string): PluginLoader

// after
type RemotePluginLoaderOptions = { apiPrefix?: string; baseUrl?: string; };
function remotePluginLoader(options?: RemotePluginLoaderOptions): PluginLoader
```

jgbernalp:
You are right. We should change the PR title so the change is recorded correctly in the change log.

Nexucis:
As a side comment, I am struggling to understand how this PR is solving the issue linked (#2579).
From what I have understand from the code changed, it will more solve the issue https://github.com/perses/perses/issues/3396

jgbernalp:
> As a side comment, I am struggling to understand how this PR is solving the issue linked (#2579). From what I have understand from the code changed, it will more solve the issue #3396

It solves both issues, https://github.com/perses/perses/issues/2579 completely by adding the [prefix](https://github.com/perses/perses/pull/3377/files#diff-237e008678bc9fff1b27463a5490c243bae0055bc1be4d35bb8bff1d4a2c5e3fR78) for view tracking. And is part of the fix for the plugins with a set of related PRs on the plugins side.

Nexucis:
ah I didn't see this line of code, my apologies. Thank you for the pointer @jgbernalp. It will be really cool to see this PR merged. That should reduce the haemorrhage of issues related to the `api_prefix`.

PR #3374, opened 2025-09-18
[BUGFIX] pass baseUrl to loadPlugin

lachieh:
# Description

Fixes #3371

Updates the `remotePluginLoader` to correctly pass the `baseUrl` to the `loadPlugin` function. This ensures that plugins are loaded from the correct base URL, preventing potential issues with loading resources.

Also added a test file for this one specifically.

# Screenshots

N/A

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [x] E2E tests are stable and unlikely to be flaky.
  See [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

PR #3412, opened 2025-10-01
[ignore] fix usage of apiPrefix vs apiURL

Nexucis:
The PR #3377 broke few functionalities like getting the list of the plugins in app.
This PR aims to fix that.

As a side effect I have renamed the field `apiUrl` by `apiURL` as URL is not a word.

/cc @lachieh

### DIFF
--- a/ui/app/src/model/auth-client.ts
+++ b/ui/app/src/model/auth-client.ts
@@ -101,7 +101,7 @@ export function useNativeAuthMutation(): UseMutationResult<void, Error, NativeAu
 }
 
 export function nativeAuth(body: NativeAuthBody): Promise<void> {
-  const url = buildURL({ resource: `${authResource}/providers/native/login`, apiPrefix: '/api' });
+  const url = buildURL({ resource: `${authResource}/providers/native/login`, apiURL: '/api' });
   return fetchJson<void>(url, {
     method: HTTPMethodPOST,
     headers: HTTPHeader,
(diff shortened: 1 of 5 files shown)


Nexucis:
by the way I saw here https://github.com/perses/perses/blob/main/ui/app/src/model/config-client.ts#L218 we are using `PERSES_APP_CONFIG.api_prefix` and I was wondering why this has not been used in every fetch call.

I know I should have asked in #3377, sorry for getting late in the party.

lachieh:
@Nexucis It does seem that there are some changes from #3377 that may not be necessary after all the iterations.

For posterity, the `remotePluginLoader` was not passing the `baseURL: string` param on to the PluginRuntime's `registerRemote`. This was partly fixed in #3374, but then I realized that it still wasn't correct because inside the baseURL will replace the whole path if it is provided:

https://github.com/perses/perses/blob/dc07984822f6d518ec895dec2bf59dc5789bc098/ui/plugin-system/src/remote/PluginRuntime.tsx#L207

That led us to determine that `remotePluginLoader()` needs the api_prefix from the backed, which is loaded via the API, which needs the api_prefix from the backend, which 🔁... so it now gets handled the same way the public assets are meaning that it is available in the UI before fetching from `/api/v1/config`.

All that to say, are both `getBasePathName()` and `PERSES_APP_CONFIG` needed?

Nexucis:
> All that to say, are both getBasePathName() and PERSES_APP_CONFIG needed?

Well I would tend to say PERSES_APP_CONFIG.api_prefix should be enough. The whole point of `getBasePathName()` was to determinate dynamically the `api_prefix` but since this is injected now statically, this is not needed anymore. Or if this is the case, then that should not and needs to be fixed.

So, let's removes the function `getBasePathName`, and hopefully it will give a coherent way to manage this prefix

PR #3416, opened 2025-10-02
[ignore] Remove usage of the function getBasePathName

Nexucis:
following what we discussed in https://github.com/perses/perses/pull/3412 with @lachieh, this PR is removing the usage of the function `getBasePathName` in favor of using the static variable `PERSES_APP_CONFIG.api_prefix` replaced at runtime by the server.


jgbernalp:
We probably should add a default value as `''` in https://github.com/perses/perses/blob/main/ui/app/src/config.ts#L16

Nexucis:
> We probably should add a default value as `''` in https://github.com/perses/perses/blob/main/ui/app/src/config.ts#L16

ah yeah good idea, I will add that

lachieh:
Actually, that part is replaced statically at compile time by rspack through the Define plugin, so it will always be defined:
https://github.com/perses/perses/blob/dc07984822f6d518ec895dec2bf59dc5789bc098/ui/app/rspack.config.mjs#L99-L102

So the actual output for production builds is:
```diff
export const PERSES_APP_CONFIG = { 
-  api_prefix: process.env.API_PREFIX ? process.env.API_PREFIX : '', 
+  api_prefix: 'PREFIX_PATH_PLACEHOLDER' ? 'PREFIX_PATH_PLACEHOLDER' : '', 
}; 
```

Then, the `PREFIX_PATH_PLACEHOLDER` is replaced by the go backend in any js file it serves:
https://github.com/perses/perses/blob/dc07984822f6d518ec895dec2bf59dc5789bc098/ui/endpoint.go#L169-L171

Nexucis:
@lachieh, I think @jgbernalp proposed that so you don't need to set the empty value for the tests. Before setting this default value, the tests were broken

lachieh:
Ahh, got it! Makes sense.

At compile time, the minification will actually take care of removing the rest since it's always truthy:
```ts
// source
  api_prefix: process.env.API_PREFIX ? process.env.API_PREFIX : '', 
// after DefinePlugin
  api_prefix: 'PREFIX_PATH_PLACEHOLDER' ? 'PREFIX_PATH_PLACEHOLDER' : '', 
// after minification:
  api_prefix: 'PREFIX_PATH_PLACEHOLDER', 
```

lachieh (approved):
I know I'm not a maintainer, but this looks good to me!

Nexucis:
> Ahh, got it! Makes sense.
> 
> At compile time, the minification will actually take care of removing the rest since it's always truthy:
> 
> ```ts
> // source
>   api_prefix: process.env.API_PREFIX ? process.env.API_PREFIX : '', 
> // after DefinePlugin
>   api_prefix: 'PREFIX_PATH_PLACEHOLDER' ? 'PREFIX_PATH_PLACEHOLDER' : '', 
> // after minification:
>   api_prefix: 'PREFIX_PATH_PLACEHOLDER', 
> ```

Great ! This is great this is simplified at build time.

> I know I'm not a maintainer, but this looks good to me!

Your inputs are valuable, thank you for approving this PR !

Issue #3448, opened 2025-10-13
Slowness when many panels are toggled


AntoineThebaud:
### What did you do?

It looks like the more panels were already rendered, slower it is to toggle more panels.

It's quite easy to reproduce with [Node Exporter full](https://demo.perses.dev/projects/perses/dashboards/nodeexporterfull) on the public demo:
- open the dashboard
- collapse+toggle again the first group -> it's quick to render
- toggle all panel groups (tip: quicker to do when going from bottom to top)
- collapse+toggle again the first group -> it's still fine at this stage, the lazy loading is doing its job (most of the panels are still not rendered) 
- scroll through the page to really load all panels at least once
- collapse+toggle again the first group -> you now get a noticeable delay (3~5 seconds) to render the graphs. 

### What did you expect to see?

The time to render a group should not be impacted by the amount of panels already present on the page, as these ones should not be impacted by rendering other unrelated panels.

This is happening because the other panels get refreshed, while this should of course not happen.

### Perses version

v0.53.0-beta.0

shahrokni:
@jgbernalp and @Nexucis 

# IMPORTANT UPDATE 03/03/2026
Based on the profiling, one of the main contributors to the sluggish behavior is the `formatDecimal` which uses `Intl.NumberFormat`
( **1,056 ms**)

The function formatDecimal creates a new instance of this object every time it runs. The initialization of an Intl.NumberFormat object is a complex process where the browser's engine must look up locale data and resolve formatting options. When called many times  as it likely is in a large dashboard, this repeated initialization becomes a massive performance bottleneck.

The profiling suggests that the main thread is spending a long time over this function. 

## Conclusion
I think we should have a caching mechanism to avoid excessive object creation. I will start immidiately.

```javaScript
const formatter = Intl.NumberFormat('en-US', formatterOptions);
```

PR #3474, opened 2025-10-16
[BUGFIX] Remove grid items content unnecessary re render by having a consistent scrollbar

shahrokni:
Relates to https://github.com/perses/perses/issues/3448

# Description 🖊️ 

One of the performance issues which has been found so far is a subtle change of the panels width when the scrollbar appears on the right side of the view port. This subtle difference changes the width of all panels, and consequently triggers an unnecessary re render. 

image

# The fix 🔧 

The fix uses a library by which overlay scrollbars are possible. So, the space is always reserved for the scrollbar and since it is overlay it doesn't affect the app visually!  

# Checklist

## UI Changes

### DIFF
--- a/ui/app/index.html
+++ b/ui/app/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <meta charset="utf-8" />
@@ -10,6 +10,10 @@
     <style>
       :root {
         color-scheme: light dark;
+      }
+
+      html {
+        overflow-y: hidden;
       }
 
       body {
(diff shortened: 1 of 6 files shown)


jgbernalp:
Not sure this is the right approach, as this behavior is different based on the OS / Browser you are using. Maybe a more simpler and robust approach would be to clamp the resize listeners to the nearest pixel, I can see in your tests that the change is a sub-pixel value. WDYT?

shahrokni:
> Not sure this is the right approach, as this behavior is different based on the OS / Browser you are using. Maybe a more simpler and robust approach would be to clamp the resize listeners to the nearest pixel, I can see in your tests that the change is a sub-pixel value. WDYT?

I am personally ok with the clamping idea. In this case it should not be a big deal, though we miss a consistent behavior for all scrollable views. However, the used library has proved to be a reliable one, and Perses app layout structure is not complex to be worried about the consequences. Using this library may help us to get rid of a lot of ugly scrollbars exit already. Anyway, let's have a 3rd voice. @Gladorme

Gladorme:
Let's try the clamping idea, but gridColWidth is only for 1 of the 12 column, no? It will be a little more than subpixels if panel is taking more than few column

shahrokni:
@jgbernalp and @Gladorme 

⚠️ I have one concern. It would be great, if you could help me to understand, if it is valid or not. 
Maybe here, the fix is not as simple as clamping the width. 

## Why the width is important in the dashboard ❓ 
In the dashboards width is consumed by `getSuggestedStepMs`. The returned value by this function determines how exactly the diagram should be depicted. **From Co-Pilot**: **_The container width plays a crucial role in rendering a time series diagram because it directly affects how the data points are visualized and how much detail can be displayed._**

## Likely issue with clamping (floor or round) the width

If the first assertion is correct, it means manipulating the width will affect the suggested_step_ms. Consequently the rendered diagram will not be accurate. 

## The consistent Scrollbar

An overlayed consistent scrollbar will never affect the width, meaning we can trust the calculated suggested_step_ms. 
Why? The overlayed nature of the scrollbar helps it to float over the viewport elements. It actually does not occupy any space.

jgbernalp:
clamping the width of the component should not affect results, as you point out the step is calculated based on the component width and the time range. Is the same as resizing the window a tiny bit, the results should still be accurate. We are just saying, do not resize the component every time there is a subpixel change.

shahrokni:
> Not sure this is the right approach, as this behavior is different based on the OS / Browser you are using. Maybe a more simpler and robust approach would be to clamp the resize listeners to the nearest pixel, I can see in your tests that the change is a sub-pixel value. WDYT?

@jgbernalp  I am wondering if the nearest value works well. The following examples show that the round, floor and ceiling functions would not work.

## Case 1
_This example shows rounding would not work_

- With_ScrollBar 95.125
- Without_ScrollBar 95.7




## Case 2
_Here floor and ceiling would not work_

- Without_ScrollBar 73.783203125
- With_ScrollBar 74.41666666666667

Gladorme:
Scrollbar style (and width) is different between OS and browsers, clamping looks to no be as easy as we think :/

IMO, overlayscrollbars looks to be a nice solution have a unified behavior for all browsers/OS and to not have the width changed because the scrollbar is appearing/disappearing + it will used only on the app, it will not impact embedded usage.

But we can still look for others solutions and check it later when we optimised other stuffs 😄

shahrokni:
> Scrollbar style (and width) is different between OS and browsers, clamping looks to no be as easy as we think :/
> 
> IMO, overlayscrollbars looks to be a nice solution have a unified behavior for all browsers/OS and to not have the width changed because the scrollbar is appearing/disappearing + it will used only on the app, it will not impact embedded usage.
> 
> But we can still look for others solutions and check it later when we optimised other stuffs 😄

@jgbernalp

PR #3486, opened 2025-10-21
[BREAKINGCHANGE] Grafana migration: change variable migration format for more flexibility

AntoineThebaud:
# Description

I found a way to simplify the variable migration scripts in most cases, by adding constraints to the grafana variable object passed instead of writing condition blocks. However to enable this I needed to adjust the logic of `GetPluginKind()` to check only for the presence of the `#var` string instead of `#var.type` . But `#var` is a quite generic naming so checking only on this pattern to assume the plugin is a variable is too risky. Hence why it gets renamed to `#grafanaVar` & plugins need to adjust their migration script accordingly (that's the breaking change - draft PR prepared for [official plugins](https://github.com/perses/plugins/pull/418)).

I don't think there are unofficial plugins out there that would get affected by this change, but who knows..

Doc adjusted in consequence.

# Checklist


docs/plugins/cue.md
```diff
@@ -173,23 +193,38 @@ package migrate
...
+
+import "strings"
+
+#grafanaVar: {
+	type: "custom" | "interval"
+	query: string
+	...
+}
+
+kind: "StaticListVariable"
```
Nexucis, line 219: yes I agree the script is easier to understand and to write. But if it is written in this form, then it will produce every time something while we would like it produces an empty string when the script does not apply for the variable targeted.

The migration of the variable is working based on what the script is producing: If it returns an empty string, then we are executing the next variable script. Until we have a script returning data.

So this change will break the migration logic of the variable and it won't work anymore.

AntoineThebaud: Actually it wont because it's checking for empty result but for [also for errors](https://github.com/perses/perses/blob/main/internal/api/plugin/migrate/variable.go#L138-L141). So if the grafana variable provided doesn't match the expected `#grafanaVar` definition, an error triggers and the migration logic will try migrating the variable with the next plugin = just like before.

Nexucis: ah yeah ok. That's a good thing.
Last thing on my side is that the breaking change is quite annoying actually. Because yeah we will release Perses, but then with this version the migration won't work anymore, until plugins are updated and then putting back in Perses then release again.

Is there anyone to avoid this breaking change ? Like can we support both model for one or two versions (like we can drop the old system in v0.54.0 or v0.55.0) ?

AntoineThebaud: Yeah I thought about that.. I was thinking to deal with that with pre-releases (we would end up with: perses v0.53.0 include new plugin versions that depend on perses v0.53.0-beta.X), WDYT?

Food for thoughts for a long term solution, as we're likely to face such situation again: as already discussed a way to deal with that would be to change the way Perses is released to make it "naked" with no plugins at all & have a wizard to help setup/install default plugins. Relates to https://github.com/perses/perses/discussions/3052. Maybe once the marketplace would be available we could think about such integration with Perses? (like: when Perses starts, if no plugins are detected it would propose to install all/selected official plugins from the marketplace?)

Nexucis: That won't work most of the time because Perses on production won't have access to internet.

And if you don't break things but instead manage it smoothly then thé release cycle will not be an issue anymore and it release the pressure around that

AntoineThebaud: Well the issue will remain for the non-standard plugins anyway (non-standard != unofficial, e.g at some point I guess we will stop embedding all the official datasource plugins in the main archive because there would be too much of them), so  the problem is just shifting here.

About avoiding the breaking change / "can we support both model for one or two versions (like we can drop the old system in v0.54.0 or v0.55.0) ?", I guess I can build something, again I think that for this precise case that's not worth it because even in our case, we have 5 official data source plugins and only 1 of them (Prom) is impacted by the change. So I really dont expect anyone outside ourselves to be impacted (hence my proposal to solve this by leveraging pre-releases).  Anyway, please confirm you really want it and I shall do it.

Nexucis: I am a bit worried, this feature is actually much more used than we thought, but I am ok to move forward on that topic.

Regarding the release logic, let's make a first beta release with a warning to user to not use it and then we can update the plugin and then release again with a proper release note ..etc.
Will do that

Issue #3484, opened 2025-10-20
`refreshInterval` can't be updated


AntoineThebaud:
### What did you do?

- Create a new dashboard from the UI
- change the refresh interval from Off to a given duration
- I would like to make this interval the new default for my dashboard (= save the value in the `refreshInterval` field of the dashboard) -> unless tweaking the JSON directly, this is not possible. When saving the dashboard it proposes to save my changes done to the time range and/or variables, but not to the refresh interval.
 
The `refreshInterval` attribute was added for this purpose of saving a default, hence why this is a bug and not a feature.

### What did you expect to see?

The UI should provide a way to save a new default value for the refresh interval of a dashboard.

### Perses version

0.53.0-beta.0

### Perses configuration file

```yaml

```

### Logs

```text

```

PR #3507, opened 2025-10-27
[BUGFIX] Save the refresh interval chnages from UI

shahrokni:
Closes https://github.com/perses/perses/issues/3484

# Description 🖊️ 

This change adds the `refresh interval` into the Save Dialog of the dashboard. If approved by the user, the new value is saved as the default value. 

image

# Test 🧪 

- [X] **1-** The refresh interval message should always appear in the dialog (Whether it changed or not)
- [X] **2-** If it was intact, the text should and the textbox should appear disabled
- [X] **3-** If it had changed, the text and the textbox should appear enabled
- [X] **3-1** User may choose to check/uncheck the change to be persisted as the default value
- [X] **4-** After the page is saved and reloaded, the according changes should be observed

# Checklist

## UI Changes

### DIFF
--- a/ui/dashboards/src/components/SaveChangesConfirmationDialog/SaveChangesConfirmationDialog.tsx
+++ b/ui/dashboards/src/components/SaveChangesConfirmationDialog/SaveChangesConfirmationDialog.tsx
@@ -13,7 +13,7 @@
 
 import { ReactElement, useState } from 'react';
 import { Checkbox, FormGroup, FormControlLabel, Typography } from '@mui/material';
-import { useTimeRange } from '@perses-dev/plugin-system';
+import { DEFAULT_REFRESH_INTERVAL_OPTIONS, useTimeRange } from '@perses-dev/plugin-system';
 import { isRelativeTimeRange } from '@perses-dev/core';
 import { Dialog } from '@perses-dev/components';
 import { useSaveChangesConfirmationDialog, useVariableDefinitionActions } from '../../context';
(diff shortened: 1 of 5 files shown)


ui/plugin-system/src/components/TimeRangeControls/TimeRangeControls.tsx
```diff
@@ -195,7 +195,16 @@ export function TimeRangeControls({
         <InfoTooltip description={TOOLTIP_TEXT.refreshInterval}>
           <RefreshIntervalPicker
             timeOptions={DEFAULT_REFRESH_INTERVAL_OPTIONS}
-            value={refreshInterval}
+            value={
```
andreasgerstmayr, line 198: Instead of this workaround, can we agree on always using `pastDuration` internally (also for storing the value in the dashboard), and the display value we get dynamically with
```
DEFAULT_REFRESH_INTERVAL_OPTIONS.find((i) => i.value.pastDuration === refreshValue)?.display ?? refreshValue
```
?
I noticed the `RefreshIntervalPicker` has functionality to add a new entry if the `value` is not part of `timeOptions`, so we need the `?? refreshValue` at the end.

That should simplify the code and allows us to remove the conditional here and in `SaveChangesConfirmationDialog.tsx`.

On a related note, I wonder why we don't use
```
  { value: { pastDuration: '1m' }, display: '1m' },
```
in the `DEFAULT_REFRESH_INTERVAL_OPTIONS`. The `60s` is a human readable string and will be translated to 60 seconds, so I'm 99% sure `1m` would also work just fine :)

ui/dashboards/src/components/SaveChangesConfirmationDialog/SaveChangesConfirmationDialog.tsx
```diff
@@ -25,25 +25,35 @@ export const SaveChangesConfirmationDialog = (): ReactElement => {
...
-  const { timeRange } = useTimeRange();
-  const currentTimeRangeText = isRelativeTimeRange(timeRange)
+  const { timeRange, refreshInterval } = useTimeRange();
+
+  const CURRENT_TIME_RANGE_TEXT = isRelativeTimeRange(timeRange)
     ? `(Last ${timeRange.pastDuration})`
     : '(Absolute time ranges can not be saved)';
 
-  const saveTimeRangeText = `Save current time range as new default ${currentTimeRangeText}`;
+  const SAVE_TIME_RANGE_MESSAGE = `Save current time range as new default ${CURRENT_TIME_RANGE_TEXT}`;
```
andreasgerstmayr: Why is the variable in uppercase, is this a new convention in Perses code?

Issue #3459, opened 2025-10-13
Loki shows up in Prometheus Explorer


AntoineThebaud:
### What did you do?

- Go to demo website > Prometheus Explorer view
- Notice that "Loki Time Series Query" appears in the Query Type dropdown, although this Explorer view is Prometheus-specific
<img width="416" height="260" alt="Image" src="https://github.com/user-attachments/assets/add4b742-2edc-4d2d-ad72-06df111f718a" />

### What did you expect to see?

Non-Prometheus stuff should not show up in this view.

### Perses version

v0.53.0-beta.0

PR #3514, opened 2025-10-29
[BUGFIX]: Apply plugin filter over Multi Query Editor

shahrokni:
Relates to https://github.com/perses/perses/issues/3459

# WARNING ⚠️ 
> This PR should be reviewed with its plugin counterpart to be effective!

> This PR should be merged first. The other PR has a decency to plugin-system 

https://github.com/perses/plugins/pull/428

# Description 🖊️ 

To load **Query Plugins**, the **Multi Query Editor** component only takes the **Query Types** into account . 
This means, all Query Plugins which are grouped under the same query type will be shown in the component. 
For instance, if the query type  is set to **TimeSeriesQuery**, both **Locki** and **Prometheus** Time Series Queries will be exposed from the component. 

The mentioned behavior is not correct in the Prometheus Explorer. As the issue points out in the Prometheus Explorer only Prometheus related plugins should be available. 

# Solution 🔧 

To be able to filter out a subgroup of Query Plugins, an optional filter has been added. 
To solve this issue, we can feed the filter from the **PrometheusExplorer** and expect to see the desired values. 
This filter is **optional**, so the change is backward compatible. 

```typescript
export function PrometheusExplorer(): ReactElement {
  const {
    data: { tab = 'table', queries = [], datasource = DEFAULT_PROM, filters = [], exploredMetric = undefined },
    setData,
    explorer,
  } = useExplorerManagerContext<MetricsExplorerQueryParams>();

  const filteredQueryPlugins = explorer ? ['PrometheusTimeSeriesQuery'] : [];
```

# Test 🧪 

## WARNING ⚠️ 
To test this fix you need to 

1. **Build** plugin-system (turbo run build)
2. **Move** the built package to the plugins \node_modules\@perses-dev
3. Run the client and server
4. Run the Prometheus plugin locally 

## Test 1 Check the Prometheus Explorer

1. You should not see Loki anymore in Table and Graph 
2. Run a Query and make sure it is working
3. Not necessary, but if you feel better take a look at the other explorers as well

## Test 2 Check the Query Editor in dashboard

1. The Query Editor in Panel Editor forms MUS HSOW both Loki  and Prometheus 

image

image

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
+++ b/ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
@@ -22,11 +22,15 @@ import { QueryEditorContainer } from './QueryEditorContainer';
 
 export interface MultiQueryEditorProps {
   queryTypes: QueryPluginType[];
+  filteredQueryPlugins?: string[];
   queries?: QueryDefinition[];
   onChange: (queries: QueryDefinition[]) => void;
 }
 
-function useDefaultQueryDefinition(queryTypes: QueryPluginType[]): {
+function useDefaultQueryDefinition(
+  queryTypes: QueryPluginType[],
+  filteredQueryPlugins?: string[]
+): {
   defaultInitialQueryDefinition: QueryDefinition;
   isLoading: boolean;
 } {
@@ -35,12 +39,18 @@ function useDefaultQueryDefinition(queryTypes: QueryPluginType[]): {
 
   // Firs the default query type
   const defaultQueryType = queryTypes[0]!;
-
   // Then the default plugin kind
   // Use as default the plugin kind explicitly set as default or the first in the list
   const { data: queryPlugins, isLoading } = useListPluginMetadata(queryTypes);
   const { defaultPluginKinds } = usePluginRegistry();
-  const defaultQueryKind = defaultPluginKinds?.[defaultQueryType] ?? queryPlugins?.[0]?.spec.name ?? '';
+
+  let defaultQueryKind: string = '';
+
+  if (filteredQueryPlugins?.length) {
+    defaultQueryKind = queryPlugins?.find((i) => filteredQueryPlugins.includes(i.spec.name))!.spec.name ?? '';
+  } else {
+    defaultQueryKind = defaultPluginKinds?.[defaultQueryType] ?? queryPlugins?.[0]?.spec.name ?? '';
+  }
 
   const { data: defaultQueryPlugin } = usePlugin(defaultQueryType, defaultQueryKind, {
     useErrorBoundary: true,
@@ -69,8 +79,8 @@ function useDefaultQueryDefinition(queryTypes: QueryPluginType[]): {
  */
 
 export const MultiQueryEditor = forwardRef<PluginEditorRef, MultiQueryEditorProps>((props, ref): ReactElement => {
-  const { queryTypes, queries = [], onChange } = props;
-  const { defaultInitialQueryDefinition, isLoading } = useDefaultQueryDefinition(queryTypes);
+  const { queryTypes, queries = [], filteredQueryPlugins, onChange } = props;
+  const { defaultInitialQueryDefinition, isLoading } = useDefaultQueryDefinition(queryTypes, filteredQueryPlugins);
   // State for which queries are collapsed
   const [queriesCollapsed, setQueriesCollapsed] = useState(queries.map(() => false));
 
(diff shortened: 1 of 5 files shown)


ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
```diff
@@ -35,12 +39,21 @@ function useDefaultQueryDefinition(queryTypes: QueryPluginType[]): {
...
   // Use as default the plugin kind explicitly set as default or the first in the list
   const { data: queryPlugins, isLoading } = useListPluginMetadata(queryTypes);
   const { defaultPluginKinds } = usePluginRegistry();
-  const defaultQueryKind = defaultPluginKinds?.[defaultQueryType] ?? queryPlugins?.[0]?.spec.name ?? '';
+
+  let defaultQueryKind: string = '';
+
+  if (
+    !filteredQueryPlugins?.length ||
+    !queryPlugins?.map((i) => i.spec.name).some((i) => filteredQueryPlugins.includes(i))
```
Gladorme: I think second condition can be removed here, it will be filtered anyway after, no?

ui/plugin-system/src/components/PluginKindSelect/PluginKindSelect.tsx
```diff
@@ -31,13 +32,17 @@ export interface PluginKindSelectProps extends Omit<TextFieldProps, 'value' | 'o
...
+  const { pluginTypes, value: propValue, onChange, filteredQueryPlugins, ...others } = props;
   const { data, isLoading } = useListPluginMetadata(pluginTypes);
 
-  const sortedData = useMemo(
-    () => data?.sort((a, b) => a.spec.display.name.localeCompare(b.spec.display.name)),
-    [data]
-  );
+  const sortedData = useMemo(() => {
+    if (!filteredQueryPlugins?.length || !data?.map((i) => i.spec.name).some((i) => filteredQueryPlugins.includes(i)))
+      return data?.sort((a, b) => a.spec.display.name.localeCompare(b.spec.display.name));
```
Gladorme: Let's add some brackets :)
```suggestion
{
      return data?.sort((a, b) => a.spec.display.name.localeCompare(b.spec.display.name));
      }
```

ui/plugin-system/src/components/PluginKindSelect/PluginKindSelect.tsx
```diff
@@ -31,13 +32,17 @@ export interface PluginKindSelectProps extends Omit<TextFieldProps, 'value' | 'o
...
-  const { pluginTypes, value: propValue, onChange, ...others } = props;
+  const { pluginTypes, value: propValue, onChange, filteredQueryPlugins, ...others } = props;
   const { data, isLoading } = useListPluginMetadata(pluginTypes);
 
-  const sortedData = useMemo(
-    () => data?.sort((a, b) => a.spec.display.name.localeCompare(b.spec.display.name)),
-    [data]
-  );
+  const sortedData = useMemo(() => {
+    if (!filteredQueryPlugins?.length || !data?.map((i) => i.spec.name).some((i) => filteredQueryPlugins.includes(i)))
```
Gladorme:
```suggestion
    if (!filteredQueryPlugins?.length)
```

ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
shahrokni: I think we should keep the second condition. Imagine the array has some items, but the items are technically garbage. Since any string could be placed in the array, we need to make sure the items do really exist in queryPlugins. queryPlugins are always reliable, because they are provided by an endpoint. 


So, if the array is not empty, and  it contains some values which match what we have from the API, then we are good to go. Otherwise, we should not use the garbage filters

ui/plugin-system/src/components/PluginKindSelect/PluginKindSelect.tsx
shahrokni: https://github.com/perses/perses/pull/3514#discussion_r2474099720

ui/plugin-system/src/components/MultiQueryEditor/MultiQueryEditor.tsx
Gladorme: I don't agree with "garbage" items. Either you specify a filter and you filter, or you don't filter.

Issue #3226, opened 2025-08-25
Variable Preview Values shows an empty item in the `Create Variable Form`


shahrokni:
### What did you do?

- Open a project
- Go to the Variable Tab
- Try to create a new variable
- Variable Preview Values shows an empty item in the `Create Variable Form`

<img width="2388" height="752" alt="Image" src="https://github.com/user-attachments/assets/a3449f57-cc70-46f9-9334-1471c4a17bf1" />

### What did you expect to see?

The empty item should not exist

### Perses version

0.52

AntoineThebaud:
Closed by https://github.com/perses/perses/pull/3517

PR #3517, opened 2025-10-30
    [BUGFIX]: Fix empty value in create variable form (#3226)

prakhar29jain:
# Description

Even when a Chip has an empty label, it still renders with minimum dimensions (around 16px height) due to Material-UI's default styling.

# Screenshots

Before: 
image

After:
image

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
+++ b/ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
@@ -77,7 +77,10 @@ export function VariablePreview(props: VariablePreviewProps): ReactElement {
       <Card variant="outlined">
         <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: 1, m: 2 }}>
           {variablePreviewState}
-          {values?.slice(0, maxValues).map((val, index) => <Chip size="small" key={index} label={val} />)}
+          {values
+            ?.slice(0, maxValues)
+            .filter((val) => val)
+            .map((val, index) => <Chip size="small" key={index} label={val} />)}
           {notShown > 0 && <Chip onClick={showAll} variant="outlined" size="small" label={`+${notShown} more`} />}
         </Box>
       </Card>


ibakshay:
@prakhar29jain,  Thank you for your contribution! Please [sign off your commit](https://github.com/perses/perses/pull/3517/checks?check_run_id=54112681294)

ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
```diff
@@ -77,7 +77,10 @@ export function VariablePreview(props: VariablePreviewProps): ReactElement {
       <Card variant="outlined">
         <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: 1, m: 2 }}>
           {variablePreviewState}
-          {values?.slice(0, maxValues).map((val, index) => <Chip size="small" key={index} label={val} />)}
+          {values
+            ?.slice(0, maxValues)
+            .filter((val) => val && val.trim().length > 0)
+            .map((val, index) => <Chip size="small" key={index} label={val} />)}
```
shahrokni, line 83: So, it seems `values` includes an item which is literally an empty string. The fix you have provided, jumps over the empty values, which is a good workaround.  However, we should find the root of the problem. Here, the main question is why the values should have an empty string? Where does it come from? Is something wrong with initiating the form?

Gladorme: > Where does it come from?

Most probably from the input default value. Inputs will be always initialized with a empty string (and not undefined) because of controlled / uncontrolled behavior

ui/plugin-system/src/components/Variables/VariableEditorForm/VariablePreview.tsx
```diff
@@ -77,7 +77,10 @@ export function VariablePreview(props: VariablePreviewProps): ReactElement {
       <Card variant="outlined">
         <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: 1, m: 2 }}>
           {variablePreviewState}
-          {values?.slice(0, maxValues).map((val, index) => <Chip size="small" key={index} label={val} />)}
+          {values
+            ?.slice(0, maxValues)
+            .filter((val) => val && val.trim().length > 0)
```
Gladorme: Maybe stupid, but I think white space could be a valid value.
```suggestion
            .filter((val) => val)
```

shahrokni: Really? So are you suggesting this is not a bug at all?

Gladorme: Empty value is a bug, but not value with only whitespace

prakhar29jain: @Gladorme Value with a whitespace do you mean we would allow users to assign give variable values = " " ? or maybe constant in dashboards if they need ( maybe in queries)... Well I can't think of any particular use case

Gladorme: @prakhar29jain yep (allow users to assign give variable values = " " ?), I told it may be stupid 😄 

But I imagine it's possible someone wants only a whitespace as value of a variable.

Gladorme (approved):
lgtm! Thanks for the contribution 😄

PR #3518, opened 2025-10-31
[BUGFIX/BREAKINGCHANGE] Remove memory leak on refresh + remove refreshKey trick

Gladorme:
# Description

The goal of this PR is to remove the trick around refreshKey for updating the time range and all associated items (panels, variables, ...) by using appropriated Tanstack-Query functions and remove a memory leak.

The memory leak is increasing every time a dashboard is refreshed: it will run queries for panels/variables and this will be cached by Tanstack-Query with a queryKey.

This queryKey for time-series queries is containing:
- definition => def of query + datasource selected. OK
- timeRange => absolute time range. OK, but if relative timerange used, it will never be overriden
- datasourceStore => NOT OK, removed: not serializable to string or json...
- mode => instant or range, OK
- variablesValueKey => variables, OK
- refreshKey => timerange + refreshCounter. NOT OK, the counter we never be the same, even with absolute timerange

The thing is we never invalidate or remove inactive cached entry when refreshing. So you can increase memory to the moon :rocket:
The issue is a little bit mitigated by Tanstack-Query with `cacheTime`, that will wipe all inactive query after 5m (default duration).

I aligned queryKeys of all query types, so we can easily invalidate them / remove them on refresh.

> [!IMPORTANT]  
> In this PR, I also separated data management from `TimeRangeProvider`, it's removing some ugly `useEffect`. If you were using it, you can migrate to `TimeRangeProviderBasic`, that component should have the same behavior.

TODO next: improve query editor "Run query" to only invalidate the selected query (currently it will refresh everything, it's calling refresh method)

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
+++ b/ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
@@ -74,8 +74,16 @@ export function TimeRangeProvider(props: TimeRangeProviderProps): ReactElement {
   const { timeRange, refreshInterval, children, setTimeRange, setRefreshInterval } = props;
 
   const queryClient = useQueryClient();
-  const [absolutTimeRange, setAbsoluteTimeRange] = useState<AbsoluteTimeRange>(
+  const [absoluteTimeRange, setAbsoluteTimeRange] = useState<AbsoluteTimeRange>(
     isRelativeTimeRange(timeRange) ? toAbsoluteTimeRange(timeRange) : timeRange
+  );
+
+  const handleSetTimeRange = useCallback(
+    (value: TimeRangeValue) => {
+      setTimeRange(value);
+      setAbsoluteTimeRange(isRelativeTimeRange(value) ? toAbsoluteTimeRange(value) : value);
+    },
+    [setTimeRange]
   );
 
   // Refresh is called when clicking on the refresh button, it refreshes all queries including variables
@@ -112,14 +120,22 @@ export function TimeRangeProvider(props: TimeRangeProviderProps): ReactElement {
   const ctx = useMemo(() => {
     return {
       timeRange: timeRange,
-      setTimeRange: setTimeRange,
-      absoluteTimeRange: absolutTimeRange,
+      setTimeRange: handleSetTimeRange,
+      absoluteTimeRange: absoluteTimeRange,
       refresh,
       refreshInterval: refreshInterval,
       refreshIntervalInMs: refreshIntervalInMs,
       setRefreshInterval: setRefreshInterval,
     };
-  }, [absolutTimeRange, refresh, refreshInterval, refreshIntervalInMs, setRefreshInterval, setTimeRange, timeRange]);
+  }, [
+    absoluteTimeRange,
+    handleSetTimeRange,
+    refresh,
+    refreshInterval,
+    refreshIntervalInMs,
+    setRefreshInterval,
+    timeRange,
+  ]);
 
   return <TimeRangeContext.Provider value={ctx}>{children}</TimeRangeContext.Provider>;
 }


ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
```diff
@@ -57,10 +54,6 @@ export function PanelEditorForm(props: PanelEditorFormProps): ReactElement {
     defaultValues: initialValues,
   });
 
-  const { dashboard } = useDashboard();
-  const dashboardDuration = dashboard?.kind === 'Dashboard' ? dashboard.spec.duration : DEFAULT_DASHBOARD_DURATION;
-  const initialTimeRange = useInitialTimeRange(dashboardDuration);
```
shahrokni, line 62: # This may introduce a bug, if not handled properly 💀 

A `Time Range Provider` was added here to fix the problem with `Query Refresh Button` 
For details about the change you can take a look at the https://github.com/perses/perses/pull/3287

## What happens if you remove this ❓ 

Removing this time range provider eventually breaks the mentioned fix. As long as I remember, the reason this time range provider was added was to **separate it from the higher time range provider which is located higher in this hierarchy.** With removing this, the higher time range provider is triggered every time the Run Query Button is pressed. Consequently, all panels of the dashboard are triggered and get update which is not desiarble behavior here. 


image

ui/dashboards/src/components/Variables/VariableEditor.tsx
```diff
@@ -190,34 +177,28 @@ export function VariableEditor(props: {
     });
   };
 
-  const { dashboard } = useDashboard();
-  const dashboardDuration = dashboard?.kind === 'Dashboard' ? dashboard.spec.duration : DEFAULT_DASHBOARD_DURATION;
-  const initialTimeRange = useInitialTimeRange(dashboardDuration);
```
shahrokni, line 195: https://github.com/perses/perses/pull/3518/files#r2489532419

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
Gladorme: With refactor this is not needed anymore, it's still working as intended.

It's way cleaner now, no more duplication of providers 😉

Gladorme: Ok, I see what you mean, still duplication providers is not perfect, imo.

> TODO next: improve query editor "Run query" to only invalidate the selected query (currently it will refresh everything, it's calling refresh method)

I will try to do it now and it will fix this issue.

ui/plugin-system/src/runtime/TimeRangeProvider/TimeRangeProvider.tsx
```diff
@@ -73,64 +73,53 @@ export function useSuggestedStepMs(width?: number): number {
...
-  useEffect(() => {
-    setLocalTimeRange(timeRange);
-  }, [timeRange, refreshCounter]);
-
-  useEffect(() => {
-    setLocalRefreshInterval(refreshInterval);
-  }, [refreshInterval]);
+  const queryClient = useQueryClient();
+  const [absolutTimeRange, setAbsoluteTimeRange] = useState<AbsoluteTimeRange>(
+    isRelativeTimeRange(timeRange) ? toAbsoluteTimeRange(timeRange) : timeRange
```
shahrokni: Spelling mistake caught by VSCODE 
Correct form: `absoluteTimeRange`

shahrokni:
Please don't forget to complete the checklist when you are ready.

Gladorme:
Found an issue when changing timerange value, checking too

Nexucis:
@jgbernalp I can see here many changes in the package `plugin-system` and `dashboards`. As you started to migrate these packages to the repository `shared`, I guess a sync will be necessary.

jgbernalp:
Yes indeed a sync is necessary. But we first need a beta release so I can use the new core without react. Planning to do it this week

Nexucis:
your wish will be granted @jgbernalp https://github.com/perses/perses/pull/3530 ;) new release coming soon

ui/dashboards/src/components/PanelDrawer/PanelEditorForm.tsx
Gladorme: Fixed in #3532

ui/dashboards/src/components/Variables/VariableEditor.tsx
Gladorme: Fixed in #3532

Gladorme:
@jgbernalp Both PRs are ready (this one and #3532), waiting for your green flag before merging 😄

jgbernalp:
ok, I'll check this tomorrow

jgbernalp:
@Gladorme I wonder if we actually need caching for the queries made by plugins. Including the time range with millisecond accuracy just adds them into memory but they will never be cached.

Gladorme:
> @Gladorme I wonder if we actually need caching for the queries made by plugins. Including the time range with millisecond accuracy just adds them into memory but they will never be cached.

Are we talking about the queries in time-series.ts, trace-series.ts, log-queries.ts (ex: https://github.com/perses/perses/pull/3518/files#diff-be60bc9e89e9ecf81601012de961afcb7947fc6c1cfccb99c6cc23a5508f4af5R60)?

We need to keep cache them, because when you open the panel editor, the preview will reuse the data. Same when you collapse/uncollapse a panel group, it will reuse the data too with the same queryKey

Issue #3522, opened 2025-11-03
insecure_skip_verify is wrong perses.dev/perses/docs/configuration/configuration/


ch-stark:
### What did you do?

it should be:
insecureSkipVerify: true
https://perses.dev/perses/docs/configuration/configuration/

### Perses configuration file

```yaml

```

### Logs

```text

```

PR #3535, opened 2025-11-06
[BREAKINGCHANGE] Use only secret tls config in the config

Nexucis:
This PR is breaking the Perses SQL configuration.

### Context

We have inconsistency in the usage of the TLS config in the Perses config. Sometimes we are using the config coming from the Perses secret package, sometimes we are using the config coming from Prometheus.

The difference between each is on the case convention used. In Prometheus, they are using the snake_case, in Perses we are using the camelCase.

Using both result having two different YAML config for the same purpose which for ends users can be weird, and can loose them.

### Resolution

This PR is removing the usage of the Prometheus TLS config in favor of the Perses one to be consistant across the config and the different datamodel used.

### Migration Guideline

In the SQL configuration, if the `tls_config` is used, then you should change your config like that:

```
ca_file -> caFile
cert_file -> certFile
key_file -> keyFile
server_name -> serverName
insecure_skip_verify -> insecureSkipVerify
min_version -> minVersion (values: TLS10, TLS11, TLS12, TLS13)
max_version -> maxVersion (values: TLS10, TLS11, TLS12, TLS13)
```

### DIFF
--- a/pkg/model/api/config/database.go
+++ b/pkg/model/api/config/database.go
@@ -14,7 +14,6 @@
 package config
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"time"
@@ -53,7 +52,7 @@ func (f *File) Verify() error {
 
 type SQL struct {
 	// TLS configuration
-	TLSConfig *secret.TLSConfig `json:"tls_config,omitempty" yaml:"tls_config,omitempty"`
+	TLSConfig *secret.PublicTLSConfig `json:"tls_config,omitempty" yaml:"tls_config,omitempty"`
 	// Username
 	User secret.Hidden `json:"user,omitempty" yaml:"user,omitempty"`
 	// Password (requires User)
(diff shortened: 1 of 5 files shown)


celian-garcia:
FYI https://github.com/perses/helm-charts/blob/main/charts/perses/values.schema.json#L470-L495

ping @ibakshay @nicolastakashi can we accept having less controls in the helm chart ? It's a bit painful to have to change the charts for such case. It forces to have a 1 - 1 mapping between perses and its chart, where it could just be a relatively free yaml for the config.

pkg/model/api/config/database.go
```diff
@@ -53,7 +53,7 @@ func (f *File) Verify() error {
 
 type SQL struct {
 	// TLS configuration
-	TLSConfig *config.TLSConfig `json:"tls_config,omitempty" yaml:"tls_config,omitempty"`
+	TLSConfig *secret.TLSConfig `json:"tls_config,omitempty" yaml:"tls_config,omitempty"`
```
celian-garcia: I believe you can directly use the Public one here no ? I mean SQL config is not saved in DB

celian-garcia (changes requested):

Nexucis:
> FYI https://github.com/perses/helm-charts/blob/main/charts/perses/values.schema.json#L470-L495

yup that means this will need to update as soon as v0.53.0 is out.
/cc @ibakshay

pkg/model/api/config/database.go
Nexucis: ah yeah you are right indeed, that should be enough !

Nexucis:
this PR will close https://github.com/perses/perses/issues/3522

ibakshay:
> > FYI https://github.com/perses/helm-charts/blob/main/charts/perses/values.schema.json#L470-L495
> 
> yup that means this will need to update as soon as v0.53.0 is out. /cc @ibakshay

sure, I will take care of it :)

ibakshay:
> FYI https://github.com/perses/helm-charts/blob/main/charts/perses/values.schema.json#L470-L495
> 
> ping @ibakshay @nicolastakashi can we accept having less controls in the helm chart ? It's a bit painful to have to change the charts for such case. It forces to have a 1 - 1 mapping between perses and its chart, where it could just be a relatively free yaml for the config.

Yep, I am with you :)

PR #3545, opened 2025-11-07
Feature/k8s provider

celian-garcia:
# Description

This is a do not merge draft PR to review overall changes in that feature branch

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/internal/cli/cmd/login/device_code.go
+++ b/internal/cli/cmd/login/device_code.go
@@ -26,11 +26,20 @@ type deviceCodeLogin struct {
 	writer                io.Writer
 	externalAuthnKind     externalAuthnKind
 	externalAuthnProvider string
+	enablePKCE            bool
 	apiClient             api.ClientInterface
 }
 
 func (l *deviceCodeLogin) Login() (*oauth2.Token, error) {
-	deviceCodeResponse, err := l.apiClient.Auth().DeviceCode(string(l.externalAuthnKind), l.externalAuthnProvider)
+	var deviceCodeOpts []oauth2.AuthCodeOption
+	var tokenOpts []oauth2.AuthCodeOption
+	if l.enablePKCE {
+		verifier := oauth2.GenerateVerifier()
+		deviceCodeOpts = append(deviceCodeOpts, oauth2.S256ChallengeOption(verifier))
+		tokenOpts = append(tokenOpts, oauth2.VerifierOption(verifier))
+	}
+
+	deviceCodeResponse, err := l.apiClient.Auth().DeviceCode(string(l.externalAuthnKind), l.externalAuthnProvider, deviceCodeOpts...)
 	if err != nil {
 		return nil, err
 	}
@@ -40,7 +49,7 @@ func (l *deviceCodeLogin) Login() (*oauth2.Token, error) {
 		return nil, outErr
 	}
 
-	return l.apiClient.Auth().DeviceAccessToken(string(l.externalAuthnKind), l.externalAuthnProvider, deviceCodeResponse)
+	return l.apiClient.Auth().DeviceAccessToken(string(l.externalAuthnKind), l.externalAuthnProvider, deviceCodeResponse, tokenOpts...)
 }
 
 func (l *deviceCodeLogin) SetMissingInput() error {
--- a/internal/cli/cmd/login/login.go
+++ b/internal/cli/cmd/login/login.go
@@ -65,6 +65,7 @@ type option struct {
 	kube                  bool
 	kubeconfig            string
 	insecureTLS           bool
+	enablePKCE            bool
 	apiClient             api.ClientInterface
 	restConfig            clientConfig.RestConfigClient
 	remoteConfig          *backendConfig.Config
@@ -226,6 +227,7 @@ func (o *option) newLoginOption() (loginOption, error) {
 			writer:                o.writer,
 			externalAuthnKind:     o.externalAuthnKind,
 			externalAuthnProvider: o.externalAuthnProvider,
+			enablePKCE:            o.enablePKCE,
 			apiClient:             o.apiClient,
 		}, nil
 	}
@@ -368,7 +370,7 @@ percli login https://demo.perses.dev --provider <slug_id> --client-id <client_id
 	cmd.Flags().StringVar(&o.accessToken, "token", "", "Bearer token for authentication to the API server")
 	cmd.Flags().BoolVar(&o.kube, "kube", false, "Sets if the login should use the users login")
 	cmd.Flags().StringVar(&o.kubeconfig, "kubeconfig-file", "", "Kubeconfig file location to load Kubernetes token from. Defaults to KUBECONFIG env variable, then HOME/.kube/config if empty")
-
 	cmd.Flags().StringVar(&o.externalAuthnProvider, "provider", "", "External authentication provider identifier. (slug_id)")
+	cmd.Flags().BoolVar(&o.enablePKCE, "enable-pkce", false, "Enable PKCE (Proof Key for Code Exchange) for the device code flow.")
 	return cmd
 }
--- a/internal/cli/cmd/refresh/k8s.go
+++ b/internal/cli/cmd/refresh/k8s.go
@@ -26,7 +26,7 @@ type k8sRefresh struct {
 }
 
 func (k *k8sRefresh) refresh() error {
-	// If the k8sAuth secret hasn't been set yet then we know that the user hasn't attempted to log in
+	// If the k8sAuth secret hasn't been set, yet then we know that the user hasn't attempted to log in
 	if config.Global.RestClientConfig.K8sAuth == nil {
 		return fmt.Errorf("kubeconfig location has not been set yet, please use the command login to set it")
 	}
--- a/internal/cli/cmd/refresh/refresh.go
+++ b/internal/cli/cmd/refresh/refresh.go
@@ -111,7 +111,7 @@ func (n *refresh) refresh() error {
 	if err != nil {
 		return err
 	}
-	if writeErr := config.SetAccessToken(response.AccessToken); writeErr != nil {
+	if writeErr := config.SetTokens(response.AccessToken, response.RefreshToken); writeErr != nil {
 		return writeErr
 	}
 	return nil
--- a/internal/cli/config/config.go
+++ b/internal/cli/config/config.go
@@ -151,9 +151,10 @@ func SetProject(project string) error {
 	})
 }
 
-func SetAccessToken(token string) error {
+func SetTokens(accessToken, refreshToken string) error {
 	return Write(&Config{
-		RestClientConfig: config.RestConfigClient{Authorization: secret.NewBearerToken(token)},
+		RestClientConfig: config.RestConfigClient{Authorization: secret.NewBearerToken(accessToken)},
+		RefreshToken:     refreshToken,
 	})
 }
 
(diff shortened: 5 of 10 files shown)


internal/api/e2e/framework/server.go
```diff
@@ -139,24 +188,35 @@ func CreateServer(t *testing.T, conf apiConfig.Config) (*httptest.Server, *httpe
...
-	server, expect, persistenceManager := CreateServer(t, conf)
-	defer persistenceManager.GetPersesDAO().Close()
+	server, expect, dependencyManager := CreateServer(t, conf)
+	defer dependencyManager.Persistence().GetPersesDAO().Close()
+	defer server.Close()
+	entities := testFunc(server, expect, dependencyManager.Persistence())
+	ClearAllKeys(t, dependencyManager.Persistence().GetPersesDAO(), entities...)
+}
+
+func WithServerAuthConfig(t *testing.T, testFunc func(*httptest.Server, *httpexpect.Expect, dependency.Manager, string) []modelAPI.Entity) {
```
celian-garcia, line 203: I'm not fan of the way the test code is closed. 
I'd prefer something a bit more modular. Something in that spirit

```go
e2eframework.WithServerConfig(t, e2eframework.DefaultAuthConfig(), ...
    token, usrEntity, globalRole, globalRoleBinding := e2eframework.BuildUserWithGlobalRoleAndLogin(expect, "alice", "admin", dependencyManager)
}
```
There is still the only native login supported that I'm not fan of. 

I think I can try to make a proposal of update it will be better than speaking.

internal/api/impl/v1/project/service.go
```diff
@@ -37,28 +37,39 @@ import (
...
+	dao                   project.DAO
+	folderDAO             folder.DAO
+	datasourceDAO         datasource.DAO
+	dashboardDAO          dashboard.DAO
+	roleDAO               role.DAO
+	roleBindingDAO        rolebinding.DAO
+	secretDAO             secret.DAO
+	variableDAO           variable.DAO
+	authz                 authorization.Authorization
+	isAuthorizationNative bool
```
celian-garcia: Make a IsNative() method in `` authorization.Authorization`` would simplify a bit

pkg/model/api/config/authorization.go
```diff
@@ -18,25 +18,102 @@ import (
...
 var (
-	defaultCacheInterval = 30 * time.Second
+	defaultCacheInterval                   = time.Second * 30
+	DefaultKubernetesAuthorizationAllowTTL = time.Minute * 5
+	DefaultKubernetesAuthorizationDenyTTL  = time.Second * 30
+	DefaultKubernetesAuthenticationTTL     = time.Minute * 2
 )
 
-type AuthorizationConfig struct {
+type KubernetesProvider struct {
```
celian-garcia: KubernetesAuthorizationProvider

pkg/model/api/config/authorization.go
```diff
@@ -18,25 +18,102 @@ import (
...
+}
+
+type AuthorizationProvider struct {
+	Kubernetes KubernetesProvider          `json:"kubernetes,omitzero" yaml:"kubernetes,omitempty"`
+	Native     NativeAuthorizationProvider `json:"native,omitzero" yaml:"native,omitempty"`
+}
+
+type AuthorizationConfig struct {
+	// DEPRECATED: use NativeAuthorizationProvider.CheckLatestUpdateInterval instead.
+	CheckLatestUpdateInterval common.Duration `json:"check_latest_update_interval,omitempty" yaml:"check_latest_update_interval,omitempty"`
```
celian-garcia, line 101: Personal opinion but I would break on my side. We're not yet a v1

pkg/model/api/config/authorization.go
```diff
@@ -18,25 +18,102 @@ import (
...
 
-type AuthorizationConfig struct {
+type KubernetesProvider struct {
+	Enable bool `json:"enable,omitempty" yaml:"enable,omitempty"`
+	// File path to a local kubeconfig file used for local testing/development. The current logged in
+	// user's bearer token will be used for both the backend and as the user being logged into Perses.
+	// The user should have "create" permissions for the `TokenReview` and `SubjectAccessReview`
+	// resources. If this parameter isn't available the pods service account token will be used. This
+	// parameter should not be set in production
+	Kubeconfig string `json:"kubeconfig,omitempty" yaml:"kubeconfig,omitempty"`
```
celian-garcia, line 36: How it works in production ? Where is the k8s URL field ? Not yet implemented ?

pkg/model/api/config/security.go
```diff
@@ -187,7 +187,16 @@ func (s *Security) Verify() error {
 	if s.EnableAuth && !s.Authentication.Providers.EnableNative &&
 		len(s.Authentication.Providers.OIDC) == 0 &&
 		len(s.Authentication.Providers.OAuth) == 0 {
-		return errors.New("impossible to enable auth if no provider is setup")
+		return errors.New("impossible to enable auth if no authentication provider is setup")
 	}
+
+	if s.EnableAuth && !s.Authorization.Provider.Kubernetes.Enable {
+		s.Authorization.Provider.Native.Enable = true
```
celian-garcia, line 195: I'd personally not set a default behavior. Better to have an error than a field that we don't know it's set for you.

internal/api/authorization/k8s/k8s.go
```diff
@@ -0,0 +1,507 @@
...
+
+// IsEnabled implements [Authorization]
+func (k *k8sImpl) IsEnabled() bool {
+	return true
+}
+
+// GetUser implements [Authorization]
+func (k *k8sImpl) GetUser(ctx echo.Context) (any, error) {
+	// Context can be nil when the function is called outside the request context.
+	// For example, the provisioning service is calling every service without any context.
```
celian-garcia, line 104: Is provisioning service calling GetUser ? I don't think so.
I find that there's a design issue here with the provisioning service if it really send some nil context requests. But this is something prior to this branch anyway.

internal/api/authorization/k8s/k8s.go
```diff
@@ -0,0 +1,507 @@
...
+	// Rather than checking if the user has access to the namespace, we check if the user has access
+	// to any of the perses scopes within the namespace, since namespaces which the user has access to
+	// but cannot view perses scopes are irrelevant
+	for _, k8sScope := range k8sScopesToCheck {
+		attributes := authorizer.AttributesRecord{
+			User:            user,
+			Verb:            string(k8sReadAction),
+			Namespace:       namespace,
+			APIGroup:        "perses.dev",
+			APIVersion:      "v1alpha1",
```
celian-garcia: To be set in config field

pkg/model/api/config/security.go
Nexucis: this is for compatibility with the previous configuration as we are deprecating some fields. This won't stay forever

internal/api/authorization/k8s/k8s.go
Nexucis: this is also for backwards compatibility. The provisioning is using the internal system and needs to bypass the authorisation system as you don't want to use an internal user.

Having a nil context is a way to bypass the authorisation system.

Nexucis:
I would advice to focus on the authentication part and not reviewing (again) the authorisation as we already many times on that part and yes there are few areas where it could be improved, but I can assure you @celian-garcia this has been carefully thought and designed.

pkg/model/api/config/authorization.go
PeterYurkovich: In production the expected behavior is to use the pods service account rather than the kubeconfig field. A warning is displayed when being used, and the parameter should only be used in development, as the token provided in the kubeconfig file is injected into requests if no Authorization header is provided. This way the frontend is able to run locally without knowledge of the kubeconfig file or needing to inject the authorization header.

pkg/model/api/config/authorization.go
PeterYurkovich: Changed in the other PR

internal/api/impl/v1/project/service.go
PeterYurkovich: Added the IsNative function on the other PR

internal/api/authorization/k8s/k8s.go
PeterYurkovich: Are you referring to the v1alpha1 APIVersion here?

pkg/model/api/config/authorization.go
celian-garcia: okay so 
- In dev. You use the kubeconfig to build and inject a fake Authorization header
- In prod. You will use the sa but for what exactly? In prod you will always have the header here. 
  You will need more a URL right ?

PeterYurkovich: There are two different actors that need to be authenticated for this setup. The user sending a request to the perses backend, and the perses backend itself. 

The perses backend needs to be authenticated to the kubernetes cluster and have the following permissions:
- `TokenReview` - Allows for the backend to check the incoming request's Authorization header and see if it is authenticated to kubernetes.
- `SubjectAccessReview` - Allows for the backend to check kubernetes RBAC permissions
- `namespaces` - Allows for the backend to list namespaces so that each users permissions can be checked for each namespace

When running in prod the Service Account information is loaded and used for the Perses backend.

As far as the URL goes, any kubeconfig (including the one that is loaded from the Service Account) will include the URL to the api server for the cluster, so we don't need to specify it.

internal/api/authorization/k8s/k8s.go
PeterYurkovich: I think this is a problem to be solved more in a future feature which looks to integrate the kubernetes resources in tighter with the perses backend, such as potentially using kubernetes resources as the perses dataplane ([discussion](https://github.com/perses/perses-operator/discussions/178)). There is a chicken and egg problem where the perses resources are defined in the perses backend codebase, which is then pulled into the perses-operator codebase and converted into the kubernetes CRD's, who's version is then checked by a deployed perses backend.

I think since there is currently only a single version, that it would be best to wait to address this in the future when we have a better grasp of the interactions that will exist, rather than introducing a config variable now only to deprecate it later.

pkg/model/api/config/authorization.go
celian-garcia: oh but then it means kube library knows automagically everything already. It know it is in a pod started with a service account? No need to explicitely configure anything in Perses ?

PeterYurkovich: Thats right, no need for any further configuration

ui/app/src/views/auth/RequireAuth.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * This component aims to redirect the user to the SignIn page if not logged in
+ * or to the delegated authn error page if there is an error with the delegated authn.
+ * Otherwise, it just loads the underlying component(s) defined as children.
+ * This is leveraging the following mechanism:
+ * https://reactrouter.com/en/main/upgrading/v5#refactor-custom-routes
+ * https://gist.github.com/mjackson/d54b40a094277b7afdd6b81f51a0393f
+ * @param children
+ * @constructor
+ */
+function RequireAuth(): ReactElement | null {
```
Nexucis: this thing should not have been moved to the views folder as it has nothing to do with the views. This thing is only used for the routing and should have stayed there.

ui/app/src/model/route.ts
```diff
@@ -14,6 +14,7 @@
 export const AdminRoute = '/admin';
 export const SignInRoute = '/sign-in';
 export const SignUpRoute = '/sign-up';
+export const DelegatedAuthnErrorRoute = '/delegated-auth-error';
```
Nexucis, line 17: when a new route is added, then it should be added in `ui/endpoint.go` as well.

Nexucis:
I am going to merge this PR. Any last comments, issues ? @celian-garcia @PeterYurkovich @jgbernalp ?

PR #3566, opened 2025-11-10
[IGNORE] bump go deps and cue version to v0.15.0

Nexucis:


cue/model/api/v1/common/reflect_go_gen.cue
```diff
@@ -0,0 +1,5 @@
+// Code generated by cue get go. DO NOT EDIT.
+
+//cue:generate cue get go github.com/perses/perses/pkg/model/api/v1/common
+
+package common
```
Nexucis, line 5: mmm cue v0.15.0 is generating this empty file, that does not look right :/

AntoineThebaud: I reported the issue to CUE team, they suspect that's indeed a regression, they're looking at it. I guess we can still move forward with this PR and get these empty files cleaned up later, as they are not harmful.

AntoineThebaud (approved):

Issue #3573, opened 2025-11-11
Dynamic height search list could improve user exprience


shahrokni:
### What did you do?

It seems Search List occupies too much space and conceals the home page content. Probably This could be improved by setting the CSS max_height property. 

<img width="2535" height="1227" alt="Image" src="https://github.com/user-attachments/assets/66f625e8-e392-458f-9d2c-b88744761f33" />

shahrokni:
@prakhar29jain May I ask yo to take care of this issue?

prakhar29jain:
Sure!

PR #3575, opened 2025-11-11
[ENHANCEMENT] Add Dynamic display for Search Bar (#3573)

prakhar29jain:
# Description

closes (#3573)

- The Search List used to take up a lot of space even though there were fewer search results. 
- The Search List had no error message if no resources were found.

# Solution

- Kept rendering of SearchList gated on the absence of errors.
- Updated the modal container to size dynamically with results, capped at 70% of the viewport height.
- Show a modal message of No Resources found

# Screenshots

- Dynamic Modal:

image

image

- Display if none of the resources are found:

image

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/Header/SearchBar/SearchBar.tsx
+++ b/ui/app/src/components/Header/SearchBar/SearchBar.tsx
@@ -13,6 +13,7 @@
 
 import { Alert, Box, Button, Chip, InputAdornment, Modal, Paper, TextField, Typography } from '@mui/material';
 import Magnify from 'mdi-material-ui/Magnify';
+import EmoticonSadOutline from 'mdi-material-ui/EmoticonSadOutline';
 import ViewDashboardIcon from 'mdi-material-ui/ViewDashboard';
 import Archive from 'mdi-material-ui/Archive';
 import DatabaseIcon from 'mdi-material-ui/Database';
@@ -33,15 +34,24 @@ function shortcutCTRL(): string {
   return isAppleDevice() ? '⌘' : 'ctrl';
 }
 
+type ResourceType = 'dashboards' | 'projects' | 'globalDatasources' | 'datasources';
+
 interface ResourceListProps {
   query: string;
   onClick: () => void;
+  isResources?: (type: ResourceType, available: boolean) => void;
 }
 
 function SearchProjectList(props: ResourceListProps): ReactElement {
   const projectsQueryResult = useProjectList({ refetchOnMount: false });
   return (
-    <SearchList list={projectsQueryResult.data ?? []} query={props.query} onClick={props.onClick} icon={Archive} />
+    <SearchList
+      list={projectsQueryResult.data ?? []}
+      query={props.query}
+      onClick={props.onClick}
+      icon={Archive}
+      isResource={(isAvailable) => props.isResources?.('projects', isAvailable)}
+    />
   );
 }
 
@@ -54,6 +64,7 @@ function SearchGlobalDatasource(props: ResourceListProps): ReactElement {
       onClick={props.onClick}
       icon={DatabaseIcon}
       buildRouting={() => `${AdminRoute}/datasources`}
+      isResource={(isAvailable) => props.isResources?.('globalDatasources', isAvailable)}
     />
   );
 }
@@ -100,7 +111,14 @@ function SearchDashboardList(props: ResourceListProps): ReactElement | null {
     );
 
   return dashboardListLoading || importantDashboardsLoading ? null : (
-    <SearchList list={list} query={props.query} onClick={props.onClick} icon={ViewDashboardIcon} chip={true} />
+    <SearchList
+      list={list}
+      query={props.query}
+      onClick={props.onClick}
+      icon={ViewDashboardIcon}
+      chip={true}
+      isResource={(isAvailable) => props.isResources?.('dashboards', isAvailable)}
+    />
   );
 }
 
@@ -116,6 +134,7 @@ function SearchDatasourceList(props: ResourceListProps): ReactElement | null {
       buildRouting={(resource) =>
         `${ProjectRoute}/${isProjectMetadata(resource.metadata) ? resource.metadata.project : ''}/datasources`
       }
+      isResource={(isAvailable) => props.isResources?.('datasources', isAvailable)}
     />
   );
 }
@@ -148,10 +167,18 @@ export function SearchBar(): ReactElement {
   const isMobileSize = useIsMobileSize();
   const [open, setOpen] = useState(false);
   const [query, setQuery] = useState('');
+  const [hasResource, setHasResource] = useState<Record<ResourceType, boolean>>({
+    dashboards: false,
+    projects: false,
+    globalDatasources: false,
+    datasources: false,
+  });
+  function handleIsResourceAvailable(type: ResourceType, available: boolean): void {
+    setHasResource((prev) => (prev[type] === available ? prev : { ...prev, [type]: available }));
+  }
   const handleOpen = (): void => setOpen(true);
   const handleClose = (): void => setOpen(false);
   useHandleShortCut(handleOpen);
-
   return (
     <Paper sx={{ width: '100%', flexShrink: 1 }}>
       <Button size="small" fullWidth sx={{ display: 'flex', justifyContent: 'space-between' }} onClick={handleOpen}>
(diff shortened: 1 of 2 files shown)


shahrokni (commented):
When the result set is empty, instead of showing an exception (The red banner), we should simply show a center-aligned `No record found`. Now the exact message could be discussed with the team to be the most appropriate one.

In my opinion though, `No record found` is good enough.  

Showing an expiation is not a valid idea, because technically nothing went wrong here. Simply the set is empty. If you take a look at the Grafana for instance, they show a sad mascot saying nothing is available or something like that. Now, we don't need to make it fancy. Instead, we can simply show the mentioned text.  

Any suggestion for the message would be appreciated. @Gladorme and @AntoineThebaud

prakhar29jain:
> In my opinion though, `No record found` is good enough.

Well Yeah I agree this is enough but lets wait for the rest of the Member's suggestions

ui/app/src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -33,27 +33,53 @@ function shortcutCTRL(): string {
...
+  return (
+    <Box sx={{ margin: 1 }}>
+      <Alert severity="error">
+        <p>Failed to load {title}.</p>
+        {messages.map((message, id) => (
+          <p key={id}>{message}</p>
+        ))}
+      </Alert>
+    </Box>
+  );
```
shahrokni: There is already an open PR for Error Handling. That PR is supposed to comprehensively cover all cases if I am not mistaken. I'd say, let's remove this and wait for that one.

ui/app/src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -33,27 +33,53 @@ function shortcutCTRL(): string {
...
 }
 
 function SearchProjectList(props: ResourceListProps): ReactElement {
-  const projectsQueryResult = useProjectList({ refetchOnMount: false });
+  const { data, error } = useProjectList({ refetchOnMount: false });
+  if (error) return <SearchBarError title="projects" errors={[error]} />;
```
shahrokni: https://github.com/perses/perses/pull/3575/files#r2517578206

ui/app/src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -88,34 +114,37 @@ function SearchDashboardList(props: ResourceListProps): ReactElement | null {
...
-          <p>Failed to load dashboards! Error:</p>
-          {importantDashboardsError?.message && <p>{importantDashboardsError.message}</p>}
-          {dashboardListError?.message && <p>{dashboardListError.message}</p>}
-        </Alert>
-      </Box>
-    );
+  if (dashboardListError || importantDashboardsError) {
+    const errors: StatusError[] = [dashboardListError, importantDashboardsError].filter(Boolean) as StatusError[];
+    return <SearchBarError title="dashboards" errors={errors} />;
+  }
```
shahrokni: https://github.com/perses/perses/pull/3575/files#r2517578206

ui/app/src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -88,34 +114,37 @@ function SearchDashboardList(props: ResourceListProps): ReactElement | null {
...
-      list={datasourceQueryResult.data ?? []}
+      list={data ?? []}
       query={props.query}
       onClick={props.onClick}
       icon={DatabaseIcon}
       chip={true}
       buildRouting={(resource) =>
         `${ProjectRoute}/${isProjectMetadata(resource.metadata) ? resource.metadata.project : ''}/datasources`
       }
+      isResource={(isAvailable) => props.isResources?.('datasources', isAvailable)}
```
shahrokni: This is new. What is this?

ui/app/src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -209,10 +249,15 @@ export function SearchBar(): ReactElement {
...
-          <SearchDatasourceList query={query} onClick={handleClose} />
+          {!!query.length && !Object.values(hasResource).some((v) => v) && (
+            <Box sx={{ margin: 1 }}>
+              <Alert severity="error">No results found for Resource {query}.</Alert>
+            </Box>
+          )}
+          <SearchDashboardList query={query} onClick={handleClose} isResources={handleIsResources} />
+          <SearchProjectList query={query} onClick={handleClose} isResources={handleIsResources} />
+          <SearchGlobalDatasource query={query} onClick={handleClose} isResources={handleIsResources} />
+          <SearchDatasourceList query={query} onClick={handleClose} isResources={handleIsResources} />
```
shahrokni: Zero is false by default in JS. So, you could have simply said: 
```typescript
/* No need to have double exclamation mark */ 
!query.length
``` 

Another thing is that logically, it is not possible to have both the `No result` and the lists together. 
 SearchDashboardList , SearchProjectList, SearchGlobalDatasource , and SearchDatasourceList  
 could only be rendered if at least one of them has data.  If you see the right behavior now is only because those components do not render anything. However, I still believe the code could be improved for the sake of clean-code.
 I suggest you have a conditional statement like this here. 
 
```typescript
!query.length && !Object.values(hasResource).some((v) => v) ? <This /> : <OrThat />
````

File: ui/app/src/components/Header/SearchBar/SearchBar.tsx [hunk 6]
```diff
@@ -148,6 +177,15 @@ export function SearchBar(): ReactElement {
...
   const [query, setQuery] = useState('');
+  const [hasResource, setHasResource] = useState<Record<ResourceType, boolean>>({
+    dashboards: false,
+    projects: false,
+    globalDatasources: false,
+    datasources: false,
+  });
+  const handleIsResources = useCallback((type: ResourceType, available: boolean) => {
+    setHasResource((prev) => (prev[type] === available ? prev : { ...prev, [type]: available }));
+  }, []);
```
shahrokni: This function handles the a resource availability object correct? If this the assumption true then,  

- The function name is not helpful and does not suggest what it really does. Shouldn't it be handleIsResourceAvilable? 
- Having a useCallBack with an empty dependency array is OK yet not very common. Was this intentional and if the answer is Yes, may I ask why?

prakhar29jain: 1. I will rename the function to handleIsResourceAvilable
2. useCallback a was deliberate... because I was afraid that after every key stroke the state will change the the function will render 4 times for the 4 different resorce types so keeping it memoized avoids passing a new function to every SearchList render, please correct me if I am wrong.
The actual working of the function will not affect with its removal though... So I will remove it :)

ui/app/src/components/Header/SearchBar/SearchBar.tsx
prakhar29jain: Hmm Okay I browsed and found that then shall I remove all the Error Handling at the api level? and stick to only if search results are not there?

ui/app/src/components/Header/SearchBar/SearchBar.tsx
shahrokni: > 1. I will rename the function to handleIsResourceAvilable
> 2. useCallback a was deliberate... because I was afraid that after every key stroke the state will change the the function will render 4 times for the 4 different resorce types so keeping it memoized avoids passing a new function to every SearchList render, please correct me if I am wrong.
>    The actual working of the function will not affect with its removal though... So I will remove it :)

A man with a plan! 👍🏼  Sounds reasonable and logical. Let me check one more time. Probably you are right

prakhar29jain:
@shahrokni I Tried a few times but seems like the Unit Tests are failing due to the Plugin Issue...
I do not have permissions to re run will check tomorrow...

AntoineThebaud:
> When the result set is empty, instead of showing an exception (The red banner), we should simply show a center-aligned `No record found`. Now the exact message could be discussed with the team to be the most appropriate one.
> 
> In my opinion though, `No record found` is good enough.
> 
> Showing an expiation is not a valid idea, because technically nothing went wrong here. Simply the set is empty. If you take a look at the Grafana for instance, they show a sad mascot saying nothing is available or something like that. Now, we don't need to make it fancy. Instead, we can simply show the mentioned text.
> 
> Any suggestion for the message would be appreciated. @Gladorme and @AntoineThebaud

I agree & I'm fine with the design we see in the last screenshots added :)

ui/app/src/components/Header/SearchBar/SearchBar.tsx
shahrokni: @prakhar29jain 
I see you changed this part. What I meant was something identical to the following block of code. 
Here, you either see the `no records` rendering-path or the result-rendering path. 
Would that make sense? 

```typescript
          {query.length && !Object.values(hasResource).some((v) => v) ? (
            <Box sx={{ margin: 1, display: 'flex', justifyContent: 'center', gap: 1 }}>
              <EmoticonSadOutline fontSize="medium" />
              <Typography>No records found for {query}</Typography>
            </Box>
          ) : (
            <>
              <SearchDashboardList query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
              <SearchProjectList query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
              <SearchGlobalDatasource query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
              <SearchDatasourceList query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
            </>
          )}
```

PR #3577, opened 2025-11-12
[FEATURE] Add notices to panel header

OrReuben:
# Description
Relates to https://github.com/perses/perses/issues/3582.

For the time series queries, we had the ability to pass down `notices` which would be used only within the `QuerySummaryTable` before.
This PR makes use of the same property, but displays it on all panels.
This feature was discussed with @AntoineThebaud, regarding a way to inform the users from within the plugins if their data was changed due to high count of series.

# Screenshots
image

# Checklist

## UI Changes


shahrokni:
Dear @OrReuben ,

I would like to thank you for the contribution. 
If I am not mistaken your commit messages are not following the convention, and for that, the pipeline is failing. May I ask you to take a look please? 

Also, if this has been already discussed with @AntoineThebaud, could we open a feature issue and put the details there? 
If it already exists may I ask you to link it to your PR?

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -129,6 +130,30 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
     }
   }, [queryResults]);
 
+  const noticesIndicator = useMemo(() => {
+    const notices = queryResults.flatMap((q) => {
+      return q.data?.metadata?.notices ?? [];
+    });
+
+    if (notices.length > 0) {
+      const lastNotice = notices[notices.length - 1]!;
```
shahrokni, line 145: No need to compare with gt 0. Checking any none zero length would be enough. 

```typescript
 if (notices.length)
```

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -129,6 +130,30 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
...
+          <HeaderIconButton aria-label="panel notices" size="small">
+            {lastNotice.type === 'warning' ? (
+              <AlertIcon fontSize="inherit" color="warning" />
+            ) : lastNotice.type === 'info' ? (
+              <InformationOutlineIcon fontSize="inherit" color="info" />
+            ) : (
+              <AlertCircleIcon color="error" />
+            )}
+          </HeaderIconButton>
+        </InfoTooltip>
```
shahrokni, line 152: Please avoid using the nested ternary operators. 

## Reason

Nested ternary operators are generally considered a suboptimal coding practice because they significantly reduce code readability and maintainability. While the ternary operator is designed as a concise substitute for simple if-else statements, nesting them multiple times (e.g., a ? b : c ? d : e) quickly creates a complex and confusing structure that is difficult for developers to parse visually and logically. This ambiguity can lead to errors, as the operator precedence might not be immediately clear to someone other than the original author. For this reason, using clear, multi-line if-else if-else statements or a switch statement is strongly recommended when dealing with multiple conditions, prioritizing clarity over brevity to ensure the code remains understandable and easier to debug.

ui/dashboards/src/components/Panel/PanelActions.tsx
```diff
@@ -129,6 +130,30 @@ export const PanelActions: React.FC<PanelActionsProps> = ({
...
+    const notices = queryResults.flatMap((q) => {
+      return q.data?.metadata?.notices ?? [];
+    });
+
+    if (notices.length > 0) {
+      const lastNotice = notices[notices.length - 1]!;
+
+      return (
+        <InfoTooltip description={lastNotice.message}>
+          <HeaderIconButton aria-label="panel notices" size="small">
```
shahrokni, line 149: If I am not mistaken the `notices` are coming from the loose part of the schema. Am I right? 

```typescript
/* The loose schema */
export interface SOMEMETADATA extends BaseMetadata {
  [key: string]: unknown;
}
```

If this is the case, it is better to make sure that the message really exists and it is indeed a string.

ui/dashboards/src/components/Panel/PanelActions.tsx
OrReuben: I am also against it, saw it was a pattern used in this repo therefore did it alike.

I believe a good approach to disable it completely is to add an eslint rule for it.

ui/dashboards/src/components/Panel/PanelActions.tsx
OrReuben: BaseMetadata consists of the notices.

ui/dashboards/src/components/Panel/PanelActions.tsx
OrReuben: hmm i don't know if i totally agree with it, while both are common practices and behave the same, i believe `notices.length > 0` makes it more readable.

what do you think?

OrReuben:
@shahrokni since we are using plugins, the idea is to just provide the user the ability to display messages on top of the panel in order to inform the user.

For my use-case, i need to inform the user about a large number of series (which breaks the UI completely), but since it's a plugin, i believe it's up for the consumer that is using the plugin to decide whats his hard cap should be.

In my repo, for the promql time series query, i just added a hard cap and incase the hard cap exceeds i append a notice.

ui/dashboards/src/components/Panel/PanelActions.tsx
shahrokni: Hmm, ok,
I am not sure why my linter plugin is showing the following exception locally. Do all of them extend the base interface ❓ 

> Property 'notices' does not exist on type 'TimeSeriesMetadata | TraceMetaData | ProfileMetaData | LogMetadata'.
  Property 'notices' does not exist on type 'TraceMetaData'.

ui/dashboards/src/components/Panel/PanelActions.tsx
shahrokni: > I am also against it, saw it was a pattern used in this repo therefore did it alike.
> 
> I believe a good approach to disable it completely is to add an eslint rule for it.

Thank you for pointing that out. Yes I was thinking about the same thing. 
So, I suggest we fix this one now, and then I will take care of adding it to the ESLINT config. 
Sounds good?

AntoineThebaud:
> Also, if this has been already discussed with @AntoineThebaud, could we open a feature issue and put the details there?
If it already exists may I ask you to link it to your PR?

I will take care of this, it's likely we'll have other PRs related to the same topic - which is about defining a series limit for data rendering.

jgbernalp:
great addition! LGTM.

OrReuben:
@AntoineThebaud I believe we will need to also have an ability to pass a config from the chart plugin to the query plugin.

As for this change, we can display notices only from within he query plugin, so if we would want to have some kind of limit based on the panel type we can't really do that atm.

Issue #3550, opened 2025-11-08
Add OIDC RP-Initiated Logout Support for Complete Session Termination


samtonyclarke:
### Summary
Currently, when users click "Logout" in Perses, only the local Perses session is terminated while the OIDC provider session remains active. This allows users to immediately re-authenticate without entering credentials, which may not be desired in security-sensitive environments.

**Current Behavior**
User clicks "Logout" in Perses UI
Perses clears JWT cookies (/api/auth/logout)
User is redirected to home page
OIDC provider session remains active
Clicking "Sign in with OIDC" immediately logs user back in without credential prompt

**Proposal**
Provide an option to perform OIDC RP-Initiated Logout that:

Clears Perses session (current behavior)
Redirects to OIDC provider's end_session_endpoint
Terminates both Perses and OIDC provider sessions
Ensures users must re-enter credentials on next login

Add logout_url ocnfiguration support to OIDC providers, similar to what's documented for OAuth providers:
```
security:
  authentication:
    providers:
      oidc:
        - slug_id: "cognito"
          name: "AWS Cognito"
          issuer: "https://cognito-idp.region.amazonaws.com/pool-id"
          logout_url: "https://cognito-domain.auth.region.amazoncognito.com/logout"  # New field
          # ... other config
```

**Use Cases**
High-security environments requiring complete session termination
Shared workstations where users want to ensure full logout
Compliance requirements mandating complete authentication cleanup
Organizations wanting to prevent immediate re-authentication bypass

Would this be something worth considering as a feature request?

Nexucis:
At a first glance I would say it makes sense yes.
@celian-garcia what do you think ? You are more knowledgable on this area than me

celian-garcia:
Hello, absolutely! I implemented without logout first because OIDC logout seemed to me a bit more complex than just an http request. What we would like to do is to remove the link perses <-> provider but not necessarily completely log you out of the OP if you have another tool connected to it.

I think everything is there 
https://openid.net/specs/openid-connect-session-1_0.html
https://openid.net/specs/openid-connect-rpinitiated-1_0.html

So yes but don't take it easy, it's a big deal that might imply some refactoring.

samtonyclarke:
@celian-garcia Understood. When you say "but not necessarily completely log you out of the OP if you have another tool connected to it". My understanding is that the RP cannot know if another RP has an active session with the OP for that user. The OP can offer the user the option if they want to log out of all apps or not. Its interesting to node that the openid spec says:

[...]

PR #3587, opened 2025-11-13
[FEATURE] Implement OIDC RP initiated logout

celian-garcia:
# Description
This PR allows a user to logout of OIDC provider when logs out of Perses. 
It will still remove the Perses session, but additionally redirects to the logout page of the provider. 

The logout URL is retrieved from the discovery endpoint ``/.well-known/openid-configuration`` of the issuer, using ``end_session_endpoint`` discovery field, as specified in the OIDC specs https://openid.net/specs/openid-connect-rpinitiated-1_0.html#OPMetadata

This feature is optional and disabled by default to not introduce breaking change. 
It can be activated easily in the OIDC provider configuration with ``logout.enabled = true``

Note also that this adds two new claims in the Perses token.
- ``pkd`` is the provider kind or "auth kind" (native, oauth, or oidc)
- ``pid`` is the provider id or "slug id" used to identify the provider 

In the logout implementation if any provider match the given claim and defined a logout implementation, this one is called in favor of the the classic '/' redirection.

relates #3550 (will need some specific to really fix the issue as AWS has a special discovery field)

# Screenshots

# Checklist

## UI Changes
No UI changes

### DIFF
--- a/docs/concepts/authentication.md
+++ b/docs/concepts/authentication.md
@@ -52,14 +52,16 @@ security:
   authentication:
     providers:
       oidc:
-        # Example with an Azure AD OIDC configuration
-        - slug_id: azure
-          name: "Azure AD"
+        # Example with a Microsoft Entra ID OIDC configuration
+        - slug_id: msft
+          name: "Microsoft"
           client_id: "<secret>"
           client_secret: "<secret>"
           issuer: "https://login.microsoftonline.com/<tenant-id>/v2.0"
           redirect_uri: "http://localhost:8080/api/auth/providers/oidc/azure/callback"
           scopes: [ "openid", "profile", "email", "User.read" ]
+          logout:
+            enabled: true
       oauth:
         - slug_id: github
           name: "Github"
@@ -82,7 +84,7 @@ sequenceDiagram
     participant rp as Perses Backend
     participant op as External Identity Provider
     
-    hu->>br: Login with OIDC provider (e.g Azure AD)
+    hu->>br: Login with OIDC provider (e.g Microsoft Entra ID)
     activate br
     br->>rp: GET /api/auth/providers/{oidc|oauth}/{slug_id}/login
     activate rp
--- a/docs/configuration/oauth-configuration-helpers.md
+++ b/docs/configuration/oauth-configuration-helpers.md
@@ -30,21 +30,23 @@ there.
 > Disclaimer: We try to keep this page up-to-date, but the provider's documentation is always the most reliable source
 > of information.
 
-#### Azure AD
+#### Microsoft Entra ID (ex Azure AD)
 
 ```yaml
 security:
   authentication:
     providers:
       oidc:
-        - slug_id: azure
-          name: "Azure AD"
+        - slug_id: msft
+          name: "Microsoft"
           client_id: "<your client ID>"
           client_secret: "<your client Secret>"
           issuer: "https://login.microsoftonline.com/<your tenant ID>/v2.0"
           scopes: [ "openid", "profile", "email", "User.read" ] # For Auth code / Device code
           client_credentials:
             scopes: [ "https://graph.microsoft.com/.default" ] # For Client Credentials
+          logout:
+            enabled: true # Generally advised, but you can disable it if you don't want to redirect to the provider's logout page
 ```
 
 !!! tip
@@ -53,6 +55,14 @@ security:
 ```
 *Ref: [https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow)*
 ```
+
+##### Logout redirection with Entra ID
+
+Logout redirection is partially supported with Entra ID. When you click on logout, Perses will redirect you to the Entra logout URL.
+But there can be some situations where Entra will not redirect you back to Perses.
+- The user logged out directly from Entra ID.
+- The user session expired.
+In this case, you will be logged out from Perses, but you will have to manually go back to Perses.
 
 #### <Place Your Provider here ...\>
 


ui/package-lock.json
```diff
@@ -18290,4 +18290,4 @@
       }
     }
   }
-}
```
celian-garcia, line 18332: Remove that change

internal/api/impl/auth/auth.go
```diff
@@ -170,6 +182,19 @@ func (e *endpoint) logout(ctx echo.Context) error {
 	ctx.SetCookie(jwtHeaderPayloadCookie)
 	ctx.SetCookie(signatureCookie)
 
+	//TODO: More control ?
+	token, _ := ctx.Get("user").(*jwt.Token)
+	providerKind := token.Claims.(*crypto.JWTClaims).ProviderKind
+	providerID := token.Claims.(*crypto.JWTClaims).ProviderID
```
celian-garcia: There should be a common way to get the JWT claims.

internal/api/impl/auth/auth.go
```diff
@@ -129,7 +141,7 @@ func (e *endpoint) CollectRoutes(g *route.Group) {
 		ep.CollectRoutes(providersGroup)
 	}
 	g.POST(fmt.Sprintf("/%s/%s", utils.PathAuth, utils.PathRefresh), e.refresh, true)
-	g.GET(fmt.Sprintf("/%s/%s", utils.PathAuth, utils.PathLogout), e.logout, true)
+	g.GET(fmt.Sprintf("/%s/%s", utils.PathAuth, utils.PathLogout), e.logout, false)
```
celian-garcia, line 146: Because of this change if you call the logout endpoint without being logged in, it will show an "invalid jwt token" message. This should not happen anyway and hard to fix without a refactoring of the way we save token in the echo context.

internal/api/impl/auth/oidc.go
```diff
@@ -198,6 +201,28 @@ func (e *oIDCEndpoint) CollectRoutes(g *route.Group) {
 	oidcGroup.POST(fmt.Sprintf("/%s", utils.PathToken), e.token, true)
 }
 
+func (e *oIDCEndpoint) GetLogoutHandler() echo.HandlerFunc {
+	if !e.logout.Enabled {
+		return nil
+	}
+	return func(ctx echo.Context) error {
+		endSessionURL, _ := url.Parse(e.relyingParty.GetEndSessionEndpoint())
```
celian-garcia: manage error

internal/api/impl/auth/auth.go
PeterYurkovich: The `authz authorization.Authorization` returns the token.Claims, but that only really works for the first line. One of your suggestions for k8s was the addition of an `IsNative` function on it. Perhaps that `IsNative` could be adjusted to a `ProviderInfo(ctx echo.Context)` function instead which is able to extract the providerKind and providerId (if applicable).

internal/api/impl/auth/auth.go
```diff
@@ -170,6 +182,19 @@ func (e *endpoint) logout(ctx echo.Context) error {
...
 
+	//TODO: More control ?
+	token, _ := ctx.Get("user").(*jwt.Token)
+	providerKind := token.Claims.(*crypto.JWTClaims).ProviderKind
+	providerID := token.Claims.(*crypto.JWTClaims).ProviderID
+
+	for _, ep := range e.endpoints {
+		if ep.GetAuthKind() == providerKind && ep.GetSlugID() == providerID {
+			if logoutHandler := ep.GetLogoutHandler(); logoutHandler != nil {
+				return logoutHandler(ctx)
```
PeterYurkovich, line 196: I think we should either rename logoutHandler to be something to the extent of `ProviderLogout` or maybe look to move the actual logout code into those providers and remove the cookie deletion from the `func (e *endpoint) logout(ctx echo.Context)` function. The LogoutHandler sounds like it should be handling all of the logout behavior, but isn't. A further benefit of this is letting each provider dictate how it logs our rather than needing to have each one follow one specific path, but this flexibility isn't really needed at this time so may be just preemtive feature adding

```go
func NativeLogout(ctx echo.Context) error {
	jwtHeaderPayloadCookie, signatureCookie := e.jwt.DeleteAccessTokenCookie()
	ctx.SetCookie(e.jwt.DeleteRefreshTokenCookie())
	ctx.SetCookie(jwtHeaderPayloadCookie)
	ctx.SetCookie(signatureCookie)
	return nil
}
```

which could then be returned from the GetLogoutHandler' 

```go
func (e *nativeEndpoint) GetLogoutHandler() echo.HandlerFunc {
	return NativeLogout
}
```

or called in the oidc logout:

```go
func (e *oIDCEndpoint) GetLogoutHandler() echo.HandlerFunc {
	if !e.logout.Enabled {
		return nil
	}
	return func(ctx echo.Context) error {
		err := NativeLogout()
		if err != nil {
			return err
		}
		endSessionURL, _ := url.Parse(e.relyingParty.GetEndSessionEndpoint())
		queryParams := endSessionURL.Query()
		rd := getRootURL(ctx.Request())
		queryParams.Add("post_logout_redirect_uri", rd.String())
		endSessionURL.RawQuery = queryParams.Encode()
		return ctx.Redirect(302, endSessionURL.String())
	}
}
```

celian-garcia: The thing is the cookie deletion is not native logout. This is a Perses session logout. It should be called for all the providers. 
For k8s external token management the token won't be managed by perses. I don't consider it an Auth provider. So the logout won't be managed by Perses neither. We'll never call logout of a the Perses session to logout from k8s

PeterYurkovich: Right, that's why I had thought maybe changing the variable to be ProviderLogout rather than LogoutHandler might make sense. But it's fine as is

internal/api/impl/auth/auth.go
celian-garcia: I added a ``GetProviderInfo``, the problem is that it can't be as static as the IsNative is because there can be several providers activated at the same time, but like that it's working well. Maybe this is what you had in mind 🙂  I let you review and tell me what you think

internal/api/impl/auth/auth.go
celian-garcia: I renamed to ``GetExtraProviderLogout`` wdyt ?

internal/api/impl/auth/auth.go
Nexucis: why it should not happen ? Calling the logout endpoint supposed you are actually log in and so you have a token. So I don't see why this is an issue to not being able to call this endpoint if you don't have a token

internal/api/impl/auth/oauth.go
```diff
@@ -140,7 +140,19 @@ type oAuthEndpoint struct {
 	loginProps      []string
 }
 
-func newOAuthEndpoint(provider config.OAuthProvider, jwt crypto.JWT, dao user.DAO, authz authorization.Authorization) (route.Endpoint, error) {
+func (e *oAuthEndpoint) GetExtraProviderLogout() echo.HandlerFunc {
+	return nil // No specific logout handler for native auth
```
Nexucis: this is not native auth here

pkg/model/api/config/authentication.go
```diff
@@ -104,12 +104,17 @@ func (p *Provider) Verify() error {
...
+	Enabled bool `json:"enabled" yaml:"enabled"`
+}
+
 type OIDCProvider struct {
 	Provider     `json:",inline" yaml:",inline"`
 	Issuer       common.URL        `json:"issuer" yaml:"issuer"`
 	DiscoveryURL common.URL        `json:"discovery_url,omitempty" yaml:"discovery_url,omitempty"`
 	URLParams    map[string]string `json:"url_params,omitempty" yaml:"url_params,omitempty"`
 	DisablePKCE  bool              `json:"disable_pkce" yaml:"disable_pkce"`
+	Logout       OIDCLogout        `json:"logout" yaml:"logout"`
```
Nexucis, line 117: I would simplify this by just

```go
Logout boolean `json:"logout" yaml:"logout"`
```

celian-garcia: I followed what has been proposed in the discussion. https://github.com/perses/perses/issues/3550#issuecomment-3518569015

To allow later some other parameters

Nexucis: mmm ok, do you plan to support the field `post_logout_redirect_uri_query_parameter_name` in a next PR ?

internal/api/impl/auth/auth.go
Nexucis: nether mind, the change was miss-leading me, but this is a fix actually. Thank you for changing it!

pkg/model/api/config/authentication.go
celian-garcia: Not necessarily me but yes. It can be a good first issue once that PR is done 👼🏻 
I'm not attached enough to make it myself, and we don't have Amazon instance to test.

internal/api/impl/auth/auth.go
Nexucis: I would name it `GetExtraProviderLogoutHandler`. Or `BuildExtraProviderLogoutHandler` because it's more a generation rather than just getting a value.

Nexucis: and having as a suffix `handler` will make it clear it will return an `handler`.

celian-garcia: You make me think of something quite interesting that this should not be built at a runtime. This is getting a function that could be built in the OIDC provider's constructor. I will modify. And yes I will add handler suffix

celian-garcia: It doesn't make a big difference performance wised as the function is declared statically and passed. It will make a diff probably only on the if on top

celian-garcia: I just pushed a new proposal of implementation. Tell me WDYT ?

pkg/model/api/config/authentication.go
Nexucis: alright let keeps it that way then.

PR #3610, opened 2025-11-21
[IGNORE] Update CODEOWNERS

AntoineThebaud:


.github/CODEOWNERS
```diff
@@ -1,5 +1,9 @@
+/cue-test @AntoineThebaud
 /cue @AntoineThebaud
+/docs @AntoineThebaud
 /go-sdk @Gladorme
+/pkg/client @Nexucis @celian-garcia
```
Nexucis: can you sort it by alphabetic order please ?

.github/CODEOWNERS
```diff
@@ -1,5 +1,9 @@
+/cue-test @AntoineThebaud
 /cue @AntoineThebaud
+/docs @AntoineThebaud
```
Nexucis, line 3: I have hesitated to do that, and then I thought the doc is the responsibility of everyone. But I am fine with you being owner as well

AntoineThebaud: indeed, actually it doesnt prevent anyone to jump in, I just thought it makes sense by default that changes in the doc ping me in particular because I often look at that part

PR #3615, opened 2025-11-24
[IGNORE] update CODEOWNERS

AntoineThebaud:
# Description

More scope assigned to @celian-garcia as discussed.

# Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,9 +2,10 @@
 /cue @AntoineThebaud
 /docs @AntoineThebaud
 /go-sdk @Gladorme
-/internal @Nexucis @celian-garcia
+/internal/api @Nexucis @celian-garcia
 /internal/api/plugin/migrate @Nexucis @AntoineThebaud
 /internal/api/plugin/schema @Nexucis @AntoineThebaud
-/pkg @Nexucis @celian-garcia
+/internal/cli @Nexucis
+/pkg/client @Nexucis @celian-garcia
 /pkg/model @Nexucis @AntoineThebaud
 /ui @Gladorme @jgbernalp @shahrokni


.github/CODEOWNERS
```diff
@@ -2,10 +2,9 @@
 /cue @AntoineThebaud
 /docs @AntoineThebaud
 /go-sdk @Gladorme
-/internal @Nexucis
-/internal/api/impl/auth @Nexucis @celian-garcia
+/internal @Nexucis @celian-garcia
 /internal/api/plugin/migrate @Nexucis @AntoineThebaud
 /internal/api/plugin/schema @Nexucis @AntoineThebaud
-/pkg/client @Nexucis @celian-garcia
+/pkg @Nexucis @celian-garcia
```
Nexucis: I don't think it will work as intended. Meaning that  for `/pkg/model` Celian will be owner as there is a dedicated rule for that.

And since there is only two sub package for `/pkg`, that means Celian is actually owner of `pkg/client`

AntoineThebaud: ah indeed nvm, reverting this one

AntoineThebaud: but well even if it would have no impact currently, isnt it better to have people defined at `pkg` level to enforce some kind of default? Otherwise if ever we add a 3rd package to this section in the future no one will own it by default.

Nexucis: well this is fine there is some folder with no owner. For me owner file define just some specific area where we need specific people to approve it because not everybody can do it.

From my point of view, the `pkg` folder is quite simple to be reviewed by anyone (maybe excepting some specific part of `pkg/client)

AntoineThebaud: Ok your point is valid, I'll revert to the previous state then.

AntoineThebaud:
(hopefully final) adjustment made, @celian-garcia you still get more scope but @Nexucis remains the main POC for the CLI part. WDYT guys?

Nexucis (approved):

PR #3642, opened 2025-11-26
Table: Prevent table column duplication when multiple settings defined

Snehith-706:
hi developers!
this pr fixes an issue where table columns were duplicated when multiple settings were defined in the configuration
this duplication occured due to repeated column mapping during render.

### DIFF
--- a/ui/components/src/Table/Table.tsx
+++ b/ui/components/src/Table/Table.tsx
@@ -144,7 +144,19 @@ export function Table<TableData>({
   }, [theme.palette.text.primary, density, getCheckboxColor, handleCheckboxChange]);
 
   const tableColumns: Array<ColumnDef<TableData>> = useMemo(() => {
-    const initTableColumns = persesColumnsToTanstackColumns(columns);
+    const initTableColumns = 78
+      (columns);
+        // Fix for issue #3639: Deduplicate columns to prevent column duplication when multiple
+    // column settings are defined for the same column
+    const columnIds = new Set<string>();
+    initTableColumns = initTableColumns.filter((col) => {
+      if (col.id && columnIds.has(col.id)) {
+        return false; // Skip duplicate columns
+      }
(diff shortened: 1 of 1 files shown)


ui/components/src/Table/Table.tsx
```diff
@@ -144,7 +144,19 @@ export function Table<TableData>({
   }, [theme.palette.text.primary, density, getCheckboxColor, handleCheckboxChange]);
 
   const tableColumns: Array<ColumnDef<TableData>> = useMemo(() => {
-    const initTableColumns = persesColumnsToTanstackColumns(columns);
```
Gladorme, line 147: Well, it will never work if you remove that 😅

ui/components/src/Table/Table.tsx
```diff
@@ -144,7 +144,19 @@ export function Table<TableData>({
...
+    const initTableColumns = 78
+      (columns);
```
Gladorme, line 148: wtf 😆

ui/components/src/Table/Table.tsx
```diff
@@ -144,7 +144,19 @@ export function Table<TableData>({
   }, [theme.palette.text.primary, density, getCheckboxColor, handleCheckboxChange]);
 
   const tableColumns: Array<ColumnDef<TableData>> = useMemo(() => {
-    const initTableColumns = persesColumnsToTanstackColumns(columns);
+    const initTableColumns = 78
```
jgbernalp, line 147: This is probably a typo,  but the code is not linted or checked for types as this is a const and cannot be re assigned. Please be mindful about the contribution guidelines.

Nexucis:
Since the issue is not an issue, we can close this PR.

Thank you for helping us @Snehith-706 for making Perses greater and sorry there was a miss-understood about wether we should fix the related issue or not.

Issue #3643, opened 2025-11-26
Search bar doesn't return any result after hitting "no records found" once, unless you fully clear the input.


AntoineThebaud:
### What did you do?

![Image](https://github.com/user-attachments/assets/dd4e74bc-fa93-4aa6-b37d-b74a3c50f6f0)

### What did you expect to see?

I should see results when my search matches entries, no matter if I hit "no records" before

### Perses version

0.53.0-beta.3

### Perses configuration file

```yaml

```

### Logs

```text

```

Issue #3638, opened 2025-11-25
Search bar doesn't return any result when no important dashboards are configured


AntoineThebaud:
### What did you do?

<img width="1904" height="666" alt="Image" src="https://github.com/user-attachments/assets/239eeb9a-a06c-4c92-8f09-f072173f8ead" />

I guess it's a side effect of https://github.com/perses/perses/pull/3572

### Perses version

0.53.0-beta.3

Nexucis:
would be great to have more tests around this functionality ...

PR #3644, opened 2025-11-26
[BUGFIX] Fix search bar wrongly returning no results

shahrokni:
Closes https://github.com/perses/perses/issues/3638
Closes https://github.com/perses/perses/issues/3643

# Description 🖊️ 
If no important project is added into the config, the search returns no results. This is due to a conditional short-circuit which doesn't let a proper initialization happen. Details can be found here => https://github.com/perses/perses/pull/3644#discussion_r2564745954

## Test  Case 1 🧪 

Remove the important projects from config. Run the app, and check if you can search any projects

image

## Tesst Case 2 🧪 

This test is similar to the first test case but with important projects added to the config. (**You have to reset the server**)

image

image

## Test Case 3 #3643  🧪 

1. Start typing a valid project name 
2. You should see the results in the Search Bar
3. Add some improper trailing characters to see the _no data found_ message
4. Remove the trailing characters
5. You should see the results again

# Checklist

## UI Changes

### DIFF
--- a/ui/app/src/components/Header/SearchBar/SearchBar.tsx
+++ b/ui/app/src/components/Header/SearchBar/SearchBar.tsx
@@ -44,27 +44,29 @@ interface ResourceListProps {
 
 function SearchProjectList(props: ResourceListProps): ReactElement {
   const projectsQueryResult = useProjectList({ refetchOnMount: false });
+  const { query, onClick, isResources } = props;
   return (
     <SearchList
       list={projectsQueryResult.data ?? []}
-      query={props.query}
-      onClick={props.onClick}
+      query={query}
+      onClick={onClick}
       icon={Archive}
-      isResource={(isAvailable) => props.isResources?.('projects', isAvailable)}
+      isResource={(isAvailable) => isResources?.('projects', isAvailable)}
     />
   );
 }
 
 function SearchGlobalDatasource(props: ResourceListProps): ReactElement {
   const globalDatasourceQueryResult = useGlobalDatasourceList({ refetchOnMount: false });
+  const { query, onClick, isResources } = props;
   return (
     <SearchList
       list={globalDatasourceQueryResult.data ?? []}
-      query={props.query}
-      onClick={props.onClick}
+      query={query}
+      onClick={onClick}
       icon={DatabaseIcon}
       buildRouting={() => `${AdminRoute}/datasources`}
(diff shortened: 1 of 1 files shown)


ui/app/src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -238,19 +245,18 @@ export function SearchBar(): ReactElement {
...
             </Box>
-          ) : (
-            <>
-              <SearchDashboardList query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
-              <SearchProjectList query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
-              <SearchGlobalDatasource query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
-              <SearchDatasourceList query={query} onClick={handleClose} isResources={handleIsResourceAvailable} />
-            </>
           )}
+          <>
```
jgbernalp: The fragment is not needed anymore. But still don't get why this change as the old logic seems correct. We should not trigger search if the query is empty. Not sure why it should be related with important dashboards.

shahrokni: > The fragment is not needed anymore. But still don't get why this change as the old logic seems correct. We should not trigger search if the query is empty. Not sure why it should be related with important dashboards.

## Clarification 🦆 

>⚠️ Not judging the approach, just explaining!

## The early return by the condition ❗ 

Lets start with the nature of the condition itself. 
The condition tries to show **no data found** instead of results, if the hasResource  object has no true value. 

```typescript
query.length && !Object.values(hasResource).some((v) => v)
```

## What is hasResource  ❓ 

hasResource  is an object with the keys referring to different resources (SearchDashboardList, SearchProjectList etc ) and the relevant Boolean values. The Boolean value determines whether the resource would return any results. This means, to show the Search Result, there should be at least one **true** value.

## What does this mean ❓ 

This means to avoid an early return of the first mentioned condition, and initial check is necessary. (To set the object values accordingly) This mechanism simply tries to answer, are there any results to be rendered? (**Even the very first time**)

## How come it doesn't happen when important dashboards available  ❓ 

Simple! Because the important dashboard in the config always returns value as soon as you click on the Search Bar. This guarantees a set of results even when the query is empty. This way the mentioned condition is bypassed and no early return happens. 

### How to check this? 🧪 

1. Click on the Search Bar
2. If a list of important dashboard exists you should see them immediately (Picture 2 from the top)

PR #3646, opened 2025-11-27
[ignore] ignore testdata folder in snyk

Nexucis:
It will avoid PR like https://github.com/perses/perses/pull/3640 to be opened


Nexucis:
moving forward on this as the change is simple and does not do anything on the code itself, it's just repo management.

Issue #1538, opened 2023-10-25
Prometheus: Datasource form: header disappears if same header key is typed


AntoineThebaud:
When configuring a PrometheusDatasource, if you already provided a header, start adding another one & type the same header key, the former item with same key will get removed.
This is happening because we are updating the state at each input, and headers are stored in a map.

Should we use an intermediary array instead? And implement a logic to decide which item to keep in case of conflict when saving. More details about the proposed technical solution here https://github.com/perses/perses/pull/1282#discussion_r1265437000.

Issue #3014, opened 2025-06-12
All official plugins: add form validation


AntoineThebaud:
### Proposal

As mentioned [here](https://github.com/perses/perses/issues/1348#issuecomment-2194100793), we have proper validation on the frontend side when filling forms to create panels, datasources & variables... but for the "core" part of their model only. The plugin part is not implemented which causes several issues:
- No clear display: when required fields are not yet provided, the "Save" button of drawer forms can still be clicked, and there's no indication about what's wrong before trying to save*.
- Even if the save action gets properly denied by the backend validation, the error message it returns comes from CUE in that case and can be quite cryptic, e.g:
![Image](https://github.com/user-attachments/assets/19bc664f-419a-42db-a1d3-ad69ab9b9dc4)

\*contrary to what we have for "core" model fields: ![Image](https://github.com/user-attachments/assets/cb8bf50e-1c0e-4e6b-b36e-6573fb41541b)

As stated in the referenced comment, the implementation consist in injecting the plugin validation schemas (based on zod) in the ValidationProvider.

AntoineThebaud:
Side note: sometimes users are even able to save technically-invalid resources (like no "Label name" provided for a PrometheusLabelValuesVariable currently) but this is due to the backend validation not being strict enough (in this case, empty string is accepted while it shouldn't). Such cases are then bugs on the backend side thus out of scope of this issue (I'll try to do a pass on these asap)

PR #3650, opened 2025-11-28
[BUGFIX] Datasource editor: Resolve request headers UX issues caused by direct map manipulation

AntoineThebaud:
# Description

Closes https://github.com/perses/perses/issues/1538.

I used the same logic as done in [this PR](https://github.com/perses/plugins/pull/485): The Request Headers in HTTPSettingsEditor now rely on an intermediate array state instead of direct map manipulation. This avoids weird UX that we had previously like:
- Duplicated keys getting removed instantly (as soon as you were typing the last character causing the duplicate).
- Entries reordering automatically while you were still editing.

Entries are now only synced to the parent map on save, preserving user input during editing.

# Screenshots

Screenshot 2025-11-28 101322

NB: you're still allowed to save when there are duplicates, in this case the last key-value takes precedence. This still needs to be improved with proper zod-based validation in the future, to prevent the Save action in such case (ref https://github.com/perses/perses/issues/3014).

# Checklist

## UI Changes

### DIFF
--- a/ui/plugin-system/src/components/HTTPSettingsEditor/HTTPSettingsEditor.tsx
+++ b/ui/plugin-system/src/components/HTTPSettingsEditor/HTTPSettingsEditor.tsx
@@ -67,6 +67,8 @@ export function HTTPSettingsEditor(props: HTTPSettingsEditor): ReactElement {
   const watchedHeaders = headersForm.watch('headers');
 
   // Check for duplicate header names
+  // TODO: duplication detection logic to be replaced by proper zod schema validation in the future
+  // ref https://github.com/perses/perses/issues/3014
   const nameMap = new Map<string, number>();
   const duplicateNames = new Set<string>();
   watchedHeaders.forEach(({ name }) => {


ui/plugin-system/src/components/HTTPSettingsEditor/HTTPSettingsEditor.tsx
```diff
@@ -38,25 +38,77 @@ export function HTTPSettingsEditor(props: HTTPSettingsEditor): ReactElement {
...
-  ): RequestHeaders | undefined => {
-    if (oldHeaders === undefined) return oldHeaders;
-    const keys = Object.keys(oldHeaders);
-    const newHeaders = keys.reduce<Record<string, string>>((acc, val) => {
-      if (val === oldName) {
-        acc[newName] = oldHeaders[oldName] || '';
-      } else {
-        acc[val] = oldHeaders[val] || '';
+  type HeaderEntry = {
+    // Unique identifier for the entry, added to avoid using array index as key
```
Gladorme: For arrays in form, you can use: https://react-hook-form.com/docs/usefieldarray

AntoineThebaud: done, see the latest commit https://github.com/perses/perses/pull/3650/commits/9fa23444ff15b5011917be30130b9eac4f58267b

ui/plugin-system/src/components/HTTPSettingsEditor/HTTPSettingsEditor.tsx
```diff
@@ -38,26 +47,61 @@ export function HTTPSettingsEditor(props: HTTPSettingsEditor): ReactElement {
...
-      return acc;
-    }, {});
+    }
+  });
+  const hasDuplicates = duplicateNames.size > 0;
 
-    return { ...newHeaders };
-  };
+  // Sync headers to parent when they change
+  useEffect(() => {
```
Gladorme: I will check tomorrow, but hopefully we can remove this useEffect 😄

AntoineThebaud: the alternative is to do like before the last commit (have this logic executed when the onChange of a header input is triggered), I don't know which one is best?

AntoineThebaud: in https://github.com/perses/perses/pull/3650/commits/1e731d1ce85b7ebec2896ad366b2b351463d8d34 I went back to soemthing onChange-based as agreed

PR #3658, opened 2025-12-03
[BUGFIX] Fix plugin path in Dockerfile

runiq:
# Description

The dev Dockerfile downloads the plugins, but it doesn't use the download directory as a COPY source. This commit fixes that.

I am not sure if the prod Dockerfile should behave like this as well, since it is built in CI. Thoughts?

# Checklist

## UI Changes

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -8,10 +8,10 @@ FROM golang:1.25 AS go-builder
 WORKDIR /go/src/github.com/perses/perses
 COPY . .
 COPY --from=node-builder /app/ui/app/dist ui/app/dist
-RUN mkdir /perses
-RUN mkdir /plugins
-RUN make build-api
-RUN make build-cli
+RUN mkdir /perses \
+  && mkdir /plugins \
+  && make build-api \
+  && make build-cli
 
 FROM gcr.io/distroless/static-debian12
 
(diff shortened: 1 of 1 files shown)


Dockerfile.dev
```diff
@@ -22,7 +23,7 @@ USER nobody
 COPY --from=go-builder --chown=nobody:nobody  /go/src/github.com/perses/perses/bin/perses        /bin/perses
 COPY --from=go-builder --chown=nobody:nobody  /go/src/github.com/perses/perses/bin/percli        /bin/percli
 COPY --chown=nobody:nobody                    LICENSE            /LICENSE
-COPY --chown=nobody:nobody                    plugins-archive/   /etc/perses/plugins-archive/
+COPY --chown=nobody:nobody                    /go/src/github.com/perses/perses/plugins-archive/   /etc/perses/plugins-archive/
```
Nexucis: this path cannot be absolute as it depends on your environment.

runiq: Argh, it seems I accidentally removed the important part of the PR. Fixed now.

Dockerfile.dev
```diff
@@ -8,10 +8,11 @@ FROM golang:1.25 AS go-builder
 WORKDIR /go/src/github.com/perses/perses
 COPY . .
 COPY --from=node-builder /app/ui/app/dist ui/app/dist
-RUN mkdir /perses
-RUN mkdir /plugins
-RUN make build-api
-RUN make build-cli
+RUN <<EOR
```
Nexucis:
```suggestion
RUN mkdir /perses \
        && mkdir /plugins \
        && make build-ap \
        && make build-cli
```

This is the usual syntax to chain bash command.

Dockerfile.dev
Nexucis: ah yeah that makes more sense. Thank you for fixing it :)

Dockerfile.dev
runiq: Done.

runiq:
Hmm, not sure about those e2e tests -- looks like the Prometheus demo instance couldn't be reached. Spurious?

Nexucis:
yeah this is not related to your PR.
It seems we are still relying on the prometheus demo instance which seems to be down currently.

PR #3659, opened 2025-12-04
[BREAKINGCHANGE] remove shared packages

jgbernalp:
# Description

This PR 

- Removes the packages that can be added to the new shared repo.
- Adds a new command to start the Perses UI app `npm run start:shared` using the link with the shared libraries.
- Fixes the package.lock file to remove local workspaces.
- Adds ignore files used by the script that links dependencies, present in the shared repo.

The counterpart PR on the shared repo is: https://github.com/perses/shared/pull/4

# Checklist

## UI Changes

No UI changes, as only dependencies where moved.


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@perses-dev/components": "file:../shared/components"
```
Nexucis: did you add this file on purpose ?

I feel this should be removed.

jgbernalp: garbage from early tests. Will delete.

Nexucis:
I think you should update the README: https://github.com/perses/perses/blob/main/ui/README.md to hightlight the new repository `shared` and where it should be located (aka alongside than the perses repo in your local path).

And maybe put a link somewhere to point to the README of the `shared` repository in case you need more details.

Nexucis (approved):
overall looks good to me. Thank you for tackling this big issue @jgbernalp.

We might have a circular dependency issue when we will need to update the `core` package. But it mights be ok.

AntoineThebaud (approved):
works like a charm! GG

Nexucis:
once the shared package are released and updated here, we can merge this PR

PR #3661, opened 2025-12-05
[BUGFIX] `format` data model: fix ambiguous schema

AntoineThebaud:
# Description

The way the format model was updated with this change https://github.com/perses/perses/pull/3035 was actually wrong as `#format` couldn't be refered to without precising a unit, as we expected with this change.

E.g with the following code:

```cue
myFormat: #format & {
    decimalPlaces: 0
    shortValues:   false
}
```

You'd get this error:
```sh
myFormat: incomplete value {decimalPlaces:0,shortValues:false,unit?:"decimal"} | {decimalPlaces:0,shortValues:false,unit?:"bits" | "decbits"} | {decimalPlaces:0,shortValues:false,unit?:"bytes" | "decbytes"} | {decimalPlaces:0,shortValues:false,unit?:"bits/sec" | "decbits/sec" | "bytes/sec" | "decbytes/sec" | "counts/sec" | "events/sec" | "messages/sec" | "ops/sec" | "packets/sec" | "reads/sec" | "records/sec" | "requests/sec" | "rows/sec" | "writes/sec"}
```

Even though the `unit` attribute is optional, the `#format` disjunction still has to resolve to a single option (struct) and it couldnt.

This PR fixes this problem, by grouping the formats that have the same attributes together, so that the evaluation resolves to a single struct.

The "old" format definitions like `#decimalFormat`, `#bytesFormat` etc are still preserved to avoid breaking change (especially for the CUE SDK users). The `unit` field in those is back as a mandatory attribute, as when referencing these precise formats directly (CUE SDK use case) it doesn't really make sense to have unit optional.

# Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,6 +28,10 @@ jobs:
           enable_npm: true
           enable_go: true
           nvmrc_path: "./ui/.nvmrc"
+      - name: start dev environment
+        uses: hoverkraft-tech/compose-action@v2.4.2
+        with:
+            compose-file: ./dev/docker-compose.e2e.yaml
       - name: Install UI deps
         working-directory: ./ui
         run: npm ci
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -155,7 +155,7 @@ jobs:
       - name: generate files
         run: make assets-compress
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v9.1.0
+        uses: golangci/golangci-lint-action@v9.2.0
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: v2.4.0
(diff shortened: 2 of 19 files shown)


scripts/validate-cue/validate-cue.go
```diff
@@ -106,7 +106,7 @@ func validateCueFiles() error {
 			}
 
 			if err := runCueVet(schemaFile, testFile); err != nil {
-				return err
+				logrus.Errorf("Validation failed for %s: %v", schemaFile, err)
```
Nexucis, line 110: if you don't return an error here, it will not make the CI failed

AntoineThebaud: ah yes I went a bit too fast sorry.. I want to have a full error report, not fail on the first file, but yes I need to make the function return an error if at least 1 file was invalid

AntoineThebaud: fixed in https://github.com/perses/perses/pull/3661/commits/9310b665e46ce3bdbe0cdc23c0459af0e89767fb

PR #3663, opened 2025-12-05
[BUGFIX] Fix loading plugin in dev with no schema

Nexucis:
When loading a plugin in development that does not require schema, previously the plugin was not load at all without any error.

This was a mistake. This PR aims to change that so plugin without schema can still be loaded in dev environment.

This error didn't exist in the regular flow.

### DIFF
--- a/internal/api/plugin/dev.go
+++ b/internal/api/plugin/dev.go
@@ -54,15 +54,13 @@ func (p *pluginFile) LoadDevPlugin(plugins []v1.PluginInDevelopment) error {
 			},
 			Module: pluginModule,
 		}
-		if !IsSchemaRequired(pluginModule.Spec) || plg.DisableSchema {
-			logrus.Debugf("schema is disabled or not required for plugin %q", pluginModule.Metadata.Name)
-			return nil
-		}
-		if pluginSchemaLoadErr := p.sch.LoadDevPlugin(plg.AbsolutePath, pluginModule); pluginSchemaLoadErr != nil {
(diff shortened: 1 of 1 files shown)


internal/api/plugin/dev.go
```diff
@@ -54,15 +54,13 @@ func (p *pluginFile) LoadDevPlugin(plugins []v1.PluginInDevelopment) error {
...
-		if !IsSchemaRequired(pluginModule.Spec) || plg.DisableSchema {
-			logrus.Debugf("schema is disabled or not required for plugin %q", pluginModule.Metadata.Name)
-			return nil
-		}
-		if pluginSchemaLoadErr := p.sch.LoadDevPlugin(plg.AbsolutePath, pluginModule); pluginSchemaLoadErr != nil {
-			return apiinterface.HandleBadRequestError(fmt.Sprintf("failed to load plugin schema: %s", pluginSchemaLoadErr))
-		}
-		if pluginMigrateLoadErr := p.mig.LoadDevPlugin(plg.AbsolutePath, pluginModule); pluginMigrateLoadErr != nil {
-			return apiinterface.HandleBadRequestError(fmt.Sprintf("failed to load plugin migration: %s", pluginMigrateLoadErr))
+		if IsSchemaRequired(pluginModule.Spec) && !plg.DisableSchema {
```
celian-garcia, line 57: What I read is that the plugin will still not be loaded if the schema is not required

celian-garcia: Mmh ok no it is the schema. Got it. Approved

celian-garcia: Mmh ok no it is the schema. Got it. Approved

PR #3670, opened 2025-12-07
[ENHANCEMENT] New home view layout

nicolastakashi:
# Description

I'm proposing a new structure for the home view.

# Screenshots

image

image

# Checklist

## UI Changes

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,11 +27,11 @@ jobs:
         with:
           enable_npm: true
           enable_go: true
-          nvmrc_path: "./ui/.nvmrc"
+          nvmrc_path: './ui/.nvmrc'
       - name: start dev environment
         uses: hoverkraft-tech/compose-action@v2.5.0
         with:
-            compose-file: ./dev/docker-compose.e2e.yaml
+          compose-file: ./dev/docker-compose.e2e.yaml
       - name: Install UI deps
         working-directory: ./ui
         run: npm ci
@@ -43,7 +43,7 @@ jobs:
         run: npx playwright install --with-deps chromium
       - name: Run Playwright tests
         working-directory: ./ui
-        timeout-minutes: 10
+        timeout-minutes: 15
         run: npm run e2e:ci
         env:
           # Outputs information from webservers started in the background by
--- a/ui/e2e/src/pages/DashboardPage.ts
+++ b/ui/e2e/src/pages/DashboardPage.ts
@@ -407,7 +407,9 @@ export class DashboardPage {
   }
 
   async goBackToProjectPage(projectName: string): Promise<void> {
-    const navigationPromise = this.page.waitForURL(`/projects/${encodeURIComponent(projectName)}`);
+    const navigationPromise = this.page.waitForURL(`/projects/${encodeURIComponent(projectName)}`, {
+      waitUntil: 'domcontentloaded',
+    });
     await this.page.getByRole('link', { name: projectName }).first().click();
     await navigationPromise;
   }
--- a/ui/e2e/src/tests/projectView.spec.ts
+++ b/ui/e2e/src/tests/projectView.spec.ts
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { expect, test } from '@playwright/test';
-import { AppProjectPage, DashboardPage } from '../pages';
+import { AppProjectPage } from '../pages';
 
 test.describe('ProjectView', () => {
   const project = 'perses';
@@ -31,9 +31,6 @@ test.describe('ProjectView', () => {
     await projectPage.goto(project);
 
     await projectPage.navigateToDashboard(project, 'Demo');
-
-    const dashboardPage = new DashboardPage(page);
-    await dashboardPage.goBackToProjectPage(project);
 
     await projectPage.goto(project);
     await projectPage.navigateToDashboardFromRecentDashboards(project, 'Demo');
--- a/ui/e2e/src/tests/variables.spec.ts
+++ b/ui/e2e/src/tests/variables.spec.ts
@@ -20,8 +20,7 @@ test.use({
 
 test.describe('Dashboard: Variables', () => {
   test('can add simple text variable', async ({ dashboardPage }) => {
-    const initialCount = await dashboardPage.variableList.count();
-    // const initialCount = await dashboardPage.variableList.getByTestId(/variable-.+/).count(); // TODO (gladorme): this test is wrong, it should based this locator
+    const initialCount = await dashboardPage.variableList.getByTestId(/variable-.+/).count();
 
     await dashboardPage.startEditing();
     await dashboardPage.startEditingVariables();
@@ -41,14 +40,12 @@ test.describe('Dashboard: Variables', () => {
     await variableEditor.applyChanges();
     await dashboardPage.saveChanges();
 
-    await expect(dashboardPage.variableList).toHaveCount(initialCount + 1);
-    // await expect(dashboardPage.variableList.getByTestId(/variable-.+/)).toHaveCount(initialCount + 1); // TODO (gladorme): this test is wrong, it should based this locator
+    await expect(dashboardPage.variableList.getByTestId(/variable-.+/)).toHaveCount(initialCount + 1);
     await expect(dashboardPage.variableList).toContainText([/Text Var/]);
   });
 
   test('can add simple list variable', async ({ dashboardPage }) => {
-    const initialCount = await dashboardPage.variableList.count(); // Builtin variables are hidden
-    // const initialCount = await dashboardPage.variableList.getByTestId(/variable-.+/).count();  // TODO (gladorme): this test is wrong, it should based this locator
+    const initialCount = await dashboardPage.variableList.getByTestId(/variable-.+/).count();
     await dashboardPage.startEditing();
     await dashboardPage.startEditingVariables();
     const variableEditor = dashboardPage.getVariableEditor();
@@ -68,8 +65,7 @@ test.describe('Dashboard: Variables', () => {
     await variableEditor.applyChanges();
     await dashboardPage.saveChanges();
 
-    await expect(dashboardPage.variableList).toHaveCount(initialCount + 1);
-    // await expect(dashboardPage.variableList.getByTestId(/variable-.+/)).toHaveCount(initialCount + 1); // TODO (gladorme): this test is wrong, it should based this locator
+    await expect(dashboardPage.variableList.getByTestId(/variable-.+/)).toHaveCount(initialCount + 1);
     await expect(dashboardPage.variableList).toContainText([/List Var/]);
   });
 });


nicolastakashi:
image
Also have this option without card on the hero section.

Gladorme:
image

Some ideas / improvements:

The information card has been removed, but we can keep it by replacing welcome text by it (and show  welcome text if not set). Creation buttons should be kept on top right, if the info card change, button will still be on the same place.

Search bar size should be the same as recently view dashboard.

Lot of empty space for project card, I think name should be at the same level as the icon. And the link button looks to take of place too :/

Nexucis:
> Creation buttons should be kept on top right, if the info card change, button will still be on the same place.

I kind of like this new design. Why not having a default info text and then it can be replaced by the info text from the config ?

Gladorme:
> Why not having a default info text and then it can be replaced by the info text from the config ?

It's what I wrote with:

> The information card has been removed, but we can keep it by replacing welcome text by it (and show welcome text if not set)

Nexucis:
yeah but I think also the buttons can be kept at the same place than what Nicolas is proposing. So it's not exactly what you are proposing

Gladorme:
I want to keep on the right for coherence. On all other pages, creation button are on the top right, it will use empty space and if user change the content of the info, it will still be on the same place

Gladorme:
Another point raised by @AntoineThebaud is to test with more data (important dashboards, etc) to see potential issues.

For example, should we limit the number of important dashboard displayed or add a scrollbar? Same for recently dashboard. Because from what I see, there no max heights defined and it can take a lot of space

ui/app/src/views/home/ImportantDashboards.tsx
```diff
@@ -11,63 +11,119 @@
...
-    <Grid container spacing={isMobileSize ? 1 : 2} data-testid="important-dashboards-mosaic">
-      {dashboards.map((dashboard) => (
-        <Grid key={`${dashboard.metadata.project}-${dashboard.metadata.name}`} item xs={6} lg={4}>
-          <DashboardCard dashboard={dashboard} hideIcon={isMobileSize}></DashboardCard>
-        </Grid>
-      ))}
-    </Grid>
-  );
-}
+const MAX_IMPORTANT_DASHBOARDS = 3;
```
AntoineThebaud: please don't 😄

ui/app/src/views/home/ProjectsAndDashboards.tsx
```diff
@@ -169,56 +192,63 @@ export function SearchableDashboards(props: SearchableDashboardsProps): ReactEle
...
-        <FormatListBulletedIcon />
-        <h2>Projects & Dashboards</h2>
-      </Stack>
-      <SearchableDashboards id="project-dashboard-list" />
-    </Stack>
+    <Box component="section">
+      <Box sx={{ display: 'flex', alignItems: 'center', justifyContent: 'space-between', mb: 3 }}>
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
+          <Archive sx={{ color: 'primary.main' }} />
+          <Typography variant="h2">Projects & Dashboards</Typography>
```
AntoineThebaud: With your changes the dashboards are no longer listed on the home page (you have to go to the project page) so this section should now be named "Projects" only imho

ui/app/src/views/home/RecentDashboards.tsx
```diff
@@ -11,27 +11,106 @@
...
+          ) : dashboards.length === 0 ? (
+            <Typography variant="body2" color="text.secondary">
+              No dashboards viewed yet.
+            </Typography>
+          ) : (
+            <Box sx={{ display: 'flex', flexDirection: 'column' }}>
+              {dashboards.map((item, index) => {
+                const updatedAt = item.date ?? item.dashboard.metadata.updatedAt;
+                const relativeTime = updatedAt ? intlFormatDistance(new Date(updatedAt), new Date()) : 'moments ago';
+                const displayName = item.dashboard.spec.display?.name ?? item.dashboard.metadata.name;
```
AntoineThebaud, line 83: Looks like there's a bug here, the display name doesnt get picked as expected. This is actually already an issue with the current home page (ref https://github.com/perses/perses/issues/3676) and it occurs here as well:
e.g 
image
versus 
image

there's sth about an API call only returning metadata instead of the full spec, that may be the cause here

AntoineThebaud: thinking about it again, a bit out of scope of this PR, your code is correct, the problem is on the hook side. to be tackled as part of https://github.com/perses/perses/issues/3676

nicolastakashi: Should I fix that on this PR or another one?

ui/app/src/views/home/ProjectsAndDashboards.tsx
nicolastakashi: Makes sense, I'll rename it.

nicolastakashi:
https://github.com/user-attachments/assets/038eb1c1-2bfb-4bc1-9c98-2f1c82742960

@AntoineThebaud @Gladorme let me know what do you think after the last few changes, addressing your comments 🙏🏽

ui/app/src/views/home/HomeViewHeroSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+          <Button
+            variant="outlined"
+            size="small"
+            startIcon={<UploadIcon />}
+            component={RouterLink}
+            to={ImportRoute}
+            disabled={userProjects.length === 0}
+            sx={{ px: 2 }}
+          >
+            Import Dashboard
```
ibakshay, line 158: How about this? Migrate dashboard is kinda hidden behind Import Dashboard otherwise. 
```suggestion
            Import/Migrate Dashboard
```

AntoineThebaud: migration happens as part of the process of importing a grafana dashboard into perses, the wording is ok like this then

ui/app/src/views/home/RecentDashboards.tsx
AntoineThebaud: I resolved the comment because it's out of scope of that PR ;)

Gladorme:
Cool, looks nice 😄 

image

Small detail, this icon is not well aligned vertically

One last remark about the layout is: what happen if there is no important dashboard defined 😅

ui/app/src/views/home/HomeViewHeroSection.tsx
nicolastakashi: I feel it breaks a bit the UI :(

nicolastakashi:
> Cool, looks nice 😄
> 
> image
> Small detail, this icon is not well aligned vertically
> 
> One last remark about the layout is: what happen if there is no important dashboard defined 😅

Good catch 😄 
I've adjusted the icon vertical alignment and also, I improve the empty state, by creating a empty state component for HomeView and reusing it on all the sections.

image

Nexucis:
I think there is an issue with the e2e tests here. It took ages while it should take 7min to run

Nexucis:
@Gladorme @shahrokni can one of you take a deeper look at this PR and help @nicolastakashi to fix the e2e tests one forever. It will be great to be able to merge this PR and I think we need your expertise to move one. Looking forward for your valuable input and fix.

Thank you by advance for your help !

ui/app/src/views/home/HomeViewHeroSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { ProjectResource } from '@perses-dev/core';
+import { CRUDButton } from '../../components/CRUDButton/CRUDButton';
+import { ImportRoute } from '../../model/route';
+import { useInformation } from '../../context/Config';
+
+interface HomeViewHeroSectionProps {
+  userProjects: ProjectResource[];
+  onAddProjectClick: () => void;
+  onAddDashboardClick: () => void;
+}
```
shahrokni, line 30: This is completely fine. I am just putting this for future contributions. 
I think the name of the interface could be as simple as IProps

ui/app/src/views/home/HomeViewHeroSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+        width: '100%',
+        borderRadius: 3,
+        px: { xs: 4, md: 6 },
+        py: { xs: 4, md: 6 },
+        background: (theme) => {
+          if (theme.palette.mode === 'dark') {
+            return `linear-gradient(135deg, ${alpha(theme.palette.primary.main, 0.1)} 0%, ${alpha(theme.palette.info?.main || theme.palette.primary.main, 0.05)} 100%)`;
+          }
+          return `linear-gradient(135deg, ${alpha(theme.palette.primary.main, 0.08)} 0%, ${alpha(theme.palette.info?.main || theme.palette.primary.main, 0.03)} 100%)`;
+        },
```
shahrokni, line 53: This is only a personal preference. All good in my opinion. I personally prefer ternary operator. Thought maybe you also find it a better option here

ui/app/src/views/home/HomeViewHeroSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+        }}
+      >
+        <Box sx={{ maxWidth: 800, flex: 1 }}>
+          {information ? (
+            <Box
+              sx={{
+                lineHeight: 1.5,
+                '& *': { marginTop: 0 },
+              }}
+              dangerouslySetInnerHTML={{ __html: information }}
```
shahrokni, line 105: This could cause a security breach. Could this be improved or replaced with a better structure somehow?

ui/app/src/views/home/HomeViewHeroSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+          >
+            Create Project
+          </CRUDButton>
+          <Button
+            variant="outlined"
+            size="small"
+            startIcon={<ViewDashboardOutlineIcon />}
+            onClick={onAddDashboardClick}
+            disabled={userProjects.length === 0}
+            sx={{ px: 2 }}
```
shahrokni, line 145: Suggestion, 

```Typescript
disabled={!userProjects?.length}
````

File: ui/app/src/views/home/HomeViewHeroSection.tsx [hunk 9]
```diff
@@ -0,0 +1,164 @@
...
+          >
+            Create Dashboard
+          </Button>
+          <Button
+            variant="outlined"
+            size="small"
+            startIcon={<UploadIcon />}
+            component={RouterLink}
+            to={ImportRoute}
+            disabled={userProjects.length === 0}
```
shahrokni, line 155: https://github.com/perses/perses/pull/3670/changes#r2736358322

ui/app/src/views/home/ImportantDashboards.tsx
```diff
@@ -11,65 +11,140 @@
...
 export function ImportantDashboards(): ReactElement {
-  const { t: translate } = useTranslation('dashboard');
   const { data: config } = useConfig();
   const { data: dashboards, isLoading } = useImportantDashboardList();
 
-  // If no important dashboard defined, hide the section
-  if (!config?.frontend.important_dashboards?.length || dashboards.length === 0) {
+  const dashboardList = useMemo(() => dashboards ?? [], [dashboards]);
+
+  const hasImportantDashboardsConfig = Boolean(config?.frontend.important_dashboards?.length);
```
shahrokni, line 30: Suggestion,
Wouldn't this work? 

```Typescript
 const hasImportantDashboardsConfig = !!config?.frontend.important_dashboards?.length;
```

ui/app/src/views/home/ImportantDashboards.tsx
```diff
@@ -11,65 +11,140 @@
...
+          </Typography>
+        </Box>
+      </CardContent>
+      <CardContent
+        sx={{
+          flex: '1 1 auto',
+          overflowY: 'auto',
+          pt: 0,
+          minHeight: 0,
+          ...(dashboardList.length === 0 && !isLoading
```
shahrokni, line 63: Suggestion,

```TypeScript
!dashboardList?.length
```

ui/app/src/views/home/ImportantDashboards.tsx
```diff
@@ -11,65 +11,140 @@
...
+            : {}),
+        }}
+      >
+        {isLoading && (
+          <Stack width="100%" sx={{ alignItems: 'center', justifyContent: 'center' }}>
+            <CircularProgress size={24} />
+          </Stack>
+        )}
+        {!isLoading && dashboardList.length === 0 && (
+          <EmptyState
```
shahrokni, line 78: https://github.com/perses/perses/pull/3670/changes#r2736370130

ui/app/src/views/home/ImportantDashboards.tsx
```diff
@@ -11,65 +11,140 @@
...
+            icon={<StarFourPointsOutline sx={{ fontSize: 32, color: 'text.secondary' }} />}
+            message="No important dashboards found."
+            hint="Configure important dashboards in your config file."
+          />
+        )}
+        {!isLoading && dashboardList.length > 0 && (
```
shahrokni, line 84: Suggestion,

In JavaScript, the comparison > 0 is not strictly necessary because of how [truthy and falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) work.

ui/app/src/views/home/Projects.tsx
```diff
@@ -0,0 +1,272 @@
...
+  }
+  const index = Math.abs(hash) % colors.length;
+  // Index is guaranteed to be within bounds due to modulo operation
+  const color = colors[index];
+  if (color) {
+    return color;
+  }
+  // Fallback to first color (should never happen, but satisfies TypeScript)
+  return colors[0]!;
+}
```
shahrokni, line 68: What do you think?

```Typescript
const PROJECT_COLORS = [
  '#3B82F6', '#8B5CF6', '#10B981', '#F97316', '#EC4899',
  '#06B6D4', '#6366F1', '#F43F5E', '#14B8A6', '#F59E0B',
] as const;

function getProjectColor(projectName: string): string {
  let hash = 0;
  
  for (let i = 0; i < projectName.length; i++) { 
    hash = projectName.charCodeAt(i) + ((hash << 5) - hash);
    hash |= 0; // Convert to 32bit integer
  }
 
  const index = (hash >>> 0) % PROJECT_COLORS.length;
  
  return PROJECT_COLORS[index];
}

```

ui/app/src/views/home/Projects.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+interface ProjectCardProps {
+  row: ProjectWithDashboards;
+}
```
shahrokni, line 72: Suggestion,
IProps would be better

ui/app/src/views/home/Projects.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+interface RenderProjectGridProps {
+  projectRows: ProjectWithDashboards[];
+  searchQuery: string;
+}
+
+function RenderProjectGrid(props: RenderProjectGridProps): ReactElement {
+  const { projectRows, searchQuery } = props;
+
+  if (projectRows.length === 0) {
```
shahrokni, line 155: Suggestion,
```Typescript
if(!projectRows.length)
```

ui/app/src/views/home/Projects.tsx
```diff
@@ -0,0 +1,272 @@
...
+          md: 'repeat(3, 1fr)',
+          lg: 'repeat(4, 1fr)',
+          xl: 'repeat(5, 1fr)',
+        },
+        gap: 1.5,
+      }}
+    >
+      {projectRows.map((row) => (
+        <ProjectCard key={row.project.metadata.name} row={row} />
+      ))}
```
shahrokni, line 182: I am not sure and need to check it myself.
Could we have two projects with the identical names? 
If yes the key would not be efficient.

ui/app/src/views/home/Projects.tsx
```diff
@@ -0,0 +1,272 @@
...
+    if (searchQuery) {
+      return kvSearch.filter(searchQuery, projectRows ?? []).map((res) => res.original);
+    } else {
+      return projectRows ?? [];
+    }
+  }, [kvSearch, projectRows, searchQuery]);
+
+  const handleSearch = useCallback((e: ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(e.target.value);
+  }, []);
```
shahrokni, line 216: A useCallback with an empty array is used in some specific scenarios. I am thinking, if this was really necessary? Is there something I am missing? Couldn't we handle the search inline? Would not cost that much right?

ui/app/src/views/home/RecentDashboards.tsx
```diff
@@ -11,27 +11,136 @@
...
 import { ErrorAlert, ErrorBoundary } from '@perses-dev/components';
-import { RecentDashboardList } from '../../components/DashboardList/RecentDashboardList';
+import { intlFormatDistance } from 'date-fns';
 import { useRecentDashboardList } from '../../model/dashboard-client';
+import { EmptyState } from '../../components/EmptyState/EmptyState';
 
 export function RecentDashboards(): ReactElement {
   const { data, isLoading } = useRecentDashboardList();
 
+  const dashboards = useMemo(() => data ?? [], [data]);
```
shahrokni, line 28: This looks a bit weird, 
Couldn't we do something like this?

```Typescript
 const { data, isLoading } = useRecentDashboardList();
  const dashboards = data ?? [];
```

ui/app/src/views/home/RecentDashboards.tsx
```diff
@@ -11,27 +11,136 @@
...
+        }}
+      >
         <ErrorBoundary FallbackComponent={ErrorAlert}>
-          <RecentDashboardList dashboardList={data} isLoading={isLoading} />
+          {isLoading && (
+            <Stack width="100%" sx={{ alignItems: 'center', justifyContent: 'center' }}>
+              <CircularProgress size={24} />
+            </Stack>
+          )}
+          {!isLoading && dashboards.length === 0 && (
```
shahrokni, line 71: Suggestion,

```Typescript
!projectRows.length
```

ui/app/src/views/home/RecentDashboards.tsx
```diff
@@ -11,27 +11,136 @@
...
+            <EmptyState
+              icon={<HistoryIcon sx={{ fontSize: 32, color: 'text.secondary' }} />}
+              message="No dashboards viewed yet."
+              hint="Your recently viewed dashboards will appear here."
+            />
+          )}
+          {!isLoading && dashboards.length > 0 && (
```
shahrokni, line 78: You don't need to compare the length with 0.

ui/app/src/views/home/HomeViewHeroSection.tsx
jgbernalp: This naming is aligned with the ui conventions: https://github.com/perses/perses/blob/main/ui/ui-guidelines.md#component-props

ui/app/src/views/home/HomeViewHeroSection.tsx
nicolastakashi: This is something working from the previous code, I'd prefer change it on a diff pr.

ui/app/src/views/home/ImportantDashboards.tsx
nicolastakashi: It could, but this is more complex to read, since you have two negations on the same expression.

ui/app/src/views/home/Projects.tsx
nicolastakashi: I don't thin k so, cuz this is the resource name and we're following k8s api spec, so resource name is unique.

ui/app/src/views/home/RecentDashboards.tsx
nicolastakashi: the current code is caching dashboards unless data is changed, on the proposal we are always triggering re render if I'm not mistaken.

shahrokni: I have some doubts that this improves the efficiency. Actually, I think it may have an adverse effect. Let me explain my idea. Since in JS no two objects, or arrays are equal (regardless of their internal value), if useRecentDashboardList is engaged (and if it is returning a new object or array every time), the useMemo will be engaged as well. So, it could be claimed, it will be executed anyway! After conferring with co-pilot, I got this answer: 

> You spent CPU cycles setting up a memoization that never actually saves a calculation

## Conclusion
Unless, useRecentDashboardList returns the same manipulated data object (which should not be the case)
the useMemo doesn't improve the performance. 

Would that make sense?

ui/app/src/views/home/Projects.tsx
jgbernalp: same here, this is aligned with our guidelines: https://github.com/perses/perses/blob/main/ui/ui-guidelines.md#component-props

Nexucis:
regardless the pending comments, I guess we will need another round of your magic @jgbernalp to fix (again) the e2e tests hopefully one forever.

jgbernalp:
I applied the e2e fixes from main, there still some flakiness unrelated to this PR that needs further investigation, but for now the review can continue.

ui/app/src/views/home/Projects.tsx
Nexucis: I think this is how it works in the current home page

Issue #3612, opened 2025-11-23
Unable to set up OIDC auth with Authelia


zhiayang:
### What did you do?

Attempted to set up OIDC with Authelia. While I can start the login flow (entering username/password in Authelia), I get the following error message afterwards:

```
invalid_state: The state is missing or does not have enough characters and is therefore considered too weak. Request parameter 'state' must be at least be 8 characters long to ensure sufficient entropy.
```

I don't see anything too interesting in the Perses logs (at trace level). I also don't see a `state` cookie set. This is the browser network log for the outgoing auth request:

```
https://auth.domain.com/api/oidc/authorization?client_id=perses&code_challenge=h0xfUo7VfJXlk51mE2Wdh4scx_sosZVxOJKBF29YS3o&code_challenge_method=S256&redirect_uri=https://perses.domain.com/api/auth/providers/oidc/authelia/callback&response_type=code&state=/
```

It indeed appears that `state` is just one character long. From what I can tell there's no way to tell Authelia to not enforce this requirement, so this needs to be changed in Perses.

Authelia client config:

```yaml
{
    client_id: "perses",
    client_name: "Perses",
    consent_mode: "implicit",
    public: false,
    authorization_policy: "one_factor",
    redirect_uris: [
        "https://perses.domain.com/api/auth/providers/oidc/authelia/callback"
    ],
    scopes: [ "openid", "profile", "email", "groups" ],
    client_secret: "xxxxx",
    userinfo_signed_response_alg: "none",
    token_endpoint_auth_method: "client_secret_basic",
}
```

### What did you expect to see?

The authentication flow should work correctly.

### System information

Docker on Linux, x86-64

### Perses version

0.53.0-beta.3

### Perses configuration file

```yaml
security:
  enable_auth: true
  encryption_key: "uwuwuuwuwuwuuwuwuw"
  authentication:
    disable_sign_up: true
    providers:
      enable_native: false
      oidc:
        - slug_id: authelia
          name: "Authelia"
          client_id: "perses"
          client_secret: "omomomomomom"

          issuer: "https://auth.domain.com"
          redirect_uri:   "https://perses.domain.com/api/auth/providers/oidc/authelia/callback"
```

### Logs

```text

perses            | time="2025-11-23T16:57:46+08:00" level=info method=GET status=200 uri=/
perses            | time="2025-11-23T16:57:46+08:00" level=info method=GET status=200 uri=/main.84b13ed57476bdd4.css
[...]

PR #3684, opened 2025-12-08
[ENHANCEMENT] Allow to configure login property from OIDC via ``custom_login_property``

celian-garcia:
# Description

We discussed about that in #3612. It's currently not possible to override the way we collect login when we use OIDC provider. 

The default behaviour is to get the first part of the mail, or take subject. But it means that in the case the provider doesn't return a mail, the result (sub claim) might be an id not very readable. 

With this PR, administrator can specify the field used to retrieve the login info.
The fields are taken from /userinfo endpoint result and id token claims.

# Screenshots

# Checklist

## UI Changes

N/A

### DIFF
--- a/internal/api/impl/auth/userinfo/oidc_test.go
+++ b/internal/api/impl/auth/userinfo/oidc_test.go
@@ -1,0 +1,209 @@
+// Copyright 2025 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package userinfo
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestOIDCUserInfoUnmarshalJSON tests the unmarshalling of OIDCUserInfo from JSON.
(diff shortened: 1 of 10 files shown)


internal/api/impl/auth/oauth_test.go
```diff
@@ -0,0 +1,7 @@
+package auth
+
+import "testing"
+
+func TestUnmarshalUserInfo(t *testing.T) {
+
```
celian-garcia: TODO

internal/api/impl/auth/userinfo/userinfo.go
```diff
@@ -0,0 +1,68 @@
...
+		return err
+	}
+	*u = tmp
+	u.rawProperties = rawProperties
+
+	return nil
+}
+
+func (u *ExternalUserInfoProfile) getLogin(loginProps []string) string {
+	if login := u.getProperty(loginProps); login != "" {
```
celian-garcia, line 67: call this first only if user gave an explicit custom, otherwise the default perses behavior is to get it from email.

internal/api/impl/auth/userinfo/oidc_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestOIDCUserInfoUnmarshalJSON tests the unmarshalling of OIDCUserInfo from JSON.
+// TODO: fix that test as it's the way OIDC library parse the user info and it's not working
```
celian-garcia: TODO

internal/api/impl/auth/userinfo/userinfo.go
celian-garcia: Fixed it using [] as default login props instead of ["sub"]

Nexucis:
based on the discussion from the issue #3851 it looks like this PR can be close in favour of a different implementation.

Issue #3434, opened 2025-10-07
Add axis scale options (linear, logarithmic) for TimeSeriesChart panels


joluc:
### Proposal

Currently, the `TimeSeriesChart` panels only support linear axis scaling. When visualizing metrics that span multiple orders of magnitude (e.g., from 1 to 1,000,000), a linear Y-axis scale makes it difficult or impossible to see smaller values. A logarithmic scale can solve this by compressing large values and expanding small values, making all data visible and comparable.

### Common use cases:

- Request rates or error counts that vary from single digits to millions
- Memory usage spanning from KB to GB
- Latency metrics with outliers
- Any exponential growth or decay patterns

### Requested feature:

Add a Y-axis scale option to TimeSeriesChart panels with support for:

- Linear (default, current behavior)
- Logarithmic with configurable base (2, 10, 32...)

This would bring Perses to feature parity with Grafana's axis scale options and significantly improve its utility for a wide range of observability use cases.

AntoineThebaud:
This was implemented with https://github.com/perses/plugins/pull/500

PR #3704, opened 2025-12-15
[FEATURE]: updates nearby-series to support yAxis:type:log TimeSeriesChart

olandr:
# Description
This is a precondition for supporting #3434. If a user has selected a log-base for their render, this will perform a base-change to convert the log-base vector space to a linear space for the logic handling nearby series finding.

## Example
For instance If this is not present and you use a log-based chart, you end up with a "nearby-search interval" of just a few pixels. That is, if you have multiple series in the realm of ~1e4, using a log scale we would have y-values of ~4. The search interval would then be <4. But as the chart coordinates still are very large ~1e4 it would be impossible to get precise enough to hover/select a particular series.

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/ui/components/src/TimeSeriesTooltip/nearby-series.ts
+++ b/ui/components/src/TimeSeriesTooltip/nearby-series.ts
@@ -17,7 +17,6 @@ import { formatValue, TimeSeriesValueTuple, FormatOptions, TimeSeries } from '@p
 import { EChartsDataFormat, OPTIMIZED_MODE_SERIES_LIMIT, TimeChartSeriesMapping, DatapointInfo } from '../model';
 import { batchDispatchNearbySeriesActions, getPointInGrid, getClosestTimestamp } from '../utils';
 import { CursorCoordinates, CursorData, EMPTY_TOOLTIP_DATA } from './tooltip-model';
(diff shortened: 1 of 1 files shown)


ui/components/src/TimeSeriesTooltip/nearby-series.ts
```diff
@@ -17,7 +17,7 @@ import { formatValue, TimeSeriesValueTuple, FormatOptions, TimeSeries } from '@p
 import { EChartsDataFormat, OPTIMIZED_MODE_SERIES_LIMIT, TimeChartSeriesMapping, DatapointInfo } from '../model';
 import { batchDispatchNearbySeriesActions, getPointInGrid, getClosestTimestamp } from '../utils';
 import { CursorCoordinates, CursorData, EMPTY_TOOLTIP_DATA } from './tooltip-model';
-
+import {pow} from 'mathjs';
```
olandr: I noticed in the source code that you are using a smaller set of math instructions to keep the bundle small. I would suppose that this would then a bad-idea.

I have not looked into how to make that possible atm.

olandr:
Closing, see https://github.com/perses/shared/pull/11

PR #3711, opened 2025-12-15
[BREAKINGCHANGE] Manage plugin version and registry

Nexucis:
### Context

This PR is coming from the discussion https://github.com/perses/perses/discussions/1186. Specially the last comments that gives details and ideas about what we should do and why.

### Description

This PR is implementing a new `Metadata` struct for the plugin. It includes the version and the registry related to the plugin.

This impacts and requires multiple things:

- First thing first, managing a version (and a registry) implies you can have multiple version of the same plugin.
- A plugin is not identified only by its name anymore. A plugin is identified by its name, its version and its registry. These three keys give us an unique way to identify a plugin.

The idea to be backward compatible is to make this field optional. As soon as `plugin.metadata` is optional, it means you need to know which latest version of a plugin you have. If the user does not provide the plugin version, then we are using the latest version to validate it.

### Breaking change

Last point: as you will figure out, I stamped this PR as a breaking change, because indeed I have broken something: the API related to the dev plugin. I changed the path so in includes the word `dev` to make the difference between a regular plugin and a dev one. Also in term of REST design, having a method POST on `/api/v1/plugins` but it actually does not create a new plugin, does not make sense.
So now, to register a dev plugin, the request will go to `/api/v1/plugins/dev`

since these APIs are used through the CLI, I don't think it will impact someone as long as they are updating the CLI version as well. Which is something that needs to be mentioned somewhere.

### What is remaining

If you read the discussion pointed at the beginning, you will see we need to support a migration script. This is a next step for the backend.

Then in the frontend, the plugin system needs to be adapted to be able to support multi plugin version and registry. The registry and the version needs to used when calling the backend to get the assets and js files related to the plugin.
For the moment for the backward compatibility, the backend is filling the version of the registry for the frontend using default value.

And then, I think it will be nice if there is a way to highlight in the frontend for a given plugin the fact it is not the latest version used and it needs to be migrated manually.


internal/api/plugin/migrate/testdata/plugins/BarChart/mf-manifest.json
```diff
@@ -1,3 +1,334 @@
 {
-  "this has no use for this test right?": true
```
AntoineThebaud, line 2: Would you mind cleaning these test manifest files to keep only the fields that are relevant for the unit tests? (thus `id`, `name` & `metadata.buildInfo.version`)

internal/api/plugin/tree/tree_test.go
```diff
@@ -0,0 +1,307 @@
...
+type treeParameter struct {
+	name     string
+	metadata plugin.ModuleMetadata
+	instance any
+}
+
+func TestTree_Add(t *testing.T) {
+	testSuite := []struct {
+		name          string
+		basedTree     Tree[any]
```
AntoineThebaud:
```suggestion
		baseTree     Tree[any]
```

unless you mean the tree is based 🤓

internal/api/plugin/migrate/testdata/plugins/BarChart/mf-manifest.json
```diff
@@ -1,3 +1,12 @@
...
-}
\ No newline at end of file
+  "id": "BarChart",
+  "name": "BarChart",
+  "metaData": {
```
jgbernalp, line 4: should it be:

```suggestion
  "metadata": {
```
?

Nexucis: According to how the file `mf-manifest.json` is generated, it is written as `metaData`.

PR #3714, opened 2025-12-16
[ENHANCEMENT] File unmarshalling: Allow .yml as well

runiq:
# Description

File unmarshalling: Allow .yml as well. Also adds a debug message. I'm aware this doesn't fit how debug messages are used in the rest of the project, but I just wasted an hour on this and it'd be nice to save others the time.

# Checklist

### DIFF
--- a/internal/cli/file/file.go
+++ b/internal/cli/file/file.go
@@ -108,8 +108,9 @@ func visit(dir string) ([]string, error) {
 			return nil
 		}
 		fileName := info.Name()
-		if filepath.Ext(fileName) != ".json" && filepath.Ext(fileName) != ".yaml" {
+		if filepath.Ext(fileName) != ".json" && filepath.Ext(fileName) != ".yaml" && filepath.Ext(fileName) != ".yml" {
 			// skip every file that doesn't have the correct extension
+			logrus.Tracef("Skipping %s", fileName)
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


internal/cli/file/file.go
```diff
@@ -108,8 +108,9 @@ func visit(dir string) ([]string, error) {
 			return nil
 		}
 		fileName := info.Name()
-		if filepath.Ext(fileName) != ".json" && filepath.Ext(fileName) != ".yaml" {
+		if filepath.Ext(fileName) != ".json" && filepath.Ext(fileName) != ".yaml" && filepath.Ext(fileName) != ".yml" {
 			// skip every file that doesn't have the correct extension
+			logrus.Debugf("Skipping %s", fileName)
```
Nexucis: I think a trace level would make more sense here

runiq: Your call, of course. But to be honest, while setting Perses up for the first time, I would have loved for it to be a bit more chatty. :)

runiq:
Spurious CI failure?

internal/cli/file/file.go
Nexucis: for sure we are not putting that much logs, we should probably add more. For this one I think a trace log makes sense as now thanks to your change, you should not really need to see this one anymore.

PR #3726, opened 2025-12-19
[BUGFIX] Fix incomplete Grafana panel migration

AntoineThebaud:
# Description

Fields like TimeSeriesChart's `querySettings` or StatChart's `metricLabel` were not populated from their Grafana equivalent while they should because the queries (called `targets` on Grafana side) were removed from the panel object at unmarshalling time.

# Checklist

### DIFF
--- a/internal/api/plugin/migrate/grafana.go
+++ b/internal/api/plugin/migrate/grafana.go
@@ -48,6 +48,8 @@ type Panel struct {
 	json.RawMessage
 }
 
+// Custom unmarshal to both store fields and keep raw data (used by the CUE-based migration logic).
+// Checks unmarshalling errors only when unmarshaling complex structs.
 func (p *Panel) UnmarshalJSON(data []byte) error {
 	var tmp map[string]json.RawMessage
 	if err := json.Unmarshal(data, &tmp); err != nil {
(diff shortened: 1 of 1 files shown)


internal/api/plugin/migrate/migrate.go
```diff
@@ -144,6 +144,11 @@ func executeCuelangScript(cueScript *build.Instance, grafanaData []byte, defID s
 		ctx.CompileBytes(grafanaData),
 	)
 
+	if logrus.IsLevelEnabled(logrus.TraceLevel) {
+		logrus.Tracef("Grafana %s to migrate:", typeOfDataToMigrate)
+		fmt.Fprintf(os.Stderr, "%# v\n", grafanaValue)
```
AntoineThebaud, line 149: FYI this printf is to have a "pretty" display of the payload, e.g:
```
time="2025-12-19T09:39:56+01:00" level=trace msg="Grafana panel to migrate:"
#panel: {
        datasource: {
                type: "prometheus"
                uid:  "myTSDB"
        }
        description: "Memory for selected containers"
        fieldConfig: {
                defaults: {
                        color: {
                                mode: "palette-classic"
                        }
                        custom: {
                                axisBorderShow:   false
[...]
```

internal/api/plugin/migrate/grafana.go
```diff
@@ -82,7 +82,6 @@ func (p *Panel) UnmarshalJSON(data []byte) error {
 		if err := json.Unmarshal(targets, &panel.Targets); err != nil {
 			return err
 		}
-		delete(tmp, "targets")
```
AntoineThebaud, line 85: @Nexucis I wonder if the change shouldnt go further here: what was the reason to have a different logic for `type, title, description, collapsed` vs `panels, gridPos, targets, links` in the first place? I mean not only the `delete()` call but also the errors getting ignored in the first case vs returned in the second.

Nexucis: so the logic here was to unmarshal anything required for the go code and remove anything not required by the cuelang script.

Saying differently, when I rewrite the migration logic, I have assumed based on what existed, the cuelang script didn't need to have access to the targets. Because the targets field is something that needs to be decoded manually depending on the struct and then passed to the query plugin.

So I supposed it does not work anymore because one script required to have access to the targets field.

AntoineThebaud: as discussed, removed all delete() calls & added comment to explain about the differentiated erro processing logic

PR #3762, opened 2026-01-05
[ENHANCEMENT] `plugin test-schemas`: match migration output against model

### LINKED ISSUE #3478: `plugin test-schemas`: match migration output against model
### Proposal

Currently, for the migration tests ran by `percli plugin test-schemas`, the logic checks if the actual output matches the expected output. However it doesn't check neither if actual nor expected are valid model-wise, which can lead to missing errors/typos..

Concretely the change would just consist in running model validation against the `expected.json` of each migration test.

AntoineThebaud:
# Description

Closes #3478.

This is how the output looks like - fyi the issue reported is actually a real one we have to fix following this change 😅: 

```bash
$ ./bin/percli plugin test-schemas --plugin.path ../plugins/timeserieschart/
[...]
✗ multiple-overrides (migrate) [migrate\tests\multiple-overrides]: Expected output failed model validation: spec.querySettings.3.colorValue: invalid value "yellow" (out of bound =~"^#(?:[0-9a-fA-F]{3}){1,2}$")
[...]
```

# Checklist

### DIFF
--- a/internal/cli/cmd/plugin/testschemas/test-schemas.go
+++ b/internal/cli/cmd/plugin/testschemas/test-schemas.go
@@ -152,7 +152,6 @@ func (o *option) runAllTests() ([]TestResult, error) {
 		if d.Name() == "tests" {
 			// Determine the plugin kind from the parent directory (which is assumed to be the entrypoint of the `model` CUE package, cf -h)
 			parentDir := filepath.Dir(currentPath)
-
 			_, buildInstance, err := schema.LoadModelSchema(parentDir)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


internal/cli/cmd/plugin/testschemas/testdata/my-datasource-plugin/schemas/my-query/model.cue
```diff
@@ -13,9 +13,14 @@
...
+	"strings"
+
+	"github.com/perses/shared/cue/common"
+)
 
 kind: "MyQuery"
 spec: close({
-	query: strings.MinRunes(1)
+	query:       strings.MinRunes(1)
+	thresholds?: common.#thresholds
```
AntoineThebaud, line 25: FYI this may seem useless as I didnt change the migration test data but it is actually covers deps resolution, as I did first a mistake for parentDir computation above - so unit tests were looking fine but the modified command was failing for real-life plugins..

internal/cli/cmd/plugin/testschemas/test-schemas.go
```diff
@@ -347,6 +347,29 @@ func (o *option) runMigrationTestsForPath(testDir string, buildInstance *build.I
 
 		logrus.Debugf("Run migration test for plugin %s of type %s", expectedPlugin.Kind, pluginKind)
 
+		// Load the model schema for validating the expected output
+		// testDir is migrate/tests, so we need to go up two levels to reach the model directory
+		parentDir := filepath.Dir(filepath.Dir(testDir))
```
Nexucis: I think before running any tests, we could actually load the schema and migration schema and shared it where it is necessary.
It will avoid this kind of hack (i.e. knowing you need to go two upper level to get the schema. 

And hopefully with the `internal/api/plugin` package this should be easy to load them at the beginning of the execute function

AntoineThebaud: I moved the call to LoadModelSchema upper, enough to remove the two-level hack, still I couldnt e.g load once and share between model & migration tests because it's making strong assumptions on the file organization (also you have to deal with the difference between single plugins like TimeSeriesChart and plugin "packages" like Prometheus). Still now it's shared among all migrate tests

Nexucis (approved):

Issue #3766, opened 2026-01-07
OIDC RP initiated logout: failed to logout from Keycloak due to missing id_token_hint in redirect URL


GrzegorzMika:
### What did you do?

1. Register Perses in Keycloak IDP
2. In Perses, enable the authentication via the OIDC provider pointing to the Keycloak instance with logout functionality enabled
3. Log in into the Perses instance via the Keycloak (works correctly)
4. Hit the logout button in Perses
5. User is correctly redirected to the Keycloak instance to terminate the session, but the Keycloak shows the error page with the error "Missing parameters: id_token_hint" (screen below) and the session in Keycloak is not terminated

<img width="793" height="517" alt="Image" src="https://github.com/user-attachments/assets/6ff413d7-afc9-4d46-a764-f81dd50cd2bf" />

### What did you expect to see?

Correct Keycloak screen allowing to terminate the Keycloak session 

<img width="793" height="517" alt="Image" src="https://github.com/user-attachments/assets/acab2c95-f603-45d7-947e-ded951c0121d" />

### System information

OS: Linux 6.8.0-90-generic x86_64, K8S: v1.34.1+rke2r1 

### Perses version

0.53.0-beta.4

### Perses configuration file

```yaml
security:
      encryption_key: <encryption key>
      authentication:
        access_token_ttl: 1m
        disable_sign_up: true
        providers:
          oidc:
          - client_id: perses
            issuer: <issuer address>
            name: Keycloak
            scopes:
            - email
            - openid
            - roles
            - profile
            slug_id: keycloak
            logout:
              enabled: true
```

### Logs

```text
No relevant logs
```

GrzegorzMika:
Per my discussion with @celian-garcia I can work on this. 

The solution would be to include the RP client id in the redirect URL to IDP instance

celian-garcia:
@GrzegorzMika thank you again, I'm closing the issue, since it's merged :)

PR #3767, opened 2026-01-07
[BUGFIX]: Add RP client_id to the redirect URL when logging out

GrzegorzMika:
[BUGFIX]: Add RP client_id to the redirect URL when logging out of the OIDC IDP

# Description

This PR enhances the RP-initiated logout process by appending the `client_id` to the URL when redirecting to the Identity Provider's (IDP) logout endpoint.

According to the OpenID Connect RP-Initiated Logout specification, when a `post_logout_redirect_uri` is provided, it is recommended to also include the `id_token_hint` (the ID Token previously issued by the OP). If providing the `id_token_hint` is not possible—as is the case in Perses where the original ID token is not retained for this step—the `client_id` can be passed as an alternative.

While the specification indicates that passing the `client_id` or `id_token_hint` is not strictly mandatory, certain IDPs (notably Keycloak) require at least one of these parameters to be present if a `post_logout_redirect_uri` is used. Without this, the logout process fails to correctly terminate the IDP's session.

Relates to [3766](https://github.com/perses/perses/issues/3766)

# Screenshots

No UI changes

# Checklist

## UI Changes

No UI changes

### DIFF
--- a/docs/concepts/authentication.md
+++ b/docs/concepts/authentication.md
@@ -58,8 +58,18 @@ security:
           client_id: "<secret>"
           client_secret: "<secret>"
           issuer: "https://login.microsoftonline.com/<tenant-id>/v2.0"
-          redirect_uri: "http://localhost:8080/api/auth/providers/oidc/azure/callback"
+          redirect_uri: "http://localhost:8080/api/auth/providers/oidc/msft/callback"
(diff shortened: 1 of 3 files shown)


GrzegorzMika:
@celian-garcia I'm glad I could contribute and I'm happy to help in the future:)

PR #3776, opened 2026-01-12
[ignore] Provide a lib to check the license header

Nexucis:
I have rewritten the way to check the license header in files.

This is an attempt to provide a common lib to check the license header across all repo under the Perses organisation.

As a side notes, you can notice the old script missed the file `Footer.tsx`.


scripts/check-license/check-license.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"flag"
+
+	"github.com/perses/perses/scripts/pkg/license"
+)
+
+func main() {
+	l := &license.License{
+		ExcludedFiles:    []string{"migrate.cue"},
```
AntoineThebaud: I forgot to do this change earlier but actually since https://github.com/perses/perses/pull/2481 the migrate.cue files can (should) be included. AFAIS we already put the license header in them anyway so shouldnt trigger much errors.

jgbernalp:
Should we add it to perses common?

Nexucis:
> Should we add it to perses common?

well since it's more for scripting purpose, I have assumed we can keep it here. But it's maybe a false assumption.

scripts/check-license/check-license.go
Nexucis: done, I had to update some migration files that are in some test folders.

PR #3779, opened 2026-01-12
Sync main branch to feature/k8s-provider

Nexucis:
sync main branch. This PR must be merged not squashed


Nexucis:
mmm it looks like there is an issue with the build that is writing too much data on the disk. That's the first time I am seeing this, I will need to investigate what is the issue here.

PR #3780, opened 2026-01-12
[ignore] add default license checker and expose interface

Nexucis:
I am proposing a default license checker. I would believe it should cover most of the use cases we have in the Perses org.

And to make it more flexible, I changed the packege to expose an interface instead of the raw struct and add more functions to manage the various pattern you would like to add.


scripts/pkg/license/license.go
```diff
@@ -165,16 +212,52 @@ func (l *License) fix() {
...
+}
+
+func (l *license) AddPattern(patterns ...string) License {
+	for _, pattern := range patterns {
+		if !slices.Contains(l.patterns, pattern) {
+			l.patterns = append(l.patterns, pattern)
+		}
+	}
+	return l
+}
```
AntoineThebaud, line 251: WDYT of factorizing like this?

```suggestion

func addUnique(dst *[]string, items ...string) {
    for _, it := range items {
        if !slices.Contains(*dst, it) {
            *dst = append(*dst, it)
        }
    }
}

func (l *license) AddExcludedFile(files ...string) License {
    addUnique(&l.excludedFiles, files...)
    return l
}

func (l *license) AddExcludedDir(dirs ...string) License {
    addUnique(&l.excludedDirs, dirs...)
    return l
}

func (l *license) AddExcludedPattern(patterns ...string) License {
    addUnique(&l.excludedPatterns, patterns...)
    return l
}

func (l *license) AddPattern(patterns ...string) License {
    addUnique(&l.patterns, patterns...)
    return l
}
```

Nexucis: I used a set datastruct instead. It's cleaner like that.

PR #3782, opened 2026-01-13
[FEATURE] proxy: drop or keep some headers

Gladorme:
# Description

Currently all requests made by the UI have "Origin"and "Referer" headers added and it's forwarded to the datasource by the proxy. However, some datasources may check these headers and fail requests because the Origin/Referer of Perses is not defined in their conf.

TODO: Add checkbox on HTTP Proxy form (UI => shared)

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/internal/api/impl/proxy/proxy.go
+++ b/internal/api/impl/proxy/proxy.go
@@ -282,6 +282,11 @@ func (h *httpProxy) serve(c echo.Context) error {
 		return apiinterface.InternalError
 	}
 
+	if h.config.RemoveOriginAndReferer {
+		req.Header.Del("Referer")
+		req.Header.Del("Origin")
+	}
+
 	// redirect the request to the datasource
 	req.URL.Path = h.path
 	logrus.WithFields(map[string]interface{}{
--- a/internal/api/impl/proxy/proxy.go
+++ b/internal/api/impl/proxy/proxy.go
@@ -282,9 +282,18 @@ func (h *httpProxy) serve(c echo.Context) error {
 		return apiinterface.InternalError
 	}
 
-	if h.config.RemoveOriginAndReferer {
-		req.Header.Del("Referer")
-		req.Header.Del("Origin")
+	if len(h.config.DropHeaders) > 0 {
+		for _, headerToDrop := range h.config.DropHeaders {
+			req.Header.Del(headerToDrop)
+		}
+	}
+
+	if len(h.config.AllowHeaders) > 0 {
+		headerKept := make(http.Header)
+		for _, header := range h.config.AllowHeaders {
+			headerKept.Add(header, req.Header.Get(header))
+		}
+		req.Header = headerKept
 	}
 
 	// redirect the request to the datasource
--- a/pkg/model/api/v1/datasource/http/http.go
+++ b/pkg/model/api/v1/datasource/http/http.go
@@ -81,6 +81,8 @@ type Config struct {
 	// Secret is the name of the secret that should be used for the proxy or discovery configuration
 	// It will contain any sensitive information such as password, token, certificate.
 	Secret string `json:"secret,omitempty" yaml:"secret,omitempty"`
+	// RemoveOriginAndReferer is removing Origin and Referer headers when proxying requests, to avoid CORS issues with some datasources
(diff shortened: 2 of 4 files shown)


ui/core/src/model/http-proxy.ts
```diff
@@ -29,6 +29,8 @@ export interface HTTPProxySpec {
   // secret is the name of the secret that should be used for the proxy or discovery configuration
   // It will contain any sensitive information such as password, token, certificate.
   secret?: string;
+  // removeOriginAndReferer is removing Origin and Referer headers when proxying requests, to avoid CORS issues with some datasources
+  removeOriginAndReferer?: boolean;
```
jgbernalp: should we add an exclude or allow headers lists?. This will allow to configure other headers rather than set a boolean for 2 of them.

internal/api/impl/proxy/proxy.go
```diff
@@ -260,6 +260,11 @@ func (h *httpProxy) serve(c echo.Context) error {
 		return apiinterface.InternalError
 	}
 
+	if h.config.RemoveOriginAndReferer {
+		req.Header.Del("Referer")
+		req.Header.Del("Origin")
```
Nexucis: mmm I would be cautious by doing that.
I need to make so research to know exactly it implies to do that and what would be the consequences

ui/core/src/model/http-proxy.ts
jgbernalp: Something like:
```yaml
headerPolicy:
   allow:
    - Authorization
    - Origin
   drop:
    - Referer
```

Gladorme: I guess it would be better: more standard and future proof, but less user friendly 😅  I guess user could only define either allow or drop. But maybe we could remove these headers by default when passing in the proxy (cf comment under Nexucis msg) 🤔

internal/api/impl/proxy/proxy.go
Gladorme: It's fixing cross-origin check done by some backends. Basically we are adding this HTTP proxy for not having this issue on front side (CORS). I was even wondering if we should not remove them by default in the proxy

ui/core/src/model/http-proxy.ts
jgbernalp: I don't think it will be a good idea to change the request in ways the user cannot control. If one of the 2 headers is needed on the backend then the user will hit a blocker.

Nexucis: I tend to think this issue would happen only on dev environment, so probably you don't need to do it in production.

Another way to handle that would be to provide the same kind of configuration @jgbernalp is proposing but on the backend side. So you can configure the backend to filter the headers.
Maybe both can be a good idea to apply.
What do you think @Gladorme @jgbernalp ?

Nexucis: By default all header, excepting `Authorization` are passed to the request, it might be easier to just have a drop list for the user perspective. If we want to maintain the `allow` list, that would need the user need to lists all headers he wants to authorise. This can be really painful don't you think @jgbernalp ?

jgbernalp: Well I think the `allow` has precedence over the `drop`. But in most cases I think allow will be used because you would like only to send what is required and avoid security issues like header injection. If we still want to be developer friendly then we can add a wildcard: `*`

Nexucis: alright, sounds good to me.
@Gladorme can you adjust your PR to reflect this proposition ? Thank you

pkg/model/api/v1/datasource/http/http.go
```diff
@@ -81,6 +81,10 @@ type Config struct {
 	// Secret is the name of the secret that should be used for the proxy or discovery configuration
 	// It will contain any sensitive information such as password, token, certificate.
 	Secret string `json:"secret,omitempty" yaml:"secret,omitempty"`
+	// AllowHeaders will drop all headers not in the list before forwarding the request to the datasource.
+	AllowHeaders []string `json:"allowHeaders,omitempty" yaml:"allowHeaders,omitempty"`
```
Nexucis, line 85: the proxy definition has been moved to `perses/spec`. Please open a PR there and modify the cue and typescript definition as well.

Nexucis:
and the form to be able to edit these new fields is required.

Issue #3786, opened 2026-01-15
Support custom column settings in Logs table


jgbernalp:
### Proposal

As other tables, log tables should support defining custom columns parsing and formatting the log data.

shahrokni:
## Clue
To add custom columns to a log table, other dashboarding applications use a feature called Transformation. 
Transformation is a strong feature by which new columns can be defined and be filled by extracting or somehow generating data based on the existing columns (simply existing data). If you agree with the assumption, let me know which kind of transformations you are interested in at this point. 

## Transformations
- Adding fields from calculation
- Concatenate fields
- Convert fields type
- Extract Fields
- etc ...

jgbernalp:
Great @shahrokni , initially we want to be able to extract fields in new columns based on the data, for example `message` or `severity` in logs. Which might be a simple transformation.

shahrokni:
> Great [@shahrokni](https://github.com/shahrokni) , initially we want to be able to extract fields in new columns based on the data, for example `message` or `severity` in logs. Which might be a simple transformation.

Wonderful!
Question, do we have anything already developed for "Other Tables" which could help us expedite the process, or we should start from scratch? Something similar to Transformation?

jgbernalp:
It seems we have a transformation already: https://github.com/perses/perses/blob/main/ui/core/src/model/transforms.ts but it seems we are missing the one that extracts data into a new column

shahrokni:
> It seems we have a transformation already: https://github.com/perses/perses/blob/main/ui/core/src/model/transforms.ts but it seems we are missing the one that extracts data into a new column

May I work on this issue?

jgbernalp:
Yes thank you!

PR #3791, opened 2026-01-16
[FEATURE] Core: add extractn column frilds transformation

shahrokni:
Relates to https://github.com/perses/perses/issues/3786

# Description 🖊️ 
This change adds basic type definition of `Extract Column Fields Transformation`

# Checklist

## UI Changes

### DIFF
--- a/ui/core/src/model/transforms.ts
+++ b/ui/core/src/model/transforms.ts
@@ -22,6 +22,15 @@ export interface JoinByColumnValueTransform {
   };
 }
 
+export interface ExtractColumnFieldsTransform {
+  kind: 'ExtractColumnFields';
+  spec: TransformCommonSpec & {
+    column: string;
+    format: 'JSON' | 'Regex' | 'SplitByDelimiter' | 'KeyValuePairs';
+    matcher?: string;
+  };
+}
+
 export interface MergeColumnsTransform {
   kind: 'MergeColumns';
   spec: TransformCommonSpec & {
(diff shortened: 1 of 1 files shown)


ui/core/src/model/transforms.ts
```diff
@@ -22,6 +22,15 @@ export interface JoinByColumnValueTransform {
   };
 }
 
+export interface ExtractColumnFieldsTransform {
+  kind: 'ExtractColumnFields';
+  spec: TransformCommonSpec & {
+    column: string;
+    format: 'JSON' | 'Regex' | 'SplitByDelimiter' | 'KeyValuePairs';
+    matcher?: string;
```
jgbernalp, line 30: IIUC, `format` is the format of the data and matcher will use either a JSON field accessed by defining matcher as `my.data.column`. And with the regex defining just the regex content. 

With the split delimiter how can i select which column from the data I want to extract?

With the key value pairs I assume the key will match the column name, is this correct?

shahrokni: With column (source) you choose the extraction source. 
For example, lets say in a hypothetical scenario you fill the **column** with `hostname`. This means you want to extract data from hostname column. 


## Split by delimiter
Now if you choose this option, you need to set the delimiter (matcher).
The delimiter splits the string, and adds corresponding columns. 

For example here, the string does not contain the delimiter itself, so the result looks odd

image

## Key Value Pair

key value pair does not need any matcher. It considers : and = as the standard indicator of key value pairs in a string. 
Every found key represent a column while corresponding values will be placed on every row cell.  The following image shows this option acting upon message column. 

image

jgbernalp (approved):

PR #3793, opened 2026-01-16
[FEATURE]: add loki to the docker-compose yml

shahrokni:
Relates to https://github.com/perses/perses/issues/3786

# Description 🖊️ 

This change adds Loki service to the docker-compose.yml

# Checklist

## UI Changes

### DIFF
--- a/dev/docker-compose.yaml
+++ b/dev/docker-compose.yaml
@@ -4,10 +4,25 @@
 
 name: "perses-dev"
 services:
+  loki:
+    image: grafana/loki:latest
+    ports:
+      - "3100:3100"
+    profiles: [loki]
+  log-generator:
+        image: mingrammer/flog
+        command:
+          - --loop
+          - --format=json
+          - --number=10
+          - --delay=100ms
+      volumes:
+        - ./loki-data:/loki
+
   mariadb:
     image: mariadb
     ports:
-      - '3306:3306'
+      - "3306:3306"
     environment:
       MARIADB_ROOT_PASSWORD: root
       MARIADB_DATABASE: perses
@@ -17,7 +32,7 @@ services:
   postgres:
     image: postgres:17
     ports:
-      - '5432:5432'
+      - "5432:5432"
     environment:
       POSTGRES_DB: perses
       POSTGRES_USER: user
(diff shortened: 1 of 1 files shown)


dev/data/mock/podman-loki-logs.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
Nexucis: I would avoid to put that in the data folder even if it is in a sub-folder. The data folder is read by the perses binary, it's better to keep clean.

Why not putting it in the `extra` folder or at the root of the `dev` folder.

jgbernalp:
we could also use: https://github.com/mingrammer/flog 

something like this in the compose file
```
log-generator:
    image: mingrammer/flog
    command:
      - --loop
      - --format=json
      - --number=10 
      - --delay=100ms
```

shahrokni:
> we could also use: https://github.com/mingrammer/flog
> 
> something like this in the compose file
> 
> ```
> log-generator:
>     image: mingrammer/flog
>     command:
>       - --loop
>       - --format=json
>       - --number=10 
>       - --delay=100ms
> ```

Wonderful. 
Should I remove the script then?

dev/docker-compose.yaml
```diff
@@ -4,10 +4,35 @@
...
+        [
+          "CMD-SHELL",
+          "wget --no-verbose --tries=1 --spider http://localhost:3100/ready || exit 1",
+        ]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    environment:
+      - TZ=UTC
+
```
Gladorme, line 21: Similar to others datasources, add `profiles` field

dev/docker-compose.yaml
```diff
@@ -4,10 +4,25 @@
 
 name: "perses-dev"
 services:
+  loki:
+    image: grafana/loki:latest
+    ports:
+      - "3100:3100"
+    profiles: [loki]
+    log-generator:
```
jgbernalp: I think the indentation is wrong...

jgbernalp:
@shahrokni I tried locally and the indentation of the log generator image is incorrect. This is a sibling image of loki. Can you check?

dev/docker-compose.yaml
shahrokni: > I think the indentation is wrong...

Are you sure? The running container is generating logs

jgbernalp: yes pretty sure. The yaml is incorrect, maybe you have a different local version?

dev/docker-compose.yaml
```diff
@@ -4,10 +4,25 @@
 
 name: "perses-dev"
 services:
+  loki:
+    image: grafana/loki:latest
+    ports:
+      - "3100:3100"
+    profiles: [loki]
+  log-generator:
+        image: mingrammer/flog
```
Nexucis, line 13: it looks like the indentation is not the same between each blocks. If you look at the `log-generator.image` and `loki.image`, they are not indented the same way. Can you please fix that?

dev/docker-compose.yaml
Nexucis: you still need to add `profiles: [loki]` here

dev/docker-compose.yaml
shahrokni: yes my yml linter doesn't validate the file! 
I am checking and fixing the problem

dev/docker-compose.yaml
Nexucis: similar to tempo where you have the server and then another services used to populate tempo

dev/docker-compose.yaml
```diff
@@ -4,10 +4,25 @@
...
+        command:
+          - --loop
+          - --format=json
+          - --number=10
+          - --delay=100ms
+      volumes:
+        - ./loki-data:/loki
```
Nexucis, line 20: I believe this volume needs to be shared between loki and log-generator.

PR #3795, opened 2026-01-16
Fix docker compose loki

jgbernalp:
# Description

Fix docker compose for loki: https://github.com/perses/perses/pull/3793

# Checklist

### DIFF
--- a/dev/docker-compose.yaml
+++ b/dev/docker-compose.yaml
@@ -41,7 +41,7 @@ services:
   mariadb:
     image: mariadb
     ports:
-      - '3306:3306'
+      - "3306:3306"
     environment:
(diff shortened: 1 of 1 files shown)


jgbernalp:
@shahrokni not sure why other commits were included, maybe i took an updated base but this should fix the issue with the compose file

shahrokni:
> @shahrokni not sure why other commits were included, maybe i took an updated base but this should fix the issue with the compose file

Thank you for the support. 
I ran and tested it. It looks fine. 
However, could we ask @Nexucis to test as well? 
I have been ruining Loki locally and what I see now may not be reliable

jgbernalp:
@shahrokni I changed the base since you closed the other PR. Now includes only the relevant commits

PR #3808, opened 2026-01-23
[BUGFIX] include lint command in plugin module generation

jgbernalp:
# Description

This PR:

- adds eslint to the plugin module template.
- Updates the rspbuild config to match the latest plugin changes
- Updates vulnerable dependencies for the package.json included with the template

# Checklist


internal/cli/cmd/plugin/generate/templates/module/.eslintrc.js.tmpl
```diff
@@ -0,0 +1,78 @@
+// Copyright The Perses Authors
```
Nexucis: since this command is to generate the squeleton of a any plugin (public or private), the Perses license might not be applicable. So I don't think it's a good idea to have the license in the generated files

PR #3813, opened 2026-01-26
[FEATURE] Add MariaDB proxy support

ZPascal:
## Summary

This PR adds comprehensive MariaDB support to Perses and significantly extends the test coverage for SQL datasources (MySQL, MariaDB, and PostgreSQL). MariaDB can now be configured as a first-class datasource with its own configuration options, while maintaining full backward compatibility with existing MySQL and PostgreSQL configurations.

## Motivation

MariaDB is a popular MySQL-compatible database that is widely used in production environments. While Perses previously had the `mariadb` driver constant defined, it lacked proper configuration support and comprehensive testing. This PR addresses these gaps by:

1. Adding dedicated MariaDB configuration support
2. Extending test coverage across all SQL data sources
3. Providing clear documentation for users

<details>
<summary>Details of the change</summary>

#### 1. SQL Model (`pkg/model/api/v1/datasource/sql/sql.go`)
- **Added** `MariaDB *MySQLConfig` field to the `Config` struct
- MariaDB uses the same configuration structure as MySQL since they are protocol-compatible
- Maintains backward compatibility - users can still use the `mysql` config field with MariaDB driver

#### 2. Proxy Implementation (`internal/api/impl/proxy/proxy.go`)
- **Updated** `openMySQL()` function to intelligently select configuration based on driver type
- When `driver: mariadb` is specified, the function now checks for `mariadb` config first, falling back to `mysql` config for compatibility
- No breaking changes to existing MySQL or PostgreSQL proxy behavior

#### 3. SQL Model Tests (`pkg/model/api/v1/datasource/sql/sql_test.go`)
Added **34 comprehensive test cases** covering:

**JSON Unmarshaling (11 tests):**
- Basic configurations for all three drivers (MySQL, MariaDB, PostgreSQL)
- MariaDB with custom parameters (charset, collation, timeouts)
- MySQL with custom parameters
- PostgreSQL with all configuration options
- Error cases: invalid SSL mode, missing driver/host/database, unsupported driver

**YAML Unmarshaling (8 tests):**
- Basic and advanced configs for all drivers
- MariaDB-specific parameter testing
- Invalid SSL mode and missing field validation

**Configuration Validation (7 tests):**
- Valid configuration detection for all drivers
- Missing required field detection
- Unsupported driver handling

**PostgreSQL SSL Validation (8 tests):**
- All SSL modes: `disable`, `allow`, `prefer`, `require`, `verify-ca`, `verify-full`
- Invalid SSL mode detection
- Empty SSL mode handling

#### 4. Proxy Tests (`internal/api/impl/proxy/proxy_test.go`)
Added **14 SQL proxy connection tests**, including:

**MariaDB-specific (4 tests):**
- Basic MariaDB connection
- Connection with username and password
- Connection with custom MariaDB configuration
- Connection with TLS/SSL support

**Extended MySQL coverage (3 tests):**
- Connection with username
- Connection with custom MySQL parameters

**Extended PostgreSQL coverage (5 tests):**
- Connection with username
- All SSL mode variations
- SSL mode with TLS config validation

**Error cases:**
- Unsupported driver detection
- TLS/SSL mode conflict validation

#### 5. Plugin Documentation (`docs/plugins/common.md`)
- **Updated** driver enum to include `mariadb`: `driver: <enum | possibleValue = 'mysql' | 'mariadb' | 'postgres'>`
- **Added** MariaDB configuration section with all available options
- Documented connection parameters, timeouts, and packet size settings

### MariaDB Datasource Configuration

```yaml
kind: "Datasource"
metadata:
  name: "production-mariadb"
  project: "myapp"
spec:
  default: false
  plugin:
    kind: "SQLProxy"
    spec:
      driver: "mariadb"
      host: "mariadb.example.com:3306"
      database: "app_database"
      secret: "mariadb-credentials"
      mariadb:
        params:
          charset: "utf8mb4"
          collation: "utf8mb4_unicode_ci"
          parseTime: "true"
        maxAllowedPacket: 33554432  # 32MB
        timeout: 30s
        readTimeout: 10s
        writeTimeout: 10s
```

</details>

## Backward Compatibility

✅ **Fully backward compatible** - no breaking changes:

- Existing MySQL datasources continue to work without modification
- Existing PostgreSQL datasources are unaffected
- The `mysql` configuration field can still be used with `driver: mariadb` for migration scenarios
- All existing tests continue to pass

## Migration Path

Users currently using the MySQL driver with MariaDB can optionally migrate:

**Before (still works):**
```yaml
driver: mysql
mysql:
  params:
    charset: "utf8mb4"
```

**After (recommended for MariaDB):**
```yaml
driver: mariadb
mariadb:
  params:
    charset: "utf8mb4"
```

## Technical Notes

1. **Driver Compatibility:** MariaDB uses the same Go MySQL driver (`github.com/go-sql-driver/mysql`) since they share the same wire protocol.

2. **Configuration Precedence:** When using `driver: mariadb`, the proxy checks for `mariadb` config first, then falls back to `mysql` config if not present.

3. **TLS/SSL Support:** MariaDB connections support TLS configuration via the Perses secret mechanism, identical to MySQL.

4. **All MySQL Parameters Supported:** MariaDB config supports all MySQL connection parameters (charset, collation, parseTime, timeouts, etc.).

## Checklist

- [x] Code changes implemented
- [x] Tests added and passing
- [x] Documentation updated
- [x] Backward compatibility maintained
- [x] No breaking changes introduced

## Related Issues

This PR enhances SQL data source support and addresses the need for explicit MariaDB configuration, while significantly improving test coverage across all SQL data sources.

### DIFF
--- a/pkg/model/api/v1/datasource/sql/sql.go
+++ b/pkg/model/api/v1/datasource/sql/sql.go
@@ -148,8 +148,7 @@ func (s *Config) UnmarshalYAML(unmarshal func(any) error) error {
 }
 
 func (s *Config) validate() error {
-	err := s.verifySupportedDriver()
-	if err != nil {
+	if err := s.verifySupportedDriver(); err != nil {
 		return err
 	}
 
--- a/pkg/model/api/v1/datasource/sql/sql_test.go
+++ b/pkg/model/api/v1/datasource/sql/sql_test.go
@@ -423,10 +423,9 @@ database: test
 
 func TestValidateConfig(t *testing.T) {
 	testSuite := []struct {
-		name      string
-		config    Config
-		expectErr bool
-		errMsg    string
+		name         string
+		config       Config
+		expectErrMsg string
 	}{
 		{
 			name: "valid mysql config",
@@ -435,7 +434,6 @@ func TestValidateConfig(t *testing.T) {
 				Host:     "localhost:3306",
 				Database: "test",
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid mariadb config",
@@ -444,7 +442,6 @@ func TestValidateConfig(t *testing.T) {
 				Host:     "localhost:3306",
 				Database: "test",
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid postgres config",
@@ -453,7 +450,6 @@ func TestValidateConfig(t *testing.T) {
 				Host:     "localhost:5432",
 				Database: "test",
 			},
-			expectErr: false,
 		},
 		{
 			name: "missing driver",
@@ -461,8 +457,7 @@ func TestValidateConfig(t *testing.T) {
 				Host:     "localhost:5432",
 				Database: "test",
 			},
-			expectErr: true,
-			errMsg:    "driver is required",
+			expectErrMsg: "driver is required",
 		},
 		{
 			name: "unsupported driver",
@@ -471,8 +466,7 @@ func TestValidateConfig(t *testing.T) {
 				Host:     "localhost:1433",
 				Database: "test",
 			},
-			expectErr: true,
-			errMsg:    "not supported",
+			expectErrMsg: "not supported",
 		},
 		{
 			name: "missing host",
@@ -480,8 +474,7 @@ func TestValidateConfig(t *testing.T) {
 				Driver:   DriverMySQL,
 				Database: "test",
 			},
-			expectErr: true,
-			errMsg:    "host cannot be empty",
+			expectErrMsg: "host cannot be empty",
 		},
 		{
 			name: "missing database",
@@ -489,19 +482,16 @@ func TestValidateConfig(t *testing.T) {
 				Driver: DriverMySQL,
 				Host:   "localhost:3306",
 			},
-			expectErr: true,
-			errMsg:    "database cannot be empty",
+			expectErrMsg: "database cannot be empty",
 		},
 	}
 
 	for _, test := range testSuite {
 		t.Run(test.name, func(t *testing.T) {
 			err := test.config.validate()
-			if test.expectErr {
+			if len(test.expectErrMsg) > 0 {
 				assert.Error(t, err)
-				if test.errMsg != "" {
-					assert.Contains(t, err.Error(), test.errMsg)
-				}
+				assert.Contains(t, err.Error(), test.expectErrMsg)
 			} else {
 				assert.NoError(t, err)
 			}
@@ -511,76 +501,67 @@ func TestValidateConfig(t *testing.T) {
 
 func TestPostgresConfigValidation(t *testing.T) {
 	testSuite := []struct {
-		name      string
-		config    PostgresConfig
-		expectErr bool
-		errMsg    string
+		name         string
+		config       PostgresConfig
+		expectErrMsg string
 	}{
 		{
 			name: "valid disable ssl mode",
 			config: PostgresConfig{
 				SSLMode: SSLModeDisable,
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid allow ssl mode",
 			config: PostgresConfig{
 				SSLMode: SSLModeAllow,
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid prefer ssl mode",
 			config: PostgresConfig{
 				SSLMode: SSLModePreferable,
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid require ssl mode",
 			config: PostgresConfig{
 				SSLMode: SSLModeRequire,
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid verify-full ssl mode",
 			config: PostgresConfig{
 				SSLMode: SSLModeVerifyFull,
 			},
-			expectErr: false,
 		},
 		{
 			name: "valid verify-ca ssl mode",
 			config: PostgresConfig{
 				SSLMode: SSLModeVerifyCA,
 			},
-			expectErr: false,
 		},
 		{
 			name: "invalid ssl mode",
 			config: PostgresConfig{
 				SSLMode: "invalid-mode",
 			},
-			expectErr: true,
-			errMsg:    "unknown ssl mode",
+			expectErrMsg: "unknown ssl mode",
 		},
 		{
 			name: "empty ssl mode is valid",
 			config: PostgresConfig{
 				SSLMode: "",
 			},
-			expectErr: false,
 		},
 	}
 
 	for _, test := range testSuite {
 		t.Run(test.name, func(t *testing.T) {
 			err := test.config.validate()
-			if len(test.errMsg) > 0 {
+			if len(test.expectErrMsg) > 0 {
 				assert.Error(t, err)
-				assert.Contains(t, err.Error(), test.errMsg)
+				assert.Contains(t, err.Error(), test.expectErrMsg)
 			} else {
 				assert.NoError(t, err)
 			}


Nexucis:
Hello @ZPascal,

Thank you for your contribution. While I am not against adding a support for a specific SQL database, I have trouble to see why we should create a dedicated configuration for `mariadb`. Like you said and like the code you write is doing,  this is exactly the same configuration than for mysql.

So I am sorry to ask, but what would be exactly the value to support this dedicated configuration as you can already use Perses proxy to contact a MariaDB datasource ?

Is it because, the driver is written `mysql` and it does not necessary mean it is compatible with `mariadb` ? If so, a simple improvement of the documentation should work, don't you think ?

ZPascal:
> Hello @ZPascal,
> 
> Thank you for your contribution. While I am not against adding a support for a specific SQL database, I have trouble to see why we should create a dedicated configuration for `mariadb`. Like you said and like the code you write is doing, this is exactly the same configuration than for mysql.
> 
> So I am sorry to ask, but what would be exactly the value to support this dedicated configuration as you can already use Perses proxy to contact a MariaDB datasource ?
> 
> Is it because, the driver is written `mysql` and it does not necessary mean it is compatible with `mariadb` ? If so, a simple improvement of the documentation should work, don't you think ?

Hello @nexucis,

Thank you for your feedback! I understand your point and agree that technically MariaDB uses the same MySQL driver.

However, I believe there's value in providing a dedicated `MariaDB` driver option:

**1. Improved User Experience:** When users work with MariaDB, having to specify driver: mysql is counterintuitive. Users expect to configure "MariaDB" when connecting to MariaDB, not "MySQL".

**2. Industry Standard:** Other established projects follow this pattern. For example, [Telegraf](https://github.com/influxdata/telegraf/blob/a7175df579623db3512664b676600cb0b659524c/plugins/inputs/sql/sql.go#L177) provides a dedicated mariadb driver option as an alias to MySQL. This makes configurations more transparent and intuitive.

**3. Self-Documenting Configuration:** While documentation would help, users would still need to remember that "MySQL" means both MySQL and MariaDB. A dedicated MariaDB option makes the configuration self-explanatory without requiring users to consult documentation.

The implementation is minimal (essentially just an alias), but it provides a meaningful usability improvement and follows established patterns.

That said, I'm open to the documentation-only approach if you prefer. I just wanted to make the case for the more intuitive setup.

What do you think?

pkg/model/api/v1/datasource/sql/sql_test.go
```diff
@@ -102,12 +277,315 @@ postgres:
...
+			},
+			expectErr: true,
+			errMsg:    "database cannot be empty",
+		},
+	}
+
+	for _, test := range testSuite {
+		t.Run(test.name, func(t *testing.T) {
+			err := test.config.validate()
+			if test.expectErr {
```
Nexucis: same comment for the way to manage the errors in these tests

internal/api/impl/proxy/proxy.go
```diff
@@ -558,11 +558,19 @@ func (s *sqlProxy) openMySQL(tlsConfig *tls.Config) (*sql.DB, error) {
...
 	}
 
-	if s.config.MySQL != nil {
-		mysqlConfig.Params = s.config.MySQL.Params
-		mysqlConfig.MaxAllowedPacket = s.config.MySQL.MaxAllowedPacket
-		mysqlConfig.ReadTimeout = s.config.MySQL.ReadTimeout
-		mysqlConfig.WriteTimeout = s.config.MySQL.WriteTimeout
+	// Use MariaDB config if driver is MariaDB, otherwise use MySQL config
+	var driverConfig *datasourceSQL.MySQLConfig
+	if s.config.Driver == datasourceSQL.DriverMariaDB && s.config.MariaDB != nil {
```
Nexucis: here you don't need to check the driver actually.

pkg/model/api/v1/datasource/sql/sql.go
```diff
@@ -149,7 +152,7 @@ func (s *Config) validate() error {
 		return errors.New("driver is required")
 	}
 
-	if s.Driver != DriverMySQL && s.Driver != DriverPostgreSQL {
+	if s.Driver != DriverMySQL && s.Driver != DriverMariaDB && s.Driver != DriverPostgreSQL {
```
Nexucis, line 182: We can add some extra validation in this function. We should check if the `MariaDB` config is not nil, then the driver `mariadb` is correctly set.

We could even set the driver by default first. If a config is set but not the driver, then we can set the matching driver. And then we can verify that the config set and the driver match.

So instead of starting the validation by

```go
	if s.Driver == "" {
		return errors.New("driver is required")
	}
```

We can have:

```go
	if s.Driver == "" {
		if s.MySQL != nil {
			s.Driver = DriverMySQL
		} else s.MariaDB != nil {
			s.Driver = DriverMariaDB
		} else if s.Postgres != nil {
			s.Driver = DriverPostgres
		}
	}

	// check again if the driver is still empty
	if s.Driver == "" {
		return errors.New("driver is required")
	}

	if s.Driver != DriverMySQL && s.Driver != DriverMariaDB && s.Driver != DriverPostgreSQL {
		return fmt.Errorf("driver %s is not supported", s.Driver)
	}

	// Then we need to verify that the driver and the config are actually matching.
	if s.Driver == DriverMySQL && (s.MariaDB != nil || s.Postgres != nil) {
		return errors.New("driver mysql cannot be set if mariaDB or postgres config is set ")
	}
	// same logic applied for other driver
[...]
```

pkg/model/api/v1/datasource/sql/sql_test.go
```diff
@@ -102,12 +277,315 @@ postgres:
...
+				SSLMode: "",
+			},
+			expectErr: false,
+		},
+	}
+
+	for _, test := range testSuite {
+		t.Run(test.name, func(t *testing.T) {
+			err := test.config.validate()
+			if test.expectErr {
```
Nexucis: I believe you can expect an error if you are expecting an error message. So you can actually simplify this part by having a single attribute `expectedErrorMessage` and checking the existence of this attribute:

```go
if len(test.expectedErrorMessage) > 0 {
	assert.Error(t,err)
	assert.Contains(t, err.Error(), test.expectedErrorMessage)
} else {
	assert.NoError(t, err)
}
```

Nexucis (commented):
Alright, thank you @ZPascal for your answer. We discussed about this topic with other maintainers and there is a consensus to be agree with you.

I have added few comments, can you take a look please ?

Let me know if you are fine with the propositions.

internal/api/impl/proxy/proxy.go
Nexucis: you could actually simplify even more this code, by setting first a config and then check if the other is not nil.

It will work if the config is correctly checked (see my other comment)

```go
// As the config is properly checked, we can set by default one config and then check if others are not nil as they cannot set both of them at the same time.
driverConfig := s.config.MySQL
if s.config.MariaDB != nil {
  driverConfig := s.config.MariaDB
}

Comment (ZPascal) [reply internal/api/impl/proxy/proxy.go]: Done

File: pkg/model/api/v1/datasource/sql/sql.go (hunk 3 above)
Comment (ZPascal) [reply pkg/model/api/v1/datasource/sql/sql.go:182]: Done

File: pkg/model/api/v1/datasource/sql/sql_test.go (hunk 4 above)
Comment (ZPascal) [reply pkg/model/api/v1/datasource/sql/sql_test.go]: Done

**Comment by ZPascal** [comment]
> Alright, thank you @ZPascal for your answer. We discussed about this topic with other maintainers and there is a consensus to be agree with you.
> 
> I have added few comments, can you take a look please ?
> 
> Let me know if you are fine with the propositions.

I've adapted the code base.

File: pkg/model/api/v1/datasource/sql/sql.go [hunk 5]
```diff
@@ -145,20 +148,52 @@ func (s *Config) UnmarshalYAML(unmarshal func(any) error) error {
 }
 
 func (s *Config) validate() error {
+	err := s.verifySupportedDriver()
+	if err != nil {
+		return err
+	}
```
Nexucis:
```suggestion
if err := s.verifySupportedDriver(); err != nil {
		return err
}
```

pkg/model/api/v1/datasource/sql/sql_test.go
```diff
@@ -102,12 +277,313 @@ postgres:
...
+			}
+		})
+	}
+}
+
+func TestPostgresConfigValidation(t *testing.T) {
+	testSuite := []struct {
+		name      string
+		config    PostgresConfig
+		expectErr bool
```
Nexucis: this boolean is not necessary anymore

pkg/model/api/v1/datasource/sql/sql_test.go
```diff
@@ -102,12 +277,313 @@ postgres:
...
+		errMsg    string
```
Nexucis: this should be defined as an expected error message. Can you please rename the field

pkg/model/api/v1/datasource/sql/sql_test.go
```diff
@@ -102,12 +277,313 @@ postgres:
...
+			},
+			expectErr: true,
+			errMsg:    "database cannot be empty",
+		},
+	}
+
+	for _, test := range testSuite {
+		t.Run(test.name, func(t *testing.T) {
+			err := test.config.validate()
+			if test.expectErr {
```
Nexucis: you can apply the same logic by checking only if there is an expected error message here

pkg/model/api/v1/datasource/sql/sql_test.go
ZPascal: Done

pkg/model/api/v1/datasource/sql/sql.go
ZPascal: Done

pkg/model/api/v1/datasource/sql/sql_test.go
ZPascal: Done

pkg/model/api/v1/datasource/sql/sql_test.go
ZPascal: Done

Nexucis (approved):
awesome, thank you for the contribution @ZPascal

Nexucis:
closing and reopening the PR to re-trigger the CI

PR #3815, opened 2026-01-26
[BREAKINGCHANGE] Improve SQL proxy error handling and switch to JSON responses

ZPascal:
# Description

This PR improves error handling and response format for SQL proxy data sources in Perses. The changes make SQL proxy errors more informative and user-friendly, and switch the response format from CSV to JSON for better frontend integration.

<details>
<summary>Implementation details</summary>

### 1. Enhanced Error Messaging for SQL Proxy

Previously, SQL proxy errors returned generic `InternalError` responses that didn't provide useful information to users. This PR replaces them with detailed HTTP error responses that include:

- **Authentication errors**: Returns `400 Bad Request` with message "Authentication setup failed" and the specific error
- **TLS configuration errors**: Returns `400 Bad Request` with message "TLS configuration failed" and the specific error  
- **Database connection errors**: Returns `400 Bad Request` with message "Database connection failed" and the specific error
- **SQL query errors**: Returns `400 Bad Request` with message "SQL query failed" and the specific error

**Before:**
```go
if err := s.setupAuthentication(); err != nil {
    logrus.WithError(err).Error("unable to setup authentication")
}
```

**After:**
```go
if err := s.setupAuthentication(); err != nil {
    logrus.WithError(err).Error("unable to setup authentication")
    return echo.NewHTTPError(http.StatusBadRequest, map[string]interface{}{
        "message": "Authentication setup failed",
        "error":   err.Error(),
    })
}
```

This applies to all SQL proxy error scenarios: authentication setup, TLS configuration, database connection, and query execution.

### 2. Switch SQL Proxy Response Format from CSV to JSON

Changed the default response format for SQL queries from CSV to JSON with structured column metadata and rows.

**Response Format:**
```json
{
  "columns": [
    {"name": "id", "type": "INT"},
    {"name": "username", "type": "VARCHAR"},
    {"name": "email", "type": "VARCHAR"}
  ],
  "rows": [
    {"id": 1, "username": "admin", "email": "admin@example.com"},
    {"id": 2, "username": "user", "email": "user@example.com"}
  ]
}
```

**Benefits:**
- Better frontend integration - JSON is the standard for API responses
- Column metadata included - the response includes column names and database types
- Type preservation - numeric and boolean values maintain their types
- Easier to parse - frontends can directly consume the JSON without parsing CSV

The CSV writer function (`writeCSVResponse`) is retained for potential future use but is no longer called by default.

### 3. Implementation Details

**New Function: `writeJSONResponse`**
- Extracts column names and types using `rows.ColumnTypes()`
- Builds structured column metadata array
- Collects all rows into a slice of maps
- Handles `[]byte` to `string` conversion for better JSON serialization
- Returns JSON with proper `Content-Type` header

**Error Response Structure:**
All SQL proxy errors now return a consistent structure:
```json
{
  "message": "Human-readable error category",
  "error": "Detailed error message"
}
```

</details>

## Testing

- Verified error messages are properly returned for authentication, TLS, connection, and query failures
- Tested JSON response format with various SQL queries (PostgreSQL)
- Confirmed column metadata is correctly extracted and included
- Validated backward compatibility - CSV function remains available if needed

## Breaking Changes

**This is a breaking change** for any clients that expect CSV responses from the SQL proxy endpoint.

- SQL proxy responses now return JSON instead of CSV by default
- Error responses now have a structured format with `message` and `error` fields instead of generic internal errors

**Migration Path:**
Frontends consuming SQL proxy responses should:
1. Update to parse JSON instead of CSV
2. Use the `columns` array for metadata
3. Iterate over the `rows` array for data
4. Update error handling to use the new structured error format

## Checklist

- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention.

## UI Changes

- [ ] N/A - This is a backend change only. However, frontends consuming SQL proxy responses will need to be updated to handle the new JSON format.

## Open Questions
- [x] How should the unused `writeCSVResponse` function be handled?

### DIFF
--- a/internal/api/impl/proxy/proxy.go
+++ b/internal/api/impl/proxy/proxy.go
@@ -54,21 +54,6 @@ import (
 	"golang.org/x/oauth2/clientcredentials"
 )
 
-// SQLColumnMetadata represents metadata for a single column in SQL result
-type SQLColumnMetadata struct {
-	Name string `json:"name"`
-	Type string `json:"type"`
-}
-
-// SQLRow represents a single row in an SQL result with column name to value mapping
-type SQLRow map[string]any
-
-// SQLResponse represents the complete SQL query response
-type SQLResponse struct {
-	Columns []SQLColumnMetadata `json:"columns"`
-	Rows    []SQLRow            `json:"rows"`
-}
-
 // projectForLog returns a meaningful log value for the project field.
 // For global datasources (where project is empty), it returns "<global>" to make logs clearer.
 func projectForLog(project string) string {
@@ -521,7 +506,7 @@ func (s *sqlProxy) serve(c echo.Context) error {
 			"project":    projectForLog(s.project),
 			"query":      q.Query,
 		}).Error("rejected query with write operations; only SELECT queries are allowed through the SQL proxy")
-		return apiinterface.BadRequestError
+		return apiinterface.HandleBadRequestError("only SELECT queries are allowed through the SQL proxy")
 	}
 
 	// add password if provided
@@ -742,6 +727,21 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
 	db := stdlib.OpenDB(*pgxConfig.ConnConfig)
 
 	return db, nil
+}
+
+// SQLColumnMetadata represents metadata for a single column in SQL result
+type SQLColumnMetadata struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+// SQLRow represents a single row in an SQL result with column name to value mapping
+type SQLRow map[string]any
+
+// SQLResponse represents the complete SQL query response
+type SQLResponse struct {
+	Columns []SQLColumnMetadata `json:"columns"`
+	Rows    []SQLRow            `json:"rows"`
 }
 
 func writeJSONResponse(c echo.Context, rows *sql.Rows, datasourceName, projectName string) error {


internal/api/impl/proxy/proxy.go
```diff
@@ -463,21 +463,30 @@ func (s *sqlProxy) serve(c echo.Context) error {
...
+			"error":   err.Error(),
+		})
 	}
 
 	// add tls.Config
 	tlsConfig, err := s.prepareTLSConfig()
 	if err != nil {
 		logrus.WithError(err).Error("unable to build the tls config")
-		return apiinterface.InternalError
+		return echo.NewHTTPError(http.StatusBadRequest, map[string]interface{}{
```
Nexucis: we already have a way to handle the HTTP error in Perses.

Besides the error type is not correct as it cannot be a bad request since it depends on the setup of the proxy. This is not related to the request itself. This is why we decided to use the `internal server error` as the user using the proxy potentially cannot do anything.

Bad request suggest your request is wrong and you need to modify it, and this is not the case actually since you cannot change it to fix the issue.

internal/api/impl/proxy/proxy.go
```diff
@@ -650,55 +662,122 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
...
+		columns[i] = map[string]string{
+			"name": col,
+			"type": dbType,
+		}
+	}
+
 	// Create a slice of interface{} to hold the column values
 	values := make([]any, len(cols))
-
-	// Create a slice of sql.RawBytes to hold the raw bytes of the column values
```
Nexucis, line 669: why did you remove this comment ? Do you think it's not accurate ?

internal/api/impl/proxy/proxy.go
```diff
@@ -650,55 +662,122 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
...
+		logrus.WithError(err).Error("unable to get column types")
+		return apiinterface.InternalError
+	}
+
+	// Build column metadata
+	columns := make([]map[string]string, len(cols))
+	for i, col := range cols {
+		dbType := colTypes[i].DatabaseTypeName()
+		columns[i] = map[string]string{
+			"name": col,
```
Nexucis: it would be great to have a proper struct we can then marshal in JSON. It's hard to understand what would be the final struct when you are defining on the fly the field.

While this is necessary to transform a generic SQL response to a JSON, you can still define properly the row struct that will handle this transformation.

internal/api/impl/proxy/proxy.go
```diff
@@ -650,55 +662,122 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
...
+		"columns": columns,
+		"rows":    rowsData,
+	}
+
+	// Set Content-Type and write JSON response
+	c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSON)
+	return c.JSON(http.StatusOK, response)
 }
+
+//func writeCSVResponse(c echo.Context, rows *sql.Rows) error {
```
Nexucis: we don't keep commented code. Git is used for that if we need to rollback. Please remove commented code. Thank you

ZPascal: Hi @Nexucis, this is only a temporary solution to fix the lint issues. If you are satisfied with the JSON response, I will clean up the code.

Nexucis: yes this is fine for me. Please go ahead @ZPascal and let me know when you are good to move forward on this PR. Thanks!

internal/api/impl/proxy/proxy.go
ZPascal: You're right, I understand the issue. The error type should remain as an `internal server error`, since it's a proxy setup issue, not a request validation issue. The user building dashboards can't fix this by modifying their request.

However, the end user doesn't necessarily have access to the logs. I'd at least like to include detailed information in the logs so that when someone with access to them investigates, they have enough context to understand what went wrong.

internal/api/impl/proxy/proxy.go
ZPascal: I removed the comment and forgot to publish the updated version. I have now fixed the problem.

internal/api/impl/proxy/proxy.go
ZPascal: I've implemented a generic structure to handle it.

internal/api/impl/proxy/proxy.go
Nexucis: A way to handle this part would be to test the connection of the datasource when defining it.
There was a tentative a couple years ago, but it was never finished and I still believe it is still accurate

ZPascal: @Nexucis, I'll think about it and try to improve the situation in a separate PR.

internal/api/impl/proxy/proxy.go
Nexucis: Just to be sure, @dougkirkley are you fine with these changes ? Does it break something on your side ?

internal/api/impl/proxy/proxy.go
```diff
@@ -449,56 +485,96 @@ func (s *sqlProxy) serve(c echo.Context) error {
...
 	}
 
+	// Validate that the query is read-only (SELECT only) to prevent data modification
+	if !isReadOnlyQuery(q.Query) {
+		logrus.WithFields(map[string]interface{}{
+			"datasource": s.name,
+			"project":    s.project,
+			"query":      q.Query,
+		}).Error("rejected query with write operations; only SELECT queries are allowed through the SQL proxy")
+		return apiinterface.InternalError
```
Nexucis: I think for this particular case we can return a bad request as the error can be fixed by the user.

internal/api/impl/proxy/proxy.go
```diff
@@ -656,55 +732,185 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
...
+			"project":    projectName,
+		}).Error("unable to get columns from query result")
 		return apiinterface.InternalError
 	}
 
+	colTypes, err := rows.ColumnTypes()
+	if err != nil {
+		logrus.WithError(err).WithFields(map[string]interface{}{
+			"datasource": datasourceName,
+			"project":    projectName,
```
Nexucis: it is possible `project` is empty because it is using a global datasource. I am wondering if it is fine to still have the field `project` with nothing.

What do you think ?

internal/api/impl/proxy/proxy.go
```diff
@@ -656,55 +732,185 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
...
+	// Remove comments to check if there's any actual query left
+	queryWithoutComments := removeSQLComments(normalizedQuery)
+	queryWithoutComments = strings.TrimSpace(queryWithoutComments)
+
+	// If nothing is left after removing comments, reject the query
+	if queryWithoutComments == "" {
+		return false
+	}
+
+	upperQuery := strings.ToUpper(normalizedQuery)
```
Nexucis: Why don't you use `queryWithoutComments` instead ?

I believe having a clean query would be safer to run and would allow more capabilities in the proxy.

Nexucis: So, what I am proposing is actually to remove any comments in the query and use the new string to analysed it for safety check and then use it afterwards to execute the query.

What do you think ?

internal/api/impl/proxy/proxy.go
ZPascal: I've updated the error.

internal/api/impl/proxy/proxy.go
ZPascal: I have implemented a function to set a default value `<global>` in such cases. What do you think?

internal/api/impl/proxy/proxy.go
ZPascal: >So, what I am proposing is actually to remove any comments in the query and use the new string to analysed it for safety check and then use it afterwards to execute the query.
>
> What do you think ?

Sounds good. I've integrated the change.

internal/api/impl/proxy/proxy.go
```diff
@@ -449,56 +494,99 @@ func (s *sqlProxy) serve(c echo.Context) error {
...
+	// Sanitize and validate that the query is read-only (SELECT only) to prevent data modification
+	// The cleaned query (without comments) is used for both validation and execution
+	cleanQuery, isValid := sanitizeAndValidateQuery(q.Query)
+	if !isValid {
+		logrus.WithFields(map[string]interface{}{
+			"datasource": s.name,
+			"project":    projectForLog(s.project),
+			"query":      q.Query,
+		}).Error("rejected query with write operations; only SELECT queries are allowed through the SQL proxy")
+		return apiinterface.BadRequestError
```
Nexucis:
```suggestion
		return apiinterface.HandleBadRequestError("only SELECT queries are allowed through the SQL proxy")
```

Nexucis: that would give a meaningful error message for the end user

internal/api/impl/proxy/proxy.go
```diff
@@ -54,6 +54,30 @@ import (
 	"golang.org/x/oauth2/clientcredentials"
 )
 
+// SQLColumnMetadata represents metadata for a single column in SQL result
+type SQLColumnMetadata struct {
```
Nexucis: I would put these structs right before the function transforming the SQL response to JSON

internal/api/impl/proxy/proxy.go
Nexucis: looks good, thank you for changing that

Nexucis:
I think once the latest the comments are addressed, we will be good to merge this one. Thank you @ZPascal for your meaningful contributions !

internal/api/impl/proxy/proxy.go
ZPascal: I have updated the error message

internal/api/impl/proxy/proxy.go
ZPascal: Done

ZPascal:
> awesome, lgtm. Thank you @ZPascal

The next PR (generic RDS plugin) will follow as soon as possible. :rocket:

PR #3820, opened 2026-01-28
[ENHANCEMENT] Go client updated to return PublicUser instead of User

celian-garcia:
# Description
This PR ensures that when the endpoint ``/api/v1/users/me`` is called in a Kubernetes authentication context, it no longer returns an error stating that the username is invalid. Kubernetes service accounts may contain colon characters, which are not valid in ``metadata.name``, and this causes issues when unmarshalling the server response.

> An alternative fix would be to allow UTF‑8 characters in metadata.name. This will eventually be required anyway to fully avoid such decoding issues.
> This would have to be done with an extra care, as we want to follow the same convention than kubernetes in order to allow some mapping and kubernetes does not support utf-8. Ref: #3237  

Additionally, since the API already returns a ``PublicUser`` object, it makes sense for the Go client to rely on the same structure for consistency.

# Screenshots

# Checklist

## UI Changes
N/A

### DIFF
--- a/cue/model/api/v1/metadata_go_gen.cue
+++ b/cue/model/api/v1/metadata_go_gen.cue
@@ -4,7 +4,24 @@
 
 package v1
 
-#Metadata: _
+import "time"
+
+#Metadata: {
+	name: string @go(Name)
+
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:validation:Type=string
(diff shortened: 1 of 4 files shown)


celian-garcia:
Doesn't work yet so I moved it to draft for now

Nexucis:
closed in favor of #3822 which has a preferred way to handle this situation

PR #3822, opened 2026-01-28
[ENHANCEMENT] Go client updated to return PublicUser instead of User

celian-garcia:
### Summary

Duplicate of **#3820**, but using a different strategy.  
Same context and same goal, except instead of removing unmarshal validation from all structs using `Metadata`, this PR introduces a dedicated `PublicMetadata` type for the `Public*` structs.

### Why

`Public*` structs are only used for **response payloads**, never for **create/update** operations. They don’t need unmarshal‑time validation.

### What this PR does

*   Add `PublicMetadata`.
*   Use it in all `Public*` structs that contains metadata.
*   Keep validation untouched for internal/write structs.

# Screenshots

# Checklist

## UI Changes

N/A

### DIFF
--- a/pkg/model/api/v1/metadata.go
+++ b/pkg/model/api/v1/metadata.go
@@ -67,17 +67,6 @@ func (m *Metadata) Flatten(sensitive bool) {
 	}
 }
 
-func NewProjectMetadata(project string, name string) *ProjectMetadata {
-	return &ProjectMetadata{
-		Metadata: Metadata{
-			Name: name,
-		},
-		ProjectMetadataWrapper: ProjectMetadataWrapper{
-			Project: project,
-		},
-	}
-}
-
 func (m *Metadata) UnmarshalJSON(data []byte) error {
 	var tmp Metadata
 	type plain Metadata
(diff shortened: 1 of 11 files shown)


pkg/model/api/v1/metadata.go
```diff
@@ -67,17 +67,6 @@ func (m *Metadata) Flatten(sensitive bool) {
 	}
 }
 
-func NewProjectMetadata(project string, name string) *ProjectMetadata {
```
celian-garcia, line 70: This is not a removal, it's been moved down to keep structs alongside their related functions and methods

internal/cli/cmd/whoami/whoami.go
```diff
@@ -81,34 +74,12 @@ func (o *option) Execute() error {
...
-		if err != nil {
-			return "", err
-		}
-
-		return result.Metadata.Name, nil
-	}
-
-	result := &v1.PublicUser{}
-	err := res.Object(result)
+	user, err := o.apiClient.V1().User().Get(utils.PathMe)
```
Nexucis, line 77: this looks weird to me actually. What would happen if your user is named `me` actually ?

celian-garcia: Nobody will be able to see him/her 😄 
This is something that has been introduced by the previous PR. If you don't want it you need another endpoint, like /api/v1/users-me or something.

Nexucis: and what about `/api/v1/user/whoami` ?

celian-garcia: I'm not fan of the confusion user / users it would introduce.
But I see where you're going. /api/v1/whoami is probably already clear enough no ?

Nexucis: we could even have `/api/whoami` then. I don't really the need to version this endpoint

celian-garcia: Since it returns a PublicUser which is a v1 model, maybe we have to ? I let you decide. 
Can you merge that PR ? So I can create the other.

PR #3826, opened 2026-01-30
[ENHANCEMENT] New endpoint `/api/v1/user/whoami`

celian-garcia:
# Description

This PR aims to fix a problem with the ``/api/v1/users/me`` introduced previously in the k8s feature branch. 
Indeed it cancel the possibility for the users to be named "me".

I propose to create a totally new "whoami" endpoint that would retrieve exactly the same data than  ``/api/v1/users/me``.

# Screenshots

image
image

# Checklist

## UI Changes

### DIFF
--- a/internal/api/generate.go
+++ b/internal/api/generate.go
@@ -422,9 +422,7 @@ func generatePersistence(ept endpoint) {
 func generateClient(ept endpoint) {
 	folder := "../../pkg/client/api/v1/"
 	fileName := fmt.Sprintf("%s.go", ept.PackageName)
-	// as the endpoint is generated and not adds in git, the client that reflects exactly what is exposed by the endpoint,
-	// then should be also ignored by git, and so we can override it.
-	generateFile(folder, fileName, clientTemplate, ept, true)
+	generateFile(folder, fileName, clientTemplate, ept, false)
 }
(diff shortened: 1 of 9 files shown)


jgbernalp:
cc @PeterYurkovich

pkg/client/api/v1/whoami.go
```diff
@@ -0,0 +1,47 @@
...
+package v1
+
+import (
+	"github.com/perses/perses/pkg/client/perseshttp"
+	v1 "github.com/perses/perses/pkg/model/api/v1"
+)
+
+const whoAmIResource = "whoami"
+
+type WhoAmIInterface interface {
```
Nexucis: instead of creating a whole interface for that, I would keep that in the user interface & endpoint.

In the HTTP client I would just put the function `WhoAmI`. That would simplify everything and it will show this is related to the users.

And still, I think I prefer `/api/v1/user/whoami` for any technical endpoint related to the user.

celian-garcia: I need to check if the fact to have two base user and users won't be a problem in the implementation but I guess it's feasible

PeterYurkovich (approved):

PR #3831, opened 2026-02-03
[BUGFIX] i18n locales files must be prefixed by api_prefix

### LINKED ISSUE #3828: api_prefix is not respected for locales
### What did you do?

api_prefix is not getting added to /locales/en.dashboard.json, due to which datasources aren’t getting created in global datasources

It should have worked ideally

### System information

_No response_

### Perses version

0.53.0-rc.0

### Perses configuration file

```yaml

```

### Logs

```text

```

Nexucis:
fix #3828

### DIFF
--- a/ui/app/src/i18n/i18n.ts
+++ b/ui/app/src/i18n/i18n.ts
@@ -14,6 +14,7 @@
 import i18n from 'i18next';
 import { initReactI18next } from 'react-i18next';
 import HttpBackend, { HttpBackendOptions } from 'i18next-http-backend';
+import { PERSES_APP_CONFIG } from '../config';
 
 i18n
   .use(HttpBackend)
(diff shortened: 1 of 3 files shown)


ui/app/src/i18n/i18n.ts
```diff
@@ -25,7 +26,7 @@ i18n
     ns: ['dashboard'],
     defaultNS: 'dashboard',
     backend: {
-      loadPath: '/locales/{{lng}}.{{ns}}.json',
+      loadPath: `${PERSES_APP_CONFIG.api_prefix}/locales/{{lng}}.{{ns}}.json`,
```
jgbernalp, line 29: This allows keeping support when Perses backend is on a different path than the ui or embedded app

```suggestion
      loadPath: `${window.PERSES_PLUGIN_ASSETS_PATH || PERSES_APP_CONFIG?.api_prefix || ""}/locales/{{lng}}.{{ns}}.json`,
```

Nexucis: I am sorry, I don't get why the load path should include the `PERSES_PLUGIN_ASSETS_PATH` while there is nothing related to the plugin and this is only for the internal application that cannot be embedded.

jgbernalp: Sorry I was confused. Plugins will also have to support i18n at some point, but that is a future battle. This is totally ok.

PR #3839, opened 2026-02-04
[BUGFIX] Store refreshed refresh token

jirikuncar:
# Description

Store (possibly) updated refresh token together with access token.

# Screenshots

N/A

# Checklist

## UI Changes

### DIFF
--- a/internal/cli/config/config.go
+++ b/internal/cli/config/config.go
@@ -151,12 +151,6 @@ func SetProject(project string) error {
 	})
 }
 
-func SetAccessToken(token string) error {
-	return Write(&Config{
-		RestClientConfig: config.RestConfigClient{Authorization: secret.NewBearerToken(token)},
-	})
-}
-
 func SetTokens(accessToken, refreshToken string) error {
 	return Write(&Config{
 		RestClientConfig: config.RestConfigClient{Authorization: secret.NewBearerToken(accessToken)},


internal/cli/cmd/refresh/refresh.go
```diff
@@ -55,7 +55,7 @@ func (o *option) Execute() error {
 	if err != nil {
 		return err
 	}
-	if writeErr := config.SetAccessToken(response.AccessToken); writeErr != nil {
```
Nexucis, line 58: you can remove the function `SetAccessToken` since it is not used anymore.

Nexucis:
The refresh token is not changed during the lifecycle of refreshing the access token, but since it is returned by the API, it does not hurt to store it again.
Thank you for the improvement ! I will merge this one once you are removing the unused function.

jirikuncar:
> The refresh token is not changed during the lifecycle of refreshing the access token, but since it is returned by the API, it does not hurt to store it again.

Some OAuth servers are updating refresh token when it was used 🤷

internal/cli/cmd/refresh/refresh.go
jirikuncar: Done

Nexucis:
ah yeah good point, I was looking at the native authentication and forget about the oauth. Sorry about that.

PR #3842, opened 2026-02-05
[ENHANCEMENT] Fix delegated auth error route & move auth check modifying the router

Nexucis:
Things modifying, activating the route should stay in the router and should not have been moved to the views at first place.

Also a new route has been added without modifying the file `ui/endpoint.go`. 

This PR aims to fix these two points


PeterYurkovich:
Thanks for fixing this

PR #3860, opened 2026-02-10
[BUGFIX] Kube discovery: fix issue with inline references in CUE

AntoineThebaud:
# Description

https://github.com/perses/plugins/commit/5b1f96f166a846330caec30c0e0216afaa31c42f broke the kube datasource discovery logic because for some reason, while the imported definitions were properly resolved with the former schema definition, the  introduced `commonProxy.#baseHTTPDatasourceSpec` that factorizes the disjunction "directUrl | proxy", used inline, doesn't get resolve & breaks the logic (see failed pipeline https://github.com/perses/perses/actions/runs/21826529550/job/62972914721?pr=3859)

This PR fixes that by calling `Eval()` on the CUE values manipulated, in order to ensure such references get resolved before further processing happens.

# Checklist

### DIFF
--- a/internal/api/discovery/cuetils/node.go
+++ b/internal/api/discovery/cuetils/node.go
@@ -23,6 +23,7 @@ import (
 	"cuelang.org/go/cue"
 	"github.com/perses/perses/pkg/model/api/v1/datasource"
 	"github.com/perses/perses/pkg/model/api/v1/datasource/http"
+	"github.com/sirupsen/logrus"
 )
 
 type NodeType string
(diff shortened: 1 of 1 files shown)


internal/api/discovery/cuetils/node.go
```diff
@@ -270,7 +276,10 @@ func buildMapFromStructValue(v cue.Value) map[string]cue.Value {
 	result := make(map[string]cue.Value)
 	it, _ := v.Fields()
 	for it.Next() {
-		result[it.Selector().String()] = it.Value()
+		// We have to evaluate the value to resolve eventual reference in the schema before going further.
+		// Otherwise inline refs (at least) are not resolved and "block" the tree construction.
+		valueWithResolvedRefs := it.Value().Eval()
```
AntoineThebaud, line 282: should we check (& log) errors here too?

Nexucis: yes that would be great

PR #3863, opened 2026-02-10
[BUGFIX] Fix native auth

Gladorme:
# Description

Fixing native authentication broken after PR #3545. I am not sure how delegated auth works and how to check it manually, it may need to do a small tweak in `useCurrentUser` to not do the `enabled` check for `isAccessTokenExist` if delegated is enabled. But if it's like native auth, it's should be fine :)

# Screenshots

# Checklist

### DIFF
--- a/ui/app/src/Router.tsx
+++ b/ui/app/src/Router.tsx
@@ -47,7 +47,7 @@ import HomeView from './views/home/HomeView';
 // Default route is eagerly loaded
 import App from './App';
 import { PERSES_APP_CONFIG } from './config';
-import { buildRedirectQueryString, useIsLoggedIn } from './model/auth/auth-client';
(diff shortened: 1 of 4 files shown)


ui/app/src/Router.tsx
Nexucis, line 230: maybe this can be simplified with the first guard ?

Nexucis: got my answer by tchat, seems we cannot simplify it because of the navigate

Nexucis:
since Guillaume didn't touch about this part and he preserves the usage of the `whoAmI` endoint, I believe we should be good.

PeterYurkovich (approved):

PeterYurkovich:
The `AlreadyLoggedIn` component handles the delegated auth, so there is no issue there

Issue #1733, opened 2024-01-29
Dashboard tags


AntoineThebaud:
### Proposal

We probably want the ability to tag dashboards, to enable another convenient way to search for the right one(s).

ibakshay:
@Nexucis Thank you very much for the summarisation.
Let us maybe discuss this feature in the next weekly meeting and brainstorm further with others. 

@jgbernalp Would love to get your feedback, especially on the `perses-operator` side :)

nicolastakashi:
I'm wondering if annotations would be a better place to store tags, usually annotations in Kubernetes are used to applications store informations in there and retrieve it later.

It would fit nice as a label if it was only one label, but since a dash can have as many labels as the user decide, I'd probably store under annotations with map of key value.

wdyt?

dougkirkley:
I guess I could see something like that working.
I was looking up some examples and there are a few in the k8s docs around comma delimited lists.

https://kubernetes.io/docs/reference/labels-annotations-taints/#applyset-kubernetes-io-additional-namespaces

```
Kind: Dashboard
metadata:
  annotations:
    dashboard.perses.dev/tags: "oncall,high_severity"
```

dougkirkley:
Also, I already had a PR merged into the operator side to enable adding annotations to the resources, so that should be supported already without any changes.

jgbernalp:
I think adding a new `tags` field in the metadata will be beneficial for the UI, I think is great to be able to group dashboards in addition to the current hierarchical project structure. From the operator side, we are already doing some mapping between Perses metadata and k8s metadata, so it would be a good idea to follow @dougkirkley example with annotations. Nothing stops users from adding more labels or annotations in the CR, but we would have a clear place to store and sync between the intended tags for dashboards created in Perses. I think we should use `tags` instead of labels as labels can be used for other purposes specific to k8s, like reconciling a dashboard in a specific instance.

nicolastakashi:
I'm wondering if UI couldn't read from annotations as well, I don't have a strong opinion on having tags as part of the spec, but it looks like metadata, so would be nice store on metadata and teach Ui leverage the metadata no?

Nexucis:
On my side, I prefer to keep the UI stupid and simple with no parsing and convention.

But overall, it seems we are reaching a consensus: 

## In Perses core (perses/perses)

- Tags are in the metadata
- Tags are in a dedicated field named `tags` like 

```yaml
Kind: Dashboard
metadata:
  name: test
  tags:
[...]

PR #3865, opened 2026-02-10
[FEATURE]: Add/Modify tags for Dashboard

ibakshay:
# Description

With this PR, 
1. Tags can be added/modified to dashboard
2. Dashboards can be filtered based on the tags. 

closes https://github.com/perses/perses/issues/1733

# Screenshots

## Demo 1- Create,modify and filter by tags in dashboards list
https://github.com/user-attachments/assets/00f8ca35-1ecb-487c-a585-74bee32492a7

## Demo 2 - Filter by tags in the central search bar 

https://github.com/user-attachments/assets/907dbd37-bba1-473e-b9b1-39a68377e0d6

# Checklist

## UI Changes

### DIFF
--- a/internal/cli/cmd/dac/build/testdata/go/go.mod
+++ b/internal/cli/cmd/dac/build/testdata/go/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/muhlemmer/gu v0.3.1 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
-	github.com/perses/common v0.30.0 // indirect
+	github.com/perses/common v0.30.1 // indirect
 	github.com/prometheus/client_golang v1.23.2 // indirect
 	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/prometheus/common v0.67.5 // indirect
--- a/internal/cli/cmd/dac/build/testdata/go/go.sum
+++ b/internal/cli/cmd/dac/build/testdata/go/go.sum
@@ -26,8 +26,8 @@ github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f h1:KUppIJq7/+
 github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f/go.mod h1:qRWi+5nqEBWmkhHvq77mSJWrCKwh8bxhgT7d/eI7P4U=
 github.com/nexucis/lamenv v0.5.2 h1:tK/u3XGhCq9qIoVNcXsK9LZb8fKopm0A5weqSRvHd7M=
 github.com/nexucis/lamenv v0.5.2/go.mod h1:HusJm6ltmmT7FMG8A750mOLuME6SHCsr2iFYxp5fFi0=
-github.com/perses/common v0.30.0 h1:Xf3PUZx8Sh1Kr6EkudcxCDCCz5c1JUmI5UHIZ4h2ZP4=
-github.com/perses/common v0.30.0/go.mod h1:DFtur1QPah2/ChXbKKhw7djYdwNgz27s5fPKpiK0Xao=
+github.com/perses/common v0.30.1 h1:XjCHCrBOyXacDA+fa1pzlq+pNiPazTj4anUhbYGU0ho=
+github.com/perses/common v0.30.1/go.mod h1:DFtur1QPah2/ChXbKKhw7djYdwNgz27s5fPKpiK0Xao=
 github.com/perses/plugins/prometheus v0.56.0 h1:5wy5BV6i76OzDA411ebI2AakwOiG9V5BCo1LKHZwcGM=
 github.com/perses/plugins/prometheus v0.56.0/go.mod h1:YwDoyIkr7pxVyNQD9CP320tAaYqv6rIXOStGejfMaI8=
 github.com/perses/plugins/timeserieschart v0.11.2 h1:aNiX2pOz0CWUQkSbX4b0VXDDrD29EsjDW3EP5esRcU0=


ui/app/src/components/DashboardList/RecentDashboardList.tsx
```diff
@@ -101,6 +102,21 @@ export function RecentDashboardList(props: RecentDashboardListProperties): React
     () => [
       { field: 'project', headerName: 'Project', type: 'string', flex: 2, minWidth: 150 },
       { field: 'displayName', headerName: 'Display Name', type: 'string', flex: 3, minWidth: 150 },
+      {
+        field: 'tags',
+        headerName: 'Tags',
+        flex: 2,
+        minWidth: 150,
+        sortable: false,
+        filterable: false,
```
jgbernalp: I think filter by tags would make total sense here

ui/app/src/components/DashboardList/RecentDashboardList.tsx
```diff
@@ -101,6 +102,21 @@ export function RecentDashboardList(props: RecentDashboardListProperties): React
...
+        renderCell: (params): ReactElement => (
+          <Box sx={{ display: 'flex', gap: 0.5, flexWrap: 'wrap', alignItems: 'center' }}>
```
jgbernalp: It seems we are missing some styling here, the tags seem aligned to the top of the cell rather than the center

pkg/model/api/v1/metadata.go
```diff
@@ -40,6 +42,7 @@ type Metadata struct {
 	// +kubebuilder:validation:Optional
 	UpdatedAt time.Time `json:"updatedAt" yaml:"updatedAt"`
 	Version   uint64    `json:"version" yaml:"version"`
+	Tags      []string  `json:"tags,omitempty" yaml:"tags,omitempty"`
```
jgbernalp: If this is optional  and max 20 we should add 
```suggestion
    // +kubebuilder:validation:Optional
    // +kubebuilder:validation:MaxItems=20
	Tags      []string  `json:"tags,omitempty" yaml:"tags,omitempty"`
```

pkg/model/api/v1/metadata.go
```diff
@@ -94,7 +97,39 @@ func (m *Metadata) UnmarshalYAML(unmarshal func(any) error) error {
...
+		if len(strings.TrimSpace(tag)) == 0 {
+			return fmt.Errorf("tag cannot be empty")
+		}
+		if strings.TrimSpace(tag) != tag {
+			return fmt.Errorf("tag %q cannot start or end with whitespace", tag)
+		}
+		if utf8.RuneCountInString(tag) > maxTagLength {
+			return fmt.Errorf("tag %q cannot contain more than %d characters", tag, maxTagLength)
+		}
+		if _, ok := seenTags[tag]; ok {
```
jgbernalp: We should probably validate non-case sensitive:

```go
normalizedTag := strings.ToLower(tag)
```

ui/app/src/components/dialogs/EditDashboardDialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+                render={({ field, fieldState }) => (
+                  <Autocomplete
+                    {...field}
+                    multiple
+                    freeSolo
+                    options={[]}
+                    value={field.value ?? []}
+                    onChange={(_, newValue) =>
+                      field.onChange(
+                        Array.from(new Set(newValue.map((tag) => tag.trim()).filter((tag) => tag.length > 0)))
```
jgbernalp: Same here, we should probably convert the tags to lowercase to match the backend checks

pkg/model/api/v1/metadata.go
```diff
@@ -94,7 +97,39 @@ func (m *Metadata) UnmarshalYAML(unmarshal func(any) error) error {
...
+func validateTags(tags []string) error {
+	const maxTags = 20
+	const maxTagLength = 50
+
+	if len(tags) > maxTags {
+		return fmt.Errorf("cannot contain more than %d tags", maxTags)
+	}
+
+	seenTags := make(map[string]struct{}, len(tags))
+	for _, tag := range tags {
```
jgbernalp: We should validate the tags have valid characters, probably only characters, space, `-` and `_`

ui/app/src/components/dialogs/EditDashboardDialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+  useEffect(() => {
+    if (!open) {
+      return;
+    }
+
+    form.reset({
+      dashboardName: getResourceDisplayName(dashboard),
+      tags: dashboard.metadata.tags ?? [],
+    });
+  }, [dashboard, form, open]);
```
jgbernalp:
```suggestion
  }, [dashboard, form.reset, open]);
```

ui/app/src/components/DashboardList/RecentDashboardList.tsx
ibakshay: Yes, fixed in https://github.com/perses/perses/pull/3865/commits/bb2ee5b3ae5a48a10e86677120cac22387f49707

image

pkg/model/api/v1/metadata.go
ibakshay: Added in https://github.com/perses/perses/pull/3865/changes/bb2ee5b3ae5a48a10e86677120cac22387f49707

pkg/model/api/v1/metadata.go
ibakshay: good catch. fixed in https://github.com/perses/perses/pull/3865/changes/bb2ee5b3ae5a48a10e86677120cac22387f49707

ui/app/src/components/dialogs/EditDashboardDialog.tsx
ibakshay: Fixed in https://github.com/perses/perses/pull/3865/changes/bb2ee5b3ae5a48a10e86677120cac22387f49707

pkg/model/api/v1/metadata.go
ibakshay: Yes, validating with this regex validator. 
https://github.com/perses/perses/blob/a0dbdce1b78e847343d4470772e96d19457061a8/pkg/model/api/v1/metadata.go#L131-L132

https://github.com/perses/perses/blob/a0dbdce1b78e847343d4470772e96d19457061a8/pkg/model/api/v1/metadata.go#L33

ui/app/src/components/dialogs/EditDashboardDialog.tsx
ibakshay: I implemented this little differently. 
https://github.com/perses/perses/blob/bb2ee5b3ae5a48a10e86677120cac22387f49707/ui/app/src/components/dialogs/EditDashboardDialog.tsx#L47-L60

Destructuring reset makes the effect depend on exactly what it uses, instead of the full form object. This avoids unnecessary reruns and makes [exhaustive-deps lint](https://react.dev/reference/eslint-plugin-react-hooks/lints/exhaustive-deps) happy.

ibakshay:
> Quick comment: I don't think #1733 should be considered closed until we also have the search by tag available from the central search bar

Sure. I already had this feature in mind ;)

ui/app/src/components/DashboardList/RecentDashboardList.tsx
ibakshay: yes good point. Fixed in https://github.com/perses/perses/pull/3865/commits/bb2ee5b3ae5a48a10e86677120cac22387f49707

image

ibakshay:
> Quick comment: I don't think #1733 should be considered closed until we also have the search by tag available from the central search bar

I have implemented now. Please watch the demo 2 in the PR description.  @AntoineThebaud

ui/app/src/validation/dashboard.ts
```diff
@@ -22,16 +22,32 @@ export const dashboardDisplayNameValidationSchema = z
   .min(1, 'Required')
   .max(75, 'Must be 75 or fewer characters long');
 
+export const tagsValidationSchema = z
+  .array(
+    z
+      .string()
+      .trim()
+      .min(1, 'Tag cannot be empty')
```
jgbernalp, line 30: It seems that by the backend validation we can create a tag with one letter
```suggestion
      .min(0, 'Tag cannot be empty')
```

ibakshay: @jgbernalp This suggestion will allow the empty string to be added as a tag. Backend does not allow this. 
image

jgbernalp: I misread the validation, this is fine

Nexucis:
can you incorporate these changes please: https://github.com/perses/perses/pull/2689 ?

pkg/model/api/v1/metadata.go
```diff
@@ -94,7 +102,44 @@ func (m *Metadata) UnmarshalYAML(unmarshal func(any) error) error {
...
+
+func validateTags(tags []string) error {
+	const maxTags = 20
+	const maxTagLength = 50
+
+	if len(tags) > maxTags {
+		return fmt.Errorf("cannot contain more than %d tags", maxTags)
+	}
+
+	seenTags := make(map[string]struct{}, len(tags))
```
Nexucis: you can actually use the set structure here. I think it will makes sense to use it to avoid the duplication of the tags

pkg/model/api/v1/metadata.go
```diff
@@ -27,6 +30,8 @@ func NewMetadata(name string) *Metadata {
 	}
 }
 
+var tagsAllowedCharactersRegexp = regexp.MustCompile(`^[a-z0-9 _-]+$`)
```
Nexucis, line 34: what is the reason to not allow all character here ? I believe here at least we can accept any UTF-8 characters

ui/app/src/components/Header/SearchBar/SearchList.tsx
Nexucis: why don't you change this list instead of adding a specific indexedKeys for the dashboards ? Tags is not something only for dashboards but for any resources managed by Perses.

pkg/model/api/v1/metadata.go
ibakshay: I added this constraint after this review comment. https://github.com/perses/perses/pull/3865#discussion_r2792843418 from @jgbernalp 
I don't have a strong opinion. I will simply implement what we all agree upon.

pkg/model/api/v1/metadata.go
Nexucis: actually I was more talking about the field metadata.tags. You can use the set struct directly in the metadata. That would simplify the whole code.

ui/app/src/components/Header/SearchBar/SearchList.tsx
ibakshay: Oh, is it? I did not know. 
[This issue description](https://github.com/perses/perses/issues/1733) mentions only`Dashboard` perses resource. 

Should I also have to display tags in all the perses resources? For example in `GlobalRoles`? 
image

Nexucis: well if it's not complicated to display them in other resources that would be great indeed !

pkg/model/api/v1/metadata.go
Nexucis: @jgbernalp do you have any reasons why we would not accept any UTF-8 characters ?

pkg/model/api/v1/metadata.go
ibakshay: Hi @Nexucis,
I looked into using set.Set[string] directly for the metadata.Tags field. However, the set.Set type in https://github.com/perses/common/blob/main/set/set.go currently only implements MarshalJSON and UnmarshalJSON, but not the YAML MarshalYAML /UnmarshalYAML.

Since Metadata supports both JSON and YAML (un)marshalling, directly using set.set[string] would break YAML parsing.

ui/app/src/components/Header/SearchBar/SearchList.tsx
ibakshay: > why don't you change this list instead of adding a specific indexedKeys for the dashboards ? Tags is not something only for dashboards but for any resources managed by Perses.

Currently, Only these resourceTypes are searchable from the central searchbar. 
https://github.com/perses/perses/blob/0d8c2652b83a33dceb6e1cd513f8e32ae3fcfc86/ui/app/src/components/Header/SearchBar/SearchBar.tsx#L37
Also, Adding/modifying tags for all other resourceTypes involve some UX as it has to be done in the edit wizard. 
Shall we create a separate issue for this feature and I can work on a separate PR. Otherwise, This PR will become too big.

pkg/model/api/v1/metadata.go
Nexucis: indeed you are right, I will address a PR to change that then.

ui/app/src/components/Header/SearchBar/SearchList.tsx
Nexucis: yeah good idea to address that in a different PR.

Nexucis: but still I would modify this list even if it's not supported for the other resources. The search will handle it if the keys do not exist.

Nexucis: I have did it with https://github.com/perses/common/pull/322. You can update the PR now :)

pkg/model/api/v1/metadata.go
ibakshay: Thank you. Could you please do a release as well in the common repo? @Nexucis

ui/app/src/components/Header/SearchBar/SearchList.tsx
ibakshay: > why don't you change this list instead of adding a specific indexedKeys for the dashboards ? Tags is not something only for dashboards but for any resources managed by Perses.

Thank you for the suggestion. I have addressed this review comment.

pkg/model/api/v1/metadata.go
ibakshay: done in https://github.com/perses/perses/pull/3865/changes/a71ff2e362c0d59cb5c5616c3a9b2cefb7e04136. Using pseudo-version pointing to the latest commit `b16ed8f4f08a69322e504c7726d057b3b6a0054` in common repository for now.

pkg/model/api/v1/metadata.go
```diff
@@ -94,7 +103,37 @@ func (m *Metadata) UnmarshalYAML(unmarshal func(any) error) error {
...
 func (m *Metadata) validate() error {
-	return common.ValidateID(m.Name)
+	if err := common.ValidateID(m.Name); err != nil {
+		return err
+	}
+
+	return validateTags(m.Tags)
+}
+
+func validateTags(tags set.Set[string]) error {
```
Nexucis:
```suggestion
func (m *Metadata) validateTags() error {
```

I believe it makes more sense it is attached to the metadata as it validates its fields

ibakshay: Fixed in https://github.com/perses/perses/pull/3865/changes/c1fab03ddb6d64cb66793a3e317ef17bce59a04c

go-sdk/test/go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/muhlemmer/gu v0.3.1 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
+	github.com/perses/common v0.30.0 // indirect
```
Nexucis: I believe you need to run again `go mod tidy` here. and in `internal/cmd/dac/build/testdata` and you should be good

Nexucis:
> can you incorporate these changes please: #2689 ?

one last thing: can you do that ☝️. You will see it includes the grafana migration.

go-sdk/test/go.mod
ibakshay: New release needs to be done in https://github.com/perses/common/releases/tag/v0.30.0. This release doesn't have this PR changes - https://github.com/perses/common/pull/322

Nexucis: v0.30.1 has been published

ibakshay:
> > can you incorporate these changes please: #2689 ?
> 
> one last thing: can you do that ☝️. You will see it includes the grafana migration.

incorporated in https://github.com/perses/perses/pull/3865/changes/6d2a8417a142c8554ce50a48af54a56e32a52cfc

PR #3866, opened 2026-02-11
[ENHANCEMENT] use project scope for non k8s perses resources

jgbernalp:
# Description

When using the k8s auth. By default, admins and other users are not allowed to create variables or read secrets. This makes Perses unusable for installations through the Perses operator, as datasources and dashboards fail to reconcile if including variables or secrets. Initially the guest permissions were used, but this is now a property only available in the native auth provider.

Since the Perses operator does not currently have support for these resources, we could do the access check using the project scope to keep Perses usable securely.

# Checklist


jgbernalp:
cc @PeterYurkovich

internal/api/authorization/k8s/translation.go
```diff
@@ -37,20 +37,48 @@ const (
...
+// globalScopesToCheck contains all scopes that should be checked at the wildcard (all-namespace)
+// level when computing permissions. This includes both global-only scopes and project-scoped
+// resources that benefit from an initial wildcard check to avoid redundant per-namespace checks.
 var globalScopesToCheck = []v1Role.Scope{
 	v1Role.GlobalDatasourceScope,
+	v1Role.GlobalVariableScope,
+	v1Role.GlobalSecretScope,
+	v1Role.GlobalRoleScope,
+	v1Role.GlobalRoleBindingScope,
+	v1Role.UserScope,
```
PeterYurkovich: Since the User, Role and RoleBinding's are disabled on Perses side when k8s are enabled, I think we shouldn't return their permissions

internal/api/authorization/k8s/k8s.go
```diff
@@ -250,12 +250,6 @@ func (k *k8sImpl) HasPermission(ctx echo.Context, requestAction v1Role.Action, r
 		return false
 	}
 
-	scope := getK8sScope(requestScope)
-	if scope == "" {
```
PeterYurkovich, line 254: We can leave this check for just the User, Role and RoleBinding's. Maybe with log included

internal/api/authorization/k8s/k8s.go
```diff
@@ -383,6 +377,14 @@ func getUnknownActions(knownActions []v1Role.Action) []v1Role.Action {
 
 func (k *k8sImpl) checkSpecificPermission(ctx echo.Context, namespace string, user user.Info, action v1Role.Action, scope v1Role.Scope) (authorized authorizer.Decision, err error) {
 	translatedK8sScope := getK8sScope(scope)
+
+	// For resources without a K8s CRD (e.g. Variable, Secret, Role, User, Folder, etc.),
+	// fall back to checking the user's permission on the project/namespace resource.
+	if translatedK8sScope == "" {
```
PeterYurkovich, line 383: My only real concern is the handling of secrets. My concern is the leaking of secrets to users who just have access to the namespace. We don't have a clear picture of what the k8s secret -> perses secrets logic should look in conjunction with RBAC. Maybe we could add a naive version which just checks if the user has k8s secret permissions (ie. including it in the getK8sScope, getK8sAPIGroup, getK8sAPIVersion)? This won't work for specific secret access, but should prevent any real escalation

jgbernalp: I mapped the perses secret and globalsecret scopes to a k8s secret. WDYT?

Nexucis:
To replace the default guess permission, you could check if the user has clusterRole. That would makes senses to have a cluster role to get access to global datasource, wdyt ?

Otherwise, from my perspective, mapping a global datasource access to the namespace does not make sense. If you need to define a datasource at namespace level then a project datasource would do the work. Is there something I am missing ?

jgbernalp:
> To replace the default guess permission, you could check if the user has clusterRole. That would makes senses to have a cluster role to get access to global datasource, wdyt ?
> 
> Otherwise, from my perspective, mapping a global datasource access to the namespace does not make sense. If you need to define a datasource at namespace level then a project datasource would do the work. Is there something I am missing ?

GlobalDatasources and GlobalDasboards were not changed, they are still driven by ClusterRoles as they have corresponding CRs. My intention is to map the resources that have no corresponding CR. 

Project scoped resources, such as variables, are now checked via the project access. Secrets are now checked directly against k8s.

internal/api/authorization/k8s/translation.go
```diff
@@ -35,22 +35,41 @@ const (
 	k8sGlobalDatasourceScope k8sScope = "persesglobaldatasources"
 	k8sDatasourceScope       k8sScope = "persesdatasources"
 	k8sProjectScope          k8sScope = "namespaces"
+	k8sSecretScope           k8sScope = "secrets"
 )
 
+// globalScopesToCheck contains all scopes that should be checked at the wildcard (all-namespace)
+// level when computing permissions. This includes both global-only scopes and project-scoped
+// resources that benefit from an initial wildcard check to avoid redundant per-namespace checks.
 var globalScopesToCheck = []v1Role.Scope{
```
Nexucis: To avoid duplication, you could something like that:

```go
globalScopesToCheck = append(projectScopesToCheck, 	v1Role.GlobalDatasourceScope,
	v1Role.GlobalVariableScope,
	v1Role.GlobalSecretScope,
)
```

internal/api/authorization/k8s/k8s.go
PeterYurkovich: looks good, thanks for the change

Issue #3137, opened 2025-07-30
Support Role Assignment Based on Scopes from Access Token


Nischitha1611:
### Proposal

Currently, Perses does not support assigning roles to users based on the scopes present in the access token after logging in using either oAuth/OIDC provider.

Add support in Perses to:
- Parse the scope field from the access token after user login.
- Allow administrators to map specific scopes to Perses roles (e.g., viewer, editor, admin) in the configuration.
- Dynamically assign roles to the user session based on these mappings, without needing explicit rolebindings per user or group.

Reference:
Grafana already supports a similar capability using the role_attribute_path in its configuration. 
https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/#role-mapping-examples

pkr4711:
@ibakshay This is a feature which I also urgently need.

gfierkasap:
I can work on this one

celian-garcia:
From a discussion we had together, some candidates for syntax that can be handy to retrieve claims in token. 
- https://jmespath.org/
- https://cel.dev/

If we want to go further, like if we want to support more complex behaviors like Azure Group Overage 
https://learn.microsoft.com/en-us/troubleshoot/entra/entra-id/app-integration/get-signed-in-users-gro
Then we'd have to find a more complex approach with dedicated server that would manage permission collection

gfierkasap:
I created a draft PR with the proposed changes. As of now it still requires some work (general issues are listed in the PR description), but I would love some feedback as to whether the general approach I took with this feature seems ok.

PR #3871, opened 2026-02-12
[FEATURE] Implement RBAC based on access token scopes (#3137)

gfierkasap:
# Description

This PR intends to add RBAC roles based on scopes found in the access token, as described in #3137 . Specific mapping of token role/group claims to Perses RBAC roles can be done via `security.authorization.claims_mapping_config` part of the config file.

TODO:

- Document implemented changes 
- Implement tests
- Improve error handling
- Improve general quality of the new code

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/pkg/model/api/config/authentication.go
+++ b/pkg/model/api/config/authentication.go
@@ -212,6 +212,7 @@ type AuthenticationConfig struct {
 	// Providers configure the different authentication providers
 	Providers AuthProviders `json:"providers" yaml:"providers"`
 	// PersistClaims defines which oAuth/oidc token claims should be persisted in a separate cookie
+	// Claims are defined via the JMESpath to the specific part of the token payload section
 	PersistClaims []string `json:"persist_claims,omitempty" yaml:"persist_claims,omitempty"`
 }
 
@@ -221,6 +222,9 @@ func (a *AuthenticationConfig) Verify() error {
 	}
 	if a.RefreshTokenTTL == 0 {
 		a.RefreshTokenTTL = common.Duration(DefaultRefreshTokenTTL)
+	}
+	if a.PersistClaims == nil {
+		a.PersistClaims = []string{}
 	}
 	return nil
 }
--- a/pkg/model/api/config/authorization.go
+++ b/pkg/model/api/config/authorization.go
@@ -41,18 +41,21 @@ func (a *Assignment) CheckRoleClaim(userRoleClaims []string) bool {
 	return false
 }
 
+// RoleAssignment contains the mapping of role found in the token to Perses role.
 type RoleAssignment struct {
 	Assignment
 	Project string `json:"project" yaml:"project"`
 	Role    *v1.Role
 }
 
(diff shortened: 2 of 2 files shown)


internal/api/authorization/native/native.go
```diff
@@ -44,16 +44,29 @@ func New(userDAO user.DAO, roleDAO role.DAO, roleBindingDAO rolebinding.DAO,
...
+	n := &native{
 		cache:                &cache{},
 		userDAO:              userDAO,
 		roleDAO:              roleDAO,
 		roleBindingDAO:       roleBindingDAO,
 		globalRoleDAO:        globalRoleDAO,
 		globalRoleBindingDAO: globalRoleBindingDAO,
-		guestPermissions:     conf.Security.Authorization.Provider.Native.GuestPermissions,
+		guestPermissions:     conf.Security.Authorization.GuestPermissions,
+		rolesJMESPath:        conf.Security.Authorization.ClaimsMappingConfig.RoleClaimsPath,
```
celian-garcia: I think that it's not necessary to have role + group claims path. You can have only one claim path and let the maintainer to decide to choose "role"or "group"or whatever other claim.

internal/api/impl/auth/oauth.go
```diff
@@ -447,6 +447,7 @@ func (e *oAuthEndpoint) tokenHandler(ctx echo.Context) error {
 	if err != nil {
 		return err
 	}
+	ctx.Set("access_token", resp.AccessToken)
```
celian-garcia: I'm not fan of saving the whole access_token from this place. Anyway I think it doesn't even do what you think, as here it's not saved for other requests.

Thinking of it, one solution could be to have a config allowing to copy at login time some claims from provider's access token into perses access token.

Then we could have a more generic role binding or project role binding, that would, instead of binding by user name as it does currently, could bind by any claim of Perses access token.

That's only a proposal. What do you think of that?

internal/api/utils/utils.go
```diff
@@ -14,6 +14,12 @@
 package utils
 
 import (
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"strings"
+
+	"github.com/jmespath/go-jmespath"
```
celian-garcia: Would be good to have a little comparison somewhere in the PR with arguments why we'd choose jmespath over cel since this has been listed in the issue.

internal/api/authorization/native/native.go
gfierkasap: Sure, I can remove that. I only added that since it was essentially no effort at all, and I saw it being mentioned in some other issue, but I thought looking air two different claims might be redundant.

internal/api/impl/auth/oauth.go
gfierkasap: This part was my biggest concert, and I wasn't exactly happy about the way I tried implementing it, nor was I sure this would work. 

I'll get back to the drawing board on this one, and looking at your suggestion it is definitely an idea worth exploring.

Issue #3874, opened 2026-02-13
Grafana migration: dashboard fails when a panel has a decimal value for position


AntoineThebaud:
### What did you do?

Grafana dashboards can sometimes have such decimal value in a panel grid position:
```json
"gridPos": {
  "h": 9,
  "w": 12,
  "x": 12,
  "y": 13.5
},
```
This is causing the migration to Perses to fail:
`bad request: json: cannot unmarshal number 13.5 into Go struct field plain.panels.y of type int`

### What did you expect to see?

The migration should still succeed. As Perses only manipulates integers here, we should round such value.

### Perses version

0.53.0-beta.3

### Perses configuration file

```yaml

```

### Logs

```text

```

PR #3883, opened 2026-02-16
[BUGFIX] Grid x,y coordination is a float in Grafana

Nexucis:
fix https://github.com/perses/perses/issues/3874


AntoineThebaud:
thanks for handling it!

PR #3887, opened 2026-02-17
[BUGFIX] Project permission is determinate based on the Perses resources in namespace 

PeterYurkovich:
# Description

Kubernetes requires both a namespace and a resource to check permissions against (ie. "persesdashboards" in "perses" namespace). Checking the permissions for a user directly against a namespace will return a `NoOpinion` verdict. 

This PR re-adds a `checkNamespaceAccess` function which changes a project scope permission request to instead check if the user has access to any k8s resources within the project itself. Ensuring that `GetUserProjects` is able to return all projects which the user has any permissions in.

Since namespaces cannot be directly checked against, this PR also changes the fallback RBAC check for Variables, EphemeralDashboards and Folders to check authorization against Dashboards.

Extra tests and comments have been added to cover these behaviors to ensure no future regressions.

# Checklist

### DIFF
--- a/internal/api/authorization/k8s/k8s.go
+++ b/internal/api/authorization/k8s/k8s.go
@@ -385,7 +385,7 @@ func getUnknownActions(knownActions []v1Role.Action) []v1Role.Action {
 }
 
 func (k *k8sImpl) checkSpecificPermission(ctx echo.Context, namespace string, user user.Info, action v1Role.Action, scope v1Role.Scope) (authorized authorizer.Decision, err error) {
-	if scope == v1Role.ProjectScope || scope == "" {
+	if scope == v1Role.ProjectScope {
 		return k.checkNamespaceAccess(ctx, namespace, user, action)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/api/authorization/k8s/k8s.go
```diff
@@ -384,13 +385,17 @@ func getUnknownActions(knownActions []v1Role.Action) []v1Role.Action {
 }
 
 func (k *k8sImpl) checkSpecificPermission(ctx echo.Context, namespace string, user user.Info, action v1Role.Action, scope v1Role.Scope) (authorized authorizer.Decision, err error) {
+	if scope == v1Role.ProjectScope || scope == "" {
```
Nexucis: I don't think scope can be empty, so it would make the second check unnecessary

Nexucis (approved):
excepting a minor comment, this is good for me @PeterYurkovich. Thanks for the fix !

internal/api/authorization/k8s/k8s.go
PeterYurkovich: Removed

PR #3890, opened 2026-02-17
OIDC logout cognito

tony-clarke-siemens:
A small tweak to control log out when working with Cognito

See https://github.com/perses/perses/issues/3550

### DIFF
--- a/cue/cue.mod/module.cue
+++ b/cue/cue.mod/module.cue
@@ -7,7 +7,7 @@ source: {
 }
 deps: {
 	"github.com/perses/shared/cue@v0": {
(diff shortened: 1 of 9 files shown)


celian-garcia:
The code looks good but I'm on mobile right now so I will check twice with a computer. Meanwhile I ran the pipeline so please check it if it's red. 😊
Thank you!

tony-clarke-siemens:
> The code looks good but I'm on mobile right now so I will check twice with a computer. Meanwhile I ran the pipeline so please check it if it's red. 😊 Thank you!

Needed to remove year from copyright header. Should be good now.

PR #3902, opened 2026-02-23
[FEATURE] Load plugin from multiple folders

### LINKED ISSUE #3881: Allow to install plugins, either at runtime or with provisioning
### Proposal

For custom Perses builds or for Perses users that want to try new plugins outside of the core Perses plugins, for example community ones, custom or beta versions. Perses should allow to install a plugin either from a file or define a folder or files in provisioning.

Idea, use percli for runtime installation:

```bash
percli plugin install path/to/my/plugin.tar
```

or through provisioning:

```yaml
provisioning:
  plugins:
    - folder/with/custom/plugins
    - plugin-file.tar
```

Nexucis:
This PR is adapting the original source where we are looking for the plugin archive.

Instead of having a single source, you can specify a list of folder where Perses will look for the archives to extract.

This will allow user to inject private plugin in the container without being forced to rebuild the docker image.

fix #3881

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -825,7 +825,13 @@ path: <path> | default = ("plugins" | "/etc/perses/plugins") # Optional
 
 # The path to the folder containing the plugins archive. 
 # When Perses is starting, it will extract the content of the archive in the folder specified in the `folder` attribute.
+# DEPRECATED: use `archive_paths` instead to specify multiple folders for the archived plugins.
 archive_path: <path> | default = ("plugins-archive" | "/etc/perses/plugins-archive") # Optional
+
+# The list of paths to the directories containing the archived plugins. It allows to specify multiple directories for the archived plugins.
+# When Perses is starting, it will extract any archive found in the folders specified in this attribute in the folder specified in the `path` attribute.
+archive_paths: 
+    - <path> | default = ("plugins-archive" | "/etc/perses/plugins-archive") # Optional
 
 # Allow use of plugins in dev mode.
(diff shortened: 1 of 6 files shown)


Nexucis:
I believe it will help for the setup of the e2e tests for the plugin in https://github.com/perses/plugins/pull/573.

Instead of loading the plugin using the CLI command `percli plugin start` we can just provide the list of plugins tested and Perses will load them directly if the archive exists.

jgbernalp:
Thank you!. IIUC this will look up for plugins in more locations but this action will still happens when Perses starts right?. Which means the image still needs to have the plugins before Perses starts. This solves the issue of separating the core plugins from other "external" or "custom" plugins. But does Perses already tracks new files added to these folders to unzip? or this will be part of the runtime installation part?, of course as part of another enhancement PR.

Nexucis:
so as you figure out, the load will happen at the start of the image. Perses does not track or make any hot reloading of the plugin. It was the case the case before but it has been removed to simplify the load management of the plugin.

Also because I have assumed if you want to add more plugins, you will necessary change the deployment / the config, so the pod will restart. So I have assuming it won't be a big issue to not load at runtime the plugin.

But perhaps you have a scenario in mind that will change these assumptions :).

jgbernalp:
This work correctly for the k8s scenario, I was thinking more on the scenario you have Perses running and found a cool plugin (maybe in the marketplace) you want to test. Currently, you would need to stop it, adjust the config and restart. But as said before it can be part of a future enhancement.

PR #3921, opened 2026-03-02
[FEATURE] Native RBAC authorization based on oidc/oAuth roles

gfierkasap:
# Description

This PR intends to solve issue #3137 ([previous PR](https://github.com/perses/perses/pull/3871) - that one will be closed, it was easier and faster to redo the PR rather than resolve conflicts)

In order to allow for RBAC authorization based on oidc/oAuth roles, this PR implements a few new functionalities: 

1. It allows to configure Perses to persist oidc/oAuth token claims, by specifying a list of JMESpaths to the desired claims in the token payload section via the `security.authentication.persist_claims` field; the claims are passed down via a separate cookie, by default called `persistedClaims`
2. Implements the `security.authorization. claims_mapping_config` field to map roles taken from the external provider token to Perses RBAC roles
3. When checking permissions it only extends the current way of native authorization works; if token roles are not present or do not allow for requested action, it still checks for any roles bound to the user by "standard" RBAC role binding

EDIT:
Right now, instead of a separate `persist_claims` cookie used to pass down the saved claims, a new server-side cache is implemented; cache entries TTLs are equal to token `expires_in` value

# Checklist

## UI Changes


internal/api/authorization/native/native.go
```diff
@@ -68,10 +83,91 @@ type native struct {
...
+	return nil
+}
+
+func (n *native) buildTokenGlobalRoles(ac config.AuthorizationConfig) error {
+	// get all global roles
+	globalRoles, err := n.globalRoleDAO.List(&globalrole.Query{})
+	if err != nil {
+		return err
+	}
+	for _, gra := range ac.ClaimsMappingConfig.GlobalRoleMapping {
```
celian-garcia: I'm not fan of this kind of shortcut. I'd prefer
- ca -> conf
- gra -> mapping

internal/api/authorization/native/native.go
```diff
@@ -215,12 +311,37 @@ func (n *native) HasPermission(ctx echo.Context, requestAction v1Role.Action, re
...
 	n.mutex.RLock()
 	defer n.mutex.RUnlock()
 	return n.cache.hasPermission(username, requestAction, requestProject, requestScope)
 }
 
+func (n *native) hasPermissionFromClaim(ctx echo.Context, project string, requestAction v1Role.Action, requestScope v1Role.Scope) bool {
+	if project != v1.WildcardProject {
+		roles := n.getUserTokenRoles(ctx)
+		for _, r := range roles {
+			if r.CheckPermission(project, requestAction, requestScope) {
```
celian-garcia: Couldn't you reuse the listHasPermission function here instead of creating some checkpermission methods that seems to do the same thing

celian-garcia (changes requested):
I think it's going into the right direction! 

I'm not sure it deserves a separate cookie though. Reusing the actual access token and adding claims in it would be simpler in my opinion. You'd have to ensure non conflicting claim with a prefix like ``ext_{claim_key} = {claim_value}`` but this is a little price to pay to get the benefit of both sessions dying at the same time. WDYT ? 

Also did you test it live ? Is it working ? I see that there are some checks failing, I let you fix them, and tell me once you have something working, I would be eager to test it with Azure provider.

internal/api/authorization/native/native.go
celian-garcia: I see that there is a difference on the permissions list, one is a list of pointer and one is a list of permission, but I don't like the duplication here.

gfierkasap: Yeah, makes sense - initially I found it more convenient to use CheckPermission methods, but looking at it it was a bit redundant; refactored the `hasPermissionFromClaim` function to use `listHasPermission` in https://github.com/perses/perses/pull/3921/commits/077c914a0d39b625a854a3cb7d2f2f92aa63dcbd

internal/api/authorization/native/native.go
gfierkasap: Changed in https://github.com/perses/perses/pull/3921/commits/0cf438675caee90b4fc3cc3b24c19e674bf52744

gfierkasap:
> I think it's going into the right direction!
> 
> I'm not sure it deserves a separate cookie though. Reusing the actual access token and adding claims in it would be simpler in my opinion. You'd have to ensure non conflicting claim with a prefix like `ext_{claim_key} = {claim_value}` but this is a little price to pay to get the benefit of both sessions dying at the same time. WDYT ?
> 
> Also did you test it live ? Is it working ? I see that there are some checks failing, I let you fix them, and tell me once you have something working, I would be eager to test it with Azure provider.

I tested it at some point during the development, but TBH that was not the final version, and it wasn't a thorough manual test. Reusing the access token might also be a good idea, i'll try and implement it shortly

gfierkasap:
> I think it's going into the right direction!
> 
> I'm not sure it deserves a separate cookie though. Reusing the actual access token and adding claims in it would be simpler in my opinion. You'd have to ensure non conflicting claim with a prefix like `ext_{claim_key} = {claim_value}` but this is a little price to pay to get the benefit of both sessions dying at the same time. WDYT ?
> 
> Also did you test it live ? Is it working ? I see that there are some checks failing, I let you fix them, and tell me once you have something working, I would be eager to test it with Azure provider.

A little update on the idea of reusing existing cookies: I tried implementing it, but it turned out that fully implementing it would be a bit more complicated than I expected. After thinking about it for a while, I changed my approach completely and used a server-side cache to store the extracted claims (it also holds the cookie, compares it to the cookie from the request, and updates cache and claims when necessary), and I think it turned out quite nicely. 

Anyhow, this still needs some testing as automated tests fail, will let you know on my progress.

gfierkasap:
PR should be ready to be manually tested

celian-garcia:
Sure sorry for the late, I should find time today or tomorrow.

internal/api/authorization/native/claimsmanager.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+type JMESPathQuery struct {
+	path  string
+	query *jmespath.JMESPath
+}
+
+// Stores claims in cache; if an entry with the same cookie and username exists, is omitted
+// returns true if cache entry set, otherwise false (not sure if useful?)
+func (cm *ClaimsManager) SetClaims(username string, cookie string, ttl time.Duration, data Claims) bool {
```
celian-garcia, line 145: I am a bit puzzled here, you say it's a cookie but it's actually a token that you save in memory. So it has nothing to do with a cookie. 

Here you try to create a kinda server side session that you store in memory. But **A Server Side Session would have to be saved in database**. Indeed, if Perses is deployed in the cloud with two replicas for example, you cannot ensure that the memory will be shared between two queries of the same user.

To me, to be honest, **introducing a Server Side Session mechanism implies too much complexity for a Perses newcomer**. I'd be you, I would really just save the needed claims in the Perses access token as I advised earlier. I can help you to get rid of some blockers.
In a nutshell, the whole logic of token creation is in ``tokenManagement`` struct. It would be something similar with [what I've done](https://github.com/perses/perses/pull/3587/changes) when I saved a new ``ProviderInfo`` into the token.

celian-garcia (changes requested):

Nexucis:
@gfierkasap would you like to continue this PR ?

gfierkasap:
@Nexucis sorry for the delay in response, I missed your comment. I was held up on a different task so no progress happened for a while, but I can keep working on this issue

Nexucis:
no worries @gfierkasap, it was more to know where we are standing with this PR. Now I have the answer, thank you very much !

Carrvin:
Any updates on this topic? We'd love to replace Grafana with Perses in our Self Service Platform, this is currently the only showstopper for us.

gfierkasap:
> Any updates on this topic? We'd love to replace Grafana with Perses in our Self Service Platform, this is currently the only showstopper for us.

Hi @Carrvin, so sorry for the late reply. After consulting a while back with one of the maintainers, we decided on a different approach than I initially took. This would require so many changes (not to mention the amount of conflicts to resolve since this PR is already quite old) that I decided on discontinuing this PR and creating a new one from scratch (which I done recently). You can find the new PR here https://github.com/perses/perses/pull/4172

To sum up, work in progress, but the new one will be ready for review shortly,

celian-garcia:
Thanks for the update @gfierkasap I close this PR then if favor of the new one #4172

Carrvin:
@gfierkasap Thanks a lot for the update!

PR #3927, opened 2026-03-03
[BUGFIX] read formatters from the cache

shahrokni:
Relates to https://github.com/perses/perses/issues/3448

# Description
Creating the formatter object is an expensive task. **This was pointed out by the browser profiler** that the main thread is spending some considerable amount of time for creating formatters. This can significantly affect huge dashboards.

The suggested fix utilizes a JS map and a short key to quickly pickup the proper formatter. So, the formatter is created once and used whenever it is needed. The following screenshots demonstrate that the max INP drops using the cache.

# Screenshots

## Before change

p1

## After using cache

image

image

# Checklist

## UI Changes

### DIFF
--- a/ui/core/src/model/units/time.ts
+++ b/ui/core/src/model/units/time.ts
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 import { MAX_SIGNIFICANT_DIGITS } from './constants';
+import { getFormatterFromCache } from './formatterCache';
 import { UnitGroupConfig, UnitConfig } from './types';
 import { hasDecimalPlaces, limitDecimalPlaces } from './utils';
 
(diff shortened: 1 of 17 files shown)


Gladorme:
Can we have some numbers? In order to validate there is significant improvement

ui/core/src/model/units/time.ts
```diff
@@ -169,6 +171,19 @@ export function formatTime(value: number, { unit, decimalPlaces }: TimeFormatOpt
     formatterOptions.maximumSignificantDigits = MAX_SIGNIFICANT_DIGITS;
   }
 
-  const formatter = Intl.NumberFormat('en-US', formatterOptions);
+  const key = [
```
jgbernalp, line 173: it seems the same logic is duplicated in all these files as only the key changes. Should we centralize this?

shahrokni: Isn't it an overkill?

jgbernalp: Why? is harder to maintain if something changes.

shahrokni: > Why? is harder to maintain if something changes.

I mean  the logic of formatter creation is not really the same. They are different. For example, compare time and temperature. Or you are talking about the hash map itself, where the pick up happens? That part should be united? **Yes the map could/should be united**

jgbernalp: The pattern is the same, values might be different, for example: the key ,formatter options, and locale. But this could easily be refactored away.

something like:

```typescript
return formatNumber({ formatterOptions, locale })
// ...
return formatNumber({ formatterOptions })
```

which internally handles the caching of the formatters based on the formatterOptions and locale

jgbernalp (approved):

PR #3944, opened 2026-03-11
[IGNORE] Add comments to clarify how schemas are loaded

Nexucis:


internal/api/plugin/schema/schema.go
```diff
@@ -64,10 +64,38 @@ func Load(pluginPath string, moduleSpec plugin.ModuleSpec) ([]LoadSchema, error)
 		}
 		currentDir, _ := filepath.Split(currentPath)
 		logrus.Tracef("Loading model package from %s", currentDir)
+		// Name is the lowest type level of the plugin. For example, for the Prometheus module, it can return PrometheusTimeseriesQuery.
+		// It is called named because in the plugin definition located in package.json, this value is present in the field name.
```
AntoineThebaud:
```suggestion
		// It is called "name" because in the plugin definition located in package.json, this value is present in the `name` field.
```

internal/api/plugin/schema/schema.go
```diff
@@ -64,10 +64,38 @@ func Load(pluginPath string, moduleSpec plugin.ModuleSpec) ([]LoadSchema, error)
...
+		//        "kind": "TimeSeriesQuery",
+		//        "spec": {
+		//          "display": {
+		//            "name": "Prometheus Time Series Query"
+		//          },
+		//          "name": "PrometheusTimeSeriesQuery"
+		//        }
+		//      },
+		//
+		// Then in the schema PrometheusTimeSeriesQuery is defined in the field `kind`
```
AntoineThebaud:
```suggestion
		// Besides in the plugin schema, "PrometheusTimeSeriesQuery" is the value of the `kind` field
```

internal/api/plugin/schema/schema.go
```diff
@@ -285,6 +313,8 @@ func (s *sch) load(pluginPath string, module v1.PluginModule) error {
 	}
 	for _, schema := range schemas {
 		if schema.Kind.IsQuery() {
+			// Here the information about the super type of the query (aka is a TimeSeriesQuery for example) disappear.
```
AntoineThebaud:
```suggestion
			// Here the information about the "super type" of the query (aka TimeSeriesQuery for PrometheusTimeSeriesQuery) disappears.
```

internal/api/plugin/schema/schema.go
```diff
@@ -285,6 +313,8 @@ func (s *sch) load(pluginPath string, module v1.PluginModule) error {
...
+			// This is not good because then you might be able to associate the super type `LogQuery` with the plugin implementation `PrometheusTimeSeriesQuery`
```
AntoineThebaud:
```suggestion
			// This is a known validation gap yet to be solved: because of this you can currently wrongly pass the validation with e.g the super type `LogQuery` & the plugin implementation `PrometheusTimeSeriesQuery`
```

AntoineThebaud (approved):

PR #3966, opened 2026-03-18
[FEATURE] Manage a sub-folder system

### LINKED ISSUE #183: Manage a sub-folder system

adrianSepiol:
# Description
Depends on: https://github.com/perses/shared/pull/90
Closes: https://github.com/perses/perses/issues/183

Implements a folder/file tree view for the dashboard using a shared Table component.

Considerations:
Folders don't support display names from what I can tell so generateMetadataName is called on users input and this is what will be saved and then shown.

Backend changes:
Empty folders can be created. This will simplify ui make for better ux. e.g. user wants to have folder that is grouping other folders and doesn't contain any dashboards. To make this possible to create we would have to create some elaborate form that would allow to create subfolders while creating folder.

Folders need to be updated after dashboard delete.

# Screenshots
Basic look:

[Screencast from 27-04-26 12:08:31.webm](https://github.com/user-attachments/assets/026e2d53-d85b-45c3-84c4-009e7958af1b)

Adding a folder: 

[Screencast from 27-04-26 12:10:56.webm](https://github.com/user-attachments/assets/fd534cf0-b6ad-4b7e-a51b-a22b7293b652)

Deleting and editing: 

[Screencast from 27-04-26 12:18:49.webm](https://github.com/user-attachments/assets/0c898885-e895-41a4-b6e0-772c3be25829)

Fuzzy search:

[Screencast from 27-04-26 12:21:03.webm](https://github.com/user-attachments/assets/07593bd9-7cde-4791-a294-0bab4a1e17bb)

Column filtering:

[Screencast from 27-04-26 12:22:09.webm](https://github.com/user-attachments/assets/84019e62-271d-4552-93d5-24706ccb8acf)

# Checklist

## UI Changes


adrianSepiol:
Pipelines are failing because of https://github.com/perses/shared/pull/90 once this one is released they should pass.

ui/app/src/components/list.tsx
```diff
@@ -15,7 +15,7 @@ import { Box, Chip, Tooltip } from '@mui/material';
 import { GridColDef, GridFilterInputValue, GridFilterItem, GridFilterOperator } from '@mui/x-data-grid';
 import { GridInitialStateCommunity } from '@mui/x-data-grid/models/gridStateCommunity';
 import { DispatchWithPromise } from '@perses-dev/core';
-import { intlFormatDistance } from 'date-fns';
+import { formatAbsoluteTime, formatRelativeTime } from '../utils/dashboardTableUtils';
```
jgbernalp, line 18: It seems this file does not exist. are we missing a commit?

ui/app/src/model/folder-client.ts
```diff
@@ -0,0 +1,145 @@
...
+  return fetchJson<FolderResource>(url, {
+    method: HTTPMethodPUT,
+    headers: HTTPHeader,
+    body: JSON.stringify(entity),
+  });
+}
+
+function deleteFolder(entity: FolderResource): Promise<Response> {
+  const url = buildURL({ resource: resource, project: entity.metadata.project, name: entity.metadata.name });
+  return fetch(url, {
```
jgbernalp, line 141: If we use fetch directly we might need to handle http responses as in the fetchJson case this are handled internally.

ui/app/src/components/DashboardList/DashboardTreeList.tsx
```diff
@@ -0,0 +1,318 @@
...
+  isLoading,
+}: DashboardTreeTableProps): ReactElement {
+  const isMobileSize = useIsMobileSize();
+  const getTableHeight = useCallback(
+    () => (isMobileSize ? 500 : (max([window.innerHeight - 350, 300]) ?? 300)),
+    [isMobileSize]
+  );
+  const [height, setHeight] = useState(getTableHeight());
+  useEffect(() => {
+    const handleResize = (): void => setHeight(getTableHeight);
```
jgbernalp: This might be read as something different, better to the explicit
```suggestion
    const handleResize = (): void => setHeight(() => getTableHeight());
```

ui/app/src/components/DashboardList/DashboardList.tsx
```diff
@@ -57,23 +94,21 @@ export interface DashboardListProperties extends ListProperties {
...
-    },
-    [dashboardList]
-  );
-
-  const rows = useMemo(() => {
-    const historyMap = new Map(navHistory.map((h) => [`${h.project}/${h.name}`, h.date]));
-    return dashboardList.map<Row>((dashboard, index) => {
+  const dashboardsRows = useMemo(() => {
+    return dashboardList.map<DashboardListRow>((dashboard, index) => {
+      const historyMap = new Map(navHistory.map((h) => [`${h.project}/${h.name}`, h.date]));
```
jgbernalp:
```suggestion
      const historyMap = new Map(navHistory.map((h) => [`${h.project}/${h.name}`, h.date]));
      return dashboardList.map<DashboardListRow>((dashboard, index) => {

```

ui/app/src/components/list.tsx
adrianSepiol: Messed up renaming. I've pushed the file now

jgbernalp:
@adrianSepiol it seems some type checks are failing, after the dependency upgrade some functions might have been moved to other packages. Can you check?

pkg/model/api/v1/folder.go
```diff
@@ -60,9 +60,6 @@ func (f *FolderSpec) validate() error {
 	if f.Kind != KindDashboard && f.Kind != KindFolder {
 		return fmt.Errorf("kind can only be %q or %q but not %q", KindDashboard, KindFolder, f.Kind)
 	}
-	if f.Kind == KindFolder && len(f.Spec) == 0 {
```
jgbernalp, line 63: Are we saying that a spec can be empty now? Is this needed in the case we create a folder from scratch and start adding folders or dashboards?

adrianSepiol: It would simplify ux, as let say user wants to create folder that would have only subfolders and those subfolders would have dashboards. If we don't allow empty folders we would have to create form that would allow to create whole structure in one go.

adrianSepiol:
> @adrianSepiol it seems some type checks are failing, after the dependency upgrade some functions might have been moved to other packages. Can you check?

This seems to be related to https://github.com/perses/shared/pull/113/changes (action moved from plugin-system to components), once app imports are adjusted I will merge the latest to this branch.

ui/app/src/components/DashboardList/DashboardTreeList.tsx
```diff
@@ -0,0 +1,318 @@
...
+  handleDuplicateButtonClick,
+  handleRenameButtonClick,
+  handleEditFolderButtonClick,
+  handleAddFolderButtonClick,
+  handleDeleteFolderButtonClick,
+  isLoading,
+}: DashboardTreeTableProps): ReactElement {
+  const isMobileSize = useIsMobileSize();
+  const getTableHeight = useCallback(
+    () => (isMobileSize ? 500 : (max([window.innerHeight - 350, 300]) ?? 300)),
```
jgbernalp: I don't think we need lodash for this, we want to remove this dependency long term:


```suggestion
    () => (isMobileSize ? 500 : (Math.max(window.innerHeight - 350, 300) ?? 300)),
```

pkg/model/api/v1/folder.go
jgbernalp: We might want to add some tests regarding this change.

jgbernalp:
This should fix the issues: 
- https://github.com/perses/perses/pull/4043 

when merged you can rebase.

jgbernalp:
@adrianSepiol It seems the branch still outdated or might need manual changes for the types. The e2e issue is a flaky test.

jgbernalp:
@adrianSepiol I found an issue with the names of the folders. Are we allowed to change the folder name? Also when creating a "root" folder if the name includes spaces they are replaced by `_`, but this does not happen for a nested folder.

jgbernalp:
I think we can move the toggle between list and folder aside the main buttons so It does not break the card flow, WDYT?

Screenshot 2026-05-01 at 16 08 34

jgbernalp:
I searched with a random string and it produced results but I'm not sure if the search is wrong or it matched something in the folders as the matched folders do not have any dashboards:

Screenshot 2026-05-01 at 16 12 53

adrianSepiol:
> I searched with a random string and it produced results but I'm not sure if the search is wrong or it matched something in the folders as the matched folders do not have any dashboards:
> Screenshot 2026-05-01 at 16 12 53

@jgbernalp @tanstack/match-sorter-utils is used for search with default threshold. In this case this matches 't' from "test_folders" and "as" from "a_main_folder_with_spaces" even thou those are not in sequence. Possible threshold settings for reference: 
export const rankings = {
  CASE_SENSITIVE_EQUAL: 7,
  EQUAL: 6,
  STARTS_WITH: 5,
  WORD_STARTS_WITH: 4,
  CONTAINS: 3,
  ACRONYM: 2,
  MATCHES: 1,
  NO_MATCH: 0,
} as const

Deafult is "MATCHES".

adrianSepiol:
> @adrianSepiol I found an issue with the names of the folders. Are we allowed to change the folder name? Also when creating a "root" folder if the name includes spaces they are replaced by `_`, but this does not happen for a nested folder.

@jgbernalp So the issue is backend doesn't accept display name for folder. Comparing it to dashboard there we have have spec.display.name here spec is array of FolderSpec and that is either folder or dashboard that is in that folder. So the name here is also a key for that object (that's why it is normalized). Nested folders exist only as an entry in array so renaming them is possible. To be able to rename root folder I think we would need to change/extend backend model.

adrianSepiol:
> I think we can move the toggle between list and folder aside the main buttons so It does not break the card flow, WDYT?
> Screenshot 2026-05-01 at 16 08 34

On that topic we had meeting within our team and I demo'ed the solution and got feedback that maybe this new view should be the default one or even the only one. What do you think?

There were also other points:
- folders should be always on top; I pushed commit with that, let me know if You are fine with that change.
- When searching folders should auto expand to see what was matched; This would be a change in shared repository. I can introduce it if you find it useful as well.

jgbernalp:
> folders should be always on top; I pushed commit with that, let me know if You are fine with that change.
When searching folders should auto expand to see what was matched; This would be a change in shared repository. I can introduce it if you find it useful as well.

If the search highlights the results I think this could be the default option. A "collapse all" and "expand all" might be helpful as well

adrianSepiol:
> > I searched with a random string and it produced results but I'm not sure if the search is wrong or it matched something in the folders as the matched folders do not have any dashboards:
> > Screenshot 2026-05-01 at 16 12 53
> 
> @jgbernalp @tanstack/match-sorter-utils is used for search with default threshold. In this case this matches 't' from "test_folders" and "as" from "a_main_folder_with_spaces" even thou those are not in sequence. Possible threshold settings for reference: export const rankings = { CASE_SENSITIVE_EQUAL: 7, EQUAL: 6, STARTS_WITH: 5, WORD_STARTS_WITH: 4, CONTAINS: 3, ACRONYM: 2, MATCHES: 1, NO_MATCH: 0, } as const
> 
> Deafult is "MATCHES".

@jgbernalp I created pr that makes this configurable and changes default to 'CONTAINS' https://github.com/perses/shared/pull/131

adrianSepiol:
> > folders should be always on top; I pushed commit with that, let me know if You are fine with that change.
> > When searching folders should auto expand to see what was matched; This would be a change in shared repository. I can introduce it if you find it useful as well.
> 
> If the search highlights the results I think this could be the default option. A "collapse all" and "expand all" might be helpful as well

@jgbernalp PR with this change https://github.com/perses/shared/pull/132

adrianSepiol:
@jgbernalp 2 follow ups we talked about:
https://github.com/perses/perses/issues/4059
https://github.com/perses/perses/issues/4060

jgbernalp (approved):

PR #3972, opened 2026-03-19
[IGNORE] replace safely Core with Spec

shahrokni:
**⚠️ MUST BE APPROVED BY** @Nexucis 
Would be great if we could have @jgbernalp opinion as well. 

The eventual goal is to drop the core. The specification types have already been moved to the Spec repo meaning we need to remove the duplicate from the Core. To keep this transition safe and avoid any breaking changes, the core plays the role of a proxy by re-exposing the spec types. (for some time) So, the consumers of the core  are still safe. Meanwhile, all duplicate types, constants and functions are removed. 

# Checklist

## UI Changes


Gladorme:
IMO, we need a big PR for replacing everything in the package core. If we only do one PR by one, it will never be finished. And for the aliasing, I never did a migration like this, but I guess it's better to use `export type DashboardSpec = DashboardSpec`.

shahrokni:
> IMO, we need a big PR for replacing everything in the package core. If we only do one PR by one, it will never be finished. And for the aliasing, I never did a migration like this, but I guess it's better to use `export type DashboardSpec = DashboardSpec`.

The main problem with using type instead of interface is that types are final and cannot be "re-opened" for declaration merging. If any code in your shared package tries to add extra properties to DashboardSpec or uses a class that implements it (extend or implement), a type alias would cause a compiler error if the spec ever becomes a union (type union check the attached example in the next comment). By using interface extends, you provide a flexible wrapper that acts like the original object while safely bridging the two packages during your migration.

Gladorme:
> cannot be "re-opened" for declaration merging

Could you please provide an example, because I don't understand what you mean

shahrokni:
> > cannot be "re-opened" for declaration merging
> 
> Could you please provide an example, because I don't understand what you mean

image

shahrokni:
> > > cannot be "re-opened" for declaration merging
> > 
> > 
> > Could you please provide an example, because I don't understand what you mean
> 
> image



> IMO, we need a big PR for replacing everything in the package core. If we only do one PR by one, it will never be finished. And for the aliasing, I never did a migration like this, but I guess it's better to use `export type DashboardSpec = DashboardSpec`.

@Gladorme 
About the big PR,
I completely understand your concern, 
I am personally fan of gradual changes, because, it gives you more room to think about the changes and self-correction at the early stages. 
Also, with huge replacements, sometimes (not saying always) it becomes overwhelming.  Actually, I have a very good example. The PR @Nexucis has opened had a problem with unit tests which had nothing to do with moving stuff around. Still it looked confusing. 
Probably the one who is maintaining the big PR, need to spend more enrgy to keep it aligned. 
Again, these are two different views for achieving the same goal. I am OK and respect the community decision

Gladorme:
The issue only happen if the original spec change from an interface to a type with unions.

The issue would have happen even without spec package and extends will change nothing 🤔 

image

shahrokni:
> The issue only happen if the original spec change from an interface to a type with unions.
> 
> The issue would have happen even without spec package and extends will change nothing 🤔
> 
> image

Could you elaborate more :D
I am not sure if I get you this time or not :D
So, what I understand is if they change it to a type, then we get the error. Correct? In any case

Gladorme:
For step 2, if you use `type ...` or `interface ... extends ...`based on a union type. For step 3, it will never work. So it's changing nothing

shahrokni:
@Gladorme 
Maybe I can move everything we have had so far so it is not really one by one? What do you think?
Maybe we have different definitions of the big PR. By Big PR, do you mean after this Big PR, Core will be dropped completely?

Gladorme:
Yes we want a big PR, like @Nexucis did for shared

Nexucis (approved):
awesome ! Thank you for your work @shahrokni.

I think we might still update the spec since Nicolas' PR (#3998) has been merged and it changed slightly the way `project.spec` is managed

PR #3987, opened 2026-03-26
[ENHANCEMENT] `percli apply`: add new flag `--create-project`

AntoineThebaud:
# Description

This change introduces a new `--create-project` flag to `percli apply`.

When enabled, if applying a project-scoped resource fails because the target project does not exist, the CLI will:
- create the missing project,
- retry the original apply once,
- and keep normal error behavior for other failure types.

Without `--create-project`, behavior is unchanged.

Why this change: the main goal is to improve the UX when dealing with a large DaC setup involving multiple projects.
 Today, applying dashboards requires the target project(s) to already exist, which is cumbersome in several common cases:
- bootstrapping a fresh environment from exported resources,
- applying dashboards for a newly introduced project in an existing setup,
- moving an existing dashboard from one project to a new one.

With `--create-project` the CLI can create the missing project automatically and retry the apply, enabling a simpler single-step workflow.

# Checklist


celian-garcia (approved):
LGTM. A little doubt if people use several apply in parallel, but I guess it would recreate the project even if existing.

AntoineThebaud:
> LGTM. A little doubt if people use several apply in parallel, but I guess it would recreate the project even if existing.

Hmm good point, currently in such case (project gets created by another parallel request) it would fail here:

```go
if _, projectCreationErr := projectService.CreateResource(projectEntity); projectCreationErr != nil && !errors.Is(projectCreationErr, perseshttp.ConflictError) {
	return projectCreationErr
}
```

It looks like I could change to call `Upsert` instead of `CreateResource` to avoid this problem, @Nexucis could you confirm?

celian-garcia:
One who'd like to ensure that could also ensure that calling an apply of the project at the beginning of the pipeline no ?

AntoineThebaud:
> One who'd like to ensure that could also ensure that calling an apply of the project at the beginning of the pipeline no ?

Yes but that's more tedious (you have to ensure the project resources you defined are & remain aligned with the different values for metadata.project found among your dashboards), and as we foresee this situation as a common one it makes sense imho to add such flag to make it easier/smoother.

internal/cli/cmd/apply/apply.go
```diff
@@ -124,6 +130,52 @@ func (o *option) applyEntity() error {
...
+		Metadata: modelV1.Metadata{
+			Name: project,
+		},
+	}
+	if _, projectCreationErr := projectService.CreateResource(projectEntity); projectCreationErr != nil && !errors.Is(projectCreationErr, perseshttp.ConflictError) {
+		return projectCreationErr
+	}
+
+	// Retry once the original request against the newly created project.
+	if retryErr := o.upsertEntity(kind, project, entity); retryErr != nil {
```
Nexucis:
```suggestion
return o.upsertEntity(kind, project, entity)
```

it is simpler since you don't do anything particular with the error returned

pkg/model/api/v1/project_error.go
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"regexp"
+)
+
+const (
```
Nexucis: I would put that in `internal/interface/error.go`

Nexucis (approved):
I have added two little comments to clean a bit the code. Nothing too serious.

> > LGTM. A little doubt if people use several apply in parallel, but I guess it would recreate the project even if existing.
> 
> Hmm good point, currently in such case (project gets created by another parallel request) it would fail here:
> 
> ```go
> if _, projectCreationErr := projectService.CreateResource(projectEntity); projectCreationErr != nil && !errors.Is(projectCreationErr, perseshttp.ConflictError) {
> 	return projectCreationErr
> }
> ```
> 
> It looks like I could change to call `Upsert` instead of `CreateResource` to avoid this problem, @Nexucis could you confirm?

I think the way it is done currently is correct. Using the `upsert` way would mean to update the project if it exists. But in the given context, updating the project does not make sense because we want to ensure the project exists. So if the project already exists, we can move forward and try to create the entity again.

AntoineThebaud:
> I think the way it is done currently is correct. Using the upsert way would mean to update the project if it exists. But in the given context, updating the project does not make sense because we want to ensure the project exists. So if the project already exists, we can move forward and try to create the entity again.

Yeah actually since we are ensuring that the error returned is not about "document already exist" (`!errors.Is(projectCreationErr, perseshttp.ConflictError)`) then we are already fine as-is, nevermind...

Nexucis (approved):

PR #3988, opened 2026-03-26
[BUGFIX] Set specific Content-Type for static UI resources

### LINKED ISSUE #3979: Set specific Content-Type for static resources
### Proposal

See also

amaury-d:
# Description

As per
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options `Content-Type` is required to avoid potential browser blocking

+ `X-Content-Type-Options: nosniff` to avoid MIME confusion attack

# Checklist

### DIFF
--- a/ui/endpoint.go
+++ b/ui/endpoint.go
@@ -19,8 +19,10 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"mime"
 	"net/http"
 	"net/http/httputil"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -103,6 +105,11 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 		// The First thing to do is to replace the URL path with the local path of the plugin.
 		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
 		// Then we just need to rely on the echo router to serve the file.
+
+		// X-Content-Type-Options: nosniff is an HTTP response header that tells browsers: "Do not try to guess the content type — trust the Content-Type header I sent you."
+		// Without it, browsers perform "MIME sniffing". For example, a file served as text/plain could be sniffed as text/html and executed as HTML, which could open the door to XSS attacks.
+		// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
+		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
 		return c.File(localPath)
 	}
 	// Otherwise, it means we are in a dev environment, and we need to proxy the request to the dev server.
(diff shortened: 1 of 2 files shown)


ui/endpoint.go
```diff
@@ -103,6 +105,7 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 		// The First thing to do is to replace the URL path with the local path of the plugin.
 		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
 		// Then we just need to rely on the echo router to serve the file.
+		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
 		return c.File(localPath)
```
Nexucis, line 113: I would assume this function is setting correctly the content type, but I am not entirely sure. Do you know if it is actually the case ?

ui/endpoint.go
```diff
@@ -103,6 +105,7 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 		// The First thing to do is to replace the URL path with the local path of the plugin.
 		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
 		// Then we just need to rely on the echo router to serve the file.
+		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
```
Nexucis, line 112: can you explain why we need this content type options ? I don't know about this one

Nexucis: ok I just saw your explanation in the PR description. Can you add that in a comment somewhere in the code please ? That would help to understand why we need that. And if necessary a pointer to a doc explaining in more details why this is necessary.

ui/endpoint.go
amaury-d: 👍  I just double checked : `echo.Context.File` uses `http.ServeFile`, which does set `Content-Type` based on the file extension.

ui/endpoint.go
amaury-d: Comment added

ui/endpoint_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}{
+		{
+			name:                "CSS file",
+			path:                "/app/dist/main.abc123.css",
+			expectedContentType: "text/css; charset=utf-8",
+		},
+		{
+			name:                "JavaScript file",
+			path:                "/app/dist/main.abc123.js",
+			expectedContentType: "text/javascript; charset=utf-8",
```
Nexucis: somehow the mime type in the windows machine is the old one for Javascript `application/javascript`. This is an obsolete mime type since 2022 replaced by `text/javascript`.

I didn't get how we can update the mime types in windows (even if it is possible) and why windows is keeping this old mime type or if it is an issue with the GitHub action machine not working.

But this is why this particular unit test is failing on windows

Nexucis: one idea would be to create a dedicated test for windows. Another one would be to get the mime types regardless the OS and use it as an expected Content type. So basically not hardcoding the string value expected

Nexucis (approved):
awesome, thank you @amaury-d for taking care about this issue !

Nexucis:
it is really weird the e2e tests are timeout on this particular PR. Even if they are flacky they are still running on other PR. any clue @jgbernalp ?

jgbernalp:
@Nexucis I'll take a look

PR #4003, opened 2026-04-03
Set specific Content-Type for static resources

tmchow:
The asset handler in `ui/endpoint.go` serves static files (CSS, JS, SVG, etc.) without setting a `Content-Type` header, forcing the browser to sniff the correct MIME type. This adds `mime.TypeByExtension` to detect the type from the file extension and sets `X-Content-Type-Options: nosniff` to prevent MIME confusion attacks.

Changed `assetHandler()` in `ui/endpoint.go` to call `mime.TypeByExtension(filepath.Ext(fileName))` before writing the response body, and set the `nosniff` header on every static response.

This contribution was developed with AI assistance (Claude Code).

Fixes #3979

PR #4004, opened 2026-04-05
Set specific Content-Type for static resources

### LINKED ISSUE #3979: Set specific Content-Type for static resources

tmchow:
Static assets were served with a generic `Content-Type`, forcing browsers to MIME-sniff the correct type. This adds `mime.TypeByExtension()` to set the correct `Content-Type` based on file extension (e.g. `text/css` for `.css`, `text/javascript` for `.js`), and sets `X-Content-Type-Options: nosniff` to prevent MIME confusion attacks.

Changes in `ui/endpoint.go`:
- `assetHandler()`: determine Content-Type from file extension via `mime.TypeByExtension()`
- `serveASTFiles()`: explicitly set `text/html; charset=utf-8` for `index.html`
- Both handlers now include `X-Content-Type-Options: nosniff`

Fixes #3979

Signed-off-by: Trevin Chow <trevin@tmchow.com>

This contribution was developed with AI assistance (Claude Code).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/ui/endpoint.go
+++ b/ui/endpoint.go
@@ -19,8 +19,10 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"mime"
 	"net/http"
 	"net/http/httputil"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


ui/endpoint.go
```diff
@@ -174,6 +176,10 @@ func (f *frontend) assetHandler() echo.HandlerFunc {
 		if strings.Contains(fileName, ".js") {
 			data = bytes.ReplaceAll(data, []byte(prefixPathPlaceholder), []byte(f.apiPrefix))
 		}
+		if ct := mime.TypeByExtension(filepath.Ext(fileName)); ct != "" {
+			c.Response().Header().Set("Content-Type", ct)
+		}
+		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
```
jgbernalp, line 182: should we keep the sniff in cases the typeByExtension returns empty?

Nexucis:
like said in https://github.com/perses/perses/pull/4003, there is already a PR opened (https://github.com/perses/perses/pull/3988) containing these changes and that is more advanced than what you are proposing.
So again, I am closing this PR.

PR #4012, opened 2026-04-09
fix(api): set specific Content-Type for static resources

### LINKED ISSUE #3979: Set specific Content-Type for static resources

mvanhorn:
## Summary

Static resources served by the frontend now get their correct Content-Type based on file extension (e.g. `text/css` for `.css`, `application/javascript` for `.js`). Also adds `X-Content-Type-Options: nosniff` to prevent MIME confusion attacks.

## Why this matters

The `assetHandler()` in `ui/endpoint.go` writes response bytes directly via `c.Response().Write(data)` without setting Content-Type. Browsers had to sniff the MIME type. Setting explicit types + nosniff follows OWASP security best practices.

## Changes

- `ui/endpoint.go`: Added `mime.TypeByExtension` lookup and `X-Content-Type-Options: nosniff` header in `assetHandler()`.

## Testing

- `go build ./ui/...` passes (note: `embedFS` undefined is pre-existing, only resolved when UI assets are built)

Fixes #3979

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/ui/endpoint.go
+++ b/ui/endpoint.go
@@ -19,8 +19,10 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"mime"
 	"net/http"
 	"net/http/httputil"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


Nexucis:
Hello,
A PR much more advanced than this one is already open.
I am closing this one in favor of the other one.

PR #4014, opened 2026-04-11
[BUGFIX] Sort query migrations to prevent greedy Prometheus matching

### LINKED ISSUE #4013: PrometheusTimeSeriesQuery: migration greedily matches non-Prometheus targets
## Problem

The [Prometheus query migration script](https://github.com/perses/plugins/blob/main/prometheus/schemas/prometheus-time-series-query/migrate/migrate.cue) matches any Grafana target containing an `expr` field, regardless of the target's datasource type. This is noted in the script itself:

> Best-effort conversion logic that may wrongly convert not-prometheus queries to PrometheusTimeSeriesQuery

This becomes a concrete problem when adding query migration scripts for other datasource plugins that also use `expr` (e.g. Loki). Since `percli` iterates query migration scripts via a Go map, iteration order is non-deterministic. If both a Prometheus and a Loki script match the same target, the result is a coin flip: sometimes `PrometheusTimeSeriesQuery`, sometimes `LokiLogQuery`.

## Investigation

We looked at two possible approaches to make the dispatch reliable:

### 1. Check `datasource.type` when available

Adding `type: "prometheus"` to the existing datasource constraint in the Prometheus script:

```cue
datasource?: {
    type: "prometheus"
    uid?: string
    ...
}
```

This works when `datasource.type` is present in the Grafana export. CUE rejects `type: "loki"` against the `type: "prometheus"` constraint, so Loki targets no longer match.

However, when `datasource.type` is absent (older dashboards, hand-crafted JSON), both scripts still match because `datasource?:` is optional and CUE fills in the constraint value on unification.

### 2. Analyze `expr` content to distinguish PromQL from LogQL

LogQL queries that use pipeline stages (`|=`, `|~`, `| json`, `| logfmt`, etc.) can be identified by a pipe operator after the stream selector closing brace:

```cue
expr: =~"^\\{[^}]*\\}\\s*\\|"
```

This correctly identifies LogQL with pipelines and rejects PromQL (including `{__name__="up"}`). However, bare LogQL stream selectors without pipes (e.g. `{container="nginx"}`) are syntactically identical to PromQL label matchers (e.g. `{__name__="up"}`), so they remain ambiguous.

### Conclusion

No single heuristic fully solves this. A layered approach covers the vast majority of real-world cases:

1. `datasource.type` present: Definitive match (handles most Grafana exports)
2. Pipe operator after `}`: Definitely LogQL
3. Starts with metric name or function: Definitely PromQL
4. Bare `{labels}` selector with no type info: Ambiguous between PromQL and LogQL; Prometheus fallback is acceptable since this is the current behavior

A combination of approach 1 (check `datasource.type` when available) and approach 2 (check for LogQL pipe syntax) would resolve the non-determinism for all practical cases.

Ref: perses/plugins#622

utafrali:
## Description

PrometheusTimeSeriesQuery was matching non-Prometheus targets because map iteration in Go is random. Sorted the query keys before iterating so migrations run in deterministic order.

Added test case with ExoticTSDB to verify it doesn't get mistaken for Prometheus.

Fixes #4013

## Screenshots

N/A

## Checklist

- [x] PR title follows `[<catalog_entry>] <commit message>` convention
- [x] All commits have DCO signoffs
- [ ] Screenshots/screencasts for UI changes
- [x] E2E tests are stable


utafrali:
doesn't look like this got any traction, closing for now

AntoineThebaud:
Hello @utafrali, well it can unfortunately happen that community PRs stay idle for a while but we always end up looking at them, we maintainers have a hard time keeping up with the pace of community contributions lately, which is good news regarding the project's traction but can be of course frustrating for the contributors. So as a general rule, please be patient & don't close your PR.

However for this precise case I think you jumped a bit too fast on the topic, a fix was already being discussed (& been [implemented](https://github.com/perses/plugins/pull/632) since) and it was about reworking the migration logic for the concerned plugin, not about introducing sorting which isn't a satisfactory solution here.

Anyway thanks for your interest in the project & don't hesitate to contribute on other topics if you wish.

PR #4037, opened 2026-04-24
[DOC] explain role and scope of each Perses package

shahrokni:
DOC to explain the role Perses packages

### DIFF
--- a/docs/packages/README.md
+++ b/docs/packages/README.md
@@ -1,0 +1,78 @@
+# Perses Packages
+
+This section outlines the purpose of each Perses package. It provides guidance on which packages to install and how developers can contribute to the repository. 
+
+
+## Spec `@perses-dev/spec`
+
+**Source:** [spec](https://github.com/perses/spec)
+
+Spec package contains the Perses dashboard and datasource formats. Its primary role is to define the structure of Perses resources so they can be understood by different tools and languages.
+Using this package ensures your dashboards are compatible with the core Perses platform.
+
+For more information about Open Specification, see the [Open Specification](https://perses.dev/perses/docs/concepts/open-specification/)
+
+> As Perses has a plugin oriented architecture, these specifications are partial as they only cover the core features of
+> the dashboard and datasource formats. Plugin authors are expected to extend these specifications with their own custom fields as needed.
+
+
+
+## Components `@perses-dev/components`
+
+**Source:** [components](https://github.com/perses/shared)
+
+This package is the foundational UI library for the Perses ecosystem. It contains the shared React components used to build dashboards and plugins.
+The package provides a unified look and feel for all visual elements and should be installed as a dependency for developers who want to integrate specific Perses visual plugins into their own applications.
+So, in a nutshell, the package offers a toolbox of pre-built components so plugin authors can focus on data logic rather than UI implementation.
+If you are developing a Perses panel or explorer plugin, it is very likely that you require the dependency. 
+
+For more information about plugins, see the [Plugin Development Guide](https://perses.dev/perses/docs/plugins/creation/).
+
+### Key Dependencies
+
+The components package relies on the following core technologies:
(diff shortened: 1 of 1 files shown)


docs/packages/README.md
```diff
@@ -0,0 +1,78 @@
...
+It offers the high-level components needed to display an entire dashboard including 
+the grid layout and panel management. Besides, it exposes the dashboard context through providers, which are explained below.
+
+### Dashboard Provider
+
+Its role is to manage the lifecycle, structural integrity, and interactive editing state of a Dashboard resource.
+It orchestrates all CRUD operations for panels and panel groups.
+This includes handling complex UI workflows like duplicating panels, editing raw JSON, and managing confirmation dialogs for saving or discarding changes.
+
+### Datasource Store Provider
```
Nexucis, line 54: this is where something should change from my opinion. The dashboard package should not be aware about this datasource hierarchy or how to decide which datasource to use because it is related on how we are managing the different resources (more specifically the datasource) in Perses.

Another project could do it differently (for example without Global or Project Datasource, but just global datasource).

This logic should be part of the App Perses, not part of the components we are providing and specially the one loading the dashboard.

docs/packages/README.md
```diff
@@ -0,0 +1,78 @@
...
+The VariableProvider acts as the central state manager for all dashboard variables.
+It tracks which variables are available (both local to the dashboard and global to the system) and manages the background loading states for dynamic dropdown options. 
+While it typically manages dashboard-level variables, it can be reused at the Project or Global level to allow admins to edit higher-scope settings using the same logic.
+
+## Plugin System `@perses-dev/plugin-system`
+
+**Source:** [plugin-system](https://github.com/perses/shared)
+
+The `@perses-dev/plugin-system` acts as the bridge between the dashboard and its plugins.
+It provides the necessary execution environment, data-fetching logic (the actual network call happens in plugins), and shared UI components that allow plugins to function within the Perses ecosystem.
```
Nexucis, line 74: here also when talking about the `data-fetching`. I would tend to believe the way to fetch the data should be implemented by the one installing the plugin-system.

The plugin-system could provide an API to be implemented by the developer. So the logic about how to fetch remain to the app in our case.
What do you think ? @shahrokni @jgbernalp ?

docs/packages/README.md
jgbernalp: But this is the default provider right, other consumers that want to implement this in a different way should implement a different provider. But by default this one will be used.

docs/packages/README.md
jgbernalp: Currently data fetching is handled internally by plugins, I don't think this should change. The plugin system main goal is to manage the plugin lifecycle inside the frontend, and offer UI components used by plugins not data fetching. 

On a different note, what the plugin system could provide is a fetch function, so it can be overwritten if clients need to add custom parameters, different retry logic or routing, but this would be jus a wrapper of the current `fetchJson` function that is used by plugins.

docs/packages/README.md
Nexucis: yes definitely, if other people can implement another provider and inject it, then this is fine for me as long as it is well explained in the documentation

docs/packages/README.md
Nexucis: based on the meeting we had yesterday, we reached a consensus to create another package called `client` being an HTTP client containing all resources supported by Perses. This package will contain a default fetch implementation

Nexucis:
I believe this PR needs to be updated to reflect the last changes in the various repositories impacted.
@shahrokni can you update your PR accordingly ? Thank you by advance.

Nexucis:
I think we can close this PR as https://github.com/perses/perses/pull/4119 has been merged and cover most of the content.

If something is missing, I think created a new PR makes sense.

Thank you for initiating this documentation @shahrokni.

PR #4048, opened 2026-05-04
[ignore] bump go to v1.26

Nexucis:


Nexucis:
it seems upgrading to go v1.26 is breaking the authorisation somehow, will need to check what's going on ...

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/perses/perses
 
-go 1.25.7
+go 1.26.0
```
celian-garcia: I think it's at 1.26.2 with security fixes

internal/api/crypto/jwt.go
```diff
@@ -135,7 +135,7 @@ func (j *jwtImpl) DeleteAccessTokenCookie() (*http.Cookie, *http.Cookie) {
 }
 
 func (j *jwtImpl) CreateRefreshTokenCookie(refreshToken string) *http.Cookie {
-	return &http.Cookie{
+	return &http.Cookie{ //nolint:gosec
```
celian-garcia, line 138: Why our cookie is not secure ?

internal/api/impl/proxy/proxy.go
```diff
@@ -277,28 +289,19 @@ func (h *httpProxy) serve(c echo.Context) error {
...
 		return apiinterface.InternalError
 	}
 
 	// redirect the request to the datasource
 	req.URL.Path = h.path
-	logrus.WithFields(map[string]interface{}{
-		"datasource": h.datasourceName,
-		"path":       h.path,
-	}).Debug("request will be redirected to datasource")
+	h.logWithDefaultEntry().Debug("request will be redirected to datasource")
```
celian-garcia: You can add the path with ``h.logWithDefaultEntry().WithField("path", h.Path).Debug`` no ?

celian-garcia (approved):

celian-garcia:
> it seems upgrading to go v1.26 is breaking the authorisation somehow, will need to check what's going on ...

Haha right after I click on approve I read your message. Ok then let's check why that is strange.

celian-garcia:
I found the difference. It is in the cookie presence in the tests
main branch, go 1.25
image

nexucis/bump2, go 1.26
image

celian-garcia:
I mobilized claude on this and he found the commit 😄 if you're curious https://github.com/golang/go/commit/7aa9ca729fcb063506ea4ed35ee9b2673a853e4f

internal/api/impl/proxy/proxy.go
Nexucis: oh yeah I removed `path` by mistake, will add it back

internal/api/crypto/jwt.go
Nexucis: honestly I didn't really get it. I guess that's because we can pass some insecure parameter, while it should not happen. But we allow it because some users did not understand why Perses was not working and that was because Perses was not served through HTTPS and the only exception was when it was served on localhost.

```
G124: http.Cookie missing or has insecure Secure, HttpOnly, or SameSite attribute (gosec)
```

Issue #2901, opened 2025-05-06
Replace CFB usage by AEAD for encryption and decryption


Nexucis:
### Proposal

When moving to go 1.24, usage of CFB for encryption/decryption has been deprecated. We will have to migrate the way to encrypt the data

AdityaPimpalkar:
Hey @Nexucis, this seems pretty straightforward, mind if I take this one? Thanks!

Nexucis:
no I don't mind.

Just keep in mind changing the way to encrypt might break actually the encryption. So for actual database that means you need to decrypt with the old way to then encrypt with the new one

celian-garcia:
Hello, I take back the topic. We discussed with @Nexucis about a plan that I will execute. Sorry @AdityaPimpalkar for the confusion. We thought it would be simpler but we want to guarantee backward compatibility.

Here is the plan (``BAD`` is non authenticated encryption and ``GOOD`` is authenticated encryption)

-------------------------------

#### Current situation ``crypto.go``
```
# s0
encrypt in BAD
decrypt in BAD
```

#### Step 1 : 
```
# s1
encrypt in BAD
decrypt in GOOD or BAD
```
⚠️ We need to do GCM (new GOOD format) first because CFB (old BAD format) always succeed
 
> wait for monthes (several releases)

#### Step 2:
```
# s2
encrypt in GOOD
decrypt in GOOD or BAD (+reencrypt script on start)
 ```

> wait for monthes (several releases)

#### Step 3:
```
# s3
encrypt in GOOD
decrypt in GOOD
```

#### Conclusion
```
s0 -> s1 -> s0 :   ok
s1 -> s2 -> s1 :   ok
s2 -> s3 -> s2 :   ok
```
⚠️ ``s0`` and ``s2+`` are backward incompatible
⚠️ ``s1`` and ``s3`` are backward incompatible
**=> We count on time to lower the risk.**

PR #4056, opened 2026-05-08
[ENHANCEMENT] Allow decrypt secret in DB with "authenticated" mode AEAD for future backward compatibility

celian-garcia:
# Description

Objective of this PR is to migrate the Perses encryption from CFB (deprecated, unauthenticated) to GCM (AEAD, authenticated) while maintaining full backward compatibility with old data.

Ref: https://go.dev/doc/go1.24#cryptocipherpkgcryptocipher

- [x] Encrypt method can encrypt in both GCM and CFB formats ~
- [x] Decrypt method use first GCM then CFB as fallback for backward compatibility
- [x] Decrypt method returns a boolean to tells user that the data has been encrypted with old model, and needs to be encrypted again
- [ ] ~Everywhere we use Decrypt after fetching a DB object, we need to reencrypt and save with new encryption.~
-> Not yet for backward compatibility. Will be done differently with a script on start.

relates #2901 

# Screenshots

# Checklist

## UI Changes

N/A

### DIFF
--- a/internal/api/crypto/crypto.go
+++ b/internal/api/crypto/crypto.go
@@ -25,13 +25,31 @@ import (
 
 	"github.com/perses/perses/pkg/model/api/config"
 	modelV1 "github.com/perses/perses/pkg/model/api/v1"
+	"github.com/sirupsen/logrus"
 )
 
-// based on https://www.golinuxcloud.com/golang-encrypt-decrypt/
+// TODO(celian-garcia): follow the plan described in the following issue to migrate old encryption to new encryption,
+//  maximizing backward compatibility: https://github.com/perses/perses/issues/2901
+//  ==========================================================
+// 	s0 (based on https://www.golinuxcloud.com/golang-encrypt-decrypt/)
+// 	- encrypt in BAD
+// 	- decrypt in BAD
+// 	*s1 (*current situation)
+// 	- encrypt in BAD
+// 	- decrypt in GOOD or BAD
+// 	s2
+// 	- encrypt in GOOD
+//  - decrypt in GOOD or BAD (+reencrypt script on start)
+// 	s3
+// 	- encrypt in GOOD
+//  - decrypt in GOOD
+//  ==========================================================
 
 type Crypto interface {
 	Encrypt(spec *modelV1.SecretSpec) error
-	Decrypt(spec *modelV1.SecretSpec) error
+	// Decrypt decrypts the spec fields in place.
+	// Returns true if the data was encrypted with the old format and needs re-encryption.
+	Decrypt(spec *modelV1.SecretSpec) (bool, error)
 }
 
 func New(security config.Security) (Crypto, JWT, error) {
@@ -45,8 +63,9 @@ func New(security config.Security) (Crypto, JWT, error) {
 		return nil, nil, err
 	}
 	return &crypto{
-			key:   key,
-			block: aesBlock,
+			usingAuthenticatedEncryption: false,
+			key:                          key,
+			block:                        aesBlock,
 		},
 		&jwtImpl{
 			accessKey:       key,
@@ -57,9 +76,17 @@ func New(security config.Security) (Crypto, JWT, error) {
 		}, nil
 }
 
+// crypto is an implementation of the encrypt / decrypt interface that allows the user to decide which algorithm to be
+// used for encryption. For decryption, it will use both for compatibility.
 type crypto struct {
-	key   []byte
-	block cipher.Block
+	// usingAuthenticatedEncryption allow you to choose between:
+	// - authenticated algorithm (AEAD): we use here GCM (Galois/Counter Mode)
+	// - unauthenticated algorithm: we use here CFB (Cipher Feedback)
+	// === /!\ Beware to not play too much with that to avoid backward incompatibility. /!\ ===
+	// === /!\ This is mainly used internally in a migration process of several months. /!\ ===
+	usingAuthenticatedEncryption bool
+	key                          []byte
+	block                        cipher.Block
 }
 
 func (c *crypto) Encrypt(spec *modelV1.SecretSpec) error {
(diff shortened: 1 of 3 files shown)


internal/api/impl/proxy/globaldatasource.go
```diff
@@ -98,3 +100,16 @@ func (e *endpoint) getGlobalSecret(dtsName, name string) (*v1.SecretSpec, error)
 
 	return &scrt.Spec, nil
 }
+
+func (e *endpoint) updateGlobalSecret(name string, spec *v1.SecretSpec) error {
+	scrt, err := e.globalSecret.Get(name)
+	if err != nil {
+		// This method is used in context of reencryption of the fly. In that context, this error will only be warned in the log.
```
celian-garcia: I'm not satisfied about that, I think I will replicate the verbose logs and error wrap of the other method, so if this error is returned in http one day, we have no surprise

celian-garcia: @Nexucis I made a proposal to avoid having different wording in the errors logged and generated in those functions. See ``error.go`` file and tell me WDYT. 

The interest is to avoid duplications and force future developer to consider all error messages when making changes.

Nexucis:
mmm I just take a look at the PR and while I didn't check in details what's happening in the proxy, I was wondering if we could just run a single process when Perses is starting to migrate all object that require an encryption.

Like that we should not need to modify the proxy and we ensure that every objects have been migrated correctly. (user included)

WDYT ?

internal/api/crypto/crypto.go
```diff
@@ -106,80 +106,107 @@ func (c *crypto) Encrypt(spec *modelV1.SecretSpec) error {
...
+	if err != nil {
+		return "", false, err
+	}
+
+	if len(cipherText) >= gcm.NonceSize() {
+		nonce := cipherText[:gcm.NonceSize()]
+		plainText, err := gcm.Open(nil, nonce, cipherText[gcm.NonceSize():], nil)
+		if err == nil {
+			return string(plainText), false, nil
+		}
```
Nexucis, line 238: here you could add a debug log, that would help I think to see if we are falling back to the old system

celian-garcia:
> mmm I just take a look at the PR and while I didn't check in details what's happening in the proxy, I was wondering if we could just run a single process when Perses is starting to migrate all object that require an encryption.
> 
> Like that we should not need to modify the proxy and we ensure that every objects have been migrated correctly. (user included)
> 
> WDYT ?

Let me summarize our interesting private discussion. You brought the idea to break down things in several steps to limit the risk of backward incompatibility 
```
v0 (now): 
encrypt in BAD
decrypt in BAD
 
v1 : 
encrypt in BAD
decrypt in BAD or GOOD
 
v2:
encrypt in GOOD
decrypt in GOOD or BAD (+reencrypt script on start)
 
v3:
encrypt in GOOD
decrypt in GOOD
 
v0 -> v1 -> v0 :   ok
v1 -> v2 -> v1 :   ok
v2 -> v3 -> v2 :  ok
```

As we said there would still be a backward incompatibility between v0 (now) and v2 but the longest the v1 stays up (monthes/years?) the lower is that risk.

celian-garcia:
I'm closing this PR in favor of the plan re-described in the issue https://github.com/perses/perses/issues/2901

internal/api/impl/proxy/globaldatasource.go
celian-garcia: This has been reverted. No reencryption mechanism for now.

celian-garcia:
PR updated with only the step 1 of the plan. 

No auto migration for now.
The encrypt can encrypt in both format (only old activated now)
The decrypt tries new, then old if new format fails. Return a ``needReencrypt`` boolean for the future. Unused for now. 
⚠️ We need to do GCM first because CFB always succeed

Now let's continue in something like 6months/1 year. The issue is still open and the plan is described in the code.

internal/api/crypto/crypto.go
```diff
@@ -25,13 +25,31 @@ import (
...
+// 	- encrypt in GOOD
+//  - decrypt in GOOD
+//  ==========================================================
 
 type Crypto interface {
 	Encrypt(spec *modelV1.SecretSpec) error
-	Decrypt(spec *modelV1.SecretSpec) error
+	// Decrypt decrypts the spec fields in place.
+	// Returns true if the data was encrypted with the old format and needs re-encryption.
+	Decrypt(spec *modelV1.SecretSpec) (bool, error)
```
Nexucis, line 52: with having in mind the script to re-encrypt all data at once before the server is ready, probably we don't need to change the signature of the function don't you think ?

internal/api/crypto/crypto.go
```diff
@@ -106,88 +126,155 @@ func (c *crypto) Encrypt(spec *modelV1.SecretSpec) error {
...
+		spec.TLSConfig.Key = decrypted
+		needsReEncryption = needsReEncryption || legacy
 	}
-	return nil
+
+	return needsReEncryption, nil
 }
 
-func (c *crypto) encrypt(stringToEncrypt string) (string, error) {
-	if len(stringToEncrypt) == 0 {
```
Nexucis, line 155: this check disappeared, worth to put it back from my opinion

internal/api/crypto/crypto.go
```diff
@@ -106,88 +126,155 @@ func (c *crypto) Encrypt(spec *modelV1.SecretSpec) error {
...
+
+func (c *crypto) encryptCFB(stringToEncrypt string) (string, error) {
 	plainText := []byte(stringToEncrypt)
 	cipherText := make([]byte, aes.BlockSize+len(plainText))
 	iv := cipherText[:aes.BlockSize]
 	if _, err := io.ReadFull(rand.Reader, iv); err != nil {
 		return "", err
 	}
 
-	// TODO use AEAD instead of CFB as recommended by Go
```
Nexucis, line 165: Did you use AEAD as recommended by Go at the end? Sorry for the silly question, I am not that much familiar with this part and from the function name it does not look like you are using AEAD.

internal/api/crypto/crypto.go
```diff
@@ -58,8 +77,9 @@ func New(security config.Security) (Crypto, JWT, error) {
 }
 
 type crypto struct {
-	key   []byte
-	block cipher.Block
+	usingAuthenticatedEncryption bool
```
Nexucis, line 87: I am not sure I understand this boolean. From the usage it looks like it is used as a switch between the old encryption and the new one.

Perhaps a more meaningful boolean would be better ? Like `useNewEncryption` ? WDYT ?

internal/api/crypto/crypto.go
celian-garcia: GCM is the advised AEAD (Authenticated Encryption with Associated Data) algorithm yes. GCM is the algorithm. AEAD is the category of algorithm.

internal/api/crypto/crypto.go
celian-garcia: I think the job will have to avoid doing some updates if not necessary, if it's executed on restart. So that Boolean will still be useful.

internal/api/crypto/crypto.go
Nexucis: great, thank you for your response !

internal/api/crypto/crypto.go
Nexucis: ah yeah that makes more sense with your comment about AEAD. A comment somewhere to explain would be helpful

internal/api/crypto/crypto.go
Nexucis: alright, it's maybe safer to have this check so if the script fails it can be restarted and continue where it stopped.

Nexucis (approved):
once the changes are done, it will be fine to merge this PR.
Thank you for doing it @celian-garcia, it's been a while this was needed !

internal/api/crypto/crypto.go
celian-garcia: it's still in the encrypt() function actually

Nexucis: oupsi, indeed !

Nexucis (approved):
awesome, thank you @celian-garcia !

PR #4064, opened 2026-05-12
[ENHANCEMENT] allow to load versioned plugins

jgbernalp:
# Description

Allows loading of versioned plugins

# Checklist

### DIFF
--- a/internal/api/plugin/schema/cue.go
+++ b/internal/api/plugin/schema/cue.go
@@ -87,12 +87,16 @@ func LoadSchemaInstance(schemaPath string, pkg string) (*build.Instance, error) 
 }
 
 func validatePlugin(plugin common.Plugin, schema *build.Instance, pluginType string, pluginName string) error {
+	version := "unknown"
+	if plugin.Metadata != nil && plugin.Metadata.Version != "" {
+		version = plugin.Metadata.Version
+	}
 	if schema == nil {
-		return fmt.Errorf("schema not found for plugin %s", plugin.Kind)
+		return fmt.Errorf("schema not found for plugin %s with version %s", plugin.Kind, version)
 	}
 	pluginData, err := plugin.JSONMarshal()
 	if err != nil {
-		logrus.WithError(err).Debugf("unable to marshal the plugin %q", plugin.Kind)
+		logrus.WithError(err).Debugf("unable to marshal the plugin %q with version %s", plugin.Kind, version)
 		return err
 	}
 	ctx := cuecontext.New()
--- a/ui/endpoint.go
+++ b/ui/endpoint.go
@@ -16,14 +16,13 @@ package ui
 import (
 	"bytes"
 	"errors"
-	"fmt"
 	"io"
 	"io/fs"
 	"mime"
 	"net/http"
 	"net/http/httputil"
+	"path"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/labstack/echo/v4"
@@ -53,7 +52,6 @@ var (
 		"/profile",
 		"/delegated-auth-error",
 	}
-	capturingPluginName = regexp.MustCompile(`/plugins/([a-zA-Z0-9_-]+)/?.*`)
 )
 
 type frontend struct {
@@ -85,16 +83,20 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
 func (f *frontend) servePluginFiles(c echo.Context) error {
 	// We are going to serve a plugin from a dev environment, let's set up the proxy to redirect the traffic.
 	req := c.Request()
-	res := c.Response()
-	pluginName := catchPluginName(req.URL.Path)
-	if len(pluginName) == 0 {
+
+	// Strip the apiPrefix so parsePluginPath sees a path starting with "/plugins/".
+	urlPath := req.URL.Path
+	if f.apiPrefix != "" {
+		urlPath = strings.TrimPrefix(urlPath, f.apiPrefix)
+	}
+
+	pluginName, pluginVersion, pluginRegistry, relPath, ok := parsePluginPath(urlPath)
+	if !ok {
 		logrus.Errorf("unable to find the plugin name in the URL path: %s", req.URL.Path)
 		return apiinterface.NotFoundError
 	}
-	// TODO: These hardcoded values should be replaced once the frontend is able to manage multiple registries and plugin versions.
-	// This suppose to update the plugin system first.
-	// These hardocded values are matching the default behavior of the plugin system and it helps to keep backward compatibility.
-	loaded, isLoaded := f.pluginService.GetLoadedPlugin(pluginName, pluginModel.LatestVersion, pluginModel.DefaultRegistry)
+	loaded, isLoaded := f.pluginService.GetLoadedPlugin(pluginName, pluginVersion, pluginRegistry)
+
 	if !isLoaded || !loaded.Module.Status.IsLoaded {
 		logrus.Errorf("unable to find the plugin %s", pluginName)
 		return apiinterface.NotFoundError
@@ -102,18 +104,24 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 	devEnvironment := loaded.DevEnvironment
 	if devEnvironment == nil {
 		// In that case, we need to read the requested files from the file system.
-		// The First thing to do is to replace the URL path with the local path of the plugin.
-		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
-		// Then we just need to rely on the echo router to serve the file.
+		// The first thing to do is to replace the URL path with the local path of the plugin.
+		localPath := filepath.Join(loaded.LocalPath, filepath.FromSlash(relPath))
 
-		// X-Content-Type-Options: nosniff is an HTTP response header that tells browsers: "Do not try to guess the content type — trust the Content-Type header I sent you."
-		// Without it, browsers perform "MIME sniffing". For example, a file served as text/plain could be sniffed as text/html and executed as HTML, which could open the door to XSS attacks.
-		// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
+		// Ensure the resolved path is contained within the plugin's directory
+		// to prevent path traversal attacks.
+		// loaded.LocalPath is cleaned separately because it comes from plugin config, not the URL.
+		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
+		if localPath != cleanedPluginDir && !strings.HasPrefix(localPath, cleanedPluginDir+string(filepath.Separator)) {
+			logrus.Errorf("path traversal attempt detected: resolved path %s escapes plugin directory %s", localPath, loaded.LocalPath)
+			return apiinterface.BadRequestError
+		}
+
 		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
 		return c.File(localPath)
 	}
 	// Otherwise, it means we are in a dev environment, and we need to proxy the request to the dev server.
 	// When developing a plugin, you will be able to serve the files of the plugin using a dev server (with rsbuild).
+	res := c.Response()
 	var proxyErr error
 	reverseProxy := httputil.NewSingleHostReverseProxy(devEnvironment.URL.URL)
 	reverseProxy.ErrorHandler = func(_ http.ResponseWriter, _ *http.Request, err error) {
@@ -250,10 +258,44 @@ func (f *frontend) serveASTFiles(c echo.Context) error {
 	return apiinterface.HandleError(err)
 }
 
-func catchPluginName(path string) string {
-	c := capturingPluginName.FindStringSubmatch(path)
-	if len(c) != 2 {
-		return ""
+// parsePluginPath parses a URL path of the form:
+//
+//	/plugins/<name>[~<version>[~<registry>]][/<relPath>]
+//
+// where <name> is the plugin name, <version> defaults to "latest", <registry> defaults to "perses.dev",
+// and <relPath> is the remaining file path after the identity segment.
+// The "~" character is used as a separator for embedding version and registry in a single URL segment.
+func parsePluginPath(urlPath string) (name, version, registry, relPath string, ok bool) {
+	urlPath = path.Clean(urlPath)
+	if !strings.HasPrefix(urlPath, "/plugins/") {
+		return "", "", "", "", false
 	}
-	return c[1]
+
+	trimmed := strings.TrimPrefix(urlPath, "/plugins/")
+	parts := strings.Split(trimmed, "/")
+	if len(parts) == 0 || parts[0] == "" {
+		return "", "", "", "", false
+	}
+
+	identitySegment := parts[0]
+
+	if len(parts) > 1 {
+		relPath = path.Join(parts[1:]...)
+	}
+
+	meta := strings.Split(identitySegment, "~")
+
+	name = meta[0]
+
+	version = pluginModel.LatestVersion
+	if len(meta) > 1 && meta[1] != "" {
+		version = meta[1]
+	}
+
+	registry = pluginModel.DefaultRegistry
+	if len(meta) > 2 && meta[2] != "" {
+		registry = meta[2]
+	}
+
+	return name, version, registry, relPath, true
 }
--- a/ui/endpoint.go
+++ b/ui/endpoint.go
@@ -110,6 +110,15 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 		// The first thing to do is to replace the URL path with the local path of the plugin.
 		localPath := filepath.Join(loaded.LocalPath, filepath.FromSlash(relPath))
 
+		// Ensure the resolved path is contained within the plugin's directory
+		// to prevent path traversal attacks.
+		// loaded.LocalPath is cleaned separately because it comes from plugin config, not the URL.
+		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
+		if localPath != cleanedPluginDir && !strings.HasPrefix(localPath, cleanedPluginDir+string(filepath.Separator)) {
+			logrus.Errorf("path traversal attempt detected: resolved path %s escapes plugin directory %s", localPath, loaded.LocalPath)
+			return apiinterface.BadRequestError
+		}
+
 		// X-Content-Type-Options: nosniff is an HTTP response header that tells browsers: "Do not try to guess the content type — trust the Content-Type header I sent you."
 		// Without it, browsers perform "MIME sniffing". For example, a file served as text/plain could be sniffed as text/html and executed as HTML, which could open the door to XSS attacks.
 		// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
--- a/ui/endpoint.go
+++ b/ui/endpoint.go
@@ -23,6 +23,7 @@ import (
 	"net/http/httputil"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/labstack/echo/v4"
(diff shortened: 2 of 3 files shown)


ui/endpoint.go
```diff
@@ -250,10 +258,33 @@ func (f *frontend) serveASTFiles(c echo.Context) error {
...
 
-func catchPluginName(path string) string {
-	c := capturingPluginName.FindStringSubmatch(path)
-	if len(c) != 2 {
-		return ""
+func parsePluginPath(urlPath string) (name, version, registry string, ok bool) {
+	urlPath = path.Clean(urlPath)
+	if !strings.HasPrefix(urlPath, "/plugins/") {
+		return "", "", "", false
+	}
```
shahrokni, line 281: I think the order of the highlighted blocks could introduce a (potential) security risk.
Probably they should be re-ordered (Not sure though). Let me make an example and see if it makes sense. 

## Hostile Scenario 💀 

An attacker can use directory traversal (`..`) to trick `path.Clean` and bypass the prefix check.

**Malicious Input**

Let's say we send a malicious URL which is => 
 
```javaScript
const urlPath = /plugins/real-plugin/../../plugins/fake-plugin"
```

The way the path.Clean works turns the malicious  input into => 

```javaScript
const result = "/plugins/fake-plugin"
```

Here the `strings.HasPrefix` is tricked and evaluates to true, opening the door to the rest of the process

```go
strings.HasPrefix("/plugins/fake-plugin", "/plugins/")
```

I am not sure yet, but I think this issue could be exploited by two types of attacks (in the subsequent steps). We can go into the details, if you also think it may cause issues

- Log Poisoning
- Denial of Service

jgbernalp: true, this should be sanitized

shahrokni: But if we are not in hurry we can do it in a follow up PR, WDYT?

jgbernalp: is done now

ui/endpoint.go
```diff
@@ -83,37 +81,43 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
-	// TODO: These hardcoded values should be replaced once the frontend is able to manage multiple registries and plugin versions.
-	// This suppose to update the plugin system first.
-	// These hardocded values are matching the default behavior of the plugin system and it helps to keep backward compatibility.
-	loaded, isLoaded := f.pluginService.GetLoadedPlugin(pluginName, pluginModel.LatestVersion, pluginModel.DefaultRegistry)
+	loaded, isLoaded := f.pluginService.GetLoadedPlugin(pluginName, pluginVersion, pluginRegistry)
+
 	if !isLoaded || !loaded.Module.Status.IsLoaded {
 		logrus.Errorf("unable to find the plugin %s", pluginName)
 		return apiinterface.NotFoundError
 	}
```
shahrokni, line 106: Not related to this change, but it seems this could be vulnerable against log forgery. 
Should we create a follow up PR? 

What if I craft a request like this
> plugins/fake-plugin%0AINFO%20%5B2026-05-13%5D%20User%20admin%20authenticated%20successfully

**_Log Poisoning?_**

shahrokni (dismissed):

shahrokni:
**_Please wait for the test ..._**

shahrokni (commented):
The backend works just fine. However, when I used `percli plugin start` to load the plugin, I encountered the following error. The URL looks OK. 
If I stop using percli it is OK again. 




Image

Image

Image

ui/endpoint.go
jgbernalp: Fixed

ui/endpoint.go
```diff
@@ -83,41 +81,47 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
+
+		// Ensure the resolved path is contained within the plugin's directory
+		// to prevent path traversal attacks.
+		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
+		if localPath != cleanedPluginDir && !strings.HasPrefix(localPath, cleanedPluginDir+string(filepath.Separator)) {
+			logrus.WithFields(logrus.Fields{"resolved_path": localPath, "plugin_dir": loaded.LocalPath}).Error("path traversal attempt detected: resolved path escapes plugin directory")
+			return apiinterface.NotFoundError
+		}
 
-		// X-Content-Type-Options: nosniff is an HTTP response header that tells browsers: "Do not try to guess the content type — trust the Content-Type header I sent you."
```
Nexucis, line 109: can you put back this comment please ?

ui/endpoint.go
```diff
@@ -83,41 +81,47 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
-		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
-		// Then we just need to rely on the echo router to serve the file.
+		localPath := filepath.Join(loaded.LocalPath, filepath.FromSlash(relPath))
+
+		// Ensure the resolved path is contained within the plugin's directory
+		// to prevent path traversal attacks.
+		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
+		if localPath != cleanedPluginDir && !strings.HasPrefix(localPath, cleanedPluginDir+string(filepath.Separator)) {
+			logrus.WithFields(logrus.Fields{"resolved_path": localPath, "plugin_dir": loaded.LocalPath}).Error("path traversal attempt detected: resolved path escapes plugin directory")
+			return apiinterface.NotFoundError
```
Nexucis: a 404 error does not sound write for this kind of error. Perhaps more an error 400 since it is the user input that is wrong

internal/api/plugin/schema/cue.go
```diff
@@ -87,12 +87,16 @@ func LoadSchemaInstance(schemaPath string, pkg string) (*build.Instance, error)
 }
 
 func validatePlugin(plugin common.Plugin, schema *build.Instance, pluginType string, pluginName string) error {
+	version := "unknown"
```
Nexucis, line 90: `unknown` does not seem right `latest` is the default one when the version is not found

Nexucis: nether mind this is for logging purpose, not to get the schema itself. My bad I read it too fast

ui/endpoint.go
```diff
@@ -83,41 +81,47 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
+
 	if !isLoaded || !loaded.Module.Status.IsLoaded {
-		logrus.Errorf("unable to find the plugin %s", pluginName)
+		logrus.WithField("plugin", pluginName).Error("unable to find the plugin")
 		return apiinterface.NotFoundError
 	}
 	devEnvironment := loaded.DevEnvironment
 	if devEnvironment == nil {
-		// In that case, we need to read the requested files from the file system.
-		// The First thing to do is to replace the URL path with the local path of the plugin.
```
Nexucis, line 105: I would keep these comments as well. It will help when we will forgot about this code and will need to fix it in one year

ui/endpoint.go
```diff
@@ -83,41 +81,47 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
-		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
-		// Then we just need to rely on the echo router to serve the file.
+		localPath := filepath.Join(loaded.LocalPath, filepath.FromSlash(relPath))
+
+		// Ensure the resolved path is contained within the plugin's directory
+		// to prevent path traversal attacks.
+		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
```
Nexucis: I am wondering if this line is really necessary as you already clean the URL right before

ui/endpoint.go
```diff
@@ -250,10 +254,37 @@ func (f *frontend) serveASTFiles(c echo.Context) error {
...
 	}
-	return c[1]
+
+	identitySegment := parts[0]
+
+	if len(parts) > 1 {
+		relPath = path.Join(parts[1:]...)
+	}
+
+	meta := strings.Split(identitySegment, "~")
```
Nexucis, line 295: I will be great to add a comment to explain what is the expected URL. I had to find the PR you did in `shared` to understand why you are doing that

ui/endpoint_test.go
```diff
@@ -106,3 +113,277 @@ func TestServeASTFilesContentType(t *testing.T) {
...
+			input:            "/plugins/plugin~1.2.3~my-registry/manifest.json",
+			expectedName:     "plugin",
+			expectedVersion:  "1.2.3",
+			expectedRegistry: "my-registry",
+			expectedRelPath:  "manifest.json",
+			expectedOk:       true,
+		},
+		{
+			name:             "Plugin with version and registry",
+			input:            "/plugins/plugin~~my-registry/manifest.json",
```
Nexucis, line 183: can you add a test with the registry `perses.dev` to see if it works.

Nexucis:
> The backend works just fine. However, when I used `percli plugin start` to load the plugin, I encountered the following error. The URL looks OK. If I stop using percli it is OK again.
> 
> Image Image Image

Yeah this is normal with the current code. The devserver does not server the file with the URL received by the application. When configuring the proxy to forward the request to the devserver, the path of the URL needs to be adapted to remove the `plugin~version` thing.

shahrokni:
> > The backend works just fine. However, when I used `percli plugin start` to load the plugin, I encountered the following error. The URL looks OK. If I stop using percli it is OK again.
> > Image Image Image
> 
> Yeah this is normal with the current code. The devserver does not server the file with the URL received by the application. When configuring the proxy to forward the request to the devserver, the path of the URL needs to be adapted to remove the `plugin~version` thing.

I had the same idea, 
Are you talking about this part?
```go
res := c.Response()
	var proxyErr error
	reverseProxy := httputil.NewSingleHostReverseProxy(devEnvironment.URL.URL)
	reverseProxy.ErrorHandler = func(_ http.ResponseWriter, _ *http.Request, err error) {
		logrus.WithError(err).Errorf("error proxying, remote unreachable: target=%s, err=%v", devEnvironment.URL.String(), err)
		proxyErr = err
	}
	if transportErr := proxyPrepareRequest(c, devEnvironment); transportErr != nil {
		return transportErr
	}
	// Reverse proxy request.
	reverseProxy.ServeHTTP(res, req)
	if proxyErr != nil {
		// we need to wrap the error with an Echo Error,
		// otherwise the error will be hidden by the middleware "middleware.HandleError".
		status := res.Status
		if status < 400 {
			// if there is an error and the status code doesn't match the error, then let's use a default one
			status = 500
		}
		return echo.NewHTTPError(status, proxyErr.Error())
	}
```
I debugged this part, but couldn't understand where exactly it encounters the issue. 
Could you please share the exact line or function which should be adjusted?

jgbernalp:
> > > The backend works just fine. However, when I used `percli plugin start` to load the plugin, I encountered the following error. The URL looks OK. If I stop using percli it is OK again.
> > > Image Image Image
> > 
> > 
> > Yeah this is normal with the current code. The devserver does not server the file with the URL received by the application. When configuring the proxy to forward the request to the devserver, the path of the URL needs to be adapted to remove the `plugin~version` thing.
> 
> I had the same idea, Are you talking about this part?
> 
> ```go
> res := c.Response()
> 	var proxyErr error
> 	reverseProxy := httputil.NewSingleHostReverseProxy(devEnvironment.URL.URL)
> 	reverseProxy.ErrorHandler = func(_ http.ResponseWriter, _ *http.Request, err error) {
> 		logrus.WithError(err).Errorf("error proxying, remote unreachable: target=%s, err=%v", devEnvironment.URL.String(), err)
> 		proxyErr = err
> 	}
> 	if transportErr := proxyPrepareRequest(c, devEnvironment); transportErr != nil {
> 		return transportErr
> 	}
> 	// Reverse proxy request.
> 	reverseProxy.ServeHTTP(res, req)
> 	// Return any error handled during proxying request.
> 	if proxyErr != nil {
> 		// we need to wrap the error with an Echo Error,
> 		// otherwise the error will be hidden by the middleware "middleware.HandleError".
> 		status := res.Status
> 		if status < 400 {
> 			// if there is an error and the status code doesn't match the error, then let's use a default one
> 			status = 500
> 		}
> 		return echo.NewHTTPError(status, proxyErr.Error())
> 	}
> ```
> 
> I debugged this part, but couldn't understand where exactly it encounters the issue. Could you please share the exact line or function which should be adjusted?

IIUC we should remove the right substring from the dev environment. As now the version and/or registry are added.

Nexucis:
> IIUC we should remove the right substring from the dev environment. As now the version and/or registry are added.

yes you understand correctly. `rsbuild` does not add the version and the registry in the URL

shahrokni:
Refers to:
https://github.com/perses/perses/pull/4064#issuecomment-4444973049

@jgbernalp and @Nexucis 
After overriding the `Rewrite` from `httputil.ReverseProxy` The linked issue was resolved. I tested the change with the new frontend. With or without `devEnvironment` it works just fine. I am not sure what I did was a workaround or not. For me it makes sense as the previous state of the code almost does the same think with `Director`

I didn't use `NewSingleHostReverseProxy` because there `Director` is already overridden and deprecated. 
The IDE suggestion was to use the  `Rewrite`, so I took it. 
 

My Commit: 
> 7b8877c

shahrokni (approved):

ui/endpoint.go
```diff
@@ -85,37 +86,58 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
-		// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
 		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
 		return c.File(localPath)
 	}
 	// Otherwise, it means we are in a dev environment, and we need to proxy the request to the dev server.
 	// When developing a plugin, you will be able to serve the files of the plugin using a dev server (with rsbuild).
+	res := c.Response()
 	var proxyErr error
-	reverseProxy := httputil.NewSingleHostReverseProxy(devEnvironment.URL.URL)
+	// TODO: I simply override the Rewrite(Director is deprecated) temporarily.
```
jgbernalp: Not sure what Director means. Wouldn't be enough to replace the substring in the `devEnvironment.URL.URL`? Instead of this workaround?

shahrokni: Based on my logs `devEnvironment.URL.URL` is is the backend address of the local development server which is `http://localhost:3008`

This should happen inside `Rewrite` function because we are modifying the incoming request path from the browser, not the destination address of the environment.

## Details of my idea

the Director is the internal control mechanism that tells the reverse proxy how to modify and forward the incoming request to the backend server

When a browser sends a request to the app, the reverse proxy intercepts it. At that moment, the request still has the browser's metadata. The `Director` function is the  step where the proxy intercepts that package and rewrites its routing details. Now it seems the Director has been deprecated. So, I did it with the Rewrite which is the suggested alternative

Nexucis: actually you just need to rewrite the path of the initial request. You don't need to do all of that.

``` go
// redirect the request to the dev server
req.URL.Path = newPath
reverseProxy := httputil.NewSingleHostReverseProxy(devEnvironment.URL.URL)
reverseProxy.ServeHTTP(res, req)
```

where new path, is the correct one without version and registry

Nexucis: If you want more info, you can check on how we are redirecting the request to the datasource in the code of the proxy. This is slightly the same thing we are doing here.

ui/endpoint.go
Nexucis: still accurate I believe :)

Nexucis (approved):
lgtm now. I still have a comment unresolved, it would be nice to have the comment back before mergin

ui/endpoint.go
```diff
@@ -85,37 +86,51 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
-		// Without it, browsers perform "MIME sniffing". For example, a file served as text/plain could be sniffed as text/html and executed as HTML, which could open the door to XSS attacks.
-		// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
 		c.Response().Header().Set("X-Content-Type-Options", "nosniff")
 		return c.File(localPath)
 	}
 	// Otherwise, it means we are in a dev environment, and we need to proxy the request to the dev server.
 	// When developing a plugin, you will be able to serve the files of the plugin using a dev server (with rsbuild).
+	res := c.Response()
 	var proxyErr error
+	if pluginPathRegex.MatchString(req.URL.Path) {
```
Nexucis: I would tend to believe this check is not necessary. You can just replace the path directly, if it does not match, it won't do anything

ui/endpoint.go
```diff
@@ -85,27 +86,38 @@ func (f *frontend) RegisterRoute(e *echo.Echo) {
...
-		// The First thing to do is to replace the URL path with the local path of the plugin.
-		localPath := strings.Replace(req.URL.Path, fmt.Sprintf("%s/plugins/%s", f.apiPrefix, pluginName), loaded.LocalPath, 1)
-		// Then we just need to rely on the echo router to serve the file.
+		// The first thing to do is to replace the URL path with the local path of the plugin.
+		localPath := filepath.Join(loaded.LocalPath, filepath.FromSlash(relPath))
 
+		// Ensure the resolved path is contained within the plugin's directory
+		// to prevent path traversal attacks.
+		// loaded.LocalPath is cleaned separately because it comes from plugin config, not the URL.
+		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
```
jgbernalp: `filepath.Join` uses `Clean` internally. This does not prevent path traversal. Do we need to protect agains path traversal when using dev mode? We probably don't need it as this reads files from a different folder location outside of Perses.

Nexucis: I don't think this is necessary as it will be unlikely used with a production instance. That would change the plugin loaded for everyone and I don't think you want that

ui/endpoint.go
```diff
@@ -107,23 +110,20 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 		// The first thing to do is to replace the URL path with the local path of the plugin.
 		localPath := filepath.Join(loaded.LocalPath, filepath.FromSlash(relPath))
 
-		// Ensure the resolved path is contained within the plugin's directory
-		// to prevent path traversal attacks.
-		// loaded.LocalPath is cleaned separately because it comes from plugin config, not the URL.
-		cleanedPluginDir := filepath.Clean(loaded.LocalPath)
```
Nexucis: since this is for the production environment, I think we should keep it.

Gabriel in the comment was thinking to have it also when it was for the dev environment. I was against it since for the dev environment you are not supposed to be in production so checks are unnecessary from my opinion.

ui/endpoint.go
```diff
@@ -114,8 +118,12 @@ func (f *frontend) servePluginFiles(c echo.Context) error {
 	}
 	// Otherwise, it means we are in a dev environment, and we need to proxy the request to the dev server.
 	// When developing a plugin, you will be able to serve the files of the plugin using a dev server (with rsbuild).
+	res := c.Response()
 	var proxyErr error
+	req.URL.Path = pluginPathRegex.ReplaceAllString(req.URL.Path, "")
```
Nexucis, line 132: does it remove only the version and the registry of the plugin here (if it exists). Or does it do something else ?

Am I not sure I understand what exactly it is removing.

Based on how the dev server is working both URL are valid:

- http://localhost:3009/plugins/Prometheus/mf-manifest.json
- http://localhost:3009/mf-manifest.json

shahrokni: So, simply speaking with this regex, everything from `/plugins/` up to the subsequent slash will be replaced with string empty. In this case `plugins/x~y~z/mf-manifest.json` the `/plugins/x~y~z` will be removed. And that is what we want because in dev mode it does not make sense to have them. 

Let's take a closer look at the regex

> Find the word /plugins/ at the start of the text, and grab absolutely everything after it until you hit the very next slash

image

## Debug
As you can see name, version, registry and urlPath have been received. 
At this point the `URL.path` is `"/plugins/PieChart~0.13.1/mf-manifest.json"`
But, this is invalid in `dev` mode.
What matters is the `mf-manifest.json`

image

As it has been explained already the pattern is found and replaced. 
And the new `Path` will be `"/mf-manifest.json"`

image

And finally, no `err` and all good!

image

Nexucis (approved):
awesome work, thanks !

PR #4065, opened 2026-05-13
[FEATURE] Add edit display name to root level folder edit action

### LINKED ISSUE #4059: Extend model for Folder with displayName
### Proposal

Current Folder model doesn't allow to define display name. This restrict us from being able to rename folders or use names with spaces.

Follow up to: https://github.com/perses/perses/pull/3966 discussed with @jgbernalp

adrianSepiol:
# Description

Since Folders are not in use this is added as feature instead of breaking change. The content of folder is now stored in spec.items and display name is in spec.display.name which matches other resources. 

Closes: https://github.com/perses/perses/issues/4059

# Screenshots

[Screencast from 13-05-26 11:22:47.webm](https://github.com/user-attachments/assets/3aab8aba-3aa3-482b-bd06-54540ef5d194)

# Checklist

## UI Changes


ui/app/src/utils/dashboardTableUtils.test.ts
jgbernalp, line 448: I think we are asserting the name of the folder after creating the table rows.

pkg/model/api/v1/folder.go
```diff
@@ -56,20 +61,25 @@ func (f *FolderSpec) UnmarshalYAML(unmarshal func(any) error) error {
 	return nil
 }
 
-func (f *FolderSpec) validate() error {
+func (f *FolderItem) validate() error {
```
jgbernalp, line 64: We should probably add some tests for this.

jgbernalp, line 64: Do we accept folders without name?

adrianSepiol: Good catch, we probably should not. I added assertion and tests for that case.

ui/app/src/utils/dashboardTableUtils.test.ts
adrianSepiol: I didn't get that one, could you elaborate?

jgbernalp: Missed an important "not".  We are not asserting that the folder name is set as expected

jgbernalp (approved):

PR #4077, opened 2026-05-21
[ignore] deprecate proxy definition in favor of the one coming from perses/spec

Nexucis:


internal/api/impl/proxy/proxy.go
```diff
@@ -673,8 +676,8 @@ func (s *sqlProxy) openPostgres(tlsConfig *tls.Config) (*sql.DB, error) {
 		query.Set("prepareThreshold", fmt.Sprintf("%d", *s.config.Postgres.PrepareThreshold))
 	}
 
-	if s.config.Postgres.ConnectTimeout != 0 {
-		query.Set("connect_timeout", s.config.Postgres.ConnectTimeout.String())
+	if len(s.config.Postgres.ConnectTimeout) > 0 {
```
celian-garcia: beware that len("0") > 0 will be yes. So there is a change here

celian-garcia (commented):
Not sure I understand the impact of this. I feel like it's only a move of code. 
Is it bigger than that?

internal/api/impl/proxy/proxy.go
celian-garcia: I see that the type is DurationString and not Duration . What is the difference between both ?

Nexucis:
> Not sure I understand the impact of this. I feel like it's only a move of code. Is it bigger than that?

We created the repo `perses/spec` that contains the Dashboard and Datasource specification. And since the creation of this new repo, we are moving code there and deprecates code that is here in favor of the spec.

So yes it is basically just moving code. To avoid am immediate breaking change we are deprecating the code here.

internal/api/impl/proxy/proxy.go
Nexucis: https://github.com/perses/spec/blob/main/go/common/durationstring.go#L18-L26

> DurationString is a string that represents a duration, such as "1h", "30m", "15s", etc.
> It is used to unmarshal a duration string from JSON or YAML, and validate that it is a valid duration string.
> 
> Not converting the duration string into a time.Duration type allows us to avoid the issue of changing the initial input with an alias.
> This is something that happens when we use Duration type, because when the duration is unmarshalled then marshaled again, the input can be changed with an equivalent duration.
> For example "14d" will be changed to "2w".
> 
> So, use DurationString instead of Duration when you want to preserve the original input string.
> If, for any reason, you need to convert the DurationString to a time.Duration, you can use the ParseDuration function.

Nexucis:
mmm I fear I will have to deactivate temporary the tests of the dac build ... It's because the go sdk of the plugins prometheus needs to be updated with the latest change from the go sdk because even if the type are similar they are not coming from the same package and so it does not compile.

celian-garcia:
> mmm I fear I will have to deactivate temporary the tests of the dac build ... It's because the go sdk of the plugins prometheus needs to be updated with the latest change from the go sdk because even if the type are similar they are not coming from the same package and so it does not compile.

You can move to an alias instead of deprecate. Like this 
```go
package http

import spechttp "github.com/perses/spec/go/datasource/proxy/http"

// Temporary migration alias.
// Old package now re-exports the new type.

type Proxy = spechttp.Proxy
```

So the type is really the same. But the version needs to be updated everywhere.

Would that help ?

Nexucis:
that won't help because then the go-sdk is still using a type from `perses/perses` rather than the one from `perses/spec`. Go is making the difference even if it is an alias.
Alias is used when you want to override a type and add a top of it other things. Like for example overriding the marshalling. You are then creating another type to do it. Like the durationString for example.

Nexucis:
perhaps this is the sign we should move the go-sdk outside of perses/perses like Guillaume was suggesting few months ago 🤔

celian-garcia (approved):

Issue #3548, opened 2025-11-07
Grafana migration: support row repeat


AntoineThebaud:
### Proposal

Since Perses now has the [panel group repeat](https://github.com/perses/perses/pull/3106) feature, it should be possible to enrich the migration logic to migrate "Row repeat" from Grafana into it.

gfierkasap:
I can take a look

PR #4079, opened 2026-05-22
[ENHANCEMENT] migrate Grafana row repeat to Perses grid layout repeatVariable

gfierkasap:
## Summary 

Solves #3548

Grafana supports repeating a row for each value of a template variable via the `repeat` field on row panels. Analogous functionality is supported in Perses via the `repeatVariable` field in grid layout, but so far the Grafana to Perses migration logic did not allow to map one to the other. This PR intends to extend its functionality to allow for repeat row migration.


Nexucis (approved):
awesome, thank you @gfierkasap !

PR #4084, opened 2026-05-26
[FEATURE] By configuration, enable or disable plugins

### LINKED ISSUE #4075: Runtime plugins management
### Proposal

Currently is not possible to:

- Install/uninstall plugins at runtime
- Disable/Enable installed plugins

It would be great that this should be possible

Nexucis:
This PR is proposing a way to disable or enable the various plugin available in the folder `plugins`. 

It fix #4075

Now, with the following config

```yaml
plugin:
  disabled:
    - "pyroscope"
    - "victoriaLogs"
    - "clickHouse"
    - "tempo"
```

I have now only one datasource activated and only the prometheus explorer available
Screenshot 2026-05-26 at 12 33 55

### DIFF
--- a/internal/api/plugin/plugin.go
+++ b/internal/api/plugin/plugin.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


internal/api/plugin/plugin.go
```diff
@@ -97,6 +100,10 @@ type pluginFile struct {
 	loaded tree.Tree[*Loaded]
 	// devLoaded is a map that contains all the loaded plugin modules in development mode.
 	devLoaded tree.Tree[*Loaded]
+	// enabled is the list of plugin or module that will be kept when loading them from the file system. If empty, all plugins/modules will be loaded.
+	enabled []string
```
celian-garcia, line 104: Usually the term is more inclusion vs exclusion.

celian-garcia (approved):
LGTM. I would suggest another name but fine to me.

internal/api/plugin/plugin.go
Nexucis: that's really a weird way to call a plugin activation.

PR #4095, opened 2026-06-01
[ignore] bump perses/common

Nexucis:


celian-garcia (approved):
Ok to me, do we have impacts to manage ?

Nexucis:
yes the flags are not initialised anymore. You need to call manually the function `InitFlag`.

PR #4100, opened 2026-06-03
[ignore] trigger workflow to cleanup old docker image

Nexucis:
I would like to add a script to cleanup the docker images we are building for the main branch older than one month.

Today we have created around 8000 tags just with the commit. It's a bit too much and I think a cleanup would help.

Perhaps we could do the same on quay.io if it is possible


celian-garcia (approved):
LGTM. Do you have the dry run result ?

Issue #3899, opened 2026-02-20
percli: add `dac watch` command


AntoineThebaud:
### Proposal

My goal is to provide a new command `percli dac watch` that would provide a development workflow similar to frontend hot-reload, where changes to Dashboard-as-Code files are immediately compiled and made to Perses.

I'm going to split the work in 3 PRs:
- [ ] base version of the command: monitor file changes to rebuild the right dashboards -> at this stage we'd already have a MVP since the output folder can be passed to provisioning. 
- [ ] add an option to replace the usage of provisioning by making proper CRUD requests to the Perses API when changes are detected
- [ ] auto browser reload - make it work at least for the locally-hosted Perses instances

celian-garcia:
Regarding the following points
> - add an option to replace the usage of provisioning by making proper CRUD requests to the Perses API when changes are detected (this is a foreseen improvement for big repos since provisioning is reloading everything)
> 
> - add automatic browser reload, to improve further the dev UX. The goal is at least to make it work for a locally-running Perses instance

I wouldn't do CRUD actually. I'd prefer something more elaborated still relying on provisioning. Here is my take on it; some new features that would be **enabled in dev only**:
- New configuration to **enable watch mode on provisioning**. Every update of provisioning update the database.
- A new endpoint ``/api/v1/watch-events`` (to be defined) exposing ``server-sent events`` ([SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events)) to the frontend
- A new frontend event listener that will invalidate the dashboards queries when anything happen on it.

With all this, in theory, you have an experience close to the frontend live development with even no refresh of the page, only refresh of the dashboard.

celian-garcia:
I just created a discussion design with more details as it might be useful for some other use cases. 
https://github.com/perses/perses/discussions/4123

celian-garcia:
Needs a lot of code review, but working so far. I am very happy of the result already! I can't wait to make some dac with that. 🥲 

Spoiler:
<img width="1868" height="1044" alt="Image" src="https://github.com/user-attachments/assets/2447493e-85ca-4649-9e0a-3d0b31f16ba9" />

PR #4126, opened 2026-06-12
[ENHANCEMENT] Add new ``api/v1/watch/*`` endpoints to watch resource updates

celian-garcia:
# Description

This PR is the implementation of the following design #4123 

Relates to #3899 

In a nutshell, this PR can be summarized with these following flows:

- Subscriber flow
``dashboardEndpoint.Watch`` – HTTP endpoints definition
  \`-> ``toolbox.Watch`` [common] – SSE implementation + permission
  \`-> ``dashboardService.Watch`` – Only transition to dashboardPersistence (minimal to avoid duplications)
  \`-> ``dashboardPersistence.Watch`` – Only transition to DB Subscriber (minimal to avoid duplications)
  \`-> ``database.EventWatcher`` [common] – Subscription implementation to broadcast properly events to all subscribers
- Publish flow (one instance by kind in the future)
``dashboardPersistence.NewDAO`` –  initiate a watchableDAO (decorator of dao watching create/update/deletes)
 \`-> ``database.WatchableDAO`` [common] – Create / Upsert / Delete decorated with publish
 \`-> ``database.EventWatcher`` [common] – do the actual publish 

Currently it's doing only the ```/api/v1/watch/(projects/<project>)?/dashboards``` endpoint. 
We could extend later to other endpoints or to a specific dashboard following the same principle.

# Screenshots

# Checklist

## UI Changes

N/A


s3onghyun (commented):
Really useful addition — a proper watch API. The locking in `eventWatcher` looks correct to me: unsubscribe removes from the map and `close()`s the channel under the write lock, while `dispatch` only sends under the read lock, so there's no send-on-closed-channel race. Two things worth considering, both around the overflow behavior.

### 1. Silent drop of the *newest* event can leave a watcher permanently stale

Both `Publish` and `dispatch` use the same "queue full → drop oldest, keep newest" pattern:

```go
select {
case p.events <- event:
default:
    select { case <-p.events: default: }   // drop one
    select { case p.events <- event: default: }  // re-send
}
```

Two concerns:

- **It doesn't actually guarantee "keep the latest" under concurrent publishers.** DAO writes happen concurrently, so two `Publish` calls can interleave between the drain and the re-send; the second `default` is then hit and the *new* event is dropped instead of an old one. So the event that gets lost can be the newest one.
- **More importantly, for a watch API the failure mode is "client misses the final update and stays stale."** If a client's subscriber buffer (64) overflows during a burst and the dropped event happens to be the last `DELETE`/`UPDATE` for a resource, the client never hears about it and remains out of sync until some unrelated event for that resource arrives — which may be never.

Kubernetes handles exactly this by not silently dropping: when the server can't keep a watch caught up it closes it with `410 Gone`, and the client re-lists to re-sync. Even if a full re-sync protocol is out of scope here, it'd be worth either (a) signalling the client on overflow so it can re-list, or (b) documenting explicitly that the watch stream is lossy so consumers periodically reconcile rather than treating it as authoritative.

### 2. No teardown for the dispatch goroutine

`NewEventWatcher` starts `go w.dispatch()`, which runs `for event := range p.events` forever — `p.events` is never closed and there's no `Close()`/`Stop()`. Since a watcher is created once per watch-enabled DAO at startup, this is benign for the running process. But there's no graceful teardown, and any test that constructs DAOs repeatedly will accumulate dispatch goroutines for the life of the test binary. A `Close()` that closes `p.events` (letting `dispatch` return) and is called from the persistence manager's shutdown path would cover both.

(Reviewing as a contributor, no merge rights — feel free to disregard #2 if lifecycle is handled elsewhere.)

celian-garcia:
Thank you @s3onghyun for your inputs.

For information, we discussed with @Nexucis, this PR might be closed in favor of a two steps work. Not yet decided, but it's probable
- A PR introducing an internal watch, that could be useful in several aspects of the code. Specially with cache sync (e.g.  dashboard search index, permissions)
- Another afterward to expose it via server sent events.

We'll consider these inputs when implementing them.

s3onghyun:
Makes sense — factoring out an internal watch first sounds like the cleaner foundation, and it could also give the schema cache in #4234 a more general invalidation signal than the current load/unload hooks. Happy to review whichever way it splits 🙂

PR #4136, opened 2026-06-17
[ENHANCEMENT] Allow provisioning by watching folders

celian-garcia:
# Description

Relates to #3899

#### 1/ Introduce a new config: 
```yaml
# Can be used to update provisioning on change instead of waiting for the configured interval.
# Interval is still used to reconcile the state.
enable_watch: <boolean> | default = false # Optional
```

Note that there is a debounce mechanism currently hardcoded to 500ms. I'm not sure it's relevant to have it in config.

As a bonus, this PR also fixes the always update behavior, that is leading to high version numbers and unnecessary writes in the DB. 
image

Now it's only updating the resource if it's different from the database one. Excluding of course the always changing fields like ``version``, ``updatedAt``, ``createdAt``.
```go
// stripTransientMetadata removes metadata fields that are managed by the server and
// should not be considered when comparing whether an entity's content has changed.
func stripTransientMetadata(m map[string]interface{}) {
	meta, ok := m["metadata"].(map[string]interface{})
	if !ok {
		return
	}
	delete(meta, "version")
	delete(meta, "updatedAt")
	delete(meta, "createdAt")
}
```

# Screenshots

# Checklist

## UI Changes

N/A

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -618,6 +618,10 @@ See the [TLS Config](../api/secret.md#tls-config-specification) specification.
 ```yaml
 interval: <duration> | default = 1h # Optional
 
+# Can be used to update provisioning on change instead of waiting for the configured interval.
+# Interval is still used to reconcile the state.
+enable_watch: <boolean> | default = false # Optional
+
 # List of folder that Perses will read periodically. 
 # Every known data found in the different folders will be injected in the database regardless what exist.
 folders:
--- a/internal/api/provisioning/provisioning.go
+++ b/internal/api/provisioning/provisioning.go
@@ -16,237 +16,151 @@ package provisioning
 import (
 	"context"
 	"fmt"
-	"sort"
+	"os"
+	"path/filepath"
+	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/perses/common/async"
-	databaseModel "github.com/perses/perses/internal/api/database/model"
 	"github.com/perses/perses/internal/api/dependency"
-	apiInterface "github.com/perses/perses/internal/api/interface"
-	"github.com/perses/perses/internal/cli/file"
-	"github.com/perses/perses/internal/cli/resource"
-	modelAPI "github.com/perses/perses/pkg/model/api"
-	modelV1 "github.com/perses/perses/pkg/model/api/v1"
 	"github.com/sirupsen/logrus"
 )
 
-type insertFunc func() (modelAPI.Entity, error)
+// watchDebounce is the quiet period after the last filesystem event before a reload is
+// triggered. This batches rapid successive changes (e.g. a directory of files being
+// written at once) into a single applyEntity call.
+const watchDebounce = 500 * time.Millisecond
 
-func New(serviceManager dependency.ServiceManager, folders []string, caseSensitive bool) async.SimpleTask {
-	return &provisioning{
+func New(serviceManager dependency.ServiceManager, folders []string, caseSensitive bool) (async.SimpleTask, async.SimpleTask) {
+	svc := &provisioningService{
 		serviceManager: serviceManager,
-		folders:        folders,
 		caseSensitive:  caseSensitive,
 	}
+	return &provisioningTask{
+			folders: folders,
+			svc:     svc,
+		}, &provisioningWatcher{
+			folders: folders,
+			svc:     svc,
+		}
 }
 
-type provisioning struct {
-	async.SimpleTask
-	serviceManager dependency.ServiceManager
-	folders        []string
-	caseSensitive  bool
+type provisioningTask struct {
+	svc     provisioningServiceInterface
+	folders []string
 }
 
-func (p *provisioning) Execute(_ context.Context, _ context.CancelFunc) error {
+func (p *provisioningTask) Execute(_ context.Context, _ context.CancelFunc) error {
 	if len(p.folders) == 0 {
 		return nil
 	}
-	var entities []modelAPI.Entity
-	for _, dir := range p.folders {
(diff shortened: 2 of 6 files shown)


internal/api/provisioning/provisioning.go
```diff
@@ -16,209 +16,154 @@ package provisioning
...
-	modelAPI "github.com/perses/perses/pkg/model/api"
-	modelV1 "github.com/perses/perses/pkg/model/api/v1"
 	"github.com/sirupsen/logrus"
 )
 
-type insertFunc func() (modelAPI.Entity, error)
+// watchDebounce is the quiet period after the last filesystem event before a reload is
+// triggered. This batches rapid successive changes (e.g. a directory of files being
+// written at once) into a single applyEntity call.
+// watchDebounce is the quiet period after the last filesystem event before a reload is
```
Nexucis: This line is the same than the first line of this comment.

pkg/model/api/config/provisioning.go
```diff
@@ -21,6 +21,9 @@ type ProvisioningConfig struct {
 	Folders []string `json:"folders,omitempty" yaml:"folders,omitempty"`
 	// Interval is the refresh frequency
 	Interval common.Duration `json:"interval,omitempty" yaml:"interval,omitempty"`
+	// EnableWatch can be used to update provisioning on change instead of waiting for the configured interval.
+	// Interval is still used to reconcile the state.
+	EnableWatch bool `json:"enableWatch,omitempty" yaml:"enableWatch,omitempty"`
```
Nexucis:
```suggestion
	EnableWatch bool `json:"enable_watch,omitempty" yaml:"enable_watch,omitempty"`
```

the config is following the snake case

internal/api/provisioning/service.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type insertFunc func() (modelAPI.Entity, error)
+type getEntityFunc func() (modelAPI.Entity, error)
+
+// specChanged returns true if the incoming entity's content differs from the existing one.
+// It compares JSON representations after stripping transient metadata fields (version,
+// updatedAt, createdAt) so that a no-op provisioning reload does NOT trigger an update.
+func specChanged(incoming, existing modelAPI.Entity) bool {
```
Nexucis: you should just compare the spec instead, that will reduce the number of marshalling and unmarshalling operation that reduce drastically the performance of the provisioning.

Nexucis:
```suggestion
reflect.DeepEqual(incoming.GetSpec(), existing.GetSpec())
```

internal/api/provisioning/provisioning.go
```diff
@@ -16,209 +16,154 @@ package provisioning
...
+	svc     provisioningServiceInterface
+	folders []string
 }
 
-type provisioning struct {
-	async.SimpleTask
-	serviceManager dependency.ServiceManager
-	folders        []string
-	caseSensitive  bool
+type provisioningWatcher struct {
```
Nexucis: can you put this struct after the functions attached to the struct `provisioningTask`. It will be easier to understand what struct and function is doing what

internal/api/provisioning/service.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		if !databaseModel.IsKeyConflict(createErr) {
+			logrus.WithError(createErr).Errorf("unable to create the %q %q", kind, name)
+			continue
+		}
+
+		// The entity already exists. Only update it if the content has actually changed
+		// to avoid triggering unnecessary SSE events and frontend refreshes.
+		existing, getErr := getFunc()
```
Nexucis: I am not a big fan of checking the state of the database. The provisioning principal is to ignore what is the current state and override no matter what.

I would be against changing this behaviour. The watching system should not impact that.

Nexucis (changes requested):
Thanks for this new feature.

I have put few comments and you need to solve the conflict you have. 

Excepting few details, I am fine with the feature.

docs/configuration/configuration.md
```diff
@@ -615,10 +615,15 @@ See the [TLS Config](../api/secret.md#tls-config-specification) specification.
...
+# Interval is still used to reconcile the state.
+enableWatch: <boolean> | default = false # Optional
+
 # List of folder that Perses will read periodically. 
-# Every known data found in the different folders will be injected in the database regardless what exist.
+# Every known data found in the different folders will be injected in the database.
+# Resources will be created if new or updated if changed.
 folders:
-  - <string>
+  - <string
```
Nexucis: this is a weird change

docs/configuration/configuration.md
```diff
@@ -615,10 +615,15 @@ See the [TLS Config](../api/secret.md#tls-config-specification) specification.
 ```yaml
 interval: <duration> | default = 1h # Optional
 
+# Can be used to update provisioning on change instead of waiting for the configured interval.
+# Interval is still used to reconcile the state.
+enableWatch: <boolean> | default = false # Optional
```
Nexucis:
```suggestion
enable_watch: <boolean> | default = false # Optional
```

internal/api/provisioning/service.go
celian-garcia: I don't  remember well. I will have to check again. I think that this was applying several time same entity because of that.

I will remove it anyway and fix after if there is a problem

internal/api/provisioning/service.go
celian-garcia: N/A because was used after the removed getFunc

Nexucis (approved):
great feature ! Well done !

Issue #4134, opened 2026-06-15
Migration from Grafana to Perses is broken


ibakshay:
### What did you do?

1. Import a Grafana dashboard. 
2. Click on `Migrate` button
3. In all the panels, Loki datasource is added

```json
{
        "kind": "Panel",
        "spec": {
          "display": {
            "name": "test"
          },
          "plugin": {
            "kind": "TimeSeriesChart",
            "spec": {}
          },
          "queries": [
            {
              "kind": "LogQuery",
              "spec": {
                "plugin": {
                  "kind": "LokiLogQuery",
                  "spec": {
                    "datasource": {
                      "kind": "LokiDatasource"
                    },
                    "query": "test"
                  }
                }
              }
            }
          ]
        }
      },
````

### What did you expect to see?

Panels in Migrated dashboard should not have any  loki datasource as it is not configured in first place.

### System information

Kubernetes

### Perses version

0.54.0-beta.1
I have tested with `0.53.1` and it works without any issues.

ibakshay:
@gfierkasap Could you please work on it?

gfierkasap:
@ibakshay Sure, let me look into it

celian-garcia:
@ibakshay @gfierkasap is that solved then ?

ibakshay:
yes, this issue can be closed. thanks to @gfierkasap!

PR #4139, opened 2026-06-17
[IMPROVEMENT] Updated migrateQuery to catch ambiguous plugin matches during migration

gfierkasap:
# Description

This PR is a result of working on https://github.com/perses/perses/issues/4134, and intends to help avoid similar bugs in the future.

This PR introduces changes to the `migrateQuery` function, responsible for populating the `spec.queries` part of the dashboard schema.
Originally, `migrateQuery` iterated over all `*queryInstance` objects, and when it encountered a fitting plugin it populated the schema and broke out of the iteration loop.
This idea behind this PR is to change the way this function works - instead of breaking out of the loop, all fitting plugins are collected, and warnings are logged when multiple plugins are found.

This way would allow to catch any situations where mistakes / bugs in plugin migrate CUE files leads to wrongly migrated dashboards


internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +159,36 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
-			result.Spec.Queries = append(result.Spec.Queries, dashboard.Query{
-				Kind: string(query.kind),
-				Spec: dashboard.QuerySpec{
-					Plugin: *queryPlugin,
-				},
-			})
-			isQueryMigrationEmpty = false
-			break
-		}
+		matchedQueries = append(matchedQueries, matchedQuery{query, plugin})
```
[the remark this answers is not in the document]
gfierkasap: Remediated in [fd337fc](https://github.com/perses/perses/pull/4139/commits/fd337fc965472263e8d481d4cf6b9da54bb7b57f)

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +159,36 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
+	}
+	if len(matchedQueries) > 1 {
+		logrus.Warnf("ambiguous query migration: %d plugins matched the same target", len(matchedQueries))
+		// should this return true as well, since there are too many matches?
+		return true
 	}
```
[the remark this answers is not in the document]
gfierkasap: TODO remoevd in [ccc9219](https://github.com/perses/perses/pull/4139/commits/ccc9219099d26a761650a7ca82baad6def3dd80c)

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +159,36 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
-			})
-			isQueryMigrationEmpty = false
-			break
-		}
+		matchedQueries = append(matchedQueries, matchedQuery{query, plugin})
+	}
+	if len(matchedQueries) > 1 {
+		logrus.Warnf("ambiguous query migration: %d plugins matched the same target", len(matchedQueries))
+		// should this return true as well, since there are too many matches?
+		return true
```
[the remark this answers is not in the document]
gfierkasap: Tests added in [7a662a3](https://github.com/perses/perses/pull/4139/commits/7a662a380c3ac063f05bc0402deec856205e6725)

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +159,42 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
 
+type matchedQuery struct {
+	query  *queryInstance
+	plugin *plugin.Plugin
+}
+
+// executeQueryScript is a package-level variable so tests can stub it without real CUE files.
+var executeQueryScript = ExecuteQueryScript
+
 func migrateQuery(queries map[string]*queryInstance, target json.RawMessage, result *dashboard.Panel) bool {
```
celian-garcia, line 180: I would challenge the name of the variable here queries, not your PR but it took me some time to understand it wasn't a query in itself but the script to execute on the target. 
Maybe I will propose a refactoring in another PR

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +159,42 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
 		}
+		matchedQueries = append(matchedQueries, matchedQuery{query, plugin})
 	}
-	return isQueryMigrationEmpty
+	if len(matchedQueries) > 1 {
+		logrus.Warnf("ambiguous query migration: %d plugins matched the same target", len(matchedQueries))
+		return true
+	}
+	if len(matchedQueries) == 0 {
+		return true
```
celian-garcia, line 207: This deserve a warning as well to me. In an ideal world we would like to be compatible with 100% of queries and thus we want to see when a target is incompatible. And print the target

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +159,42 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
-			isQueryMigrationEmpty = false
-			break
+		if isEmpty {
+			continue
 		}
+		matchedQueries = append(matchedQueries, matchedQuery{query, plugin})
 	}
-	return isQueryMigrationEmpty
+	if len(matchedQueries) > 1 {
+		logrus.Warnf("ambiguous query migration: %d plugins matched the same target", len(matchedQueries))
```
celian-garcia: Can you print the target ?

celian-garcia (changes requested):
If I understand the PR, it means that it doesn't fix the issue but now, instead of applying the wrong migration script (first found among several compatible) it will apply no script, precisely because several are compatible. Am I correct ? 

Some little changes proposed, but look fine, thanks !


Edit: Please can you see what happened into the commit history and check why you have a DCO missing ?

gfierkasap:
> If I understand the PR, it means that it doesn't fix the issue but now, instead of applying the wrong migration script (first found among several compatible) it will apply no script, precisely because several are compatible. Am I correct ?
> 
> Some little changes proposed, but look fine, thanks !
> 
> Edit: Please can you see what happened into the commit history and check why you have a DCO missing ?

That's exactly right - this does not fix the mentioned bug, but rather helps catch and log any situations where ambiguity  in CUE migration files might lead to wrongly assigned plugins.

As to the DCO signoff, seems something weird happened when trying to merge main when I updated the branch. Let me see if I can untangle this mess

EDIT: signoff issue mitigated

internal/api/plugin/migrate/panel.go
gfierkasap: Sure, I added a new warning in [f27add5](https://github.com/perses/perses/pull/4139/commits/f27add58d4aba607661a7be1958d7383609ffdbd)

internal/api/plugin/migrate/panel.go
gfierkasap: That's a good idea, this will make the warning a bit more clear; added it in [9980bef](https://github.com/perses/perses/pull/4139/commits/9980bef9ab546c97045eaf9017a39085c679aae5)

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +160,52 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
 	}
 }
 
+type matchedTarget struct {
```
celian-garcia: you renamed to matchedTarget which introduce confusion actually

internal/api/plugin/migrate/panel.go
```diff
@@ -159,26 +160,52 @@ func (m *completeMigration) migrateQueries(targets []json.RawMessage, result *da
...
-			})
-			isQueryMigrationEmpty = false
-			break
+		if isEmpty {
+			continue
 		}
+		matchedQueries = append(matchedQueries, matchedTarget{query, plugin, queryKind})
+	}
+	if len(matchedQueries) > 1 {
+		logrus.Warnf("ambiguous query migration: %d plugins matched the same target: %s", len(matchedQueries), sprintKinds(matchedQueries))
```
celian-garcia: Instead of – or on top of – printing the matched queries, you could print the ``target json.RawMessage`` variable as it's really the guilty part when an error occurs to me. It can then easily be used as test data in plugins repo.

internal/api/plugin/migrate/panel.go
celian-garcia: Here again can you print the guilty ``target json.RawMessage``

celian-garcia (changes requested):

internal/api/plugin/migrate/panel.go
gfierkasap: Yeah, later when I thought about it I came to the same conclusion. Reverted this change in [ddca96a](https://github.com/perses/perses/pull/4139/commits/ddca96a814748062d632c4130b4d79f46a088945)

internal/api/plugin/migrate/panel.go
gfierkasap: That's a good idea. I added it over a few commits, ultimately I wrapped it in a function `targetDebugMsg`. The one small thing I changed was instead of extending the already present log warning, I added a separate log containing the target json with log lvl debug, as it seemed more appropriate. Wdyt?

EDIT: I kind of overcomplicated it before with this `targetDebugMsg` function, refactored the logging in [8fba1c1](https://github.com/perses/perses/pull/4139/commits/8fba1c12a159239be2177ad68e514adcf15d289a)

celian-garcia (approved):

PR #4154, opened 2026-06-19
[BUGFIX] refuse provisioning of resources in a non-existing project (#3936)

### LINKED ISSUE #3936: Provisioning: can provision dashboards for a project that doesn't exist
### What did you do?

Currently no error fires if you try to provision Dashboard resources for a project that doesn't exist - I guess the same applies to other project resources like Variable & Datasource.

This is the case at least for the DB in file mode. In this case the files get created without triggering any error and you wont see them on the UI, until you create the project with the corresponding name.

I didn't check but the problem should also occur with a database in SQL mode.

Perses should refuse to provision the project-level resources that belong to non-existing projects.

### Perses version

0.53.0

MohammadYusif:
# Description

The provisioning service iterated over provisioned entities and called each service's `Create` without verifying that the parent project existed. As a result, a project-scoped resource (Dashboard, Datasource, Variable, Role, RoleBinding, Secret, Folder) declared for a project that doesn't exist was silently persisted to the database and stayed invisible in the UI until a project with the matching name was later created.

This adds a project-existence check in `applyEntity`: any project-scoped resource whose project doesn't exist is skipped and reported via an error log. Global resources and Project resources (which carry no project) are unaffected. A lookup error other than "not found" is treated as "exists" so a transient database error doesn't silently drop the resource.

A regression test (`internal/api/e2e/api/provisioning_test.go`) provisions one resource into an existing project and one into a missing project, and asserts the former is created while the latter is refused.

Fixes #3936

# Screenshots

N/A — no UI changes.

# Checklist

- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention (`BUGFIX`).

## UI Changes

- [ ] N/A — no UI changes.


internal/api/e2e/api/provisioning_test.go
```diff
@@ -0,0 +1,69 @@
...
+	e2eframework "github.com/perses/perses/internal/api/e2e/framework"
+	"github.com/perses/perses/internal/api/provisioning"
+	testUtils "github.com/perses/perses/internal/test"
+	modelAPI "github.com/perses/perses/pkg/model/api"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProvisioningSkipsResourcesInUnknownProject(t *testing.T) {
+	conf := e2eframework.DefaultConfig()
+	_, _, manager := e2eframework.CreateServer(t, conf)
```
celian-garcia: Cool to see some provisioning tests popping there!
Any reason why you don't use the simpler ``WithServer`` function ? 
https://github.com/perses/perses/blob/807f6de317981c1877cd6a3addcfccecde279562/internal/api/e2e/framework/server.go#L200

You should find a lot of examples althrough the e2e/api folder

celian-garcia: Oh I'm answering myself, you need the service manager. It seems the ``WithServerAuthConfig`` addressed this need giving to the test the global dependency manager instead of just the persistence. I will propose a little refactoring to align.

https://github.com/perses/perses/blob/807f6de317981c1877cd6a3addcfccecde279562/internal/api/e2e/framework/server.go#L209

celian-garcia (changes requested):
Hey @MohammadYusif, thank you for that. I'm currently touching the provisioning in another PR https://github.com/perses/perses/pull/4136 and I like the way you did tests! I might steal you the idea :) 
I'm not 100% satisfied of the ``WithServer`` test interface, let me propose a refactor first, that will simplify your test.

celian-garcia:
It looks like the UI tests are failing because of this fix
```
[WebServer] time="2026-06-20T07:26:48Z" level=error msg="unable to provision the \"Variable\" \"job\": project \"perses\" doesn't exist"
[WebServer] time="2026-06-20T07:26:48Z" level=error msg="unable to provision the \"Folder\" \"demos\": project \"perses\" doesn't exist"
[WebServer] time="2026-06-20T07:26:48Z" level=error msg="unable to provision the \"Folder\" \"infrastructure\": project \"perses\" doesn't exist"
[WebServer] time="2026-06-20T07:26:48Z" level=error msg="unable to provision the \"Folder\" \"panels\": project \"testing\" doesn't exist"
[WebServer] time="2026-06-20T07:26:48Z" level=error msg="unable to provision the \"Folder\" \"charts\": project \"testing\" doesn't exist"
[WebServer] time="2026-06-20T07:26:48Z" level=error msg="unable to provision the \"Folder\" \"features\": project \"testing\" doesn't exist"
```

Can you make sure the missing projecta are provisioned properly first ?
It makes me think that now the order will be important in the provisioning folder. So is it a breaking change in itself and do we accept it ? @Nexucis

MohammadYusif:
Good catch — thanks. The UI fixtures declare projects and their project-scoped resources (folders, variables) in the same provisioning batch, and since `filepath.Walk` reads the folder in lexical order, a child could be processed before its parent project, so the new guard refused it.

I pushed `b8eff14`: `applyEntity` now applies every `Project` first (stable sort, relative order otherwise) so any project declared in the same batch is created before the resources that reference it.

On the breaking-change concern: this actually makes provisioning **less** order-dependent, not more. The user's file ordering inside the provisioning folder no longer matters — projects are always applied before their children regardless of filenames, so the fixtures work without reordering. The only behaviour change remains the original one: a resource targeting a project that exists *nowhere* (not on disk, not in the batch) is refused instead of silently persisted.

Added an integration test (`TestProvisioningCreatesProjectBeforeItsResources`) that provisions a project + a role belonging to it in one batch with the role file read first, asserting both are created (it fails without the ordering fix). Happy to align the test setup with whatever framework helper you land for exposing the service manager.

internal/api/provisioning/provisioning.go
```diff
@@ -70,11 +71,28 @@ func (p *provisioning) String() string {
...
 	for _, entity := range entities {
 		entity.GetMetadata().Flatten(p.caseSensitive)
 		kind := modelV1.Kind(entity.GetKind())
 		name := entity.GetMetadata().GetName()
 		project := resource.GetProject(entity.GetMetadata(), "")
+		// A project-scoped resource cannot be provisioned if its parent project doesn't exist:
+		// the resource would be silently persisted but never visible until the project is created.
+		// So we refuse to provision it and report the missing project.
+		if len(project) > 0 && !p.projectExists(project) {
+			logrus.Errorf("unable to provision the %q %q: project %q doesn't exist", kind, name, project)
```
Nexucis:
```suggestion
logrus.Errorf("unable to provision the %q/%q: project %q doesn't exist", kind, name, project)
```

Nexucis:
excepting my little comment, that looks good to me.
Thank you for doing that @MohammadYusif !

@celian-garcia order mattered since the beginning in the provisioning. That's why we have prefixed the dev files with a number to ensure it is taken in the correct order.

@MohammadYusif can you signoff your commits please. We cannot merge the PR until the DCO is happy.

Regarding the tests, the PR from @celian-garcia  is going to be merged soon (as soon as the CI is passing). Once it is done, can you rebase your PR and apply the change @celian-garcia  proposed ? Thank you by advance.

Nexucis:
@MohammadYusif let us know when you update the PR with the changes required. Once it is done we will be able to merge this PR.

MohammadYusif:
Thanks for your patience @Nexucis and @celian-garcia — I've updated the format string to use `%q/%q` as suggested. The branch is ready for re-review whenever you have a chance.

internal/api/e2e/api/provisioning_test.go
celian-garcia: @MohammadYusif can you try to use the ``WithServer`` helper instead of managing the setup/teardown of the server yourself ?

celian-garcia (changes requested):
Thanks @MohammadYusif there is still some actions to take on your side. And also the DCO is a must please https://github.com/perses/perses/pull/4154/checks?check_run_id=83238709697

MohammadYusif:
Done — two updates in the latest push:

1. **DCO sign-offs:** Added `Signed-off-by: MohammadYusif <MohammadYusif@users.noreply.github.com>` to all four commits via `git rebase --exec 'git commit --amend --no-edit -s'`. DCO should now be happy.

2. **Test refactor:** Added `WithServiceManager` to the e2e framework (`internal/api/e2e/framework/server.go`) — analogous to `WithServer` but passing the full `dependency.Manager` so that `ServiceManager` is available for tests that drive internal services directly. Both provisioning tests now use it, removing the manual `CreateServer` / `defer` / `ClearAllKeys` boilerplate and aligning with the pattern used by the rest of the `e2e/api` suite.

internal/api/e2e/framework/server.go
```diff
@@ -197,7 +197,19 @@ func CreateServer(t *testing.T, conf apiConfig.Config) (*httptest.Server, *httpe
...
+	defer dependencyManager.Persistence().GetPersesDAO().Close()
+	defer server.Close()
+	entities := testFunc(server, expect, dependencyManager.Persistence())
+	ClearAllKeys(t, dependencyManager.Persistence().GetPersesDAO(), entities...)
+}
+
+// WithServiceManager is like WithServer but exposes the full dependency.Manager
+// (including Service()) rather than just PersistenceManager. Use this for tests
+// that need to drive internal services directly (e.g. provisioning).
+func WithServiceManager(t *testing.T, testFunc func(*httptest.Server, *httpexpect.Expect, dependency.Manager) []modelAPI.Entity) {
```
celian-garcia: You did a copy of the new WithServer function. Let me update your branch and you'll see.

celian-garcia (changes requested):

internal/api/e2e/framework/server.go
celian-garcia: Updated. Now you can simply remove this function and use directly ``WithServer`` 🙂

internal/api/provisioning/provisioning.go
MohammadYusif: Already applied in this format — the `kind/name` pattern was in the original commit. Thanks!

internal/api/e2e/api/provisioning_test.go
MohammadYusif: Done — switched both provisioning tests to use `WithServer` directly and removed the `WithServiceManager` helper. Thanks for the pointer!

internal/api/e2e/framework/server.go
MohammadYusif: Done — `WithServiceManager` removed and both tests updated to use `WithServer` directly. Thanks for making the update!

celian-garcia (approved):
Looks good now thanks!

You give me some work with that on my watch feature 😄 if I want the orphans to be automatically created when the parent pops up...

Nexucis:
great, thank you @MohammadYusif.
Can you fix the code format? The import in the go code are not in the correct order.
Once this is done I think we will good to merge this one.

MohammadYusif:
Fixed — moved `github.com/gavv/httpexpect/v2` ahead of the perses imports so the block is in alphabetical order, matching the convention in the rest of the e2e suite.

Nexucis (approved):
awesome, thank you for your contribution @MohammadYusif !

PR #4167, opened 2026-06-24
[DOC] Fix go-sdk doc about the listvariable

Nexucis:
import path are wrong in the doc

### DIFF
--- a/docs/dac/getting-started.md
+++ b/docs/dac/getting-started.md
@@ -7,8 +7,8 @@
 
 ### Prerequisites
 
-- `percli`, the [CLI of Perses](../cli.md). ⚠️ The version should be at least `v0.51.0`.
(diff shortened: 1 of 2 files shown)


docs/dac/getting-started.md
```diff
@@ -7,8 +7,8 @@
 
 ### Prerequisites
 
-- `percli`, the [CLI of Perses](../cli.md). ⚠️ The version should be at least `v0.51.0`.
```
celian-garcia, line 10: Do you really want to say that the minimum should be this ? People can actually use older version no ?

celian-garcia (approved):

docs/dac/getting-started.md
Nexucis: this won't be the case after the version since the different import path have been changed and also it contains new query type. So to avoid people complaining, I prefer to say use the last version.
Hopefully yes you won't need to upgrade afterwards the CLI for few versions

PR #4169, opened 2026-06-24
[BUGFIX] Pin timeZone to UTC in date unit tests to prevent locale/tim…

Teja079:
While working on date formatting in the codebase I noticed that several tests  in `ui/core/src/model/units/date.test.ts` were silently relying on the test  runner being in UTC to pass.

The affected tests cover the *-local date units — `datetime-local`, `date-local`,  and `time-local`. These units are designed to format timestamps using the  browser's local timezone, so the `formatDate` function intentionally inherits  the host timezone when none is specified. The tests, however, were asserting  exact time strings like `"12:30:45"` or `"01/01/2024, 12:30:45"` which are  only correct if the runner happens to be in UTC.

The `npm test` script sets `cross-env TZ=UTC` as a band-aid, but this doesn't  hold in all environments. On containerized CI runners the OS-level timezone is  set independently of the process environment, so `TZ=UTC` has no effect and  the tests fail. On a developer machine set to a non-UTC timezone (e.g.  Asia/Tokyo, America/Los_Angeles) the same failures show up locally.

To reproduce: running the suite with `TZ=Asia/Tokyo` (UTC+9) before this fix  produces 7 failures like:

  Expected: "12:30:45"
  Received: "21:30:45"

  Expected: "01/01/2024, 12:30:45"
  Received: "01/01/2024, 21:30:45"

The fix is to make each of these test cases self-contained. Since the tests are  validating locale-specific formatting (not timezone conversion), passing  `timeZone: 'UTC'` alongside the existing `locale` option is the right call —  it pins the time component to a known value without changing what the test is  actually checking.

The `"should handle default options"` test is also updated separately. That test  calls `formatDate` with no options at all, which means it uses the system timezone by design. Asserting on the exact time string there is wrong in principle — the test should only verify that the output looks like a date, not what the time portion says. It now checks for the `MM/DD/YYYY` pattern only.

After the fix, all 63 tests pass with both `TZ=UTC` and `TZ=Asia/Tokyo`.

### DIFF
--- a/ui/core/src/model/units/date.test.ts
+++ b/ui/core/src/model/units/date.test.ts
@@ -49,7 +49,7 @@ const TEST_CASES: DateTestCase[] = [
   // DateTime Local format (depends on locale)
   {
     value: TEST_TIMESTAMP_MS,
-    format: { unit: 'datetime-local', locale: 'en-US' } as DateFormatOptions,
+    format: { unit: 'datetime-local', locale: 'en-US', timeZone: 'UTC' } as DateFormatOptions,
     expected: '01/01/2024, 12:30:45',
   },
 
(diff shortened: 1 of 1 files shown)


Teja079:
Hi, could a maintainer approve the CI workflows to run? Thanks!

jgbernalp:
@Teja079 the core package is being deprecated and will be removed in future releases. This specific test was moved to https://github.com/perses/shared/blob/main/components/src/model/date.test.ts would you mind update it there?

Nexucis:
Hello @Teja079,
As stated by Gabriel, this PR needs to go to the repository `perses/shared`.
Please open the PR there. In the meantime I am closing this PR.

Thank you for your time and your contribution !

Issue #3891, opened 2026-02-19
Pass user id_token through proxy


abogdanov37:
### Proposal

Hi everyone!
I try migrate my own datasource from Grafana to Perses.
At that time I have BFF which use OIDC to authenticate and can separate returned data by user. User 1  can see all data, User 2 can see part of data, User 3 can see another part of data.

I deploy Perses and my BFF and connect there to the same Keycloak using OIDC protocol

Perses correctly authenticate user!

I setap my datasource use proxy.
But I can't find way to send user token through Perses proxy to my BFF. 
Than I setup my datasource to use direct access but it does not send token too (((

My idea is add new secret type, named PassThrough for example, which will collect user token from cookies or use token stored on Perses backend (if it stored) and send request to my BFF.  

I also view Perses source code to find ability to implement custom secret as a plugin but not found any points to extend ((((

Nexucis:
Hello @abogdanov37,

as you figure out, we currently don't store the token received from external authentication provider in any way. So probably a way to mitigate this issue would be to simply store the token in a cookie or to provide it in a header and then in the datasource configuration provide a way to decide if yes or not you want to use this token.

@celian-garcia do you think this is as simple ? Or do I miss something ?

celian-garcia:
Hello, indeed currently we use the provider's access token and id token only at login time and the reason for that is the tokens TTL.
If we need to keep a valid OIDC access token, we would have to synchronize the TTL of the sessions OIDC and Perses. Making sure the access & refresh tokens are both usable would be painful in my opinion.

We can think of a solution but that will not be simple. How was it configured in Grafana ? I don't think they store the token they do ?

Last dumb question. What is BFF ?

abogdanov37:
> Hello [@abogdanov37](https://github.com/abogdanov37),
> 
> as you figure out, we currently don't store the token received from external authentication provider in any way. So probably a way to mitigate this issue would be to simply store the token in a cookie or to provide it in a header and then in the datasource configuration provide a way to decide if yes or not you want to use this token.
> 
> [@celian-garcia](https://github.com/celian-garcia) do you think this is as simple ? Or do I miss something ?

Small clarification. At that time Perses **store** token in cookies. But I can't use it ((( 

Periodicaly Perses refresh token from the UI and get id_token. But I can't inject code in this proccess (((

And another one moment through proxy I can't send Authorization header because it will be dropped

abogdanov37:
> Last dumb question. What is BFF ?

[...]

PR #4172, opened 2026-06-26
[FEATURE] Add claim-based role assignment from OAuth/OIDC access tokens

### LINKED ISSUE #3137: Support Role Assignment Based on Scopes from Access Token

gfierkasap:
# Description

Closes #3137

## Summary

Perses currently assigns roles exclusively via database RoleBinding objects, requiring per-user manual setup. This PR adds the ability to map claims from an OAuth/OIDC token (e.g. roles: ["admin"]) to Perses roles dynamically, so users automatically receive the correct permissions based on their identity provider's attributes — no RoleBinding required.

## How it works

1. Config — operators define claim→role mappings per provider:
```yaml
  authentication:
    providers:
      oidc:
        - slug_id: keycloak
          claims:
            - claim_name: roles
              mappings:
                - claim_value: admin
                  role_name: admin          # → GlobalRole "admin" (due to no project defined)
                - claim_value: ops
                  role_name: editor
                  project: platform         # → project-scoped Role "editor"
```
2. Token — at login, matching claim values are extracted from the ID token and stored in the Perses JWT as `prc` (persisted claims). Both string and []string claim value types are supported.
3. Authorization — on every request, HasPermission, GetPermissions, and GetUserProjects resolve permissions from prc in addition to DB-backed RoleBinding objects. Unknown role names produce a warning log rather than a hard error. Claim-derived permissions are evaluated after guest checks and before DB cache lookups.

## Manual test stack

For the purpose of manual tests, a full Keycloak-based test stack has been added under dev/ for manual end-to-end verification (requires docker compose):

- dev/docker-compose.keycloak.yaml — spins up Keycloak on :8080 and Perses (built from source) on :8090
- dev/keycloak-realm.json — preconfigured realm with 4 users (alice, bob, carol, dave), roles, and a roles protocol mapper
  - dev/config.keycloak.yaml — Perses config with claim mappings for all 4 roles
  - dev/keycloak-provisioning/ — seed data (GlobalRoles, project, project Role)
  - dev/TESTING.txt — step-by-step runbook with expected results for each test user

To start: `docker compose -f dev/docker-compose.keycloak.yaml up --build --detach`

NOTE: This manual test stack is just a temporary part of this PR, in case someone wants to test this and not have to build a testing environment. Ultimately, once they're not needed, these files should not be merged and will be removed from this PR in a following commit.

# Screenshots

# Checklist


dev/keycloak-provisioning/1-globalroles.json
```diff
@@ -0,0 +1,30 @@
+[
```
celian-garcia, line 1: ouch I'm not sure. Include a full keycloak environment to make some tests will be a nightmare to maintain. It's quite opinionated. To me e2e tests with a mock server returning token with claims should be the way to test.

celian-garcia (changes requested):
A first message, reading at a glance. Now I'm pulling and review deeper.

Edit: At least from the description, I like what I read 🙂

dev/keycloak-provisioning/1-globalroles.json
gfierkasap: Hi @celian-garcia, thanks for taking a look at it! 

I think I wrote it poorly in the description, my idea with that docker compose was to just temporarily add it to the PR, in case someone wanted to test the solution manually and not have to create the entire testing environment from scratch :) I don't think these parts should ultimately be merged.

I'll update the PR description to highlight the purpose of these files

raphaelpierre:
Hi, thanks for working on this! Scope-based role mapping would be really useful for our setup. Is there a rough timeline or milestone for when this might land? Happy to help test if that's useful.

celian-garcia:
Thank you @raphaelpierre for manifesting your interest! It's waiting for my review sorry. I'll make it tomorrow.

No clear ETA
0.54.0 is in the pipeline but it won't be in for sure.
We will try to get it in 0.55 that won't come before a couple of month.

ibakshay:
> Thank you @raphaelpierre for manifesting your interest! It's waiting for my review sorry. I'll make it tomorrow.
> 
> No clear ETA 0.54.0 is in the pipeline but it won't be in for sure. We will try to get it in 0.55 that won't come before a couple of month.

Thank you @celian-garcia! I have tested manually last week and it works as expected. 😀

internal/api/authorization/native/native.go
celian-garcia:
```suggestion
	// cache is used to store in memory the permissions of all users and all the roles defined
```

internal/api/authorization/native/native.go
```diff
@@ -245,41 +326,50 @@ func (n *native) GetPermissions(ctx echo.Context) (map[string][]*v1Role.Permissi
...
+		for project, perms := range n.claimPermissions(claims) {
+			userPermissions[project] = append(userPermissions[project], perms...)
+		}
+	}
 	return userPermissions, nil
 }
 
 func (n *native) RefreshPermissions() error {
-	permissions, err := n.loadAllPermissions()
+	permissions, globalRoles, roles, err := n.loadAllPermissions()
```
celian-garcia: loadAllPermissionsAndRoles

internal/api/authorization/native/native.go
```diff
@@ -245,41 +326,50 @@ func (n *native) GetPermissions(ctx echo.Context) (map[string][]*v1Role.Permissi
...
+	usr, _ := n.GetUser(ctx)
+	if claims, ok := usr.(*crypto.JWTClaims); ok {
+		for project, perms := range n.claimPermissions(claims) {
+			userPermissions[project] = append(userPermissions[project], perms...)
+		}
+	}
 	return userPermissions, nil
 }
 
 func (n *native) RefreshPermissions() error {
```
celian-garcia: RefreshPermissionsAndRoles

internal/api/authorization/native/native.go
```diff
@@ -176,6 +227,25 @@ func (n *native) GetUserProjects(ctx echo.Context, requestAction v1Role.Action,
 		logrus.Error("failed to get username from context to list the user projects")
 		return nil, apiInterface.InternalError
 	}
+
+	// Claim-based check: if wildcard permission found via claims, short-circuit.
+	usr, _ := n.GetUser(ctx)
+	if claims, ok := usr.(*crypto.JWTClaims); ok {
+		claimPerms := n.claimPermissions(claims)
+		if listHasPermission(claimPerms[v1.WildcardProject], requestAction, requestScope) {
```
celian-garcia: The process to retrieve projects from permissions is the same there and right after. It could be put into a function to avoid duplicate. Also there is a slight difference in yours. You don't check the that it contains WildProject permission before accessing. This is probably not necessary indeed, but to be sure I would keep it

internal/api/authorization/native/native.go
```diff
@@ -217,6 +287,17 @@ func (n *native) HasPermission(ctx echo.Context, requestAction v1Role.Action, re
 	if ok := listHasPermission(n.guestPermissions, requestAction, requestScope); ok {
 		return true
 	}
+	// Claim-based check (stateless, from JWT)
+	usr, _ := n.GetUser(ctx)
```
celian-garcia: multiple time you do GetUser without returning the error. Can you return the error every time ?

internal/api/e2e/framework/server.go
```diff
@@ -417,3 +417,105 @@ func NewOIDCProviderTestServer(t *testing.T) (*httptest.Server, apiConfig.OIDCPr
 
 	return server, conf
 }
+
+// NewOIDCProviderTestServerWithClaims is like NewOIDCProviderTestServer but the /token
+// endpoint returns tokens that carry the given extraClaims (e.g. map[string]any{"roles": []string{"admin"}}).
+// Use this to test claim→role mapping without modifying the standard mock server.
+func NewOIDCProviderTestServerWithClaims(t *testing.T, extraClaims map[string]any) (*httptest.Server, apiConfig.OIDCProvider) {
```
celian-garcia: Modify the existing to not copy paste. If the map is set it uses it, if nil it doesn't use it. WDYT ?

internal/api/impl/auth/oauth.go
```diff
@@ -373,7 +375,8 @@ func (e *oAuthEndpoint) codeExchangeHandler(ctx echo.Context) error {
 		return err
 	}
 
-	_, err = e.performUserSync(uInfo, ctx.SetCookie)
+	persistedClaims := extractPersistedClaims(uInfo.(*oauthUserInfo).RawProperties, e.claimConfigs)
```
celian-garcia, line 378: Here we actually get "claims" from the output of /userinfos and not the token claim. Probably this is something that will need to be fixed or improved in the future. 

Not coming from your code though, this is the requestUserInfos that could be improved. So that's ok.

internal/api/impl/auth/oidc.go
celian-garcia: That would be good to parse the client credentials token as well.

celian-garcia (changes requested):
Well done! We're totally in the right direction to me!

Generally I think there should be a function to parse any token to retrieve the claim that you could reuse everywhere. Here you are taking sometimes userinfo result, sometimes claims from id token already parses. If you had a function that parse the token you could reuse it directly to parse oauth2 token, oidc id token raw, or oidc client credentials token. It would simplify the logic of saving claims to persist from external provider.

Also can you update the documentation with your new configs. They're perfectly fine to me. I was originally thinking of a support more elaborated with a language or even just regex but it can be made later if we have the need.

internal/api/authorization/native/native.go
gfierkasap: [71b444b](https://github.com/perses/perses/pull/4172/commits/71b444b816906fe7b8732e0eab514ce2560eeb71)

internal/api/authorization/native/native.go
gfierkasap: [b662704](https://github.com/perses/perses/pull/4172/commits/b66270480ba9a1bbf883625f4bc3b270993e75a8)

internal/api/impl/auth/oidc.go
gfierkasap: [e9a8b9e](https://github.com/perses/perses/pull/4172/commits/e9a8b9e14c5e23f23c304295f2c5cb7742f8f7a8)

internal/api/e2e/framework/server.go
gfierkasap: Sure, this makes sense. No need to redo the entire thing just for a separate scenario with claims, added in [71ef3fd](https://github.com/perses/perses/pull/4172/commits/71ef3fdd298e06487add0243622828d1ca98b7f5)

internal/api/authorization/native/native.go
gfierkasap: Added missing error handlers in [21e58d3](https://github.com/perses/perses/pull/4172/commits/21e58d3564e98a2851064e439690f917b6adcaa0)

internal/api/authorization/native/native.go
gfierkasap: Fair point, a lot of duplication going on. I added a new helper called `projectsWithPermission` in [5e9fdbf](https://github.com/perses/perses/pull/4172/commits/5e9fdbf5f09c725d228c6dd870829ac6151202ae), I think it solves this

internal/api/authorization/native/native.go
gfierkasap: [de7fa34](https://github.com/perses/perses/pull/4172/commits/de7fa3438198e824b5572046656b89bc48cfe159)

internal/api/authorization/native/native.go
```diff
@@ -283,7 +286,10 @@ func (n *native) HasPermission(ctx echo.Context, requestAction v1Role.Action, re
 		return true
 	}
 	// Claim-based check (stateless, from JWT)
-	usr, _ := n.GetUser(ctx)
+	usr, err := n.GetUser(ctx)
+	if err != nil {
+		return false
```
celian-garcia, line 292: Can you log the error before returning false ?

docs/concepts/authentication.md
```diff
@@ -114,6 +111,55 @@ security:
...
+                  role_name: "dashboard-editor"
+                  project: "myproject"   # maps to Role "dashboard-editor" in project "myproject"
+```
+
+### How it works
+
+1. At login, Perses extracts the configured claim values from the upstream access token (or ID token for OIDC).
+2. The extracted values are embedded in the Perses session JWT.
+3. On every request, Perses evaluates the claim→role mappings and resolves the corresponding permissions — stateless, no extra database lookup.
+4. Resolved permissions are additive with any explicit `RoleBinding`/`GlobalRoleBinding` the user may have.
```
celian-garcia: Additive is maybe not clear of what it does in case of conflicts. 
Can you provide an example where a user is set to admin role by claim and editor (less priviledged role) by config?

celian-garcia (changes requested):
Hi @gfierkasap great! Just some nit and that is good for me 🙂 
Thanks for your patience and dedication! ❤️ 

I tested it locally with Azure OIDC and Github oauth2, using email as mapping ^^ 
And it works perfectly! 🚀 

Can you please
- [x] Make the few minor changes
- [x] Make the tests pass
- [x] Remove the keycloak test

And that will be good for me

s3onghyun (commented):
The claim→role resolution in `claimPermissions` looks solid from a security standpoint — default-deny when there are no claims/mappings, exact `slices.Contains` matching (no case-folding or substring widening), per-provider mapping isolation via `providerKey`, correct GlobalRole-vs-project scoping, and unknown roles grant nothing. Nice.

One design point worth surfacing (or at least documenting) around **revocation semantics**, which I don't think has come up yet:

Claim-based grants are captured once at login and then carried forward on every refresh without re-consulting the IdP. From the PR's own `TESTING.txt`:

> The refresh endpoint reads PersistedClaims from the refresh token and copies them to the new access token without re-querying Keycloak.

Because each refresh also re-persists those same claims into a **new** refresh token with a fresh TTL, an actively-used session keeps its original claim snapshot alive indefinitely. The practical consequence: if an operator removes a user's `roles: [admin]` (or the equivalent group) at the IdP, that user keeps Perses admin until their refresh-token chain lapses and they're forced to re-login — which, for a user who stays active, may be never.

Two things make this easy to miss:

- It's asymmetric with the rest of the system. The `RefreshPermissionsAndRoles` cron keeps the *role definitions* current, so changing what `admin` is allowed to do propagates quickly — but a user's *membership* in `admin` (via claim) is frozen at login.
- It's weaker than the existing `RoleBinding`/`GlobalRoleBinding` path, which is evaluated live against the DB, so revoking a binding takes effect immediately. Operators moving to claim-based mapping would reasonably expect the same and be surprised.

Options, roughly in increasing cost:

1. At minimum, document this explicitly in `authorization.md` — "IdP-side revocation of a claim does not take effect until the user's refresh token expires; size `access_token_ttl` / `refresh_token_ttl` accordingly." Operators can then bound the exposure.
2. Bound it structurally by not extending the claim lifetime past the original refresh token (i.e. don't reset the claim snapshot's effective expiry on each refresh), so revocation is guaranteed within one refresh-token TTL.
3. Re-fetch/re-validate claims from the IdP on refresh (proper, but adds an IdP round-trip and depends on the provider returning updated claims on the refresh grant).

Not a blocker — the feature is genuinely useful as-is — but the revocation window is the kind of thing that should be a conscious, documented choice rather than an implicit one. (Reviewing as a contributor, no merge rights.)

celian-garcia:
Hi @s3onghyun, thank you for the review. May I ask you if you're using an AI generally for your reviews / comments? I have no objection and this is a good thing to have context and all, but also keep in mind that it's a human who reads this. So if you can maybe low down the cursor on the verbosity, we would appreciate a lot ^^'. Also you can comment on the piece of code that it is about. A general comment like this ask some effort to find back the right location.

---

Back to the main point. If I understand properly, the concern is about the fact that the roles computed at login time will be valid during the entire lifetime of the session. I think this is quite structural on the way we handle the sessions in perses. **Once we login, the IdP is not taken in account anymore, the session and its lifetime fully belong to Perses. Even the TTL can be different.** 

> 1. At minimum, document this explicitly in authorization.md — "IdP-side revocation of a claim does not take effect until the user's refresh token expires; size access_token_ttl / refresh_token_ttl accordingly." Operators can then bound the exposure.

If it's not clear enough we can emphasize this. Don't hesitate to make a suggestion.

> 2. Bound it structurally by not extending the claim lifetime past the original refresh token (i.e. don't reset the claim snapshot's effective expiry on each refresh), so revocation is guaranteed within one refresh-token TTL.

This would revoke the access after the first refresh ??
> 3. Re-fetch/re-validate claims from the IdP on refresh (proper, but adds an IdP round-trip and depends on the provider returning updated claims on the refresh grant).

Probably valid, the big missing point here is that we don't keep a trace of the original tokens from the IdP. This is discussed here https://github.com/perses/perses/issues/3891.

internal/api/authorization/native/native.go
gfierkasap: Logged the error in [7b768c4](https://github.com/perses/perses/pull/4172/commits/7b768c491030b1c5f7b5c36fea87bc0e6cafed4b)

gfierkasap:
@celian-garcia @s3onghyun Thank you both for your reviews. I hope in the recently pushed changes address your concerns. 

The test stack was removed in [f7e1f96](https://github.com/perses/perses/pull/4172/commits/f7e1f965aa12d67b3c8f5184d442cea268b4df1d) and the documentation (along with some clarification and examples) were changed in [e5307a7](https://github.com/perses/perses/pull/4172/commits/e5307a79b65be9a373fbfca73aaca8e860d7b289).

docs/concepts/authentication.md
gfierkasap: An example was added in [e5307a7](https://github.com/perses/perses/pull/4172/commits/e5307a79b65be9a373fbfca73aaca8e860d7b289)

celian-garcia (approved):
Thank you @gfierkasap. LGTM you addressed my last nits, and did the minimal for the @s3onghyun concern.

PR #4173, opened 2026-06-26
[ENHANCEMENT] migrate G2P dashboard links

### LINKED ISSUE #1642: Dashboard links
### Proposal

Today we are lacking a way to quickly jump from a dashboard to another with already-set variables. Thus it would be great to have links as part of the dashboard data model, ideally both at top level & at panels level.

### LINKED ISSUE #4162: [grafana dashboard migration] Dashboard links
### Proposal

It seems the dashboard links are not migrated in the grafana migration logic.

shahrokni:
fixes https://github.com/perses/perses/issues/4162
fixes https://github.com/perses/perses/issues/1642

This enhancement migrates the Grafana Dashboard Links into the Perses Dashboard Spec structure. 
The following screenshots demonstrate a successful migration.

The variables are replaced by client. Therefore, the original URL with the template variables could be kept intact and as it is.  For instance here, 

1. I used both `${myvar}` and `$myvar`
2. Since the variable had been created already, it was successfully replaced

image

image

# Checklist

## UI Changes

### DIFF
--- a/internal/api/plugin/migrate_test.go
+++ b/internal/api/plugin/migrate_test.go
@@ -116,9 +116,24 @@ func TestMigrateDashboardLinks(t *testing.T) {
 	testCases := []struct {
 		name          string
 		grafanaLink   migrate.GrafanaLink
-		expectedLink  *dashboard.Link
-		expectNilLink bool
+		expectedLinks []dashboard.Link
 	}{
+		{
+			name: "Link with URL only",
+			grafanaLink: migrate.GrafanaLink{
+				URL:     "http://fakedomain/${samplevar}/$samplevar",
+				Tooltip: "",
+			},
+			expectedLinks: []dashboard.Link{
+				{
+					URL:             "http://fakedomain/${samplevar}/$samplevar",
+					TargetBlank:     false,
+					RenderVariables: false,
+					Tooltip:         "",
+					Name:            "",
+				},
+			},
+		},
 		{
 			name: "Full link with variables and target blank",
 			grafanaLink: migrate.GrafanaLink{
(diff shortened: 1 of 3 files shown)


jgbernalp:
@shahrokni I think we are missing the tests for this

internal/api/plugin/migrate_test.go
```diff
@@ -109,6 +109,48 @@ func TestMig_MigrateTags(t *testing.T) {
 	assert.Equal(t, set.New("ops", "prod"), persesDashboard.Metadata.Tags)
 }
 
+func TestMigrateDashboardLinks(t *testing.T) {
+	pl := LoadTestPlugins()
+	grafanaDashboard := &migrate.SimplifiedDashboard{
```
jgbernalp: We probably should follow the format on other test and something I see in go regularly. To create a collection with the input and expected output and iterate over the test cases. In this case creating 3 dashboards each one with a set of links configuration., it will make it easier to rad and maintain if new cases are needed.

internal/api/plugin/migrate_test.go
```diff
@@ -109,6 +110,80 @@ func TestMig_MigrateTags(t *testing.T) {
...
+			persesDashboard, err := pl.Migration().Migrate(grafanaDashboard, false)
+			assert.NoError(t, err)
+			assert.NotNil(t, persesDashboard)
+
+			if tc.expectNilLink {
+				assert.Len(t, persesDashboard.Spec.Links, 0)
+			} else {
+				assert.Len(t, persesDashboard.Spec.Links, 1)
+				actualLink := persesDashboard.Spec.Links[0]
+				assert.Equal(t, tc.expectedLink.Name, actualLink.Name)
```
Nexucis: you could just do `assert.Equal(t, tc.expectedLink, actualLink)`

internal/api/plugin/migrate_test.go
```diff
@@ -109,6 +110,80 @@ func TestMig_MigrateTags(t *testing.T) {
...
+				Title: "Dashboard with links",
+				Links: []migrate.GrafanaLink{tc.grafanaLink},
+			}
+
+			persesDashboard, err := pl.Migration().Migrate(grafanaDashboard, false)
+			assert.NoError(t, err)
+			assert.NotNil(t, persesDashboard)
+
+			if tc.expectNilLink {
+				assert.Len(t, persesDashboard.Spec.Links, 0)
```
Nexucis: and instead of doing  `expectNilLink` you can have a slice of link as expected so if if needs to be empty then you just need to have an empty slice as expected.

So in summary you can just do `assert.Equal(t, tc.expectedLinks, persesDashboard.Spec.Links)`

celian-garcia:
@shahrokni hope you don't mind, I allowed myself to edit the desc with 2 fixes issues. Thanks for working on that <3

Nexucis (approved):
perfect,  thanks !

celian-garcia (approved):

PR #4174, opened 2026-06-26
fix(cli): skip non-main Go packages when building a directory (#3819)

### LINKED ISSUE #3819: CLI/DAC build: failed building go files successfully when using -d flag
### What did you do?

I have created a go module with the following struct:

```
main.go
|
|__ foo
      |
      |__ foo.go
```

When using `percli dac build -d my_module`, the command produce the following output:

```
Succesfully built my_module/main.go at my_module/main_output.yaml
  FAILED:
error processing file "my_module/foo/foo.go": failed to build my_module/foo/foo.go: package command-line-arguments is not a main package

Error: processing directory "my_module" failed, see the message(s) above
```

The command should not fail and should not process other things than the main package.

### System information

_No response_

### Perses version

v0.52.0, v0.53.0-rc.0

### Perses configuration file

```yaml

```

### Logs

```text

```

MohammadYusif:
## Summary

- `percli dac build -d <dir>` recursed into every subdirectory and tried to pass each `.go` file directly to `go run`. Files in non-main support packages (e.g. a helper library in a sub-directory) are not `package main`, so `go run` rejected them with *"package command-line-arguments is not a main package"* and the whole build failed.
- Fix: parse just the package clause of each `.go` file encountered during a directory walk using `go/parser.ParseFile(..., parser.PackageClauseOnly)`. Files not declaring `package main` are silently skipped with a debug-level log. The explicit `-f` / single-file path is unchanged — there the user pointed at a specific file and `go run` will report any error naturally.

## Changes

- `internal/cli/cmd/dac/build/build.go`: split the `goExtension` and `cueExtension` branches in the `filepath.Walk` callback; add `isGoMainPackage` helper that parses the package clause and returns `false` for non-main packages.
- `internal/cli/cmd/dac/build/build_test.go`: add regression test `"directory with non-main sub-packages skips them without error"` that walks `testdata/go` (which already contains `args/main.go` and `cmd/main.go` as `package main` alongside `prometheus/query/*.go` as `package query`) and asserts only the two main-package files are built.

Fixes #3819

## Checklist

- [x] Existing Go tests still pass (`nominal case with a single Go file`, `nominal case with a Go project`)
- [x] New regression test passes (`directory with non-main sub-packages skips them without error`)
- [x] `go build ./internal/cli/cmd/dac/build/...` clean
- [x] Only relevant files staged (no lockfiles, no generated code)
- [x] DCO `Signed-off-by` present in commit

### DIFF
--- a/internal/cli/cmd/dac/build/build.go
+++ b/internal/cli/cmd/dac/build/build.go
@@ -42,6 +42,20 @@ const (
 
 var folderToIgnore = []string{".git", ".github", ".idea", config.DefaultOutputFolder, "cue.mod"}
 
+// isGoMainPackage reports whether the Go source file at path belongs to
+// package main. Files in non-main packages (library helpers, sub-packages)
+// cannot be passed to `go run` directly and must be skipped during a
+// directory walk.
+func isGoMainPackage(path string) bool {
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, path, nil, parser.PackageClauseOnly)
+	if err != nil {
+		// If parsing fails, fall through so `go run` surfaces the real error.
+		return true
+	}
+	return f.Name.Name == "main"
+}
+
 type option struct {
 	persesCMD.Option
 	opt.FileOption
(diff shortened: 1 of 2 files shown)


internal/cli/cmd/dac/build/build.go
```diff
@@ -119,6 +133,20 @@ func (o *option) Execute() error {
 	return nil
 }
 
+// isGoMainPackage reports whether the Go source file at path belongs to
+// package main. Files in non-main packages (library helpers, sub-packages)
+// cannot be passed to `go run` directly and must be skipped during a
+// directory walk.
+func isGoMainPackage(path string) bool {
```
Nexucis: this function should at the top of the file to avoid to be mixed with the functions associated to `option`

internal/cli/cmd/dac/build/build.go
```diff
@@ -91,7 +93,19 @@ func (o *option) Execute() error {
 		}
 
 		extension := filepath.Ext(path)
-		if extension == goExtension || extension == cueExtension {
+		if extension == goExtension {
```
Nexucis: I prefere like this. That avoid a duplication of the code.
```go
if extension == goExtension || extension == cueExtension {	
	if extension == goExtension {
   		// Only process files declared as package main. Supporting packages that
		// live alongside a main entrypoint (e.g. helper packages in sub-dirs)
		// are valid Go code but cannot be passed to `go run` directly.
		if !isGoMainPackage(path) {
			logrus.Debugf("file %q has been ignored (not in package main)", path)
			return nil
		}
	}
	err = o.processFile(path, extension)
	if err != nil {
		// Append the error to highlight the issue to the user on a later stage but don't stop the processing
		errs = append(errs, fmt.Errorf("error processing file %q: %w", path, err))
	}
}
```

internal/cli/cmd/dac/build/build.go
MohammadYusif: Done — moved `isGoMainPackage` to package level above the `option` type.

internal/cli/cmd/dac/build/build.go
MohammadYusif: Applied — collapsed the go/cue branches into the nested-if form you suggested, removing the duplicated `processFile` call.

Nexucis (approved):
Thank you for the fix @MohammadYusif !

Issue #3994, opened 2026-03-30
`percli dac build` does not build json file


Nexucis:
### What did you do?

I tried to specify the output when building my dashboard using the CLI and it build a file with the extension `json`. But the content is written in YAML.

### What did you expect to see?

I am expecting that the json file generated by `percli` is actually a json file and not yaml file

### Perses version

v0.53.1

### Perses configuration file

```yaml

```

### Logs

```text

```

celian-garcia:
Little precision from my investigation so far. 
Cue supports well the json and yaml format. This is working the json output file is well in json.

The problem is scoped to Go SDK. 
Still investigating..

celian-garcia:
You were probably missing a ``flag.Parse()`` call in the beginning. It's in the doc. But I linked a PR to see if we can improve this.

celian-garcia:
Let's close the PR and the issue. When we follow the doc (sorry 😝 ) then it works well 🙂

PR #4175, opened 2026-06-26
[ENHANCEMENT] Ensure ``flag.Parse()`` is called by the go sdk itself.

celian-garcia:
# Description

This is fixing #3994. When ``flag.Parse()`` is called in corresponding main.go, it works well. I don't know if there's a reason why this needs to be called by the user but I prefer to encapsulate it. From what I undetstood ``sdk.NewExec()`` should be called only once at the beginning so it seems to be the perfect place to have that call.

# Screenshots

# Checklist

### DIFF
--- a/docs/dac/getting-started.md
+++ b/docs/dac/getting-started.md
@@ -184,7 +184,6 @@ import (
 )
 
 func main() {
-	flag.Parse()
 	exec := sdk.NewExec()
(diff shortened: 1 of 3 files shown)


celian-garcia:
It's quite dangerous actually as it gives less control for user on the args and cancel the ability to call it twice

PR #4187, opened 2026-06-30
[BUGFIX] reject users with both native and oauth providers

### LINKED ISSUE #2429: Users with both nativeProvider and oauthProviders can be created.

anxkhn:
# Description

A `User` spec is only valid with either a native provider (password) **or** one
or more oauth providers, never both. As described in #2429, the maintainers
decided these must be mutually exclusive to prevent account impersonation, and
oauth providers are meant to be attached internally at first login only.
Previously a spec setting both was silently persisted via `POST /api/v1/users`,
but the resulting user could not be used to log in.

This extends `User.validate()` to reject a spec that sets a native password and
at least one oauth provider, so the create API returns `HTTP 400` instead of
quietly saving an unusable user. The internal first-login OAuth flow
(`internal/api/impl/auth/service.go` `syncUser`) builds the user struct directly
and does not go through `UnmarshalJSON` / `validate()`, so it is unaffected. This
mirrors the existing invariant in `saveProviderInfo`, which already refuses to
attach an oauth provider to a user that has a native password.

A previous attempt at this (#2431) was closed because it also changed
`UserSpec.NativeProvider` from a value to a pointer (rippling into the CUE
schema, the e2e framework, and the auth tests). This change keeps the scope to
the validation guard only, by checking `len(NativeProvider.Password) > 0` rather
than nil-ness, so no type or schema change is needed.

Fixes #2429

# Screenshots

n/a (backend-only change)

# Checklist

## UI Changes

n/a - no UI changes.


Nexucis (approved):
Thank you for the fix @anxkhn !

anxkhn:
Thanks for the review! Just flagging that the only red check here is the `tests with file DB on windows` leg, which is failing on `main` as well (e.g. run 29001462936 on commit cc6e105) with the same e2e timeout, so it looks unrelated to this change (which only touches user spec validation). Happy to rebase if that would help re-trigger CI.

Nexucis:
yes can you rebase it please ? At least the job will run faster. But yes this is a flaky job, we are aware about it don't worry

PR #4191, opened 2026-06-30
[BUGFIX] Fix wrong minimum version and maximum version assignment in TLS config

Nexucis:


pkg/model/api/v1/secret/tls_config.go
```diff
@@ -146,5 +147,11 @@ func (c *TLSConfig) Verify() error {
 	if len(c.MaxVersion) == 0 {
 		c.MaxVersion = "TLS13"
 	}
+	if promConfig.TLSVersions[c.MinVersion] == 0 {
```
celian-garcia, line 150: this should be under the if len() > 0 no ?

Nexucis: that's not in the same function

celian-garcia: aaaah :D ok

celian-garcia (approved):

PR #4203, opened 2026-07-02
[BUGFIX] check all markdown files in checkdocs

vprashar2929:
# Description

`checkdocs` ran `fmt-docs` over every markdown file but only verified root-level *.md files with git diff.
Formatting drift under docs/ and other nested paths was never caught by the docs-fmt CI job.
In this commit we share the find command between fmt-docs and checkdocs so both targets operate on the same file set.

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,16 +46,21 @@ In the usual workflow, all PRs are squashed. There is two exceptions to this rul
 
 ## Documentation
 
-Documentation is written in Markdown. To ensure some quality on our documentation, we are
-running [mdox](https://github.com/bwplotka/mdox) that will ensure the doc is well formatted and all links are working.
+Documentation is written in Markdown. To ensure some quality on our documentation, we are running
+[mdox](https://github.com/bwplotka/mdox) on Markdown files that use GitHub Flavored Markdown. mdox ensures these files
+are well formatted and all links are working.
 
-To format the docs, you will have to install the tool mentioned above. An easy way is to run the following command:
+The `docs/` tree is imported by the [Perses website](https://github.com/perses/website) and rendered with MkDocs
+Material, so it is not formatted with mdox.
+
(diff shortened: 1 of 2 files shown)


Nexucis:
I am not sure we should do that actually. Because there are many things that has been done to format correctly the doc for the website (like correct space). And the rules are not the same unfortunately

vprashar2929:
> I am not sure we should do that actually. Because there are many things that has been done to format correctly the doc for the website (like correct space). And the rules are not the same unfortunately

You’re right docs follows `MkDocs` conventions and `mdox fmt` conflicts with that. Happy to drop the bulk reformat and change the fix to exclude docs from mdox while still aligning what `fmt-docs` formats with what `checkdocs` verifies for the remaining markdown. Thoughts?

Nexucis:
yup that sounds good to me @vprashar2929 🙏

docs/dac/cue/variable/list.md
```diff
@@ -3,8 +3,8 @@
 The List Variable builder helps creating list variables in the format expected by Perses.
 
 !!! warning
-	This lib is not meant for direct usage in DaC, as List variables are not standalone objects (they embed a plugin part).
-	This lib should be used by plugin developers to develop wrappers for their variable plugins - as it was done for official plugins.
+This lib is not meant for direct usage in DaC, as List variables are not standalone objects (they embed a plugin part).
```
Nexucis: I think with mkdocs the space is not correct anymore :(

vprashar2929: Agree the `docs/`  is used in publishing and rendered with Mkdocs so running mdox formatter on it will break Mkdocs specific syntax. I will scope mdox to GFM managed markdown and keep the `docs/` for Mkdocs specific formatter. WDYT?

Nexucis: sounds good :)

Nexucis (approved):
awesome, thank you!

PR #4218, opened 2026-07-06
[IGNORE]: Fix issue_template

ibakshay:
The end users are not able to create issues at all due to this bug.


jgbernalp (approved):

celian-garcia (approved):
Oops

PR #4226, opened 2026-07-07
[DOC] Schema configuration is deprecated

Nexucis:


ibakshay (approved):
I can add a deprecation flag for `schemas` property in helm-charts as wel, right? 

https://github.com/perses/helm-charts/blob/377f493ec5a78a289f80c8dcabf07543233f19a9/charts/perses/values.schema.json#L379-L404

Nexucis:
yes definitely you can @ibakshay.

The schema configuration is not used by the backend since two or three versions, so we can either remove it or deprecate it in the helm chart, depending how sensitive the helm chart is if you are adding key that does not exist

PR #4232, opened 2026-07-08
[BUGFIX] Auth: fake latency when user does not exist

Nexucis:


celian-garcia (approved):
Overkill to me, the latency of password comparison is minimal compared to DB query don't you think so ?
Anyway approved

Nexucis:
the function `bcrypt.CompareHashAndPassword` is intentionally slow, so the difference between returning directly an error if the user is not found vs comparing the password is visible for an attacker.

Basically the function comparing the password is slower than a DB query

PR #4234, opened 2026-07-08
[IMPROVEMENT] Cache schema endpoint responses to avoid redundant CUE recomputation

gfierkasap:
# Description

This PR is a result of one of the review suggestions in https://github.com/perses/perses/pull/4031

The GET `/api/v1/schemas/dashboards` and GET `/api/v1/schemas/plugins[/:name]` endpoints recomputed expensive CUE operations on every HTTP request. This PR adds an in-process cache for the computed schemas to `completeSchema` struct.

# Checklist

## UI Changes


internal/api/plugin/schema/schema.go
```diff
@@ -371,6 +405,94 @@ func (s *completeSchema) GenerateDashboardSchema() (cue.Value, error) {
 	return generateDashboardCueValue(ctx, plugins)
 }
 
+func (s *completeSchema) GenerateDashboardSchemaBytes() ([]byte, error) {
```
Nexucis: instead of doing that why not caching the response returned by `GenerateDashboardSchema()` ?

for others, I would not cache them for the moment until we have feedback it's necessary

gfierkasap: That's actually a really good point, especially since one of the other improvement suggestions is to implement json schema, and this would make it easier. Changed it in [5daaa10](https://github.com/perses/perses/pull/4234/commits/5daaa10771ec5a7fb658d372ca061da990b77ceb) and in following commits.

gfierkasap: As for the plugins caching, I added it since generating the plugin schema still seems like a decently expensive process. But it can be removed if necessary

Nexucis:
@gfierkasap as you might notice, the feature has been removed because there is an issue with the last version of cuelang crashing the migration from grafana to perses.

I have added a non-reg tests to ensure we are tracking this issue. In the meantime I have created a feature branch `feature/dashboard-schema-generation` containing the code about the dashboard schema generation.

Can you rebase your PR and target this branch instead of the main one ? Thank you 🙏

ibakshay:
> @gfierkasap as you might notice, the feature has been removed because there is an issue with the last version of cuelang crashing the migration from grafana to perses.
> 
> I have added a non-reg tests to ensure we are tracking this issue. In the meantime I have created a feature branch `feature/dashboard-schema-generation` containing the code about the dashboard schema generation.
> 
> Can you rebase your PR and target this branch instead of the main one ? Thank you 🙏

Just FYI: Gabor is on vacation. He will do after he is back next week:)

gfierkasap:
> @gfierkasap as you might notice, the feature has been removed because there is an issue with the last version of cuelang crashing the migration from grafana to perses.
> 
> I have added a non-reg tests to ensure we are tracking this issue. In the meantime I have created a feature branch `feature/dashboard-schema-generation` containing the code about the dashboard schema generation.
> 
> Can you rebase your PR and target this branch instead of the main one ? Thank you 🙏

@Nexucis @ibakshay thanks for the patience! Changed the pr to target the new feature branch

EDIT: after discussing the topic with @ibakshay today, I switched this PR back to target `main`, since from what I saw CUE was bumped back to v0.17 and the rollback of the PR with the dashboard schema was reverted.

EDIT 2: retargeted feature branch once again, I got misled by the release notes (they did mention the schemas were included in v0.54, but that's not true)

Nexucis:
@gfierkasap I think you are bringing more commits than the one you created.

A way to fix it, is to create a new branch based on the feature branch, and then cherry-pick the commit you created. Like that we should only the changes you made.

gfierkasap:
> @gfierkasap I think you are bringing more commits than the one you created.
> 
> A way to fix it, is to create a new branch based on the feature branch, and then cherry-pick the commit you created. Like that we should only the changes you made.

@Nexucis Good spot, I actually didn't notice the extra commits. I'll start working on this asap

EDIT: should be good now

internal/api/impl/v1/schema/endpoint.go
```diff
@@ -57,16 +55,13 @@ func (e *endpoint) CollectRoutes(g *route.Group) {
...
+	schema, err := e.pluginSvc.Schema().GenerateDashboardSchema()
 	if err != nil {
 		logrus.WithError(err).Error("unable to generate dashboard schema")
 		return apiinterface.InternalError
 	}
-
-	data, err := apiCue.Marshal(result)
+	data, err := apiCue.Marshal(schema)
 	if err != nil {
-		logrus.WithError(err).Error("unable to export dashboard schema as CUE")
```
ibakshay, line 69: [NOTE] Any reason why this logging is removed?

internal/api/impl/v1/schema/endpoint.go
```diff
@@ -84,43 +79,34 @@ func (e *endpoint) PluginDefinition(ctx echo.Context) error {
...
+		return apiinterface.InternalError
+	}
+	if !list.Exists() {
 		return ctx.Blob(http.StatusOK, contentType, []byte("{}"))
 	}
-	data, err := generateCUEbytes(schemas)
+	data, err := apiCue.Marshal(list)
 	if err != nil {
-		return err
+		return apiinterface.InternalError
```
ibakshay, line 109: [SUGGESTION] does it make sense to add logging here?

internal/api/impl/v1/schema/endpoint_test.go
```diff
@@ -56,6 +56,32 @@ func (s *stubSchema) GenerateDashboardSchema() (cue.Value, error) {
 	return cuecontext.New().CompileString("{}"), nil
 }
 
+func (s *stubSchema) GetAllPluginSchemas() (cue.Value, error) {
+	schemas := s.GetAllSchemas()
```
ibakshay, line 60: [SUGGESTION] No new tests verify that schemas are actually cached and not recomputed on repeated calls. Consider adding a test that checks the cache behavior.

internal/api/impl/v1/schema/endpoint_test.go
```diff
@@ -113,8 +139,8 @@ func TestDashboardSchemaWithNoPlugins(t *testing.T) {
 func TestPluginListWithNoSchemas(t *testing.T) {
 	ep := newEndpointWithSchemas(nil)
 	ctx, rec := newEchoContext(t, "/api/v1/schemas/plugins")
-
```
ibakshay, line 116: [NIT] Is this whitespace required?

ibakshay (commented):
Added some nits :)

internal/api/plugin/schema/schema.go
```diff
@@ -371,6 +395,77 @@ func (s *completeSchema) GenerateDashboardSchema() (cue.Value, error) {
...
+	s.mutex.RLock()
+	if cached, ok := s.cachedPluginByName[name]; ok {
+		s.mutex.RUnlock()
+		return cached, true, nil
+	}
+	s.mutex.RUnlock()
+
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	if cached, ok := s.cachedPluginByName[name]; ok {
```
Nexucis, line 453: this line is the same as before, so this is not necessary

internal/api/plugin/schema/schema.go
```diff
@@ -144,12 +144,19 @@ type Schema interface {
 	GetSchema(name, version, registry string) (LoadSchema, bool)
 	GetInstance(kind plugin.Kind, name string) (*build.Instance, error)
 	GenerateDashboardSchema() (cue.Value, error)
+	// GetAllPluginSchemas returns all plugin schemas serialised as CUE bytes.
+	// Results are cached until the next plugin load/unload event.
+	GetAllPluginSchemas() (cue.Value, error)
```
Nexucis, line 149: this new function makes `GetAllSchemas` unnecessary.

Nexucis (commented):
I have added few comments on top of what Askhay already reviewed.

Also I have few notes to make that question the purpose of this PR:

- Since it can cost some memories to cache all plugins schemas, I would avoid to do that until it is proved it is necessary.

- The generation of the dashboard schema can be interesting to cache in memory if it really takes time to generate it. I have the feeling it can be necessary for the dashboard. But it would be nice to have the proof it is the case.

- We already know the dashboard schema will be used for AI. But the plugin schema, we have actually no idea if it will be used or not. We have added it because it was possible. So until we have feedback this is used and without a cache it's not really usable, I would avoid to do that.

s3onghyun (commented):
One concurrency note on the caching, building on @ibakshay's suggestion to add a cache-behavior test: the three cached getters (`GenerateDashboardSchema`, `GetAllPluginSchemas`, `GetPluginSchema`) don't re-check the cache after upgrading from the read lock to the write lock, so the double-checked-locking is incomplete:

```go
s.mutex.RLock()
if s.cachedDashboardSchema.Exists() { ... return }  // miss
s.mutex.RUnlock()

s.mutex.Lock()
defer s.mutex.Unlock()
val, err := s.generateDashboardSchemaLocked()        // recomputes unconditionally
s.cachedDashboardSchema = val
```

If several requests hit a cold cache concurrently (e.g. a burst of dashboard loads right after a plugin load invalidates it), they all pass the `RLock` check, then each acquires the write lock in turn and regenerates — so you get N redundant CUE generations instead of 1, which is the cost this PR is trying to remove, just narrowed to the cold-cache window. Re-checking after taking the write lock closes it:

```go
s.mutex.Lock()
defer s.mutex.Unlock()
if s.cachedDashboardSchema.Exists() {
    return s.cachedDashboardSchema, nil
}
// ... generate + store
```

If the cache test @ibakshay suggested exercises a couple of concurrent first-callers, it'd naturally cover this. (Noting the retarget to `feature/dashboard-schema-generation` — flagging now since the pattern will likely carry over.)

Nexucis:
thank you for your investigation @s3onghyun this should be indeed considered if we make some benchmark with the cache activated.

For the moment the idea is more to make the benchmark without the cache to see if it is actually necessary.

@gfierkasap What do you want to do to move forward in this PR ?

Nexucis:
this PR has been closed automatically when we have merged the feature branch to main. It was not intentional, I am sorry about that.

But I think this PR can remain close for the moment until a proper benchmark is done and we have a proof it is necessary to do it and what needs to be improved:

- everything should be cached ?
- Actually just the dashboard schema
- None need to be in cache

PR #4236, opened 2026-07-08
[BUGFIX] Allow client to auto-refresh token

### LINKED ISSUE #3977: Automatic OAuth2 token refresh in client-credentials flow not happening for long-running operations
### What did you do?

(In below explanation, T = access token lifetime, which can be like 15min or 1h depending on the setup)

When interacting with Perses programmatically for batch operations with client-credentials OAuth2 authentication, for a long duration that exceeds T, token expiry causes all subsequent API requests to fail with `something wrong happened with the request to the API. Message: invalid or expired jwt StatusCode: 401"`. The HTTP client does not automatically refresh the token before expiry or when stale.

A workaround is on client side to detect 401 errors and recreate the REST client to force a new token acquisition, or recreate it every T, but that's an anti-pattern that shifts responsibility from the HTTP client library to application code.

⚠️ It remains unclear why 401 errors occur only in this case, knowing that:
- If a sync batch completes in < T, it succeeds entirely
- X hours later (much more than T) when the next batch gets triggered (that reuses the same client object) it succeeds initially before failing again after T
- This suggests the OAuth token refresh mechanism works during idle periods but fails under prolonged API load?

The Perses HTTP client's OAuth2 transport should transparently:
- Refresh the access token before expiry (proactive), OR
- Detect token expiry on 401 and automatically refresh without requiring client recreation (reactive)

### System information

_No response_

### Perses version

`github.com/perses/perses v0.52.0`

### Perses configuration file

```yaml

```

### Logs

```text

```

celian-garcia:
# Description

fix #3977 

We need to return the ``expires_in`` in the token answer to be sure that the client library will automatically do the refresh of the token. 

Ref: https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2

Specially the oauth2 client_credentials go library that is explicitely checking this field. 

Retrieval of the info https://github.com/golang/oauth2/blob/master/internal/token.go#L279-L322
verify if expired: https://github.com/golang/oauth2/blob/master/token.go#L138-L148

# Screenshots

# Checklist

### DIFF
--- a/internal/api/crypto/jwt.go
+++ b/internal/api/crypto/jwt.go
@@ -66,6 +66,7 @@ type JWT interface {
 	CreateRefreshTokenCookie(refreshToken string) *http.Cookie
 	DeleteRefreshTokenCookie() *http.Cookie
 	ValidateRefreshToken(token string) (*JWTClaims, error)
+	GetExpiresIn() int64
 }
 
 type jwtImpl struct {
@@ -165,4 +166,9 @@ func (j *jwtImpl) ValidateRefreshToken(token string) (*JWTClaims, error) {
 		return nil, err
 	}
 	return parsedToken.Claims.(*JWTClaims), nil
+}
+
+// GetExpiresIn returns the number of seconds until the access token expires.
+func (j *jwtImpl) GetExpiresIn() int64 {
+	return int64(j.accessTokenTTL.Seconds())
 }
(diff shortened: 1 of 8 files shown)


Nexucis (approved):
awesome fix ! Thank you @celian-garcia 🙏

PR #4245, opened 2026-07-11
[BUGFIX] use provider HTTP client for OAuth device code token exchange

anxkhn:
The OAuth device code token exchange was sending its token request through
`http.DefaultClient` instead of the provider-scoped client the endpoint builds
for itself.

When an `oAuthEndpoint` is constructed, it builds an `*http.Client` via
`newHTTPClient(provider.HTTP)`, which carries the provider's configured TLS
(custom CA, client certificates, `InsecureSkipVerify`) and timeout, and stores it
on `e.httpClient`. Every other flow uses it:

- the authorization code flow, via `newQueryContext`
  (`context.WithValue(..., oauth2.HTTPClient, e.httpClient)`) passed to
  `e.conf.Exchange`;
- the device code *initiation* step (`DeviceAuth`), which also goes through
  `newQueryContext`;
- the OIDC device code flow, via `rp.WithHTTPClient(httpClient)`.

Only the OAuth device code *token retrieval* (`retrieveDeviceAccessToken`), which
is hand-rolled because the `oauth2` package only exposes the device poll as a
whole, sent the request with `http.DefaultClient`. That silently dropped the
configured TLS and timeout, so device code login failed against an OAuth provider
served behind a private CA (`x509: certificate signed by unknown authority`),
even though the authorization code flow against the same provider worked.

This changes that one call to use `e.httpClient`, making the device code token
exchange consistent with the other flows.

The `http` package is still used in the same file (request construction, cookies,
status codes), so no import changes.

### Test

Added `TestOAuthRetrieveDeviceAccessToken_UsesProviderHTTPClient` next to the
existing auth tests. It stands up an `httptest.NewTLSServer` (self-signed cert)
and assigns `srv.Client()` (which trusts that cert) to `e.httpClient`. The
exchange succeeds only when the code uses `e.httpClient`; reverting the fix to
`http.DefaultClient` makes the test fail with the private-CA `x509` error. It also
asserts the server received
`grant_type=urn:ietf:params:oauth:grant-type:device_code` and the device code, so
the request is proven to flow through the intended client.

# Checklist

(No UI changes; the UI section of the template does not apply.)

### DIFF
--- a/internal/api/impl/auth/auth_test.go
+++ b/internal/api/impl/auth/auth_test.go
@@ -14,12 +14,16 @@
 package auth
 
 import (
+	"context"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"testing"
 
 	"github.com/perses/perses/internal/api/utils"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/oauth2"
 )
 
 func TestGetRedirectURI_WithAPIPrefix(t *testing.T) {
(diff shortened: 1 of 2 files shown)


celian-garcia (approved):
Valid as well :)

PR #4249, opened 2026-07-11
[BUGFIX] honor datasource TLS config when fetching OAuth token

anxkhn:
When an HTTP datasource proxy is configured with an OAuth 2.0 client-credentials
secret, `httpProxy.getToken` (`internal/api/impl/proxy/proxy.go`) builds an
`*http.Transport` from the datasource secret's TLS config and stores it on an
HTTP client under the `oauth2.HTTPClient` context key. It stored the client as a
value (`http.Client{...}`), but `golang.org/x/oauth2` retrieves it with a
`ctx.Value(oauth2.HTTPClient).(*http.Client)` type assertion. That assertion
fails for a value, so oauth2 silently falls back to `http.DefaultClient`.

The effect: the datasource secret's TLS settings (custom CA, client certificate
for mTLS, `InsecureSkipVerify`) and the transport's dial/handshake timeouts are
ignored specifically when fetching the OAuth token. A datasource whose token
endpoint uses a private CA or requires a client certificate would fail the token
handshake (or bypass the intended verification), even though the same TLS config
is correctly applied to the proxied data requests themselves.

The two other places in the codebase that put a client under this context key
already pass a pointer (`pkg/client/config/config.go` and
`internal/api/impl/auth/oauth.go`, the latter referencing golang/oauth2#187,
which is the upstream note that TLS-via-context needs a `*http.Client`). This
change makes the proxy path consistent with them.

Fix: pass `&http.Client{...}` so the configured transport is carried through the
context and used for the token request.

A regression test (`TestHTTPProxy_getToken_honorsTLSConfig`) serves the token
endpoint over TLS with a self-signed certificate wired in as the secret's CA. It
only passes when the transport built from the secret's TLS config is honored;
with the previous value-typed client it fails with
`x509: certificate signed by unknown authority` (oauth2 having fallen back to
`http.DefaultClient`).

# Screenshots

# Checklist

## UI Changes

- [ ] N/A - no UI changes (Go backend only).

### DIFF
--- a/internal/api/impl/proxy/proxy.go
+++ b/internal/api/impl/proxy/proxy.go
@@ -402,7 +402,7 @@ func (h *httpProxy) getToken(ctx context.Context, oauth *secretModel.OAuth) (*oa
 		return nil, err
 	}
 
-	httpClient := http.Client{
+	httpClient := &http.Client{
 		Transport: transport,
 	}
 
--- a/internal/api/impl/proxy/proxy_test.go
+++ b/internal/api/impl/proxy/proxy_test.go
@@ -14,9 +14,15 @@
 package proxy
 
 import (
+	"context"
 	"crypto/tls"
+	"encoding/pem"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
+	v1 "github.com/perses/perses/pkg/model/api/v1"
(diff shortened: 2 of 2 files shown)


celian-garcia (approved):
thanks !

PR #4258, opened 2026-07-14
[BUGFIX] api: prevent path traversal in the file database backend

anxkhn:
The file database backend (`internal/api/database/file`) is the default database
for the Perses API (`database.file`, `defaultFileDBFolder = "./local_db"`). Its DAO
builds a filesystem path by joining the configured data folder with a key derived
from the resource `name` (and `project`) that comes from the request URL. Those
values are not validated before they reach the DAO on the read and delete paths, so
a resource name containing `..` segments resolves to a path outside the data folder.

Concretely, a request such as `GET /api/v1/globaldatasources/../../<file>` (or the
`DELETE` equivalent) reaches `DAO.Get` / `DAO.Delete`, where the name flows through
`generateID` into `buildPath` and then into `os.ReadFile` / `os.Remove`. Because
`filepath.Join` cleans the `..` segments, the resolved path escapes the data folder,
allowing arbitrary `.json` / `.yaml` files on the host to be read or deleted. When
authentication is disabled (the default per `SECURITY.md`), this is reachable
unauthenticated.

This change adds a containment check in `buildPath`: after the path is joined, it is
rejected if it does not resolve inside `d.Folder`, returning a not-found error
instead of touching the filesystem. `buildPath` is the single chokepoint for
`Create`, `Get`, `Delete` and `upsert`, so one guard covers read, write and delete.
The SQL backend and the shared request toolbox are left untouched.

Two regression tests are added (`internal/api/database/file/file_test.go`):
`TestDAO_GetPathTraversal` and `TestDAO_DeletePathTraversal`. Each writes a sentinel
file outside the data folder, calls `Get` / `Delete` with a `../../secret` name, and
asserts the call returns key-not-found and that the outside file is neither read nor
removed. Both fail without the guard and pass with it.

Scope note: this targets the single-resource read/delete path, which is the proven
vector. The list/query path (`Query` / `DeleteByQuery`) is not modified here.

# Screenshots

N/A (backend-only change; no UI impact).

# Checklist

  convention (`BUGFIX`).

## UI Changes

- [x] N/A - no UI changes.

### DIFF
--- a/internal/api/database/file/file.go
+++ b/internal/api/database/file/file.go
@@ -68,7 +68,10 @@ func (d *DAO) Create(entity modelAPI.Entity) error {
 	if generateIDErr != nil {
 		return generateIDErr
 	}
-	filePath := d.buildPath(key)
+	filePath, buildPathErr := d.buildPath(key)
+	if buildPathErr != nil {
+		return buildPathErr
+	}
 	if _, err := os.Stat(filePath); err == nil {
 		// The file exists, so we should return a conflict error.
 		return &databaseModel.Error{Key: key, Code: databaseModel.ErrorCodeConflict}
@@ -89,7 +92,10 @@ func (d *DAO) Get(kind modelV1.Kind, metadata modelAPI.Metadata, entity modelAPI
 	if generateIDErr != nil {
 		return generateIDErr
 	}
-	filePath := d.buildPath(key)
+	filePath, buildPathErr := d.buildPath(key)
+	if buildPathErr != nil {
+		return buildPathErr
+	}
 	data, err := os.ReadFile(filePath) //nolint: gosec
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -239,7 +245,10 @@ func (d *DAO) Delete(kind modelV1.Kind, metadata modelAPI.Metadata) error {
 	if generateIDErr != nil {
 		return generateIDErr
 	}
-	filePath := d.buildPath(key)
+	filePath, buildPathErr := d.buildPath(key)
+	if buildPathErr != nil {
+		return buildPathErr
+	}
 	err := os.Remove(filePath)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -286,7 +295,10 @@ func (d *DAO) GetLatestUpdateTime(_ []modelV1.Kind) (*string, error) {
 }
 
 func (d *DAO) upsert(key string, entity modelAPI.Entity) error {
-	filePath := d.buildPath(key)
+	filePath, buildPathErr := d.buildPath(key)
+	if buildPathErr != nil {
+		return buildPathErr
+	}
 	if err := os.MkdirAll(filepath.Dir(filePath), 0750); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


celian-garcia (commented):
Hello, this is addressed in another PR by @Nexucis I let him compare with what he's doing

s3onghyun (commented):
Nice fix — the `filepath.Rel` containment check is the right approach, and I like that the test puts the sensitive file *outside* the data folder with a matching name so a successful read couldn't be masked by the case-sensitivity handling. Threading the error through `Create`/`Get`/`Delete`/`upsert` all looks correct.

One question on completeness: this protects the by-key paths that go through `buildPath`, but the query paths don't go through it. `RawQuery`/`Query`/`DeleteByQuery` build their folder via `buildQuery` → `generateProjectResourceQuery` (`query.go:54`):

```go
return filepath.Join(d.Folder, v1.PluralKindMap[kind], project)
```

`project` (`param:"project" query:"project"`) and `NamePrefix` (`query:"name"`) are the same kind of user-controlled values, and they aren't run through the new containment check before reaching:

- `os.RemoveAll(folder)` in `DeleteByQuery` (`file.go:271`)
- `os.ReadFile(file)` in `RawQuery`/`Query` (`file.go:145`, `:214`) via `visit(folder, prefix)`

So a `project` containing `..` segments looks like it could still resolve outside `d.Folder` on those paths — and the `os.RemoveAll` one is arguably more impactful than a read, since it's a recursive delete.

Am I missing an upstream validation that already rejects `..` in `project`/`name` for the list/delete-by-query endpoints? If not, it might be worth factoring the containment check into a small helper (e.g. `ensureWithinFolder(path string) error`) and applying it in `buildQuery` as well, so both entry points share one guard. Happy to help with a follow-up if useful.

Nexucis:
I am unable today to reproduce the case you are describing @anxkhn. Do you mind to provide a e2e test case scenario that proves the attack you are describing is feasible ?

@s3onghyun I guess this is what you are looking for: https://github.com/perses/perses/blob/main/internal/api/toolbox/list.go#L47-L58

s3onghyun (commented):
Thanks for the pointer @Nexucis — I dug into it and I think your skepticism is right: I can't see an end-to-end path either.

The reason is the metadata name/project validation (`^[a-zA-Z0-9_.-]+$`). It allows `.` but rejects `/`. Since `generateID` builds the key with `filepath.Join(kind, project, name)` and `buildPath` then appends the extension, without a `/` you can't form a `../..` sequence: a lone `..` field just collapses against the `kind` segment and stays inside the data folder (or ends up as part of the filename). To actually climb out of the folder you'd need `/` separators, which validation refuses.

So the traversal only "works" at the DAO unit level — @anxkhn's test constructs `NewMetadata("../../secret")` and calls the DAO directly, which bypasses the API-layer name validation that a real request goes through. That matches "unable to reproduce": through the HTTP API the name/project never contains `/`, so it never reaches the DAO in a dangerous shape.

Net: the DAO-level check (and my `buildQuery` note) are defense-in-depth at best, not a live vuln — assuming name validation always runs before the DAO on the read/delete/list paths, which is the thing an e2e test would actually pin down. If it's useful I'm happy to write that e2e case: a request with `..`/`%2f` in the `project`/`name` path param, asserting it's rejected (400) before touching the filesystem. That would either close this cleanly or surface a real gap if some path skips validation. Want me to put that together?

anxkhn:
closing this because the api validation rejects path separators before the dao, so the dao-only test does not demonstrate a reachable traversal. thanks for checking this.

PR #3583, opened 2025-11-12
[BUGFIX] Add bits/sec convertation to human-readable sizes, add bits units(#3441)

lomobot:
# Description

This PR resolve problem #3441 with incorrect convertation of bits unit. 

In this PR were added (or updated):
- units bits, decbits, bit/sec and decbits/sec
- migration for these types
- tests for new units
- info about new type to docs/plugins/common.md 

# Screenshots

image

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).
- [x] E2E tests are stable and unlikely to be flaky.
  See [e2e](https://github.com/perses/perses/tree/main/ui/e2e#visual-tests) docs for more details. Common issues include:
  - Is the data inconsistent? You need to mock API requests.
  - Does the time change? You need to use consistent time values or mock time utilities.
  - Does it have loading states? You need to wait for loading to complete.

PR #4259, opened 2026-07-14
[BUGFIX] accept bits/decbits/sek/sgd/usd units in go-sdk Format validation

anxkhn:
---

`go-sdk` `common.Format.validate()` rejected five of the package's own exported
unit constants with an `unknown format` error. The allow-list `switch` in
`validate()` was not kept in sync when two features added new units:

- `BinaryBitsUnit` (`bits`) and `DecimalBitsUnit` (`decbits`) were declared in
  #3583 but only `BitsDecPerSecondsUnit` got wired into the switch; the two
  base bits units were left out.
- The currency cases added in #3132 stop at `NewZealandDollarUnit`, so
  `SwedishKronaDollarUnit` (`sek`), `SingaporeDollarUnit` (`sgd`) and
  `USDollarUnit` (`usd`) fall through to the `default` branch.

Because `validate()` runs from `UnmarshalJSON`/`UnmarshalYAML`, any
dashboard-as-code `Format` built with one of these otherwise valid units fails
to unmarshal, even though the same package publicly exports the constant as a
legal value.

All five are legitimate Perses units with full formatting support in the
frontend (`ui/core/src/model/units/bits.ts` defines `BitsUnit = 'bits' |
'decbits'`; `ui/core/src/model/units/currency.ts` lists `sek`, `sgd`, `usd`
with config entries), which confirms the validator was wrong to reject them,
not the constants.

This adds the five missing constants to the switch (no other behavior change)
and adds a table test in a new `go-sdk/common/format_test.go`. One of the tests
iterates every exported unit constant so a newly added unit that is forgotten
in `validate()` is caught in the future.

# Screenshots

N/A (Go-only change, no UI impact).

# Checklist

  convention (`BUGFIX`).

## UI Changes

N/A - no UI changes.

---

## Diff summary

- `go-sdk/common/format.go` (+2/-2 on the switch): add
  `string(BinaryBitsUnit), string(DecimalBitsUnit)` next to the byte units and
  `string(SwedishKronaDollarUnit), string(SingaporeDollarUnit),
  string(USDollarUnit)` after `NewZealandDollarUnit`.
- `go-sdk/common/format_test.go` (new, +83): `TestFormatUnmarshalJSONValidUnits`
  (every exported unit round-trips), `TestFormatUnmarshalJSON` (the 5 regression
  units, a no-unit case, and a negative `not-a-unit` case).

### DIFF
--- a/go-sdk/common/format.go
+++ b/go-sdk/common/format.go
@@ -114,14 +114,14 @@ func (f *Format) validate() error {
 	switch *f.Unit {
 	case string(NanoSecondsUnit), string(MicroSecondsUnit), string(MilliSecondsUnit), string(SecondsUnit), string(MinutesUnit),
 		string(HoursUnit), string(DaysUnit), string(WeeksUnit), string(MonthsUnit),
-		string(YearsUnit), string(PercentUnit), string(PercentDecimalUnit), DecimalUnit, string(BinaryBytesUnit), string(DecimalBytesUnit),
+		string(YearsUnit), string(PercentUnit), string(PercentDecimalUnit), DecimalUnit, string(BinaryBitsUnit), string(DecimalBitsUnit), string(BinaryBytesUnit), string(DecimalBytesUnit),
 		string(BitsPerSecondsUnit), string(BitsDecPerSecondsUnit), string(BytesPerSecondsUnit), string(BytesDecPerSecondsUnit), string(CountsPerSecondsUnit), string(EventsPerSecondsUnit),
 		string(MessagesPerSecondsUnit), string(OpsPerSecondsUnit), string(PacketsPerSecondsUnit),
 		string(ReadsPerSecondsUnit), string(RecordsPerSecondsUnit), string(RequestsPerSecondsUnit),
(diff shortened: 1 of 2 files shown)


Nexucis (approved):

s3onghyun (commented):
Nice catch bringing these in line with the frontend. While cross-checking against the UI unit registry (`ui/core/src/model/units/units.ts:61-62`), I noticed two more groups it exposes that `Format.validate` will still reject as `unknown format` even after this PR:

- **Temperature** — `celsius`, `fahrenheit` (`ui/core/src/model/units/temperature.ts:21`)
- **Date** — `datetime-iso`, `date-iso`, `relative-time`, `unix-timestamp`, … (`ui/core/src/model/units/date.ts`)

There are no matching constants in `go-sdk/common/format.go` (the switch has no temperature/date cases), so a dashboard authored in the UI with one of those units would fail go-sdk validation the same way `bits`/`sek` did before this fix. Since cue doesn't enumerate units, `format.go` and the TS registry are the two sources of truth, and they'd still be out of sync for these two groups.

Would it make sense to fold them into this PR, or track them as a follow-up? Happy to pick up the follow-up if useful.

Nexucis:
ah yes indeed, if you want to open a PR to fix this issue I would be happy to review it :)

PR #4260, opened 2026-07-14
[BUGFIX] plugin: skip migrate.cue whose package is not migrate

anxkhn:
# Description

While loading plugin schemas, `internal/api/plugin/migrate/migrate.go` walks the
plugin tree and, for each `migrate` folder, checks that `migrate.cue` belongs to
the `migrate` package before loading it. Folders whose `migrate.cue` declares a
different package are meant to be skipped. That skip is currently unreachable
because the check is nested inside the error branch of the same `if`:

```go
if isMigrate, openFileErr := isPackageMigrate(migrateFilePath); openFileErr != nil {
    if openFileErr != nil {
        return openFileErr
    }
    if !isMigrate {
        return fs.SkipDir
    }
}
```

The outer condition only enters when `openFileErr != nil`, so `if !isMigrate {
return fs.SkipDir }` can never run. When `isPackageMigrate` returns `(false, nil)`
(a readable `migrate.cue` that simply declares another package), the folder is not
skipped and instead reaches `LoadMigrateSchema`, which fails with:

```
build constraints exclude all CUE files ...: package is <other>, want migrate
```

instead of being silently ignored as intended.

This restructures the guard so the open error is returned and the package
mismatch is otherwise handled via `else if`:

```go
if isMigrate, openFileErr := isPackageMigrate(migrateFilePath); openFileErr != nil {
    return openFileErr
} else if !isMigrate {
    return fs.SkipDir
}
```

A regression test (`TestLoadSkipsMigrateFolderWithWrongPackage`) loads a `migrate`
folder containing a `migrate.cue` with a mismatching package and asserts it is
skipped rather than producing an error.

The model schema loader (`internal/api/plugin/schema/schema.go`) had the identical
dead-guard shape; it was already restructured to the reachable form as part of the
recent "Exposing complete dashboard and plugin CUE schema endpoint" change (#4031),
but without a test covering the skip path. This adds
`TestSch_load_SkipsCueFileWithWrongPackage` there too, so the skip behavior stays
covered against a future regression.

No behavior changes for well-formed plugins: valid `migrate`/`model` packages load
exactly as before; only the intended skip of mismatching-package `.cue` files is
restored (migrate) and locked in with tests (both).

# Checklist

### DIFF
--- a/internal/api/plugin/migrate/migrate.go
+++ b/internal/api/plugin/migrate/migrate.go
@@ -79,12 +79,9 @@ func Load(pluginPath string, moduleSpec v1.ModuleSpec) ([]schema.LoadSchema, err
 		migrateFilePath := filepath.Join(currentPath, "migrate.cue")
 		// We are verifying if the package is a package migrate. Otherwise, we won't be able to use it.
 		if isMigrate, openFileErr := isPackageMigrate(migrateFilePath); openFileErr != nil {
-			if openFileErr != nil {
-				return openFileErr
-			}
-			if !isMigrate {
-				return fs.SkipDir
-			}
+			return openFileErr
+		} else if !isMigrate {
+			return fs.SkipDir
 		}
 
 		instance, schemaErr := LoadMigrateSchema(currentPath)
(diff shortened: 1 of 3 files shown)


Nexucis (approved):
thanks for the improvement and nice catch again !

PR #4266, opened 2026-07-16
[BUGFIX] accept temperature and date units in go-sdk Format validation

s3onghyun:
### Description

Follow-up to #4259 (thanks @Nexucis for the go-ahead).

The frontend unit registry (`ui/core/src/model/units/units.ts`) exposes two groups that `go-sdk`'s `Format.validate` did not accept:

- **Temperature** — `celsius`, `fahrenheit` (`ui/core/src/model/units/temperature.ts`)
- **Date** — `datetime-iso`, `datetime-us`, `datetime-local`, `date-iso`, `date-us`, `date-local`, `time-local`, `time-iso`, `time-us`, `relative-time`, `unix-timestamp`, `unix-timestamp-ms` (`ui/core/src/model/units/date.ts`)

So a dashboard authored in the UI with one of those units was rejected as `unknown format` by the Go SDK — the same class of issue #4259 fixed for bits/currency.

This adds the `TemperatureUnit` and `DateUnit` constants (mirroring the frontend strings exactly) and accepts them in `validate`. Since cue doesn't enumerate units, `format.go` and the TS registry are the two sources of truth; this brings them back in sync.

### Testing

- The existing `validUnits` guard test now covers the 14 new constants (each is round-tripped through `Format` unmarshalling).
- Added regression cases for `celsius`, `fahrenheit`, `datetime-iso`, `unix-timestamp`.
- `go test ./go-sdk/common/`, `go vet`, `gofmt` all clean.

### DIFF
--- a/go-sdk/common/format.go
+++ b/go-sdk/common/format.go
@@ -19,60 +19,76 @@ import (
 )
 
 type (
-	TimeUnit       string
-	PercentageUnit string
-	ThroughputUnit string
-	CurrencyUnit   string
-	BytesUnit      string
+	TimeUnit        string
+	PercentageUnit  string
+	ThroughputUnit  string
+	CurrencyUnit    string
+	BytesUnit       string
+	TemperatureUnit string
+	DateUnit        string
 )
 
 const (
-	NanoSecondsUnit        TimeUnit       = "nanoseconds"
-	MicroSecondsUnit       TimeUnit       = "microseconds"
-	MilliSecondsUnit       TimeUnit       = "milliseconds"
-	SecondsUnit            TimeUnit       = "seconds"
-	MinutesUnit            TimeUnit       = "minutes"
-	HoursUnit              TimeUnit       = "hours"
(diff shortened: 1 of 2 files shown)


Nexucis:
Actually in cue we also have the various format available here: https://github.com/perses/shared/blob/main/cue/common/format.cue

But I think it is up to date comparing to go

Nexucis (approved):
overall looks good to me. Thank you for taking care about this part @s3onghyun !

s3onghyun:
Thanks for merging! And you're right — I was wrong in the description when I said cue doesn't enumerate units; it does, in `cue/common/format.cue`. I just checked it against this change:

- `#temperatureFormat: "celsius" | "fahrenheit"` ✓
- `#dateFormat: "datetime-iso" | "datetime-us" | ... | "unix-timestamp-ms"` ✓ (all 12)

So cue and the frontend registry were already in sync on these — go-sdk was the one lagging behind, which this PR brings back in line. All three sources match now. Nothing further needed on the cue side.

PR #4268, opened 2026-07-16
[FEATURE] Add server defaults for user preferences

### LINKED ISSUE #4265: [FEATURE] Support server-side default configuration for user preferences (timezone, rows per page, theme)
### Description

Currently, user preferences (timezone, rows per page, theme) are per-user only. There is no way for an admin to configure server-side defaults that apply to all users who haven't set their own preferences.

### Use case

In self-hosted deployments, it is common to want all users to default to UTC timezone, a specific rows-per-page value, or a consistent theme — without requiring each user to configure this manually. This is especially important for ephemeral deployments where the user database is reset on restart (e.g. file-based DB with emptyDir volumes in Kubernetes), causing user preferences to be lost on every restart.

### Proposed solution

Add a `frontend.default_user_preferences` section to the server config:

```yaml
frontend:
  default_user_preferences:
    timezone: UTC
    theme: dark
```

These values would be used as the default when a user has not yet set their own preference, following the existing precedence: `dashboard.spec.timezone > user preference > server default > local`.

### Related

- #4170 — add user preference (merged)
- #4184 — pass user preferences tz to ViewDashboard (merged)

PCRedHot:
## What changed

- add `frontend.default_user_preferences` for timezone, rows per page, and theme
- use server defaults only when the user has not selected that preference
- apply the configured page size across dashboard and administration tables
- validate and document the new configuration
- add backend and frontend regression tests

Closes #4265.

## Validation

- `npm --prefix ui/app run type-check`
- `npm --prefix ui/app run lint`
- `npm --prefix ui/app test -- --runInBand` (12 suites, 148 tests)
- `npm --prefix ui/app run build`
- `go test ./pkg/model/api/config` using Go 1.26.5
- `gofmt` and `git diff --check`


Nexucis:
@jgbernalp @Gladorme @shahrokni can you review this PR please ?

ui/app/src/context/UserPreferences.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const [storedPreferences, setUserPreferences] = useLocalStorage<UserPreferences | null>(USER_PREFERENCES_KEY, null);
+  const userPreferences = storedPreferences ?? props.defaultPreferences;
+  const contextValue = useMemo(() => ({ userPreferences, setUserPreferences }), [userPreferences, setUserPreferences]);
+
+  return <UserPreferencesContext.Provider value={contextValue}>{props.children}</UserPreferencesContext.Provider>;
+}
+
+export function useUserPreferences(): UserPreferencesContextType {
+  const ctx = useContext(UserPreferencesContext);
+  if (ctx === undefined) {
```
jgbernalp: Not sure if this provider should be mandatory. In the case there is no config provider in the current component tree, we should probably return an empty object. To match the current behavior when reading from local storage

ui/app/src/context/DarkMode.tsx
```diff
@@ -40,8 +41,13 @@ export const DarkModeContext = createContext<DarkModeContext | undefined>(undefi
...
 export function DarkModeContextProvider(props: { children: React.ReactNode }): ReactElement {
   const browserPrefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
+  const { config } = useConfigContext();
+  const defaultDarkMode = config.frontend.default_user_preferences?.theme
+    ? config.frontend.default_user_preferences.theme === 'dark'
+    : browserPrefersDarkMode;
 
-  const [isDarkModeEnabled, setDarkMode] = useLocalStorage<boolean>(DARK_MODE_PREFERENCE_KEY, browserPrefersDarkMode);
+  const [storedDarkMode, setDarkMode] = useLocalStorage<boolean | null>(DARK_MODE_PREFERENCE_KEY, null);
+  const isDarkModeEnabled = storedDarkMode ?? defaultDarkMode;
```
jgbernalp, line 49: current users will have either `true` or `false` stored in their local storage as the UI eagerly writes the preference. So this would never use the defaults as the user cannot "unset" the darkmode. This might not be needed as it would be better to align the darkmode with user or browser preferences for a11y, rather than deciding the a11y for users.

ui/app/src/context/UserPreferences.tsx
```diff
@@ -0,0 +1,44 @@
...
+const USER_PREFERENCES_KEY = 'PERSES_USER_PREFERENCES';
+
+const UserPreferencesContext = createContext<UserPreferencesContextType | undefined>(undefined);
+
+export function UserPreferencesContextProvider(props: {
+  children: React.ReactNode;
+  defaultPreferences: UserPreferences;
+}): ReactElement {
+  const [storedPreferences, setUserPreferences] = useLocalStorage<UserPreferences | null>(USER_PREFERENCES_KEY, null);
+  const userPreferences = storedPreferences ?? props.defaultPreferences;
```
jgbernalp: same here, for timezone the UI writes the "local" value eagerly so not sure in which case the defaults would be selected.

pkg/model/api/config/frontend.go
```diff
@@ -61,6 +63,27 @@ type TimeRange struct {
 	Options                []common.DurationString `json:"options,omitempty" yaml:"options,omitempty"`
 }
 
+// DefaultUserPreferences contains the preferences used when the user has not
+// stored an explicit preference in their browser.
+type DefaultUserPreferences struct {
+	Timezone    string `json:"timezone,omitempty" yaml:"timezone,omitempty"`
```
jgbernalp: timezone is not being validated in the Verify

ui/app/src/context/UserPreferences.tsx
jgbernalp: also if we add a new field and only default that one, this will never be merged with the stored preferences.

pkg/model/api/config/frontend.go
```diff
@@ -61,6 +63,27 @@ type TimeRange struct {
...
+	RowsPerPage int    `json:"rows_per_page,omitempty" yaml:"rows_per_page,omitempty"`
+	Theme       string `json:"theme,omitempty" yaml:"theme,omitempty"`
```
Nexucis: since `Theme` is an enum, I think this should be reflected by created a dedicated type.

```go
type FrontendTheme string

const (
    DarkTheme FrontendTheme = "dark"
    LightTheme FrontendTheme = "light"
)
```

pkg/model/api/config/frontend.go
```diff
@@ -61,6 +63,27 @@ type TimeRange struct {
...
+}
+
+func (p *DefaultUserPreferences) Verify() error {
+	if p.RowsPerPage < 0 {
+		return fmt.Errorf("frontend.default_user_preferences.rows_per_page cannot be negative")
+	}
+	if p.RowsPerPage != 0 && !slices.Contains(allowedRowsPerPage, p.RowsPerPage) {
+		return fmt.Errorf("frontend.default_user_preferences.rows_per_page must be one of: 10, 25, 50, 100")
+	}
+	if p.Theme != "" && p.Theme != "light" && p.Theme != "dark" {
```
Nexucis: in case theme is empty, you can set one by default

Nexucis:
Hello @PCRedHot,

any chance you can continue to work on this PR ?

PCRedHot:
@Nexucis Yes I am working on it, but I am current on vacation, will back working on it later today if possible.

pkg/model/api/config/frontend.go
PCRedHot: Done. I added the `FrontendTheme` type with `DarkTheme` and `LightTheme` constants, and updated `DefaultUserPreferences`.

pkg/model/api/config/frontend.go
PCRedHot: Done. `DefaultUserPreferences.Verify` now sets an empty theme to `LightTheme`. I also added a test covering this defaulting behavior.

pkg/model/api/config/frontend.go
PCRedHot: Added timezone validation in `DefaultUserPreferences.Verify` using `time.LoadLocation`, while allowing the local UI sentinel.

ui/app/src/context/UserPreferences.tsx
PCRedHot: Addressed both points. Preference storage now starts as `null` instead of eagerly storing local, and resolved preferences merge the server defaults with stored fields individually. Existing stored values remain explicit overrides, and the reset action clears them.

ui/app/src/context/DarkMode.tsx
PCRedHot: Good point. I changed the storage initialization so new users no longer eagerly persist the browser-derived value. The precedence is now: stored user choice > configured server default > browser preference when no server default exists.

ui/app/src/context/UserPreferences.tsx
PCRedHot: I made the provider optional by giving the context an empty default value. When no provider is present, `useUserPreferences()` now returns empty preferences with a no-op setter, matching the previous local-storage behavior. I also added a test for this case.

Nexucis:
oh no worries @PCRedHot I was just going through the PRs and asking the same questions.

Take your time and enjoy your holidays !

Nexucis (commented):
lgtm for the config side.
Over to @jgbernalp for the front if it is good for him

pkg/model/api/config/frontend.go
```diff
@@ -61,6 +72,35 @@ type TimeRange struct {
...
+	Theme       FrontendTheme `json:"theme,omitempty" yaml:"theme,omitempty"`
+}
+
+func (p *DefaultUserPreferences) Verify() error {
+	if len(p.Timezone) > 0 && p.Timezone != "local" {
+		if _, err := time.LoadLocation(p.Timezone); err != nil {
+			return fmt.Errorf("invalid frontend.default_user_preferences.timezone %q: %w", p.Timezone, err)
+		}
+	}
+	if p.RowsPerPage < 0 {
```
Nexucis: well instead of checking that, you can define a `uint8` type instead. So the number will always be positive

if it is equal to 0 you can set a default value at 25

PCRedHot: Thanks, updated. `RowsPerPage` now uses a `uint8`-based type, and 0 defaults to 25. I also aligned the frontend fallback and tests.

pkg/model/api/config/frontend.go
```diff
@@ -76,7 +80,7 @@ type TimeRange struct {
 // stored an explicit preference in their browser.
 type DefaultUserPreferences struct {
 	Timezone    string        `json:"timezone,omitempty" yaml:"timezone,omitempty"`
-	RowsPerPage int           `json:"rows_per_page,omitempty" yaml:"rows_per_page,omitempty"`
+	RowsPerPage RowsPerPage   `json:"rows_per_page,omitempty" yaml:"rows_per_page,omitempty"`
```
Nexucis: ah sorry I miss-leading you, I was more thinking at that:

```suggestion
	RowsPerPage uint8   `json:"rows_per_page,omitempty" 
```

creating the type does not provide anything useful as there is no function related to this custom type

Nexucis (approved):
I am approving this PR. Code looks good to me.

I will wait one or two days before merging it in case @jgbernalp or @Gladorme would like to say something on this PR

Nexucis:
No news so far so I am merging this one.

Thank you for this PR @PCRedHot !

Nexucis:
@PCRedHot aarg, I am sorry we have merged a PR that changed the linter. Do you mind to update your PR and solve the conflict please ?

Once this is done I will merge this PR

PCRedHot:
@Nexucis Done ;)

Nexucis:
awesome thank you !

PR #4269, opened 2026-07-17
Fix SecretSpec validation of mutually exclusive auth methods

chuenchen309:
`SecretSpec.validate()` guards the three auth methods with `&&`, so it only rejects a secret that sets **all three** at once. Every pair passes:

```go
if s.BasicAuth != nil && s.Authorization != nil && s.OAuth != nil {
    return fmt.Errorf("basicAuth, authorization and oauth are mutually exclusive, use one of them")
}
```

| payload | expected | actual on main |
|---|---|---|
| only basicAuth / only authorization / only oauth | accepted | accepted |
| all three | rejected | rejected |
| **basicAuth + oauth** | rejected | **accepted** |
| **basicAuth + authorization** | rejected | **accepted** |
| **authorization + oauth** | rejected | **accepted** |

That contradicts the error message the guard itself raises, and `docs/api/secret.md` ("Basic Auth, Authorization and OAuth are mutually exclusive").

**It matters downstream.** `httpProxy.setupAuthentication` (`internal/api/impl/proxy/proxy.go:365`) applies all three in sequence to the same `Authorization` header — `SetBasicAuth` first, then `Header.Set` for authorization, then `Header.Set` for oauth. Last one wins. So a secret with `basicAuth` + `oauth` is accepted today and then sends only the OAuth bearer token; the basicAuth credentials never reach the datasource, silently. This validation is what should stop that combination from existing.

The fix counts the configured methods, which is how `RestConfigClient.Validate` already does the same job in `pkg/client/config/config.go:85-103`.

Added `pkg/model/api/v1/secret_test.go` — the file had no test coverage, which is presumably how this survived. It covers all seven combinations; the three pair cases fail on `main` and pass here, and the four existing-behaviour cases pass on both.

`go test ./pkg/...` is green, `gofmt` and `go vet` clean.

Disclosure: written with Claude Code. I verified the reproduction, the downstream header behaviour, and the test red/green myself.

### DIFF
--- a/pkg/model/api/v1/secret_test.go
+++ b/pkg/model/api/v1/secret_test.go
@@ -1,0 +1,82 @@
+// Copyright The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	secretBasicAuth     = `"basicAuth": {"username": "Magical", "password": "Girl"}`
+	secretAuthorization = `"authorization": {"type": "Bearer", "credentials": "token"}`
+	secretOAuth         = `"oauth": {"clientID": "id", "clientSecret": "secret", "tokenURL": "https://example.com/token"}`
+)
+
+func TestUnmarshalJSONSecretSpecMutuallyExclusiveAuth(t *testing.T) {
+	mutuallyExclusiveErr := fmt.Errorf("basicAuth, authorization and oauth are mutually exclusive, use one of them")
+
+	testSuite := []struct {
+		title string
+		jason string
+		err   error
+	}{
+		{
+			title: "only basicAuth",
+			jason: fmt.Sprintf(`{%s}`, secretBasicAuth),
+		},
+		{
+			title: "only authorization",
+			jason: fmt.Sprintf(`{%s}`, secretAuthorization),
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/model/api/v1/secret_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	secretBasicAuth     = `"basicAuth": {"username": "Magical", "password": "Girl"}`
```
Nexucis:
```suggestion
	secretBasicAuth     = `"basicAuth": {"username": "Magical", "password": "Girl"}` // nolint:gosec
```
that will ignore the linter since it is a false positive

Nexucis:
@chuenchen309 thank you for the fix. Can you add the comments I have suggested to solve the linter issues ?

Once it is done we should be able to merge this PR

chuenchen309:
Sorry for the long delay here, and thanks for the review.

The gosec finding is fixed in 502d10b — but I didn't apply your suggestion verbatim, so please say the word if you'd rather have it your way and I'll switch.

Instead of suppressing G101 I renamed the three constants: `secretBasicAuth`/`secretAuthorization`/`secretOAuth` → `basicAuthJSON`/`authorizationJSON`/`oauthJSON`. They're JSON fixtures rather than credentials, so the new names describe them better and the warning goes away on its own. Verified with the v2.12.1 the CI pins, on the two commits:

```
$ golangci-lint run ./pkg/model/api/v1/...     # cdc7f9c, before
pkg/model/api/v1/secret_test.go:25:2: G101: Potential hardcoded credentials (gosec)
pkg/model/api/v1/secret_test.go:27:2: G101: Potential hardcoded credentials (gosec)
2 issues:
* gosec: 2

$ golangci-lint run ./pkg/model/api/v1/...     # 502d10b, after
0 issues.
```

I did see the existing `//nolint:gosec` in `internal/api/crypto/jwt.go` and elsewhere, so suppression is clearly idiomatic here — I only preferred the rename because the warning was about the names rather than anything real in the values.

One thing that needs you rather than me: all five workflows on this PR are sitting at `action_required` (first-time contributor), so `lint` hasn't re-run against the fix — the green you're waiting for can't show up until someone approves the runs.

*AI-assisted; the two lint runs above are my own, on the pinned v2.12.1.*

Nexucis (approved):
thanks !

PR #4278, opened 2026-07-20
[ignore] add no reg migration test

### LINKED ISSUE #4272: Migration API fails with Cue schema validation error when panels contain targets
### Summary
The Perses migration API (`POST /api/migrate`) fails with a Cue schema validation error when processing Grafana dashboards that have panels containing `targets` field. The error occurs during the marshaling phase of the migrated dashboard.

### Environment
- **Perses Version**: 0.54.0-rc.0
- **API Endpoint**: `POST /api/migrate`
- **HTTP Response**: 200 OK with `{"message": "internal server error"}`
- **Server Log**: `time="..." level=error msg="unexpected error not handled" error="cue: marshal error: spec: error in call to len: incomplete argument list (type list)"`

### Steps to Reproduce

1. Create a Perses instance (version 0.54.0-rc.0)
2. Authenticate and obtain a bearer token
3. Send a migration request with a Grafana dashboard containing panels with targets:

```json
{
  "grafanaDashboard": {
    "title": "Test Dashboard",
    "panels": [
      {
        "id": 1,
        "title": "Panel 1",
        "type": "timeseries",
        "targets": [
          {
            "expr": "rate(metric[5m])"
          }
        ]
      }
    ]
  }
}
```

### Actual Behavior
The API returns HTTP 200 with an error message:
```json
{
  "message": "internal server error"
}
```

### Workaround
The migration API works successfully when:
- Panels do NOT contain `targets` field
- Only templating variables are included without panels containing targets

### Additional Context
- **Panels without targets**: ✅ 
- **Panels with targets**: ❌ 
- **Templating variables only**: ✅ 
- **Combined (templating + panels with targets)**: ❌

The error originates from Cue's marshal validation function, suggesting a schema definition issue rather than request data validation.

### Expected Behavior
The migration API should successfully migrate the Grafana dashboard and return a valid Perses Dashboard resource with:
- `kind`: "Dashboard"
- `metadata`: with name, createdAt, updatedAt, version
- `spec`: with migrated dashboard configuration including panels

### System information

Linux 6.8.0-124-generic x86_64

### Perses version

0.54.0-rc.0

### Perses configuration file

```yaml

```

### Logs

```text
unexpected error not handled: cue: marshal error: spec: error in call to len: incomplete argument list (type list)
```

Nexucis:
This PR is adding a no regression test on the migration using a real plugin (timeseriesChart).
As you can see the test is failing and it will be the case until we rollback the cuelang version to v0.16 (v0.17 currently)

It also includes the fix. So this PR can close #4272 

This PR needs to merged not squashed to preserve the two commits


celian-garcia (commented):
While very efficient, I find it a bit not future proof. We have already the plugins archive. Why not make a e2e test for migrate with a timeseries chart. ?

Nexucis:
> While very efficient, I find it a bit not future proof. We have already the plugins archive. Why not make a e2e test for migrate with a timeseries chart. ?

this is too avoid loading too much plugin and also to avoid to depend on a plugin that change overtime. Ideally I should take more time to find what syntax is causing the error in the cue migration file to reproduce and create a fake plugin containing this isolation.

Since at some point we need to report it to Cuelang to see the version fixed, this is something that will need to be done anyway. But for the moment it is quicker to just copy the plugin, so we can move on, patch and release quickly.

celian-garcia:
> this is too avoid loading too much plugin and also to avoid to depend on a plugin that change overtime. Ideally I should take more time to find what syntax is causing the error in the cue migration file to reproduce and create a fake plugin containing this isolation.

I see, so you created a minimal plugin, that appears to be a copy paste but that is not designed to follow the original one. It is also complex enough to trigger the problem. Alright that's good to me.

celian-garcia (approved):

PR #4279, opened 2026-07-21
Revert "[FEATURE] Exposing complete dashboard and plugin CUE schema endpoint"

Nexucis:
Reverts perses/perses#4031

see https://github.com/perses/perses/issues/4272 for more context


Gladorme (approved):

internal/cli/cmd/dac/build/testdata/go/go.mod
```diff
@@ -1,6 +1,6 @@
 module dac
 
-go 1.26.2
+go 1.26.5
```
celian-garcia, line 4: This commit was downgrading go version of tests ? This diff and all til the end looks to not be a revert

celian-garcia (approved):
Ok with an acknowledgement of the owners of this code

internal/cli/cmd/dac/build/testdata/go/go.mod
Nexucis: it is a mixup because the PR is merged into the branch where there is the non-reg migration test to ensure this PR is actually fixing the probleme

Nexucis: so it also includes the commit from Gabriel upgrading the golang version

Nexucis: I have removed the commit from Gabriel to avoid this mixup

PR #4288, opened 2026-07-22
[ignore] ignore upgrade cuelang.org/go in dependabot

Nexucis:


Nexucis:
moving forward to avoid an unfortunate merge with dependabot

Issue #3001, opened 2025-06-10
Graph annotations


AntoineThebaud:
### Proposal

The idea would be to be able to display certain data as annotations (= precise point / vertical line) on graphs instead of regular trends. For instance you could use that with the TimeSeriesChart on a pod memory graph to show when restarts / OOMKilled happened.

AntoineThebaud:
So we discussed this topic with @Gladorme and @Nexucis and because we believe any annotation coming from any datasource should be treated/displayed the same (it comes with a single timestamp or a range + metadata like title & description to be displayed in the popup on hover), we thus think we should define a new type of plugin named Annotation & each Datasource could then come with its own implem (e.g PrometheusAnnotation, LokiAnnotation...).

Besides, annotations could be defined at two levels
- dashboard
- panel

Dashboard annotations would get displayed on all panels that "subscribe" to them = all panels that are able to display annotations - maybe we can think about adding a on/off toggle for that at panel level that would be on by default?. About "panels that are able to display annotations", concretely this would mean we'd have a new hook to consume annotations and each individual panel plugin could call it & render the data. As said above at the beginning we'll start with TimeSeriesChart only but further plugins will be considered after.

nicolastakashi:
Hey @Gladorme @AntoineThebaud @Nexucis 👋

Now that the dashboard-level annotations are in, I'd like to pick up the panel-level half described in the two-level design above. Before I open any PRs I want to check the data-model shape and the toggle behavior with you.

For the model, I'd add an optional annotations block to `PanelSpec` (in `perses/spec`) and reuse the existing `AnnotationSpec` / `AnnotationDisplay` types for the panel-local definitions:

```go
type PanelSpec struct {
    Display     *PanelDisplay     `json:"display,omitempty" yaml:"display,omitempty"`
    Plugin      plugin.Plugin     `json:"plugin" yaml:"plugin"`
    Queries     []Query           `json:"queries,omitempty" yaml:"queries,omitempty"`
    Links       []Link            `json:"links,omitempty" yaml:"links,omitempty"`
    Annotations *PanelAnnotations `json:"annotations,omitempty" yaml:"annotations,omitempty"` // NEW
}

// PanelAnnotations controls annotation behavior for a single panel.
type PanelAnnotations struct {
    // nil == enabled (default on), matching "on by default" from the design.
    Enabled     *bool            `json:"enabled,omitempty" yaml:"enabled,omitempty"`
[...]

PR #4299, opened 2026-07-24
[FEATURE] Validate panel-level annotation definitions

nicolastakashi:
## What this does

This validates panel-local annotation definitions on the backend, the same way dashboard-level annotations are already validated.

`ValidatePanels` in `internal/api/plugin/schema/schema.go` now walks `panel.Spec.Annotations.Definitions` for each panel and runs every definition's plugin through the existing `ValidateAnnotations` path. No new code path, and no new plugin kind: `KindAnnotation` already exists and the definitions reuse `AnnotationSpec`, so a panel-local annotation goes through the same schema check as a dashboard-level one.

The `Enabled` toggle is not validated, since it only controls whether dashboard annotations show on the panel. A panel with no `annotations` block is skipped.

## Dependency

This is stacked on perses/spec#61, which adds the `PanelSpec.Annotations` field to the data model. It will not compile until that lands and the `github.com/perses/spec` version is bumped here. I developed it against a local `replace` pointing at the spec branch, so I left the go.mod bump out of this PR on purpose.

## Testing

There were no annotation fixtures under `internal/api/plugin/schema/testdata`, so I added a small annotation schema (`FirstAnnotation`) and two sample plugins, one valid and one with an unknown kind. `TestValidatePanelsWithAnnotations` covers four cases: a valid panel-local definition, an unknown annotation kind that gets rejected, a panel with only the toggle set, and a panel with no annotations block.

`go test ./internal/api/plugin/schema/...` passes. Two failures in `internal/api/validate` also happen on a clean checkout, because those tests need built plugin archives that are not present locally.

## Open questions

Same points as on perses/spec#61 and the design thread in #3001:

1. Is a single per-panel boolean enough for the toggle, or would you want per-annotation opt-out by name?
2. Should panel-local definitions be in the first iteration, or a follow-up after the toggle?

### DIFF
--- a/go-sdk/test/go.mod
+++ b/go-sdk/test/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/perses/plugins/staticlistvariable v0.9.0
 	github.com/perses/plugins/table v0.13.0
 	github.com/perses/plugins/timeserieschart v0.13.0
-	github.com/perses/spec v0.2.0
+	github.com/perses/spec v0.3.0-beta.0
 	github.com/stretchr/testify v1.11.1
 )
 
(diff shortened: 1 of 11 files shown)


Nexucis (approved):
perfect, thank you @nicolastakashi !

PR #4303, opened 2026-07-27
[BUGFIX] Downgrade zitadel/oidc to 3.45.4 to keep support with hydra

celian-garcia:
# Description

zitadel/oidc made some changes betwee 3.45.4 and 3.47.7 that act as a regression for hydra provider. 
https://github.com/perses/perses-operator/pull/431#discussion_r3659275823

I propose to downgrade back to a version that was working for hydra, before trying to upgrade it back later when the 3.47 will be more mature and tested.

# Screenshots

# Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/gjson v1.19.0
-	github.com/zitadel/oidc/v3 v3.47.7
+	github.com/zitadel/oidc/v3 v3.45.4
 	golang.org/x/crypto v0.54.0
 	golang.org/x/mod v0.38.0
 	golang.org/x/oauth2 v0.36.0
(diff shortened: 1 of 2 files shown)


Nexucis:
mmm is there an issue on OIDC side that is tracking the problem with hydra ? Otherwise we won't be able to upgrade anymore the packages 😟

celian-garcia:
> mmm is there an issue on OIDC side that is tracking the problem with hydra ? Otherwise we won't be able to upgrade anymore the packages 😟

Unfortunately not, we just discovered it and I am not knowledgeable enough on hydra. We need to test their next patch before creating an issue.

The thing is the time is too short on me to make that and I thought it could be better to revert the break before trying to fix.

celian-garcia:
If you want I can create an issue on perses side to follow up and test before creating the issue on zitadel.

Nexucis:
yes that will be nice to have something tracking this issue at least on our side

Nexucis (approved):

PR #4304, opened 2026-07-27
[IGNORE] Add upgrade guide about package re-structure

shahrokni:
Add package reconstruction instructions to the `upgrade guide`.
In a simple short paraph it explains that the core has been deprecated and the relevant files could be found in other packages.

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -84,6 +84,134 @@ max_version -> maxVersion
 
 ### Upgrading from v0.53.0 to v0.54.0
 
+#### Core package deprecated
+
+We have deprecated the `core package` and accordingly moved all its types and functionalities into other packages including `spec`, `dashboards`, `components`, `plugin-system`, and `client`. We are planing to drop the `core package` completely in the subsequent release. Therefore, we kindly ask all contributors to avoid importing `core package` members, and instead use the relevant types from the mentioned packages. Furthermore, should your PRs already import `core` members, you should replace them with the relevant types from the mentioned packages. This also means that meanwhile we cease developing `core`, because as it has been already mentioned the subsequent release will drop core completely.
+
+The dependencies from `core` had been used in all repositories and different packages including `perses/perses` (the Perses app), `perses/plugins`, and `perses/shared`. Please note that the `shared` is the host of the of the `components`, `plugin-system`, `dashboards`, and `client` packages. The `spec` as its name suggests exposes the specifications and had no dependency to `core`. However, some of the core members have been moved to `spec` already.
+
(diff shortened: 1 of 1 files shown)


docs/upgrade-guide.md
```diff
@@ -11,6 +11,10 @@ application, as a library, etc.). Therefore, the upgrade process may vary based
 
 ### Upgrading from v0.53.0 to v0.54.0
 
+#### Core package have been deprecated
+
+We have deprecated the `core package` and accordingly moved all its types and functionalities into other packages including `spec`, `dashboards`, `components`, `plugin-system`, and `client`. We are planing to drop the `core package` completely in the subsequent release. Therefore, we kindly ask all contributors to avoid importing `core package` members, and instead use the relevant types from the mentioned packages. Furthermore, should your PRs already import `core` members, you should replace them with the relevant types from the mentioned packages.
+
```
Nexucis: As you might notice, we have different section depending on the user profile.

Here you are in the section that is about how to upgrade the Perses application.

It can either go in the `Plugin Developer` section or in a new one `Embedded component`.
https://github.com/perses/perses/blob/main/docs/upgrade-guide.md#upgrading-from-v0530-to-v0540-1

Also it would be nice if you can provide different example because here we have no clue on what do to depending on the situation. So it would be nice to describe precisely what you need to do depending on the situation.

You can say if you import `Dashboard` for example, then this is in the spec and you provide a code snippet.

And provide examples/snippets for other cases as well.

docs/upgrade-guide.md
```diff
@@ -84,6 +84,45 @@ max_version -> maxVersion
...
+  FormatOptions,
+  formatValue,
+  Table,
+  TableCellConfigs,
+  TableColumnConfig,
+  transformData,
+  useSelection,
+} from '@perses-dev/components';
+```
+You as the contributor need to make sure that no `core` dependency is used in your changes. If it has been used already, it should be simply replaced with its equivalent from the mentioned packages. This should be no challenge as most of the IDEs suggest substitutes. For example, in VSCODE, if you can remove the `core` dependency, the IDE will suggest the substitutes. Simply select the suggested import and it will be added to your file automatically.
```
Nexucis, line 122: Here you are supposing people are using IDE which might not be the case.

So please provide more examples when you need the package spec, the client and so one. With what you have before and now what is look like. Like it has been done for the go-sdk for example.

You will see you have multiple examples that explains in details how the code changed.

Please can we have an detail and exhaustive documentation for people that are not as knowledgable as you ?

Nexucis (approved):
awesome, thank you !

PR #4332, opened 2026-08-03
[ENHANCEMENT]: Multi yaml file support for provisioning

### LINKED ISSUE #4341: Support multiple YAML documents in a single provisioning file
Currently, provisioning supports a single YAML object or a list of objects per file. 
It would be useful to also support multiple YAML documents separated by YAML doc markers `---` within the same provisioning file. 

@prakhar29jain

prakhar29jain:
Closes https://github.com/perses/perses/issues/4341

# Description
Support for multi yaml separated with dashlanes (---) .

image

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/internal/cli/file/file.go
+++ b/internal/cli/file/file.go
@@ -14,6 +14,7 @@
 package file
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/cli/file/file.go
```diff
@@ -189,17 +190,59 @@ func (u *unmarshaller) read() error {
...
-		if yamlErr := yaml.Unmarshal(data, &objects); yamlErr != nil {
-			if yamlErr = yaml.Unmarshal(data, &object); yamlErr != nil {
-				return newReadFileErr(yamlErr)
+		decodedObjects, decodeErr := decodeYAMLDocumentsAsObjects(data)
+		if decodeErr != nil {
+			if yamlErr := yaml.Unmarshal(data, &objects); yamlErr != nil {
+				if yamlErr = yaml.Unmarshal(data, &object); yamlErr != nil {
+					return newReadFileErr(yamlErr)
+				}
+				objects = append(objects, object)
```
shahrokni: ## Likely Logic Duplication
It seems `decodeYAMLDocumentsAsObjects` is already covering both following cases
- yaml.Unmarshal(data, &objects);
- yaml.Unmarshal(data, &object);

This means if your function fails, it is not likely that calling Unmarshal(s) give you any valid results. So those attempts are already doomed. 

## Why?
The first and second Unmarshal(s) are trying to fill objects and object respectively, each dealing with different kind of data structure. The first one is an slice of objects and the second one is a single object. This is exactly what your function is trying to do. `Case one` **tries** to find an slice of objects while `Case Two` **tries** to deal with an object. 

```go
// YOUR CODE
		switch doc := document.(type) {
       // Slice of any detection, hopefully objects
		case []any:
			for _, item := range doc {
				obj, ok := item.(map[string]any)
				if !ok {
					return nil, fmt.Errorf("yaml document list contains a non-object item")
				}
				objects = append(objects, obj)
			}
        // Single object detection
		case map[string]any:
			objects = append(objects, doc)
		default:
			return nil, fmt.Errorf("yaml document must be an object or list of objects")
		}
```
So the final conclusion would be if your function can not produce the desired results, then the unmarshal(s) will fail as well. 

## How to avoid this

```go
decodedObjects, decodeErr := decodeYAMLDocumentsAsObjects(data)
		if decodeErr != nil {
          // No need to call the unmarshal(s) just deal with the error
```

internal/cli/file/file.go
```diff
@@ -189,17 +190,59 @@ func (u *unmarshaller) read() error {
...
+				if !ok {
+					return nil, fmt.Errorf("yaml document list contains a non-object item")
+				}
+				objects = append(objects, obj)
+			}
+		case map[string]any:
+			objects = append(objects, doc)
+		default:
+			return nil, fmt.Errorf("yaml document must be an object or list of objects")
+		}
```
shahrokni, line 234: I think the error message is generalized, and could benefit from detailed information.
- What doc caused the issue
- What was the expected types (static message) and  what has been received. 

When dealing with a huge yaml file or many objects, this would help the developers to quickly remedy the issue.

shahrokni (approved):

PR #4339, opened 2026-08-05
[FEATURE] add k8s user impersonation support

jgbernalp:
# Description

Add support for k8s impersonation by checking the `Impersonate-User` header, if present, a SAR is done to check if the user can impersonate and uses the username provided.

# Checklist

### DIFF
--- a/internal/api/authorization/k8s/k8s_test.go
+++ b/internal/api/authorization/k8s/k8s_test.go
@@ -15,6 +15,7 @@ package k8s
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -22,6 +23,7 @@ import (
 	"testing"
 
 	"github.com/labstack/echo/v4"
+	apiInterface "github.com/perses/perses/internal/api/interface"
 	v1Role "github.com/perses/perses/pkg/model/api/v1/role"
 	"github.com/stretchr/testify/assert"
 	authnv1 "k8s.io/api/authentication/v1"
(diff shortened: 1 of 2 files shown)


internal/api/authorization/k8s/k8s_test.go
```diff
@@ -982,3 +997,170 @@ func TestGetPermissions(t *testing.T) {
...
+		req.Header.Set("Impersonate-User", userZero)
+		rec := httptest.NewRecorder()
+
+		perms, err := mockK8s.GetPermissions(e.NewContext(req, rec))
+		assert.NoError(t, err)
+
+		// user0 has read GlobalDatasource in wildcard and read Dashboard/Datasource/Secret/Variable/EphemeralDashboard/Folder in project0
+		assert.Contains(t, perms, "*")
+		assert.Contains(t, perms, projectZero)
+		assert.NotContains(t, perms, projectOne)
```
PeterYurkovich: These lines don't actually test the users permissions. They should test for equality of the actual permission structure like how the TestGetPermission does. I think it would probably be helpful to extract out the permission definitions from the test into variables if they are going to be reused. While `TestGetUserImpersonation` is fine as a separate test, I think it may be best to combine the `TestHasPermissionWithImpersonation` and `TestGetPermissionsWithImpersonation` with the existing test functions and make use of the request mocking you are doing in the existing tests as well.

internal/api/authorization/k8s/k8s_test.go
```diff
@@ -971,14 +1033,104 @@ func TestGetPermissions(t *testing.T) {
...
+		},
 	}
 	for i := range testSuites {
 		test := testSuites[i]
 		t.Run(test.title, func(t *testing.T) {
 			req.Header.Set("Authorization", fmt.Sprintf("bearer %s-token", test.user))
+			if test.impersonateAs != "" {
+				req.Header.Set("Impersonate-User", test.impersonateAs)
+			} else {
+				req.Header.Del("Impersonate-User")
```
PeterYurkovich: We should probably recreate the request rather than modifying in place if we are needing to do things like delete headers between tests

PeterYurkovich (approved):

Nexucis (approved):

PR #4342, opened 2026-08-06
[BUGFIX] Fix data race when listing project resources across authorized projects

anxkhn:
### Problem

When native RBAC is enabled and a user has read access to more than one project, listing a project-scoped resource without a project in the URL path or query fans out one request per authorized project. All of those requests received the same `*Query` pointer and each async closure mutated it with `SetProjectQueryParam(project)` before the service and DAO read the field.

`async.Async` starts every supplied function immediately in its own goroutine, so the closures write and read `Query.Project` concurrently with no synchronization or copy. The outcome is nondeterministic: two goroutines can end up querying the same project, so the response can contain duplicates and silently omit resources. The race detector flags the concurrent access too.

This affects the global list endpoints for dashboards, datasources, variables, roles, role bindings, secrets, folders and ephemeral dashboards, which is what the UI uses for the cross-project views.

Reproduction: enable native RBAC, grant one user dashboard read access to two projects, create one dashboard in each, then repeatedly `GET /api/v1/dashboards` under `-race`.

### Fix

Make the per-project query immutable instead of shared:

- Replace `SetProjectQueryParam` on `databaseModel.Query` with a new `databaseModel.ProjectQuery` interface exposing `CloneWithProject(project string) ProjectQuery`.
- Implement `CloneWithProject` on every project-scoped query type. It returns a copy with `Project` set and leaves the caller's query untouched.
- Clone once per project before the goroutine is launched, so each async request owns its own query.

The interface change is internal only, so there is no impact on the public API or on stored data.

### Test

Added `TestListAcrossAuthorizedProjects` in `internal/api/toolbox`. A fake service blocks until both project requests are in flight, which makes the concurrency deterministic rather than timing dependent, then the test asserts that each goroutine received a distinct query, that the caller's query was not mutated, and that the response contains exactly one entry per authorized project. It fails on the previous code and passes with the fix.

```
go test -race ./internal/api/toolbox -run TestListAcrossAuthorizedProjects -count=50
```

No UI change, so no screenshots apply.

### DIFF
--- a/internal/api/toolbox/list.go
+++ b/internal/api/toolbox/list.go
@@ -18,9 +18,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/brunoga/deep"
 	"github.com/labstack/echo/v4"
 	"github.com/perses/common/async"
-	databaseModel "github.com/perses/perses/internal/api/database/model"
 	apiInterface "github.com/perses/perses/internal/api/interface"
 	"github.com/perses/perses/pkg/model/api"
 	modelV1 "github.com/perses/perses/pkg/model/api/v1"
@@ -194,8 +194,12 @@ func (t *toolbox[T, K, V]) metadataOrFullList(query V) (any, error) {
 }
 
 func (t *toolbox[T, K, V]) asyncMetadataOrFullList(project string, query V) func() (any, error) {
-	projectQuery := any(query).(databaseModel.ProjectQuery).CloneWithProject(project).(V)
 	return func() (any, error) {
-		return t.metadataOrFullList(projectQuery)
+		q, err := deep.Copy(query)
+		if err != nil {
+			return nil, err
+		}
+		q.SetProjectQueryParam(project)
+		return t.metadataOrFullList(q)
 	}
 }
--- a/internal/api/toolbox/list.go
+++ b/internal/api/toolbox/list.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/labstack/echo/v4"
 	"github.com/perses/common/async"
+	databaseModel "github.com/perses/perses/internal/api/database/model"
 	apiInterface "github.com/perses/perses/internal/api/interface"
 	"github.com/perses/perses/pkg/model/api"
 	modelV1 "github.com/perses/perses/pkg/model/api/v1"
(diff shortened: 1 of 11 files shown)


internal/api/toolbox/list.go
```diff
@@ -193,8 +194,8 @@ func (t *toolbox[T, K, V]) metadataOrFullList(query V) (any, error) {
 }
 
 func (t *toolbox[T, K, V]) asyncMetadataOrFullList(project string, query V) func() (any, error) {
+	projectQuery := any(query).(databaseModel.ProjectQuery).CloneWithProject(project).(V)
 	return func() (any, error) {
-		query.SetProjectQueryParam(project)
-		return t.metadataOrFullList(query)
+		return t.metadataOrFullList(projectQuery)
```
Nexucis: actually there is a simpler way to fix this issue, you can simply copy the object and use it afterwards

```suggestion
		q, err := deep.Copy(query)
		if err != nil {
			return nil, err
		}
		q.SetProjectQueryParam(project)
		return t.metadataOrFullList(q)
```

Nexucis (commented):
Thank you for rising this issue.

I have proposed another way to fix this issue that I think simpler. Let me know what you think about it

anxkhn:
@nexucis replaced the public query cloning api with a local deep copy per project and kept the race fix contained in the toolbox. rebased onto the latest main; the focused package tests and race test pass. @celian-garcia could you please take another look?

internal/api/toolbox/list_test.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (s *synchronizedDashboardService) RawList(query *dashboard.Query) ([]json.RawMessage, error) {
+	s.entered <- query
+	<-s.release
+	result, err := json.Marshal(query.Project)
+	return []json.RawMessage{result}, err
+}
+
+func TestListAcrossAuthorizedProjects(t *testing.T) {
```
Nexucis, line 71: I really don't understand how this test is working / testing the current bugfixes.

Can you add comments and details about the logic here please ? Thank you

Nexucis (approved):
thanks @anxkhn

PR #4343, opened 2026-08-06
[BUGFIX] Honor kubeconfig exec credential plugins during Kubernetes login

anxkhn:
## Description

Kubernetes login only ever used `rest.Config.BearerToken`. For any kubeconfig that authenticates through a client-go credential plugin (`user.exec`), that field is empty: client-go applies exec credentials through its HTTP transport wrappers instead of populating `BearerToken`.

As a result `percli login --kube` sent `Authorization: Bearer ` with an empty token, the exec command was never executed, and `WhoAmI` failed with 401. The same applied to a persisted client built from `K8sAuth`, which snapshotted the (empty) token once and could never rotate short-lived credentials.

## Changes

- Replace `K8sAuth.GetToken` with `K8sAuth.WrapRoundTripper`, which builds the client-go authentication wrappers via `rest.HTTPWrappersForConfig` and layers them over the existing Perses transport. Only the auth wrappers are taken from the kubeconfig, so the Perses TLS/proxy configuration is untouched.
- Use that helper in `k8sLogin.Login` instead of writing a static `Authorization` header.
- Use the same helper in `NewRESTClient` so a persisted Kubernetes-authenticated client re-executes and refreshes the credential plugin per request instead of caching a token taken at construction time.

Static-token and basic-auth kubeconfigs keep working, since `HTTPWrappersForConfig` installs the corresponding round trippers for those too.

Note for reviewers: `K8sAuth.GetToken` is exported and is removed here rather than deprecated, because the value it returned is meaningless for exec-based kubeconfigs. Happy to keep a deprecated shim instead if backwards compatibility of `pkg/client/config` matters for this release.

## Testing

Added a test that writes a kubeconfig whose `exec` command is the test binary itself, emitting an `ExecCredential` with an already-expired `expirationTimestamp` and an incrementing token. It asserts that:

- `login` triggers the plugin and sends `Authorization: Bearer token-1`,
- a persisted REST client built from `K8sAuth` re-executes the plugin and sends `token-2` and `token-3` on subsequent requests, proving credential rotation.

```
go test ./pkg/client/config ./internal/cli/cmd/login -run 'Test.*K8s.*Exec' -count=1
```

No UI changes.

### DIFF
--- a/internal/cli/cmd/login/k8s.go
+++ b/internal/cli/cmd/login/k8s.go
@@ -27,6 +27,7 @@ import (
 type k8sLogin struct {
 	apiClient          api.ClientInterface
 	kubeconfigLocation string
+	baseTransport      http.RoundTripper
 }
 
 func NewK8sLogin(apiClient api.ClientInterface, kubeconfigLocation string) *k8sLogin {
@@ -43,12 +44,14 @@ func (k *k8sLogin) Login() (*oauth2.Token, error) {
 		httpClient = &http.Client{Timeout: http.DefaultClient.Timeout}
 		restClient.Client = httpClient
 	}
-	roundTripper := httpClient.Transport
-	if roundTripper == nil {
-		roundTripper = http.DefaultTransport
+	if k.baseTransport == nil {
+		k.baseTransport = httpClient.Transport
+		if k.baseTransport == nil {
+			k.baseTransport = http.DefaultTransport
+		}
 	}
 
-	roundTripper, err := (&config.K8sAuth{KubeconfigFile: k.kubeconfigLocation}).WrapRoundTripper(roundTripper)
+	roundTripper, err := (&config.K8sAuth{KubeconfigFile: k.kubeconfigLocation}).WrapRoundTripper(k.baseTransport)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/cli/cmd/login/k8s.go
+++ b/internal/cli/cmd/login/k8s.go
@@ -14,7 +14,7 @@
 package login
 
 import (
-	"fmt"
+	"net/http"
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


gangadhar-res (commented):
Took a look at this since I've been in kubeconfig code elsewhere recently. The approach looks right, and the part worth spelling out is why: `rest.HTTPWrappersForConfig` goes through `config.TransportConfig()`, which is where an `ExecProvider` gets turned into a `WrapTransport`, and `transport.HTTPWrappersForConfig` applies that first in the chain. So the plugin is consulted per request instead of once at login, which is exactly what the old `BearerToken` read could never do.

The empty `oauth2.Token` from `Login()` looks deliberate rather than an oversight. With `accessToken` empty, `login.go` skips `restConfig.Authorization` and persists `K8sAuth` instead, so credentials get re-resolved on every client build and no static token ends up in the config file.

One gap that might deserve a docs line: cert based kubeconfigs. `HTTPWrappersForConfig` never touches the TLS section, so those send no credentials at all and fail as a bare 401. Not a regression and probably not fixable here, but it'd be easy to hit and annoying to diagnose.

Nice test too, forcing an already expired `expirationTimestamp` proves the rotation rather than just the wiring.

celian-garcia:
@PeterYurkovich @jgbernalp do you have any comment on that? 
I feel it legit, as long as it's not breaking the support of kubeconfig generated via kubectl or oc

internal/cli/cmd/login/k8s.go
```diff
@@ -37,24 +37,29 @@ func NewK8sLogin(apiClient api.ClientInterface, kubeconfigLocation string) *k8sL
...
+		restClient.Client = httpClient
+	}
+	roundTripper := httpClient.Transport
+	if roundTripper == nil {
+		roundTripper = http.DefaultTransport
 	}
 
-	if len(k.apiClient.RESTClient().Headers) == 0 {
-		k.apiClient.RESTClient().Headers = map[string]string{}
+	roundTripper, err := (&config.K8sAuth{KubeconfigFile: k.kubeconfigLocation}).WrapRoundTripper(roundTripper)
```
jgbernalp: Would this generate a resource leak?, it seems is wrapping on the previous transport, if `Refresh` is executed multiple times in the same instance then multiple layers of wrapping will be added.

jgbernalp (approved):

PR #4347, opened 2026-08-09
[ENHANCEMENT] Add option to generate dashboard name from title during Grafana migration

### LINKED ISSUE #3718: Grafana migration: add option to regenerate dashboard name from its title
### Proposal

Perses currently often displays the "technical" name of dashboards rather than their display name, for reasons explained [here](https://github.com/perses/perses/issues/3676#issuecomment-3656210173). Which means that if you migrated a dashboard from Grafana that potentially had a cryptic, auto-generated UID, you end up dealing with this meaningless information in search bar, home page etc rather than with its pretty display name.

While we can (should?) think about ways to (re)introduce the display name in these cases, we can also take the problem by the other end by making the technical name meaningful - both strategies could live together in the end, these are not mutually-exclusive.

So the idea is simply to do the same when importing dashboards than what is happening when creating a brand new one: in the latter case Perses generates the technical name from the display name provided by the user (simply takes the same string but replaces some forbidden char). Thus on the import screen we could have a new checkbox to enable the generation of a new technical name based on the display name (called "title" on Grafana side) of the dashboard to import, instead of taking the UID.

Something like this:

Image

thisis-manan:
# Description

When migrating a Grafana dashboard, the resulting Perses dashboard reused the Grafana UID as its `metadata.name`. Since the UID is often a cryptic auto-generated string (e.g. `aead1k2js`), it ends up being displayed in the search bar, home page, etc. rather than something readable.

This adds an opt-in to generate the technical name from the dashboard **title** instead — the same way a brand new dashboard is named (via the existing `generateMetadataName` helper). The Grafana UID is kept as a fallback when the title is empty or only contains characters that get stripped out during name generation.

The option is available across the three migration entry points:

- `generateDashboardName` field on the `/migrate` API request
- `--generate-dashboard-name` flag on `percli migrate`
- a checkbox in the Grafana import view

Example: a dashboard titled `Node Exporter Full` with UID `aead1k2js-4f21` migrates to `metadata.name: Node_Exporter_Full` when the option is enabled, and to `aead1k2js-4f21` when it isn't.

One note on naming: title-derived names are not guaranteed unique, which is why this stays opt-in and the UID remains the default behaviour.

Closes #3718

# Screenshots

New "Generate dashboard name from title" checkbox in the Grafana import view, next to the existing "Use default datasource in Perses" option (unchecked / checked):

_Screenshots attached below._
image

image

# Checklist

- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`, `IGNORE`.

## UI Changes

- [x] Code follows the UI guidelines.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -71,6 +71,15 @@ Note: In case you would like to have the result as a K8s CustomResource, you can
     percli migrate -f grafana-dashboard.json --online --use-default-datasource -o json > perses-dashboard.json
     ```
 
+!!! tip
+    By default the migrated dashboard reuses the Grafana UID as its technical name (`metadata.name`),
+    which is often a cryptic auto-generated string. If you want the technical name to be generated from
+    the dashboard title instead, you can use the `--generate-dashboard-name` flag.
+
+    ```bash
+    percli migrate -f grafana-dashboard.json --online --generate-dashboard-name -o json > perses-dashboard.json
+    ```
+
 - You should check the unsupported migrations. For example, in case of a variable, you will get a static variable like this:
 
 ```json
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,7 @@ require (
 	golang.org/x/crypto v0.54.0
 	golang.org/x/mod v0.38.0
 	golang.org/x/oauth2 v0.36.0
+	golang.org/x/text v0.40.0
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.36.3
 	k8s.io/apimachinery v0.36.3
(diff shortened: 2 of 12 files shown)


Nexucis:
Hello,
thank you for your contributions.

I am currently working on an improvement of the search bar so it can display the field `spec.display.name` instead of the UID.

And in the home page it should already be the case (displaying the `spec.display.name` instead of the field `metadata.name`).

Do you think if the search bar is improved in this direction, this PR is not necessary ?

thisis-manan:
Yes that makes sense displaying spec.display.name in the search bar is the better fix since it covers all dashboards, not just migrated ones. Closing this in favour of that direction.

PR #4352, opened 2026-08-11
[ENHANCEMENT] Config: `frontend.time_range.options` is sorted in ascending order and values are verified

pujitha24:
# Description

The `frontend.time_range.options` config value lets an admin customize the list of
quick time range options shown in the UI's time range dropdown. `TimeRange.Verify()`
only checked whether the list was empty (falling back to the built-in defaults) but
never validated or sorted the values, so if an admin listed the durations out of
order in `config.yaml` (e.g. `1h, 5m, 7d, 30m`), the UI dropdown rendered them in
that same out-of-order sequence.

This PR sorts `frontend.time_range.options` ascending by parsed duration inside
`TimeRange.Verify()`, using `common.ParseDuration` from `github.com/perses/spec/go/common`.
As a side effect, a duration string that fails to parse now causes `Verify()` to
return an error instead of silently being kept in the list — previously such a value
could only reach `Verify()` via a value with an empty duration or a value assembled by
Go code (not by loading a real YAML/JSON config, since `common.DurationString`
already validates non-empty strings on unmarshal). This is a defensive tightening,
not expected to affect any currently working deployment.

The built-in default list (`5m, 15m, 30m, 1h, 6h, 12h, 24h, 7d, 14d`) was already in
ascending order, so default deployments are unaffected; only deployments with a
custom, out-of-order `options` list in their config will see a behavior change (the
dropdown will now render sorted).

# Screenshots

N/A — this is a backend config-validation change; no UI code in this repository was
modified. (The dropdown rendering component itself now lives in a separate
`perses/shared` repository and simply renders the `options` array in the order it
receives it, so sorting at the config layer here fixes the root cause for the config-driven case.)

# Checklist

## UI Changes

## Validation

- `go build ./pkg/model/api/config/...` — passes.
- `go vet ./pkg/model/api/config/...` — passes.
- `gofmt -l pkg/model/api/config/frontend.go pkg/model/api/config/frontend_test.go` — no output (correctly formatted).
- `golangci-lint run --timeout 5m ./pkg/model/api/config/...` (v2.12.2, matching CI's v2.12.1) — 0 issues.
- `go test ./pkg/model/api/config/... -count=1 -v` — all tests pass, including pre-existing tests
  (`TestJSONMarshalConfig`, `TestUnmarshalYAMLConfig`, etc.) and 5 new tests covering: sorting an
  out-of-order custom list, keeping an already-sorted list unchanged, stable ordering of
  equal-duration values, the (already-sorted) defaults, and rejecting an invalid duration string.
- Confirmed the new `TestTimeRange_VerifySortsCustomOptions` test fails on the pre-fix code
  (verified via `git stash`) and passes after the fix — this is the reproduction for the bug.
- Not run: `go build ./...` / `make test` — these require a prebuilt UI frontend dist to satisfy an
  `embedFS` symbol in `ui/endpoint.go`, and `make integration-test` requires Postgres/Prometheus
  service containers per CI. This is unrelated to the change (confirmed the same `embedFS` build
  failure exists on `main` without this diff) and out of scope for a change scoped entirely to
  `pkg/model/api/config`.

### DIFF
--- a/pkg/model/api/config/frontend_test.go
+++ b/pkg/model/api/config/frontend_test.go
@@ -20,39 +20,50 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestTimeRange_VerifyDefaultsAreSorted(t *testing.T) {
-	tr := &TimeRange{}
-	assert.NoError(t, tr.Verify())
-	assert.Equal(t, defaultTimeRangeOptions, tr.Options)
-}
-
-func TestTimeRange_VerifySortsCustomOptions(t *testing.T) {
-	tr := &TimeRange{
-		Options: []common.DurationString{"1h", "5m", "7d", "30m"},
+func TestTimeRange_Verify(t *testing.T) {
+	testSuite := []struct {
+		title           string
+		options         []common.DurationString
+		errMessage      string
+		expectedOptions []common.DurationString
+	}{
+		{
+			title:           "defaults are used and already sorted when no options are provided",
+			options:         nil,
+			expectedOptions: defaultTimeRangeOptions,
+		},
+		{
+			title:           "custom options are sorted ascending",
+			options:         []common.DurationString{"1h", "5m", "7d", "30m"},
+			expectedOptions: []common.DurationString{"5m", "30m", "1h", "7d"},
+		},
+		{
+			title:      "invalid option returns an error",
+			options:    []common.DurationString{"1h", "not-a-duration"},
+			errMessage: "not-a-duration",
+		},
+		{
+			title:           "already sorted options are kept as-is",
+			options:         []common.DurationString{"5m", "1h", "7d"},
+			expectedOptions: []common.DurationString{"5m", "1h", "7d"},
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/model/api/config/frontend_test.go
```diff
@@ -0,0 +1,58 @@
...
+package config
+
+import (
+	"testing"
+
+	"github.com/perses/spec/go/common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTimeRange_VerifyDefaultsAreSorted(t *testing.T) {
```
Nexucis: Can you create a single test function containing a struct that will list the different case.

You have various examples in the different test files to see how it should be represented. Thank you !

Nexucis (changes requested):
Hello,

Thank you for the improvement !

Overall it looks fine to me. I just would like a refactoring of the unit tests and we should be good.
Thank you by advance for making the changes !

Nexucis (approved):
LGTM. Thank you !

Issue #3249, opened 2025-08-28
Search bar: Show display name of resource


AntoineThebaud:
### Proposal

I think the search results should show the display name of the resource first, rather than the technical name. Then, if there's room for it the technical name could still be shown next to it (grayed out maybe)

<img width="1064" height="248" alt="Image" src="https://github.com/user-attachments/assets/b12b0ac1-840a-4547-92fa-6665c9138983" />

mockup of my proposal:

<img width="1066" height="249" alt="Image" src="https://github.com/user-attachments/assets/4e17743b-c946-4e8a-b654-997c1e3e5d6b" />

⚠️ open question: should the search logic be ran against both the display name & technical name attributes? Today we are checking technical name only.

Nexucis:
the search is working only on the metadata. The display name is part of the spec and as such this is not possible to make this search or display as we don't download the spec for performance issue

AntoineThebaud:
Oh indeed I forgot about that.. well the use case still makes sense, especially if you have many dashboards you imported from Grafana where the ID has good chances to be a random character sequence, in such case not being able to search based on the display name becomes a strong limitation..

AntoineThebaud:
In the same way we added a dedicated endpoint to return only the metadata, could we have one returning metadata + `display`?

PR #4355, opened 2026-08-13
[FEATURE] Add new search endpoint

Nexucis:
This PR is adding a new search endpoint `/api/v1/search/dashboards` that increases the search capabilities for the dashboard.

### How it works ?

1. In memory, the dashboards are indexed. The key indexed can be configured. For example if you want to search through all panels name, then you can index them all.

2. When a dashboard is added or deleted the index is updated. Then if you are using SQL, there is a cron task running checking if the table dashboards has been modified. Like that every instances of Perses will synchronised and the index will be the same between the instances

3. I have implemented a little fuzzy search inside Perses. It is more or less the same algorithm used for the fuzzy search we are using in the front end

Example:

```bash
curl -XGET http://localhost:8080/api/v1/search/dashboards
```

Result:

```json
[
  {
    "metadata": {
      "name": "perses_memory",
      "createdAt": "2024-12-11T13:07:24.451467Z",
      "updatedAt": "2024-12-11T13:15:09.277842Z",
      "version": 6,
      "project": "perses"
    },
    "displayName": "perses_memory",
    "original": "perses_memory",
    "intervals": [
    {
      "from": 0,
      "to": 3
    }
    ],
    "score": 16
  },
  {
    "metadata": {
      "name": "panelgroups",
      "createdAt": "2024-10-03T14:46:03.572542Z",
      "updatedAt": "2026-08-13T08:23:09.345955Z",
      "version": 293,
      "project": "testing"
    },
    "displayName": "panelgroups",
    "original": "panelgroups",
    "intervals": [
    {
      "from": 0,
      "to": 0
    },
    {
      "from": 3,
      "to": 3
    },
    {
      "from": 6,
      "to": 6
    },
    {
      "from": 10,
      "to": 10
    }
    ],
    "score": 4
  }
]
```

Related to:
- https://github.com/perses/perses/issues/3249
- https://github.com/perses/perses/discussions/3798

Edit:

I have added a way to list the dashboard using this new endpoint. This thing can be used to list the dashboards in the project view for example.


Nexucis:
I need to figure out how to fix the cue issue 🤔

ibakshay:
Maybe @gfierkasap can help you out with the CUE issue :)

docs/configuration/configuration.md
```diff
@@ -907,3 +910,27 @@ message: <string>
...
+# But the more keys you add, the more time it will take to index the resources and the more memory it will take to store the index. So be careful when adding keys.
+# Example of keys:
+# - "metadata.name": it will index the name of the resource (that was the default behavior before the search feature was added)
+# - "metadata.tags": it will index the tags of the resource (if it has any)
+# - "spec.display.name": it will index the display name of the resource (if it has one)
+# - "spec.panels.@dig:display.name": It will index the display name of all panels in a dashboard. The @dig is used to dig into the panels array and index the display name of each panel.
+# We are using gjson to find the value of the key in the resource. So you can use any valid gjson path to index the value you want.
+# Syntax is available here: https://github.com/tidwall/gjson/blob/master/SYNTAX.md
+index_keys:
+  dashboard: <list of strings | default = ["metadata.name", "spec.display.name"]> # Optional
```
celian-garcia, line 935: if we want to use it to list dashboards with display name, then full metadata + display name will be the minimum. Can we say that by default, we have full metadata and display name and people can add some ?

internal/api/database/file/file.go
```diff
@@ -110,57 +110,48 @@ func (d *DAO) Get(kind modelV1.Kind, metadata modelAPI.Metadata, entity modelAPI
...
 		}
+		ch <- jsonData
 	}
-	return result, nil
+	return nil
+}
+
+func (d *DAO) RawMetadataQuery(_ databaseModel.Query, _ modelV1.Kind) ([]json.RawMessage, error) {
+	// this is implemented in the dao struct in database.go. This is just here to satisfy the interface.
+	return nil, fmt.Errorf("raw metadata query not implemented")
```
celian-garcia, line 149: I'm not fan of that. Maybe you can move rawquery and rawmetadataquery into standalone functions. A kind of helper with such definition

```go
func RawQuery(d *dao.DAO, query databaseModel.Query) ([]json.RawMessage, error) {
	ch := make(chan json.RawMessage)
	var err error
	go func() {
		err = d.StreamRaw(query, ch)
	}()
	result := make([]json.RawMessage, 0)
	for data := range ch {
		result = append(result, data)
	}
	return result, err
} 
```


so you wouldn't have to respect that interface in low level database

internal/api/impl/v1/search/search.go
```diff
@@ -0,0 +1,59 @@
...
+
+func NewEndpoint(index index.Client) route.Endpoint {
+	return &endpoint{
+		index: index,
+	}
+}
+
+func (e *endpoint) CollectRoutes(g *route.Group) {
+	group := g.Group(fmt.Sprintf("/%s", utils.PathSearch))
+	group.GET(fmt.Sprintf("/%s", utils.PathDashboard), e.Dashboard, false)
```
celian-garcia, line 40: Don't you have to check permissions manually when you're not using toolbox ?

celian-garcia (changes requested):
I reviewed without the business of the indexer + mapping that is not very readable but tested ^^'

docs/configuration/configuration.md
Nexucis: the usage of the `index_keys` does not impact the way to get the display name. This is a different process where user does not have the hand on this.

`Index_keys` is telling you what part of the JSON document you are going to indexed and so what part of the document you want to use for the search.

The default value is just here to be able to search through the `metadata.name` and the `spec.display.name`

internal/api/impl/v1/search/search.go
Nexucis: this is the case, you need to look at the code `internal/api/index/index.go`

docs/configuration/configuration.md
celian-garcia: oh you mean that if we make a list calling the endpoint without query params, then it will give still the display name in the output ? Even if index_keys are empty ?

internal/api/database/file/file.go
Nexucis: the way to manage the database is a different topic and I did not want to change the way it is working currently.

I just took the opportunity of the function `StreamRaw` to optimise a bit the code. But I did not mean to change the way we are managing that.

If we want to refactor the code of the database, this is a different topic and I am opened to see improvement. Currently this is working fine and I did not see any particular issue with the current implementation.

celian-garcia (approved):

PR #4359, opened 2026-08-14
[ENHANCEMENT] Track datasource origin and delete stale discovered datasources 

### LINKED ISSUE #2747: Datasource Discovery: it should delete datasource when the discovery is updated
### Proposal

In the datasource discovery, when the response from discovery is changing, specially when there is less datasource returned, it's not reflected in the Perses database.

Saying differently, when a datasource is removed in the discovery, it's removed from the Perses database.

This is because today it's not possible to identify a datasource created from the discovery. Since we cannot identify them, we are not able to make a difference between the list of datasource coming from the discovery and the one stored.

A way to identify the datasource would be to add a field `source` with the possible value : `provisioning`, `manual`, `discovery`. 

Probably for the `provisioning` and the `discovery` we would need additional information such as the type and the name of discovery, or for the provisioning the path / the name of the file where it comes from

gfierkasap:
Resolves #2747 

# Description

When datasource discovery returns fewer datasources than in a previous run, stale entries were never removed from the database. There was also no way to distinguish datasources created manually via the API from those managed by discovery or provisioning.

This PR implements a new field `GlobalDatasource.Metadata.Source` with the possible values: `manual`, `provisioning` and `discovery`, that allows to distinguish the differently sourced datasources, extends the datasource querying capabilities to filter for the source, and introduces automatic management of `discovery` datasources.


pkg/model/api/v1/metadata.go
```diff
@@ -24,12 +24,68 @@ import (
...
+	DiscoverySource    MetadataSource = "discovery"
+)
+
 func NewMetadata(name string) *Metadata {
 	return &Metadata{
 		Name: name,
 	}
 }
 
+type DatasourceMetadata struct {
```
celian-garcia, line 45: Don't we want to add source in ``Metadata``? This is the same problem with provisioning and the scope of the provisioning is larger.

internal/api/discovery/service/service.go
```diff
@@ -53,5 +61,27 @@ func (a *ApplyService) Apply(entities []*v1.GlobalDatasource) {
...
+	foundDatasources, err := a.svc.List(
+		&globaldatasource.Query{
+			Source:        v1.DiscoverySource,
+			DiscoveryName: a.discoveryName,
+		})
+	if err != nil {
+		logrus.WithError(err).Error("unable to get discovered globaldatasources")
+	}
+	for _, ds := range foundDatasources {
+		if !slices.Contains(currentNames, ds.Metadata.Name) {
```
celian-garcia: Database can be set in a case insensitive mode. 
I'm wondering how the discovery mechanism behaves in that case. For example if a discovered datasource change of name going from ``myDatasource`` to ``MyDatasource``.

I'm thinking of that because here the contains is case sensitive, so maybe there's a gap

pkg/model/api/v1/metadata.go
```diff
@@ -24,12 +24,68 @@ import (
...
+	Metadata      `json:",inline" yaml:",inline"`
+	Source        MetadataSource `json:"source,omitempty" yaml:"source,omitempty"`
+	DiscoveryName string         `json:"discovery_name,omitempty" yaml:"discovery_name,omitempty"`
```
celian-garcia, line 48: You have to add Discovery name and type, this is two different things

pkg/model/api/v1/metadata.go
celian-garcia: (that said, I just heard augustin in my ear saying "don't touch metadata" so I am not sure hehe)

celian-garcia (commented):
Thank you for working on that, this is interesting! It does not contain the provisioning part and if the enum value is possible, it should be supported in the code as well in my opinion.

Here are some comments, but I am also interested in the @Nexucis opinion about the location in the model

s3onghyun (commented):
Nice feature — pruning stale discovered datasources closes a real gap (the #2747 case). One correctness concern on the pruning logic in `internal/api/discovery/service/service.go`, plus a small nit.

### Pruning deletes everything when a discovery cycle returns an empty-but-successful result

`Apply` gathers the names it just saw into `currentNames`, then deletes every stored datasource for this `discoveryName` that isn't in that set:

```go
foundDatasources, err := a.svc.List(&globaldatasource.Query{
    Source: v1.DiscoverySource, DiscoveryName: a.discoveryName,
})
...
for _, ds := range foundDatasources {
    if !slices.Contains(currentNames, ds.Metadata.Name) {
        a.svc.Delete(nil, ...)
    }
}
```

If `Apply` is called with an empty `entities` slice, `currentNames` is empty and **every** datasource previously discovered by this discovery gets deleted. That empty-but-successful case is reachable on both discoverers — and note it's distinct from the error case, which is already handled safely (both `return nil` *before* `Apply`):

- **HTTP** (`discovery/http/http.go`): an endpoint returning `200` with `[]` — e.g. during its own restart/deploy, or a transient upstream — passes the `err != nil` guard with an empty `result`, then calls `Apply(result)`.
- **Kubernetes** (`discovery/kubernetes/service.go`): if the label selector / `ServiceType` filter matches nothing this cycle, `discover()` returns `(nil, nil)` and `Apply(nil)` runs.

The result is a delete-then-recreate churn on the next good refresh (dashboards referencing those datasources break in the gap), and if the empty state persists (misconfigured endpoint or selector) the datasources stay gone. A guard to skip pruning when there's nothing to reconcile avoids it:

```go
if len(entities) == 0 {
    return
}
```

If an intentional "prune to empty" is ever desired, it'd be safer as an explicit separate path than as the default behavior of an empty poll.

### Nit: `List` error falls through to the delete loop

When `a.svc.List(...)` fails, the error is logged but execution continues into the delete loop over a nil slice. Harmless today (nil ⇒ no deletes), but an explicit `return` after the log would make the "don't prune if we couldn't read the current state" intent clear and keep it correct if the loop changes later.

(Reviewing as a contributor — no merge rights here. The model-location/provisioning points are already well covered by @celian-garcia above.)

internal/api/discovery/service/service.go
gfierkasap: I actually wasn't aware that it can be case insensitive, thanks for pointing it out! I've implemented a small fix in [4965f78](https://github.com/perses/perses/pull/4359/commits/4965f78231ddb1b3d87177ad7db1f1b3f7d35338), wdyt?

pkg/model/api/v1/metadata.go
gfierkasap: Sure, I added it in [a2210b8](https://github.com/perses/perses/pull/4359/commits/a2210b8484121ff56373968e1021733b6d0a4df9) and [a0774d5](https://github.com/perses/perses/pull/4359/commits/a0774d53717789dffb9d995ea798774fcbdf260a) 

EDIT: kinda forgot to update the Query struct as well, will update asap
EDIT 2: updated `Query` struct in [e74eb03](https://github.com/perses/perses/pull/4359/commits/e74eb039c0cafe122554b0a71fa3de9cdee64138), with a minor refactor of the `applyListFilter` function

pkg/model/api/v1/metadata.go
gfierkasap: I actually actively wanted to avoid changing `Metadata` as it is used basically everywhere, hence the new `DatasourceMetadata` ;) Embedding it seemed like a better option

gfierkasap:
@s3onghyun Thank you for the comprehensive review and explanation! You raised two very good points, as I actually didn't consider the scenario of empty `entities` slice being passed to the `Apply` method. I implemented the suggested safeguard in [0ad3df4](https://github.com/perses/perses/pull/4359/commits/0ad3df4a18d486d39c849a6ec1729c7d55713888), as well as the one for the scenario where `a.svc.List()` errors out in [0d16e8c](https://github.com/perses/perses/pull/4359/commits/0d16e8c5796d4ce2c4b2bc6d060c2c414ceeb6be)

s3onghyun:
Thanks for the quick turnaround, @gfierkasap — I had a look at the new commits and both fixes land exactly where they should:

- the `if len(entities) == 0 { return }` guard at the top of `Apply` (0ad3df4) closes the empty-cycle mass-deletion cleanly, and
- the `return` after the failed `List` (0d16e8c) makes the "don't prune when we couldn't read current state" intent explicit.

The `EqualFold` helper for the case-insensitive comparison is a nice touch too. Looks good from my side on these points 👍

s3onghyun:
Hi @celian-garcia — honest answer: yes, I use an AI assistant to help me read through the code and draft comments. But I verify every claim against the source myself before posting, and the conclusions are my own. Thanks for asking directly.

And you're right on both counts — I'll keep it much more concise and put comments inline on the relevant lines from now on. Appreciate the nudge 🙏

PR #4366, opened 2026-08-20
[ENHANCEMENT] UI: use dashboard list from search endpoint

Nexucis:
In the project list, we should not get the list of the dashboard with the complete dashboard definition (aka containing the spec) because it can cost a lot.

Instead we can now use the new endpoint `/api/v1/search/dashboards` that can give you the list of the dashboard without the spec but including the display name.

As we still need the spec to duplicate a dashboard or when editing the name, we are getting the full dashboard one by one when editing or duplicating a dashboard

NB: I have used AI to assist me on editing the file `DashboardList`. I am not sure the way to get a dashboard when editing and duplicating a dashboard is the best way to do it

Edit: The complete list of dashboard was also used during the creation of a folder and during the creation of a dashboard (in the dialog). I have changed that as well


ui/app/src/components/DashboardList/DashboardList.tsx
```diff
@@ -240,10 +253,10 @@ export function DashboardList(props: DashboardListProperties): ReactElement {
         handleDeleteFolderButtonClick={handleDeleteFolderButtonClick}
         isLoading={isLoading}
       />
-      {activeDialog.type === 'editDashboard' && (
+      {activeDialog.type === 'editDashboard' && editDashboardData && (
```
jgbernalp, line 291: we need to handle the case the dashboard for edit is loading, otherwise the dialog won't show without any indication

ui/app/src/components/DashboardList/DashboardList.tsx
```diff
@@ -262,7 +275,18 @@ export function DashboardList(props: DashboardListProperties): ReactElement {
...
+          resource={{
+            kind: 'Dashboard',
+            metadata: activeDialog.target.metadata,
+            spec: {
+              panels: {},
+              layouts: [],
+              variables: [],
+              display: { name: activeDialog.target.displayName },
+              duration: '1h' as DurationString,
+              refreshInterval: '30s' as DurationString,
```
jgbernalp: Do we need to provide this full spec for deleting?

ui/app/src/components/DashboardList/DashboardList.tsx
jgbernalp: Should we guard here for the duplicateDashboardTarget?

ui/app/src/components/DashboardList/DashboardList.tsx
Nexucis: well it's because it is requesting the type `resource` and so a complete `DashboardResource`. So I had to put these dummy values to make it compile

ui/app/src/components/DashboardList/DashboardList.tsx
Nexucis: yes good point, we need to do that

ui/app/src/components/DashboardList/DashboardList.tsx
Nexucis: ah, I thought it was magically handling it. But ok I will find a way to handle it

ui/app/src/components/DashboardList/DashboardList.tsx
Nexucis: ah wait, maybe not if we do not need the spec. Or I did not get your point maybe 🤔

jgbernalp: > ah wait, maybe not if we do not need the spec. Or I did not get your point maybe 🤔

If you mean the previous comment yes. We might not need the spec to delete IIUC only kind and name are required to do that

ui/app/src/model/search-client.ts
```diff
@@ -0,0 +1,50 @@
...
+  metadata: ProjectMetadata;
+  displayName: string;
+}
+
+export function useSearchDashboards(
+  project?: string,
+  searchQuery?: string,
+): UseQueryResult<SearchProjectResource[], StatusError> {
+  return useQuery<SearchProjectResource[], StatusError>({
+    queryKey: ['search/dashboards', project, searchQuery],
```
Gladorme: If someone edit or delete a dashboard, this cache will not clear / update them. Small issue, but if we want to have this cache up to date, we need to invalidate it when action are made on dashboards

ui/app/src/views/projects/tabs/ProjectDashboards.tsx
```diff
@@ -17,14 +17,14 @@ import type { ReactElement } from 'react';
...
-import { useDashboardList } from '../../../model/dashboard-client';
 import { useFolderList } from '../../../model/folder-client';
+import { useSearchDashboards } from '../../../model/search-client';
 
 interface ProjectDashboardsProps extends BoxProps {
   projectName: string;
 }
 export function ProjectDashboards({ projectName, ...props }: ProjectDashboardsProps): ReactElement {
-  const { data, isLoading } = useDashboardList({ project: projectName });
+  const { data, isLoading } = useSearchDashboards(projectName);
```
Gladorme: Why change this? 🤔

ui/app/src/views/projects/ProjectTabs.tsx
```diff
@@ -430,7 +431,7 @@ export function ProjectTabs(props: DashboardVariableTabsProps): ReactElement {
   const hasEphemeralDashboards = (data ?? []).length > 0;
 
   // Fetch counts for tab badges
-  const { data: dashboards } = useDashboardList({ project: projectName, metadataOnly: true });
+  const { data: dashboards } = useSearchDashboards(projectName);
```
Gladorme: Why change this? 🤔

ui/app/src/model/dashboard-client.ts
```diff
@@ -54,12 +55,17 @@ export function useCreateDashboardMutation(
...
+  project: string,
+  name: string,
+  enabled: boolean = true,
+): UseQueryResult<DashboardResource, StatusError> {
   return useQuery<DashboardResource, StatusError>({
     queryKey: [resource, project, name],
     queryFn: () => {
       return getDashboard(project, name);
     },
+    enabled,
```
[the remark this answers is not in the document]
Nexucis: @copilot keepPreviousData is deprecated and as such cannot be used and tanstak does not provide a replacement so far

AntoineThebaud (approved):
❤️

ui/app/src/components/DashboardList/DashboardList.tsx
```diff
@@ -241,20 +304,39 @@ export function DashboardList(props: DashboardListProperties): ReactElement {
         handleDeleteFolderButtonClick={handleDeleteFolderButtonClick}
         isLoading={isLoading}
       />
-      {activeDialog.type === 'editDashboard' && (
-        <EditDashboardDialog
-          open={activeDialog.type === 'editDashboard'}
-          dashboard={activeDialog.target}
-          onClose={closeDialog}
-        />
+      {(isEditDashboardLoading || isDuplicateDashboardLoading) && (
```
Gladorme, line 307: Look strange these conditions. Loading indicator + error must be handled directly by the EditDialog if necessary

ui/app/src/components/DashboardList/DashboardList.tsx
```diff
@@ -118,11 +120,72 @@ export function DashboardList(props: DashboardListProperties): ReactElement {
 
   const [activeDialog, setActiveDialog] = useState<openDialogAction>({ type: 'none' });
 
+  // We must make sure the dashboard has been fetched *after* the dialog was opened: when stale cached data exists,
+  // `isLoading` is false while a background refetch is in flight, and the user could edit/duplicate an outdated version
+  // of the dashboard and overwrite a newer server spec.
+  // We compare `dataUpdatedAt` with the time the dialog was opened rather than relying on `isFetching`: the update
```
Gladorme, line 126: Not 100% clear to me this comment.

ui/app/src/model/search-client.ts
```diff
@@ -0,0 +1,51 @@
...
+  metadata: ProjectMetadata;
+  displayName: string;
+}
+
+export function useSearchDashboards(
+  project?: string,
+  searchQuery?: string,
+): UseQueryResult<SearchProjectResource[], StatusError> {
+  return useQuery<SearchProjectResource[], StatusError>({
+    // The queryKey contains the key dashboards and not simply the resource `search/dashboards` to ensure it is invalidated when a dashboard is created or deleted
```
Gladorme, line 32: Strange hack, I prefer to have a dedicated key for search and invalidate it mutation dashboard functions

Nexucis: mmm I don't see it as a hack personally, I see it more like a way to use the key in tanstack to sync things together and surprisedly it works pretty well for once.

ui/app/src/model/search-client.ts
```diff
@@ -0,0 +1,51 @@
...
+    queryKey: ['dashboards', 'search', project, searchQuery],
```
Gladorme: The key will conflict with `dashboard-client.ts`. If a project is called `search`, it may do some strange stuff. But it will well invalidate cache in case a dashboard is deleted, edited or created. What do you think about using a key less generic?
```suggestion
    queryKey: ['dashboards', '__INTERNAL_SEARCH__', project, searchQuery],
```

Nexucis: yes good idea. Thank you !

Gladorme (approved):

PR #4368, opened 2026-08-20
[FEATURE] Add default flag to datasource discovery configuration

### LINKED ISSUE #4351: Allow Perses Kubernetes ServiceDiscovery to mark a discovered datasource as default
## Background

At the moment, datasources created by [Perses Kubernetes service discovery](https://perses.dev/perses/docs/configuration/datasource-discovery/?h=datasource+discovery#kubernetes-service-discovery)
are always non-default, and there's no way to change that. Even setting one as
default by hand doesn't help, since it gets reset on the next refresh.

A common need is to discover datasources (e.g. Prometheus, OpenSearch,
Alertmanager) through this feature and, for each kind, pick one that acts as the
default. That way users can rely on the default when creating dashboards instead
of selecting the correct datasource every time. This matters a lot at scale, when
an organization manages hundreds of dashboards as expecting every user to pick the
right datasource each time isn't practical. The default datasources for every plugin kind should be
something the Perses administrator sets up once, centrally.

## Proposal

Add a `default` flag to `kubernetes_sd`. When it's `true`, the discovered
datasource is created as the default one.

```yaml
datasource:
  global:
    discovery:
    - kubernetes_sd:
        datasource_plugin_kind: "AlertManagerDatasource"
        namespace: "alertmanager"
        service_configuration:
          enable: true
          port_name: "http-web"
        labels:
          perses.dev/default-datasource: "true"
        default: true  # <<<- this one
```
Because a default is per datasource kind, administrators can have one default for
each kind at the same time. The typical setup is two discovery entries per
datasource kind: one for the default service and one for everything else.
Here's how it would look:

```yaml
datasource:
  global:
    discovery:
      # --- Prometheus ---
      - name: "prometheus-default"
        kubernetes_sd:
          datasource_plugin_kind: "PrometheusDatasource"
          service_configuration:
            enable: true
            port_name: "http-web"
          labels:
            perses.dev/default-datasource: "true"
          default: true
      - name: "prometheus-others"
        kubernetes_sd:
          datasource_plugin_kind: "PrometheusDatasource"
          service_configuration:
            enable: true
            port_name: "http-web"
          labels:
            perses.dev/datasource: "true"
          default: false

      # --- OpenSearch ---
      - name: "opensearch-default"
        kubernetes_sd:
          datasource_plugin_kind: "OpenSearchDatasource"
          service_configuration:
            enable: true
            port_name: "http"
          labels:
            perses.dev/default-datasource: "true"
          default: true
      - name: "opensearch-others"
        kubernetes_sd:
          datasource_plugin_kind: "OpenSearchDatasource"
          service_configuration:
            enable: true
            port_name: "http"
          labels:
            perses.dev/datasource: "true"
          default: false

      # --- Alertmanager ---
      - name: "alertmanager-default"
        kubernetes_sd:
          datasource_plugin_kind: "AlertManagerDatasource"
          namespace: "alertmanager"
          service_configuration:
            enable: true
            port_name: "http-web"
          labels:
            perses.dev/default-datasource: "true"
          default: true
      - name: "alertmanager-others"
        kubernetes_sd:
          datasource_plugin_kind: "AlertManagerDatasource"
          namespace: "alertmanager"
          service_configuration:
            enable: true
            port_name: "http-web"
          labels:
            perses.dev/datasource: "true"
          default: false
```

One thing to keep in mind: there can only be one default per kind. 
If the `default: true` selector matches more than one service of the same kind, Perses
already handles it, it keeps the first one and logs an error for the rest. So no extra logic should be needed for it.

if the there is no `default` property in the config, then it means the datasource should be created with `default: false`.

gfierkasap:
Resolves #4351 

# Description

This PR adds a `default` flag to datasource discovery configuration, allowing administrators to designate discovered datasources as defaults for their plugin kind. This prevents users from repeatedly selecting the same datasource during dashboard creation at organizational scale.


Nexucis:
from my point of view, discovery will be used to discover multiple datasource at the same time. So, setting all of them as default won't work because you can only have one datasource as default per type of datasource. 

Perhaps we need a more complete way to designate which datasource should be the default one ? For example a label or something k8s way to select one datasource through others. What do you think ?

gfierkasap:
> from my point of view, discovery will be used to discovery multiple datasource at the same time. So, setting all of them as default won't work because you can only have one datasource as default per type of datasource.
> 
> Perhaps we need a more complete way to designate which datasource should be the default one ? For example a label or something k8s way to select one datasource through others. What do you think ?

AFAIK the situation where all k8s datasources would be marked as default would be handled by currently existing Perses logic, basically taking the first found as default, and logging an error (as per @ibakshay issue description). In any case, I would consider a situation like that a configuration error on the user side, and unless we want to change this default behavior I don't see how this PR changes anything. Or am I missing something?

On another hand, adding a field in the config that would define a pod tag/annotation, which would mark the default sources, at first glance seems like a nice touch (although I haven't given any thought just yet about how much effort it would take to implement it, and whether it wouldn't end up being more complicated than it initially seems). @ibakshay what do you think?

ibakshay:
> from my point of view, discovery will be used to discover multiple datasource at the same time. So, setting all of them as default won't work because you can only have one datasource as default per type of datasource.
> 
> Perhaps we need a more complete way to designate which datasource should be the default one ? For example a label or something k8s way to select one datasource through others. What do you think ?

Hi @Nexucis, 
I wrote my reasoning behind this [here](https://github.com/perses/perses/issues/4351#issue-5121134983). We already log if there are more than 1 default datasources  per datasource kind and from my point of view it is totally up to the user to maintain only one default datasource per datasource kind. You can check how the config with default datasource discovery will look like as well in the [issue description here](https://github.com/perses/perses/issues/4351#issue-5121134983).

>Perhaps we need a more complete way to designate which datasource should be the default one ? For example a label or something k8s way to select one datasource through others. What do you think ?

However, I am also open to your suggestion. Could you please give us an example of how the default datasource discovery config will look like with the label selector? :)

Nexucis:
Coming back to this PR, I think this struct will be more flexible.

And I think even if you have multiple datasources discovery, it might be better to take only the first one as the default one.

```go
	// Default is used to indicate that the discovered resource should be used as the default datasource for the plugin kind.
	// If multiple resources are discovered, the first one will be used as the default datasource.
	// Labels and Annotations can be used to filter even more the list of resource discovered to select only one resource to use as the default datasource for the plugin kind.
	// The rest of the discovered resources will be used to create additional datasources for the plugin kind.
	Default struct {
		Enable      bool              `json:"enable,omitempty" yaml:"enable,omitempty"`
		Labels      map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
		Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
	} `json:"default,omitempty" yaml:"default,omitempty"`
```

gfierkasap:
> Coming back to this PR, I think this struct will be more flexible.
> 
> And I think even if you have multiple datasources discovery, it might be better to take only the first one as the default one.
> 
> ```go
> 	// Default is used to indicate that the discovered resource should be used as the default datasource for the plugin kind.
> 	// If multiple resources are discovered, the first one will be used as the default datasource.
> 	// Labels and Annotations can be used to filter even more the list of resource discovered to select only one resource to use as the default datasource for the plugin kind.
> 	// The rest of the discovered resources will be used to create additional datasources for the plugin kind.
> 	Default struct {
> 		Enable      bool              `json:"enable,omitempty" yaml:"enable,omitempty"`
> 		Labels      map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
> 		Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
> 	} `json:"default,omitempty" yaml:"default,omitempty"`
> ```

Hi @Nexucis 

It took me a while to figure out and implement this way of selecting the default source. Below I've added a small flow chart of how it works right now:

```

  Execute()                           ← kubernetes.go
    │
    ├─ discover()                     ← fetches all matching K8s resources → []*GlobalDatasource
    │
    ├─ resolveDefaultName()           ← kubernetes.go                                           
    │     if Default.Enable == false → return ""
    │     else:                                      
    │       List Services or Pods using Default.Labels as label selector
    │       iterate results, client-side filter by Default.Annotations
    │       return "namespace.name" of first match  (or "" if none)     
    │                                                                                                                                                                                                                                                                        
    └─ svc.Apply(result, defaultName) ← service.go                                                                                                                                                                                                                           
          for each entity:                                                                                                                                                                                                                                                   
            entity.Spec.Default = (defaultName != "" && entity.Name == defaultName)                                                                                                                                                                                          
            Create or Update in DB                                                                                                                                                                                                                                           
                                      
```

Let me know if this is what you had in mind.

internal/api/discovery/kubernetes/kubernetes.go
```diff
@@ -108,10 +111,56 @@ func (d *discovery) Execute(_ context.Context, _ context.CancelFunc) error {
 		logrus.Errorf("failed to execute kube discovery %q: %v", d.name, err)
 		return nil
 	}
-	d.svc.Apply(result)
+	defaultName, err := d.resolveDefaultName()
```
Nexucis: instead of doing that, I would suggest to actually filter / found during the discovery which datasource would be the default one.

It will avoid another call to the kubernetes API and also the additional labels and annotations used to filter are applied after the one used for the discovery. It's not meant to be used separately.

Nexucis:
Hi @gfierkasap

Thank you for taking care about these changes. I have added a comment where I am trying to explain a bit more what I have in mind. Hopefully it will simplify a bit the code and you won't have to change the apply function.

Let me know if you need more explanations. A quick call is also possible if you prefer.

internal/api/discovery/kubernetes/kubernetes.go
gfierkasap: I was wondering whether a separate api call would be the best option; in this case I'll revert the most recent changes. I think I get the idea, but to be on the safe side, let me walk you through it:

1. introducing a wrapper struct around `GlobalDatasource` that will hold the labels and annotations, something like: 
```go
type discoveredDatasource struct {                                                                                                                                                                                                                                           
      datasource  *v1.GlobalDatasource                                                                                                                                                                                                                                       
      labels      map[string]string                                                                                                                                                                                                                                          
      annotations map[string]string                                                                                                                                                                                                                                          
  }
```
2. `discovery()`  would additionally populate the labels and annotations and return `[]*v1.discoveredDatasource` instead of `[]*v1.GlobalDatasource`
3. Implement filtering in `Execute()`

Does this approach make sense?

Nexucis: yes that makes sense !

gfierkasap: Awesome :) I've implemented the changes mentioned above, let me know what do you think about them. Just FYI, I'll be off for a few days and should be back on tuesday, so please be patient, I will respond as soon as I'm back :)

internal/api/discovery/kubernetes/kubernetes.go
```diff
@@ -103,15 +111,55 @@ func (d *discovery) Execute(_ context.Context, _ context.CancelFunc) error {
...
+		entities = append(entities, r.datasource)
+	}
+	d.svc.Apply(entities, defaultName)
 	return nil
 }
 
+// resolveDefaultName returns the name of the first discovered datasource whose
+// labels and annotations match Default.Labels and Default.Annotations.
+// Returns empty string if Default.Enable is false or no match is found.
+func (d *discovery) resolveDefaultName(resources []*discoveredDatasource) string {
```
Nexucis, line 131: instead of returning the name of the datasource, why not setting directly the boolean default when you find the datasource ?

Something like that:

```go
func (d *discovery) setDefaultDatasource(resources []*discoveredDatasource) {
	if !d.cfg.Default.Enable {
		return
	}
	for _, r := range resources {
		if matchesLabels(r.labels, d.cfg.Default.Labels) && matchesAnnotations(r.annotations, d.cfg.Default.Annotations) {
			r.datasource.spec.default = true
			return
		}
	}
	return
}
```

Like that you don't need to modify the function `apply`

gfierkasap: That's a good point, at a glance this makes a lot more sense that what I did before. I'll implement these changes, and push them asap

internal/api/discovery/kubernetes/kubernetes_test.go
```diff
@@ -59,3 +61,89 @@ func TestBuildLabelSelector(t *testing.T) {
...
+		},
+	}
+	resources := []*discoveredDatasource{
+		newDiscoveredDatasource("ns.prometheus", map[string]string{"app": "prometheus"}, nil),
+	}
+	d.setDefaultDatasource(resources)
+	assert.False(t, resources[0].datasource.Spec.Default)
+}
+
+func TestSetDefaultDatasource_NoMatch(t *testing.T) {
```
Nexucis: can you factorise all these tests in a single one with a struct containing the different cases ?

Nexucis:
This is almost good. I have added a last comment. Once fixed, we will be good to merge this one :).

Nexucis:
Also can you modify the doc to reflect this new configuration ?

Nexucis (approved):
awesome ! Thank you @gfierkasap !

Nexucis:
arg @gfierkasap can you format the code to unlock the CI. I will be able to merge this PR once this is done. Thank you !

gfierkasap:
> arg @gfierkasap can you format the code to unlock the CI. I will be able to merge this PR once this is done. Thank you !

@Nexucis just spotted that the checkformat failed; pushed the fix commit rn :)

Nexucis (approved):
perfect, thanks !

PR #4369, opened 2026-08-20
[IGNORE] Move from jest to vitest

Nexucis:
I have used AI for this migration


ui/package.json
```diff
@@ -58,7 +54,8 @@
     "ts-node": "^10.9.2",
```
Nexucis: perhaps `ts-node` is not needed anymore

Nexucis: I have removed it, it does not seem it breaks the build or the load of the plugins

ui/vitest.shared.ts
```diff
@@ -0,0 +1,40 @@
...
+export default defineConfig({
+  plugins: [react()],
+  resolve: {
+    alias: [
+      // Use polyfill for jsdom environment
+      { find: 'use-resize-observer', replacement: 'use-resize-observer/polyfilled' },
+
+      // Tell Vitest where other Perses packages live since it doesn't know about project references.
+      // Exclude `spec` and `dashboards` since they live outside of the perses/ui workspace.
+      { find: /^@perses-dev\/(?!spec|dashboards|components|client)(.*)$/, replacement: resolve(__dirname, '$1/src') },
```
Gladorme: Now, all perses package live outside of perses/ui, no?

I wonder if this conf is still usefull

ui/package.json
```diff
@@ -30,22 +30,18 @@
     "internal-utils"
   ],
   "devDependencies": {
-    "@swc/cli": "^0.8.1",
-    "@swc/core": "^1.15.33",
-    "@swc/jest": "^0.2.39",
     "@testing-library/dom": "^10.1.0",
-    "@testing-library/jest-dom": "^5.16.4",
+    "@testing-library/jest-dom": "^6.6.3",
```
jgbernalp, line 34: Do we still need this?

Nexucis: this is used actually in `src/components/DashboardList/NameCell.test.tsx` for example

ui/vitest.shared.ts
Nexucis: removed, it still works

ui/package.json
Nexucis: i tried to remove and these tests are failing 😐

jgbernalp: ok, I checked and is needed for some matchers.

Gladorme (approved):

PR #4374, opened 2026-08-21
[BUGFIX] Group changelog entries by pull request

vprashar2929:
# Description

This commit addresses the issue where a pull request containing several commits of the same type were listed as individual entries in the changelog. It now uses the PR title and number as a single changelog entry.

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/scripts/pkg/changelog/changelog.go
+++ b/scripts/pkg/changelog/changelog.go
@@ -35,6 +35,12 @@ const (
 	doc            = "DOC"
 )
 
+const (
+	gitLogFieldSeparator  = "\x00"
+	gitLogRecordSeparator = "\x1e"
+	gitLogFormat          = "--pretty=format:%h%x00%P%x00%s%x00%b%x1e"
+)
+
 // kind represents the type of change.
 type kind int
 
@@ -47,6 +53,13 @@ const (
 	KindToBeIgnored
 	kindDoc
 )
+
+type gitLogEntry struct {
+	hash    string
+	parents string
+	subject string
+	body    string
+}
 
 func getStringInBetweenTwoString(str string, startS string, endS string) (result string, found bool) {
 	s := strings.Index(str, startS)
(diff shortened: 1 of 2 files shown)


Nexucis:
while I am not necessary agains this change, I am wondering why you are not squashing the commit instead ? So you have one commit per pull request ?

Also Prometheus got a new approach to handle the changelog. It is requested to have a line in the Pull request description to asses what would be the entry in the changelog. Perhaps this can be handled more easily than reading all commits.
No opinion currently on that, but I think we should reconsider how to manage the commits / changelog to be able to generate it properly. Perhaps by using the semantic commit ?

vprashar2929:
> while I am not necessary agains this change, I am wondering why you are not squashing the commit instead ? So you have one commit per pull request ?
> 
> Also Prometheus got a new approach to handle the changelog. It is requested to have a line in the Pull request description to asses what would be the entry in the changelog. Perhaps this can be handled more easily than reading all commits. No opinion currently on that, but I think we should reconsider how to manage the commits / changelog to be able to generate it properly. Perhaps by using the semantic commit ?



> while I am not necessary agains this change, I am wondering why you are not squashing the commit instead ? So you have one commit per pull request ?
> 
> Also Prometheus got a new approach to handle the changelog. It is requested to have a line in the Pull request description to asses what would be the entry in the changelog. Perhaps this can be handled more easily than reading all commits. No opinion currently on that, but I think we should reconsider how to manage the commits / changelog to be able to generate it properly. Perhaps by using the semantic commit ?

@Nexucis You are right squashing would avoid it. Since the operator uses the changelog package from here and it's PR are currently merged with merge commits which is why this change was made. We could enforce squash merge commits in operator as well. @slashpai @jgbernalp Thoughts? 

The prometheus approach also sounds interesting and may give us a cleaner source for release notes than trying to reconstruct from git history. I am happy to adjust the change once we agree on the preferred direction

slashpai:
We added [commit lint checker](https://github.com/perses/perses-operator/blob/main/.github/workflows/commit-lint.yaml) recently to operator that should also help to reduce some noise while generating changelog. But I think having the option to group by pull request is good as sometimes we might need to split commits in same PR where squash may not be possible

jgbernalp:
IMO we should squash the commits into one so it generates a single entry in the changelog and it addresses a single fix or feature. PRs with multiple commits with the same or different messages should be avoided as we risk to mix different fixes into one.

vprashar2929:
> IMO we should squash the commits into one so it generates a single entry in the changelog and it addresses a single fix or feature. PRs with multiple commits with the same or different messages should be avoided as we risk to mix different fixes into one.

@jgbernalp I agree we should do squash merging for each PR. If we can enforce that in operator I think we are good to close this PR.

jgbernalp:
I configured a merge queue with squash and merge for the operator

PR #4381, opened 2026-08-25
[FEATURE] Enable decompress middleware for all route

Nexucis:
This PR enables the middleware to decompress body that would have been compressed. The header `Content-Encoding` needs to be set with the value `gzip` to use this functionality.


celian-garcia (approved):
That's it ? So if it will adapt automatically according to the Content-Encoding ?

Nexucis:
> That's it ? So if it will adapt automatically according to the Content-Encoding ?

yes: https://echo.labstack.com/middleware/decompress/

PR #4386, opened 2026-08-27
[FEATURE] (go-sdk): add dashboard.Timezone option for Spec.timezone

colivi:
# Description

Implementing dahboards with Dashboard-as-Code solution, I could not set spec.timezone via the Go SDK, even though the field already exists on github.com/perses/spec (dashboard.Spec.Timezone) and is part of the UI timezone resolution hierarchy:

dashboard.spec.timezone > user preference > server default > browser local

This PR adds:

- dashboard.Timezone(tz string) Option in go-sdk/dashboard
- Validation aligned with frontend preferences: empty, "local", or a valid IANA name (time.LoadLocation)
- Unit tests for UTC, local, and invalid values
- Docs update in docs/concepts/timezone.md (hierarchy + DaC example)

**Example:**

```go
builder, err := dashboard.New("MyDashboard",
    dashboard.ProjectName("myProject"),
    dashboard.Timezone("UTC"),
)
```

### DIFF
--- a/go-sdk/dashboard/options_timezone_test.go
+++ b/go-sdk/dashboard/options_timezone_test.go
@@ -1,0 +1,86 @@
+// Copyright The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dashboard
+
+import (
+	"encoding/json"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


go-sdk/dashboard/options_timezone_test.go
```diff
@@ -0,0 +1,69 @@
...
+	var m map[string]any
+	if err := json.Unmarshal(raw, &m); err != nil {
+		t.Fatal(err)
+	}
+	spec, _ := m["spec"].(map[string]any)
+	if spec == nil {
+		t.Fatalf("missing spec in %s", raw)
+	}
+	if spec["timezone"] != "UTC" {
+		t.Fatalf("json timezone = %v, want UTC", spec["timezone"])
```
Nexucis: can you contact these tests into a single one please ?

You can create a struct that will contains the various tests data. There are many examples in the code on how to do it. Like this one: https://github.com/perses/perses/blob/main/internal/api/index/index_test.go#L251

Nexucis (changes requested):
Thanks @colivi for the improvement !

Do you mind to signoff your commit please to make DCO happy ?

Once my comment is fixed we should be good to merge this PR

Nexucis (approved):
thanks !

PR #4387, opened 2026-08-27
[BUGFIX] handle semver comparission with plugins that do not include 'v' as version prefix

jgbernalp:
# Description

Plugins that do not include the prefix v in their version will lead to erroneous semver comparison in the backend. This caused the issue that a perses backend running with multiple plugin versions wont load correctly the plugin assets.

# Checklist


celian-garcia (approved):

Nexucis (approved):
oupsi oupsi. I am wondering why this is even working before

PR #4390, opened 2026-08-29
fix: correct 'not knowm' typo in the same_site error message

vaibhav8a:
The error returned for an unrecognised `cookie same_site` value reads `mode not knowm`. Corrects it to `known`.

User-facing error text only, no behaviour change.


Nexucis (approved):
good catch ! Thanks !

PR #4394, opened 2026-08-30
[DOC] Fix typos in troubleshooting doc, client error message and comments

mihhhir08:
Found a few typos while reading through the repo. The troubleshooting doc for plain HTTP datasources quotes the browser error and spells the Prometheus endpoint as guery_range, it should be query_range. In pkg/client/perseshttp/request.go the returned error says "something horrible occured" and everywhere else in the codebase it is spelled occurred, so I made it consistent, and the comment right above it said "append" where it meant "happened". Last one is in internal/api/authorization/authorization.go where a doc comment says "dependant" and "theses decisions". All of these are text only, no logic changed and no test depends on the strings.

### DIFF
--- a/docs/troubleshooting-tls-plain-http-datasource.md
+++ b/docs/troubleshooting-tls-plain-http-datasource.md
@@ -8,7 +8,7 @@ When you try now to create some dashboards you will see a "Load failed" error me
 
 ## Diagnostics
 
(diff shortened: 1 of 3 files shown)


Nexucis (approved):
good catch ! Thanks !

PR #4399, opened 2026-08-31
[BUGFIX] Fix e2e test for moving panel to different group after Layout tab move

adrianSepiol:


Nexucis (approved):
awesome @adrianSepiol, thank you very much 🙏

PR #4404, opened 2026-09-01
[FEATURE]: implement oauthPassThru, handles OIDC refreshToken as well

kubehubdev:
# Description

Implement a feature that is well established in Grafana: https://grafana.com/developers/plugin-tools/how-to-guides/data-source-plugins/add-authentication-for-data-source-plugins#forward-oauth-identity-for-the-logged-in-user 

it use current user's OIDC token to authenticate against datasources (datasources have to config secret to opt-in).

```yaml
kind: GlobalDatasource
metadata:
  name: kubehub-loki
spec:
  default: true
  plugin:
    kind: LokiDatasource
    spec:
      proxy:
        kind: HTTPProxy
        spec:
          allowedEndpoints: []
          oauthPassThrough: true
          url: https://{{OAuthProtectedLokiURL}}/
```

# Screenshots

N/A, backend changes only.

# Checklist

## UI Changes


internal/api/impl/proxy/proxy.go
```diff
@@ -291,7 +291,7 @@ func (h *httpProxy) serve(c echo.Context) error {
 
 	if err := h.prepareRequest(c); err != nil {
 		h.logWithDefaultEntry().WithError(err).Error("unable to prepare the HTTP request")
-		return apiinterface.InternalError
+		return err
```
Nexucis, line 300: why did you change the error returned ?

internal/api/impl/proxy/proxy.go
```diff
@@ -366,6 +366,24 @@ func (h *httpProxy) setupAuthentication(req *http.Request) error {
 	if h.secret == nil {
 		return nil
 	}
+	if h.secret.OAuthPassThru {
+		var oidcToken string
+		for _, cookie := range req.Cookies() {
```
Nexucis: If you add in the parameter the context, this loop can be then simplified by getting the cookie you wanted directly.

```go
func (h *httpProxy) setupAuthentication(c echo.Context, req *http.Request) error {
		[ ... ]
		oidcCookie, err := c.Cookie(crypto.CookieKeyOIDCToken)
		if errors.Is(err, http.ErrNoCookie) {
				// TODO : manage the error
		}
		req.Header.Set(echo.HeaderAuthorization, fmt.Sprintf("Bearer %s", oidcCookie.Value))
		return nil
}

File: pkg/model/api/v1/secret.go [hunk 3]
```diff
@@ -29,6 +29,10 @@ type SecretSpec struct {
 	OAuth *secret.OAuth `json:"oauth,omitempty" yaml:"oauth,omitempty"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig *secret.TLSConfig `json:"tlsConfig,omitempty" yaml:"tlsConfig,omitempty"`
+	// OAuthPassThru when true forwards the incoming user's OAuth/OIDC access token to the upstream datasource
+	// using an Authorization: Bearer header. This is only applicable for proxy-type datasources (HTTP proxy and SQL proxy).
+	// When enabled, the user's token is used instead of any static auth configured in basicAuth, authorization, or oauth.
+	OAuthPassThru bool `json:"oauthPassThru,omitempty" yaml:"oauthPassThru,omitempty"`
```
Nexucis: why not writing like that : `OAuthPassThrough` ?

Nexucis (commented):
Thanks for the improvement, this is an interesting way to forward the OIDC token.

While I am not against this feature (on the contrary), I am wondering about the utility to support the refresh token in Perses directly. Just curious about your point of view.

And I have added few comments, but it looks already good.

Nexucis:
this should solve this discussion: https://github.com/perses/perses/discussions/2942

pkg/model/api/v1/secret.go
kubehubdev: just use same property name from grafana, that `Thru` does annoy me, I can rename it to OAuthPassThrough, really not worth to save few characters.

internal/api/impl/proxy/proxy.go
kubehubdev: I think I wanted to surface more info to the UI, when login as native user (username/password), this thing isn't going to work. 

let me get this fixed in a proper way, while I can surface the error to user.

internal/api/impl/proxy/proxy.go
kubehubdev: thanks a lot! let me look into this.

internal/api/impl/proxy/proxy.go
kubehubdev: introduced UserVisibleError, so I return this error in setupAuthentication, and condition with if errors.Is, then return 400 error with actionable message.

internal/api/impl/proxy/proxy.go
kubehubdev: fixed, this does simplified the code.

pkg/model/api/v1/secret.go
kubehubdev: renamed to `OAuthPassThrough`

kubehubdev:
tested heavily, and fix another issue where perses token expiration is longer than oidcToken, so have to decouple oidcToken refresh from persesToken refresh.

internal/api/impl/auth/oauth.go
```diff
@@ -479,6 +479,16 @@ func (e *oAuthEndpoint) performUserSync(userInfo externalUserInfo, persistedClai
 		return nil, err
 	}
 
+	// Store the upstream OIDC token and refresh token for oauthPassThru datasource proxy support
```
Nexucis:
```suggestion
	// Store the upstream OIDC token and refresh token for oauthPassThrough datasource proxy support
```

pkg/model/api/v1/secret.go
```diff
@@ -71,6 +75,9 @@ func (s *SecretSpec) validate() error {
 	if nbAuthConfigured > 1 {
 		return fmt.Errorf("basicAuth, authorization and oauth are mutually exclusive, use one of them")
 	}
+	if s.OAuthPassThrough && nbAuthConfigured > 0 {
+		return fmt.Errorf("oauthPassThru cannot be used together with basicAuth, authorization, or oauth")
```
Nexucis:
```suggestion
		return fmt.Errorf("oauthPassThrough cannot be used together with basicAuth, authorization, or oauth")
```

pkg/model/api/v1/secret_test.go
```diff
@@ -80,3 +80,56 @@ func TestUnmarshalJSONSecretSpecMutuallyExclusiveAuth(t *testing.T) {
...
+	for _, test := range testSuite {
+		t.Run(test.title, func(t *testing.T) {
+			result := &SecretSpec{}
+			err := json.Unmarshal([]byte(test.jason), result)
+			if test.wantErr {
+				assert.Error(t, err)
+				assert.Equal(t, test.errMsg, err.Error())
+			} else {
+				assert.NoError(t, err)
+				if test.jason == `{"oauthPassThru": true}` {
```
Nexucis: tests need to be updated with the new field

internal/api/impl/proxy/proxy.go
```diff
@@ -560,6 +593,10 @@ func (s *sqlProxy) setupAuthentication() error {
 		return nil
 	}
 
+	if s.secret.OAuthPassThrough {
+		logrus.Warnf("oauthPassThru is configured for SQL datasource %s but is not supported for SQL proxy; only HTTP proxy supports passing user tokens", s.name)
```
Nexucis:
```suggestion
		logrus.Warnf("oauthPassThrough is configured for SQL datasource %s but is not supported for SQL proxy; only HTTP proxy supports passing user tokens", s.name)
```

internal/api/impl/proxy/proxy.go
```diff
@@ -359,13 +368,37 @@ func (h *httpProxy) prepareRequest(c echo.Context) error {
...
+			// before giving up.
+			if h.tokenRefresher != nil {
+				if _, refreshErr := c.Cookie(crypto.CookieKeyOIDCRefreshToken); refreshErr == nil {
+					h.tokenRefresher(c)
+					// Re-read the OIDC token cookie after the refresh attempt.
+					oidcCookie, err = c.Cookie(crypto.CookieKeyOIDCToken)
+				}
+			}
+			if errors.Is(err, http.ErrNoCookie) {
+				return apiinterface.HandleUserVisibleError(fmt.Sprintf(
```
Nexucis: I think you don't need to create another type of error. You can use the function to handle a 400 error instead. This will be enough

internal/api/impl/proxy/proxy.go
```diff
@@ -359,13 +368,37 @@ func (h *httpProxy) prepareRequest(c echo.Context) error {
...
+			// before giving up.
+			if h.tokenRefresher != nil {
+				if _, refreshErr := c.Cookie(crypto.CookieKeyOIDCRefreshToken); refreshErr == nil {
+					h.tokenRefresher(c)
+					// Re-read the OIDC token cookie after the refresh attempt.
+					oidcCookie, err = c.Cookie(crypto.CookieKeyOIDCToken)
+				}
+			}
+			if errors.Is(err, http.ErrNoCookie) {
+				return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf(
```
Nexucis:
```suggestion
				return apiinterface.HandleBadRequestError(fmt.Sprintf(
```

I prefere that way since we have this kind of function that avoid having to define manually the HTTP error code

then you will need to change the type of the error check at ligne 299

Nexucis:
I am wondering if this parameter `OauthPassThrough` should not be here https://github.com/perses/spec/blob/main/go/datasource/proxy/http/http.go#L73 instead. I am not sure it makes sense this is part of the secret.

Do you have any opinion on that @kubehubdev ?

Nexucis:
> I am wondering if this parameter `OauthPassThrough` should not be here https://github.com/perses/spec/blob/main/go/datasource/proxy/http/http.go#L73 instead. I am not sure it makes sense this is part of the secret.
> 
> Do you have any opinion on that @kubehubdev ?

@celian-garcia if you can give me your opinion on that also, that would be cool. Thank you !

kubehubdev:
> I am wondering if this parameter `OauthPassThrough` should not be here https://github.com/perses/spec/blob/main/go/datasource/proxy/http/http.go#L73 instead. I am not sure it makes sense this is part of the secret.
> 
> Do you have any opinion on that @kubehubdev ?

I initially track through a static JWT token to this secret, and then naturally implement it in Secret object. 

Now in overall picture, this meas the datasource object with OAuthPassthrough doesn't need a secret, we can attribute this to Datasource instead of Secret (without secret).

I am ok to move o datasource object, the datasource is self sufficient, and we don't need a secret object that doesn't have real secret.  @celian-garcia  let me know what do you think, I can make the change if we all aligned.

kubehubdev:
@Nexucis  if you are aligned on attach this property to datasource, I can start to change the code.

Nexucis:
Hello @kubehubdev,

I have checked with @celian-garcia and we agree to put this boolean into the HTTP proxy definition.

Can you open a PR to the `spec` and add this boolean please ?

Nexucis:
And you will need to update also the proxy form in the shared repository. If you can do all of that, it will be great ! I know moving this boolean is harder than keeping it in the secret, but like you, I believe it will be cleaner.

We also don't have to check anything for the SQL proxy because the boolean won't appear.

kubehubdev:
thanks, will follow that direction.

kubehubdev:
created https://github.com/perses/spec/pull/99/commits

kubehubdev:
@Nexucis  I need help get a  release tag github.com/perses/spec@v0.3.0-beta.8 .

Nexucis:
> @Nexucis I need help get a release tag github.com/perses/spec@v0.3.0-beta.8 .

yes I am on it :)

Nexucis:
@kubehubdev done :) v0.3.0-beta.8 is available

kubehubdev:
I think I need plugins PR as well:   https://github.com/perses/plugins/pull/805
hit this during my test:
```
time="2026-09-09T19:51:18Z" level=error msg="unable to create the \"GlobalDatasource\" \"kubehub-mimir\"" func="github.com/perses/perses/internal/api/provisioning.(*provisioningService).applyEntity" file="/workspace/perses/perses/internal/api/provisioning/service.go:96" error="bad request: invalid datasource kubehub-mimir: spec: 2 errors in empty disjunction:\nspec.proxy: field not allowed:\n    1:40\nspec.proxy.spec.oauthPassThrough: field not allowed:\n    1:500\n"
```

kubehubdev:
@Nexucis I think I need some dev instruction on how can I test this locally with version mixing across repo, spec/plugings/perses, maybe shared as well.

Nexucis:
you may need these instructions https://github.com/perses/shared#linking-with-local-projects @kubehubdev.

The CLI can help you also to run the plugin in a dev mode.

And perhaps you will have to open a PR in shared to modify the form of the proxy to include this new boolean to be able to have it in the UI

Nexucis:
but I think you can already modify this PR so we can merge it. You will have to update the spec dependencies.

internal/api/impl/proxy/proxy.go
```diff
@@ -384,10 +393,17 @@ func (h *httpProxy) filterHeaders(headers http.Header) {
...
+func (h *httpProxy) setupAuthentication(c echo.Context) error {
+	if h.config.OauthPassthrough {
+		return h.setupOAuthPassthrough(c)
+	}
 	if h.secret == nil {
 		return nil
 	}
+
+	req := c.Request()
+	h.filterHeaders(req.Header)
```
Nexucis: you should not move this line here because we want to filter the header regardless there is an authentication setup to make

internal/api/impl/proxy/proxy.go
```diff
@@ -416,6 +432,32 @@ func (h *httpProxy) setupAuthentication(req *http.Request) error {
...
+		// before giving up.
+		if h.tokenRefresher != nil {
+			if _, refreshErr := c.Cookie(crypto.CookieKeyOIDCRefreshToken); refreshErr == nil {
+				h.tokenRefresher(c)
+				// Re-read the OIDC token cookie after the refresh attempt.
+				oidcCookie, err = c.Cookie(crypto.CookieKeyOIDCToken)
+			}
+		}
+		if errors.Is(err, http.ErrNoCookie) {
+			return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf(
```
Nexucis: here you can use our function to handle such error:

```go
return apiinterface.HandleBadRequestError(fmt.Sprintf(	"you are querying datasource %q which is configured to use OAuthPassThrough, but no OAuth token is available in this session; try logging out and logging in again with the correct authentication provider",
h.datasourceName))

File: internal/api/impl/proxy/proxy.go [hunk 12]
```diff
@@ -292,6 +297,10 @@ func (h *httpProxy) serve(c echo.Context) error {
 
 	if err := h.prepareRequest(c); err != nil {
 		h.logWithDefaultEntry().WithError(err).Error("unable to prepare the HTTP request")
+		var echoHttpErr *echo.HTTPError
+		if errors.As(err, &echoHttpErr) {
+			return echoHttpErr
+		}
 		return apiinterface.InternalError
```
Nexucis: it will mean also instead of making a difference between the echo error and the rest, you can just return the error directly.

We have a middleware that will return an internal error for every error not managed, so simply returning the error will do exactly what you want

kubehubdev:
thanks, addressed all comments, and verified I can see the 400 message:
Screenshot from 2026-09-11 01-44-27

internal/api/impl/proxy/proxy.go
Nexucis:
```suggestion
		return err
```

Nexucis: one last thing and we should be good. You need to change the error returned here to avoid filtering. This is quite weird you have the correct error returned in the UI while the error returned currently is a `500 internal error`

Nexucis:
to fix the ci @kubehubdev you need to run `go mod tidy` in `go-sdk/test` and in `internal/cli/cmd/dac/build/testdata/go`

Nexucis:
you can ignore the error in the windows tests. It seems I was too quick when changing the setup of postgresql on windows

Nexucis (approved):
thank you @kubehubdev for your contribution. It looks good to me !

kubehubdev:
thanks a lot for helping me going through this PR.

PR #4420, opened 2026-09-07
[FEATURE] Add infinite scrolling option to project dashboards list (#4418)

### LINKED ISSUE #4418: Use infinite scrolling in projects stack view
## Description

Based on our user feedback, when opening a project, for example, projects/myproject, the stack view currently displays items using pagination, with 25 rows shown by default. Pagination works, but the preferred, default behavior is to load additional rows automatically as the user scrolls down.

Image
 

## Current behavior

- The view shows 25 rows by default
-  User needs to use pagination to see additional items

## Expected behavior

- Keep loading additional rows automatically when the user reaches the bottom of the current list
- Infinite scrolling should be the default behavior

> Infinite scrolling is a web design pattern that loads new content continuously as you scroll down a page, removing the need for page numbers or "next page" buttons

## Note

I'd say, we keep the current behavior and add the ability to switch between the normal pagination view and infinite scrolling.

thisis-manan:
# Description

Adds a "Pages / Infinite" segmented toggle above the project dashboards stack view. Pagination remains the default; selecting "Infinite" removes the page controls and renders all rows in the Table's virtualized scroll area. The preference is persisted in localStorage so it survives reloads.

Closes #4418

# Screenshots

https://github.com/user-attachments/assets/25f654db-685e-499e-b014-b0eef5e411af

# Checklist

- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention.

## UI Changes

- [x] Code follows the UI guidelines.
- [x] E2E tests are stable and unlikely to be flaky (default behavior unchanged; no selector collisions).

### DIFF
--- a/ui/app/src/components/DashboardList/DashboardTreeList.test.tsx
+++ b/ui/app/src/components/DashboardList/DashboardTreeList.test.tsx
@@ -21,25 +21,36 @@ import type { DashboardListRow } from './DashboardList';
 import DashboardTreeList from './DashboardTreeList';
 
 interface MockTableProps {
-  pagination: { pageIndex: number; pageSize: number };
-  onPaginationChange: (pagination: { pageIndex: number; pageSize: number }) => void;
+  pagination?: { pageIndex: number; pageSize: number };
+  onPaginationChange?: (pagination: { pageIndex: number; pageSize: number }) => void;
 }
 
 function MockTable({ pagination, onPaginationChange }: MockTableProps): ReactElement {
-  const showTenRows = useCallback((): void => onPaginationChange({ pageIndex: 0, pageSize: 10 }), [onPaginationChange]);
+  const showTenRows = useCallback(
+    (): void => onPaginationChange?.({ pageIndex: 0, pageSize: 10 }),
+    [onPaginationChange],
+  );
   return (
     <>
-      <span>Rows per page: {pagination.pageSize}</span>
-      <button type="button" onClick={showTenRows}>
-        Show 10 rows
-      </button>
+      {/* In infinite scrolling mode the Table receives no pagination. */}
+      <span>Rows per page: {pagination ? pagination.pageSize : 'all'}</span>
+      {onPaginationChange && (
+        <button type="button" onClick={showTenRows}>
+          Show 10 rows
+        </button>
+      )}
     </>
   );
 }
 
-vi.mock('@perses-dev/components', () => ({
-  Table: MockTable,
-}));
+vi.mock('@perses-dev/components', async () => {
+  const { useState } = await import('react');
+  return {
+    Table: MockTable,
+    // Back the persisted display mode with local component state for the test.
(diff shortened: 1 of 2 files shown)


thisis-manan:
A few design decisions I'd like maintainer input on:

1. Default is pagination (following the note in the issue), though the description says infinite should be default happy to flip if you prefer.
2. Scoped to the dashboards stack view the other project lists use MUI Community DataGrid (100-row pageSize cap), so infinite scroll there would need DataGridPro or a component swap.
3. Mode is persisted in localStorage; I can move it into UserPreferences if that's preferred.

Nexucis:
Hello @thisis-manan, thank you for opening this PR.

As you say yourself, we cannot apply this behaviour to every list / table. So I think we need to discuss furthermore about this new need.
From my perspective, either this infinity scroll is available for all list/ table, either it does not exist.

Happy to discuss about this opinion.

thisis-manan:
@Nexucis Yeah, fair point, that makes sense.

The tricky part is that the lists aren't all the same under the hood. The dashboards list uses our internal Table component which is virtualized, so infinite scroll is easy there. But the other lists (variables, datasources, secrets, roles, etc.) are on MUI's Community DataGrid, and that one caps pageSize at 100 and doesn't do infinite/lazy scroll at all  you'd need the Pro version for that. So I can't just flip it on everywhere.

The way to make it consistent would be to move those DataGrid lists over to the internal Table too. I don't mind doing that, but the Table's still marked experimental so I didn't want to go refactoring everything without checking with you first. Do you want to go that route, or did you have something else in mind? Happy to follow whatever makes sense for where you want the tables to go.

Nexucis:
> The way to make it consistent would be to move those DataGrid lists over to the internal Table too. I don't mind doing that, but the Table's still marked experimental so I didn't want to go refactoring everything without checking with you first. Do you want to go that route, or did you have something else in mind? Happy to follow whatever makes sense for where you want the tables to go.

On that I would need the opinion from @jgbernalp @Gladorme and @shahrokni.

So we can do what you are suggesting. Or we don't do anything actually and we drop the issue and the PR.

From my side I am not that convinced this is a game changer to have an infinite scroll on the table. Or perhaps it's something desired for plugin (like for displaying the logs or the timeseries).

Anyway I think we need more opinions to move forward. I cannot give you the right decision as it impacts deeply the projects

Nexucis:
Hi @thisis-manan,
I am sorry but after considering and further investigation about why the infinite scroll was requested, this feature is not necessary.

I have added more details about why we won't do it in the issue.
I am sorry you spent time on this feature already, but anyway it was useful you did it to show a part of the limitation of this feature.

Thank you for the time you spent to contribute to Perses. I hope you will continue with us :).

thisis-manan:
Thanks for looking into it and for the feedback!

No worries about closing the PR. I actually learned quite a bit from working on this, especially around the different table implementations and their limitations.

Appreciate the guidance, and I'll definitely look for another issue to work on!

Issue #4408, opened 2026-09-02
Document Annotations


AntoineThebaud:
The graph annotations have been implemented in the last release but we are lacking doc related to this feature, namely:
- mention the new plugin kind in the [Concept doc](https://perses.dev/perses/docs/concepts/plugin/).
- mention the new annotation plugin among the [prometheus plugins](https://perses.dev/plugins/docs/prometheus/)

PR #4427, opened 2026-09-08
[DOC] Document the Annotation plugin kind

7487:
# Description

`Annotation` has been a first-class plugin kind (`plugin.KindAnnotation`) since the annotation support released in #4040, and `percli plugin generate` already accepts it, but the documentation never mentions it.

- `docs/concepts/plugin.md`: add `Annotation` to the object types available as plugins, plus the matching capability bullet.
- `docs/plugins/creation.md`: add `Annotation` to the plugin types and to the `--plugin.type` values, and describe the layout of an annotation plugin (`schemas/annotations/<plugin-name>` and `src/annotations/<plugin-name>`), matching `internal/cli/cmd/plugin/generate/templates/annotations`.

This covers the first bullet of the issue. The second one (listing the annotation plugin among the Prometheus plugins) belongs to `perses/plugins`, so it is handled in perses/plugins#804.

Unrelated to this change, but noticed while editing that line: the `--plugin.type` list also advertises `TraceQuery`, `ProfileQuery` and `LogQuery`, which are not part of `availablePluginTypes` in `internal/cli/cmd/plugin/generate/generate.go` and are therefore rejected by `Validate()`. I left them untouched here — happy to remove them in a follow-up if the doc is the side that is wrong.

Part of #4408

# Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/concepts/plugin.md
+++ b/docs/concepts/plugin.md
@@ -6,6 +6,7 @@ In Perses, the following object types are available as plugins:
 - [Datasource](./datasource.md)
 - Query
 - [Variable](./variable.md)
+- Annotation
 - Explorer
 
 The goal with this is to eventually empower users to seamlessly enhance Perses' native capabilities through custom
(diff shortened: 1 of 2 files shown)


AntoineThebaud (approved):
Thanks!

PR #4432, opened 2026-09-09
[FEATURE] add configurable refresh interval

### LINKED ISSUE #4385: Refresh interval picker should be configurable
## Descriptions
As mentioned by our users, the available `refresh interval` list is limited to very low range numbers which may not be appropriate for some cases. Investigating the code based, it turned out that the corresponding control is fed by a list of limited `Duration Strings` 

> plugin-system\src\components\TimeRangeControls\TimeRangeControls.tsx 

```TypeScript

export const DEFAULT_REFRESH_INTERVAL_OPTIONS: TimeOption[] = [
  { value: { pastDuration: '0s' }, display: 'Off' },
  { value: { pastDuration: '5s' }, display: '5s' },
  { value: { pastDuration: '10s' }, display: '10s' },
  { value: { pastDuration: '15s' }, display: '15s' },
  { value: { pastDuration: '30s' }, display: '30s' },
  { value: { pastDuration: '60s' }, display: '1m' },
];

          <RefreshIntervalPicker
            timeOptions={DEFAULT_REFRESH_INTERVAL_OPTIONS}
            value={
              DEFAULT_REFRESH_INTERVAL_OPTIONS.some((i) => i.value.pastDuration === refreshInterval)
                ? refreshInterval
                : DEFAULT_REFRESH_INTERVAL_OPTIONS.find((i) => i.display === refreshInterval)?.value.pastDuration
            }
            onChange={handleRefreshIntervalChange}
            height={height}
          />
```

## Proposal
The list of `Duration Strings` should be configurable. The current behavior could be used as a fallback mechanism.

shahrokni:
Closes https://github.com/perses/perses/issues/4385

# Description

This feature adds `configurable refresh interval` by which the users could add their desired auto refresh intervals. 

# How to test 🧪 

You need to provide the proper (desired )config
For example, 

```yaml
frontend:
  enable_keyboard_shortcuts: true
  explorer:
    enable: true
  auto_refresh:
    disable: false
    options: 
    - "8s"
    - "16s"
```

To assure your config is picked you can add a simple console.log in `Config.tsx`
```javaScript
console.log('auto refresher options', data?.frontend.auto_refresh?.options);
```
image

The following screenshot depicts an `8 seconds` interval

image

Now You can also save the changes

image

If the config is not provided, the fallback mechanism returns the default list.
⚠️ I just added 63s for demonstration, and removed after my test

image

# Old Bug 🐛 

Please note that there is an old bug which is not relevant to this change and should be addressed separately. The following screenshot has been provided from the `demo.perses.dev`
Scenario: Set a refresh interval from URL which does not exist in the list. Consequently, the list select item would be empty

image

# Checklist

## UI Changes


docs/configuration/configuration.md
```diff
@@ -852,6 +852,8 @@ disable_zoom:  <bool> | default = false # Optional
 ```yaml
 # Allow you to disable dashboard auto-refresh (refresh interval picker hidden; refreshInterval and ?refresh= ignored)
 disable:  <bool> | default = false # Optional
+# Use duration format. The display will be computed automatically. Eg: "10s: will be display "5 seconds"
```
adrianSepiol: wouldn't 10s be "10 seconds"?

adrianSepiol (approved):

ui/app/src/context/Config.tsx
```diff
@@ -37,11 +38,23 @@ export function ConfigContextProvider(props: { children: React.ReactNode }): Rea
...
   );
 
+  const autoRefreshOptions = useMemo((): TimeOption[] | undefined => {
+    return data?.frontend.auto_refresh?.options?.map((option: DurationString) => {
+      const timeOption = buildRelativeTimeOption(option);
+      /**
+       * `Last` does not make sense for refresh interval. It should be either `Every` or string.empty
+       * TODO: .replace should be removed when the buildRelativeTimeOption is fixed in the shared package (In progress)
+       */
+      return { ...timeOption, display: timeOption.display.replace('Last', '') };
```
adrianSepiol: shouldn't it be 'Last ' so we get rid of space at the beginning as well?

adrianSepiol (approved):

ibakshay (approved):

pkg/model/api/config/frontend.go
```diff
@@ -75,7 +75,8 @@ type TimeRange struct {
 }
 
 type AutoRefresh struct {
-	Disable bool `json:"disable,omitempty" yaml:"disable,omitempty"`
+	Disable bool                    `json:"disable,omitempty" yaml:"disable,omitempty"`
+	Options []common.DurationString `json:"options,omitempty" yaml:"options,omitempty"`
```
celian-garcia, line 88: Isn't there some default computation missing there?

ui/app/src/context/Config.tsx
```diff
@@ -37,11 +38,23 @@ export function ConfigContextProvider(props: { children: React.ReactNode }): Rea
...
   );
 
+  const autoRefreshOptions = useMemo((): TimeOption[] | undefined => {
+    return data?.frontend.auto_refresh?.options?.map((option: DurationString) => {
+      const timeOption = buildRelativeTimeOption(option);
+      /**
+       * `Last` does not make sense for refresh interval. It should be either `Every` or string.empty
+       * TODO: .replace should be removed when the buildRelativeTimeOption is fixed in the shared package (In progress)
+       */
+      return { ...timeOption, display: timeOption.display.replace('Last ', '') };
```
celian-garcia, line 54: Can't you wait for shared fixed before? What is the PR corresponding ?

docs/configuration/configuration.md
```diff
@@ -852,6 +852,8 @@ disable_zoom:  <bool> | default = false # Optional
 ```yaml
 # Allow you to disable dashboard auto-refresh (refresh interval picker hidden; refreshInterval and ?refresh= ignored)
 disable:  <bool> | default = false # Optional
+# Use duration format. The display will be computed automatically. Eg: "5s: will be display "5 seconds"
+options: <duration[]> | default = [ "0s", "5s", "10s", "15s", "30s", "60s" ]
```
Nexucis, line 856: If you want to have a default value here, the backend needs to provide it.

For that the `AutoRefresh` struct needs to implement the function `Verify() error` where you can set a default value if none has been provided by the config

It will be nice to explain that the `0` value means it is `off`

ui/app/src/context/Config.tsx
shahrokni: No, let's not wait, a small .replace would not do any harm.

ui/app/src/context/Config.tsx
Nexucis, line 64: I don't get why this is part of the TimeRangeSettingProvider since it's not about the timeRange selector at all

Nexucis: I get it it is because it is currently implemented like that, but I don't think it makes sense this is mixed actually

shahrokni: As you said it has been like this. I propose to keep a separate change for the improvement.

ui/app/src/context/Config.tsx
```diff
@@ -37,11 +38,23 @@ export function ConfigContextProvider(props: { children: React.ReactNode }): Rea
...
   );
+
   const timeRangeOptions = useMemo(
     () => data?.frontend.time_range?.options?.map((option: DurationString) => buildRelativeTimeOption(option)),
     [data?.frontend.time_range?.options],
   );
 
+  const autoRefreshOptions = useMemo((): TimeOption[] | undefined => {
+    return data?.frontend.auto_refresh?.options?.map((option: DurationString) => {
+      const timeOption = buildRelativeTimeOption(option);
```
celian-garcia, line 49: Makes me think, how do you handle the Off option ? 

Image

Nexucis: it's the 0 value that means "off".

ui/app/src/context/Config.tsx
```diff
@@ -37,11 +38,23 @@ export function ConfigContextProvider(props: { children: React.ReactNode }): Rea
...
+      /**
+       * `Last` does not make sense for refresh interval. It should be either `Every` or string.empty
+       * TODO: .replace should be removed when the buildRelativeTimeOption is fixed in the shared package (In progress)
+       */
+      return { ...timeOption, display: timeOption.display.replace('Last ', '') };
+    });
+  }, [data?.frontend.auto_refresh?.options]);
```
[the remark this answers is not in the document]
Nexucis: seems legit indeed. Probably not the best elegant way to achieve it, but the issue at least seems legit

pkg/model/api/config/frontend.go
```diff
@@ -166,7 +183,7 @@ type Frontend struct {
 	// TimeRange contains the time range configuration for the dropdown
 	TimeRange *TimeRange `json:"time_range,omitempty" yaml:"time_range,omitempty"`
 	// AutoRefresh contains the auto-refresh configuration for dashboards
-	AutoRefresh *AutoRefresh `json:"auto_refresh,omitempty" yaml:"auto_refresh,omitempty"`
+	AutoRefresh AutoRefresh `json:"auto_refresh,omitzero" yaml:"auto_refresh,omitempty"`
```
celian-garcia: Ok for the remove of the pointer, to make sure your verify is called I guess.

But do you need this omitzero ? what does it mean already ? 
We mostly have yaml tag used for the configs not really json tags. I'd prefer to keep it aligned

celian-garcia (changes requested):

celian-garcia (approved):

PR #4437, opened 2026-09-10
[BUGFIX] default OIDC/OAuth redirect path to api_prefix, not router basename-relative root

### LINKED ISSUE #4436: Bug: Default redirect path `rd` does not include `api_prefix`, causing redirect to root path after OIDC login
### Description

When Perses is deployed with a non-empty `api_prefix` (e.g., `/perses`), the default redirect path used after a successful OIDC/OAuth login ignores `api_prefix` and falls back to `/`. This causes the browser to be redirected to the root path of the host (e.g., `https://example.com/`), which is typically not served by Perses in a sub-path deployment, resulting in a 404 or landing on an unrelated application's page.

This is a standalone bug. It is exposed in combination with two other bugs (see "Related issues" below), but the root cause is this single default value is independent of them.

### Environment

- Perses version: v0.54.0
- Deployment: Kubernetes with Ingress, sub-path deployment (`api_prefix: /perses`)

### Root cause

`ui/app/src/model/auth/auth-client.ts:28-32`

```typescript
export function useRedirectQueryParam(): string {
  const [path] = useQueryParam<string | undefined>(redirectQueryParam);
  return path ?? '/';
}
```

When a user visits `https://example.com/perses` (the Perses landing page), there is no `rd` query parameter in the URL. `path` is `undefined`, so the function returns the hardcoded default `'/'`.

This default value is then passed to `buildRedirectQueryString(path)` in `SignWrapper.tsx:197`:

```typescript
window.location.href = `/api/auth/providers/${provider.path}/login?${buildRedirectQueryString(path)}`;
//                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// path = '/', so the query becomes: ?rd=%2F
```

The Perses backend receives `rd=/` and encodes it into the OAuth `state`:

```go
// internal/api/impl/auth/oidc.go:266-267
redirectPath := ctx.Request().URL.Query().Get(redirectQueryParam)  // "rd" → "/"
return encodeOAuthState(redirectPath)  // state = "random--/"
```

After the OIDC callback, Perses decodes the state and redirects:

```go
// internal/api/impl/auth/oidc.go:284
redirectURI := decodeOAuthState(state)  // ""
http.Redirect(w, r, redirectURI, http.StatusFound)  // Location: /
```

The browser follows the redirect to `https://example.com/` — the root path of the host, not the Perses sub-path. In a shared-host deployment (multiple apps behind the same Ingress), this lands on a different application or not Perses.

### Expected behavior

The default redirect path should respect `api_prefix`. The application already exposes `api_prefix` to the frontend at runtime via `PERSES_APP_CONFIG.api_prefix` (injected by the backend through `PREFIX_PATH_PLACEHOLDER` replacement in `ui/endpoint.go`), and other call sites already use it:

```typescript
// ui/app/src/components/Header/AccountMenu.tsx:78 — logout, correctly uses api_prefix ✓
<MenuItem component="a" href={`${PERSES_APP_CONFIG.api_prefix}/api/auth/logout`}>
```

The fix should use `api_prefix` as the default:

```typescript
export function useRedirectQueryParam(): string {
  const [path] = useQueryParam<string | undefined>(redirectQueryParam);
  return path ?? PERSES_APP_CONFIG.api_prefix;
}
```

### Impact

Any user that directly visits the Perses landing page (without a `rd` query parameter) and logs in via OIDC/OAuth will be redirected to the host root `/` instead of `/perses`. In a sub-path deployment, this either:

- 404s (if the host root is not served by Perses)
- Lands on a different application sharing the same host

### Related issues

This bug is part of a chain of three independent bugs that together break the OIDC login flow when `api_prefix` is set. Each bug is independent and should be fixed separately:

1. **#4431** — `SignWrapper.tsx:197` hardcodes `/api/auth/providers/...` without `api_prefix` (causes `HandleAPIPrefix` 302)
2. **#4435** — `HandleAPIPrefix` drops query string on 302 redirect — `apiprefix.go:27` only concatenates `URL.Path`, losing `URL.RawQuery` (causes `rd` parameter to be lost)
3. **This issue** — `auth-client.ts:31` defaults `rd` to `/` instead of `api_prefix` (causes redirect to root path `/`)

Fixing any three bugs together fully resolves the OIDC login flow in a sub-path deployment. Fixing this bug alone resolves the redirect-to-root issue regardless of the other two.

_No response_

### System information

_No response_

### Perses version

v0.54.0

### Perses configuration file

```yaml

```

### Logs

```text

```

pujitha24:
# Description

When Perses is deployed with a non-empty `api_prefix` (e.g. `/perses`), a user who visits the Perses landing page directly (no `rd` query param in the URL) and then signs in via an OIDC/OAuth provider button is redirected, after a successful login, to the host root `/` instead of the sub-path `/perses`. In a shared-host deployment this either 404s or lands on a different application.

Root cause: `useRedirectQueryParam()` in `ui/app/src/model/auth/auth-client.ts` hardcoded its fallback (used when there's no `rd` query param) to `'/'`. `SignWrapper.tsx` feeds that value into the `rd` query param sent to the backend's OIDC/OAuth login endpoint; the backend later issues a raw `http.Redirect` to that literal path once the OAuth flow completes, so it must be an absolute, `api_prefix`-inclusive path to land back inside the app.

`useRedirectQueryParam()` is also used by three React Router-based consumers (`SignInView.tsx`, `Router.tsx`'s `AlreadyLoggedIn`, `DelegatedAuthnErrorView.tsx`) via `navigate()`/`<Navigate>`. Those go through a router created with `api_prefix` as its `basename`, which already prepends `api_prefix` automatically — so simply changing the hook's shared default to `api_prefix` would fix `SignWrapper.tsx` but double-prefix those three (e.g. `/perses/perses`).

# Approach

Added an optional `{ absolute?: boolean }` parameter to `useRedirectQueryParam()`. When there's no `rd` query param:
- by default (`absolute` unset/false — used by the three React Router consumers, unchanged call sites), the fallback stays `'/'`, exactly matching prior behavior;
- with `absolute: true` (used only by `SignWrapper.tsx`, which redirects outside of React Router), the fallback is `PERSES_APP_CONFIG.api_prefix || '/'`, so it defaults back to `'/'` for the common no-`api_prefix` deployment too.

Only `ui/app/src/views/auth/SignWrapper.tsx` was updated to opt into `{ absolute: true }`; the other three call sites are untouched.

# Validation

Ran from `ui/`:
- `npx turbo run test --filter=@perses-dev/app` — all 152 tests pass (14 files), including new/updated tests in `ui/app/src/model/auth/auth-client.test.ts` (new — covers all four fallback branches: `rd` present, `rd` absent with `absolute` false/true, and `absolute: true` with an empty `api_prefix`) and `ui/app/src/views/auth/SignWrapper.test.tsx` (updated — asserts the OIDC login redirect URL now carries `rd=%2Fperses` instead of `rd=%2F` for a `/perses` `api_prefix` deployment).
- `npx turbo run type-check --filter=@perses-dev/app` — passes.
- `npx turbo run lint --filter=@perses-dev/app` — 0 errors (repo-wide pre-existing warnings unrelated to this change; confirmed via `git stash` that the touched files' warning set is identical before and after this diff).
- `npx oxfmt --check` on the changed files — formatted correctly.
- `npm run build` (full `ui/` workspace) — succeeds.

# Screenshots

No UI-visible change; this only affects the URL a browser is redirected to after an OIDC/OAuth login when no `rd` query param was present.

# Checklist

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes. (N/A — no visual change, only the redirect target URL in one edge case.)

Fixes #4436

### DIFF
--- a/ui/app/src/model/auth/auth-client.ts
+++ b/ui/app/src/model/auth/auth-client.ts
@@ -28,15 +28,16 @@ const redirectQueryParam = 'rd';
  *
  * By default, the fallback path (used when there is no `rd` query param) is relative to the app's `api_prefix`,
  * which is what React Router's `navigate()`/`<Navigate>` expect since the router is created with `api_prefix` as
- * its basename. Pass `absolute: true` when the path is instead going to be used outside of React Router (e.g. a
- * raw `window.location.href` redirect to the backend), where the fallback must include `api_prefix` itself.
+ * its basename. Pass `true` when the path is instead going to be used outside of React Router (e.g. a raw
+ * `window.location.href` redirect to the backend), where the fallback must include `api_prefix` itself.
+ * @param absolute whether the fallback path must be absolute (include `api_prefix`) rather than router-relative.
  */
-export function useRedirectQueryParam(options: { absolute?: boolean } = {}): string {
+export function useRedirectQueryParam(absolute?: boolean): string {
   const [path] = useQueryParam<string | undefined>(redirectQueryParam);
   if (path !== undefined) {
     return path;
   }
-  return options.absolute ? PERSES_APP_CONFIG.api_prefix || '/' : '/';
+  return absolute ? PERSES_APP_CONFIG.api_prefix || '/' : '/';
 }
 
 /**
--- a/ui/app/src/model/auth/auth-client.ts
+++ b/ui/app/src/model/auth/auth-client.ts
@@ -14,6 +14,7 @@
 import { fetch } from '@perses-dev/client';
 import { useQueryParam } from 'use-query-params';
 
+import { PERSES_APP_CONFIG } from '../../config';
 import { HTTPHeader, HTTPMethodPOST } from '../http';
 import buildURL from '../url-builder';
 import { useCurrentUser } from '../user-client';
@@ -24,10 +25,18 @@ const redirectQueryParam = 'rd';
 /**
  * Get the redirect path from URL's query params.
  * This is used to retrieve the original path that a user desired before being redirected to the login page.
+ *
+ * By default, the fallback path (used when there is no `rd` query param) is relative to the app's `api_prefix`,
+ * which is what React Router's `navigate()`/`<Navigate>` expect since the router is created with `api_prefix` as
+ * its basename. Pass `absolute: true` when the path is instead going to be used outside of React Router (e.g. a
+ * raw `window.location.href` redirect to the backend), where the fallback must include `api_prefix` itself.
  */
-export function useRedirectQueryParam(): string {
+export function useRedirectQueryParam(options: { absolute?: boolean } = {}): string {
   const [path] = useQueryParam<string | undefined>(redirectQueryParam);
-  return path ?? '/';
+  if (path !== undefined) {
+    return path;
+  }
+  return options.absolute ? PERSES_APP_CONFIG.api_prefix || '/' : '/';
 }
 
 /**
--- a/ui/app/src/model/auth/auth-client.test.ts
+++ b/ui/app/src/model/auth/auth-client.test.ts
@@ -1,0 +1,75 @@
+// Copyright The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { renderHook } from '@testing-library/react';
+
+import { useRedirectQueryParam } from './auth-client';
+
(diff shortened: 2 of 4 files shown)


ui/app/src/model/auth/auth-client.ts
```diff
@@ -24,10 +25,18 @@ const redirectQueryParam = 'rd';
...
  * Get the redirect path from URL's query params.
  * This is used to retrieve the original path that a user desired before being redirected to the login page.
+ *
+ * By default, the fallback path (used when there is no `rd` query param) is relative to the app's `api_prefix`,
+ * which is what React Router's `navigate()`/`<Navigate>` expect since the router is created with `api_prefix` as
+ * its basename. Pass `absolute: true` when the path is instead going to be used outside of React Router (e.g. a
+ * raw `window.location.href` redirect to the backend), where the fallback must include `api_prefix` itself.
  */
-export function useRedirectQueryParam(): string {
+export function useRedirectQueryParam(options: { absolute?: boolean } = {}): string {
```
Nexucis: I think the parameter can be simplified.

```suggestion
export function useRedirectQueryParam(absolute?: boolean): string {
```

Nexucis (approved):
thank you @pujitha24 for the contribution.

PR #4446, opened 2026-09-14
[BUGFIX] include dashboard tags in project view search

### LINKED ISSUE #4440: Support search by tags in project view
### Proposal

It's not possible currently to rely on the search bar within the project view to filter based on tags. The string match is only based on the dashboard name:

Image

Image

Searching among tags could be part of the same input or be a dedicated one just aside/below, to be discussed.

NB: With https://github.com/perses/perses/issues/4425 it would also be possible at the end to fulfill the need of searching by tags within a single project. Once we have the latter implemented, this issue may feel like a duplicate. Still, putting myself in the shoes of an end user, seing this searchbar in the project view I tend to go to this one to search among my dashboards, instead of going back to the top navbar, as it feels "closer" to what I want to reach.

Sarthak-Shreshtha01:
## Summary
- The dashboard list search bar within a project view (`DashboardTreeList`) filters by dashboard/folder name but silently excludes tags, because TanStack Table's default global-filter eligibility check only accepts `string`/`number` typed column values, and the `tags` column exposed a `string[]`.
- Adds a derived `tagsSearchValue` field (tags joined into a single string) that the table uses for filtering/sorting, while the tag chips still render from the original `tags` array.

Fixes #4440

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npx oxlint` on changed files — no new warnings
- [x] `npx vitest run` — 152/152 tests passing (full `app` package suite)
- [x] Manually reproduced the bug locally (search by dashboard name worked, search by tag returned nothing) and confirmed the fix resolves it


Sarthak-Shreshtha01:
Fixed in 9265cbae. Nested subfolders (a folder inside another folder) go through a separate mapping function (`mapFolderItemsToTableRow`'s `case 'Folder':`), which was missing `tagsSearchValue`, leaving it `undefined` at runtime for that row shape. Added the missing field plus a regression test asserting `tagsSearchValue` is always a string for nested folder rows.

Nexucis:
Hello @Sarthak-Shreshtha01 thanks for the PR. Can you provide a screen shot showing in the UI the changes you are proposing ?

adrianSepiol:
@Sarthak-Shreshtha01 what do you think about implementing getColumnCanGlobalFilter in "Table" component instead. We use search @tanstack/match-sorter-utils which is fork of https://github.com/kentcdodds/match-sorter and in the Readme there is this mentioned: "Array of values: When the specified key matches an array of values, the best match from the values of in the array is going to be used for the ranking.". The issue right now is that tanstack table calls filter function only on columns that have string or number value but we can safely call it on arrays.

Sarthak-Shreshtha01:
Hi @Nexucis ! Thanks for the review !
Screenshots attached — showing the tagged dashboard, the search failing to find it before the fix, and the same search working after.

Screenshot 2026-09-15 143045
Screenshot 2026-09-15 143052
Below is after the Fix:-
Screenshot 2026-09-15 143307

Sarthak-Shreshtha01:
@adrianSepiol 
That did cross my mind, and you're right that the fuzzy matcher itself already handles arrays fine (match-sorter's ranking) — the actual gate is TanStack's default `getColumnCanGlobalFilter`, which only allows string/number before the matcher ever runs.

I kept this PR scoped to just the reported bug though, since `Table` lives in `perses/shared`, a separate repo, and touching it would pull in a bigger, cross-repo change for what's meant to be a small, focused fix. I'll open a follow-up issue on `perses/shared` for the `getColumnCanGlobalFilter` change.

Sarthak-Shreshtha01:
Hi @Nexucis , just a quick follow-up on this PR.

I've addressed the review feedback, including the nested-folder "tagsSearchValue" case and the tag-search test coverage, and also added the requested UI screenshots.

The current checks are passing, and the latest review only has a non-blocking test-coverage nit remaining.

Whenever you get a chance, could you please take another look and let me know if anything else is needed from my side? Thanks!

PR #4449, opened 2026-09-15
[BUGFIX] Fix drop of query parameter when redirection is running with api_prefix

### LINKED ISSUE #4435: HandleAPIPrefix middleware drops query string on 302 redirect
### Description

When `api_prefix` is configured (e.g., `api_prefix: /perses`), the `HandleAPIPrefix` middleware redirects requests that lack the prefix via HTTP 302. However, the redirect URL is built using only `c.Request().URL.Path`, omitting `c.Request().URL.RawQuery`. This causes any query parameters on the original request to be silently lost.

This is a general issue that affects **any** request arriving without the `api_prefix` — not just the OIDC login flow described below.

This issue is related to #4431, where the OIDC/OAuth login redirect URL is hardcoded without `api_prefix`. That bug causes requests to arrive without the `api_prefix`, which triggers the `HandleAPIPrefix` middleware and exposes the query string loss described here.

### Environment

- Perses version: v0.54.0
- Deployment: Kubernetes with Ingress, sub-path deployment (`api_prefix: /perses`)

### Root cause

`internal/api/core/middleware/apiprefix.go`

```go
func HandleAPIPrefix(apiPrefix string) echo.MiddlewareFunc {
    return func(next echo.HandlerFunc) echo.HandlerFunc {
        return func(c echo.Context) error {
            if !strings.HasPrefix(c.Request().URL.Path, apiPrefix) {
                return c.Redirect(http.StatusFound, apiPrefix+c.Request().URL.Path)
                //                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
                // Only Path is concatenated; RawQuery is dropped
            }
            return next(c)
        }
    }
}
```

`c.Request().URL.Path` contains only the path portion of the URL (e.g., `/api/auth/providers/oidc/keycloak/login`). The query string (e.g., `?rd=%2F`) is available in `c.Request().URL.RawQuery` but is never appended to the redirect target.

Echo's `c.Redirect(code, url)` simply sets the `Location` header to the provided `url` string — it does not automatically preserve or re-attach query parameters from the original request.

### Impact: OIDC login flow broken (real-world example)

This bug, combined with the hardcoded login URL in `SignWrapper.tsx` (which omits `api_prefix`), breaks the entire OIDC/OAuth login flow when `api_prefix` is set.

#### Step-by-step trace

1. **User clicks "Sign in with Keycloak"**

   The frontend (`SignWrapper.tsx:197`) redirects the browser with a hardcoded path (no `api_prefix`):
   ```
   GET /api/auth/providers/oidc/keycloak/login?rd=%2F
   ```
   Note: `rd=%2F` carries the redirect path the user should land on after login.

2. **HandleAPIPrefix middleware fires 302**

   The request path `/api/auth/...` does not start with `/perses`, so the middleware redirects:
   ```
   302 Location: /perses/api/auth/providers/oidc/keycloak/login
   ```
   **The `?rd=%2F` query string is lost.**

3. **Perses processes login with empty `rd`**

   ```go
   redirectPath := ctx.Request().URL.Query().Get(redirectQueryParam) // "rd" → "" (empty)
   return encodeOAuthState(redirectPath) // state = "random--"
   ```
   The OAuth `state` parameter becomes `random--` (empty redirect path after `--`).

4. **Keycloak authenticates the user and calls back**

   ```
   302 Location: /perses/api/auth/providers/oidc/keycloak/callback?state=random--&code=xxx
   ```

5. **Perses decodes state and redirects to empty URL**

   ```go
   redirectURI := decodeOAuthState("random--") // returns "" (empty string)
   http.Redirect(w, r, redirectURI, http.StatusFound) // Location: (empty)
   ```

6. **Browser resolves empty redirect to parent directory**

   The browser interprets an empty `Location` header relative to the current URL:
   ```
   Current: /perses/api/auth/providers/oidc/keycloak/callback
   Redirect: /perses/api/auth/providers/oidc/keycloak/  (parent directory)
   ```
   Result: **404 Not Found**

#### Other affected scenarios

Any request that arrives without the `api_prefix` and carries query parameters will lose them on redirect. For example:

- Direct browser access: `https://example.com/api/v1/projects?name=zhg` → redirects to `/perses/api/v1/projects` (loses `?name=zhg`)
- Third-party callbacks with query parameters
- percli calls without `api_prefix` configured

### Expected behavior

The `HandleAPIPrefix` middleware should preserve query parameters when redirecting:

```go
func HandleAPIPrefix(apiPrefix string) echo.MiddlewareFunc {
    return func(next echo.HandlerFunc) echo.HandlerFunc {
        return func(c echo.Context) error {
            if !strings.HasPrefix(c.Request().URL.Path, apiPrefix) {
                target := apiPrefix + c.Request().URL.Path
                if c.Request().URL.RawQuery != "" {
                    target += "?" + c.Request().URL.RawQuery
                }
                return c.Redirect(http.StatusFound, target)
            }
            return next(c)
        }
    }
}
```

### Current workaround

Ensure all client-side URLs include the `api_prefix` so that `HandleAPIPrefix` is never triggered. For the OIDC login specifically, this also requires fixing `SignWrapper.tsx:197` to prepend `PERSES_APP_CONFIG.api_prefix` to the login URL.

_No response_

### System information

_No response_

### Perses version

Perses version: v0.54.0

### Perses configuration file

```yaml

```

### Logs

```text

```

Nexucis:
fix https://github.com/perses/perses/issues/4435 based on the suggestion written in the issue.


AntoineThebaud (approved):
🚀

PR #4454, opened 2026-09-15
[IGNORE] Generate frontend SBOM on each PR + fix peer deps

celian-garcia:
# Description

## New CI check
During a release, we had an issue that could have been detected before. This is on the CycloneDX library call, responsible of generating the SBOM for the release 
https://github.com/perses/perses/actions/runs/34983935316/job/104431071264#step:10:23
```
Run npx @cyclonedx/cyclonedx-npm --output-file ../ui-sbom.cdx.json --output-format JSON
  npx @cyclonedx/cyclonedx-npm --output-file ../ui-sbom.cdx.json --output-format JSON
  shell: /usr/bin/bash -e {0}
  env:
    NPM_CONFIG_USERCONFIG: /home/runner/work/_temp/.npmrc
npm warn exec The following package was not found and will be installed: @cyclonedx/cyclonedx-npm@6.0.1
npm warn deprecated prebuild-install@7.1.3: No longer maintained. Please contact the author of the relevant native addon; alternatives are available.
npm warn deprecated glob@10.5.0: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
  Command failed: /opt/hostedtoolcache/node/24.19.0/x64/bin/node -- /opt/hostedtoolcache/node/24.19.0/x64/lib/node_modules/npm/bin/npm-cli.js ls --json --long --all
  npm error code ELSPROBLEMS
  npm error invalid: ajv@6.14.0 /home/runner/work/perses/perses/ui/node_modules/ajv
  npm error invalid: typescript@7.0.2 /home/runner/work/perses/perses/ui/node_modules/typescript
  npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2026-09-15T14_47_45_071Z-debug-0.log
  
  npm error code ELSPROBLEMS
  npm error invalid: ajv@6.14.0 /home/runner/work/perses/perses/ui/node_modules/ajv
  npm error invalid: typescript@7.0.2 /home/runner/work/perses/perses/ui/node_modules/typescript
  npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2026-09-15T14_47_45_071Z-debug-0.log
```

So I am running this CI step on each PR to be sure we fail fast on that.

## About the fix

AI helped me on identifying the root cause and the fix. Here is the summary: 
[...]

PR #4459, opened 2026-09-16
[IGNORE] Activate sbom generation in CI all the time

Nexucis:


Nexucis:
closing in favor of #4454 that reached the same conclusion about what needs to be done

PR #4465, opened 2026-09-16
[FEATURE] (ui): set browser tab title from dashboard display name

colivi:
# Description

Browser tab should show the dashboard name like Grafana, to navigate multi-tab investigations.

Sets "document.title" to "{displayName} · {project} | Perses" while a dashboard is open ("getResourceDisplayName" + project).

Restores the previous title on leave.

No API change — uses existing dashboard resource fields.

# Checklist
- [x] Descriptive title ("[FEATURE] …")
- [x] DCO signoff


Gladorme (commented):
I wanted to use React 19 native feature for changing metadata page info, like title. But the upgrade is still not planned, if really necessary, I think we should use: https://github.com/staylor/react-helmet-async and implement it on all pages and if possible implement it in the router, instead of components. Not sure if possible to be implem in router, but it would be the best

Nexucis:
Hi @colivi,

Thank you for helping us around this issue. @Gladorme has opened a PR that is covering the complete routing not just the dashboard page. Hopefully you should be satisfied with what has been proposed.

Thanks again. I am closing this PR in favour of the one opened by @Gladorme. 

Looking forward to the others contributions you have made or in the future.
Cheers;

PR #4468, opened 2026-09-17
[ENHANCEMENT] Support custom login property for the OIDC provider

### LINKED ISSUE #4456: Support `preferred_username` claim as login username in OIDC provider
### Summary

The OIDC provider currently hardcodes the login username extraction to use the email address (falling back to `subject`). This is inconsistent with the OAuth provider, which supports a `custom_login_property` configuration to flexibly select the username field. The `preferred_username` claim — a standard OIDC `profile` scope claim — is available in the `oidcUserInfo` struct but is never used by `GetLogin()`.

### Current Behavior

**OIDC provider** (`internal/api/impl/auth/oidc.go:58-63`):

```go
func (u *oidcUserInfo) GetLogin() string {
    login := buildLoginFromEmail(u.Email)
    if len(login) > 0 {
        return login
    }
    return u.Subject
}
```

The login is derived from the email address (`user@domain.com` → `user`), falling back to the OIDC `subject` claim. There is no way to configure an alternative field such as `preferred_username`.

**OAuth provider** (`internal/api/impl/auth/oauth.go:106-111`):

```go
func (u *oauthUserInfo) GetLogin() string {
    if login := u.getProperty(u.loginKeys); login != "" {
        return login
    }
    return buildLoginFromEmail(u.Email)
}
```

The OAuth provider supports `custom_login_property` (e.g., `preferred_username`) via the `OAuthProvider` config struct (`config/authentication.go:165`):

```go
CustomLoginProperty string `json:"custom_login_property,omitempty" yaml:"custom_login_property,omitempty"`
```

### Problem

When using OIDC with Keycloak (or any OIDC IdP), the `preferred_username` claim is a standard claim returned by the userinfo endpoint under the `profile` scope. However, the OIDC provider ignores it and uses the email prefix as the login name.

This causes several issues:

1. **Inconsistent usernames**: If a user's email is `john.doe@example.com` but their preferred username is `jdoe`, the OIDC provider creates a user named `john.doe` instead of `jdoe`.
2. **Email dependency**: Users without an email claim in their token cannot log in via OIDC (falls back to `subject`, which is often a UUID).
3. **Inconsistency between OIDC and OAuth**: The OAuth provider supports `custom_login_property`, but OIDC does not offer an equivalent option.
4. **Migration friction**: When migrating from OAuth to OIDC (or vice versa), users get different login names, breaking existing `GlobalRoleBinding`/`RoleBinding` references.

### Proposed Solution

**Option A**: Add `custom_login_property` field to `OIDCProvider` config struct, mirroring the OAuth provider. This allows users to specify any userinfo field as the login name.

```go
// pkg/model/api/config/authentication.go
type OIDCProvider struct {
    Provider            `json:",inline" yaml:",inline"`
    Issuer              common.URL        `json:"issuer" yaml:"issuer"`
    DiscoveryURL        common.URL        `json:"discovery_url,omitempty" yaml:"discovery_url,omitempty"`
    URLParams           map[string]string `json:"url_params,omitempty" yaml:"url_params,omitempty"`
    DisablePKCE         bool              `json:"disable_pkce" yaml:"disable_pkce"`
    Logout              OIDCLogout        `json:"logout" yaml:"logout"`
    CustomLoginProperty string            `json:"custom_login_property,omitempty" yaml:"custom_login_property,omitempty"` // NEW
}
```

Update `GetLogin()` to check the custom property first, then fall back to email and subject:

```go
func (u *oidcUserInfo) GetLogin() string {
    if u.loginProperty != "" && u.RawProperties[u.loginProperty] != nil {
        if login, ok := u.RawProperties[u.loginProperty].(string); ok && login != "" {
            return login
        }
    }
    login := buildLoginFromEmail(u.Email)
    if len(login) > 0 {
        return login
    }
    return u.Subject
}
```

**Option B**: Use `preferred_username` by default. The `PreferredUsername` field already exists in `externalUserInfoProfile` (`internal/api/impl/auth/userinfo.go:32`) and is already populated from the OIDC userinfo response — it's just not used by `GetLogin()`.

```go
type externalUserInfoProfile struct {
    // ...
    PreferredUsername string `json:"preferred_username,omitempty"`  // already parsed!
    Email             string `json:"email,omitempty"`
}
```

Simply update `GetLogin()` to check `PreferredUsername` before email:

```go
func (u *oidcUserInfo) GetLogin() string {
    if u.PreferredUsername != "" {
        return u.PreferredUsername
    }
    login := buildLoginFromEmail(u.Email)
    if len(login) > 0 {
        return login
    }
    return u.Subject
}
```

Either approach would make OIDC behavior consistent with the OAuth provider.

### Impact

- **OIDC users**: Can use `preferred_username` as login name, consistent with OAuth provider behavior
- **No breaking change**: Default behavior (email → subject) is preserved when `custom_login_property` is not set
- **Migration**: Users switching between OIDC and OAuth providers get consistent login names

### Environment

- Perses: v0.54.0
- Reproducible on `main` branch (OIDC `GetLogin()` unchanged since v0.54.0)

pujitha24:
# Description

The OIDC provider's `GetLogin()` always derives the login name from the email
address (falling back to `subject`), even though the standard OIDC
`preferred_username` claim is already parsed into
`externalUserInfoProfile.PreferredUsername` and simply unused. The OAuth
provider already supports selecting the login field via
`custom_login_property`; OIDC had no equivalent.

This PR adds a `custom_login_property` option to the OIDC provider
configuration, mirroring the existing OAuth provider option. When set to one
of the userinfo fields Perses already parses (`name`, `given_name`,
`family_name`, `middle_name`, `nickname`, `preferred_username`, `email`), and
that field is non-empty in the userinfo response, it is used as the login.
Otherwise, the behavior is unchanged: email prefix, then subject.

This is opt-in and defaults to empty, so existing deployments keep their
current login names unless they explicitly set the new option.

# Screenshots

N/A (backend-only, no UI changes).

# Checklist

## UI Changes

Fixes #4456


kubehubdev:
ah, I have this requirement as well, I have an implementation with go-template. like ``` "{{.claims.email}}-{{.oidcProvider}}"```. but we can converge to you implementation. 

couple things I noticed:
1. perses have strong limitation on username, `@` is not accepted, so you will need to do some transform on email.
2. there could be multiple oidcProvider configured, and with this go-template, user have the power to tight up RBAC based on different IDP. i.e. same email loggin through github oidc and company oidc.

pujitha24:
Good points. On #1: this mirrors what the OAuth provider already does today — `getProperty`/`getProfileProperty` return the raw claim value with no sanitization, so setting `custom_login_property` to `email` would hit the same `@`-in-username validation issue on the existing OAuth provider too. It's not something this PR introduces, and I'd rather not bolt on ad-hoc sanitization just for OIDC when OAuth has the identical gap — feels like it should be fixed (if at all) for both providers together, as a separate change. On #2, the go-template / per-provider approach sounds more flexible, but that's a bigger scope change than what this PR does (just bringing OIDC to parity with the existing `custom_login_property` mechanism on OAuth) — happy to see that discussed as its own issue if there's interest.

pkg/model/api/config/authentication.go
```diff
@@ -200,6 +200,11 @@ type OIDCProvider struct {
 	URLParams    map[string]string `json:"url_params,omitempty" yaml:"url_params,omitempty"`
 	DisablePKCE  bool              `json:"disable_pkce" yaml:"disable_pkce"`
 	Logout       OIDCLogout        `json:"logout" yaml:"logout"`
+	// CustomLoginProperty is the name of the userinfo property to use as the "login" of the user
+	// (e.g. "preferred_username"). It must be one of the properties Perses already extracts from
+	// the userinfo response. If not set or not found in the response, it falls back to the email,
+	// then to the subject.
+	CustomLoginProperty string `json:"custom_login_property,omitempty" yaml:"custom_login_property,omitempty"`
```
Nexucis: since it is an enum, we can reflect it as such.

It will also avoid putting a random string here.

internal/api/impl/auth/oidc.go
```diff
@@ -46,16 +46,47 @@ type oidcUserInfo struct {
...
+	case "family_name":
+		return u.FamilyName
+	case "middle_name":
+		return u.MiddleName
+	case "nickname":
+		return u.Nickname
+	case "preferred_username":
+		return u.PreferredUsername
+	case "email":
+		return u.Email
```
Nexucis:
```suggestion
		return buildLoginFromEmail(u.Email)
```

internal/api/impl/auth/oidc.go
```diff
@@ -46,16 +46,47 @@ type oidcUserInfo struct {
...
+	default:
+		return ""
```
Nexucis:
```suggestion
		return buildLoginFromEmail(u.Email)
```

internal/api/impl/auth/oidc.go
```diff
@@ -46,16 +46,47 @@ type oidcUserInfo struct {
...
+// If a custom login property is configured and present in the userinfo response, it is used.
+// Otherwise, it uses the first part of the email to create the username, falling back to the subject.
 func (u *oidcUserInfo) GetLogin() string {
+	if u.loginProperty != "" {
+		if login := u.getProfileProperty(u.loginProperty); login != "" {
+			return login
+		}
+	}
 	login := buildLoginFromEmail(u.Email)
 	if len(login) > 0 {
```
Nexucis, line 88: you can actually simplify a log this function by simply doing that:

```go
func (u *oidcUserInfo) GetLogin() string {
	login := u.getProfileProperty(u.loginProperty);
	if len(login) > 0 {
		return login
	}
	return u.Subject
}
```

Of course this is working only if the function `getProfileProperty` is returning by default the login from the email (see my suggestion)

docs/configuration/configuration.md
```diff
@@ -292,6 +292,12 @@ logout:
   enabled: <boolean> | default = false # Optional
   # A config option to use a different query parameter for the redirect uri on logout. Some providers (e.g. Cognito) require this.
   logout_redirect_param_name: <string> | default = post_logout_redirect_uri # Optional
+
+# Name of the userinfo property to use as the "login" of the user (e.g. "preferred_username").
+# It must be one of the properties Perses already extracts from the userinfo response
+# (name, given_name, family_name, middle_name, nickname, preferred_username, email).
+# If not set, or not present in the userinfo response, it falls back to the email, then to the subject.
+custom_login_property: <string> # Optional
```
Nexucis: please align this documentation with the same space than above.

Nexucis (commented):
Hello @pujitha24,

Thank you for your contribution. I have done a little review. Regarding the question about the email, I have added few comments that will tackle it.

docs/configuration/configuration.md
```diff
@@ -292,6 +292,10 @@ logout:
   enabled: <boolean> | default = false # Optional
   # A config option to use a different query parameter for the redirect uri on logout. Some providers (e.g. Cognito) require this.
   logout_redirect_param_name: <string> | default = post_logout_redirect_uri # Optional
+
+# Name of the userinfo property to use as the "login" of the user.
+# If not set, or not present in the userinfo response, it falls back to the email, then to the subject.
+custom_login_property: < enum | possibleValue = 'name' | 'given_name' | 'family_name' | 'middle_name' | 'nickname' | 'preferred_username' | 'email' > # Optional
```
Nexucis: can you please align the space at the beginning of this new entry in the doc with the rest of the documentation ?

```suggestion
  # Name of the userinfo property to use as the "login" of the user.
  # If not set, or not present in the userinfo response, it falls back to the email, then to the subject.
  custom_login_property: < enum | possibleValue = 'name' | 'given_name' | 'family_name' | 'middle_name' | 'nickname' | 'preferred_username' | 'email' > # Optional

Nexucis (approved):

Nexucis (approved):
Thank you @pujitha24 for your contributions !

PR #4475, opened 2026-09-18
[ENHANCEMENT] Add Cache-Control headers for static UI assets

### LINKED ISSUE #4474: Feature request: add caching and content type headers for static assets
### Proposal

As of v0.53.0, it seems that assets like `main.[...].js` are served with Content-Type: text/plain, and without any cache headers. These asset names bear some sort of hash and I suspect they are immutable for a given asset name, so should be cacheable.

pujitha24:
# Description

Static JS/CSS bundles served by the embedded Go frontend server (`ui/endpoint.go`) are built with a content hash in their filename (e.g. `main.<hash>.js`, via `ui/app/rspack.config.mjs`'s `[name].[contenthash].js`), so a given filename's content never changes across a build. Despite that, `assetHandler` was not setting any `Cache-Control` header, so browsers could not cache these assets long-term and had to re-fetch them on every load.

This PR adds `Cache-Control: public, max-age=31536000, immutable` for JS/CSS assets served through `assetHandler` (reusing the existing `.js`/`.css` substring check already used there for placeholder replacement), and `Cache-Control: no-cache` for `index.html` (served by `serveASTFiles`), since it references the current hashed bundle filenames and must always be revalidated so a client doesn't keep using stale asset references after a new deploy.

# Screenshots

N/A (no UI changes, only HTTP response headers on the Go static-file server).

# Checklist

## UI Changes

Fixes #4474


Nexucis (approved):
great, thank you @pujitha24 !

